PR #105, opened 2015-12-30
Fix a bug that exception will throw if there're duplicate parameters in developer comments.

chenkennt:
@vicancy @vwxyzh


src/Microsoft.DocAsCode.EntityModel/Parsers/TripleSlashCommentParser.cs
```diff
@@ -243,7 +258,18 @@ private static List<string> GetExamples(XPathNavigator nav, ITripleSlashCommentP
                 string name = nav.GetAttribute("name", string.Empty);
                 string description = nav.Value;
                 if (context.Normalize) description = NormalizeContentFromTripleSlashComment(description);
-                if (!string.IsNullOrEmpty(name)) result.Add(name, description);
```
vicancy: code could be shared with L234?

vicancy:
:+1:

PR #130, opened 2016-01-22
Fix issue that missing cell at table last row

fenxu:
Currently if the last cell of table is empty. It will be replace with empty and then we missing a cell.
 I fix that by adding the missing cells back. I'm not sure if it is a correct method. Please have a review and tell me if it is OK.


vicancy:
LGTM. please rebase to make only one commit to dev

PR #140, opened 2016-02-04
Extract content in FileModel to a ModelWithCache class so that Apply Template can reuse the cache

vicancy:
@vwxyzh @chenkennt


vwxyzh:
:+1:

vwxyzh:
pls. add broken change in release note.

PR #142, opened 2016-02-04
apply aspiscan metadata

ansyral:
@vwxyzh


vwxyzh:
is it better to move it into plug-in?

PR #147, opened 2016-02-06
fix for docs global toc

superyyrrzz:
When building global toc in docs, `href` means the child repo's absolute path. Like this: https://github.com/kimizhu/CapsToOpsTest_EMS/tree/live/CapsToOpsTest_EMS/v1

Robin reflected that after update docfx in op pipeline, the global toc does't work because the `href` in output `toc.json` is overwritten by `homepage`.

@herohua @vicancy @zhenlan @chenkennt


openpublishoperation:
LGTM

I'm okay with the current fix. However, we'd better have a clearer definition on these values and document the behaviors. We can do this later.

herohua:
LGTM

I'm okay with the current fix. However, we'd better have a clearer definition on these values and document the behaviors. We can do this later.

src/docfx.website.themes/docs.html/toc.json.js
```diff
@@ -28,7 +28,11 @@ function transform(model, _attrs){
       }else{
         if (item.href.indexOf('~/') == 0) item.href = item.href.substring(2);
         item.href = removeExtension(item.href);
-        item.relative_path_in_depot = directory + removeExtension(item.href);
+        if (item.href.indexOf('/') == 0) {
```
vicancy: directory should be empty when item.href.indexOf('/') == 0 ?

superyyrrzz: Because the href is already an absolute path, no need to append it.

PR #152, opened 2016-02-08
Fix external reference bug and add external reference to docfx website

vicancy:
@vwxyzh @chenkennt @superyyrrzz


chenkennt:
:+1:

Issue #151, opened 2016-02-07
Building from sources fails


atifaziz:
I cloned the repo, checked out tag [1.4.2](https://github.com/dotnet/docfx/releases/tag/1.4.2) (latest release at the time of reporting this issue) and ran `build.cmd` from a regular Windows 10 Command Prompt. This was more or less per the instructions laid out in the “[Build from source code](http://dotnet.github.io/docfx/tutorial/docfx_getting_started.html#3-2-build-from-source-code)” section of your documentation but it failed with many warnings & errors.

See [buildlog.txt](https://github.com/dotnet/docfx/files/120762/buildlog.txt) for the complete build details.

vwxyzh:
It looks like: running `build.cmd` with opening solution by `Visual Studio`.
`Visual Studio` will lock some files, e.g. `project.lock.json`.

atifaziz:
@vwxyzh I never had VS open. I tried again with a `git clean -dxf` followed by `build.cmd` (all from the command-line) and got the same errors.

vicancy:
According to the log,
```
  NotFound https://www.myget.org/F/raboof/api/v3/flatcontainer/microsoft.codeanalysis.workspaces.common/index.json 351ms
  NotFound https://www.myget.org/F/raboof/api/v3/flatcontainer/microsoft.codeanalysis.visualbasic.workspaces/index.json 368ms
  NotFound https://www.myget.org/F/raboof/api/v3/flatcontainer/microsoft.codeanalysis.csharp.workspaces/index.json 363ms
  NotFound https://www.myget.org/F/raboof/api/v3/flatcontainer/microsoft.codeanalysis.visualbasic/index.json 386ms
```
These packages are not restored correctly.
Could you remove `https://www.myget.org/F/raboof/api/v3/flatcontainer` from your nuget.config temporarily, close you VS, delete all `project.lock.json` file and have a retry?

atifaziz:
@vicancy If I search the log for `microsoft.codeanalysis.workspaces.common` using…

    findstr /i microsoft.codeanalysis.workspaces.common buildlog.txt

…then I see that it does get restored eventually over another feed…

```
  GET https://www.myget.org/F/aspnetrelease/api/v2/FindPackagesById()?id='Microsoft.CodeAnalysis.Workspaces.Common'
  GET https://www.myget.org/F/raboof/api/v3/flatcontainer/microsoft.codeanalysis.workspaces.common/index.json
[...]

PR #170, opened 2016-02-24
Build script fix for doc project packages restore

atifaziz:
2e4073e05067056af185352860b72d9b818b6bb5 renamed `docs` folder to `Documentation` but `build.cmd` was still attempting to build `docs`.

P.S. Not sure if this is related to #151 but certainly is a step in the right direction.

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -94,7 +94,7 @@ EXIT /B %ERRORLEVEL%
 :RestorePackage
 
 :RestoreDnuPackage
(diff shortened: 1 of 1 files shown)


atifaziz:
I noticed this needed fixing when I saw the following error message in [a build log output](https://github.com/dotnet/docfx/files/120762/buildlog.txt):

```
The system cannot find the path specified.
```

vicancy:
Thanks @atifaziz

PR #186, opened 2016-03-02
“Does the generation contain source code files?” grammar

atifaziz:
Grammar fix for the first prompt that appears when running `docfx init`:

```
A:\test> docfx init
Is the generation contains source code files? (Default: Yes)
Choose Answer (Yes/No):
```


vicancy:
Thanks @atifaziz, merged

PR #193, opened 2016-03-04
Trim common whitespace from included code snippets

### LINKED ISSUE #192: Included code snippets end up indented too far
### Title
Included code snippets end up indented too far

### Functional impact
When code is imported from code files into DFM files using the `[!code-...` syntax,  no indenting is removed from the beginning of the line. This results in code snippets that can potentially be shifted far to the right in the rendered output.

### Minimal repro steps
Create a code file with normally-indented code:
```
using System;

namespace mynamespace
{
    class myclass
    {
        public void Main()
        {
            // <mytag>
            Console.WriteLine("information");
            // </mytag>
        }
    }
}
````
Include a snippet of this file into a DFM file:
````
[!code-csharp[mysection](mycodefile.cs?name=mytag)]
````

### Expected result
Ideally, this inclusion would result in a nice, left-aligned code snippet:
````
Console.WriteLine("information");
````

### Actual result
Instead, the snippet is rendered at the same level of indent as the original code:
````
            Console.WriteLine("information");
````
When the original code was highly indented and/or includes long lines, the rendered output is far from ideal, and can be quite difficult to understand.

godefroi:
When a set of lines are included into a DFM file, no indenting is
removed from the beginning of the line. This results in code snippets
that can potentially be shifted far to the right in the rendered
output.

This patch adds code that looks through the included lines for leading
spaces and tabs, and removes enough that any indent that is common to
all the lines is removed.

Fixes #192.

Note that this patch is UNTESTED, as I was not successful in building the project.

### DIFF
--- a/src/Microsoft.DocAsCode.Dfm/DfmRendererHelper.cs
+++ b/src/Microsoft.DocAsCode.Dfm/DfmRendererHelper.cs
@@ -25,7 +25,27 @@ namespace Microsoft.DocAsCode.Dfm
                 var name = string.IsNullOrEmpty(token.Name) ? null : $" name=\"{StringHelper.HtmlEncode(token.Name)}\"";
                 var title = string.IsNullOrEmpty(token.Title) ? null : $" title=\"{StringHelper.HtmlEncode(token.Title)}\"";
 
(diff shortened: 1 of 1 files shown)


chenkennt:
Thanks @godefroi , we already fixed this in 2077f527486d8963384db4a2217347c048e9fae7

PR #206, opened 2016-03-13
Improved Coding Standards document

jozefizso:
Fixed typos and markdown formatting in Coding Standards document.

### DIFF
--- a/Documentation/guideline/csharp_coding_standards.md
+++ b/Documentation/guideline/csharp_coding_standards.md
@@ -26,14 +26,14 @@ It does not pretend to be a tutorial on C#. It only includes a set of limitation
 Tools
 ----------------
 
-* [Resharper](http://www.jetbrains.com/resharper/) is a great 3rd party code cleanup and style tool.
-* [StyleCop](http://stylecop.codeplex.com/) analyzes C# srouce code to enforce a set of style and consistency rules and has been integrated into many 3rd party development tools such as Resharper.
-* [FxCop](http://codebox/SDLFxCop) is an application that analyzes managed code assemblies (code that targets the .NET Framework common language runtime) and reports information about the assemblies, such as possible design, localization, performance, and security improvements.
+* [ReSharper](http://www.jetbrains.com/resharper/) is a great 3rd party code cleanup and style tool.
+* [StyleCop](http://stylecop.codeplex.com/) analyzes C# source code to enforce a set of style and consistency rules and has been integrated into many 3rd party development tools such as ReSharper.
+* [FxCop](http://codebox/SDLFxCop) is an application that analyzes managed code assemblies (code that targets the .NET Framework common language runtime) and reports information about assemblies, such as possible design, localization, performance, and security improvements.
 
 Highlights of Coding Standards
 ------------------------------
 
-This section is not intended to give a summary of all the coding standards that enabled by our customized StyleCop, but to give a highlight of some rules one will possibly meet in daily coding life. It also provides some recommended however not mandatory(which means not enabled in StyleCop) coding standards.
+This section is not intended to give a summary of all the coding standards that are enabled by our customized StyleCop, but to give a highlight of some rules one will possibly meet in daily coding life. It also provides some recommended however not mandatory (which means not enabled in StyleCop) coding standards.
 
 ### File Layout (Recommended)
 Only one public class is allowed per file.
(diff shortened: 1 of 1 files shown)


Documentation/guideline/csharp_coding_standards.md
```diff
@@ -26,14 +26,14 @@ It does not pretend to be a tutorial on C#. It only includes a set of limitation
 Tools
 ----------------
 
-* [Resharper](http://www.jetbrains.com/resharper/) is a great 3rd party code cleanup and style tool.
-* [StyleCop](http://stylecop.codeplex.com/) analyzes C# srouce code to enforce a set of style and consistency rules and has been integrated into many 3rd party development tools such as Resharper.
-* [FxCop](http://codebox/SDLFxCop) is an application that analyzes managed code assemblies (code that targets the .NET Framework common language runtime) and reports information about the assemblies, such as possible design, localization, performance, and security improvements.
+* [ReSharper](http://www.jetbrains.com/resharper/) is a great 3rd party code cleanup and style tool.
+* [StyleCop](http://stylecop.codeplex.com/) analyzes C# soruce code to enforce a set of style and consistency rules and has been integrated into many 3rd party development tools such as ReSharper.
```
vicancy: source

Documentation/guideline/csharp_coding_standards.md
```diff
@@ -112,24 +118,24 @@ The class definition contains class members in the following order, from *less*
...
 
-* **DO** use `Enviroment.NewLine` instead of hard-coding the line break instead of `\r\n`, as Windows uses `\r\n` and OSX/Linux uses `\n`.
+* **DO** use `Enviroment.NewLine` instead of hard-coding the line break, as Windows uses `\r\n` and OSX/Linux uses `\n`.
 
-> **Note**
->
-> Be aware that thes line-endings may cause problems in code when using `@""` text blocks with line breaks.
+  > **Note**
+  >
+  > Be aware that thes line-endings may cause problems in code when using `@""` text blocks with line breaks (it spans over multiple lines).
```
vicancy: (it spans over multiple lines) is not accurate. `@""` text blocks with line breaks may cause problems under different OS with different git crlf config. e.g. under Windows, files may be checked out with CRLF line endings, thus, the line breaks in `@""` change to `\r\n`.

Documentation/guideline/csharp_coding_standards.md
jozefizso: Can you clarify what's wrong with this code? Thanks.

Documentation/guideline/csharp_coding_standards.md
jozefizso: I understood the **`@""` text blocks with line breaks** as a string literal that contains hard-coded   newline character, eg. `@"text\ntext"`.

I wanted to also point out a case when `@""` is used in a multiline string literal, like this:

```csharp
var x = @"line1
line2";

Assert.AreEqual("line1\nline2", x);
```

That's why I used the "spans over multiple lines" expression to explicitly denote this case.

Is this a wrong way to call it like that?

Documentation/guideline/csharp_coding_standards.md
MarzinZ: L30: analyzes C# srouce code -> source

jozefizso: I am blind to mine own typos :)

Documentation/guideline/csharp_coding_standards.md
vicancy: The following is exactly the one that may cause problems, as under Windows, `x` may equal to `"line1\r\nline2"`. However I feel "spans over multiple lines" does not point out the issue explictly.
```
var x = @"line1
line2";

Assert.AreEqual("line1\nline2", x);
```

jozefizso: Well, the code example is definitely the best way to express it :)

Documentation/guideline/csharp_coding_standards.md
```diff
@@ -112,24 +118,32 @@ The class definition contains class members in the following order, from *less*
...
+  > **Note**
+  >
+  > Be aware that these line-endings may cause problems in code when using `@""` text blocks with line breaks.
+  >
+  > ```cs
+  > var x = @"line1
+  > line2";
+  >
+  > Assert.AreEqual("line1\nline2", x); // may fail on Windows when checkout uses CRLF line endings
+  > Assert.AreEqual($"line1{Environment.NewLine}line2", x);
```
vicancy: `Environment.NewLine` does not solve the issue. `Environment.NewLine` is always `\r\n` on Windows, however the line endings of the file is configured by GIT config.
So if `@""` is used, the safest way is to assert: 
`Assert.AreEqual("line1\nline2", x.Replace("\r\n", "\n"));`
Or also use `@""`, which is at most cases keep the same line ending with the above `@""` in the same file:
```
Assert.AreEqual(@"
...", x);
```

vicancy:
:+1: LGTM. Thanks @jozefizso for your help!

PR #233, opened 2016-03-22
Add introduction to overwrite files

vicancy:
@chenkennt @vwxyzh @superyyrrzz @qinezh

### DIFF
--- a/Documentation/tutorial/intro_overwrite_files.md
+++ b/Documentation/tutorial/intro_overwrite_files.md
@@ -22,6 +22,7 @@ The format of *Overwrite File*s
 ```md
 ---
 uid: microsoft.com/docfx/Contacts
+some_property: value
 ---
 Further description for `microsoft.com/docfx/Contacts`
 ```
@@ -55,7 +56,7 @@ uid: microsoft.com/docfx/Contacts
 footer: <p>Footer for <code>microsoft.com/docfx/Contacts</code></p>
 ```
 
-`uid` for a *Overwrite Model* stands for the Unique IDentifier of the *Model* it will overwrite. So it is allowed to have multiple *Overwrite Section*s with YAML Header containing the same `uid`. For one *Overwrite File*, the latter *Overwrite Section* overwrites the former one with the same `uid`. For different *Overwrite File*s, the order of overwrite is **Undetermined**. So it is suggested to have *Overwrite Sections* with the same `uid` in the same *Overwrite File*.
+`uid` for an *Overwrite Model* stands for the Unique IDentifier of the *Model* it will overwrite. So it is allowed to have multiple *Overwrite Section*s with YAML Header containing the same `uid`. For one *Overwrite File*, the latter *Overwrite Section* overwrites the former one with the same `uid`. For different *Overwrite File*s, the order of overwrite is **Undetermined**. So it is suggested to have *Overwrite Sections* with the same `uid` in the same *Overwrite File*.
 
 When processing *Conceptual File*s and *Metadata File*s, *Overwrite Model*s with the same `uid` are applied to the processed *Model*s. Different *Model*s have different overwrite principles, [Overwrite principles](#overwrite-principles) section describes the them in detail.
 
@@ -99,8 +100,6 @@ Key | Type | Overwrite behavior
 Key | Type | Overwrite behavior
 --- | --- | ---
 *uid* | string | Key
-
-#### REST API item model
 
 ### Conceptual model
 Key | Type | Overwrite behavior
--- a/Documentation/tutorial/intro_overwrite_files.md
+++ b/Documentation/tutorial/intro_overwrite_files.md
@@ -1,0 +1,110 @@
+Overwrite Files
+===============================
+
+Introduction
+--------------
+DocFX supports processing Markdown files, as well as structured data model in YAML or JSON format.
+
+We call Markdown files *Conceptual File*s, and the structured data model files *Metadata File*s.
+
+Current supported *Metadata File*s include:
+1. YAML files presenting managed reference model following [Metadata Format for .NET Languages](../spec/metadata_dotnet_spec.md).
+2. Swagger JSON files presenting Swagger REST API model following [Swagger Specification Version 2.0](http://swagger.io/specification).
+
+Inside DocFX, both *Conceptual File*s and *Metadata File*s are represented as *Model*s with different properties. Details on *Model* structure for these files are described in [Data model inside DocFX](#data-model-inside-docfx) section.
+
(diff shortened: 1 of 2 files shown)


Documentation/tutorial/intro_overwrite_files.md
```diff
@@ -0,0 +1,110 @@
...
+Further description for `microsoft.com/docfx/Contacts`
+```
+
+Each *Overwrite Section* is transformed to *Overwrite Model* inside DocFX. For the above example, the *Overwrite Model* represented in YAML format is:
+
+```yaml
+uid: microsoft.com/docfx/Contacts
+some_property: value
+conceptual: <p><b>Content</b> in Markdown</p>
+```
```
sergey-vershinin: I would synchronize the overwrite file content with the resulting overwrite model. So the following .md section:
```md
---
uid: microsoft.com/docfx/Contacts
some_property: value
---
**Content** in Markdown
``` 
will be transformed to the YAML, you provided. However, I'm not sure that `**Content** in Markdown` will be in `conceptual` property. Please correct me if needed.

Documentation/tutorial/intro_overwrite_files.md
```diff
@@ -0,0 +1,110 @@
...
+---
+Footer for `microsoft.com/docfx/Contacts`
+```
+
+In the above example, the value for `*content` is `<p>Footer for <code>microsoft.com/docfx/Contacts</code></p>`, and the *Overwrite Model* represented in YAML format is:
+
+```yaml
+uid: microsoft.com/docfx/Contacts
+footer: <p>Footer for <code>microsoft.com/docfx/Contacts</code></p>
+```
```
vwxyzh: should we change `summary` and other properties in managed reference with same behavior?

Documentation/tutorial/intro_overwrite_files.md
```diff
@@ -0,0 +1,110 @@
...
+uid: microsoft.com/docfx/Contacts
+some_property: value
+conceptual: <p><b>Content</b> in Markdown</p>
+```
+
+### Anchor `*content`
+
+`*content` is the keyword invented and used specifically in *Overwrite File*s to represent the Markdown content following YAML header. We leverage [Anchors](http://www.yaml.org/spec/1.2/spec.html#id2765878) syntax in YAML specification for `*content`.
+
+The value for `*content` is always transformed from Markdown content to HTML. When `*content` is not used, the Markdown content below YAML header will be set to `conceptual` property; When `*content` is used, the Markdown content below YAML header will no longer be set to `conceptual` property. With `*content`, we can easily add Markdown content to any properties.
```
vwxyzh: should we rename `conceptual` property to `content`?

Documentation/tutorial/intro_overwrite_files.md
vwxyzh: @vicancy I don't think it is allowed, consider following case:
```md
---
uid: foo
conceptual: bar
---
foobar
```

Documentation/tutorial/intro_overwrite_files.md
vicancy: I agree, `content` sounds more natural. @chenkennt What's your opinion?

chenkennt: sounds good. but make sure it's backward compatible.

---
In reply to: [57546551](https://github.com/dotnet/docfx/pull/233#discussion_r57546551) [](ancestors = 57546551)

Documentation/tutorial/intro_overwrite_files.md
```diff
@@ -0,0 +1,110 @@
...
+*children* | [REST API item model](#rest-api-item-model) | Overwrite when *uid* of the item model matches
+*summary* | string | Overwrite
+*description* | string | Overwrite
+
+#### REST API item model
+Key | Type | Overwrite behavior
+--- | --- | ---
+*uid* | string | Key
+
+#### REST API item model
```
chenkennt: remove

Documentation/tutorial/intro_overwrite_files.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+In the above example, the value for `*content` is `<p>Footer for <code>microsoft.com/docfx/Contacts</code></p>`, and the *Overwrite Model* represented in YAML format is:
+
+```yaml
+uid: microsoft.com/docfx/Contacts
+footer: <p>Footer for <code>microsoft.com/docfx/Contacts</code></p>
+```
+
+`uid` for a *Overwrite Model* stands for the Unique IDentifier of the *Model* it will overwrite. So it is allowed to have multiple *Overwrite Section*s with YAML Header containing the same `uid`. For one *Overwrite File*, the latter *Overwrite Section* overwrites the former one with the same `uid`. For different *Overwrite File*s, the order of overwrite is **Undetermined**. So it is suggested to have *Overwrite Sections* with the same `uid` in the same *Overwrite File*.
```
superyyrrzz: a *Overwrite Model* => an *Overwrite Model*

Documentation/tutorial/intro_overwrite_files.md
```diff
@@ -0,0 +1,110 @@
...
+---
+Further description for `microsoft.com/docfx/Contacts`
+```
+
+Each *Overwrite Section* is transformed to *Overwrite Model* inside DocFX. For the above example, the *Overwrite Model* represented in YAML format is:
+
+```yaml
+uid: microsoft.com/docfx/Contacts
+some_property: value
+conceptual: <p><b>Content</b> in Markdown</p>
```
chenkennt: test

Documentation/tutorial/intro_overwrite_files.md
vicancy: exactly

---
In reply to: [57170647](https://github.com/dotnet/docfx/pull/233#discussion_r57170647) [](ancestors = 57170647)

Documentation/tutorial/intro_overwrite_files.md
vicancy: backward compatiblity costs a lot, on hold the change

---
In reply to: [57547775](https://github.com/dotnet/docfx/pull/233#discussion_r57547775) [](ancestors = 57547775,57546551)

Documentation/tutorial/intro_overwrite_files.md
vicancy: its already the same behavior. How to handle these properties is the job of plugins and templates

---
In reply to: [57272638](https://github.com/dotnet/docfx/pull/233#discussion_r57272638) [](ancestors = 57272638)

Issue #122, opened 2016-01-13
Date when .md file was committed


DovydasNavickas:
It is always convenient to see when the documentation was written to have sense if it is still relevant or not.
It would be nice to have that in the default template.

chenkennt:
Yes, it's in our plan.

vicancy:
Postpone to v1.6

vicancy:
On hold the change as the implementation has perf issue for large repo...

DovydasNavickas:
Any updates? :smile:

vicancy:
Is it possible to move OPS plugin to docfx? @herohua

herohua:
@vicancy It is possible, but we need to take care some issues: 
1. OPS plugin generates push time, not commit time. Current implementation needs update.
2. Shall we consider cross-platform support if we move the plugin into DocFX?

yufeih:
This is addressed in v3

BenBo-slc:
Hi @yufeih, would you have any idea when v3 would be released?
Or at the very least this feature would become available?
Because as far as I can tell this currently hasn't been released yet.
kind regards

PR #238, opened 2016-03-31
Implement #122, get last commit detail for files

vicancy:
Implement https://github.com/dotnet/docfx/issues/122
@chenkennt @vwxyzh @qinezh @superyyrrzz


vicancy:
On-hold the change as it has perf issue for large repos...

vicancy:
@chenkennt @vwxyzh @superyyrrzz @qinezh Please have a second round review.
It has perf issue for large repos, and I will add an option to enable it later.

Issue #240, opened 2016-04-05
Support Cross reference for external content


johnneycao:

vwxyzh:
eb2d2e5f1b70a87f4a3385997ff048d50bd9c7af

PR #242, opened 2016-04-06
add sitemap file.

vwxyzh:
#240

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
@@ -20,6 +20,7 @@ namespace Microsoft.DocAsCode.Build.Engine
     public class DocumentBuilder : IDisposable
     {
         public const string PhaseName = "Build Document";
+        public const string SiteMapFileName = "sitemap.yml";
 
         private CompositionHost GetContainer(IEnumerable<Assembly> assemblies)
         {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -107,6 +107,16 @@ public void Build(DocumentBuildParameters parameters)
                         UpdateHref(manifest, context);
 
                         var generatedManifest = TemplateProcessor.Transform(processor, manifest.Select(s => s.Item).ToList(), context, parameters.ApplyTemplateSettings);
+
+                        // export sitemap
+                        YamlUtility.Serialize(
+                            Path.Combine(parameters.OutputBaseDir, "site.xref"),
```
vicancy: "site.xref" move to constant? or make it configurable?

vicancy: Should we add file extension `yml` to it?

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -107,6 +107,16 @@ public void Build(DocumentBuildParameters parameters)
...
+                            from xref in context.XRefSpecMap.Values
+                            select new XRefSpec(xref)
+                            {
+                                Href = ((RelativePath)context.FileMap[xref.Href]).RemoveWorkingFolder().ToString() + "#" + XrefDetails.GetHtmlId(xref.Uid),
+                            });
```
vicancy: extract to a data model? And from the experience of Manifest file/TOC, a object model is more extensible than an array model....

vicancy: what about other properties, e.g. Name, FullName...

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -107,6 +107,16 @@ public void Build(DocumentBuildParameters parameters)
...
+
```
vicancy: Log info when xref file is generated?

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
vwxyzh: other properties is copied from old instance.

PR #267, opened 2016-04-18
Sometimes symbol.ContainingNamespace is null and result in a ExtractMetadataException

xen2:
Sometimes symbol.ContainingNamespace is null.

This currently happens when using Attribute constructor with unbound typeof (such as [MyAttribute(typeof(A<>))]).

This internally translate to UnboundArgumentErrorTypeSymbol, which have its ContainingNamespace set to null.


xen2:
Actually it still fails later, I will look into why it resolves into an error type symbol at all.

PR #270, opened 2016-04-20
throw exception if site folder doesn't exist

Eonasdan:
If the passed in site folder doesn't exist (user forgot to run build command), throw a useful exception instead of a DirectoryNotFound.

image

image


vicancy:
Thanks, looks cool

PR #287, opened 2016-05-04
don't traverse excluded folders to avoid PathTooLongException

### LINKED ISSUE #156: PathTooLongException
Hi, I'm wondering if anyone else has experienced a `PathTooLongException` when generating docs, and if there is a workaround.

superyyrrzz:
fix #156 
@chenkennt @vicancy @vwxyzh @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Glob/FileGlob.cs
+++ b/src/Microsoft.DocAsCode.Glob/FileGlob.cs
@@ -34,15 +34,9 @@ namespace Microsoft.DocAsCode.Glob
 
         private static IEnumerable<string> GetFilesFromSubfolder(string baseDirectory, string cwd, IEnumerable<GlobMatcher> globs, IEnumerable<GlobMatcher> excludeGlobs)
         {
-            var relativePath = GetRelativeFilePath(cwd, baseDirectory);
-            foreach (var exclude in excludeGlobs)
-            {
-                if (exclude.Match(relativePath)) yield break;
-            }
-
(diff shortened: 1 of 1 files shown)


chenkennt:
:+1:

src/Microsoft.DocAsCode.Glob/FileGlob.cs
```diff
@@ -34,9 +34,15 @@ private static IEnumerable<string> GetFilesCore(string cwd, IEnumerable<GlobMatc
 
         private static IEnumerable<string> GetFilesFromSubfolder(string baseDirectory, string cwd, IEnumerable<GlobMatcher> globs, IEnumerable<GlobMatcher> excludeGlobs)
         {
+            var relativePath = GetRelativeFilePath(cwd, baseDirectory);
```
vicancy: When I wanna exclude node_modules folder? should it be `node_modules/` instead of `node_modules/**`? Consider the following case: `exclude`: `**/node1`, while there are folder `node1` and file `a/node2`, folder `node1` should not be excluded.

superyyrrzz: @vicancy fixed. Now both `node_modules/` and `node_modules/**` can work, and only `node_modules/` can be used to avoid `PathTooLongException`.

vicancy:
:+1:

superyyrrzz:
committed in https://github.com/dotnet/docfx/commit/22dcba0373b3435e91ad9b02bf48f354d736687c

PR #292, opened 2016-05-06
add commentid in XRefSpec

ansyral:
@vicancy @vwxyzh @chenkennt


src/Microsoft.DocAsCode.Plugins/XRefSpec.cs
```diff
@@ -69,6 +70,17 @@ public string Href
...
 
+        public string CommentId
+        {
+            get
+            {
+                string value;
+                _dict.TryGetValue(CommentIdKey, out value);
+                return value;
+            }
+            set { _dict[CommentIdKey] = value; }
```
vwxyzh: remove key when null?

vwxyzh:
👍

vicancy:
:+1:

PR #297, opened 2016-05-09
fix bug when outputFolder, basedirectory and destination are all not set

vicancy:
@chenkennt

`Path.GetFullPath(string.Empty)` throws exception


chenkennt:
:+1:

chenkennt:
:+1:

PR #299, opened 2016-05-10
extract virtual method to update model content for managedreferenceprocessor

ansyral:
@vwxyzh @superyyrrzz @chenkennt @vicancy


vwxyzh:
👍

PR #303, opened 2016-05-11
Fix name of DfmFencesBlockRule

hellosnow:
Fix DfmFencesBlockRule's name from "RestApiFences" to "Fences".

@chenkennt @vicancy @superyyrrzz @vwxyzh @ansyral


src/Microsoft.DocAsCode.Dfm/Rules/DfmFencesBlockRule.cs
```diff
@@ -20,7 +20,7 @@ public class DfmFencesBlockRule : IMarkdownRule
         private const string DedentQueryStringKey = "dedent";
         private const char RegionSeparatorInRangeQueryString = ',';
 
-        public string Name => "RestApiFences";
+        public string Name => "Fences";
```
vwxyzh: duplicate with GFM fence.

vwxyzh:
👍

PR #305, opened 2016-05-13
Implement statictoc template

vicancy:
@chenkennt @vwxyzh @qinezh @superyyrrzz


src/docfx.website.themes/statictoc/ManagedReference.html.primary.js
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.
+var common = require('./common.js');
+var util = require('./statictoc.util.js');
+
+exports.transform = function (model) {
```
chenkennt: extract some of this file to a separate file that can be required.

PR #307, opened 2016-05-16
For container node with no href, click it expand the children nodes

vicancy:
@chenkennt @superyyrrzz @vwxyzh @qinezh


chenkennt:
:+1:

PR #318, opened 2016-05-18
Rest - overwrite the value when key already defined in reference object

hellosnow:
For example, if "b" key has already defined in json, and it's merging reference object which also has "b" key, should take the former and overwrite it.

@superyyrrzz @chenkennt @ansyral @vwxyzh @qinezh @vicancy

### DIFF
--- a/src/Microsoft.DocAsCode.Build.RestApi/Swagger/Internals/SwaggerObjectConverter.cs
+++ b/src/Microsoft.DocAsCode.Build.RestApi/Swagger/Internals/SwaggerObjectConverter.cs
@@ -34,16 +34,8 @@ namespace Microsoft.DocAsCode.Build.RestApi.Swagger.Internals
                         // Preserve value inside swagger.Token
                         foreach (var k in swagger.Token)
                         {
-                            JToken existing;
-                            if (jObject.TryGetValue(k.Key, out existing))
-                            {
-                                // Overwrite the value if the key is already defined.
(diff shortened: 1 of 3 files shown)


src/Microsoft.DocAsCode.Build.RestApi/Swagger/Internals/SwaggerObjectConverter.cs
```diff
@@ -37,10 +37,13 @@ public override void WriteJson(JsonWriter writer, object value, JsonSerializer s
                             JToken existing;
                             if (jObject.TryGetValue(k.Key, out existing))
```
vwxyzh: remove if, and set indexer directly

hellosnow: Resolved

---
In reply to: [63802675](https://github.com/dotnet/docfx/pull/318#discussion_r63802675) [](ancestors = 63802675)

superyyrrzz:
Should we log warning when conflicts?

vwxyzh:
👍

qinezh:
:shipit:

hellosnow:
Trying in petstore.swagger.io and it's behavior is overwritten. not log for now following the behavior

---
In reply to: [220218314](https://github.com/dotnet/docfx/pull/318#issuecomment-220218314) [](ancestors = 220218314)

PR #319, opened 2016-05-18
add ApiTypeAndSpec to avoid loop reference

superyyrrzz:
To fix the bug introduced in https://github.com/dotnet/docfx/commit/712a1dcecd64ee89d3fd63f08cb1610b8aa08178, avoid using ApiReferenceBuildOutput in ApiCrefBuildOutput/ApiParameterBuildOutput, which will cause loop ref.

@chenkennt @vwxyzh @qinezh @hellosnow @ansyral @vicancy


vwxyzh:
👍

PR #320, opened 2016-05-18
Update to be expanded by default

DuncanmaMSFT:


superyyrrzz:
@DuncanmaMSFT Please run `gulp` under `src/docfx.website.themes`, and then include the updated `src/docfx/Template/docs.html.zip` in the change. This zip file will be finally embedded into `docfx.exe`. Thanks.

DuncanmaMSFT:
Added new commit with zip file included. Thanks for the instructions!

superyyrrzz:
The `class.tmpl.partial` in zip is still the old version. Need to make sure the file has been updated before running `gulp`.

DuncanmaMSFT:
I sync'd with the upstream dev branch, then ran the gulp task again, so I believe it is now set. If it isn't, I might need hand holding to know how to make this update correctly

superyyrrzz:
:+1: Thanks @DuncanmaMSFT . It's a little troublesome to update ZIP for now... We will consider to improve the flow.

PR #334, opened 2016-05-26
Refactor ApplyOverwriteDocument, to support different overwrite model for Rest Api

hellosnow:
Plan to refactor REST API from single ItemVIewModel to RootItemVIewModel  and ChildItemVIewModel. Before implement this, need to refactor ApplyOverwriteDocument, to support different overwrite model.

@vwxyzh @qinezh @ansyral @chenkennt @superyyrrzz


vwxyzh:
👍

PR #338, opened 2016-05-27
separate category for extension methods and EIIs

superyyrrzz:
@vwxyzh @qinezh @hellosnow @ansyral @chenkennt


hellosnow:
:shipit:

vwxyzh:
👍

src/docfx.website.themes/default/ManagedReference.common.js
```diff
@@ -52,7 +54,13 @@ function groupChildren(model, typeChildrenItems) {
   var grouped = {};
 
   model.children.forEach(function (c) {
-    var type = c.type.toLowerCase();
+    if (c.isEii) {
+      var type = "eii";
+    } else if (c.isExtensionMethod) {
+      var type = "extensionMethod"
```
ansyral: extensionmethod is a bit different because the method's definition isn't in the class's page, namely it won't be put in Children property.

chenkennt: this is extension methods defined in current class. MSDN doesn't show this as a separate category. Do we really want to do this?

---
In reply to: [65029735](https://github.com/dotnet/docfx/pull/338#discussion_r65029735) [](ancestors = 65029735)

superyyrrzz:
@ansyral @chenkennt I misunderstood the `isExtensionMethod` property... I open another PR https://github.com/dotnet/docfx/pull/341 and only seperate EII for now. Will add extension method after Xuan's change.

PR #345, opened 2016-05-31
Update to template, as per docs bug # 549692

DuncanmaMSFT:
“Introduction” doesn’t seem like a very good term to use – it doesn’t really make clear what the content is. “Summary Description” might be better.  - Suggestion - Move it above inheritance and remove the need for title 'Introduction'


mairaw:
We have made a similar request before to address this. /cc @sankethka

superyyrrzz:
:+1: LGTM

sankethka:
@mairaw That same item has been assigned to Duncan. Duncan owns this item now.

mairaw:
Ah, cool! Glad to see this coming along then. :smile:

PR #346, opened 2016-06-01
Rest api - Uid and overwrite logic for tag

hellosnow:
- Implement uid and Conceptual for tag:
  - Tag's uid format is like "graph.windows.net/myorganization/Contacts/1.0/**tag**/pet store"
  - Writer can overwrite conceptual or description in overwrite document.
- In tag, if no **x-bookmark-id** is provided, get html id from **name**.
- Remove sort by section logic, only sort by tags and its original sequence in json.

@vwxyzh @superyyrrzz @ansyral @superyyrrzz @qinezh


src/Microsoft.DocAsCode.Build.RestApi/ApplyOverwriteDocumentForRestApi.cs
```diff
@@ -47,6 +47,21 @@ public class ApplyOverwriteDocumentForRestApi : ApplyOverwriteDocument
...
+                    overwriteModel,
+                    uid,
+                    s => BuildRestApiDocument.BuildTag(host, s, overwriteModel, content => content != null && content.Trim() == Constants.ContentPlaceholder));
+            }));
+
+        public Func<FileModel, string, IHostService, IEnumerable<RestApiTagViewModel>> GetTagItemsToOverwrite =
+            (((articleModel, uid, host) =>
+            {
+                return ((RestApiRootItemViewModel)articleModel.Content).Tags.Where(c => c.Uid == uid);
+            }));
```
vwxyzh: why not using method directly?
```csharp
public IEnumerable<RestApiTagViewModel> GetTagItemsToOverwrite(FileModel articleModel, string uid, IHostService host)
{
     return ((RestApiRootItemViewModel)articleModel.Content).Tags.Where(c => c.Uid == uid);
}
```

vwxyzh:
👍

src/Microsoft.DocAsCode.Build.RestApi/ApplyOverwriteDocumentForRestApi.cs
hellosnow: good catch!

---
In reply to: [65476613](https://github.com/dotnet/docfx/pull/346#discussion_r65476613) [](ancestors = 65476613)

PR #355, opened 2016-06-06
Rest api - use includeInTags instead of Tags.0 to indicate 'Other apis'

hellosnow:
use includeInTags instead of "Tags.0" to indicate 'Other apis', since tags are also defined in children.

@vwxyzh @ansyral @chenkennt @qinezh


src/docfx.website.themes/default/partials/rest.tmpl.partial
```diff
@@ -28,9 +28,9 @@
 {{/tags}}
 {{!if some children are not tagged while other children are tagged, add default title}}
 {{#children.0}}
-{{#tags.0}}
+{{^includedInTags}}
```
hellosnow: still are re-factoring

vwxyzh:
👍

Issue #325, opened 2016-05-19
Unclear how to write a toc outside the root directory


jskeet:
I wanted to make my "docs" root directory content-free - i.e. all content should be in subdirectories.

I created a `pages` directory containing `index.md` and `toc.yml`, and the metadata for my API is in an `obj` directory. So I have:

- docs
  - README.md (how to build the docs)
  - docfx.json
  - pages
    - toc.yml
    - index.md
  - obj
    - api
 
The `docfx.json` file has a `build` section like this:

```json
"build": {
  "content": [
    {
      "files": [ "**/*.yml" ],
      "src": "obj/api",
      "dest": "api"
    },
    {
      "files": [ "*.md", "toc.yml" ],
      "src": "pages"
    }
  ],
  // etc
}
```

But I can't work out how to refer to the `obj/api` directory from `toc.yml` as an `href`. I've tried:

- `/obj/api/` - that ends up with a link to literally `/obj/api`
- `../obj/api/` - docfx complains that it can't find the link
- `obj/api/` - docfx complains that it can't find the link

How is this meant to work? From my `index.md` file I'm able to use `../obj/api/SomeType.yml` but it doesn't seem to work for `toc.yml`.

qinezh:
Hi @jskeet, thanks for reporting the issue and we fix it in the latest release of docfx.

jskeet:
@qinezh: Excellent - so is `../obj/api` the right way to refer to it?

qinezh:
yes, docfx should find the link now.

PR #357, opened 2016-06-06
Fix bug when write relative folder in toc.yml

qinezh:
Fix issue #325 , test case would be added in another pull request  @vwxyzh @chenkennt @hellosnow @ansyral


vwxyzh:
👍

PR #363, opened 2016-06-07
refactor document builder.

vwxyzh:
@chenkennt @hellosnow @ansyral @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
@@ -95,6 +95,11 @@ namespace Microsoft.DocAsCode.Build.Engine
                 parameters.Metadata = ImmutableDictionary<string, object>.Empty;
             }
 
+            BuildCore(parameters);
+        }
+
+        private void BuildCore(DocumentBuildParameters parameters)
+        {
             using (new LoggerPhaseScope(PhaseName))
             {
                 Logger.LogInfo($"Max parallelism is {parameters.MaxParallelism.ToString()}.");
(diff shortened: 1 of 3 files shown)


hellosnow:
:shipit:

qinezh:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -427,45 +431,42 @@ private static void BuildArticle(IDocumentProcessor processor, HostService hostS
...
                     }
                 });
         }
 
-        private IEnumerable<ManifestItemWithContext> ExportManifest(InnerBuildContext buildContext, DocumentBuildContext context)
+        private IEnumerable<ManifestItemWithContext> ExportManifest(HostService hostService, DocumentBuildContext context)
         {
-            var hostService = buildContext.HostService;
-            var processor = buildContext.Processor;
-            var templateProcessor = buildContext.TemplateProcessor;
```
ansyral: Question, why move all these to HostService? what does a HostService instance mean?

vwxyzh: I don't want `build context` and `inner build context`, and I think no one know the different between them.
The `inner build context` is a named tuple for `host service`, `document processor` and `template processor`. And `host service` is an internal class, so merge them into that.
And now `host service` means the context for a document processor, and `IHostService` is a set of services providing for plug-ins.

ansyral: Thanks, i get it now. maybe later if we need to add more context related info, a HostServiceContext is still needed for better management

---
In reply to: [66018261](https://github.com/dotnet/docfx/pull/363#discussion_r66018261) [](ancestors = 66018261)

ansyral:
:shipit:

PR #379, opened 2016-06-13
Auto pack template in build.cmd

qinezh:
@chenkennt @superyyrrzz @vwxyzh @hellosnow @ansyral


UpdateTemplate.cmd
```diff
@@ -0,0 +1,25 @@
...
+SETLOCAL
+SETLOCAL ENABLEDELAYEDEXPANSION
+
+:: Check if node exists globally
+WHERE node >NUL
+IF NOT '%ERRORLEVEL%'=='0' (
+    ECHO ERROR: UpdateTemplate.cmd requires node installed globally.
+    GOTO :Exit
+)
+
```
superyyrrzz: Also check gulp?

qinezh: No need to check gulp since the step `npm install` will install it and other dependency

superyyrrzz: But I meet the error `'gulp' is not recognized as an internal or external command,
operable program or batch file.`... Seems need to install gulp globally?

qinezh: You're right, I should add a step to global install gulp

UpdateTemplate.cmd
```diff
@@ -0,0 +1,27 @@
...
+PUSHD %DefaultTemplate%
+CALL npm install
+CALL gulp
+POPD
+
+PUSHD %TemplateHome%
+CALL npm install
+CALL gulp
+POPD
```
chenkennt: missed one popd in the end

chenkennt: actually only one pair of pushd and popd is needed at beginning and end of the script.

---
In reply to: [66747755](https://github.com/dotnet/docfx/pull/379#discussion_r66747755) [](ancestors = 66747755)

UpdateTemplate.cmd
```diff
@@ -0,0 +1,27 @@
...
+PUSHD %~dp0
+
+SETLOCAL
+SETLOCAL ENABLEDELAYEDEXPANSION
+
+:: Check if node exists globally
+WHERE node >NUL
+IF NOT '%ERRORLEVEL%'=='0' (
+    ECHO ERROR: UpdateTemplate.cmd requires node installed globally.
+    GOTO :Exit
```
hellosnow: add definition for Exit?

hellosnow:
:shipit:

superyyrrzz:
:+1:

superyyrrzz:
@DuncanmaMSFT We've removed the ZIPs from our repo, as `build.cmd` will generate them automatically. So the ZIPs are no longer required in commits.

PR #381, opened 2016-06-13
Refactoring the templates a bit, changing enums to generate a table for their members

DuncanmaMSFT:


superyyrrzz:
As default template is shared by all templates, the change in `default/ManagedReference.common.js` will break others. What about adding `isEnum` in docs template? Here's some place need to be changed:
1. Copy `ManagedReference.html.primary.js` from `default` to `docs`, then add `isEnum` and remove `isClass` in `model` after `model = mrefCommon.transform(model);`
2. Remove this line: https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/gulpfile.js#L146
3. And remember to revert `ManagedReference.common.js` :)

DuncanmaMSFT:
I've updated the pull request with the changes from the earlier comment, but I'm getting a commit check issue (AppVeyor build failed). I can't see what the problem is viewing the details.

qinezh:
@DuncanmaMSFT There are some unknown issues for AppVeyor to build docfx, you could ignore the commit check status before we fix it.

superyyrrzz:
:+1:

PR #386, opened 2016-06-14
Use indexof instead of endswith as current version doesn't support en…

fenxu:
…dswith


src/docfx.website.themes/op.html/op.common.js
```diff
@@ -18,6 +18,14 @@ exports.getCanonicalUrl = function (canonicalUrlPrefix, path, layout) {
   return canonicalUrl;
 }
 
+exports.endsWith = function(str, suffix) {
```
superyyrrzz: Move this function inside `exports.getCanonicalUrl`? `exports.xxx` is used to expose the function to other js.

PR #388, opened 2016-06-15
When merging ReferenceItem.IsDefinition, use the same resolution as ReferenceItem.Definition

xen2:
I had some problem with https://github.com/dotnet/docfx/blob/master/src/Microsoft.DocAsCode.Metadata.ManagedReference/Models/ReferenceItem.cs#L66
The Debug.Assert fails because I have a merge with IsDefinition == true on one side and IsDefinition == false on the other.

I changed it so that IsDefinition pick the same side as Merge of ReferenceItem.Definition.
Not sure if this is the right things to do, but at least it prevents it from Debug.Assert => crash


vwxyzh:
Hi @xen2, I think `IsDefinition` conflict is a bug in our code, and this fix is just a work around, not really fix the bug.

xen2:
Ok thanks for the clarification.
I thought so as well, probably not a good fix, as mentioned in the first PR comment.
Probably I should have opened an issue rather than a PR?

xen2:
Seems to be fixed properly by #390, thanks! (sorry about that PR that should just have been an issue)

PR #394, opened 2016-06-17
fix commentid lost for cref defined in see/seealso/exception

ansyral:
@chenkennt @vwxyzh @hellosnow @superyyrrzz @qinezh


src/Microsoft.DocAsCode.Build.ManagedReference/ManagedReferenceDocumentProcessor.cs
```diff
@@ -136,7 +136,7 @@ private IEnumerable<XRefSpec> GetXRefFromReference(PageViewModel vm)
             }
             foreach (var reference in vm.References)
             {
-                if (reference?.IsExternal == true)
+                if (reference?.IsExternal != false)
```
vwxyzh: what about: reference == null

vwxyzh:
👍

PR #395, opened 2016-06-17
fix bug for extension method: 1. overload non-generic extension methods 2. multiple retargeted symbol founded

ansyral:
@chenkennt @vwxyzh @hellosnow @superyyrrzz @qinezh


vwxyzh:
👍

PR #398, opened 2016-06-20
Add Migration Rule and Token for Azure content

fenxu:
Add new migration rule/token instead of modify the existed ones of azure. So that we can switch between them.


vwxyzh:
👍

PR #409, opened 2016-06-21
Fix bug in see-also

qinezh:
@chenkennt @superyyrrzz @vwxyzh @hellosnow @ansyral

### DIFF
--- a/src/docfx.website.themes/docs.html/partials/class.tmpl.partial
+++ b/src/docfx.website.themes/docs.html/partials/class.tmpl.partial
@@ -286,7 +286,7 @@
 <div class="level">
   <section class="primary">
      {{#type.specName}}
-        <code class="lang-{{lang}}">{{{value}}}</code>
+        <span class="lang-{{lang}}">{{{value}}}</span>
      {{/type.specName}}
   </section>
   <section class="secondary">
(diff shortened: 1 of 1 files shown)


src/docfx.website.themes/docs.html/partials/class.tmpl.partial
```diff
@@ -286,7 +286,7 @@
 <div class="level">
   <section class="primary">
      {{#type.specName}}
-        {{{value}}}
+        <code class="lang-{{lang}}">{{{value}}}</code>
```
hellosnow: {{{?

hellosnow:
:shipit:

src/docfx.website.themes/docs.html/partials/class.tmpl.partial
qinezh: the same as L165

hellosnow: how about update together?

---
In reply to: [67851987](https://github.com/dotnet/docfx/pull/409#discussion_r67851987) [](ancestors = 67851987)

superyyrrzz:
I find a page with seealso: https://docs.microsoft.com/en-us/rights-management/sdk/2.1/api/win/environment%20properties#msipc_environment_properties
This PR will change the current UI. What about `<span class="lang-{{lang}}">{{{value}}}</span>`?

qinezh:
@superyyrrzz it doesn't work after trying to use <span>

superyyrrzz:
:+1:

PR #413, opened 2016-06-22
support parameters in path, to be applicable for all the operations under this path

### LINKED ISSUE #358: Rest Swagger - support `parameters` applicable for all the operations under `path`
Currently, docfx rest swagger support `parameters` directly under `get`, `put` operation etc. But according [Swagger paths schema](http://swagger.io/specification/#pathsObject), `paths` should support `parameters` that are applicable for all the operations described under each path.
```json
{
  "paths": {
    "/test": {
      "parameters": [{
        "$ref": "#/parameters/A"
      }],
      "get": {
        "operationId": "B",
        "parameters": [{
          "name": "kind",
          "in": "path",
          "required": true,
          "type": "string"
        }]
      }
    }
  }
}
```

hellosnow:
Fix #358, parameters in [PathsObject](http://swagger.io/specification/#pathsObject), which is:

- A list of parameters that are applicable for all the operations described under this path. These parameters can be overridden at the operation level, but cannot be removed there.
- The list MUST NOT include duplicated parameters. 
- A unique parameter is defined by a combination of a name and location.

@chenkennt @ansyral @vwxyzh @superyyrrzz @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Build.RestApi/ViewModels/RestApiRootItemViewModel.cs
+++ b/src/Microsoft.DocAsCode.Build.RestApi/ViewModels/RestApiRootItemViewModel.cs
@@ -188,7 +188,7 @@ namespace Microsoft.DocAsCode.Build.RestApi.ViewModels
 
             // Path parameters can be overridden at the operation level.
             var uniquePathParams = pathParameters.Where(
-                p => operationParameters.Any(o => IsParameterEquals(p, o)) != true).ToList();
+                p => operationParameters.Any(o => !IsParameterEquals(p, o))).ToList();
 
             return operationParameters.Union(uniquePathParams).ToList();
         }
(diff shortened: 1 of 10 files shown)


src/Microsoft.DocAsCode.Build.RestApi/ViewModels/RestApiRootItemViewModel.cs
```diff
@@ -38,7 +40,7 @@ public class RestApiRootItemViewModel : RestApiItemViewModelBase
         [JsonProperty("children")]
         public List<RestApiChildItemViewModel> Children { get; set; }
 
-        public static RestApiRootItemViewModel FromSwaggerModel(Swagger.SwaggerModel swagger)
+        public static RestApiRootItemViewModel FromSwaggerModel(SwaggerModel swagger)
```
hellosnow: plan to move out of this class

vwxyzh:
👍

PR #415, opened 2016-06-22
Html encode to the text that will be used as attributes value.

fenxu:
As title, do some html encode operation. Otherwise, the generated file is not a valid xhtml although we set the xhtml output flag to true.


vwxyzh:
👍

PR #422, opened 2016-06-23
Fix issue 359 - unescape definition in swagger

### LINKED ISSUE #359: Rest Swagger - $ref with `[` and `]` not parsed successfully
When ref with "#/definitions/KeyValuePair[String,String]", the `[` and `]` are decoded to added as key in dict, but it's get without decoding.

Expected behavior: Swagger is parsed successfully.
Actual behavior: Could not resolve reference '#/definitions/KeyValuePair[String,String]' in the document

```json
{
  "definitions": {
    "A": {
      "type": "object",
      "properties": {
        "properties": {
          "description": "Properties",
          "type": "array",
          "items": {
            "$ref": "#/definitions/KeyValuePair[String,String]"
          }
        }
      }
    },
    "KeyValuePair[String,String]": {
      "type": "object",
      "properties": {
        "key": {
          "type": "string",
          "readOnly": true
        },
        "value": {
          "type": "string",
          "readOnly": true
        }
      }
    }
  }
}
```

hellosnow:
To fix #359 , unescape swagger definition key

@vwxyzh @ansyral @chenkennt @qinezh @superyyrrzz

### DIFF
--- a/src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
+++ b/src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
@@ -9,33 +9,39 @@ namespace Microsoft.DocAsCode.Build.RestApi
     public static class RestApiHelper
     {
         /// <summary>
-        /// Use JSON String Representation instead of URI Fragment Identifier Representation, refer to: https://tools.ietf.org/html/rfc6901#section-5
         /// Reverse to reference unescape described in http://tools.ietf.org/html/rfc6901#section-4
         /// </summary>
         /// <param name="reference"></param>
         /// <returns></returns>
         public static string FormatDefinitionSinglePath(string reference)
         {
-            return Uri.UnescapeDataString(reference.Replace("~", "~0").Replace("/", "~1"));
+            return reference.Replace("~", "~0").Replace("/", "~1");
         }
 
+        /// <summary>
+        /// When the reference starts with '#/', treat it as URI Fragment Identifier Representation and decode.
+        /// When the reference starts with '/', treat it as JSON String Representation and keep it as.
+        /// Refer to: https://tools.ietf.org/html/rfc6901#section-5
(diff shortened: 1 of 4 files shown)


src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        /// <summary>
+        /// Use JSON String Representation instead of URI Fragment Identifier Representation, refer to: https://tools.ietf.org/html/rfc6901#section-5
+        /// Reverse to reference unescape described in http://tools.ietf.org/html/rfc6901#section-4
+        /// </summary>
+        /// <param name="reference"></param>
+        /// <returns></returns>
+        public static string FormatDefinitionSinglePath(string reference)
+        {
+            return Uri.UnescapeDataString(reference.Replace("~", "~0").Replace("/", "~1"));
```
hellosnow: still need to careful consideration

hellosnow:
Refer to: https://tools.ietf.org/html/rfc6901#section-5:
- When the json pointer starts with '#/', treat it as **URI Fragment Identifier Representation** and **decode**.
- When the json pointer starts with '/', treat it as **JSON String Representation** and**keep it as**.

@chenkennt @vwxyzh please help to check whether the logic is correct.

src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+namespace Microsoft.DocAsCode.Build.RestApi
```
hellosnow: add one line before

vwxyzh:
👍

PR #424, opened 2016-06-23
Encode invalid html tag as that should be regarded as normal text to keep same behavior with GitHub

fenxu:
as title. Sample:
<1-100>, <_hello>, <?world>, <1_2 href=""good"">, <1 att='bcd'> should be encoded.
<abc> <a-hello> <a?world> <a_b href=""good""> <AC att='bcd'> should keep same.


vwxyzh:
👍

qinezh:
:+1:

PR #430, opened 2016-06-27
Fix filter return value in rest api common javacript

hellosnow:
filter function should return Boolean.

@qinezh @vwxyzh @ansyral @superyyrrzz


qinezh:
:+1:

Issue #396, opened 2016-06-19
How should I organize a multi-version site?


jskeet:
My [Noda Time](http://nodatime.org) site hosts the API reference documentation for multiple versions (all past releases, and the current "unstable" code). Each version also has its own user guide, which needs to refer to the API reference.

Obviously these contain the same type and member names by and large, so even though I can easily generate the metadata (yml files) from the source of previous releases easily enough, I can't use `docfx build` due to the collisions.

My current plan is to have a separate docfx.json file per release, build the sites as if they were entirely separate, and then stitch them together... but if there's any better approach, I'd love to hear it.

(I considered post-processing the YML files to add the version number to disambiguate, but that would be awkward in various ways, and lead to ugly URLs.)

bradygaster:
@jskeet I can help you with this. I agree with your opinion of post-processing of YML files. I'll check with our engineering team via @johnneycao to see if there's a "quick-and-easy" way to do this or if they'd have specific guidance on the solution. We'll update you here.

jskeet:
@bradygaster: Can that work with nice URLs? I'm aiming for URLs such as http://nodatime.org/1.0.x/api/NodaTime.LocalDate.html and http://nodatime.org/1.1.x/api/NodaTime.LocalDate.html - I wouldn't want the "1.0.x" in there twice.

bradygaster:
Unsure at this time. Would you be comfortable flipping the api and version segment of your URL? cc @johnneycao

jskeet:
@bradygaster: I'd *prefer* not to, but I could probably live with it if the upsides were significant. (I think the "current plan" described in the first comment would work, with the downside being that no part of docfx would know about the whole site, so it couldn't generate a title bar or whatever.)

bradygaster:
I chatted with @johnneycao about this and he's talking it over with the devs and plans on getting some guidance to you shortly. The goal here is obviously to have one DocFX JSON file to drive them all, right? The "workaround" would be if you had to create N sites and wire everything up manually.

jskeet:
The ideal would be to have a single docfx.json file, yes. But I'd rather sacrifice build cleanliness than end-user URL sanity.

chenkennt:
[...]

PR #461, opened 2016-07-04
Docset Part 1: Support docset in config #396

superyyrrzz:
@chenkennt @vwxyzh @qinezh @hellosnow @ansyral 

sample docfx.json:
```json
{
  "build": {
    "docsets": {
      "v2.0": {
        "content": [
          {
            "files": "**/*.yml",
            "src": "v2.0",
            "dest": "api"
          }
        ],
        "globalMetadata": {
          "apiVersion": "v2.0"
        },
        "dest": "v2.0"
      },
      "v2.1": {
        "content": [
          {
            "files": "**/*.yml",
            "src": "v2.1",
            "dest": "api"
          }
        ],
        "globalMetadata": {
          "apiVersion": "v2.1"
        },
        "dest": "v2.1"
      }
    },
    "content": [
      {
        "files": [
          "*.md"
        ]
      }
    ],
    "dest": "_site"
  }
}
```


src/docfx/Models/BuildJsonConfig.cs
```diff
@@ -96,5 +58,9 @@ public class BuildJsonConfig
         /// </summary>
         [JsonProperty("noLangKeyword")]
         public bool NoLangKeyword { get; set; }
+
+        [JsonProperty("docsets")]
+        public List<BuildJsonConfigCommon> Docsets { get; set; }
```
ansyral: need update our website's docfx.json?

ansyral:
:shipit:

src/docfx/Models/BuildJsonConfig.cs
superyyrrzz: As I only add a property "docset" property, it's compatible with the old docfx.json.

ansyral: but you moved some properties to `BuildJsonConfigCommon`, shouldn't the old docfx.json be updated to contain a `docsets` property with a list that has only one element?

vwxyzh:
also update sample :)

vwxyzh:
👍

Documentation/tutorial/docfx.exe_user_manual.md
```diff
@@ -277,6 +278,26 @@ Please refer to [How to Create Custom Templates](howto_create_custom_template.md
 }
 ```
 
+#### 3.2.2 Properties for docset config
```
qinezh: Also add sample here?

qinezh:
:shipit:

Documentation/tutorial/docfx.exe_user_manual.md
```diff
@@ -277,6 +278,70 @@ Please refer to [How to Create Custom Templates](howto_create_custom_template.md
...
+
+*Docsets* are used to build multiple docsets with a single run of DocFX. The behavior is like running DocFX multiple times with different `docfx.json`. It's useful when building multiple versions of the same repository, which will lead to `uid` collusion if not put in different docsets.
+**NOTE** Cross Reference between different docsets is not allowed.
+
+Key                      | How to Merged with root config                                                                          | Description
```
hellosnow: merge

Documentation/tutorial/docfx.exe_user_manual.md
```diff
@@ -277,6 +278,70 @@ Please refer to [How to Create Custom Templates](howto_create_custom_template.md
 }
 ```
 
+#### 3.2.2 Properties for docset config
+
+*Docsets* are used to build multiple docsets with a single run of DocFX. The behavior is like running DocFX multiple times with different `docfx.json`. It's useful when building multiple versions of the same repository, which will lead to `uid` collusion if not put in different docsets.
+**NOTE** Cross Reference between different docsets is not allowed.
```
hellosnow: I tried and found it's appending to previous line, add one line before to start a new line?

hellosnow:
:shipit:

superyyrrzz:
@jskeet To support multiple versions, now we have 2 designs of `docfx.json`:

Design 1. Add a dictionary named `versions` to the top level of `docfx.json`. The value of this dictionary is like a full `docfx.json` for a version. Outside `docsets`, all things remain the same:
```json
{
  "build": {
    "versions": {
      "v2.0": {
        "content": [
          {
            "files": "**/*.yml",
            "src": "v2.0",
            "dest": "api"
          }
        ],
        "dest": "v2.0"
      },
      "v2.1": {
        "content": [
          {
            "files": "**/*.yml",
            "src": "v2.1",
            "dest": "api"
          }
        ],
        "dest": "v2.1"
      }
    },
    "content": [
      {
        "files": [
          "*.md"
        ]
      }
    ],
    "dest": "_site"
  }
}
```
Design 2. A conciser way to solve uid conflict is to directly add `version` attribute in `content`. Files in each "version" have separated scope:
```json
{
  "build": {
    "content": [
      {
        "files": "*.yml",
        "src": "v2.0",
        "dest": "v2.0/api",
        "version": "v2.0"
      },
      {
        "files": "*.yml",
        "src": "v2.1",
        "dest": "v2.1/api",
        "version": "v2.1"
      },
      {
        "files": "conceptual/*.md",
        "src": "v2.0",
        "dest": "v2.0",
        "version": "v2.0"
      },
      {
        "files": "conceptual/*.md",
        "src": "v2.1",
        "dest": "v2.1",
        "version": "v2.1"
      }
    ],
    "dest": "_site"
  }
} 
```
From user's perspective, which one do you prefer?

jskeet:
Option 2 looks a bit cleaner to me - but what's the impact of the two options beyond `docfx.json`? Two things I'm concerned about:

- How I refer to specific versions from top-level documentation - think of a home page with "Read the in-development documentation *here*, or the latest stable release *here*."
- How I refer to other items "within" a specific version, e.g. from a conceptual doc to a specific method.

In particular, for the latter, it would be a pain if that meant the doc code containing the version number - as moving from (say) v2.0 to v2.1 should start off by just copying the documentation source.

superyyrrzz:
Thank you for your feedback:) We also tend to the later one. As both options aim to **separate** contents to avoid uid conflict, the impact of them are the same:

- For the first, due to the separated scope, one version can't refer to another. That's to say, DocFX can't resolve `[here](v2.0/index.md)` in top-level version, as it refers to a different version v2.0. A workaround is to use the correct relative path in output like`[here](v2.0/index.html)`. Reference cross versions still remains a problem that we will solve later.
- For the second, to refer to a specific method, you can use the method's uid directly. No extra version info need to be added as long as they have the same version in `docfx.json`.

jskeet:
Great - that sounds fabulous to me. Will look forward to seeing how it progresses. Thanks very much for working on this :)

PR #464, opened 2016-07-05
Rest - loop reference - Step 2 - Add reference name for complex type and loop reference

hellosnow:
Add reference name for complex type and loop reference, so that will show as A->B->A.

@chenkennt @vwxyzh @superyyrrzz @qinezh @ansyral

### DIFF
--- a/src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
+++ b/src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
@@ -27,33 +27,25 @@ namespace Microsoft.DocAsCode.Build.RestApi
         /// </summary>
         /// <param name="reference"></param>
         /// <returns></returns>
-        public static string FormatReferenceFullPath(string reference)
+        public static Tuple<string, string> FormatReferenceFullPath(string reference)
         {
             // Decode for URI Fragment Identifier Representation
             if (reference.StartsWith("#/"))
             {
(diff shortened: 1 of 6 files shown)


src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
```diff
@@ -44,5 +45,15 @@ public static string FormatReferenceFullPath(string reference)
...
         }
+
+        public static string GetReferenceName(string formattedPath)
+        {
+            if (!formattedPath.StartsWith("#/") && !formattedPath.StartsWith("/"))
+            {
+                throw new ArgumentException($"Formatted reference path \"{formattedPath}\" must start with '/' or '#/'");
+            }
+
+            return formattedPath.Split('/').Last();
```
vwxyzh: for `#/` decode first.

vwxyzh:
👍

src/Microsoft.DocAsCode.Build.RestApi/RestApiHelper.cs
hellosnow: before calling this method, formatted path has been decoded. But for better undestanding, I will merge this two methods into one.

---
In reply to: [69504557](https://github.com/dotnet/docfx/pull/464#discussion_r69504557) [](ancestors = 69504557)

PR #476, opened 2016-07-07
add yaml-mime

vwxyzh:

### DIFF
--- a/src/Microsoft.DocAsCode.Common/YamlMime.cs
+++ b/src/Microsoft.DocAsCode.Common/YamlMime.cs
@@ -1,0 +1,34 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Common
+{
+    using System.IO;
+
+    public static class YamlMime
+    {
+        public const string ManagedReference = nameof(ManagedReference);
+        public const string TableOfContent = nameof(TableOfContent);
+        public const string XRefMap = nameof(XRefMap);
+
+        public static string ReadMime(TextReader reader)
+        {
+            var line = reader.ReadLine();
+            if (!line.StartsWith("#"))
+            {
+                return null;
+            }
+            return line.TrimStart('#', ' ');
+        }
+
+#if !NetCore
+        public static string ReadMime(string file)
+        {
+            using (var reader = File.OpenText(file))
+            {
+                return ReadMime(reader);
+            }
+        }
+#endif
+    }
+}
--- a/src/Microsoft.DocAsCode.Common/YamlMime.cs
+++ b/src/Microsoft.DocAsCode.Common/YamlMime.cs
@@ -3,27 +3,36 @@
 
 namespace Microsoft.DocAsCode.Common
 {
+    using System;
     using System.IO;
 
     public static class YamlMime
     {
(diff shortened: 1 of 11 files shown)


src/Microsoft.DocAsCode.Common/YamlMime.cs
```diff
@@ -0,0 +1,34 @@
...
+            var line = reader.ReadLine();
+            if (!line.StartsWith("#"))
+            {
+                return null;
+            }
+            return line.TrimStart('#', ' ');
+        }
+
+#if !NetCore
+        public static string ReadMime(string file)
```
hellosnow: check file existence?

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Common/YamlMime.cs
```diff
@@ -0,0 +1,34 @@
...
+        public const string XRefMap = nameof(XRefMap);
+
+        public static string ReadMime(TextReader reader)
+        {
+            var line = reader.ReadLine();
+            if (!line.StartsWith("#"))
+            {
+                return null;
+            }
+            return line.TrimStart('#', ' ');
```
superyyrrzz: Commens is serialized in multiple lines In YamlUtility.Serialize(), but only read and return the first line here. Read more lines?

vwxyzh: "mime" must be first line

superyyrrzz:
:+1:

src/Microsoft.DocAsCode.Common/YamlMime.cs
ansyral: if so, need add some comments for the parameter "comments" in YamlUtility that its first line would be used as mime?

---
In reply to: [69884493](https://github.com/dotnet/docfx/pull/476#discussion_r69884493) [](ancestors = 69884493)

vwxyzh: yes, first line should be mime, and then other comments.
and there is no difference between "###" and "#" in yaml. just better experience for reader.

qinezh:
:shipit:

ansyral:
:shipit:

src/Microsoft.DocAsCode.Common/YamlMime.cs
vwxyzh: I can't check all cases following, so just let it throw.
```csharp
        // Exceptions:
        //   T:System.UnauthorizedAccessException:
        //     The caller does not have the required permission.
        //
        //   T:System.ArgumentException:
        //     path is a zero-length string, contains only white space, or contains one or more
        //     invalid characters as defined by System.IO.Path.InvalidPathChars.
        //
        //   T:System.ArgumentNullException:
        //     path is null.
        //
        //   T:System.IO.PathTooLongException:
        //     The specified path, file name, or both exceed the system-defined maximum length.
        //     For example, on Windows-based platforms, paths must be less than 248 characters,
        //     and file names must be less than 260 characters.
        //
        //   T:System.IO.DirectoryNotFoundException:
        //     The specified path is invalid, (for example, it is on an unmapped drive).
        //
        //   T:System.IO.FileNotFoundException:
        //     The file specified in path was not found.
        //
        //   T:System.NotSupportedException:
        //     path is in an invalid format.
```

Issue #449, opened 2016-06-30
Support incremental build for conceptual docs


chenkennt:
1. Extract dependency information
     - Code snippet
     - File include
     - Cross reference

2. Implement incremental build

PR #479, opened 2016-07-08
collect dependency for conceptual document.

vwxyzh:
#449

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/DependencyGraph.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/DependencyGraph.cs
@@ -1,0 +1,68 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Build.Engine
+{
+    using System.Collections.Generic;
+
+    public class DependencyGraph
+    {
+        private readonly Dictionary<string, SortedSet<string>> _dictionary =
+            new Dictionary<string, SortedSet<string>>();
+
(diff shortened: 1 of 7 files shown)


src/Microsoft.DocAsCode.Build.Engine/DependencyGraph.cs
```diff
@@ -0,0 +1,68 @@
...
+namespace Microsoft.DocAsCode.Build.Engine
+{
+    using System.Collections.Generic;
+
+    public class DependencyGraph
+    {
+        private readonly Dictionary<string, SortedSet<string>> _dictionary =
+            new Dictionary<string, SortedSet<string>>();
+
+        public void ReportDependency(string key, string value)
```
hellosnow: reuse the overload method below?

hellosnow:
:shipit:

chenkennt:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/DependencyGraph.cs
```diff
@@ -0,0 +1,68 @@
...
+                    if (_dictionary.TryGetValue(current, out set))
+                    {
+                        foreach (var item in set)
+                        {
+                            queue.Enqueue(item);
+                        }
+                    }
+                }
+            }
+            return result;
```
superyyrrzz: Can we record the result for later use to avoid some repeated computation?

vwxyzh: I want to keep it simple, if you think cache is necessary, I will create a cache decorator for it.

superyyrrzz:
:+1:

PR #495, opened 2016-07-13
add interface ISupportIncrementalBuild.

vwxyzh:
#449

### DIFF
--- a/src/Microsoft.DocAsCode.Build.ConceptualDocuments/BuildConceptualDocument.cs
+++ b/src/Microsoft.DocAsCode.Build.ConceptualDocuments/BuildConceptualDocument.cs
@@ -17,11 +17,12 @@ namespace Microsoft.DocAsCode.Build.ConceptualDocuments
     using Microsoft.DocAsCode.Utility;
 
     [Export(nameof(ConceptualDocumentProcessor), typeof(IDocumentBuildStep))]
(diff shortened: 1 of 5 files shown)


src/Microsoft.DocAsCode.Build.ConceptualDocuments/BuildConceptualDocument.cs
```diff
@@ -84,6 +85,20 @@ public override void Build(FileModel model, IHostService host)
             host.ReportDependency(model, result.Dependency);
         }
 
+        #region ISupportIncrementalBuild Members
+
+        public bool CanIncrementalBuild(FileAndType fileAndType)
```
ansyral: >CanIncrementalBuild [](start = 20, length = 19)

Question: in what case the return value is false? force-rebuild?

qinezh:
:shipit:

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.ConceptualDocuments/BuildConceptualDocument.cs
vwxyzh: some processor/step always need force-rebuild, for example, it will generate a date time now in some content.

ansyral:
:+1:

PR #506, opened 2016-07-18
fix metadata incremental check bug

ansyral:
take the following case as an example:
the src contains project A, B C.
1. project A depends on B,
2. project C depends on B, E
3. project E depends on A.

the original code might omit the dependency relationship 3 and generate the topological order BCA, which is wrong.


ansyral:
#507

vwxyzh:
✔️

hellosnow:
:shipit:

qinezh:
:100:

Issue #509, opened 2016-07-19
Neither 'list' nor export is found


superyyrrzz:
https://github.com/docascode/docfx-seed/issues/11

superyyrrzz:
@peterennis fixed in dev branch.

Improved message:
![image](https://cloud.githubusercontent.com/assets/3831744/17050979/82805394-5028-11e6-8c48-dafeafda8b97.png)

PR #520, opened 2016-07-21
improve message for docfx template

superyyrrzz:
#509 
@chenkennt @vwxyzh @qinezh @hellosnow @ansyral

### DIFF
--- a/src/docfx/Program.cs
+++ b/src/docfx/Program.cs
@@ -45,16 +45,7 @@ namespace Microsoft.DocAsCode
                 Logger.LogError(fe.Message);
                 return 1;
             }
-            catch (OptionParserException e)
-            {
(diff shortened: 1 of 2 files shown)


qinezh:
💯

src/docfx/Program.cs
```diff
@@ -54,6 +54,15 @@ private static int ExecSubCommand(string[] args)
                 }
                 return 1;
             }
+            catch (InvalidOptionException e)
```
hellosnow: >catch (InvalidOptionException e) [](start = 12, length = 32)

how about merge it with OptionParserException like catch (Exception ex) when (ex is OptionParserException|| ex is InvalidOptionException)

hellosnow:
:shipit:

PR #525, opened 2016-07-23
Ignore azure note content leading white spaces to avoid code block issue

fenxu:
As title. Confirmed with Azure team to do this action.


vwxyzh:
:+1:

PR #526, opened 2016-07-24
Add video syntax

fenxu:
docfx video syntax. It's first version of video syntax. Only contains link, the other properties such as width and height may should be configurable and we can modify the syntax later.

Also, we may need to extend the first version of video syntax based on users requirement.


src/Microsoft.DocAsCode.Dfm/DfmBlockquoteHelper.cs
```diff
@@ -21,6 +21,11 @@ public static List<SplitToken> SplitBlockquoteTokens(ImmutableArray<IMarkdownTok
                     splitToken = new SplitToken(token);
                     splitTokens.Add(splitToken);
                 }
+                else if (token is DfmVideoBlockToken)
```
superyyrrzz: Can merge with the above if by adding a `|| token is ...`?

src/Microsoft.DocAsCode.Dfm/Rules/DfmVideoBlockRule.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Microsoft.DocAsCode.Dfm
+{
+    using System.Text.RegularExpressions;
+
+    using Microsoft.DocAsCode.MarkdownLite;
+
+    public class DfmVideoBlockRule : IMarkdownRule
+    {
+        private static readonly Regex _videoRegex = new Regex(@"^ *\[\!Video +(?<link>https?\:\/\/.+?) *\] *(\n|$)", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
superyyrrzz: >^ *\[\!Video +(?<link>https?\:\/\/.+?) *\] *(\n|$) [](start = 64, length = 50)

Question: is it possible use syntax like GFM? [https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#youtube-videos](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#youtube-videos)

fenxu: The syntax you mentioned has already been supported in our syntax.
However, That's a picture link of video syntax but not embeded video in the page. They are different.

---
In reply to: [72000890](https://github.com/dotnet/docfx/pull/526#discussion_r72000890) [](ancestors = 72000890)

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Dfm/DfmBlockquoteHelper.cs
```diff
@@ -16,7 +16,7 @@ public static List<SplitToken> SplitBlockquoteTokens(ImmutableArray<IMarkdownTok
             SplitToken splitToken = null;
             foreach (var token in tokens)
             {
-                if (token is DfmSectionBlockToken || token is DfmNoteBlockToken)
+                if (token is DfmSectionBlockToken || token is DfmNoteBlockToken || token is DfmVideoBlockToken)
                 {
```
vwxyzh: what about as an interface?

src/Microsoft.DocAsCode.Dfm/DfmEngineBuilder.cs
```diff
@@ -57,10 +57,11 @@ public DfmEngineBuilder(Options options, string baseDir = null) : base(options)
                 throw new ArgumentException("MarkdownNewLineBlockRule should exist!");
             }
             blockRules.Insert(index + 1, new DfmIncludeBlockRule());
-            blockRules.Insert(index + 2, new DfmYamlHeaderBlockRule());
-            blockRules.Insert(index + 3, new DfmSectionBlockRule());
-            blockRules.Insert(index + 4, new DfmFencesBlockRule());
-            blockRules.Insert(index + 5, new DfmNoteBlockRule());
+            blockRules.Insert(index + 2, new DfmVideoBlockRule());
```
vwxyzh: >Insert [](start = 23, length = 6)

InsertRange instead of Insert +1, +2, ...

vwxyzh:
:shipit:

PR #538, opened 2016-07-27
Support relative path in breadcrumb_path

qinezh:
@chenkennt @vwxyzh @superyyrrzz @ansyral

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/Constants.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/Constants.cs
@@ -17,5 +17,7 @@ namespace Microsoft.DocAsCode.Build.Engine
 
         public const string ManifestFileName = "manifest.json";
 
+        public const string BreadCrumbPath = "breadcrumb_path";
+
     }
 }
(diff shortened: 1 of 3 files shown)


qinezh:
Three ways to use breadcrumb_path:

1. `"breadcrumb_path": "/a/b/toc.json"` (absolute path of generated TOC file)
2. `"breadcrumb_path": "~/c/toc.yml" `(relative path of source TOC file)
3. `"breadcrumb_path": "c/toc.yml" ` (will get warning that it needs `~`)

src/Microsoft.DocAsCode.Build.Engine/Constants.cs
```diff
@@ -17,5 +17,7 @@ internal static class Constants
 
         public const string ManifestFileName = "manifest.json";
 
+        public const string BreadCrumbPath = "breadcrumb_path";
```
superyyrrzz: BreadcrumbPath

src/Microsoft.DocAsCode.Build.Engine/TemplateModelTransformer.cs
```diff
@@ -176,6 +177,32 @@ internal ManifestItem Transform(InternalManifestItem item)
             return appended;
         }
 
+        private IDictionary<string, object> ResolveBreadCrumb(IDictionary<string, object> model)
```
chenkennt: >ResolveBreadCrumb [](start = 44, length = 17)

Make this a general function for resolve "path" metadata in model.

chenkennt: And we should let template to decide which metadata should be treated as path.

---
In reply to: [72409208](https://github.com/dotnet/docfx/pull/538#discussion_r72409208) [](ancestors = 72409208)

PR #548, opened 2016-07-29
Fix unescape bug in StringHelper

fenxu:
1. &amp; should return & in unescape method instead of string.Empty to match escape method in StringHelper
2. Add a wrong case, that the ' in the title part of markdown link syntax should be translated to &amp;. Add this case to reminder. @vwxyzh, I don't have a good idea to fix this, may be you should take a look.


vwxyzh:
@fenxu I've fix the double escape.

PR #564, opened 2016-08-03
Refine azure migration code

fenxu:
1. Split includes file and normal file list.
2. Extract class for azure video helper

### DIFF
--- a/src/Microsoft.DocAsCode.AzureMarkdownRewriters/AzureMarked.cs
+++ b/src/Microsoft.DocAsCode.AzureMarkdownRewriters/AzureMarked.cs
@@ -26,17 +26,17 @@ namespace Microsoft.DocAsCode.AzureMarkdownRewriters
                 context = engine.Context.CreateContext(engine.Context.Variables.SetItem("path", path));
             }
 
-            if (azureMarkdownFileInfoMapping?.Count != 0)
+            if (azureMarkdownFileInfoMapping != null && azureMarkdownFileInfoMapping.Count != 0)
             {
                 context = context.CreateContext(context.Variables.SetItem("azureMarkdownFileInfoMapping", azureMarkdownFileInfoMapping));
             }
 
-            if (azureVideoInfoMapping?.Count != 0)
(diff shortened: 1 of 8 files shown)


src/Microsoft.DocAsCode.AzureMarkdownRewriters/AzureMarked.cs
```diff
@@ -26,17 +26,17 @@ public class AzureMarked
                 context = engine.Context.CreateContext(engine.Context.Variables.SetItem("path", path));
             }
 
-            if (azureMarkdownFileInfoMapping != null)
+            if (azureMarkdownFileInfoMapping?.Count != 0)
```
ansyral: >azureMarkdownFileInfoMapping [](start = 16, length = 28)

what if azureMarkdownFileInfoMapping is null?

fenxu: Just not set the video info in the context and will skip the video transform.

---
In reply to: [73485155](https://github.com/dotnet/docfx/pull/564#discussion_r73485155) [](ancestors = 73485155)

tools/AzureMarkdownRewriterTool/AzureFileInformationCollection.cs
```diff
@@ -0,0 +1,56 @@
...
+    public class AzureFileInformationCollection
+    {
+        public AzureFileInformationCollection()
+        {
+            AzureMarkdownFileInfoMapping = new Dictionary<string, AzureFileInfo>();
+            AzureResourceFileInfoMapping = new Dictionary<string, AzureFileInfo>();
+            AzureIncludeMarkdownFileInfoMapping = new Dictionary<string, AzureFileInfo>();
+            AzureIncludeResourceFileInfoMapping = new Dictionary<string, AzureFileInfo>();
+            AzureVideoInfoMapping = new Dictionary<string, AzureVideoInfo>();
+        }
```
vwxyzh: use property initializer

vwxyzh:
:shipit:

PR #565, opened 2016-08-04
add some perf traces

herohua:
@hellosnow @vwxyzh @fenxu @superyyrrzz @qinezh @ansyral


herohua:
Revisit this after @vwxyzh 's change of `LoggerPhaseScope`. Close it for now.

PR #577, opened 2016-08-05
lower log level to diagnostic for model transform

herohua:
@vwxyzh @hellosnow @qinezh @superyyrrzz @ansyral @chenkennt


qinezh:
:100:

PR #581, opened 2016-08-08
readmine might throw null reference exception

ansyral:


hellosnow:
:shipit:

qinezh:
:+1:

src/Microsoft.DocAsCode.Common/YamlMime.cs
```diff
@@ -20,7 +20,7 @@ public static string ReadMime(TextReader reader)
                 throw new ArgumentNullException(nameof(reader));
             }
             var line = reader.ReadLine();
-            if (!line.StartsWith("#"))
+            if (line == null || !line.StartsWith("#"))
```
vwxyzh: >. [](start = 37, length = 1)

?.

vwxyzh:
:shipit:

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Common/YamlMime.cs
ansyral: the return value for `?.` is `bool?`

PR #582, opened 2016-08-08
use jobject to parse json, to judge whether json file is supported as swagger

hellosnow:
JsonUtility.Deserialize< Dictionary < string, object > > will fail for such case, since 'required' has been added twice, use JObject.Parse instead to only care about top keys.
```json
"A": {
      "required": [
        "B",
        "C"
      ],
      "required": [
        "B"
      ],
      "description": "Deployment properties."
    },
```
@vwxyzh @superyyrrzz @qinezh @chenkennt @ansyral


qinezh:
:shipit:

ansyral:
:shipit:

hellosnow:
Need to discuss with partner, whether the following syntax is wanted:
```json
"A": {
      "required": [
        "B",
        "C"
      ],
      "required": [
        "B"
      ],
      "description": "Deployment properties."
    },
```

PR #590, opened 2016-08-09
Rest - fix when 'definitions' has direct child with $ref defined

hellosnow:
Previously we didn't support following case, when "definitions" has direct child with "$ref" defined, we will treat it as SwaggerReferenceObject which is to be resolved, and not add into cache, so it will fail when resolving the reference in next steps.
```json
"definitions": {
    "containerUrl": {
      "$ref": "#/definitions/url",
      "description": "fake description."
    }
}
```

@ansyral @superyyrrzz @qinezh @chenkennt @vwxyzh


vwxyzh:
:shipit:

PR #611, opened 2016-08-16
Add NonDNX.sln

qinezh:


vwxyzh:
what about add new config `NonDnxDebug` and `NonDnxRelease` in `all.sln`?

qinezh:
Add an extra solution for non-DNX seems more clear since it no need load DNX projects?

hellosnow:
:shipit:

ansyral:
:+1:

chenkennt:
:shipit:

PR #627, opened 2016-08-22
support icremental for resourcedocumentprocessor and tocdocumentprocessor

ansyral:


src/Microsoft.DocAsCode.Build.TableOfContents/BuildTocDocument.cs
```diff
@@ -99,5 +99,19 @@ private void BuildCore(TocItemViewModel item, FileModel model, IHostService host
...
+        #region ISupportIncrementalBuild Members
+
+        public bool CanIncrementalBuild(FileAndType fileAndType)
+        {
+            return true;
+        }
+
+        public string GetIncrementalContextHash()
+        {
+            return null;
```
hellosnow: quesion: seems currently IncrementalContextHash all set to null? will we implement it later or not?

herohua: I think `IncrementalContextHash == null` means the step is stateless and can process those changed files only. @ansyral, correct?

ansyral: agree with @herohua . the hash would be used to check whether the processor/step can be incremental, currently seems no context info would influence whether the processor can be incremental

superyyrrzz:
:shipit:

qinezh:
:shipit:

hellosnow:
:shipit:

PR #629, opened 2016-08-23
refactor code: load/save intermediate files in a concentrated place so it could be atom operation

ansyral:
commit1.
move incrementals code back to engine project since they are mutual-referenced.
commit2.
load/save intermediate files during serialization/deserialization so that once the build failed, buildinfo won't record partial result.
intermediate files could use fixed name so that there is no need to clean up files from last build


qinezh:
`NonDNX.sln` also need update...

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalUtility.cs
```diff
@@ -0,0 +1,61 @@
...
+            {
+                return null;
+            }
+            using (var reader = new StreamReader(dependencyFile))
+            {
+                return DependencyGraph.Load(reader);
+            }
+        }
+
+        public static void SaveDependency(string fileName, DependencyGraph dg)
```
superyyrrzz: >SaveDependency [](start = 27, length = 14)

change to SaveDependencyGraph to be consistent with LoadDependencyGraph?

src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildVersionInfo.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// The file link for the XRefMap file(type is <see cref="XRefMap"/>).
+        /// </summary>
+        public string XRefSpecMapFile { get; set; }
+
+        #region Deserialized content
+        /// <summary>
+        /// deserialized dependency graph
+        /// </summary>
+        [JsonIgnore]
+        public DependencyGraph Dependency { get; set; }
```
superyyrrzz: >Dependency [](start = 31, length = 10)

What about move these properties to a new class like BuildVersionInfoModel(maybe other name..), and move the OnDeserializedMethod to BuildVersionInfoModel.From(bvi)?

ansyral: i don't think it is more clear and we need to track two objects then.

---
In reply to: [75987892](https://github.com/dotnet/docfx/pull/629#discussion_r75987892) [](ancestors = 75987892)

superyyrrzz:
:shipit:

qinezh:
:shipit:

ansyral:
#449

PR #632, opened 2016-08-24
fix required of properties for swagger

hellosnow:
Some required array is not handled and filled in each required field of properties.

@superyyrrzz @qinezh @chenkennt @ansyral


superyyrrzz:
:+1:

qinezh:
:shipit:

Issue #604, opened 2016-08-12
YamlModelGenerator error


adrienjt:
Hi,
I get an error when I run docfx.exe with the following docfx.json (pretty standard; just replaced my solution's name with dashes here):
```
{
  "metadata": [
    {
      "src": [
        {
          "src": "../BUILD",
          "files": [
            "----------.sln"
          ],
          "exclude": [
            "**/obj/**",
            "**/bin/**",
            "_site/**"
          ]
        }
      ],
      "dest": "api"
    }
  ],
  "build": {
    "content": [
      {
        "files": [
          "api/**.yml",
          "api/index.md"
        ]
      },
      {
        "files": [
          "articles/**.md",
          "articles/**/toc.yml",
          "toc.yml",
          "*.md"
        ],
        "exclude": [
          "obj/**",
          "_site/**"
        ]
      }
    ],
    "resource": [
      {
        "files": [
          "images/**"
        ],
        "exclude": [
          "obj/**",
          "_site/**"
        ]
      }
    ],
    "overwrite": [
      {
        "files": [
          "apidoc/**.md"
        ],
        "exclude": [
          "obj/**",
          "_site/**"
        ]
      }
    ],
    "dest": "_site",
    "globalMetadataFiles": [],
    "fileMetadataFiles": [],
    "template": [
      "default"
    ],
    "postProcessors": [],
    "noLangKeyword": false
  }
}
```
Here is the anonymized error (pretty verbose and with repeats, sorry):
```
Info: Config file docfx.json found, start generating metadata...
Info: 'C:/----------.csproj' keep up-to-date since '8/12/2016 8:57:02 PM', cached intermediate result 'C:/----------/obj/xdoc/cache/obj\z5sqmqw1.0d4' is used.
Info: 'C:/----------.csproj' keep up-to-date since '8/12/2016 8:57:03 PM', cached intermediate result 'C:/----------/obj/xdoc/cache/obj\ygbufxu1.4dd' is used.
Info: 'C:/----------.csproj' keep up-to-date since '8/12/2016 8:56:18 PM', cached intermediate result 'C:/----------/obj/xdoc/cache/obj\opjttcqf.0l5' is used.
Warning: DocFX currently only supports generating metadata with namespace defined. The following types in assembly "----------" will have no metadata generated: ----------, ----------, ----------, ----------, ----------, ----------. 
Warning: Invalid triple slash comment is ignored: <!-- Badly formed XML comment ignored for member "M:----------(----------)" -->

[...]

PR #633, opened 2016-08-24
when event type isn't defined in a namespace, resolving its parent might throw null reference exception

ansyral:
#604


hellosnow:
:shipit:

superyyrrzz:
:shipit:

qinezh:
:shipit:

PR #638, opened 2016-08-25
not overwrite buildinfo intermediate files when build status is failed

ansyral:
#449

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
@@ -318,7 +318,10 @@ namespace Microsoft.DocAsCode.Build.Engine
             }
             if (_intermediateFolder != null)
             {
-                // to-do: add check for build status. if failed, not overwrite buildinfo
+                if (_currentBuildInfo.Versions.Any(v => v.Status == BuildStatus.Failed))
+                {
+                    return;
+                }
                 JsonUtility.Serialize(
                     Path.Combine(_intermediateFolder, BuildInfo.FileName),
                     _currentBuildInfo);
--- a/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
@@ -318,10 +318,6 @@ namespace Microsoft.DocAsCode.Build.Engine
             }
             if (_intermediateFolder != null)
             {
-                if (_currentBuildInfo.Versions.Any(v => v.Status == BuildStatus.Failed))
-                {
-                    return;
-                }
                 JsonUtility.Serialize(
                     Path.Combine(_intermediateFolder, BuildInfo.FileName),
                     _currentBuildInfo);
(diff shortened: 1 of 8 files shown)


src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
```diff
@@ -125,6 +126,11 @@ private Manifest BuildCore(DocumentBuildParameters parameters)
                             manifest.AddRange(BuildCore(hostService, context));
                         }
 
+                        if (ShouldTraceIncrementalInfo)
+                        {
+                            UpdateUidFileDependency(context, hostServices);
```
herohua: wrap in `LoggerPhaseScope`

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -318,7 +318,10 @@ public void Dispose()
             }
             if (_intermediateFolder != null)
             {
-                // to-do: add check for build status. if failed, not overwrite buildinfo
+                if (_currentBuildInfo.Versions.Any(v => v.Status == BuildStatus.Failed))
```
superyyrrzz: This shuold work for OP as it can provide the latest succeeded output. Will it break for local if output and intermediate file don't match?

qinezh:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
ansyral: yes, this could be a problem for the case that build failed. might need to store the output to intermediate files

---
In reply to: [76353748](https://github.com/dotnet/docfx/pull/638#discussion_r76353748) [](ancestors = 76353748)

ansyral: i used a temporary walk around: always fully rebuild if last build failed

---
In reply to: [76365473](https://github.com/dotnet/docfx/pull/638#discussion_r76365473) [](ancestors = 76365473,76353748)

src/Microsoft.DocAsCode.Common/Loggers/WarningLogListener.cs
```diff
@@ -0,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Common
+{
+    using System;
+    using System.Collections.Concurrent;
+
+    public sealed class WarningLogListener : ILoggerListener
```
vwxyzh: Better naming, I think it is an in-memory log listener with default warning log level.

ansyral: how about LogWithSpecifiedLevelListener?

---
In reply to: [76367221](https://github.com/dotnet/docfx/pull/638#discussion_r76367221) [](ancestors = 76367221)

superyyrrzz:
:shipit:

ansyral:
use new skeleton.

PR #640, opened 2016-08-25
Replace Environment.CurrentDirectory in code

qinezh:
Replace `Environment.CurrentDirectory` with `Directory.GetCurrentDirectory()` since it's not supported in .NET Core. https://docs.microsoft.com/en-us/dotnet/core/api/system.environment


hellosnow:
:shipit:

herohua:
:+1:

PR #658, opened 2016-09-01
fix bugs in markdown table.

vwxyzh:
1. fix nptable rows start with `|`
2. fix columns more than or less than header count.
@chenkennt @hellosnow @ansyral @superyyrrzz @qinezh


superyyrrzz:
:shipit:

Issue #650, opened 2016-08-30
Problems with accented characters in index.json 


reinaldoarrosi:
I'm having trouble using full-text search (ExtractSearchIndex post-processor) when my markdown files contains accented characters.

For example:

/articles/aplicacao.md
`#Aplicação e módulos
Uma aplicação é um conjunto de módulos integrados.`

generates this index.json file
`{
  "toc.html": {
    "href": "toc.html",
    "title": "",
    "keywords": ""
  },
  "articles/aplicacao.html": {
    "href": "articles/aplicacao.html",
    "title": "Aplicação e módulos",
    "keywords": "AplicaÃ§Ã£o e mÃ³dulos Uma aplicaÃ§Ã£o Ã© um conjunto de mÃ³dulos integrados."
  }
}`

The .html output file (_site/articles/aplicacao.html) is generated correctly.

It seems to be there is a problem with index.json file encoding.

evil-shrike:
Indexing definitely has problems with encoding.
It's unclear in what encoding index.json is. I have a source file which is in UTF-8 encoding and contains Russian letters:
```
	/// <summary>
	/// Типы шрифтов
	/// </summary>
	public enum FontType
	{
		WOFF,
		EOT,
		TTF, // он же OTF
		SVG
	}
```
In index.json I have:
```
  "api-server/Croc.XFW3.Web.Bootloader.FontType.html": {
    "href": "api-server/Croc.XFW3.Web.Bootloader.FontType.html",
    "title": "Enum FontType | API .Net Server | WebClient",
    "keywords": "Enum FontType Р СћР С‘Р С—РЎвЂ№ РЎв‚¬РЎР‚Р С‘РЎвЂћРЎвЂљР С•Р Р† Namespace :...skipped"
  },
```
it's not utf8, not ansi 1251, I have no idea what it is. 
Obviously in search results I can see the same gibberish.

PR #660, opened 2016-09-01
fixes index.json encoding problems with non-english characters

reinaldoarrosi:
It appears that the problem is caused when reading .html files used to generate the index.json file in ExtractSearchIndex class.

I've changed the encoding used to read the .html files to UTF-8 since, it appears to me, that .html files are always generated in UTF-8.

#650

### DIFF
--- a/test/Microsoft.DocAsCode.Build.Engine.Tests/ExtractSearchIndexFromHtmlTest.cs
+++ b/test/Microsoft.DocAsCode.Build.Engine.Tests/ExtractSearchIndexFromHtmlTest.cs
@@ -3,14 +3,13 @@
 
 namespace Microsoft.DocAsCode.Build.Engine.Tests
 {
-    using Microsoft.DocAsCode.Build.Common;
-
     using HtmlAgilityPack;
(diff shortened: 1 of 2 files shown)


test/Microsoft.DocAsCode.Build.Engine.Tests/ExtractSearchIndexFromHtmlTest.cs
```diff
@@ -7,6 +7,10 @@ namespace Microsoft.DocAsCode.Build.Engine.Tests
 
     using HtmlAgilityPack;
     using Xunit;
+    using System.IO;
+    using Plugins;
+    using System.Collections.Generic;
```
vwxyzh: - Sort namespaces
- Use full name, e.g.: `Microsoft.DocAsCode.Plugins`

vwxyzh:
:shipit:

superyyrrzz:
:shipit:

hellosnow:
:shipit:

qinezh:
:shipit:

PR #692, opened 2016-09-12
not support postbuild incremental now

ansyral:
@vwxyzh @superyyrrzz @hellosnow @qinezh @chenkennt

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
@@ -373,7 +373,7 @@ namespace Microsoft.DocAsCode.Build.Engine
                 MarkdownService = markdownService,
                 DependencyGraph = new DependencyGraph(),
             };
-            BuildCore(new List<HostService> { hostService }, parameters.MaxParallelism, null, null, null, null);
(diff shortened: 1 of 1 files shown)


vwxyzh:
:+1:

src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
```diff
@@ -373,7 +373,7 @@ into model
                 MarkdownService = markdownService,
                 DependencyGraph = new DependencyGraph(),
             };
-            BuildCore(new List<HostService> { hostService }, parameters.MaxParallelism, null, null, null, null);
+            BuildCore(new List<HostService> { hostService }, parameters.MaxParallelism, null, null);
```
herohua: maybe use a switch to turn it off rather than delete the code?

hellosnow:
:shipit:

herohua:
👍

src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
ansyral: i didn't delete the updater method. it is easy to change the code later

PR #717, opened 2016-09-19
Refactor codes for incremental build.

vwxyzh:
@chenkennt @hellosnow @qinezh 
#449

### DIFF
--- a/src/Microsoft.DocAsCode.Build.ConceptualDocuments/BuildConceptualDocument.cs
+++ b/src/Microsoft.DocAsCode.Build.ConceptualDocuments/BuildConceptualDocument.cs
@@ -90,30 +90,13 @@ namespace Microsoft.DocAsCode.Build.ConceptualDocuments
             }
         }
 
-        #region ISupportIncrementalBuild Members
+        #region ISupportIncrementalBuildStep Members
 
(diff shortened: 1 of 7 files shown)


src/Microsoft.DocAsCode.Build.ConceptualDocuments/BuildConceptualDocument.cs
```diff
@@ -90,30 +90,13 @@ public override void Build(FileModel model, IHostService host)
...
-            {
-                new DependencyType
-                {
-                    Name = DependencyTypeName.Include,
-                    IsTransitive = true,
-                    TriggerBuild = true,
-                }
-            };
-        }
+        public IEnumerable<DependencyType> GetDependencyTypesToRegister() => null;
```
herohua: no dependency type to register here?

vwxyzh: move to default (at DependencyGraph.cs line 19).

---
In reply to: [79351820](https://github.com/dotnet/docfx/pull/717#discussion_r79351820) [](ancestors = 79351820)

herohua:
:shipit:

PR #725, opened 2016-09-22
Add fallback folders

fenxu:


src/Microsoft.DocAsCode.Dfm/DocfxFlavoredIncHelper.cs
```diff
@@ -18,21 +18,24 @@ namespace Microsoft.DocAsCode.Dfm
     internal sealed class DocfxFlavoredIncHelper : IDisposable
     {
         private readonly FileCacheLite _cache;
+        private readonly FileCacheLite _fallbackCache;
```
vwxyzh: >_fallbackCache [](start = 39, length = 14)

why we need this?

fenxu: Just want to separate the cache. It could reuse the cache if you think it is redundant. 

---
In reply to: [79969518](https://github.com/dotnet/docfx/pull/725#discussion_r79969518) [](ancestors = 79969518)

vwxyzh:
:shipit:

PR #732, opened 2016-09-26
fix IsRelativePath in PathUtility.

vwxyzh:
@chenkennt @ansyral @hellosnow @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Utility/PathUtility.cs
+++ b/src/Microsoft.DocAsCode.Utility/PathUtility.cs
@@ -7,14 +7,16 @@ namespace Microsoft.DocAsCode.Utility
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
-    using System.Web;
+    using System.Text.RegularExpressions;
 
     /// <summary>
     /// 
     /// </summary>
     public static class PathUtility
     {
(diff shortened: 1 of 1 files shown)


src/Microsoft.DocAsCode.Utility/PathUtility.cs
```diff
@@ -7,14 +7,16 @@ namespace Microsoft.DocAsCode.Utility
...
-    using System.Web;
+    using System.Text.RegularExpressions;
 
     /// <summary>
     /// 
     /// </summary>
     public static class PathUtility
     {
-        private static char[] InvalidFilePathChars = Path.GetInvalidFileNameChars();
+        private static readonly Regex UriWithProtocol = new Regex(@"^\w{2,}\:", RegexOptions.Compiled);
```
hellosnow: >\w [](start = 69, length = 2)

[a-zA-Z] is enough?

vwxyzh: [https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax)

---
In reply to: [80407718](https://github.com/dotnet/docfx/pull/732#discussion_r80407718) [](ancestors = 80407718)

hellosnow:
:shipit:

PR #739, opened 2016-09-27
fix duplicate warning message.

vwxyzh:
@chenkennt @fenxu @ansyral @hellosnow


src/Microsoft.DocAsCode.Build.TableOfContents/TocDocumentProcessor.cs
```diff
@@ -169,7 +170,7 @@ private void RegisterTocMap(TocItemViewModel item, string key, IDocumentBuildCon
             }
         }
 
-        private string ResolveHref(string pathToFile, string originalPathToFile, FileModel model, IDocumentBuildContext context)
```
fenxu: I can only see error message change here. How does it fix duplicate warning message issue?

fenxu:
:shipit:

PR #742, opened 2016-09-27
incremental: relay warnings from last build

ansyral:
@vwxyzh @chenkennt  @fenxu @hellosnow

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildMessageInfo.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildMessageInfo.cs
@@ -4,7 +4,6 @@
 namespace Microsoft.DocAsCode.Build.Engine.Incrementals
 {
     using System;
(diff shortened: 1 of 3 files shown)


src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildMessageInfo.cs
```diff
@@ -3,18 +3,46 @@
...
     using System.Linq;
 
     using Microsoft.DocAsCode.Common;
+    using Microsoft.DocAsCode.Plugins;
+    using Microsoft.DocAsCode.Utility;
 
     public sealed class BuildMessageInfo
     {
         private Listener _listener;
+        private readonly ConcurrentDictionary<string, List<LogItem>> _logs;
```
vwxyzh: I check the Logger class, there is a lock in Logger.cs at 64, so no need to handle thready safety here.

PR #786, opened 2016-10-14
fix the bug when the filePath contains backslash

928PJY:
fix the bug when the filePath contains backslash
@hellosnow @Huachao @qinezh @vwxyzh


src/VscPreviewExtension/media/js/previewMatch.js
```diff
@@ -6,6 +6,8 @@ var lastLocation = 0;
 
 var fileName = document.body.childNodes[1].data;
 
+var fileNameEscape = replaceAllInString(fileName, "\\", "\\\\");
```
928PJY: yes, that would be better!

hellosnow:
:shipit:

src/VscPreviewExtension/media/js/previewMatch.js
```diff
@@ -6,6 +6,8 @@ var lastLocation = 0;
 
 var fileName = document.body.childNodes[1].data;
```
hellosnow: >fileName [](start = 4, length = 8)

should it be filePath? filename shouldn't contain backslash?

Huachao:
:shipit:

PR #799, opened 2016-10-19
fix xref's text being double escaped

superyyrrzz:
@vwxyzh @chenkennt @ansyral @hellosnow @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
@@ -114,7 +114,6 @@ namespace Microsoft.DocAsCode.Build.Engine
                 string value = InnerHtml;
                 if (string.IsNullOrEmpty(value))
                 {
-                    value = StringHelper.HtmlEncode(Text);
                     if (string.IsNullOrEmpty(value) && Spec != null)
                     {
                         value = StringHelper.HtmlEncode(GetLanguageSpecificAttribute(Spec, language, Uid, DisplayProperty, "name"));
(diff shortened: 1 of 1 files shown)


src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
```diff
@@ -114,7 +114,6 @@ public HtmlAgilityPack.HtmlNode ConvertToHtmlNode(string language)
                 string value = InnerHtml;
                 if (string.IsNullOrEmpty(value))
                 {
-                    value = StringHelper.HtmlEncode(Text);
                     if (string.IsNullOrEmpty(value) && Spec != null)
```
hellosnow: >string.IsNullOrEmpty(value) [](start = 24, length = 27)

string.IsNullOrEmpty(value) must be true?

src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
```diff
@@ -114,7 +114,6 @@ public HtmlAgilityPack.HtmlNode ConvertToHtmlNode(string language)
                 string value = InnerHtml;
                 if (string.IsNullOrEmpty(value))
                 {
-                    value = StringHelper.HtmlEncode(Text);
```
hellosnow: >value [](start = 20, length = 5)

should value fall back to text first? even not encode

superyyrrzz: yes!

---
In reply to: [83991038](https://github.com/dotnet/docfx/pull/799#discussion_r83991038) [](ancestors = 83991038)

hellosnow:
:shipit:

PR #805, opened 2016-10-20
work around plugin loading multiple times

ansyral:
@herohua @hellosnow


src/docfx/SubCommands/DocumentBuilderWrapper.cs
```diff
@@ -160,6 +160,12 @@ private static IEnumerable<Assembly> LoadPluginAssemblies(string pluginDirectory
                         Logger.LogWarning("Skipping assembly: Microsoft.DocAsCode.EntityModel.");
                         continue;
                     }
+                    if (assemblyName == "Microsoft.DocAsCode.Build.Common")
```
hellosnow: not sure whether typeof("..").Assembly.GetName().Name is better?

ansyral: nice

PR #807, opened 2016-10-21
value from xref query string should be html encoded

superyyrrzz:
@chenkennt @vwxyzh @hellosnow @qinezh @ansyral


src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
```diff
@@ -71,10 +71,10 @@ public static XRefDetails From(HtmlAgilityPack.HtmlNode node)
             }
 
             xref.InnerHtml = node.InnerHtml;
-            xref.DisplayProperty = node.GetAttributeValue("displayProperty", queryString?.Get("displayProperty") ?? XRefSpec.NameKey);
-            xref.AltProperty = node.GetAttributeValue("altProperty", queryString?.Get("altProperty") ?? "fullname");
-            xref.Text = node.GetAttributeValue("text", node.GetAttributeValue("name", queryString?.Get("text")));
-            xref.Alt = node.GetAttributeValue("alt", node.GetAttributeValue("fullname", queryString?.Get("alt")));
+            xref.DisplayProperty = node.GetAttributeValue("displayProperty", StringHelper.HtmlEncode(queryString?.Get("displayProperty")) ?? XRefSpec.NameKey);
```
hellosnow: question: will DisplayProperty be encoded again in line 133?

superyyrrzz: L133 encodes the value with DisplayProperty as the key.
Seems DisplayProperty and AltProperty will not show directly in HTML, so they indeed needn't be encoded here.

PR #809, opened 2016-10-21
use new xref syntax when expanding mref

superyyrrzz:
@chenkennt @vwxyzh @hellosnow @qinezh @ansyral

### DIFF
--- a/src/docfx.website.themes/default/partials/class.header.tmpl.partial
+++ b/src/docfx.website.themes/default/partials/class.header.tmpl.partial
@@ -19,10 +19,10 @@
 {{#inheritedMembers}}
   <div>
   {{#definition}}
-    <xref href="{{definition}}" fullName="{{fullName.0.value}}" name="{{nameWithType.0.value}}"/>
+    <xref href="{{definition}}" altProperty="fullName" displayProperty="nameWithType"/>
   {{/definition}}
   {{^definition}}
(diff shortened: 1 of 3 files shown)


src/docfx.website.themes/default/partials/class.header.tmpl.partial
```diff
@@ -19,10 +19,10 @@
 {{#inheritedMembers}}
   <div>
   {{#definition}}
-    <xref href="{{definition}}" fullName="{{fullName.0.value}}" name="{{nameWithType.0.value}}"/>
+    <xref href="{{definition}}" altProperty="fullName" displayProperty="nameWithType"/>
   {{/definition}}
   {{^definition}}
-    <xref href="{{uid}}" fullName="{{fullName.0.value}}" name="{{nameWithType.0.value}}"/>
+    <xref href="{{uid}}" altProperty="fullName" displayProperty="nameWithType"/>
```
hellosnow: needn't we use {{}} like altProperty="{{fullName}}"?

superyyrrzz: No. By using altProperty=fullName, resolver will use "fullName" as the key to find the value from XRefSpec's dict. An alternative is alt="{{fullName}}".

---
In reply to: [84432531](https://github.com/dotnet/docfx/pull/809#discussion_r84432531) [](ancestors = 84432531)

qinezh:
:shipit:

PR #810, opened 2016-10-21
add html post processor.

vwxyzh:


src/Microsoft.DocAsCode.Build.Common/HtmlPostProcessor.cs
```diff
@@ -0,0 +1,75 @@
...
+    using System.IO;
+    using System.Linq;
+    using System.Text;
+
+    using HtmlAgilityPack;
+
+    using Microsoft.DocAsCode.Plugins;
+    using Microsoft.DocAsCode.Common;
+
+    internal sealed class HtmlPostProcessor : IPostProcessor
```
hellosnow: need we export the name?

vwxyzh: I think this is an internal post processor, and run force.

src/Microsoft.DocAsCode.Build.Common/HtmlPostProcessor.cs
```diff
@@ -0,0 +1,75 @@
...
+                        continue;
+                    }
+                    foreach (var handler in Handlers)
+                    {
+                        handler.Handle(document, tuple.Item, tuple.SrcRelativePath, tuple.RelativePath);
+                    }
+                    document.Save(filePath, Encoding.UTF8);
+                }
+            }
+            foreach (var handler in Handlers)
```
hellosnow: if there's no any .html file in menifest, should we skip calling complete?

superyyrrzz:
:shipit:

Issue #662, opened 2016-09-01
Multiple preprocessors


evil-shrike:
docfx supports template merging. It's useful for custom templates when we have not to reimplement all logic from default template in our custom one. But if I want to add my own preprocessor (to add a field into view model) I have to copy almost all default template items. 
For example if I want to extend `ManagedReference.html.primary.js` and created such the file in my template then standard login from default template won't run. So I have to copy default `ManagedReference.html.primary.js` and patch it. But then I also have to copy `common.js` and `ManagedReference.common.js`. So I'm quickly loosing merging capabilities and have to synchronize my template with every new version.
It would be nice to have an ability to add a preprocessor in custom template and keep standard preprocessors running.

superyyrrzz:
You can put your `ManagedReference.html.primary.js` in a folder, like `customized_template`, and specify template in `docfx.json` like `"template": ["default", "customized_template"]`. Then you needn't copy `common.js` and `ManagedReference.common.js`. You need to synchronize only `ManagedReference.html.primary.js` in new version, and this won't change frequently.

evil-shrike:
When I created `ManagedReference.html.primary.js` in my template `ManagedReference.html.primary.js` in default template stopped running.

superyyrrzz:
Your `ManagedReference.html.primary.js` will overwrite the one in default template, so you should copy all the codes in the default, then add your parts.

evil-shrike:
This is exactly what I talked about - quickly we have to copy everything.

superyyrrzz:
@evil-shrike Now we expose 2 methods for preprocessor file. You can find documents [here](https://github.com/dotnet/docfx/blob/dev/Documentation/tutorial/howto_create_custom_template.md#extension-for-preprocessor-file) This will ship with today's release v2.7

PR #811, opened 2016-10-21
separate default template for easy customization

superyyrrzz:
#662 
provide 2 functions for each type: preSteps() and postSteps, which defined in type.html.primary.extension.js. Users can change this file to modify the model.
@chenkennt @vwxyzh @hellosnow @ansyral @qinezh


qinezh:
:shipit:

src/docfx.website.themes/default/ManagedReference.html.primary.extension.js
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) Microsoft. All rights reserved. Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+exports.preSteps = function (model) {
+  return model;
```
hellosnow: question, how about process model directly and not return model itself?

superyyrrzz: As discussed, remain this as user may want to return a new model.

---
In reply to: [84629091](https://github.com/dotnet/docfx/pull/811#discussion_r84629091) [](ancestors = 84629091)

hellosnow:
:shipit:

PR #821, opened 2016-10-25
add anchor info for the warning message of invalid link

ansyral:
@vwxyzh @chenkennt @hellosnow @superyyrrzz @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/HostService.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/HostService.cs
@@ -224,6 +224,7 @@ namespace Microsoft.DocAsCode.Build.Engine
                     sources.Add(new LinkSourceInfo
                     {
                         Target = file,
+                        Anchor = anchor,
                         SourceFile = pair.Node.GetAttributeValue("sourceFile", null),
                         LineNumber = pair.Node.GetAttributeValue("sourceLineNumber", 0),
                     });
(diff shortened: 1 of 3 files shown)


src/Microsoft.DocAsCode.Build.Engine/HostService.cs
```diff
@@ -224,6 +224,7 @@ public MarkupResult ParseHtml(string html, FileAndType ft)
                     sources.Add(new LinkSourceInfo
                     {
                         Target = file,
+                        Anchor = anchor,
```
hellosnow: >anchor [](start = 33, length = 6)

Note whether we need to remove anchor info in RemoveDebugInfo, which just added by Zhenghui?

hellosnow:
:shipit:

superyyrrzz:
:shipit:

qinezh:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/HostService.cs
ansyral: this commit only add the anchor info into LinkSourceInfo, not add an attribute into the doc

superyyrrzz: Finally find it's introduced in [this](https://github.com/dotnet/docfx/commit/051548b6ecf6b4469a2dc56951154545ba309593) commit... This anchor is an intermediate attribute used in resolving xref, and has been removed [here](https://github.com/dotnet/docfx/commit/051548b6ecf6b4469a2dc56951154545ba309593#diff-7bc60e42b31ac047d2df1c6b05f100d5R450).

PR #829, opened 2016-10-26
improve ihtmlpostprocessor logs: add scopewrapper for handlers

ansyral:
@vwxyzh @herohua @superyyrrzz @hellosnow @qinezh @chenkennt

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Common/HtmlDocumentHandler.cs
+++ b/src/Microsoft.DocAsCode.Build.Common/HtmlDocumentHandler.cs
@@ -1,0 +1,47 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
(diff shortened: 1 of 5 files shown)


src/Microsoft.DocAsCode.Build.Common/HtmlDocumentHandler.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                Handle(document, manifestItem, inputFile, outputFile);
+            }
+        }
+
+        public Manifest PostHandleWithScopeWrapper(Manifest manifest)
+        {
+            string phase = this.GetType().Name;
+            using (new LoggerPhaseScope(phase, false))
+            using (new PerformanceScope(phase, LogLevel.Verbose))
```
hellosnow: >Verbose [](start = 56, length = 7)

minor: could remove? it has an overload method to call verbose as default log level.

hellosnow:
:shipit:

PR #849, opened 2016-11-02
fix bookmark: need transform path from relative path to pathFromRoot when validating bookmark

ansyral:
@superyyrrzz @vwxyzh @chenkennt @hellosnow @qinezh


hellosnow:
:shipit:

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Build.Common/ValidateBookmark.cs
```diff
@@ -40,7 +40,7 @@ public override void Handle(HtmlDocument document, ManifestItem manifestItem, st
                  let link = node.GetAttributeValue("href", null)
                  let index = link.IndexOf("#")
```
superyyrrzz: @vwxyzh Do we need to consider query string here?

PR #853, opened 2016-11-03
improve toc in default template

superyyrrzz:
1. fix toc item with empty href getting parent toc's href by mustache fallback...
2. fix breadcrumb rendering when current page occurs more than once in toc.

@chenkennt @hellosnow @qinezh @vwxyzh @ansyral

### DIFF
--- a/src/docfx.website.themes/default/toc.html.js
+++ b/src/docfx.website.themes/default/toc.html.js
@@ -7,7 +7,7 @@ exports.transform = function (model) {
 
   function transformItem(item, level) {
     // set to null incase mustache looks up
-    item.href = item.href || null;
+    item.topicHref = item.topicHref || null;
     item.tocHref = item.tocHref || null;
     item.name = item.name || null;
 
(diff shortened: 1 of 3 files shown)


src/docfx.website.themes/default/toc.html.js
```diff
@@ -6,6 +6,11 @@ exports.transform = function (model) {
   return model;
 
   function transformItem(item, level) {
+    // set to null incase mustache looks up
+    item.href = item.href || null;
+    item.tocHref = item.tocHref || null;
```
hellosnow: >tocHref [](start = 9, length = 7)

not sure whether to update topicHref also?

hellosnow: topicHref is not used in js, so may not update?

---
In reply to: [86301821](https://github.com/dotnet/docfx/pull/853#discussion_r86301821) [](ancestors = 86301821)

hellosnow:
:shipit:

src/docfx.website.themes/default/toc.html.js
superyyrrzz: It's not used in mustache template, so no need to prevent fallback.

PR #856, opened 2016-11-03
refine bookmark

ansyral:
1.  add link text
2.  use sourcefile info from markdown parser as the srcfile isn't correct in the case of token inclusion

@vwxyzh @chenkennt @superyyrrzz @hellosnow @qinezh


hellosnow:
:shipit:

superyyrrzz:
:shipit:

qinezh:
:shipit:

PR #868, opened 2016-11-08
REST - export xref map

hellosnow:
@superyyrrzz @qinezh @ansyral @chenkennt @vwxyzh


hellosnow:
need to merge with @superyyrrzz 's code

PR #876, opened 2016-11-09
remove the usage of FileModel.LocalPathFromRepoRoot

ansyral:
@vwxyzh @herohua @superyyrrzz @hellosnow @qinezh @chenkennt 

still keep the property "LocalPathFromRepoRoot" and the assignment of it.


superyyrrzz:
:shipit:

Issue #873, opened 2016-11-08
Simple config for replacing logo


roji:
#239 describes how to override the navbar partial in order to replace the logo (upper-left corner).

It seems like there should be a much easier config option for doing this - practically every documentation site out there will want to put their project's logo, and overriding template partials is dangerous for compatibility as the template changes. Would you consider docfx global metadata option(s) for managing this?

superyyrrzz:
Yes, we can provide a metadata `_appLogoPath` to replace `logo.svg` [here](https://github.com/dotnet/docfx/blob/27f7f55746dc48f0d7700205c52dff071b51427b/src/docfx.website.themes/default/partials/navbar.tmpl.partial#L13). Then by minor customizing the template to provide your own logo file, you can replace logo while keep the template compatible.

superyyrrzz:
Now in [DocFX 2.8](https://github.com/dotnet/docfx/releases/tag/v2.8), you can use `_appLogoPath` and `_appFaviconPath` in global metadata to customize it.

roji:
@superyyrrzz great! Should I leave this open for tracking splitting out the logo into its own separate partial?

superyyrrzz:
It's also implemented. See: https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/default/partials/navbar.tmpl.partial#L12

roji:
Oh awesome! Thanks!

PR #892, opened 2016-11-14
support customizing logo and favicon

superyyrrzz:
#873
@chenkennt @vwxyzh @hellosnow @ansyral @qinezh


hellosnow:
:shipit:

roji:
This is great, but note that it still imposes `class="svg"` (I'm not sure what that means in your CSS and whether it's possible to put in a png etc.).

One idea, probably in addition to this PR, is to extract the logo into its own separate partial, allowing people to override it with whatever they want without having to duplicate the rest of the navbar.

superyyrrzz:
@roji We use `svg` class to provide hover effect in [docfx.css](https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/default/styles/docfx.css#L88). Putting a png here can also work. BTW, it's nice to have a logo partial.

ansyral:
:shipit:

PR #907, opened 2016-11-18
Make [!VIDEO] output correctly responsive

DuncanmaMSFT:
Updating the rendered code for the video element to correctly handle being displayed at different widths. Previous code was a fixed 640x320 that wouldn't scale on smaller or larger width pages.

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.css
+++ b/src/docfx.website.themes/default/styles/docfx.css
@@ -807,6 +807,23 @@ pre > code .line-highlight {
   background-color: #ffffcc;
 }
 
+/* For Embedded Video */
+div.embeddedvideo {
+    padding-top: 56.25%;
+    position: relative;
+    width: 100%;
+}
+
+div.embeddedvideo iframe {
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    width: 100%;
+    height: 100%;
+}
+
 /* For printer */
 @media print{
   .article.grid-right {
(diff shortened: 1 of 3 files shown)


superyyrrzz:
@DuncanmaMSFT Please update unit tests. @fenxu Please help review.

fenxu:
The change is OK and can work in my test. Just need to update the test cases.

---
In reply to: [261432070](https://github.com/dotnet/docfx/pull/907#issuecomment-261432070) [](ancestors = 261432070)

src/docfx.website.themes/default/styles/docfx.css
```diff
@@ -807,6 +807,23 @@ pre > code .line-highlight {
...
+    position: relative;
+    width: 100%;
+}
+
+div.embeddedvideo iframe {
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
```
hellosnow: is it enough for width: 100%; and height: 100%; without top/left/right/bottom?

DuncanmaMSFT: No, the absolute position (and top/left/right/bottom) keep it in the right spot as the container resizes.

DuncanmaMSFT:
I updated the tests, sorry about missing that on the first commit, thanks!

DuncanmaMSFT:
So, the tests pass in VS locally... can someone explain the failure from above?

superyyrrzz:
Sorry this is caused by a problem in DocFX's own test case... Your change is good and I've merged it.

PR #927, opened 2016-11-25
REST - mark up swagger root other than children and tags

hellosnow:
@ansyral @chenkennt @vicancy @superyyrrzz @vwxyzh @qinezh


superyyrrzz:
:shipit:

PR #931, opened 2016-11-28
Add tags in toc.yml for subcommand merge

qinezh:
@vwxyzh @vicancy @superyyrrzz @hellosnow @ansyral @chenkennt

### DIFF
--- a/src/docfx/SubCommands/MetadataMerger.cs
+++ b/src/docfx/SubCommands/MetadataMerger.cs
@@ -13,6 +13,7 @@ namespace Microsoft.DocAsCode.SubCommands
     using Microsoft.DocAsCode.Build.Engine;
     using Microsoft.DocAsCode.Build.ManagedReference;
     using Microsoft.DocAsCode.DataContracts.Common;
+    using Microsoft.DocAsCode.DataContracts.ManagedReference;
     using Microsoft.DocAsCode.Plugins;
     using Microsoft.DocAsCode.Utility;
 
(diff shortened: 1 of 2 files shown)


src/docfx/SubCommands/MetadataMerger.cs
```diff
@@ -104,6 +109,53 @@ private static void MergeToc(MetadataMergeParameters parameters, string outputBa
...
+            if (content != null)
+            {
+                var metadata = content.Metadata;
+                object tagsObj;
+                if (metadata != null && metadata.TryGetValue("tags", out tagsObj))
+                {
+                    var tags = tagsObj as object[];
+                    if (tags != null)
+                    {
+                        var tagsList = tags.Select(tag => tag as string).Where(tag => tag != null).ToArray();
```
ansyral: >Select(tag => tag as string).Where(tag => tag != null) [](start = 44, length = 54)

.Where(tag => tag is string) ?

qinezh: in that way, tagsList would be a list of object instead of a list of string

qinezh:
Closed as the design is changed.

src/docfx/SubCommands/MetadataMerger.cs
ansyral: i see

---
In reply to: [89735853](https://github.com/dotnet/docfx/pull/931#discussion_r89735853) [](ancestors = 89735853)

PR #935, opened 2016-11-30
use css instead of <em> for param refs

superyyrrzz:
USER STORY 790938: [Reference] Inline code for language keywords and param refs from triple slash comments instead of `<strong>` and `<em>`
@vwxyzh @vicancy @hellosnow @ansyral @qinezh @chenkennt

### DIFF
--- a/src/docfx.website.themes/default/partials/class.header.tmpl.partial
+++ b/src/docfx.website.themes/default/partials/class.header.tmpl.partial
@@ -50,7 +50,7 @@
 {{#syntax.parameters}}
     <tr>
       <td>{{{type.specName.0.value}}}</td>
-      <td><em>{{{id}}}</em></td>
+      <td><span class="parametername">{{{id}}}</span></td>
       <td>{{{description}}}</td>
     </tr>
 {{/syntax.parameters}}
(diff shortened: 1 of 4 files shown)


src/docfx.website.themes/default/styles/docfx.css
```diff
@@ -85,6 +85,10 @@ h6 mark {
     margin-left: 5em;
 }
 
+.parametername {
```
hellosnow: >.parametername [](start = 0, length = 14)

span.parametername?

hellosnow:
:shipit:

src/docfx.website.themes/default/partials/class.header.tmpl.partial
```diff
@@ -50,7 +50,7 @@
 {{#syntax.parameters}}
     <tr>
       <td>{{{type.specName.0.value}}}</td>
-      <td><em>{{{id}}}</em></td>
+      <td><span class="parametername">{{{id}}}</span></td>
```
vicancy: ><span class="parametername"> [](start = 10, length = 28)

this is breaking change for default template, make sure to mention this change in our release note

vicancy:
:shipit:

src/docfx.website.themes/default/partials/class.header.tmpl.partial
superyyrrzz: 👌

PR #945, opened 2016-12-02
Implements disposable for fal.

vwxyzh:
@chenkennt @vicancy @superyyrrzz @ansyral @hellosnow @qinezh


src/Microsoft.DocAsCode.Common/FileAbstractLayer/FileAbstractLayer.cs
```diff
@@ -74,18 +81,39 @@ public void Copy(RelativePath sourceFileName, RelativePath destFileName)
...
             return mapping.Properties;
         }
 
         #endregion
 
+        #region IDisposable Support
+
+        private bool _disposed = false;
+
+        protected virtual void Dispose(bool disposing)
```
vwxyzh: dispose pattern

qinezh:
:shipit:

PR #946, opened 2016-12-02
Remove unneeded step in GitUtility

qinezh:
no need to parse remote origin url since it could be handled in template.

@superyyrrzz @vwxyzh @hellosnow @ansyral @vicancy @chenkennt


superyyrrzz:
:shipit:

PR #953, opened 2016-12-05
incremental: support postbuild phase

ansyral:
@vwxyzh @vicancy @superyyrrzz @hellosnow @qinezh @chenkennt

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/DocumentBuildParameters.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/DocumentBuildParameters.cs
@@ -10,48 +10,47 @@ namespace Microsoft.DocAsCode.Build.Engine
 
     public sealed class DocumentBuildParameters : MarshalByRefObject
     {
+        [IncrementalIgnore]
         public FileCollection Files { get; set; }
 
+        [IncrementalIgnore]
         public string OutputBaseDir { get; set; }
 
-        [IncrementalCheck]
         public ImmutableArray<string> ExternalReferencePackages { get; set; } = ImmutableArray<string>.Empty;
 
-        [IncrementalCheck]
         public ImmutableArray<string> XRefMaps { get; set; } = ImmutableArray<string>.Empty;
 
-        [IncrementalCheck]
         public ImmutableDictionary<string, object> Metadata { get; set; } = ImmutableDictionary<string, object>.Empty;
 
-        [IncrementalCheck]
         public FileMetadata FileMetadata { get; set; }
 
-        [IncrementalCheck]
         public ImmutableArray<string> PostProcessors { get; set; } = ImmutableArray<string>.Empty;
 
+        [IncrementalIgnore]
         public TemplateManager TemplateManager { get; set; }
 
-        // todo : move the check to the prehandle of postbuild phase
-        [IncrementalCheck]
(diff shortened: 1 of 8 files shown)


src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -454,6 +454,15 @@ private bool GetCanVersionIncremental()
                 Logger.LogVerbose($"Disable incremental build by force rebuild option.");
                 return false;
             }
+            if (_parameters.ApplyTemplateSettings != null)
```
hellosnow: >ApplyTemplateSettings [](start = 28, length = 21)

it's create by internal and won't be null?

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/DocumentBuildParameters.cs
```diff
@@ -31,8 +31,8 @@ public sealed class DocumentBuildParameters : MarshalByRefObject
 
         public TemplateManager TemplateManager { get; set; }
 
-        // todo : partial properties should check.
-        //[IncrementalCheck]
+        // todo : move the check to the prehandle of postbuild phase
+        [IncrementalCheck]
         public ApplyTemplateSettings ApplyTemplateSettings { get; set; }
```
vwxyzh: `ApplyTemplateSettings` contains absolute path, that's why I remove the attribute `IncrementalCheck`

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
ansyral: checked in Create Method

src/Microsoft.DocAsCode.Build.Engine/DocumentBuildParameters.cs
ansyral: i should use IncrementalIgnore Attribute as the customized ContractResolver would check the property recursively

src/Microsoft.DocAsCode.Build.Engine/DocumentBuildParameters.cs
```diff
@@ -10,47 +10,47 @@ namespace Microsoft.DocAsCode.Build.Engine
...
 
         public string RootTocPath { get; set; }
 
+        [IncrementalIgnore]
         public string TemplateDir { get; set; }
 
+        [IncrementalIgnore]
         public ImmutableDictionary<string, ChangeKindWithDependency> Changes { get; set; }
 
+        [IncrementalIgnore]
```
vicancy: >IncrementalIgnore [](start = 9, length = 17)

when force rebuild, should we always full build? What happens when IncrementalIgnore this property?

ansyral: when force rebuild, we'll always full build. the attribute IncrementalIgnore means that the property/field would be ignored when calculating config hash

---
In reply to: [90815153](https://github.com/dotnet/docfx/pull/953#discussion_r90815153) [](ancestors = 90815153)

Issue #227, opened 2016-03-21
Have the option (configurable in docfx.json) for search results to open on same tab


lidermanrony:
Currently search results are being opened on a new tab.

For a lot of sites it makes more sense to open the search result on the same tab (just like your typical search engine).

Using the back browser button should return the user back to search results page

saguiitay:
+1

pascalberger:
As a workaround, you can quite easily achieve this by overwriting the `docfx.js` file in a custom theme. See [How-to: Create A Custom Template](http://dotnet.github.io/docfx/tutorial/howto_create_custom_template.html).

In the custom theme remove the `.attr("target", "_blank")` on [this line](https://github.com/dotnet/docfx/blob/0213400ab3cfc4193548ed4c1bea3b760b7e1395/src/docfx.website.themes/default/styles/docfx.js#L243).

The drawback is that you'll need to keep the `docfx.js` file in your theme up to date with the one from the default theme.

lidermanrony:
Yeah that's what I ended up doing. Another issue with that is that the back
button doesn't return you to search results after you clicked on a link

lanthonyneville:
Can this be looked at please? This feature doesn't make a lot of sense to me .... its so easy in browsers these days to manually control whether to open in a new tab/window, but with the link target hard-coded you have no choice. 

For me, as I embed a DocFX site in another site via an IFrame, clicking search links means the user leaves the context of the parent site and it is not intuitive for them often how to get back.

The ability to control the link target via a meta-data property would be best. As others have said, making a whole new template just for this is not ideal.

lanthonyneville:
This is still an issue. Please can this be looked at.

superyyrrzz:
@lanthonyneville bot will not close this issue any more as it is marked as `feature-request`.

Back to this issue, #956 provides an implementation but back button cannot work. I think it need to change the search experience with concrete result URL to support this.

filzrev:
> Back to this issue, https://github.com/dotnet/docfx/pull/956 provides an implementation but back button cannot work.
> I think it need to change the search experience with concrete result URL to support this.

Modern browser supports `bfcache` (Back-forward cache) by default.

It seems to works as expected by removing `target="_blank"` attribute.
When `Back` event occurred.  search-results view states are automatically restored.

PR #956, opened 2016-12-05
Don't open search links in a new window/tab

pascalberger:
Don't open search links in a new window/tab.

If there's a possibility to access settings from `docfx.json` in the `docfx.js` file I could also implement it as a metadata switch. 

Please also mention if you expect a test case for this.

#227


qinezh:
Just like @lidermanrony said in #227 , how could you resolve the issue that the back
button doesn't return you to search results after you clicked on a link in this way?

pascalberger:
@qinezh That's a valid point and with the current implementation of the client search I don't see any quick workaround. Will close this PR as a consequence.

Issue #948, opened 2016-12-02
Search not working on root page


pascalberger:
### Title
I've build a website out of only conceptual documentation and added `"postProcessors": ["ExtractSearchIndex"]` in the `docfx.json` file. While search generally works, the required files are not loaded on the root `index.html` page but search box is shown.

### Functional impact
Entering something in the search box on the root `index.html` file doesn't start search.

### Minimal repro steps
1. Clone [Seed Project](https://github.com/docascode/docfx-seed) (Commit d3bf65df3e54297c8dd3ff86dfed09c34056461f)
2. Run `docfx --serve`
3. Open `http://localhost:8080/index.html`
4. Enter `foo` in the top right search box. No search is started

### Expected result
Search should work on all pages.

### Actual result
Search doesn't work on root `index.html`

### Further technical details
It seems like the JavaScript files required for client search to work are not loaded on the root `index.html` page.

Tested with DocFx 2.8.2.0

qinezh:
Thanks your feedback, the issue would be fixed in next release.

PR #961, opened 2016-12-05
Fix bug that full text search not works in index page

qinezh:
#948 
@ansyral @hellosnow @superyyrrzz @vicancy @vwxyzh @chenkennt

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -82,13 +82,11 @@ $(function () {
   // Support full-text-search
   (function () {
     var query;
-    var relHref = $("meta[property='docfx\\:rel']").attr("content");
-    if (relHref) {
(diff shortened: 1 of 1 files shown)


hellosnow:
:shipit:

src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -82,13 +82,11 @@ $(function () {
...
   (function () {
     var query;
-    var relHref = $("meta[property='docfx\\:rel']").attr("content");
-    if (relHref) {
-      var search = searchFactory();
-      search();
-      highlightKeywords();
-      addSearchEvent();
-    }
+    var relHref = $("meta[property='docfx\\:rel']").attr("content") || "";
```
qinezh: `rel` always exists in our default template. It seems a bug that jQuery return null when `rel`'s value is `""`

PR #962, opened 2016-12-05
DFM is not 100% compatible with GFM

pascalberger:
As stated [here](https://github.com/dotnet/docfx/blob/dev/Documentation/spec/docfx_flavored_markdown.md#docfx-flavored-markdown) DFM is not 100% compatible with GFM. Make documentation consistent.


hellosnow:
:shipit:

vicancy:
99%... should we list what are the 1% not compatible? or we should target for 100%? @vwxyzh

PR #963, opened 2016-12-05
Add source and region code attribute support.

### LINKED ISSUE #910: <code> elements should support source and region attributes
### Functional impact

Sandcastle supported a few XML doc comment extensions that were in heavy use by a lot of code. DocFX not supporting these extensions is inconvenient at best. In this case, though, the loss of the capability is really hard to deal with. Being able to do something similar with overwrite files isn't really a good solution, because it requires a LOT more effort and doesn't keep code documentation with the code.

### Minimal repro steps

You should be able to include something like the following in the triple slash comments.

```
<example>
<code source="..\Examples\SomeSourceFile.cs" region="Example 1" lang="csharp" title="Example 1"/>
</example>
```

### Expected result

The generated metadata should include the following.

```
  example:
  - >-
    <pre>

    <code>the source code found in the file ..\Examples\SomeSourceFile.cs in the region named "Example 1"</code>

    </pre>
```

### Actual result

The generated metadata only includes the following.

```
  example:
  - >-
    <pre>

    <code></code>

    </pre>
```

### Further technical details

See [Taming Sandcastle: A .NET Programmer’s Guide to Documenting Your Code](https://www.simple-talk.com/dotnet/net-tools/taming-sandcastle-a-net-programmers-guide-to-documenting-your-code/) for more on using these attributes.

I'd even be willing to work on this and provide a pull request if I knew this was functionality that would be accepted. I find not being able to use this feature with DocFX to be much more than inconvenient, and feel pretty strongly that this is something that should be supported.

wekempf:
Resolve #910

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs
@@ -26,18 +26,28 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
         private static Regex CommentIdRegex = new Regex(@"^(?<type>N|T|M|P|F|E):(?<id>" + idSelector + ")$", RegexOptions.Compiled);
         private static Regex LineBreakRegex = new Regex(@"\r?\n", RegexOptions.Compiled);
         private static Regex CodeElementRegex = new Regex(@"<code[^>]*>([\s\S]*?)</code>", RegexOptions.Compiled);
+        private static Regex RegionRegex = new Regex(@"^\s*#region\s*(.*)$");
+        private static Regex EndRegionRegex = new Regex(@"^\s*#endregion\s*.*$");
 
         private readonly ITripleSlashCommentParserContext _context;
         private readonly TripleSlashCommentTransformer _transformer = new TripleSlashCommentTransformer();
 
         public string Summary { get; private set; }
+
         public string Remarks { get; private set; }
+
         public string Returns { get; private set; }
+
         public List<ExceptionInfo> Exceptions { get; private set; }
+
         public List<LinkInfo> Sees { get; private set; }
+
         public List<LinkInfo> SeeAlsos { get; private set; }
+
         public List<string> Examples { get; private set; }
+
         public Dictionary<string, string> Parameters { get; private set; }
+
         public Dictionary<string, string> TypeParameters { get; private set; }
 
         private TripleSlashCommentModel(string xml, SyntaxLanguage language, ITripleSlashCommentParserContext context)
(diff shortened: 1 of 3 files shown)


src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs
```diff
@@ -218,6 +230,72 @@ private List<string> GetExamples(XPathNavigator nav, ITripleSlashCommentParserCo
...
+                {
+                    continue;
+                }
+
+                var region = node.Attribute("region");
+
+                var path = source.Value;
+                if (!Path.IsPathRooted(path))
+                {
+                    path = Path.Combine(context.Source?.BasePath ?? ".", path);
```
ansyral: >BasePath [](start = 56, length = 8)

path is relative to current comments' src file path?

wekempf: Good catch. Is the question what BasePath points at, or what the example source path should be relative to? I'll be honest, I'm not 100% sure what BasePath represents, but as is obvious from the unit test changes, I assumed it was the current comments' src file path. So, my intention was that the code would be relative to the current source.

In Sandcastle, this wasn't the case. The path was relative to the SHFB project. I always found that confusing, and since there's little chance of existing code not having to update these paths regardless, I thought I'd do what made sense to me. That is, assuming BasePath meant what I thought it did. If it didn't, then this behavior is probably closer to what Sandcastle did. I can change this part of the code if anyone thinks it needs to be different, though. Just let me know what it should be instead.

wekempf: Oh, and I realize ```Source?.BasePath ?? "."``` isn't 100% correct since "." is unlikely to be correct no matter where the intent is here. It's my understanding (could very well be wrong) that BasePath shouldn't ever be null when running in the actual tool, but it was always null in unit tests. I took a fast/cheap way out to ensure we didn't throw a NullReferenceException here, being fine with likely producing a warning message in this unexpected scenario. If that's not good enough, I can fix that however is deemed appropriate as well.

ansyral: sorry, my question might not be clear. i think you should use context.Source.Path, which means current comments' src file path. BTW, we have IsRelativePath() method in PathUtility.cs, which i think cover more cases like a url

---
In reply to: [91070662](https://github.com/dotnet/docfx/pull/963#discussion_r91070662) [](ancestors = 91070662)

ansyral: sorry, my question might not be clear. i think you should use context.Source.Path, which means current comments' src file path. Currently BasePath is not used.

ansyral: hi @wekempf, any update about the pr?

hellosnow:
@wekempf , do you have any updates for this pr?

vicancy:
Out of date, close this one and using #1494

PR #965, opened 2016-12-06
REST - Add overwrite logic for parameters

hellosnow:
@ansyral @chenkennt @vicancy @vwxyzh @superyyrrzz @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Common/ApplyOverwriteDocument.cs
+++ b/src/Microsoft.DocAsCode.Build.Common/ApplyOverwriteDocument.cs
@@ -14,7 +14,7 @@ namespace Microsoft.DocAsCode.Build.Common
 
     public abstract class ApplyOverwriteDocument : BaseDocumentBuildStep
     {
-        protected virtual MergerFacade Merger { get; set; } = new MergerFacade(
+        protected virtual MergerFacade Merger { get; } = new MergerFacade(
             new DictionaryMerger(
                 new KeyedListMerger(
                     new ReflectionEntityMerger())));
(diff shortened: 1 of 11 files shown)


src/Microsoft.DocAsCode.Common/EntityMergers/JObjectMerger.cs
```diff
@@ -0,0 +1,63 @@
...
+    public class JObjectMerger : MergerDecorator
+    {
+        public JObjectMerger(IMerger inner)
+            : base(inner)
+        {
+        }
+
+        public override void Merge(ref object source, object overrides, Type type, IMergeContext context)
+        {
+            var stopMerge = false;
```
vwxyzh: add `if (type == typeof(object))`?

src/Microsoft.DocAsCode.Build.Common/ApplyOverwriteDocument.cs
```diff
@@ -14,10 +14,10 @@ namespace Microsoft.DocAsCode.Build.Common
 
     public abstract class ApplyOverwriteDocument : BaseDocumentBuildStep
     {
-        private readonly MergerFacade Merger = new MergerFacade(
-                new DictionaryMerger(
-                    new KeyedListMerger(
-                        new ReflectionEntityMerger())));
+        protected virtual MergerFacade Merger { get; set; } = new MergerFacade(
```
vwxyzh: `get` only?

vicancy: >protected virtual [](start = 8, length = 17)

how about use MergerFacade factory classes?

hellosnow: Add todo and consider to add later

---
In reply to: [91022771](https://github.com/dotnet/docfx/pull/965#discussion_r91022771) [](ancestors = 91022771)

PR #966, opened 2016-12-06
support environment variable in fal.

vwxyzh:
@chenkennt @hellosnow @ansyral @vicancy @superyyrrzz @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Common/FileAbstractLayer/LinkFileReader.cs
+++ b/src/Microsoft.DocAsCode.Common/FileAbstractLayer/LinkFileReader.cs
@@ -3,6 +3,7 @@
 
 namespace Microsoft.DocAsCode.Common
 {
+    using System;
     using System.Collections.Generic;
     using System.Collections.Immutable;
     using System.IO;
@@ -29,7 +30,7 @@ namespace Microsoft.DocAsCode.Common
                     if (m.AllowMoveOut || localPath.ParentDirectoryCount == 0)
                     {
                         var physicalPath = Path.Combine(m.PhysicalPath, localPath.ToString());
-                        if (File.Exists(physicalPath))
+                        if (File.Exists(Environment.ExpandEnvironmentVariables(physicalPath)))
                         {
                             return new PathMapping(path, physicalPath) { Properties = m.Properties };
                         }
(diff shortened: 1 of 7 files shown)


ansyral:
:shipit:

superyyrrzz:
>         public bool Exists(RelativePath file)

Does this need to support env var?

---
Refers to: src/Microsoft.DocAsCode.Common/FileAbstractLayer/FileAbstractLayer.cs:49 in 20d7d60. [](commit_id = 20d7d6025e2511626e184c86bc6553813bf8329e, deletion_comment = False)

src/Microsoft.DocAsCode.Common/FileAbstractLayer/LinkFileReader.cs
```diff
@@ -29,7 +30,7 @@ public LinkFileReader(IEnumerable<PathMapping> mappings)
                     if (m.AllowMoveOut || localPath.ParentDirectoryCount == 0)
                     {
                         var physicalPath = Path.Combine(m.PhysicalPath, localPath.ToString());
-                        if (File.Exists(physicalPath))
+                        if (File.Exists(Environment.ExpandEnvironmentVariables(physicalPath)))
```
vicancy: >ExpandEnvironmentVariables [](start = 52, length = 26)

move expand into FileMapping?

vwxyzh: No, we cannot do it, otherwise restore from file mapping with different environment variables will break.

vicancy: saving with initial %env% while getting with expand?
feels not save to let the caller make the expanding

vwxyzh: op always change the %env% for every build. @herohua

vwxyzh:
>         public bool Exists(RelativePath file)

% is not allowed in RelativePath

---
In reply to: [265095340](https://github.com/dotnet/docfx/pull/966#issuecomment-265095340) [](ancestors = 265095340)

---
Refers to: src/Microsoft.DocAsCode.Common/FileAbstractLayer/FileAbstractLayer.cs:49 in 20d7d60. [](commit_id = 20d7d6025e2511626e184c86bc6553813bf8329e, deletion_comment = False)

PR #969, opened 2016-12-07
Fix bug in GitUtility

qinezh:


hellosnow:
:shipit:

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -60,7 +60,15 @@ public static GitRepoInfo GetRepoInfo(string directory)
         #region Private Methods
         private static GitDetail GetFileDetailCore(string filePath)
```
vicancy: >filePath [](start = 58, length = 8)

is it possible that filePath is a relative path and GetDirectoryName returns empty?

qinezh: nice catch, `GitUtility` should handle the case.

---
In reply to: [91219667](https://github.com/dotnet/docfx/pull/969#discussion_r91219667) [](ancestors = 91219667)

PR #977, opened 2016-12-08
fix bug: generating overload names from generic method.

vwxyzh:
@chenkennt @ansyral @hellosnow @vicancy @superyyrrzz @qinezh


superyyrrzz:
:shipit:

hellosnow:
:shipit:

PR #978, opened 2016-12-08
continue fixing bug: report file dependnecy shouldn't crash when link contains illegal characters

ansyral:


hellosnow:
:shipit:

PR #982, opened 2016-12-08
perf: add cache for hostserviceCreatorWithCache

ansyral:
when compared pre-release docfx with last release docfx, the perf shows that Load phase takes much more time. Add cache to fix it.

@vwxyzh @vicancy @hellosnow @qinezh @superyyrrzz @chenkennt

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
@@ -31,27 +31,13 @@ namespace Microsoft.DocAsCode.Build.Engine
         public override bool ShouldProcessorTraceInfo(IDocumentProcessor processor)
         {
             var key = $"trace-{processor.Name}";
-            bool result;
-            if (_cache.TryGetValue(key, out result))
-            {
-                return result;
-            }
-            result = ShouldProcessorTraceInfoCore(processor);
-            _cache[key] = result;
-            return result;
+            return _cache.GetOrAdd(key, ShouldProcessorTraceInfoCore(processor));
         }
 
         public override bool CanProcessorIncremental(IDocumentProcessor processor)
         {
             var key = $"canIncremental-{processor.Name}";
-            bool result;
-            if (_cache.TryGetValue(key, out result))
-            {
-                return result;
-            }
-            result = CanProcessorIncrementalCore(processor);
-            _cache[key] = result;
-            return result;
+            return _cache.GetOrAdd(key, CanProcessorIncrementalCore(processor));
         }
 
         public override HostService CreateHostService(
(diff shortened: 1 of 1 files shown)


src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
```diff
@@ -111,5 +116,29 @@ private void PostCreate(HostService hostService, IEnumerable<FileAndType> files)
                 IncrementalContext.ReportModelLoadInfo(hostService, loadedFiles, BuildPhase.PreBuildBuild);
             }
         }
+
+        private bool ShouldProcessorTraceInfoCore(IDocumentProcessor processor)
```
hellosnow: >ShouldProcessorTraceInfoCore [](start = 21, length = 28)

question: why the method is checking incremental, but naming with should processor trace info?

superyyrrzz:
:clock1:

src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
ansyral: if a processor, like resourceProcessor, doesn't support incremental, we won't save any intermediate files for it, which is the name's meaning.

qinezh:
:shipit:

ansyral:
close it as it turns out this isn't the root cause that Load becomes so much slower

ansyral:
after not calling gitutility in ResourceDocumentProcessor.Load, the perf is back to normal. still open this as it could improve perf when files count is large

src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
```diff
@@ -27,26 +30,28 @@ public HostServiceCreatorWithIncremental(DocumentBuildContext context) : base(co
 
         public override bool ShouldProcessorTraceInfo(IDocumentProcessor processor)
         {
-            if (!(processor is ISupportIncrementalDocumentProcessor))
+            var key = $"trace-{processor.Name}";
+            bool result;
+            if (_cache.TryGetValue(key, out result))
```
ansyral: nice

src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
```diff
@@ -111,5 +116,29 @@ private void PostCreate(HostService hostService, IEnumerable<FileAndType> files)
                 IncrementalContext.ReportModelLoadInfo(hostService, loadedFiles, BuildPhase.PreBuildBuild);
             }
         }
+
+        private bool ShouldProcessorTraceInfoCore(IDocumentProcessor processor)
+        {
+            if (!(processor is ISupportIncrementalDocumentProcessor))
+            {
+                Logger.LogVerbose($"Processor {processor.Name} cannot suppport incremental build because the processor doesn't implement {nameof(ISupportIncrementalDocumentProcessor)} interface.");
```
hellosnow: update the suppport to support? :)

---
In reply to: [91652735](https://github.com/dotnet/docfx/pull/982#discussion_r91652735) [](ancestors = 91652735)

hellosnow:
:shipit:

PR #994, opened 2016-12-12
Ignore derivedClasses temporarily

qinezh:


src/Microsoft.DocAsCode.DataContracts.ManagedReference/ItemViewModel.cs
```diff
@@ -279,7 +279,7 @@ public string FullNameForVB
         [JsonProperty("inheritance")]
         public List<string> Inheritance { get; set; }
 
-        [YamlMember(Alias = "derivedClasses")]
+        [YamlIgnore]
```
vwxyzh: Try `ApiBuildOutput.DerivedClasses`

PR #996, opened 2016-12-12
Refine merger definition in ApplyOverwriteDocument

hellosnow:
@vwxyzh @qinezh @chenkennt @superyyrrzz @ansyral @vicancy

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Common/ApplyOverwriteDocument.cs
+++ b/src/Microsoft.DocAsCode.Build.Common/ApplyOverwriteDocument.cs
@@ -14,12 +14,12 @@ namespace Microsoft.DocAsCode.Build.Common
 
     public abstract class ApplyOverwriteDocument : BaseDocumentBuildStep
     {
+        private readonly MergerFacade _merger;
+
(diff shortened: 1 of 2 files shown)


superyyrrzz:
:shipit:

qinezh:
:shipit:

qinezh:
one question: why choose `override method` instead of `override field` directly in this case?

hellosnow:
Fields cannot be overridden...

---
In reply to: [266609831](https://github.com/dotnet/docfx/pull/996#issuecomment-266609831) [](ancestors = 266609831)

qinezh:
Sorry, what I mean is Property `protected override MergerFacade Merger { get; }`

hellosnow:
Since want to reuse IMerger as MergerDecorator, to fill in MergerFacade, so the variable is IMerger instead of MergerFacade. For the GetMerger() method, I think it's both ok for override method or override property.

---
In reply to: [266618725](https://github.com/dotnet/docfx/pull/996#issuecomment-266618725) [](ancestors = 266618725)

PR #1002, opened 2016-12-15
Fix bug that see/seealso href without link text not shows in browser

qinezh:
example:

`<see href="http://bing.com"/>` in triple slash comments will be rendered as `<div><span><a href="http://bing.com"></a></span></div>
` and can't be displayed in browser.

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs
@@ -375,7 +375,7 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
                 }
                 else if (!string.IsNullOrEmpty(url))
                 {
-                    yield return new LinkInfo { AltText = altText, LinkId = url, LinkType = LinkType.HRef };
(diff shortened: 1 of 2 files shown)


hellosnow:
>                         yield return new LinkInfo { AltText = altText, LinkId = type, CommentId = commentId, LinkType = LinkType.CRef };

should this altText check null as well?

---
Refers to: src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs:373 in a13d857. [](commit_id = a13d857727738510b3b2dc98728b11d49bfc69e8, deletion_comment = False)

hellosnow:
:shipit:

qinezh:
>                         yield return new LinkInfo { AltText = altText, LinkId = type, CommentId = commentId, LinkType = LinkType.CRef };

no need, cref with comment id has its own fallback logic.

---
In reply to: [267251030](https://github.com/dotnet/docfx/pull/1002#issuecomment-267251030) [](ancestors = 267251030)

---
Refers to: src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs:373 in a13d857. [](commit_id = a13d857727738510b3b2dc98728b11d49bfc69e8, deletion_comment = False)

superyyrrzz:
:shipit:

vicancy:
:shipit:

PR #1018, opened 2016-12-19
try fix e2e

superyyrrzz:
Add some code to make sure the page has been loaded, then resuming checking.
An element will be waited up to 10 seconds: https://github.com/dotnet/docfx/blob/dev/test/docfx.E2E.Tests/DocfxSeedSiteFixture.cs#L33

### DIFF
--- a/test/docfx.E2E.Tests/DocfxSeedSiteTest.cs
+++ b/test/docfx.E2E.Tests/DocfxSeedSiteTest.cs
@@ -99,7 +99,14 @@ namespace Microsoft.DocAsCode.E2E.Tests
             // go to reference
             _driver.FindElement(By.LinkText("API Documentation")).Click();
 
+            // Make sure the namespace page has been loaded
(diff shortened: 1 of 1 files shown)


test/docfx.E2E.Tests/DocfxSeedSiteTest.cs
```diff
@@ -99,8 +99,15 @@ public void TestReferencePage()
             // go to reference
             _driver.FindElement(By.LinkText("API Documentation")).Click();
 
+            // Make sure the namespace page has been loaded
+            _driver.FindElement(By.XPath("//title[starts-with(.,'Namespace')]"));
+
+            // go to class page
             _driver.FindElements(By.XPath("//h4/a"))[0].Click();
 
+            // Make sure the class page has been loaded
```
hellosnow: >Make [](start = 15, length = 4)

minor: to be consistent, starts with lower case?

hellosnow:
:shipit:

PR #1026, opened 2016-12-20
add more perf logs

ansyral:


src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
```diff
@@ -209,14 +212,17 @@ select new
...
             {
-                var hostService = creator.CreateHostService(
-                    parameters,
-                    templateProcessor,
-                    MarkdownService,
-                    MetadataValidators,
-                    pair.processor,
-                    pair.item?.Select(f => f.file));
-                yield return hostService;
+                using (new LoggerPhaseScope(pair.processor.Name, true))
```
hellosnow: >pair.processor.Name [](start = 44, length = 19)

need we add some verb here?

hellosnow:
:shipit:

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
ansyral: already have a `Load` prefix

PR #1033, opened 2016-12-21
try fix e2e again

superyyrrzz:
From Selenium's comments:
```
    /// Increasing the implicit wait timeout should be used judiciously as it
    ///             will have an adverse effect on test run time, especially when used with
    ///             slower location strategies like XPath.
```

### DIFF
--- a/test/docfx.E2E.Tests/DocfxSeedSiteTest.cs
+++ b/test/docfx.E2E.Tests/DocfxSeedSiteTest.cs
@@ -100,13 +100,13 @@ namespace Microsoft.DocAsCode.E2E.Tests
             _driver.FindElement(By.LinkText("API Documentation")).Click();
 
             // make sure the namespace page has been loaded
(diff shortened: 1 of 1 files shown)


test/docfx.E2E.Tests/DocfxSeedSiteTest.cs
```diff
@@ -100,13 +100,13 @@ public void TestReferencePage()
             _driver.FindElement(By.LinkText("API Documentation")).Click();
 
             // make sure the namespace page has been loaded
-            _driver.FindElement(By.XPath("//title[starts-with(.,'Namespace')]"));
+            _driver.FindElement(By.Id("classes"));
```
hellosnow: >classes [](start = 39, length = 7)

from the comments, should it be namespace page?

superyyrrzz: Yes, classes is in namespace

hellosnow:
:shipit:

PR #1035, opened 2016-12-22
add info for yaml header parse fail.

vwxyzh:
@chenkennt @superyyrrzz @qinezh @hellosnow @ansyral @vicancy

### DIFF
--- a/src/Microsoft.DocAsCode.Dfm/Rules/DfmYamlHeaderBlockRule.cs
+++ b/src/Microsoft.DocAsCode.Dfm/Rules/DfmYamlHeaderBlockRule.cs
@@ -42,6 +42,7 @@ namespace Microsoft.DocAsCode.Dfm
             }
             catch (Exception)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Dfm/Rules/DfmYamlHeaderBlockRule.cs
```diff
@@ -42,6 +42,7 @@ public IMarkdownToken TryMatch(IMarkdownParser parser, IMarkdownParsingContext c
             }
             catch (Exception)
             {
+                Logger.LogInfo("Invalid yaml header.", file: context.File, line: context.LineNumber.ToString());
```
vicancy: also print exception message?

vwxyzh: I think no one understand the exception message.

vicancy:
:shipit:

hellosnow:
:shipit:

superyyrrzz:
:shipit:

qinezh:
:shipit:

PR #1039, opened 2016-12-22
ensure intermediate folder is full path in document builder.

vwxyzh:
@chenkennt @superyyrrzz @vicancy @hellosnow @ansyral @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
@@ -69,7 +69,14 @@ namespace Microsoft.DocAsCode.Build.Engine
                 Logger.LogVerbose($"\t{processor.Name} with build steps ({string.Join(", ", from bs in processor.BuildSteps orderby bs.BuildOrder select bs.Name)})");
             }
             _postProcessors = GetPostProcessor(postProcessorNames);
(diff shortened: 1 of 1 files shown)


src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -69,7 +69,14 @@ public class DocumentBuilder : IDisposable
                 Logger.LogVerbose($"\t{processor.Name} with build steps ({string.Join(", ", from bs in processor.BuildSteps orderby bs.BuildOrder select bs.Name)})");
             }
             _postProcessors = GetPostProcessor(postProcessorNames);
-            _intermediateFolder = intermediateFolder;
+            if (intermediateFolder != null)
+            {
+                if (intermediateFolder.Length == 0)
+                {
+                    _intermediateFolder = Path.GetFullPath(".");
```
qinezh: `_intermediateFolder` here would always be overwritten ?

vwxyzh: Path.GetFullPath("") will throw expection.

---
In reply to: [93723348](https://github.com/dotnet/docfx/pull/1039#discussion_r93723348) [](ancestors = 93723348)

hellosnow:
:shipit:

PR #1048, opened 2016-12-23
init fileAttributes: copy md5 from last if the file doesn't change

ansyral:


src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -557,6 +554,43 @@ private bool GetCanVersionIncremental(IncrementalStatus buildInfoIncrementalStat
...
+            {
+                foreach (var f in CurrentBuildVersionInfo.Dependency.GetAllDependentNodes())
+                {
+                    var p = RelativePath.TryParse(f);
+                    if (p == null)
+                    {
+                        continue;
+                    }
+                    var fullPath = PathUtility.GetFullPath(EnvironmentContext.BaseDirectory, p.RemoveWorkingFolder());
+                    if (!keys.Contains(f) && File.Exists(fullPath))
```
hellosnow: >File.Exists(fullPath) [](start = 45, length = 21)

question: in what case the file not exist? incorrect file reference?

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -557,6 +554,43 @@ private bool GetCanVersionIncremental(IncrementalStatus buildInfoIncrementalStat
...
+                    FullPath = f.FullPath,
+                    IsFromSource = true,
+                };
+            }
+            if (CurrentBuildVersionInfo.Dependency != null)
+            {
+                foreach (var f in CurrentBuildVersionInfo.Dependency.GetAllDependentNodes())
+                {
+                    var p = RelativePath.TryParse(f);
+                    if (p == null)
```
vicancy: when p == null, does it mean the dependency is corrupted?

ansyral: when collect dependency, we won't check whether the dependency is valid. for example, user could add an invalid path as link, this way, p could be null

---
In reply to: [93839673](https://github.com/dotnet/docfx/pull/1048#discussion_r93839673) [](ancestors = 93839673)

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
ansyral: yes. if the user writes a link that doesn't exist at all, we don't need to calculate the md5 info for it.

---
In reply to: [93815073](https://github.com/dotnet/docfx/pull/1048#discussion_r93815073) [](ancestors = 93815073)

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
vicancy: if we validate dependencies when saving them, will it make the dependency graph smaller?

ansyral: we can only check whether the path contains illegal character. for the case that path doesn't exist. we should still keep it considering the non-existed file might be added in the next run.

---
In reply to: [93843377](https://github.com/dotnet/docfx/pull/1048#discussion_r93843377) [](ancestors = 93843377)

ansyral: we can only check whether the path contains illegal character(not sure whether this is a common user error). for the case that path doesn't exist. we should still keep it considering the non-existed file might be added in the next run.

PR #1051, opened 2016-12-26
Move mref model.Uids from Load to Build

vicancy:


vwxyzh:
This also means: you cannot find file by uid in all prebuild.

vicancy:
Keep Uid in Load

PR #1057, opened 2016-12-26
Thread safety for LinkFileWriter.

vwxyzh:
@chenkennt @ansyral @hellosnow @vicancy @qinezh @superyyrrzz

### DIFF
--- a/src/Microsoft.DocAsCode.Common/FileAbstractLayer/LinkFileWriter.cs
+++ b/src/Microsoft.DocAsCode.Common/FileAbstractLayer/LinkFileWriter.cs
@@ -20,32 +20,44 @@ namespace Microsoft.DocAsCode.Common
         public override void Copy(PathMapping sourceFilePath, RelativePath destFilePath)
         {
             var key = destFilePath.GetPathFromWorkingFolder();
-            _mapping[key] = new PathMapping(key, sourceFilePath.PhysicalPath)
+            var pm = new PathMapping(key, sourceFilePath.PhysicalPath)
             {
                 Properties = sourceFilePath.Properties,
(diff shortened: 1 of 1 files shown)


hellosnow:
:shipit:

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Common/FileAbstractLayer/LinkFileWriter.cs
```diff
@@ -20,32 +20,44 @@ public LinkFileWriter(string outputFolder)
         public override void Copy(PathMapping sourceFilePath, RelativePath destFilePath)
         {
             var key = destFilePath.GetPathFromWorkingFolder();
-            _mapping[key] = new PathMapping(key, sourceFilePath.PhysicalPath)
+            var pm = new PathMapping(key, sourceFilePath.PhysicalPath)
             {
                 Properties = sourceFilePath.Properties,
             };
+            lock (_mapping)
```
ansyral: >lock (_mapping) [](start = 12, length = 15)

question: what's the advantage over directly using ConcurentDictionary?

vwxyzh: https://msdn.microsoft.com/en-us/library/dd287131(v=vs.110).aspx#Anchor_1
> The enumerator returned from the dictionary is safe to use concurrently with reads and writes to the dictionary, however it does not represent a moment-in-time snapshot of the dictionary. The contents exposed through the enumerator may contain modifications made to the dictionary after GetEnumerator was called.

PR #1059, opened 2016-12-26
Framework for document processors to restructure TOC

vicancy:
So that each Processor can register delegates to restructure TOC in its own processor

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandler.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandler.cs
@@ -3,8 +3,8 @@
 
 namespace Microsoft.DocAsCode.Build.Engine
 {
-    using System;
     using System.Collections.Generic;
+    using System.Collections.Immutable;
     using System.IO;
     using System.Linq;
 
@@ -13,6 +13,8 @@ namespace Microsoft.DocAsCode.Build.Engine
 
     internal class CompilePhaseHandler : IPhaseHandler
     {
+        private readonly List<RestructureTableOfContent> _delegates = new List<RestructureTableOfContent>();
+
         public string Name => nameof(CompilePhaseHandler);
 
         public BuildPhase Phase => BuildPhase.Compile;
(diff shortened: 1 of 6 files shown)


src/Microsoft.DocAsCode.Build.Engine/HostService.cs
```diff
@@ -123,6 +124,43 @@ public MarkupResult Markup(string markdown, FileAndType ft, bool omitParse)
...
+        {
+            if (methods == null || methods.Length == 0)
+            {
+                return;
+            }
+
+            var list = new List<RestructureTableOfContent>(methods);
+            lock (_tocSyncRoot)
+            {
+                if (RestructureTableOfContentDelegates != null)
```
hellosnow: >RestructureTableOfContentDelegates [](start = 20, length = 34)

it won't be null? has initialized?

qinezh:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/HostService.cs
vicancy: It is public accessible, so it can be set to null by callers.

hellosnow: get it, so it's better also check null in ShouldRestructureTableOfContent and InvokeRestructuringTableOfContent?

---
In reply to: [93891813](https://github.com/dotnet/docfx/pull/1059#discussion_r93891813) [](ancestors = 93891813)

src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandler.cs
```diff
@@ -38,8 +40,25 @@ public void Handle(List<HostService> hostServices, int maxParallelism)
                     {
                         Prebuild(hostService);
                     }
+
+                    // Register all the delegates to handler
+                    if (hostService.RestructureTableOfContentDelegates != null)
+                    {
+                        _delegates.AddRange(hostService.RestructureTableOfContentDelegates);
+                    }
```
ansyral: how about put these into Prepare()?

ansyral:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandler.cs
```diff
@@ -38,8 +40,25 @@ public void Handle(List<HostService> hostServices, int maxParallelism)
                     {
                         Prebuild(hostService);
                     }
+
+                    // Register all the delegates to handler
+                    if (hostService.RestructureTableOfContentDelegates != null)
+                    {
+                        _delegates.AddRange(hostService.RestructureTableOfContentDelegates);
```
hellosnow: >AddRange [](start = 35, length = 8)

question: if same delegate are added twice, we should invode twice?

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandler.cs
vicancy: yes, hostservice decides

hellosnow:
:shipit:

PR #1076, opened 2016-12-28
make perf logs verbose so we could get server logs

ansyral:
some perf logs are diagnostics, so when we do the perf test, loglevel=diagnostics might introduce another variable


hellosnow:
:shipit:

PR #1077, opened 2016-12-28
Reduce the number of times to call git command

qinezh:


src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -59,7 +59,31 @@ public static GitRepoInfo GetRepoInfo(string directory)
...
+        public static bool IsGitRoot(string directory)
+        {
+            if (!PathUtility.IsDirectory(directory))
+            {
+                throw new ArgumentException($"{directory} should be a directory");
+            }
+
+            if (!Directory.Exists(directory))
+            {
+                throw new FileNotFoundException($"{directory} can't be found");
```
vicancy: DirectoryNotFoundException

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -59,7 +59,31 @@ public static GitRepoInfo GetRepoInfo(string directory)
...
             }
 
-            return Cache.GetOrAdd(directory, GetRepoInfoCore);
+            if (IsGitRoot(directory))
+            {
+                return Cache.GetOrAdd(directory, GetRepoInfoCore);
+            }
+
+            var parentDirInfo = Directory.GetParent(directory);
+            return Cache.GetOrAdd(directory, GetRepoInfo(parentDirInfo?.FullName));
```
vicancy: when directory is root directory, exception will throw?

superyyrrzz: >GetRepoInfo [](start = 45, length = 11)

This recursion will cause more calling ExistGitCommand(). Extract a GetRepoInfoCore() or put this check somewhere else?

qinezh: Yes, `Directory.GetParent` will throw the exception.

---
In reply to: [94093186](https://github.com/dotnet/docfx/pull/1077#discussion_r94093186) [](ancestors = 94093186)

qinezh: why are there more calling ExistGitCommand()? I think it's the same as before.  However, I should reduce the number of times to call ExistGitCommand too...

---
In reply to: [94094370](https://github.com/dotnet/docfx/pull/1077#discussion_r94094370) [](ancestors = 94094370)

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -12,6 +12,7 @@ public static class GitUtility
     {
         private static readonly string CommandName = "git";
         private static readonly int GitTimeOut = 1000;
+        private static int _existGitCommand = -1;
```
vwxyzh: >int [](start = 23, length = 3)

bool?

vwxyzh:
:shipit:

PR #1085, opened 2016-12-29
change to not markup placeholder content

vicancy:
It feels more natural not marking up placeholder content behind the scene


superyyrrzz:
:shipit:

PR #1087, opened 2016-12-29
relay logs should include logs from tokens

ansyral:
file A includes a token B which isn't in docfx scope.
if B has warning, incremental build should also relay the warning


src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandlerWithIncremental.cs
```diff
@@ -89,6 +87,24 @@ private void PostHandle(List<HostService> hostServices)
             Logger.UnregisterListener(CurrentBuildMessageInfo.GetListener());
         }
 
+        private IEnumerable<string> GetFilesToRelayMessages(HostService hs)
```
superyyrrzz: >GetFilesToRelayMessages [](start = 36, length = 23)

Seems same with the one in LinkPhaseHandlerWithIncremental. Can it be reused?

superyyrrzz:
:shipit:

PR #1093, opened 2016-12-30
fix e2e sometimes failure

superyyrrzz:


test/docfx.E2E.Tests/DocfxSeedSiteTest.cs
```diff
@@ -281,8 +281,7 @@ public void TestXRefMap()
         private void TestPageCommon()
         {
             // check logo
-            IWebElement element = _driver.FindElement(By.Id("logo"));
-            Assert.Equal("svg", element.TagName);
+            Assert.NotEmpty(_driver.FindElements(By.Id("logo")));
```
hellosnow: >FindElements [](start = 36, length = 12)

question: the possible failure is caused by findElement or tag name? If not empty, will it fail by tage name != 'svg'?

superyyrrzz: In page source, it's a `img` tag with id = "logo". When finishing loading, it will become a `svg` tag. There's a chance to fail when tagName is still img.

superyyrrzz: In page source, it's a `img` tag with id = "logo". When finishing loading, it will become a `svg` tag. There's a chance to fail when tagName is still img.

hellosnow:
:shipit:

PR #1095, opened 2016-12-30
path should be case insensitive

ansyral:


hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildVersionInfo.cs
```diff
@@ -65,7 +66,7 @@ public class BuildVersionInfo
         /// deserialized attributes
         /// </summary>
         [JsonIgnore]
-        public IDictionary<string, FileAttributeItem> Attributes { get; set; } = new Dictionary<string, FileAttributeItem>();
+        public IDictionary<string, FileAttributeItem> Attributes { get; set; } = new Dictionary<string, FileAttributeItem>(FilePathComparer.OSPlatformSensitiveComparer);
```
vicancy: is OSPlatformSensitiveStringComparer enough? OSPlatformSensitiveComparer calculates FullPath while OSPlatformSensitiveStringComparer compares string directly

ansyral: you are right

Issue #1097, opened 2016-12-31
Getting branch name fails on AppVeyor


richardschneider:
### Title
*Getting the branch name fails on AppVeyor*  

### Functional impact
Seeing this error:

    Info: Can't find remote branch in this repo and fallback to use local branch [HEAD]: fatal: HEAD does not point to a branch

The `View source` link is wrong.  It is using `HEAD` instead of the branch that is being built.

### Minimal repro steps

I've cloned `docfx-seed`  at https://github.com/richardschneider/docfx-seed and created a branch named `msdn-xref` (see issue #1078).

- goto https://ci.appveyor.com/project/richardschneider/docfx-seed/build/1.0.5
- click `re-build commit`

### Further technical details

Most CIs, such as AppVeyor, use a git [detached head](https://www.google.co.nz/url?sa=t&rct=j&q=&esrc=s&source=web&cd=3&cad=rja&uact=8&ved=0ahUKEwiEjMPGtZ_RAhWIzLwKHQyECskQFggjMAI&url=https%3A%2F%2Fwww.git-tower.com%2Flearn%2Fgit%2Ffaq%2Fdetached-head-when-checkout-commit&usg=AFQjCNHu2MA59odXEKg_hTT-EweZa9gJLQ&sig2=RPTn-EzdlPgfmAvuVKtsfg) to build a new commit.

richardschneider:
`GitUtility.GetRepoInfoCore` needs to take a detach head into consideration.  Most CIs create an environment variable with the current branch name

qinezh:
Just like my comments in your PR, I don't think it's a bug in DocFX.

cameron-elliott:
I will leave my thoughts on this too, in case it is helpful.
I have run into the same problem using AppVeyor.


Richard> You might find a 'git checkout master' helps you, even if it is not a nice/clean solution.

Qinezh>
I don't personally like how Docfx trys scanning the file-system and git-repo in order to determine whether "Improve this document" links should be generated.
I wonder if "Microsoft.DocAsCode.Common/Git/*" could be removed, and 
the flags _gitContribute, _gitUrlPattern, and maybe another new flag could be used to specify git-repo linking without Docfx having to be able to read Git repo files, etc.
Just a crazy idea. :)

qinezh:
I submit a PR #1100  to fix the wrong branch name when doc repo in detached head. In this case, we'll use commit id instead of `HEAD` as the branch name.

richardschneider:
[...]

PR #1098, opened 2017-01-01
Branch name for git detached head

richardschneider:
Many build systems use a "detached head", which means that the normal git commands to get branch names do not work.  Thankfully, they set an environment variable.  The various environment variable names was obtained from https://github.com/GitTools/GitVersion

Bug #1097 

Here's an example of it now working on AppVeyor https://ci.appveyor.com/project/richardschneider/docfx-seed/build/1.0.8


qinezh:
Hi @richardschneider , thanks for your contribution. From my point, I don't think DocFX should handle the case. Consider a scenario, what if users just need generate the`view source` url with commit id instead of branch name? And, it seems a litter weird to add these specific logic in an utility class...Maybe you could add a post processor to it if you really need it.

My suggestion is that you need resolve the issue in CI tools instead of in DocFX. Such as running command `git checkout master` after step that CI has cloned you repo.

What do you think about it?

richardschneider:
`git checkout master` will not work because then I will be building the doc against the latest source code.  AppVeyor uses the detached head so the doc is built against the specified commit or a Pull Request.  An advantage is that the CI can be used to build from history.

AppVeyor is popular in open source dev because it uses a Windows VM and is free.  My workflow is simple: Build code, Test Code, Publish to Nuget, Generate Doc, publish to gh-pages. 

docfx does not work in this environment.  It has the wrong branch name in the generated .yml files.  How is this not a docfx bug?  

The logic was placed in the Utility class because this is where the bug is.  It uses `git rev-parse --abbrev-ref HEAD` to get the local branch name.  This command does not work in a detached head and always returns 'HEAD`.

qinezh:
???

richardschneider:
Did you read
https://github.com/dotnet/docfx/pull/1098#issuecomment-269916435

I'm of the opinion that this is docfx bug.

qinezh:
What if users just need generate the `view source` url with commit id instead of branch name in AppVeyor?

richardschneider:
@qinezh I don't understand your concern about `view source` with commit id..  

I am simply fixing the branch name for an edge case (a git detached head).  

Existing code will work as before.  This fix has nothing to do with a `commit id`.

vicancy:
The concern is that the code here greatly dependents on CI's implementation.
How about a environment variable for docfx to fallback to, e.g. DOCFX_SOURCE_BRANCH_NAME? @qinezh @richardschneider

richardschneider:
@vicancy good idea.  I'll add it add the head of list, so it overrides all other environment variables.

I want to keep the other env names so that that docfx works with other build systems without having to make to make any changes.

vicancy:
@richardschneider How about a section talking about setting up CI in `docfx` documentation? We can put that in Readme.md so that people can easily find that section. I would prefer removing those CI env names, considering that these names are prone to change, for instance, Jenkins already contains 2 env names.

richardschneider:
@vicancy excellent. I was just looking at folder structure to find out where to this info.  Could you tell me the folder/file to update.

vicancy:
1. Mention it in Readme here: https://github.com/dotnet/docfx/blob/dev/README.md
2. And add a new section in https://github.com/dotnet/docfx/blob/dev/Documentation/tutorial/docfx_getting_started.md. Or add a new file talking about it in detail and reference the file in https://github.com/dotnet/docfx/blob/dev/Documentation/tutorial/toc.yml

Really appreciate your contributions @richardschneider  😄 ,

Documentation/tutorial/docfx_getting_started.md
```diff
@@ -64,7 +64,27 @@ namespace WebApplication1
...
+---------------
+
+*DocFX* can be used in a Continuous Integration environment. 
+
+Most build systems do not checkout the branch that is being built, but 
+use a `detached head` for the specific commit.  DoxFX needs the the branch name to implement the `View Source` link in the API documentation.
+
+Setting the environment variable `DOCFX_SOURCE_BRANCH_NAME` tells DocFX which branch name to use.
+
+Many build systems set an environment variable with the branch name.  DocFX uses the following:
```
vicancy: Support `DOCFX_SOURCE_BRANCH_NAME` is enough? Use documentation to tell CI users how to configure environment variables with different build systems?

test/Microsoft.DocAsCode.Common.Tests/GitUtilityTest.cs
```diff
@@ -0,0 +1,39 @@
...
+    using Xunit;
+    using YamlDotNet.Core;
+
+    using Microsoft.DocAsCode.Common.Git;
+    using Microsoft.DocAsCode.YamlSerialization;
+
+    [Trait("Owner", "makaretu")]
+    public class GitUtilityTest
+    {
+
```
vicancy: remove extra empty lines

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -112,22 +124,35 @@ private static GitDetail GetFileDetailCore(string filePath)
...
-            try
+            if (localBranch != null)
             {
-                remoteBranch = RunGitCommandAndGetFirstLine(repoRootPath, GetRemoteBranchCommand);
-                var index = remoteBranch.IndexOf('/');
-                if (index > 0)
-                {
-                    remoteBranch = remoteBranch.Substring(index + 1);
-                }
+                remoteBranch = localBranch;
```
vicancy: LogInfo to tell users that branch name is found in env

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -112,22 +124,35 @@ private static GitDetail GetFileDetailCore(string filePath)
         private static GitRepoInfo GetRepoInfoCore(string directory)
         {
             var repoRootPath = RunGitCommandAndGetFirstLine(directory, GetRepoRootCommand);
-            var localBranch = GetLocalBranchName(repoRootPath);
-
+          
+            // Many build systems use a "detach head", which means that the normal git commands
+            // to get branch names do not work.  Thankfully, they set an environment variable.
+            string localBranch = BuildSystemBranchName
```
vicancy: what if the BuildSystemBranchName is found in env, however the repo is not in a detached mode?

Documentation/tutorial/docfx_getting_started.md
richardschneider: I prefer to have docfx run on common the build system without any special instructions.

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
richardschneider: The environment variable names are overrides.  If any is specified, then they specify the branch name.

vicancy:
Hi @richardschneider, the concern with your PR is that it introduces a new overwriting mechanism through Environment Variables for a specific case. 

Actually `docfx` allows overwriting URL for documents through `_gitContribute` in http://dotnet.github.io/docfx/tutorial/docfx.exe_user_manual.html#322-reserved-metadata.

To support your case, we plan to support overwriting the git url for source files, aka, `href` for `View Source` button. So after the change, in CI machines, passing %BRANCH_NAME% through `docfx` `--globalMatadata` command line can work.

richardschneider:
@vicancy I was just trying to get doxfx to work, as **expected**, with appyevor.  If you want developers to **jump through hoops** and spend hours googling to get it to work then go for it and document it. 

I thought I had a simple fix for this, but obviously I am wrong.

vicancy:
Hi @richardschneider I agree with you that the simplicity of setting up `docfx` on CI machine is quite important. After further discussion, we plan to support both overwriting from `globalMetadata` and Environment Variables. We will take your PR to support Environment Variables.
For the PR, do you think it is more reasonable to change the logic to:
1. If `DOCFX_SOURCE_BRANCH_NAME` is set, set branch name to `DOCFX_SOURCE_BRANCH_NAME`
2. If the repo is in *detached* mode, set branch name to CI env variable if it exists
3. Otherwise, use current commit id

So that `DOCFX_SOURCE_BRANCH_NAME` has the highest priority, and CI env variables only take effect when the repo is in detached mode.

Again, really appreciate your patience and continuous contribution to `docfx`. 😄

richardschneider:
Hi @vicancy and thanks for all comments.

> If DOCFX_SOURCE_BRANCH_NAME is set, set branch name to DOCFX_SOURCE_BRANCH_NAME

PR already does this.

> If the repo is in detached mode, set branch name to CI env variable if it exists

Okay, will make a change.

> Otherwise, use current commit id

PR already does this.

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -112,22 +123,47 @@ private static GitDetail GetFileDetailCore(string filePath)
...
-            try
+
+            // Many build systems use a "detached head", which means that the normal git commands
+            // to get branch names do not work.  Thankfully, they set an environment variable.
+            if (string.IsNullOrEmpty(localBranch))
             {
-                remoteBranch = RunGitCommandAndGetFirstLine(repoRootPath, GetRemoteBranchCommand);
-                var index = remoteBranch.IndexOf('/');
-                if (index > 0)
+                var isDetached = "HEAD" == RunGitCommandAndGetFirstLine(repoRootPath, GetLocalBranchCommand);
```
vicancy: > "HEAD" == Run [](start = 32, length = 14)

merge with GetLocalBranchName?

richardschneider: I think the addition of `GetLocalBranchName` is wrong.  It changes the semantics of both  `RepoInfo.LocalBranch` and `RepoInfo.RemoteInfo`.  Previously they held a git branch name.  Now they can also hold a `commit id`.

I want nothing to do with this.  Its **evil**.

Not sure what you mean by `"HEAD" == Run [](start = 32, length = 14)`

vicancy: However the code here is exactly `GetLocalBranchName`. And they have duplicate logic checking `detached`. Please make the code change cleaner rather than more complex.

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -112,22 +123,47 @@ private static GitDetail GetFileDetailCore(string filePath)
...
+                        .FirstOrDefault();
                 }
             }
-            catch (Exception ex)
+
+            if (!string.IsNullOrEmpty(localBranch))
             {
-                Logger.LogInfo($"Can't find remote branch in this repo and fallback to use local branch [{localBranch}]: {ex.Message}");
                 remoteBranch = localBranch;
+                Logger.LogInfo($"Using branch '{localBranch}' from the environment variable.");
```
vicancy: also log the environment variable name

superyyrrzz:
:shipit:

PR #1100, opened 2017-01-03
Fallback to use commit id when doc repo is in detached head

qinezh:


src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -143,6 +144,19 @@ private static GitRepoInfo GetRepoInfoCore(string directory)
...
 
+        private static string GetLocalBranchName(string repoRootPath)
+        {
+            var localBranch = RunGitCommandAndGetFirstLine(repoRootPath, GetLocalBranchCommand);
+
+            // Fallback to use commit id
+            if (localBranch.Equals("HEAD"))
+            {
+                localBranch = RunGitCommandAndGetFirstLine(repoRootPath, GetLocalBranchCommitIdCommand);
+                Logger.LogInfo("Fallback to use commit id as the branch name.");
```
vicancy: also tell user which repo fallbacks to use commit id

PR #1103, opened 2017-01-03
invalid path: add check for wildchar ?*

ansyral:


src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -573,12 +573,20 @@ private IEnumerable<FileItem> GetFilesToCalculateAttributes()
...
+                    try
+                    {
+                        var p = RelativePath.TryParse(f);
+                        if (p == null)
+                        {
+                            continue;
+                        }
+                        fullPath = PathUtility.GetFullPath(EnvironmentContext.BaseDirectory, p.RemoveWorkingFolder());
+                    }
+                    catch (Exception)
```
vwxyzh: IOException

PR #1104, opened 2017-01-03
Update packageSourceUrl to point to the folder in the repository

pascalberger:
Update packageSourceUrl of the Chocolatey Package to point to the folder in the repository where the source of the Chocolatey Package can be found.


qinezh:
See from chocolatey.nuspec https://github.com/chocolatey/choco/blob/master/nuget/chocolatey/chocolatey.nuspec#L15, the packageSourceUrl is just the same as projectUrl ?

pascalberger:
@qinezh eg. [Core Team Package Guidelines](https://github.com/chocolatey/chocolatey-coreteampackages/wiki/Contributing-guidelines#121-obligatory-metadata) also expect it to be to the directory inside of the repository where the .nuspec file is (but they have several packages in the same repo). But in this case I think it a little bit confusing, since it might not be clear to users that the package is maintained by DocFx itself, or if the packageUrl was just by mistake the same as the projectUrl. With the packageUrl pointing to the effectiv package source this would be clear.

PR #1105, opened 2017-01-03
relativepath could contain ?

ansyral:
path might contain query string


superyyrrzz:
Or check after `UriUtility.GetPath()`?

PR #1112, opened 2017-01-04
dependencytype: make isTransitive bool -> enum

ansyral:
file A include a token B, the dependency type should be transitive on all , namely, if token B contains a link to file C, it is equivalent to "A contains a link to file C"
another case about transitivity is to only transit a dependency of the same kind, for example , documentid

@fenxu @herohua @vwxyzh


src/Microsoft.DocAsCode.Plugins/DependencyType.cs
```diff
@@ -24,5 +25,27 @@ public class DependencyType
...
+        {
+            if (other == null)
+            {
+                throw new ArgumentNullException(nameof(other));
+            }
+            if (Transitivity == Plugins.Transitivity.All)
+            {
+                return true;
+            }
+            if ((Transitivity == Plugins.Transitivity.SameType || IsTransitive == true) && Name == other.Name)
```
fenxu: >IsTransitive == true [](start = 66, length = 20)

simplify to IsTransitive

src/Microsoft.DocAsCode.Plugins/DependencyType.cs
```diff
@@ -24,5 +25,27 @@ public class DependencyType
...
+        /// </summary>
+        public Transitivity? Transitivity { get; set; }
+
+        public bool CouldTransit(DependencyType other)
+        {
+            if (other == null)
+            {
+                throw new ArgumentNullException(nameof(other));
+            }
+            if (Transitivity == Plugins.Transitivity.All)
```
fenxu: >Plugins [](start = 32, length = 7)

Useless prefix?

src/Microsoft.DocAsCode.Plugins/DependencyType.cs
```diff
@@ -24,5 +25,27 @@ public class DependencyType
         /// the build phase that this type of dependency could have an effect on. [TO-DO]: remove the nullable ? when old value is overwritten
         /// </summary>
         public BuildPhase? Phase { get; set; }
+
+        /// <summary>
+        /// the transitivity of the dependency. [TO-DO]: remove the nullable ? when old value is overwritten
```
fenxu: >[TO-DO]: remove the nullable ? when old value is overwritten [](start = 48, length = 60)

For current behavior, if we assign default value to None, is it OK? So we don't need to add the nullable paramter.

ansyral: yes, you're right

src/Microsoft.DocAsCode.Plugins/DependencyType.cs
ansyral: compiler is confused about the type and variable name

ansyral: turns out just intellisense would get confused, removed

vwxyzh: rename type to DependencyTransitivity?

---
In reply to: [94542184](https://github.com/dotnet/docfx/pull/1112#discussion_r94542184) [](ancestors = 94542184)

fenxu:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/Incrementals/DependencyGraph.cs
```diff
@@ -177,20 +180,20 @@ private void RegisterDependencyTypeCore(DependencyType dt)
...
                 {
-                    // to-do: add check for phase when new value overwrites old value
+                    // to-do: add check for phase/transitivity when new value overwrites old value
                     if (stored.IsTransitive != dt.IsTransitive)
                     {
                         Logger.LogError($"Dependency type {JsonUtility.Serialize(dt)} isn't registered successfully because a different type with name {dt.Name} is already registered. Already registered one: {JsonUtility.Serialize(stored)}.");
                         throw new InvalidDataException($"A different dependency type with name {dt.Name} is already registered");
                     }
-                    if (stored.Phase != null)
+                    if (stored.Phase != null && stored.Transitivity == dt.Transitivity)
```
hellosnow: stored.Transitivity won't be null now?

PR #1114, opened 2017-01-04
Add TreeNavigator for TreeItems

vicancy:


src/Microsoft.DocAsCode.Plugins/TreeNavigator.cs
```diff
@@ -0,0 +1,144 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+
+    public class TreeNavigator
+    {
+        private NavigatorTreeItem _tree;
+        private NavigatorTreeItem _current;
+
+        public TreeNavigator(TreeItem tree)
```
superyyrrzz: >tree [](start = 38, length = 4)

check null?

PR #1120, opened 2017-01-05
link phase save output: manifest might contain multiple items with same path

ansyral:
@herohua

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/LinkPhaseHandlerWithIncremental.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/LinkPhaseHandlerWithIncremental.cs
@@ -214,7 +214,7 @@ namespace Microsoft.DocAsCode.Build.Engine
                     }
                     foreach (var path in items)
                     {
-                        // for redirection, path might be duplicate
+                        // path might be duplicate. for example, files with same name in different input folders are mapped to same output folder.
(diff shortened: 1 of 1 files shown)


hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/LinkPhaseHandlerWithIncremental.cs
```diff
@@ -235,7 +240,7 @@ select new
                             }
 
                             File.Copy(fullPath, Path.Combine(IncrementalContext.BaseDir, fileName));
-                            CurrentBuildVersionInfo.BuildOutputs.Add(path, fileName);
```
herohua: don't think this line needs update

src/Microsoft.DocAsCode.Build.Engine/LinkPhaseHandlerWithIncremental.cs
```diff
@@ -214,6 +214,11 @@ select new
                     }
                     foreach (var path in items)
                     {
+                        // for redirection, path might be duplicate
```
herohua: update the comment here

src/Microsoft.DocAsCode.Build.Engine/LinkPhaseHandlerWithIncremental.cs
ansyral: double assurance

PR #1123, opened 2017-01-06
add Execute method for logger file scope.

vwxyzh:
@chenkennt @vicancy @ansyral @hellosnow @herohua @qinezh @superyyrrzz


src/Microsoft.DocAsCode.Common/Loggers/LoggerFileScope.cs
```diff
@@ -51,6 +51,48 @@ public static LoggerFileScope Restore(object captured)
             return new LoggerFileScope(capturedScope.FileName);
         }
 
+        public void Execute(Action action)
+        {
+            try
+            {
+                action();
+            }
+            catch (Exception)
```
herohua: filter out DocumentException?

PR #1128, opened 2017-01-09
Fix bug in git utility cache

qinezh:
file path such as "../RELEASENOTE.md" need to be normalized before adding into cache.


hellosnow:
:shipit:

PR #1132, opened 2017-01-09
for xref in html, href is the encoded one while uid is the not encode…

vicancy:
…d one

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
@@ -82,7 +82,7 @@ namespace Microsoft.DocAsCode.Build.Engine
             }
             else
             {
-                xref.Uid = node.GetAttributeValue("uid", null);
+                xref.Uid = uid;
             }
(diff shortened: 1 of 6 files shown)


src/Microsoft.DocAsCode.Build.Engine/XrefDetails.cs
```diff
@@ -46,22 +46,28 @@ public static XRefDetails From(HtmlAgilityPack.HtmlNode node)
         {
             if (node.Name != "xref") throw new NotSupportedException("Only xref node is supported!");
             var xref = new XRefDetails();
-
-            var rawUid = GetRawUid(node);
+            var uid = node.GetAttributeValue("uid", null);
```
superyyrrzz: >var uid = node.GetAttributeValue("uid", null); [](start = 12, length = 46)

remove

superyyrrzz:
:shipit:

vicancy:
@DuncanmaMSFT Hi Duncan, is there any usage of `xref` in your template? Please take a look of the review if there is any.

PR #1153, opened 2017-01-11
Fix YAML serialization bug

qinezh:


hellosnow:
:shipit:

superyyrrzz:
:shipit:

vicancy:
The same concern with MSDN xref map, should we update it as well? @vwxyzh @qinezh ?

vwxyzh:
@vicancy let me do to tomorrow.

PR #1161, opened 2017-01-12
Only load the XSL once.

richardschneider:
A compiled XSL is thread safe.  Just load and compile in the static initialiser.

This change improved `docfx metadata` by 2 seconds.  Before it took ~12s now its down to 10s.

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentTransformer.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentTransformer.cs
@@ -12,14 +12,15 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
 
     using Microsoft.DocAsCode.Common;
     using Microsoft.DocAsCode.DataContracts.ManagedReference;
+    using System.Reflection;
 
     public class TripleSlashCommentTransformer
     {
-        private readonly XslCompiledTransform _transform;
+        private static readonly XslCompiledTransform _transform;
 
-        public TripleSlashCommentTransformer()
+        static TripleSlashCommentTransformer()
         {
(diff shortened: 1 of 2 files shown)


vicancy:
How about using Singleton?

richardschneider:
Yes, it could be just a `static` class.  Just didn't want to change the caller.  Will do later (in an hour or so)

src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentTransformer.cs
```diff
@@ -12,14 +12,15 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
...
     public class TripleSlashCommentTransformer
     {
-        private readonly XslCompiledTransform _transform;
+        private static readonly XslCompiledTransform _transform;
 
-        public TripleSlashCommentTransformer()
+        static TripleSlashCommentTransformer()
         {
-            var assembly = this.GetType().Assembly;
+            var assembly = Assembly.GetExecutingAssembly();
```
vwxyzh: `typeof(TripleSlashCommentTransformer).Assembly`?

src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentTransformer.cs
```diff
@@ -12,14 +12,15 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
 
     using Microsoft.DocAsCode.Common;
     using Microsoft.DocAsCode.DataContracts.ManagedReference;
+    using System.Reflection;
```
hellosnow: move namespace to 'System' group?

PR #1169, opened 2017-01-13
Improve extensibility for MREF default template

vicancy:


src/docfx.website.themes/default/ManagedReference.html.primary.tmpl
```diff
@@ -48,6 +48,9 @@
             {{#isEnum}}
               {{>partials/enum}}
             {{/isEnum}}
+            {{#_customContent}}
```
superyyrrzz: >_customContent [](start = 15, length = 14)

_isCustomContent?

src/docfx.website.themes/default/ManagedReference.html.primary.tmpl
```diff
@@ -48,6 +48,9 @@
...
+              {{>partials/customContent}}
```
superyyrrzz: >customContent [](start = 26, length = 13)

add an empty file?

vicancy: by default _customContent is false, so I don't think empty file is needed

vicancy: Or add an empty file instead of using `_customContent`

hellosnow:
:shipit:

PR #1171, opened 2017-01-13
Triple slash note

richardschneider:
Process the `<note>` XML comment as defined by [Sandcaslte](http://ewsoftware.github.io/XMLCommentsGuide/html/4302a60f-e4f4-4b8d-a451-5f453c4ebd46.htm).

Sorry about the messy commit history, but only **one** file is being changed.


src/Microsoft.DocAsCode.Metadata.ManagedReference/Transform/TripleSlashCommentTransform.xsl
```diff
@@ -111,6 +111,34 @@
...
+    <div>
+      <xsl:attribute name="class">
+        <xsl:choose>
+          <xsl:when test="$type = 'warning'">WARNING alert alert-warning</xsl:when>
+          <xsl:when test="$type = 'caution'">CAUTION alert alert-danger</xsl:when>
+          <xsl:when test="$type = 'important'">IMPORTANT alert alert-danger</xsl:when>
+          <xsl:otherwise>NOTE alert alert-info</xsl:otherwise>
+        </xsl:choose>    
+      </xsl:attribute>
+      <h5>
```
vwxyzh: this is the default behavior of DFM, but user can change it by [token.json](https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/default/token.json#L39)

richardschneider: But having a `<note>` in a triple slash comment did not work.  I'm assuming DFM never got called.

src/Microsoft.DocAsCode.Metadata.ManagedReference/Transform/TripleSlashCommentTransform.xsl
```diff
@@ -111,6 +111,34 @@
     </xsl:if>
   </xsl:template>
 
+  <xsl:template match="note">
+    <xsl:variable name="type">
+      <xsl:value-of select="normalize-space(@type)"/>
```
hellosnow: >normalize-space(@type) [](start = 28, length = 22)

should to lowercase to make it case insensitive?

richardschneider: The spec says it is lower case.  Don't really have to use `normalise-space` but was just following the code already present to process a `<list>`.

hellosnow: if so, update line 133 to lowercase the "Note" to "note", to be consistent with lowercase $type?

---
In reply to: [95945168](https://github.com/dotnet/docfx/pull/1171#discussion_r95945168) [](ancestors = 95945168)

src/Microsoft.DocAsCode.Metadata.ManagedReference/Transform/TripleSlashCommentTransform.xsl
ansyral: @vwxyzh  our Rule `DfmNoteBlockRule` only matched markdown style [!Note]

src/Microsoft.DocAsCode.Metadata.ManagedReference/Transform/TripleSlashCommentTransform.xsl
```diff
@@ -111,6 +111,34 @@
     </xsl:if>
   </xsl:template>
 
+  <xsl:template match="note">
+    <xsl:variable name="type">
```
ansyral: >type [](start = 24, length = 4)

be consistent with DFM behavior which uses Uppercase?

richardschneider: As I mentioned earlier, this is not DFM but an XML comment.  The spec uses lowercase.

ansyral: i meant that the output should be consistent, at html level?

---
In reply to: [96158289](https://github.com/dotnet/docfx/pull/1171#discussion_r96158289) [](ancestors = 96158289)

ansyral: i meant that the output should be consistent, at html level?

richardschneider: What is the issue?  I looked at the generated html for https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html and I think I'm consistent. But I could be wrong!

richardschneider:
@vwxyzh @ansyral @hellosnow Is there anything else I should do to get this PR approved?

src/Microsoft.DocAsCode.Metadata.ManagedReference/Transform/TripleSlashCommentTransform.xsl
ansyral: @richardschneider, the default behavior of DFM would transform to `<h5>NOTE</h5>`, but as @vwxyzh pointed, user can configure to use , for example, `<h6>NOTE</h6>` through `token.json`. Could the transformation of the xml comments also follow this?

richardschneider:
@ansyral said

>@richardschneider, the default behavior of DFM would transform to `<h5>NOTE</h5>`, but as 
> @vwxyzh pointed, user can configure to use , for example, `<h6>NOTE</h6`> through token.json. Could the transformation of the xml comments also follow this?

Okay I think I now understand.  You want me to generate the `<h5>...</h5>` from a predefined string.  Will look into it.

BTW: `token.json` is not in documentation.  Perhaps you should raise a documentation issue.

PS: Instead of emailing me directly, it should be sent to Github so that everyone knows.

richardschneider:
@vwxyzh Cannot implement `token.json` support.  Token processing is another assembly (Build Engine) that Metadata does not use.  Also, token processing relies upon a Resource Collection and Metadata doesn't have this concept.

richardschneider:
when #1214 is merged. I'll update documentation.

dotMorten:
@superyyrrzz @richardschneider 
I wonder if we could also add "Note" to the list? Seems to be used quite a lot in the Microsoft doc. Example:
https://learn.microsoft.com/en-us/dotnet/api/system.windows.window.windowstate?view=windowsdesktop-8.0#remarks
image

dotMorten:
oh wait nevermind... I just realized that is the fallback 🤦

PR #1175, opened 2017-01-13
cache files os platform case sensitive(II)

ansyral:


src/Microsoft.DocAsCode.Build.Engine/Incrementals/OSPlatformSensitiveDictionary.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Microsoft.DocAsCode.Build.Engine.Incrementals
```
superyyrrzz: copyright header

superyyrrzz:
:shipit:

PR #1184, opened 2017-01-16
incremental: only load dependnecyitems from last for unloaded files

ansyral:


hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/Incrementals/DependencyGraph.cs
```diff
@@ -27,6 +27,13 @@ public sealed class DependencyGraph
             },
             new DependencyType
             {
+                Name = DependencyTypeName.UidInclude,
```
vwxyzh: >UidInclude [](start = 42, length = 10)

rename to `OverwriteDocument`?

ansyral: how about `overwrite`? a little shorter:)

vicancy:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/Incrementals/DependencyGraph.cs
vwxyzh: Sound good!

---
In reply to: [96593710](https://github.com/dotnet/docfx/pull/1184#discussion_r96593710) [](ancestors = 96593710)

vwxyzh:
:shipit:

Issue #1200, opened 2017-01-18
c# region support for code snippets broken by #endregion with extra text


tobiasbunyan:
### Title
The regex for detecting #endregion in c# code snippet support expects to only have whitespace to the end of the line after #endregion.  If there is extra text the #endregion is not detected, resulting in an error.

### Functional impact
Documentation build fails.

### Minimal repro steps
1. Create a class with text after the endregion:
```
#region TestRegion
public class someclass
{
}
#endregion TestRegion
```
2. Reference this file in a conceptual document using a code snippet markdown:
```
[!code-csharp[Main](someclass.cs?name=TestRegion)]
```
3. Build the documentation

### Expected result
The code snippet is generated as normal.

### Actual result
Error occurs and build fails.

> Document.CompilePhaseHandler.ConceptualDocumentProcessor.Build.BuildConceptualDocument](articles/intro.md)Tag name TestRegion is not found	Web.Documentation	C:\Users\tbunyan\Source\Workspaces\Web.Documentation\EXEC

### Further technical details
Whilst adding text after an #endregion has no effect, it is not actually prohibited, and so should be allowed for.

Changing the CSharpCodeSnippetCommentEndLineRegex value in Microsoft.DocAsCode.Dfm.TagNameBlockPathQueryOption should alleviate this.

Untested suggestion:
From
```
@"^\s*#\s*endregion\s*$"
```
To
```
@"^\s*#\s*endregion\s?.*$"
```

vwxyzh:
@tobiasbunyan This bug should be fixed in current release version.

PR #1202, opened 2017-01-19
fix region support for code snippets broken by end region with extra text

vwxyzh:
@chenkennt @vicancy @ansyral @hellosnow @superyyrrzz @qinezh 
#1200


superyyrrzz:
:shipit:

Issue #1155, opened 2017-01-11
Full name in xrefmap.yml


pascalberger:
### Functional impact
It seems like name from yaml metadata inside contextual files is truncated to 30 characters in the `xrefmap.yml` file. Using cross-site linking afterwards results in truncated titles inside the documents.

### Minimal repro steps
1. Create conceptual document with the followin yaml header:
```
---
uid: someuid
title: Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus et. 
---
```
2. Run `docfx`
3. Check generated `xrefmap.yml`

### Expected result
I would expect full title in the `xrefmap.yml` file, or at least on option to define behavior

### Actual result
Title is truncated to 30 chars: 
```
- uid: someuid
  name: Lorem ipsum dolor sit amet, co...
  href: index.html
```

pascalberger:
Thanks @vicancy. This seems to work fine with 2.12

PR #1205, opened 2017-01-20
For #1155, it is more reasonable for front end to decide if name is t…

vicancy:
…runcated or not while docfx provides the full name


hellosnow:
:shipit:

qinezh:
:shipit:

PR #1212, opened 2017-01-21
add command to export dg

ansyral:


superyyrrzz:
:shipit:

src/docfx/SubCommands/DependencyCommand.cs
```diff
@@ -0,0 +1,61 @@
...
+            var buildInfo = BuildInfo.Load(intermediateFolder);
+            if (buildInfo == null)
+            {
+                Logger.LogWarning($"Cache files in the folder {intermediateFolder} are corrupted!");
+                return;
+            }
+            var dg = buildInfo.Versions.FirstOrDefault(v => v.VersionName == versionName)?.Dependency;
+            if (dg == null)
+            {
+                Logger.LogWarning($"Cache files for version {versionName} is not found!");
```
herohua: should be error

src/docfx/SubCommands/DependencyCommandCreator.cs
```diff
@@ -0,0 +1,17 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.SubCommands
+{
+    using Microsoft.DocAsCode;
+    using Microsoft.DocAsCode.Plugins;
+
+    [CommandOption("dependency", "Export dependency file")]
```
herohua: can we define a subcommand as `export`? I think there might be other cases we want to add into the `dependency` command

src/docfx/SubCommands/DependencyCommand.cs
```diff
@@ -0,0 +1,66 @@
...
+    using System.Linq;
+
+    using Microsoft.DocAsCode;
+    using Microsoft.DocAsCode.Build.Engine.Incrementals;
+    using Microsoft.DocAsCode.Build.Engine.Incrementals.Outputs;
+    using Microsoft.DocAsCode.Common;
+    using Microsoft.DocAsCode.Exceptions;
+    using Microsoft.DocAsCode.Plugins;
+
+    internal sealed class DependencyCommand : ISubCommand
```
hellosnow: >DependencyCommand [](start = 26, length = 17)

add some tests? refer to ArgsParserTest and BuildCommandTest

PR #1216, opened 2017-01-22
add CompositeDictionary.

vwxyzh:

### DIFF
--- a/src/Microsoft.DocAsCode.Common/CompositeDictionary.cs
+++ b/src/Microsoft.DocAsCode.Common/CompositeDictionary.cs
@@ -1,0 +1,245 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Common
+{
+    using System;
+    using System.Collections;
+    using System.Collections.Generic;
+    using System.Collections.Immutable;
+    using System.Linq;
+
+    public class CompositeDictionary
+        : IDictionary<string, object>
+    {
+
+        private readonly ImmutableArray<Entry> _entries;
+
+        private CompositeDictionary(ImmutableArray<Entry> entries)
+        {
+            _entries = entries;
+        }
+
+        public object this[string key]
+        {
+            get
+            {
+                if (key == null)
+                {
+                    throw new ArgumentNullException(nameof(key));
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Common/CompositeDictionary.cs
```diff
@@ -0,0 +1,245 @@
...
+                }
+                foreach (var entry in _entries)
+                {
+                    if (key.StartsWith(entry.Prefix))
+                    {
+                        entry.SetValue(key.Substring(entry.Prefix.Length), value);
+                        return;
+                    }
+                }
+            }
```
hellosnow: also throw new KeyNotFoundException() if not found?

hellosnow: I mean whether should set the value when the key not starts with prefix?

---
In reply to: [97214579](https://github.com/dotnet/docfx/pull/1216#discussion_r97214579) [](ancestors = 97214579)

src/Microsoft.DocAsCode.Common/CompositeDictionary.cs
```diff
@@ -0,0 +1,262 @@
...
+                foreach (var entry in _entries)
+                {
+                    if (key.StartsWith(entry.Prefix))
+                    {
+                        var pair = entry.TryGetValue(key.Substring(entry.Prefix.Length));
+                        if (pair.Key)
+                        {
+                            return pair.Value;
+                        }
+                        break;
```
hellosnow: >break [](start = 24, length = 5)

remove?

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Common/CompositeDictionary.cs
```diff
@@ -0,0 +1,262 @@
...
+                }
+                if (dict == null)
+                {
+                    throw new ArgumentNullException(nameof(dict));
+                }
+                if (valueConverter == null)
+                {
+                    valueConverter = o => (TValue)o;
+                }
+                _entries.Add(new Entry
```
ansyral: >_entries.Add [](start = 16, length = 12)

check whether the prefix already exists?

ansyral:
:shipit:

vicancy:
:shipit:

PR #1247, opened 2017-01-25
Cache and restore manifest in incremental post processing

hellosnow:
1. Manifest should be cached to prevent output files are modified during incremental post processing.
2. Change some test cases to use FAL.


superyyrrzz:
:shipit:

PR #1252, opened 2017-01-28
Open links to a different host in a new tab/window

### LINKED ISSUE #949: Open external links in a new window
Currently all links in markdown files are opened in the current window / tab. This behavior makes sense for internal links between different pages of the documentation. External links, where in the Markdown a full URL is provided (eg. `[see here](http://example.com)`) the link should have a `target=_blank`.

pascalberger:
Add JavaScript to rewrite links pointing to a different host to open in a new window/tab.

Fixes #949


qinezh:
@pascalberger E2E test has failed...

PR #1259, opened 2017-02-03
fix bug: markdown link rule is not allowed in link text.

vwxyzh:
@chenkennt @superyyrrzz @hellosnow


superyyrrzz:
:shipit:

hellosnow:
:shipit:

PR #1267, opened 2017-02-06
fix exception call stack in load.

vwxyzh:
@chenkennt @ansyral @hellosnow @superyyrrzz @qinezh

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
@@ -220,7 +220,7 @@ namespace Microsoft.DocAsCode.Build.Engine
             }
             catch (AggregateException ex)
             {
-                throw ex.InnerException;
+                throw new DocfxException(ex.InnerException?.Message, ex);
             }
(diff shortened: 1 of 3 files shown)


src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
```diff
@@ -220,7 +220,7 @@ private void Cleanup(HostService hostService)
             }
             catch (AggregateException ex)
             {
-                throw ex.InnerException;
+                throw new DocfxException(ex.InnerException?.Message, ex);
```
hellosnow: >ex [](start = 69, length = 2)

ex.InnerException?

hellosnow:
:shipit:

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
vwxyzh: If multi exceptions occurs, we should give all exceptions.

PR #1271, opened 2017-02-06
Also merge ItemViewModel's property for tocMetadata option

qinezh:


src/docfx/SubCommands/MetadataMerger.cs
```diff
@@ -139,8 +168,14 @@ private void CopyMetadataToToc(TocViewModel vm)
 
         private void CopyMetadataToTocItem(TocItemViewModel item)
         {
+            ApplyTocMetadata(item, _propTable);
+            ApplyTocMetadata(item, _metaTable);
```
superyyrrzz: switch these 2 lines to allow item's metadata overwrite page's metadata?

PR #1312, opened 2017-02-15
improve perf: set thread pool min threads.

vwxyzh:


ansyral:
:shipit:

PR #1324, opened 2017-02-21
Move ExtractSearchIndex post processor out 

vicancy:
Move ExtractSearchIndex post processor out from Microsoft.DocAsCode.Common. When plugins refrences to Microsoft.DocAsCode.Common, multiple Imports are found for ExtractSearchIndex. In general, Microsoft.DocAsCode.Common should not contain any Exports


hellosnow:
:shipit:

PR #1331, opened 2017-02-22
should CSS escape before using jQuery

superyyrrzz:


src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -622,6 +621,12 @@ $(function () {
         .replace(/&gt;/g, '>')
         .replace(/&amp;/g, '&');
     }
+
+    function cssEscape(str) {
+      // see: http://stackoverflow.com/questions/2786538/need-to-escape-a-special-character-in-a-jquery-selector-string#answer-2837646
+      return String(str)
```
vicancy: String(undefined) is `'undefined'`, is that intended?

superyyrrzz: Yes, the same with `htmlEncode()`.

superyyrrzz:
Note that this PR only partially solves the issue. Remaining issue: anchorjs will fail Bootstrap's [Scrollspy](https://v4-alpha.getbootstrap.com/components/scrollspy/) when the generated `id` contains `<` or `>`, because it has special meaning in CSS. See [jQuery's Selectors](http://api.jquery.com/category/selectors/).

PR #1343, opened 2017-02-23
fix bug in incremental build.

vwxyzh:
@chenkennt @vicancy @hellosnow @ansyral @qinezh @superyyrrzz @herohua


hellosnow:
:shipit:

PR #1344, opened 2017-02-23
improve error message for include

superyyrrzz:
before: `Absolute path "" is not supported.at line 3.`
after: `Unable to resolve "[!INCLUDE[vsprvslong]()]" at line 3: Absolute path "" is not supported.`

### DIFF
--- a/src/Microsoft.DocAsCode.Dfm/DocfxFlavoredIncHelper.cs
+++ b/src/Microsoft.DocAsCode.Dfm/DocfxFlavoredIncHelper.cs
@@ -56,7 +56,7 @@ namespace Microsoft.DocAsCode.Dfm
 
                 if (parents.Contains(currentPath, FilePathComparer.OSPlatformSensitiveComparer))
                 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Dfm/DocfxFlavoredIncHelper.cs
```diff
@@ -56,7 +56,7 @@ private string LoadCore(IMarkdownRenderer adapter, string currentPath, string ra
 
                 if (parents.Contains(currentPath, FilePathComparer.OSPlatformSensitiveComparer))
                 {
-                    return GenerateErrorNodeWithCommentWrapper("INCLUDE", $"Unable to resolve {raw}: Circular dependency found in \"{parent}\"", raw, sourceInfo);
+                    return GenerateErrorNodeWithCommentWrapper("INCLUDE", $"Circular dependency found in \"{parent}\"", raw, sourceInfo);
```
hellosnow: >Circular [](start = 76, length = 8)

also update some cases?

hellosnow:
:shipit:

PR #1348, opened 2017-02-24
modify manifest class.

vwxyzh:
@chenkennt @vicancy @ansyral @hellosnow @qinezh @superyyrrzz @herohua @fenxu

### DIFF
--- a/src/Microsoft.DocAsCode.Plugins/ManifestItemCollection.cs
+++ b/src/Microsoft.DocAsCode.Plugins/ManifestItemCollection.cs
@@ -1,0 +1,159 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Plugins
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Collections.ObjectModel;
+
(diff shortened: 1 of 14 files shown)


src/Microsoft.DocAsCode.Plugins/ManifestItemCollection.cs
```diff
@@ -0,0 +1,159 @@
...
+        public new ReadOnlyCollection<ManifestItem> AsReadOnly() => base.AsReadOnly();
+
+        [Obsolete]
+        public new int BinarySearch(ManifestItem item) => base.BinarySearch(item);
+
+        [Obsolete]
+        public new int BinarySearch(ManifestItem item, IComparer<ManifestItem> comparer) => base.BinarySearch(item, comparer);
+
+        [Obsolete]
+        public new int BinarySearch(int index, int count, ManifestItem item, IComparer<ManifestItem> comparer) => base.BinarySearch(index, count, item, comparer);
```
ansyral: why these methods are obsolete. seems strange considering the following case:
```
List<ManifestItem> = new ManifestItemCollection();
```
could it revoke the method or not?

vwxyzh: we will change the inheritance in next pr.

ansyral:
:shipit:

PR #1354, opened 2017-02-27
add serializable for VersionInfo, add VersionName to manifest

superyyrrzz:


src/Microsoft.DocAsCode.Plugins/OutputFileInfo.cs
```diff
@@ -26,6 +26,9 @@ public string RelativePath
             }
         }
 
+        [JsonProperty("version_name")]
+        public string VersionName { get; set; }
```
hellosnow: >VersionName [](start = 22, length = 11)

is it possible to add a mapping in root manifest?

src/docfx/Models/VersionConfig.cs
```diff
@@ -3,9 +3,12 @@
 
 namespace Microsoft.DocAsCode
```
vwxyzh: namespace

src/Microsoft.DocAsCode.Plugins/Manifest.cs
```diff
@@ -45,6 +45,9 @@ public Manifest(IEnumerable<ManifestItem> files)
         [JsonProperty("incremental_info")]
         public List<IncrementalInfo> IncrementalInfo { get; set; }
 
+        [JsonProperty("version_info")]
+        public Dictionary<string, VersionInfo> VersionInfo { get; set; }
```
hellosnow: >VersionInfo [](start = 47, length = 11)

VersionInfos?

superyyrrzz: want to align with `IncrementalInfo` above

hellosnow:
:shipit:

PR #1364, opened 2017-02-28
when uid is long, trim to avoid PathTooLong filename

vicancy:


plugins/Microsoft.DocAsCode.Build.MemberLevelManagedReference/SplitClassPageToMemberLevel.cs
```diff
@@ -22,6 +22,7 @@ public class SplitClassPageToMemberLevel : BaseDocumentBuildStep
         private const char Separator = '.';
         private const string SplitReferencePropertyName = "_splitReference";
         private const string IsOverloadPropertyName = "_isOverload";
+        private const int MaximumFileNameLength = 180;
```
vwxyzh: Can we disable it?

superyyrrzz:
:shipit:

qinezh:
:shipit:

plugins/Microsoft.DocAsCode.Build.MemberLevelManagedReference/SplitClassPageToMemberLevel.cs
vicancy: No, we cannot. Plan to remove this when FAL is enabled.

PR #1367, opened 2017-02-28
incremental:restructure modelmanifest to fit for split case

ansyral:
in prebuild step, models might be split. 
`ModelManifest`  used to record the mapping between a src file and its intermediate model file,
now it is used to record the mapping between src file and all the intermediate model files that are split.
namely, for example, a.md is split to a.md, a-1.md and a-2.md during prebuild step, then the mapping item would be a.md -> [`model for a.md`, `model for a-1.md`, `model for a-2.md`].

all other cache files like xrefmap, filemap, dependnecygraph also follow this way. namely, key is always src file path.


superyyrrzz:
:shipit:

vicancy:
:shipit:

PR #1369, opened 2017-03-01
allow duplicate file in manifest index.

vwxyzh:


vicancy:
:shipit:

PR #1371, opened 2017-03-01
enable incrementalbuild for managedreference plugins except splitmemberlevel

ansyral:
1.  some methods in IncrementalBuildContext aren't  thread-safe and when managedreferenceprocessor enables incremental, two processors run in parallel might have some problems, fix it.
2.  enable plugins incremental except for splitmemberlevel.

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
@@ -15,6 +15,7 @@ namespace Microsoft.DocAsCode.Build.Engine.Incrementals
 
     internal class IncrementalBuildContext
     {
+        private readonly object _sync = new object();
         private DocumentBuildParameters _parameters;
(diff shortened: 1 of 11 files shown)


hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -371,7 +375,10 @@ public ProcessorInfo CreateProcessorInfo(IDocumentProcessor processor)
                     IncrementalContextHash = ((ISupportIncrementalBuildStep)step).GetIncrementalContextHash(),
                 });
             }
-            CurrentBuildVersionInfo.Processors.Add(cpi);
+            lock (_sync)
+            {
+                CurrentBuildVersionInfo.Processors.Add(cpi);
```
superyyrrzz: >CurrentBuildVersionInfo [](start = 16, length = 23)

why should lock this and _modelLoadInfo>

ansyral: processors are run in parallel

vwxyzh:
:shipit:

PR #1375, opened 2017-03-02
dg: ignore self dependency

ansyral:
if `a` depends on itself, or depends on its uid, then ignore the dependencyitem


hellosnow:
:shipit:

PR #1376, opened 2017-03-02
Use os platform sensitive file name dictionary

vicancy:


plugins/Microsoft.DocAsCode.Build.MemberLevelManagedReference/SplitClassPageToMemberLevel.cs
```diff
@@ -85,7 +85,7 @@ private SplittedResult SplitModelToOverloadLevel(FileModel model)
             }
 
             // Make sure new file names generated from current page is unique
-            var newFileNames = new Dictionary<string, int>();
+            var newFileNames = new Dictionary<string, int>(FilePathComparer.OSPlatformSensitiveStringComparer);
```
hellosnow: >Dictionary [](start = 35, length = 10)

use OSPlatformSensitiveDictionary?

vicancy: It's currently inside Microsoft.DocAsCode.Build.Engine.dll and plugin don't want to take this dependency. We can move the dictionary to Microsoft.DocAsCode.Build.Common later and all the projects can reuse the dictionary. However I don't think declaring Dictionary with Comparer is heavy here and I tend to keep it this way for now.

hellosnow:
:shipit:

PR #1390, opened 2017-03-06
Token/Code snippets resolve failed should throw document exception and return non-zero result

fenxu:
as title


hellosnow:
:shipit:

PR #1396, opened 2017-03-08
add _versionName and _versionPath to systemMetadata

superyyrrzz:


src/Microsoft.DocAsCode.Build.Engine/SystemMetadata.cs
```diff
@@ -65,5 +65,17 @@ internal sealed class SystemMetadata
         /// </summary>
         [JsonProperty("_tocKey")]
         public string TocKey { get; set; }
+
+        /// <summary>
+        /// Current file's version name
+        /// </summary>
+        [JsonProperty("_versionName")]
```
fenxu: >_versionName [](start = 23, length = 12)

To keep same this should be _version?

src/Microsoft.DocAsCode.Plugins/IDocumentBuildContext.cs
```diff
@@ -74,5 +74,15 @@ public interface IDocumentBuildContext
         /// The Root Toc Path of current version
         /// </summary>
         string RootTocPath { get; }
+
+        /// <summary>
+        /// Current file's version name
+        /// </summary>
+        string VersionName { get; }
```
fenxu: >VersionName [](start = 15, length = 11)

so does this

fenxu:
:shipit:

PR #1398, opened 2017-03-09
use matcher in rules.

vwxyzh:


src/Microsoft.DocAsCode.MarkdownLite/MarkdownParsingContext.cs
```diff
@@ -3,25 +3,41 @@
 
 namespace Microsoft.DocAsCode.MarkdownLite
 {
+    using Matchers;
```
vicancy: >Matchers [](start = 10, length = 8)

fullname and order

src/Microsoft.DocAsCode.MarkdownLite/Basic/BlockRules/MarkdownCodeBlockRule.cs
```diff
@@ -3,35 +3,54 @@
 
 namespace Microsoft.DocAsCode.MarkdownLite
 {
-    using System.Text.RegularExpressions;
+    using Matchers;
```
vicancy: >Matchers [](start = 10, length = 8)

fullname...

PR #1399, opened 2017-03-09
fix bug for incremental: need to record tocrestruction info for split  case

ansyral:


hellosnow:
:shipit:

vwxyzh:
For temporary solution, I agree.

PR #1408, opened 2017-03-10
Conversion from .NET Core Tooling 1.0 Preview 2 to .NET Core Tooling 1.0 RTM

fubar-coder:
This project is a migration to .NET Core Tooling 1.0 RTM. All projects are building without error.


fubar-coder:
Scrap this PR. All this stuff doesn't work, because the unit tests aren't run under .NET Core and binding redirects aren't working with the new tooling. I'll investigate further.

PR #1409, opened 2017-03-13
Allow log null in JS

vicancy:


src/Microsoft.DocAsCode.Build.Engine/TemplatePreprocessors/TemplateJintPreprocessor.cs
```diff
@@ -53,14 +53,16 @@ public class TemplateJintPreprocessor : ITemplatePreprocessor
...
+
         private object _utilityObject;
         private static readonly object ConsoleObject = new
         {
-            log = new Action<object>(s => Logger.Log(s)),
-            info = new Action<object>(s => Logger.LogInfo(s.ToString())),
-            warn = new Action<object>(s => Logger.LogWarning(s.ToString())),
-            err = new Action<object>(s => Logger.LogError(s.ToString())),
-            error = new Action<object>(s => Logger.LogError(s.ToString())),
+            log = new Action<object>(s => Logger.Log(s ?? NullString)),
```
hellosnow: >Log [](start = 49, length = 3)

can we check in the logger? if it's null, not log anything?

hellosnow: >s [](start = 37, length = 1)

can we check s first, if it's null, not log? or else the log message is "null"?

vicancy: the change is to make the behavior of console.log the same as browser / node.js, with console.log(null) logs "null"

hellosnow:
:shipit:

PR #1410, opened 2017-03-13
validatebookmark for local link of files that are not in the root folder doesn't work

ansyral:
"C/d.html" + "" -> "C" rather than "C/d.html"


hellosnow:
:shipit:

PR #1438, opened 2017-03-19
add interface to load/save context info for plugins

ansyral:
1. add interface to load/save context info for plugins
2. plugin `FillReferenceInformation` to implement the interface

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandlerWithIncremental.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandlerWithIncremental.cs
@@ -73,6 +73,7 @@ namespace Microsoft.DocAsCode.Build.Engine
                                               select IncrementalUtility.GetDependencyKey(f.OriginalFileAndType),
                                                   FilePathComparer.OSPlatformSensitiveStringComparer);
             ReloadDependency(fileSet);
+            LoadContextInfo(hostServices);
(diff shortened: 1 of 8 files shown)


src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandlerWithIncremental.cs
```diff
@@ -88,6 +89,7 @@ private void PostHandle(List<HostService> hostServices)
                 h.SaveIntermediateModel(IncrementalContext);
             }
             IncrementalContext.UpdateBuildVersionInfoPerDependencyGraph();
+            SaveContextInfo(hostServices);
```
vwxyzh: why save when build completed?

ansyral: i'd like to allow plugin writer to decide which phase to load/save the context info. but since it might be confusing , remove it.

vwxyzh:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandlerWithIncremental.cs
ansyral: just find that some article might build incrementally in `Compile` phase while non-incrementally in `Link` phase, this way we have to have phase data

PR #1442, opened 2017-03-20
fix e2e hang when IO exception

superyyrrzz:


test/docfx.E2E.Tests/DocfxSeedSiteFixture.cs
```diff
@@ -48,14 +49,16 @@ public DocfxSeedSiteFixture()
...
                 WebApp.Start(Url, builder => builder.UseFileServer(fileServerOptions));
             }
             catch (System.Reflection.TargetInvocationException)
             {
                 Console.WriteLine($"Error serving \"{folder}\" on \"{Url}\", check if the port is already being in use.");
             }
-
+            catch (Exception e)
+            {
+                Console.WriteLine($"Error serving \"{folder}\" on \"{Url}\": {e}");
```
hellosnow: >e [](start = 78, length = 1)

e.message or whole e?

hellosnow:
:shipit:

test/docfx.E2E.Tests/DocfxSeedSiteFixture.cs
superyyrrzz: I think whole e containing calling stack and inner exception is more clear here.

PR #1444, opened 2017-03-21
Also dedent when code include the whole file

vicancy:
To keep the behavior consistent


superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Dfm/DfmCodeExtractor.cs
```diff
@@ -58,20 +63,25 @@ public DfmExtractCodeResult ExtractFencesCode(DfmFencesToken token, string fence
...
+                    FencesCodeLines = Dedent(includedLines, token.PathQueryOption.DedentLength)
                 };
             }
         }
 
+        private static string[] Dedent(IEnumerable<string> lines, int? dedentLength = null)
+        {
+            var length = dedentLength ??
+                               (from line in lines
+                                where !string.IsNullOrEmpty(line) && !string.IsNullOrWhiteSpace(line)
```
hellosnow: >!string.IsNullOrEmpty(line) [](start = 38, length = 27)

can be removed? when string is empty, string.IsNullOrWhiteSpace is true?

hellosnow:
:shipit:

PR #1462, opened 2017-03-22
improve matcher perf.

vwxyzh:


superyyrrzz:
:shipit:

hellosnow:
:shipit:

test/Microsoft.DocAsCode.MarkdownLite.Tests/PerfTest.cs
```diff
@@ -33,1116 +32,6 @@ public void TestPerf()
             GC.Collect();
         }
 
-        [Fact]
-        [Trait("Related", "Markdown")]
-        [Trait("Related", "Perf")]
```
ansyral: why remove?

vicancy:
>     public class PerfTest

How about moving perf test out to a benchmark project?

---
Refers to: test/Microsoft.DocAsCode.MarkdownLite.Tests/PerfTest.cs:17 in f2fb530. [](commit_id = f2fb530a95e80f4866f647e6b51f9b70a28b7401, deletion_comment = True)

test/Microsoft.DocAsCode.MarkdownLite.Tests/PerfTest.cs
vwxyzh: no use.

vwxyzh:
@vicancy I will do it later.

PR #1466, opened 2017-03-23
fix bug for saving plugin context info: streamwriter would overwrite a file even though nothing writes to the stream

ansyral:


src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -517,7 +517,7 @@ public void SaveContextInfo(HostService hostService, BuildPhase phase)
                 {
                     continue;
                 }
-                using (var writer = new StreamWriter(Path.Combine(Environment.ExpandEnvironmentVariables(BaseDir), stepInfo.ContextInfoFile)))
```
vicancy: Is append the expected behavior from the very beginning? Move using writer outside of the `foreach` loop instead?

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -479,22 +501,22 @@ public void LoadContextInfo(HostService hostService, BuildPhase phase)
             {
                 return;
             }
-            var traceContext = CreateTraceContext(hostService, phase);
+
             foreach (var step in hostService.Processor.BuildSteps.Where(s => s is ICanTraceContextInfoBuildStep))
```
vwxyzh: `OfType<ICanTraceContextInfoBuildStep>`

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -479,22 +501,22 @@ public void LoadContextInfo(HostService hostService, BuildPhase phase)
...
                 var stepInfo = lpi.Steps.Find(s => s.Name == step.Name);
                 if (stepInfo == null || stepInfo.ContextInfoFile == null)
                 {
                     continue;
                 }
-                using (var reader = new StreamReader(Path.Combine(Environment.ExpandEnvironmentVariables(LastBaseDir), stepInfo.ContextInfoFile)))
+                using (var stream = File.OpenRead(Path.Combine(Environment.ExpandEnvironmentVariables(LastBaseDir), stepInfo.ContextInfoFile)))
                 {
-                    ((ICanTraceContextInfoBuildStep)step).LoadFromContext(traceContext, reader);
+                    ((ICanTraceContextInfoBuildStep)step).LoadContext(stream);
```
vwxyzh: var step is `ICanTraceContextInfoBuildStep` now.

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.ManagedReference/FillReferenceInformation.cs
```diff
@@ -18,6 +18,7 @@ namespace Microsoft.DocAsCode.Build.ManagedReference
     [Export(nameof(ManagedReferenceDocumentProcessor), typeof(IDocumentBuildStep))]
     public class FillReferenceInformation : BaseDocumentBuildStep, ICanTraceContextInfoBuildStep
     {
+        private IEnumerable<SourceInfo> _lastContextInfo;
         private Dictionary<string, SourceInfo> _items = new Dictionary<string, SourceInfo>();
```
vicancy: >_items [](start = 47, length = 6)

All the ItemViewModel are stored to memory? That could be another 2G.....

vicancy: Only store what the buildstep needs?

---
In reply to: [107822615](https://github.com/dotnet/docfx/pull/1466#discussion_r107822615) [](ancestors = 107822615)

src/Microsoft.DocAsCode.Build.ManagedReference/FillReferenceInformation.cs
```diff
@@ -26,6 +27,7 @@ public class FillReferenceInformation : BaseDocumentBuildStep, ICanTraceContextI
 
         public override void Postbuild(ImmutableList<FileModel> models, IHostService host)
```
vicancy: >Postbuild [](start = 29, length = 9)

So PostBuild is safe, but for "build", if context is changed, build step should guarantee thread safety?

src/Microsoft.DocAsCode.Build.ManagedReference/FillReferenceInformation.cs
ansyral: storing the whole set of itemviewmodel is already in the stable bits.
if only store what the step needs, then it is all the itemviewmodel that the articles reference, but considering the following case: A references B, which would lead to a dependnecyitem: A->B. now if A updates to reference B and C, now we don't have C's model in our cache because C isn't changed.

ansyral: i got your point, i would only store the needed properties

src/Microsoft.DocAsCode.Build.ManagedReference/FillReferenceInformation.cs
ansyral: @herohua

---
In reply to: [107823555](https://github.com/dotnet/docfx/pull/1466#discussion_r107823555) [](ancestors = 107823555)

vicancy:
:shipit:

PR #1501, opened 2017-03-29
fix bug in markdown html rule.

vwxyzh:


test/Microsoft.DocAsCode.MarkdownLite.Tests/GfmTest.cs
```diff
@@ -805,6 +805,24 @@ public void TestTable_Html_Div()
 
         [Fact]
         [Trait("Related", "Markdown")]
+        public void TestInlineHtml_Span()
+        {
+            // 1. Prepare data
+            var source = @"
+<span>a</span>
```
yufeih: Consider adding a `<span data-ttu-id="abcd-123">b</span>` case to test attributes in html tags.

vwxyzh: test case updated.

hellosnow:
:shipit:

PR #1504, opened 2017-03-29
Add error code to duplicated uids

hellosnow:


src/Microsoft.DocAsCode.Build.Engine/ErrorCode.cs
```diff
@@ -7,6 +7,6 @@ internal static class ErrorCode
     {
         public const string InvalidInternalBookmark = "InvalidInternalBookmark";
         public const string InvalidExternalBookmark = "InvalidExternalBookmark";
-
+        public const string DuplicateUids = "DuplicateUids";
```
vicancy: This is not ErrorCode....

PR #1505, opened 2017-03-29
fix: warning of overwrite files are relayed multiple times

ansyral:


hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/LinkPhaseHandlerWithIncremental.cs
```diff
@@ -275,12 +275,21 @@ private void UpdateXrefMap(IEnumerable<HostService> hostServices)
 
         private void RelayBuildMessage(IEnumerable<HostService> hostServices)
```
vicancy: share with CompilePhaseHandlerWithIncremental?

PR #1525, opened 2017-04-05
Support a named global namespace

### LINKED ISSUE #1519: DocFX currently only supports generating metadata with namespace defined
I have a large code base with many types without a namespace. I understand from #417 that this is viewed as bad form, but I cannot change this about the structure of the code at this point.

I'm absolutely willing to add a solution for this, but would love to get feedback on the recommended change to support this. At a quick glance it looks like I could traverse the members of the global namespace in SymbolVisitorAdapter.cs instead of logging the warning.

jwittner:
Allows specifying an id for the global namespace and includes the global namespace in metadata generation if one is provided.

Resolves #1519

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference/Visitors/SymbolVisitorAdapter.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference/Visitors/SymbolVisitorAdapter.cs
@@ -133,13 +133,17 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
             item.Type = MemberType.Assembly;
             _references = new Dictionary<string, ReferenceItem>();
 
-            var typeMembers = symbol.GlobalNamespace.GetTypeMembers().Where(member => member.DeclaredAccessibility == Accessibility.Public);
-            if (typeMembers.Any())
+            IEnumerable<INamespaceSymbol> namespaces;
+            if (!string.IsNullOrEmpty(VisitorHelper.GlobalNamespaceId))
             {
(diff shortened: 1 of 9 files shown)


jwittner:
Can I just make a Team City account or something? To see why this test is failing.

src/Microsoft.DocAsCode.Metadata.ManagedReference/Visitors/SymbolVisitorAdapter.cs
```diff
@@ -133,13 +133,17 @@ public override MetadataItem VisitAssembly(IAssemblySymbol symbol)
             item.Type = MemberType.Assembly;
             _references = new Dictionary<string, ReferenceItem>();
 
-            var typeMembers = symbol.GlobalNamespace.GetTypeMembers().Where(member => member.DeclaredAccessibility == Accessibility.Public);
-            if (typeMembers.Any())
+            IEnumerable<INamespaceSymbol> namespaces;
+            if (!string.IsNullOrEmpty(VisitorHelper.GlobalNamespaceId))
             {
-                Logger.LogWarning($"DocFX currently only supports generating metadata with namespace defined. The following types in assembly \"{symbol.MetadataName}\" will have no metadata generated: {string.Join(", ", typeMembers.Select(m => m.MetadataName))}. ");
+                namespaces = Enumerable.Repeat( symbol.GlobalNamespace, 1);
```
superyyrrzz: >  [](start = 47, length = 1)

minor: remove

superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Metadata.ManagedReference/Visitors/SymbolVisitorAdapter.cs
jwittner: Fixed! =)

jwittner:
Thanks a ton for getting this in!

PR #1531, opened 2017-04-07
Adding snippet support for Java and Python

adkinn:
Not sure why line 1 comes up in the diff, must be just whitespace


superyyrrzz:
:shipit:

hellosnow:
:shipit:

Issue #1518, opened 2017-04-03
Support specific build configurations for metadata


jskeet:
As far as I can tell, the `docfx metadata` command configuration doesn't specify which project configuration or target framework to extract metadata for. In Noda Time we have some members which are present in one build but not another, and I'd really like to make this clear in the documentation... is there any way to either tell docfx which project configuration / tfm to build, or at least specify C# preprocessor symbols to define?

MichaCo:
Not sure if that's the same request @jskeet but I'd also love to see some indication in the docs of e.g. classes, for which target frameworks the code is available/visible..

At least IF you have multitarget projects, like NET40, NET45, netstandard14... and you ifdef parts of the code away, it would make sense to indicate that.

Advanced version could be something like the version switch at the top of the screen (MSDN docs).
The minimal solution could be an additional field near namespace and assembly? like:

    Namespace:MyNameSapce
    Assembly:Awesome.dll
    Target:.NET45, .netstandard1.3

jskeet:
@MichaCo: Yes, that's basically what I was intending to achieve, but I'd expected to have to do that with my own tooling, based on the results of running docfx twice. If docfx could do it itself, that would be even better :)

(I have something similar for which version a member was introduced in - see http://nodatime.org/2.0.x/api/NodaTime.Duration.html for example, and look at the "Since" point in each member.)

Issue #1254, opened 2017-01-30
Support for .NET Core Tooling Preview3+ projects


foxbot:
### Support for .NET Core Tooling Preview3+ projects
When building API documentation for a project that uses .NET Core Preview 3 or higher (csproj), the build fails, because (roslyn?) is unable to parse PackageReference tags

### Functional impact
API documentation can not be built for projects that use .NET Core

### Minimal repro steps
1. Create a docFx project
2. Add a metadata reference to a csproj file
### Expected result
The build would succeed and API documentation would be built.

### Actual result
During the 'generating metadata' phase of the build, a warning is thrown opening the project.
`Warning: Error opening project <project>.csproj: The attribute "Version" in element <PackageReference> is unrecognized.. Ignored.`

No API documentation is built.

### Further technical details
N/A

[Sample csproj/docfx.json for reference](https://gist.github.com/foxbot/961588dab3e70c0e0c7ffd8a241067e1)

superyyrrzz:
Yes, Roslyn should be updated to support it. Currently we are waiting for its official release.

superyyrrzz:
Seems Roslyn 2.0-rc4 can open new csproj, but fail to extract info from it.

cpx86:
Is there any work being done on this? VS2017 is just around the corner, and for libraries targeting the latest dotnet version, not being able to use DocFx is a bit of a problem.

dustinkingen:
I encountered this today too. I'm using .NET Core SDK 1.0 rc4 build 004771.

I created a new project:

```
C:\Code> mkdir MyApp\src\App

C:\Code> cd MyApp

C:\Code\MyApp> dotnet new console -o .\src\App
Content generation time: 42.4024 ms
The template "Console Application" created successfully.

C:\Code\MyApp> dir .\src\App
 Directory of C:\Code\MyApp\src\App

02/24/2017  08:34 PM    <DIR>          .
02/24/2017  08:34 PM    <DIR>          ..
02/24/2017  08:34 PM               170 App.csproj
02/24/2017  08:34 PM               177 Program.cs
```

Added docfx:

```
C:\Code\MyApp> docfx init -q
Created folder C:\Code\MyApp\docfx_project\src
Created folder C:\Code\MyApp\docfx_project\api
Created folder C:\Code\MyApp\docfx_project\apidoc
Created folder C:\Code\MyApp\docfx_project\articles
Created folder C:\Code\MyApp\docfx_project\images
[...]

PR #1549, opened 2017-04-14
Provide option for msbuild properties

vicancy:
for #1518 & #1254

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference/Incremental/CacheBase.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference/Incremental/CacheBase.cs
@@ -38,7 +38,7 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
             return GetConfig(key);
         }
 
-        public void SaveToCache(IEnumerable<string> inputProjects, IDictionary<string, List<string>> containedFiles, DateTime triggeredTime, string outputFolder, IList<string> fileRelativePaths, bool shouldSkipMarkup)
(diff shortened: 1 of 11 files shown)


qinezh:
:shipit:

src/Microsoft.DocAsCode.Metadata.ManagedReference/Incremental/CacheBase.cs
```diff
@@ -38,7 +38,7 @@ public BuildInfo GetValidConfig(IEnumerable<string> inputProjects)
             return GetConfig(key);
         }
 
-        public void SaveToCache(IEnumerable<string> inputProjects, IDictionary<string, List<string>> containedFiles, DateTime triggeredTime, string outputFolder, IList<string> fileRelativePaths, bool shouldSkipMarkup)
+        public void SaveToCache(IEnumerable<string> inputProjects, IDictionary<string, List<string>> containedFiles, DateTime triggeredTime, string outputFolder, IList<string> fileRelativePaths, bool shouldSkipMarkup, IDictionary<string, string> msbuildProperties)
```
vwxyzh: create a class for parameters?

vwxyzh:
:shipit:

src/Microsoft.DocAsCode.Metadata.ManagedReference/Incremental/CacheBase.cs
vicancy: will do refactor later

ericdrobinson:
Does this Issue perchance cover preprocessor symbol handling as suggested in #1518?

>or at least specify C# preprocessor symbols to define?

We have some APIs that are only generated for certain builds where certain Preprocessor Symbols are defined. It would be awesome if the same symbols specified when building assemblies could similarly affect the generated documentation (i.e. make generated documentation content match the generated code content)...

macrogreg:
The `merge` command described in the release notes for this PR is not actually working (https://github.com/dotnet/docfx/issues/3067, https://github.com/dotnet/docfx/issues/7106).

PR #1562, opened 2017-04-18
Also rename non-overload when its path exceeds 180

vicancy:


hellosnow:
:shipit:

superyyrrzz:
:shipit:

qinezh:
:shipit:

PR #1593, opened 2017-04-25
splitmembers: add first item's metadata to overload item

ansyral:


vicancy:
No, overload member should not copy metadata from the first method, it copies what defined in its own metadata in References section, https://github.com/dotnet/docfx/blob/dev/plugins/Microsoft.DocAsCode.Build.MemberLevelManagedReference/SplitClassPageToMemberLevel.cs#L315

ansyral:
make sense, i'll close the pr

PR #1594, opened 2017-04-25
Fix bug that wrong sha1 value was computed in some case

qinezh:


tools/Deployment/lib/chocolatey.ts
```diff
@@ -28,6 +29,10 @@ export class Chocolatey {
             return Promise.resolve();
         }
 
+        if (!fs.existsSync(assetZipPath)) {
```
vicancy: could this assetZip be an old version?

PR #1605, opened 2017-04-26
add case to repro incremental bug: if dest changed, incremental doesn't work

ansyral:
original `CompareDir` doesn't compare directory info


hellosnow:
:shipit:

qinezh:
:shipit:

PR #1606, opened 2017-04-27
add markup to templateUtility

superyyrrzz:


hellosnow:
:shipit:

src/Microsoft.DocAsCode.Build.Engine/TemplatePreprocessors/TemplateUtility.cs
```diff
@@ -3,7 +3,11 @@
 
 namespace Microsoft.DocAsCode.Build.Engine
 {
+    using System;
+
     using Microsoft.DocAsCode.Common;
+    using Microsoft.DocAsCode.Plugins;
+    using Microsoft.DocAsCode.Exceptions;
```
vicancy: order

src/Microsoft.DocAsCode.Build.Engine/TemplatePreprocessors/TemplateUtility.cs
```diff
@@ -54,5 +58,33 @@ public string GetHrefFromRoot(string originalHref, string sourceFileKey)
             }
             return file.UrlEncode().ToString() + UriUtility.GetQueryStringAndFragment(originalHref);
         }
+
+        public string Markup(string markdown, string sourceFileKey)
```
vicancy: sourceFileKey can be provided by us?

superyyrrzz: yes, added it to system metadata when adding templateUtility.ResolveSourceRelativePath

vicancy: so that no need to be a parameter to markup?

superyyrrzz: As it's hard to pass it here, so preserve as a parameter

---
In reply to: [113619942](https://github.com/dotnet/docfx/pull/1606#discussion_r113619942) [](ancestors = 113619942)

superyyrrzz:
Seems this feature is needed before BUILD(5/10), it should be removed later as it breaks incremental build's design.

vicancy:
Add a TODO if this is a temp solution

PR #1622, opened 2017-05-03
Initial version for pdf

vicancy:
Pdf as a postprocessor, require wkhtmltopdf set to PATH.
Need to specify in the *build* section of **docfx.json**:
```json
"build":{
...
   "postProcessors": "PdfConverterPostProcessor",
   "template": [
       "default",
       "pdfConverter\\tools"
   ]
}
```


plugins/Microsoft.DocAsCode.Build.PdfConverter/PdfConverterPostProcessor.cs
```diff
@@ -10,6 +10,7 @@ namespace Microsoft.DocAsCode.Build.PdfConverter
     using Microsoft.DocAsCode.Common;
     using Microsoft.DocAsCode.Exceptions;
     using Microsoft.DocAsCode.Plugins;
+    using System.IO;
```
superyyrrzz: group

plugins/Microsoft.DocAsCode.Build.PdfConverter/PdfConverterPostProcessor.cs
```diff
@@ -23,7 +24,32 @@ public class PdfConverterPostProcessor : IPostProcessor
...
+                DestDirectory = outputFolder,
+                GenerateAppendices = true,
+            };
+            var p = new ConverterImpl(options);
+            try
+            {
+                p.Convert(manifest, outputFolder);
+                return manifest;
+            }
+            catch (Exception e) when (e is IOException)
```
superyyrrzz: >Exception [](start = 19, length = 9)

why not catch IOException?

vicancy:
Use a subcommand instead of a postprocessor

PR #1625, opened 2017-05-04
add seealsoContent for udp

superyyrrzz:


hellosnow:
:shipit:

src/Microsoft.DocAsCode.DataContracts.Common/Constants.cs
```diff
@@ -48,6 +48,7 @@ public static class PropertyName
             public const string ExtensionMethods = "extensionMethods";
             public const string Overload = "overload";
             public const string Return = "return";
+            public const string SeeAlsoContent = "seealsoContent";
```
vicancy: seeAlsoContent?

src/Microsoft.DocAsCode.DataContracts.Common/Constants.cs
```diff
@@ -48,6 +48,7 @@ public static class PropertyName
...
 
             public const string SystemKeys = "_systemKeys";
```
vicancy: is _systemKey used in udp?

src/Microsoft.DocAsCode.DataContracts.Common/Constants.cs
superyyrrzz: want to be consistent with seealso

---
In reply to: [114701331](https://github.com/dotnet/docfx/pull/1625#discussion_r114701331) [](ancestors = 114701331)

src/Microsoft.DocAsCode.DataContracts.Common/Constants.cs
superyyrrzz: not used now. can add it later

---
In reply to: [114701341](https://github.com/dotnet/docfx/pull/1625#discussion_r114701341) [](ancestors = 114701341)

Issue #1554, opened 2017-04-14
Connection refused by server on Mono/Ubuntu.


kaby76:
### Title
docfx is unreachable in Ubuntu 16.x due to the call to self-hosting OWIN "WebApp.Start("localhost",...). I would recommend using "*:port" instead of "localhost:port" for the URL in the method call. (I have written a self-hosting NancyFX server, and it doesn't have the same problem.)

### Functional impact
I can execute "wget localhost:8080" from sh, but "wget actual-ip-addr:8080" on the server machine outputs a "connection refused". I also cannot talk to the server from another machine.

### Steps to reproduce
1) Create a vanilla Ubuntu 16.04.2 desktop in Virtualbox. Or you can use DigitalOcean and create a droplet with the same version.

2) Install Mono on the new box:
a)  sudo apt-get update
b)  sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
c)  echo "deb http://download.mono-project.com/repo/debian wheezy main" | sudo tee /etc/apt/sources.list.d/mono-xamarin.list
d)  sudo apt-get install mono-complete

3) Install current version of docfx.exe (Version 2.15.5)
a) sudo apt-get install zip unzip
b) cd
c) wget https://github.com/dotnet/docfx/releases/download/v2.15.5/docfx.zip
d) mkdir docfx; cd !$; mv ~/docfx.zip .; unzip docfx.zip

4) Get a complete BUILT source project. You can use my example, so you do not need VS C++ Build Tools:
a) cd; git clone https://github.com/kaby76/swigged-llvm-docfx.git

5) Run:
a) cd ~/docfx;
b) sudo mono docfx.exe serve -p 9000  ~/swigged-llvm-docfx/_site/

6) In a new terminal shell, try:
a) wget localhost:9000
-----> This works fine.
b) wget what-ever-the-ip-addr-si:9000     (substituting the IP of the machine into the command)
-----> Connecting to 192.168.1.11:9000... failed: Connection refused.

Connecting from another box gives the same problem. No firewalls enabled.

### Further technical details
[...]

PR #1638, opened 2017-05-06
Added '-n hostname' option for workaround of running docfx server on …

kaby76:
…Ubuntu with Mono. The default is 'localhost'. Note, alternative would be to hardwire '*' instead of 'localhost' in the two occurrences in the entire source, but this change is more general. Note, it's still hardwired into tools/DfmHttpService/Program.cs as 'localhost' because there isn't any arg parsing there, and  I don't want to write that.

This change fixes Issue 1554. I tested the change built binaries (using VS 2017.1), Mono 4.8.1, on Ubuntu 16.04:

$sudo mono docfx.exe serve -p 9000 -n '*' /home/ken/swigged-llvm-docfx/_site/
[sudo] password for ken: xxxxxxxxxxxxxxxxxxxx
Serving "/home/ken/swigged-llvm-docfx/_site/" on http://*:9000

.......

$ wget 10.0.2.15:9000
--2017-05-06 15:58:32--  http://10.0.2.15:9000/
Connecting to 10.0.2.15:9000... connected.
HTTP request sent, awaiting response... 200 OK
Length: 6708 (6.6K) [text/html]
Saving to: ‘index.html.6’

index.html.6        100%[===================>]   6.55K  --.-KB/s    in 0s      

2017-05-06 15:58:32 (1015 MB/s) - ‘index.html.6’ saved [6708/6708]

### DIFF
--- a/tools/DfmHttpService/Program.cs
+++ b/tools/DfmHttpService/Program.cs
@@ -19,7 +19,7 @@ namespace DfmHttpService
                     new ExitHandler()
                 });
 
-            var service = new DfmHttpServer(handler, args.Length > 0 ? args[0] : null);
+            var service = new DfmHttpServer(handler, "localhost", args.Length > 0 ? args[0] : null);
             service.Start();
             service.WaitForExit();
         }
--- a/src/docfx/Models/BuildCommandOptions.cs
+++ b/src/docfx/Models/BuildCommandOptions.cs
@@ -54,6 +54,9 @@ namespace Microsoft.DocAsCode
         [Option('s', "serve", HelpText = "Host the generated documentation to a website")]
         public bool Serve { get; set; }
 
+        [Option('n', "hostname", HelpText = "Specify the hostname of the hosted website (e.g., 'localhost' or '*')")]
+        public string Host { get; set; }
+
         [Option('p', "port", HelpText = "Specify the port of the hosted website")]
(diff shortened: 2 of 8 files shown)


superyyrrzz:
:shipit:

tools/DfmHttpService/Program.cs
```diff
@@ -19,7 +19,7 @@ public static void Main(string[] args)
                     new ExitHandler()
                 });
 
-            var service = new DfmHttpServer(handler, args.Length > 0 ? args[0] : null);
+            var service = new DfmHttpServer(handler, "localhost", args.Length > 0 ? args[0] : null);
```
vicancy: Should we use "" instead of "localhost" so that it can run under Linux? #1554

kaby76: As I mentioned, DfmHttpServer/Program.cs does not do arg "parsing" per se. It passes arg[0] directly into the function DfmHttpServer(). The better thing would be to parse the args like docfx/Program.cs does. That was a bit of machinery I didn't feel comfortable doing, e.g., -n '*' 5000. Or, -n '*' -p 5000?

vicancy: DfmHttpServer is the background server used by vscode extension. Let's keep it "localhost" for now.

PR #1640, opened 2017-05-08
Add pdfcommand

vicancy:
PdfCommandOptions inherits from BuildCommandOptions instead of using the result from 'build', considering that PDF uses a totally different template from normal build.


vwxyzh:
:shipit:

PR #1643, opened 2017-05-08
Return exit code 1 when ignore to sync in the third week

qinezh:


tools/Deployment/gulpfile.js
```diff
@@ -187,15 +187,15 @@ gulp.task("publish:chocolatey", () => {
...
+gulp.task("syncBranchCore", () => {
     Guard.argumentNotNullOrEmpty(config.docfx.repoUrl, "config.docfx.repoUrl", "Can't find docfx repo url in configuration.");
     Guard.argumentNotNullOrEmpty(config.docfx.home, "config.docfx.home", "Can't find docfx home directory in configuration.");
     Guard.argumentNotNullOrEmpty(config.sync.fromBranch, "config.sync.fromBranch", "Can't find source branch in sync configuration.");
     Guard.argumentNotNullOrEmpty(config.sync.targetBranch, "config.sync.targetBranch", "Can't find target branch in sync configuration.");
 
     if (Common.isThirdWeekInSprint()) {
         console.log("Ignore to sync in the third week of a sprint");
-        process.exit(0);
+        process.exit(1);
```
vicancy: does it mean an error? this case does not sound like an error to me

PR #1668, opened 2017-05-18
fix bug: dfm code fences should not work in paragraph.

vwxyzh:


vicancy:
Also re emphasis this in our docs?

PR #1683, opened 2017-05-23
incremental: move template hash check to link phase

ansyral:


src/Microsoft.DocAsCode.Build.Engine/LinkPhaseHandlerWithIncremental.cs
```diff
@@ -65,7 +65,14 @@ private void PreHandle(List<HostService> hostServices)
         {
             using (new LoggerPhaseScope("ReloadModelsPerChanges", LogLevel.Verbose))
             {
-                ReloadModelsPerChanges(hostServices);
+                if (IncrementalContext.IsTemplateUpdated)
+                {
+                    ReloadModels(hostServices);
```
hellosnow: question: for what case we need to move template hash check to link phase?

ansyral: for perf consideration. if mustache template is updated, no need to fully build.

PR #1690, opened 2017-05-24
enable incremental by default

ansyral:


src/docfx/SubCommands/BuildCommand.cs
```diff
@@ -286,6 +286,7 @@ internal static void MergeOptionsToConfig(BuildCommandOptions options, BuildJson
             {
                 config.IntermediateFolder = options.IntermediateFolder;
             }
+            config.IntermediateFolder = config.IntermediateFolder ?? Path.Combine(Path.GetTempPath(), "inter");
```
hellosnow: how about prepend "docfx" such as ""docfx-inter"?

vicancy: >Path.GetTempPath(), "inter" [](start = 82, length = 27)

using obj folder under base dir as default?

src/docfx/SubCommands/BuildCommand.cs
```diff
@@ -286,6 +286,7 @@ internal static void MergeOptionsToConfig(BuildCommandOptions options, BuildJson
             {
                 config.IntermediateFolder = options.IntermediateFolder;
             }
+            config.IntermediateFolder = config.IntermediateFolder ?? Path.Combine(optionsBaseDirectory,"obj", "docfx", "build");
```
vicancy: >optionsBaseDirectory [](start = 82, length = 20)

should be using the directory of docfx.json

PR #1692, opened 2017-05-24
Fix bug that collection is modified when iterate

vicancy:


superyyrrzz:
What modification may occur? Will the Count change?

vicancy:
Nice catch, there should not be any modification. The issue happens when a same model is handled in different threads simultaneously.

PR #1699, opened 2017-05-26
log warning for invalid inline code snippet.

vwxyzh:


src/Microsoft.DocAsCode.Dfm/Rules/DfmFencesInlineRule.cs
```diff
@@ -34,6 +38,10 @@ public override IMarkdownToken TryMatch(IMarkdownParser parser, IMarkdownParsing
             var title = match.Groups["title"]?.Value;
             var pathQueryOption = ParsePathQueryString(match.Groups["option"]?.Value, match.Groups["optionValue"]?.Value);
 
+            if (!parser.Context.GetIsInTable())
+            {
+                Logger.LogWarning("Invalid inline code.", line: sourceInfo.LineNumber.ToString(), code: WarningCodes.Markdown.InvalidInlineCodeSnippet);
```
vicancy: more descriptive warning message, e.g. "Inline code snippet is only allowed inside tables"

PR #1702, opened 2017-05-26
normalize path for manifest index.

vwxyzh:


src/Microsoft.DocAsCode.Plugins/Manifest.cs
```diff
@@ -166,24 +164,30 @@ private void OutputFileInfoPropertyChanged(object sender, PropertyChangedEventAr
...
         private void RemoveItem(string relativePath, OutputFileInfo item)
         {
-            if (_index.TryGetValue(relativePath, out List<OutputFileInfo> list))
+            if (_index.TryGetValue(NormalizePath(relativePath), out List<OutputFileInfo> list))
             {
                 list.Remove(item);
             }
         }
 
+        private static string NormalizePath(string relativePath)
```
hellosnow: >NormalizePath [](start = 30, length = 13)

can use string extension ToNormalizedPath()?

vwxyzh: It is in Common project, Common depend on Plugins...

hellosnow:
:shipit:

Issue #1708, opened 2017-05-28
metadata command not working on Linux (Mono)


dlech:
### Functional impact

`docfx metadata` is not usable on Linux.

### Minimal repro steps

* Ubuntu 16.04
* Mono from [Mono repository](http://www.mono-project.com/download/#download-lin-ubuntu)

        mono --version
        Mono JIT compiler version 5.0.1.1 (2017-02/5077205 Thu May 25 09:19:18 UTC 2017)
* [docfx v2.17.7](https://github.com/dotnet/docfx/releases/tag/v2.17.7)
* Clone [docfx-seed repo](https://github.com/docascode/docfx-seed)
* Run `mono /path/to/docfx.exe metadata`

### Expected result

Metadata should be created.

### Actual result

If fails to create any data...

```
[17-05-28 12:35:53.204]Info:Config file docfx.json found, start generating metadata...
Could not load signature of Microsoft.CodeAnalysis.SymbolSearch.SymbolSearchUpdateEngine+RemoteControlService:CreateClient due to: Could not load file or assembly 'Microsoft.V
isualStudio.RemoteControl, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. assembly:Microsoft.VisualStudio.RemoteControl, Versi
on=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a type:<unknown type> member:<none>
Could not load signature of Microsoft.CodeAnalysis.SymbolSearch.IRemoteControlService:CreateClient due to: Could not load file or assembly 'Microsoft.VisualStudio.RemoteContro
l, Version=14.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. assembly:Microsoft.VisualStudio.RemoteControl, Version=14.0.0.0, Culture=neu
tral, PublicKeyToken=b03f5f7f11d50a3a type:<unknown type> member:<none>
[17-05-28 12:35:53.682]Warning:Error opening solution /home/david/work/junk/docfx-seed/src/SampleClass1/SampleClass1.sln: Could not resolve the signature of a virtual method.
Ignored.
[17-05-28 12:35:53.698]Warning:No metadata is generated for .
[17-05-28 12:35:53.729]Info:Completed executing in 393.7172 milliseconds.
```

### Further technical details

[...]

Issue #1686, opened 2017-05-23
Hard-to-diagnose "No metadata is generated"


matkoch:
In relation to #165.

Even though I have `--logLevel Verbose` I cannot tell why the project is being skipped with `Warning:No metadata is generated for XXX.`.

vicancy:
1. Better logging
2. Troubleshooting guide

ThadHouse:
Is there anything else we can try rather than just install MS Build Tools 2015? I have multiple systems that have both 2015 and 2017 installed, and still can't generate metadata. I don't think that is the only dependency, or maybe something in VS 2017 is overriding something in the GAC causing docfx to fail.

matkoch:
@ThadHouse in my case I was targeting multiple target frameworks. In case you have the same, you need to specify them in your docfx.json as described here: https://github.com/dotnet/docfx/issues/1254#issuecomment-294080535

ThadHouse:
I'll double check that, but I think I've tried that before. I'm getting the error where no files are being returned by the MSBuildProject. But I'll try again. I might just end up putting a project.json back in my projects, because that does seem to actually wor.

matkoch:
What do you have instead? I think I've read something about xproj not being supported. But not sure :)

ThadHouse:
I'm using new VS 2017 csproj files, which have never worked for me in docfx, even though theyre supposed to.

matkoch:
Yes, for me that works. Public types/members? Maybe you can reference the project?

ThadHouse:
Yeah I've tried a ton of different things. Even example projects posted by both the docfx team, and other people that have posted issues don't work for me. Nothing related to the new csproj works for me.

ThadHouse:
I was able to get it working on one of my machines (I don't know how...) However, AppVeyor still does not work for me, so something still is not right with the dependencies that are thought to be required.

vicancy:
Hi ThadHouse, could you share your project & AppVeyor settings, to allow me to take a look?

ThadHouse:
https://github.com/robotdotnet/WPILib-Docs/tree/metadatatest

Here's my repo that I have to generate the docs. The runBuild.ps1 script downloads the latest version of docfx, clone my repos, and then attempt to run docfx. The appveyor settings are in appveyor.yml. 

https://github.com/robotdotnet/FRC-Utilities

[...]

PR #1734, opened 2017-06-06
Print warning message when project file does not contain any documents.

dlech:
Currently, roslyn does not seem to pick up source files from the "new" csproj format where .cs source files are implicitly included (I even updated to the latest pre-release roslyn nuget packages). So, to help diagnose this lack of support, print a warning message.


dlech:
This could have helped me in troubleshooting https://github.com/dotnet/docfx/issues/1708#issuecomment-306601538

Also, this handles one possible case for #1686

hellosnow:
:shipit:

PR #1738, opened 2017-06-07
Log WorkspaceFail events

dlech:
MSBuildWorkspace.OpenSolutionAsync() and MSBuildWorkspace.OpenProjectAsync() do not give any notification that loading a solution or project failed. However, there is a WorkspaceFailed event that we can use to be notified when there is a problem. This logs the diagnotic information from this event to help the user troubleshoot.

Related:
* https://github.com/dotnet/roslyn/issues/19978#issuecomment-306877880
* https://github.com/dotnet/docfx/issues/1686
* https://github.com/dotnet/docfx/issues/1708#issuecomment-306678871


dlech:
Just realized I am making the assumption that the Logger is thread-safe. Hopefully that is the case.

PR #1746, opened 2017-06-08
Extract base document processor for toc

hellosnow:


src/Microsoft.DocAsCode.Build.TableOfContents/TocDocumentProcessor.cs
```diff
@@ -35,42 +28,14 @@ public override ProcessingPriority GetProcessingPriority(FileAndType file)
...
-            // todo : metadata.
-            return new FileModel(file, toc)
-            {
-                Uids = new[] { new UidDefinition(file.File, displayLocalPath) }.ToImmutableArray(),
-                LocalPathFromRoot = displayLocalPath
-            };
-        }
-
-        public override SaveResult Save(FileModel model)
+        public override void RegisterTocToContext(TocItemViewModel toc, FileModel model, IDocumentBuildContext context)
```
hellosnow: >TocItemViewModel toc,  [](start = 50, length = 22)

better remove the view model from interface, to prevent view model change

vicancy:
:shipit:

src/Microsoft.DocAsCode.Build.TableOfContents/TocDocumentProcessorBase.cs
```diff
@@ -0,0 +1,171 @@
...
+            UpdateTocItemHref(toc, model, context);
+
+            RegisterTocToContext(model, context);
+        }
+
+        #endregion
+
+        #region Abstract methods
+
+        public abstract void RegisterTocToContext(FileModel model, IDocumentBuildContext context);
```
hellosnow: >public [](start = 8, length = 6)

protected

src/Microsoft.DocAsCode.Build.TableOfContents/TocDocumentProcessor.cs
hellosnow: reload TocItemViewModel will break original bahavior, add back, later only child item will call register method so it's ok.

---
In reply to: [121039829](https://github.com/dotnet/docfx/pull/1746#discussion_r121039829) [](ancestors = 121039829)

PR #1758, opened 2017-06-13
update RegisterToc methods to be protectes

hellosnow:


superyyrrzz:
:shipit:

src/Microsoft.DocAsCode.Build.TableOfContents/Constants.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace Microsoft.DocAsCode.Build.TableOfContents
 {
-    internal static class Constants
+    public static class Constants
```
vwxyzh: why `public`?

hellosnow: delete this class and move into common

---
In reply to: [121577705](https://github.com/dotnet/docfx/pull/1758#discussion_r121577705) [](ancestors = 121577705)

PR #1764, opened 2017-06-14
extract toc load logic

hellosnow:


src/Microsoft.DocAsCode.Build.TableOfContents/TocResolverUtility.cs
```diff
@@ -11,7 +12,7 @@ namespace Microsoft.DocAsCode.Build.TableOfContents
     using Microsoft.DocAsCode.DataContracts.Common;
     using Microsoft.DocAsCode.Plugins;
 
-    public class TocResolverUtility
+    public class ResolveTocUtility
```
hellosnow: >ResolveTocUtility [](start = 17, length = 17)

is there any suggest name?

PR #1790, opened 2017-06-21
fix bug in DocumentBuildContext.GetXrefSpec

vwxyzh:


src/Microsoft.DocAsCode.Build.Engine/DocumentBuildContext.cs
```diff
@@ -410,7 +410,7 @@ public XRefSpec GetXrefSpec(string uid)
 
             if (XRefMaps != null && XRefMaps.Count > 0)
```
qinezh: `XRefMaps` would always be null ?

src/Microsoft.DocAsCode.Build.Engine/DocumentBuildContext.cs
```diff
@@ -410,7 +410,7 @@ public XRefSpec GetXrefSpec(string uid)
...
             {
-                xref = (from map in XRefMaps select new BasicXRefMapReader(map).Find(uid)).FirstOrDefault();
+                xref = _reader.Result.Find(uid);
```
qinezh: Do we need to check if `_reader` is null?

hellosnow:
:shipit:

PR #1812, opened 2017-06-29
Make AmbientContext able to cross AppDomain, using object[] to save inside CallContext

vicancy:
for cross app domain cases


src/Microsoft.DocAsCode.Common/AmbientContext.cs
```diff
@@ -96,23 +98,53 @@ public void Dispose()
...
+        }
+
+        private static AmbientContext ToAmbientContext(object[] objs)
+        {
+            if (objs == null || objs.Length < 3)
+            {
+                return null;
+            }
+
+            return new AmbientContext()
```
vwxyzh: too many new instance, what about using structure?

PR #1815, opened 2017-06-29
Fix #1803 that reload model reload duplicate model

### LINKED ISSUE #1803: Overriding example with *content in same file as other overrides doesn't work
Overriding `example` with the `*content` anchor in the same file that overrides another field (also with the `*content` anchor) causes the `example` value to be ignored.

For example, using the following override file:

<pre><code>
---
uid: MyApp.MyService
remarks: *content
---

Here are some remarks about `MyService`.

---
uid: MyApp.MyService
example: [*content]
---

This is an example.

```csharp
var service = new MyService();

service.DoSomethingCool()
```
</code></pre>

The result is that the remarks show up, but the example does not.

I've verified that I can override multiple sections with the `*content` anchor in the same file, as long as one of them is not `example`. (For example, I've done both `remarks` and `summary`).

I tried putting `example` first in the file, but that did not change the result.

The only thing that worked was to create a separate file for just the `example`. This is less than ideal though because of how the *Improve this Doc* link works.

vicancy:
When overwrite document model contains multiple uids in one file, for example:
```markdown
---
uid: src.Class1.#ctor
remarks: *content
---
Hello there 1212.~

---
uid: src.Class1.#ctor
example: [*content]
---
This is example
```hello```
```
When in incremental case, it reloads the overwrite model, `m.Uids` stores `UidDefinition`, which distinct for different line numbers, and thus, the overwrite file is added to `_uidIndex` 2 times, thus, `merge` occurs 2 times. 2nd merge tries to merge `string` properties with `null` value and `array` properties with `null` value as well. `string` merge ignores null value while `array` merge *replace*s current one with `null` value, this is why #1803 takes place

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/HostService.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/HostService.cs
@@ -496,12 +496,12 @@ namespace Microsoft.DocAsCode.Build.Engine
                 m.FileOrBaseDirChanged += fileOrBaseDirChangedHandler;
                 m.UidsChanged += uidsChangedHandler;
                 m.ContentAccessed += contentAccessedHandler;
-                foreach (var uid in m.Uids)
+                foreach (var uid in m.Uids.Select(s => s.Name).Distinct())
                 {
-                    if (!_uidIndex.TryGetValue(uid.Name, out List<FileModel> list))
+                    if (!_uidIndex.TryGetValue(uid, out List<FileModel> list))
                     {
                         list = new List<FileModel>();
-                        _uidIndex.Add(uid.Name, list);
+                        _uidIndex.Add(uid, list);
                     }
                     list.Add(m);
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Build.Engine/HostService.cs
```diff
@@ -496,12 +496,12 @@ private void LoadCore(IEnumerable<FileModel> models)
                 m.FileOrBaseDirChanged += fileOrBaseDirChangedHandler;
                 m.UidsChanged += uidsChangedHandler;
                 m.ContentAccessed += contentAccessedHandler;
-                foreach (var uid in m.Uids)
+                foreach (var uid in m.Uids.Select(s => s.Name).Distinct())
```
hellosnow: >Select [](start = 43, length = 6)

how about add one case?

hellosnow:
:shipit:

PR #1826, opened 2017-07-03
Add plugins to split rest into tag pages

hellosnow:


hellosnow:
Will open another clean one PR.

PR #1828, opened 2017-07-04
Extract SplitModelBaseDocumentBuildStep

hellosnow:


hellosnow:
Discuss offline, SplittedInfo class could be removed and no need to extract then.

PR #1852, opened 2017-07-11
fix _rel unfound when href is url encoded

superyyrrzz:


hellosnow:
>             if (toc.Homepage != null)

question: will the homepage be encoded?

---
Refers to: src/Microsoft.DocAsCode.Build.TableOfContents/TocDocumentProcessor.cs:41 in 6eee194. [](commit_id = 6eee1943053e93484ac1d91f266d1df3ed014e46, deletion_comment = False)

superyyrrzz:
>             if (toc.Homepage != null)

yes, should

---
In reply to: [314342797](https://github.com/dotnet/docfx/pull/1852#issuecomment-314342797) [](ancestors = 314342797)

---
Refers to: src/Microsoft.DocAsCode.Build.TableOfContents/TocDocumentProcessor.cs:41 in 6eee194. [](commit_id = 6eee1943053e93484ac1d91f266d1df3ed014e46, deletion_comment = False)

hellosnow:
:shipit:

PR #1881, opened 2017-07-17
fix bug during download nuget

928PJY:
$lastexitcode won't change by PowerShell command, if some error happens, an exception will be thrown and script stop.

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -25,6 +25,8 @@ $scriptHome = Split-Path $scriptPath
 $versionCsFolderPath = $scriptHome + "\TEMP\"
 $versionCsFilePath = $versionCsFolderPath + "version.cs"
 
+$global:LASTEXITCODE = $null
+
 Push-Location $scriptHome
 
 function NugetPack {
--- a/build.ps1
+++ b/build.ps1
@@ -65,8 +65,7 @@ if (-not(ValidateCommand($nugetCommand))) {
     mkdir -Path "$env:LOCALAPPDATA\Nuget" -Force
     $ProgressPreference = 'SilentlyContinue'
     [Net.WebRequest]::DefaultWebProxy.Credentials = [Net.CredentialCache]::DefaultCredentials
-    Invoke-WebRequest 'https://dist.nuget.org/win-x86-commandline/latest/nuget.exe' -OutFile $nugetCommand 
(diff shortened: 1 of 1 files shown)


build.ps1
```diff
@@ -65,8 +65,7 @@ if (-not(ValidateCommand($nugetCommand))) {
     mkdir -Path "$env:LOCALAPPDATA\Nuget" -Force
     $ProgressPreference = 'SilentlyContinue'
     [Net.WebRequest]::DefaultWebProxy.Credentials = [Net.CredentialCache]::DefaultCredentials
-    Invoke-WebRequest 'https://dist.nuget.org/win-x86-commandline/latest/nuget.exe' -OutFile $nugetCommand 
-    ProcessLastExitCode $lastexitcode "Download nuget.exe"
```
vicancy: what's the side effect check $lastexitcode?

928PJY: Invoke-WebRequest won't change the value of $lastexitcode even if some error happens
So if $lastexitcode default to be `not 0`, even if Invoke-WebRequest runs successfully,  an error message will be print and this script will exit.

---
In reply to: [127640655](https://github.com/dotnet/docfx/pull/1881#discussion_r127640655) [](ancestors = 127640655)

build.ps1
```diff
@@ -65,8 +65,7 @@ if (-not(ValidateCommand($nugetCommand))) {
...
+    Invoke-WebRequest 'https://dist.nuget.org/win-x86-commandline/latest/nuget.exe' -OutFile $nugetCommand
```
hellosnow: >Invoke [](start = 4, length = 6)

For the case such as uri is not avaiable, need we catch exception and update last exit code?

build.ps1
vicancy: so $lastexitcode is `not 0` from last command line call, we should exit with that?

hellosnow:
:shipit:

build.ps1
hellosnow: >$lastexitcode [](start = 24, length = 13)

initialize $LastExitCode=0 in the very first beginning?

928PJY: this is the first time using `$lastexitcode`, so it depends on the default value.

---
In reply to: [127884787](https://github.com/dotnet/docfx/pull/1881#discussion_r127884787) [](ancestors = 127884787)

PR #1883, opened 2017-07-17
Add new templates to dashboard.

roel4ez:
Updated the `templates-dashboard.md` file with two new templates. Both templates are in the same repository, and can be used independently (depending on your preference).

Templates are compatible with `default` template.

Sorry about the branch name, `patch-1` was not my preferred name, but somehow it defaulted to it.


hellosnow:
@roel4ez thank you for contributing!

PR #1884, opened 2017-07-17
Add possibility to opt-out of anchors for specific links

pascalberger:
Adds the possibility to opt-out of anchors for specific links by setting the `no-anchor` class on the heading


superyyrrzz:
@pascalberger How can user add no-anchor to a heading?

pascalberger:
@superyyrrzz If you have HTML inside your contextual markdown files or as part of your liquid/moustache template

PR #1891, opened 2017-07-19
upgrade full-text search library to new version(2.1.2)

928PJY:
new version feature: separate search pipeline with index pipeline


src/docfx.website.themes/default/styles/search-worker.js
```diff
@@ -50,4 +44,35 @@
...
+    }
+  }
+
+  function isEmpty(obj) {
+    for (var prop in obj) {
+      if (obj.hasOwnProperty(prop))
+        return false;
+    }
+
+    return JSON.stringify(obj) === JSON.stringify({});
```
vicancy: is `return true` enough?

928PJY: yeah, for this situation, searchData has been initialed to {}.
Or obj maybe null but it will return true(wrong judge)?

---
In reply to: [128408409](https://github.com/dotnet/docfx/pull/1891#discussion_r128408409) [](ancestors = 128408409)

vicancy: stringify sounds heavy, how about
```js
if (!obj) return true;
if (typeof obj !== 'object') return true;
check props...
```

src/docfx.website.themes/default/styles/search-worker.js
```diff
@@ -31,11 +27,9 @@
...
-      if (searchData.hasOwnProperty(prop)) {
-        lunrIndex.add(searchData[prop]);
-      }
-    }
+
+    buildIndex();
+
     postMessage({ e: 'index-ready' });
   }
   searchDataRequest.send();
```
hellosnow: can extract a method? the logic is same as fetch stopWord?

src/docfx.website.themes/default/styles/search-worker.js
928PJY: offline discuss
should check if obj us null!

---
In reply to: [128416672](https://github.com/dotnet/docfx/pull/1891#discussion_r128416672) [](ancestors = 128416672)

hellosnow:
:shipit:

Issue #1886, opened 2017-07-17
Possible bug on getGitInfo (common.js)


leopignataro:
Dear DocFX team,

While debugging an issue with "contribute links" on my project, I stumbled upon something that I think might be a bug on the file [docfx.website.themes/common/common.js](https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/common/common.js), function `getGitInfo`:

```javascript
function getGitInfo(gitContribute, gitRemote) {
    // apiSpecFolder defines the folder contains overwrite files for MRef, the default value is apiSpec
    var defaultApiSpecFolder = "apiSpec";

    if (!gitContribute && !gitRemote) {
        return {
            apiSpecFolder: defaultApiSpecFolder
        }
    }

    if (!gitContribute) {
        return {
            apiSpecFolder: defaultApiSpecFolder,
            repo: gitRemote.repo,
            branch: gitRemote.branch,
            // path defines the relative path from current git repo.
            path: gitRemote.path
        }
    }

    return {
        apiSpecFolder: gitContribute.apiSpecFolder || defaultApiSpecFolder,
        repo: gitContribute.repo || gitRemote.repo,
        branch: gitContribute.branch || gitRemote.branch,
        path: gitRemote.path
    }
}
```

This function results in error if the parameter `gitContribute` is filled but `gitRemote` is not (which is exactly what is happening on my project).

Shouldn't this

```javascript
if (!gitContribute && !gitRemote) {
```

Be this?

```javascript
if (!gitRemote) {
```

Since the last two return statements depend on `gitRemote` being defined?

superyyrrzz:
@leopignataro Yes, it's a bug. I'll fix it.

superyyrrzz:
This is fixed in v2.22

PR #1892, opened 2017-07-19
fix gitRemote null reference

superyyrrzz:
#1886

### DIFF
--- a/src/docfx.website.themes/common/common.js
+++ b/src/docfx.website.themes/common/common.js
@@ -150,29 +150,26 @@ function getRemoteUrl(remote, startLine, gitContribute, gitUrlPattern) {
 
 function getGitInfo(gitContribute, gitRemote) {
     // apiSpecFolder defines the folder contains overwrite files for MRef, the default value is apiSpec
-    var defaultApiSpecFolder = "apiSpec";
+    var defaultApiSpecFolder = 'apiSpec';
 
-    if (!gitContribute && !gitRemote) {
-        return {
-            apiSpecFolder: defaultApiSpecFolder
-        }
+    var result = {};
+    if (gitContribute && gitContribute.apiSpecFolder) {
+        result.apiSpecFolder = gitContribute.apiSpecFolder;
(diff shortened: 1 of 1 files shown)


hellosnow:
:shipit:

src/docfx.website.themes/common/common.js
```diff
@@ -150,29 +150,26 @@ function getRemoteUrl(remote, startLine, gitContribute, gitUrlPattern) {
...
-    }
+    return result;
 
-    return {
-        apiSpecFolder: gitContribute.apiSpecFolder || defaultApiSpecFolder,
-        repo: gitContribute.repo || gitRemote.repo,
-        branch: gitContribute.branch || gitRemote.branch,
-        path: gitRemote.path
+    function mergeKey(source, sourceFallback, dest, key) {
+        if (source && source[key]) {
```
vicancy: as a general function, should we provide a more strict check to check only `undefined` ? For example, `false` or string.Empty all returns false currently

superyyrrzz: agree. fixed it.

src/docfx.website.themes/common/common.js
```diff
@@ -150,29 +150,26 @@ function getRemoteUrl(remote, startLine, gitContribute, gitUrlPattern) {
...
-    }
+    return result;
 
-    return {
-        apiSpecFolder: gitContribute.apiSpecFolder || defaultApiSpecFolder,
-        repo: gitContribute.repo || gitRemote.repo,
-        branch: gitContribute.branch || gitRemote.branch,
-        path: gitRemote.path
+    function mergeKey(source, sourceFallback, dest, key) {
+        if (source && key in source) {
```
vicancy: use `hasOwnProperty`? `in` returns true for `prototype` properties.

superyyrrzz: resolved

928PJY:
:shipit:

PR #1893, opened 2017-07-19
Allow user define heading id.

vwxyzh:
Allow user define heading id.
For example:
```md
# <a id="x"></a>Title
```
will generate following html:
```html
<h1 id="x">Title</h1>
```

### DIFF
--- a/src/Microsoft.DocAsCode.Dfm/Aggregators/HeadingIdAggregator.cs
+++ b/src/Microsoft.DocAsCode.Dfm/Aggregators/HeadingIdAggregator.cs
@@ -1,0 +1,59 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Dfm
+{
+    using System.Text.RegularExpressions;
+
+    using Microsoft.DocAsCode.MarkdownLite;
+
+    public class HeadingIdAggregator : MarkdownTokenAggregator<MarkdownHeadingBlockToken>
+    {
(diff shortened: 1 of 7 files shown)


src/Microsoft.DocAsCode.Dfm/Aggregators/HeadingIdAggregator.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Microsoft.DocAsCode.Dfm
+{
+    using System.Text.RegularExpressions;
+
+    using Microsoft.DocAsCode.MarkdownLite;
+
+    public class HeadingIdAggregator : MarkdownTokenAggregator<MarkdownHeadingBlockToken>
+    {
+        private static readonly Regex OpenARegex = new Regex(@"^\<a +(?:name|id)=\""([\w \-\.]+)\"" *\>$", RegexOptions.Compiled);
```
yufeih: <a id="x" class="xliff"></a>

yufeih: `<a id="x" class="xliff"></a>`

hellosnow:
:shipit:

src/Microsoft.DocAsCode.Dfm/Aggregators/HeadingIdAggregator.cs
vwxyzh: Do not apply id in such heading, keep out put like following:
```html
<h1 id="{autogenid}">
<a id="x" class="xliff"></a>Title
</h1>
```

PR #1921, opened 2017-08-03
encode invalid chars in version for xrefmap filename

superyyrrzz:


src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
```diff
@@ -308,6 +307,24 @@ private static string ExportXRefMap(DocumentBuildParameters parameters, Document
...
 
+        private static string GetXrefMapFileNameWithVersion(string version)
+        {
+            if (string.IsNullOrEmpty(version))
+            {
+                return XRefMapFileName;
+            }
+
+            // TODO: full check of invalid chars
+            version = Regex.Replace(
```
vwxyzh: what about `Uri.EscapeDataString`?

vicancy: what for? should JSON serializer handle the case?

superyyrrzz: In monikerrange, write to `> netcore1.0.xrefmap.yml` will fail... Exception is thown by `Path.Combine` before passed into serializer.

src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
```diff
@@ -308,6 +306,16 @@ private static string ExportXRefMap(DocumentBuildParameters parameters, Document
...
         }
 
+        private static string GetXrefMapFileNameWithVersion(string version)
+        {
+            if (string.IsNullOrEmpty(version))
+            {
+                return XRefMapFileName;
+            }
+
+            return Uri.EscapeDataString(version) + "." + XRefMapFileName;
```
vicancy: As it is a file name, should we use Path.GetInvalidFileNameChars?

PR #1926, opened 2017-08-04
extract different token render function from blockquote token render

928PJY:


src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
```diff
@@ -83,68 +83,19 @@ public override StringBuffer Render(IMarkdownRenderer renderer, MarkdownBlockquo
...
-                    content += "<blockquote";
-                    content = AppendSourceInfo(content, renderer, splitToken.Token);
-                    content += ">";
-                    foreach (var item in splitToken.InnerTokens)
-                    {
-                        content += renderer.Render(item);
-                    }
-                    content += "</blockquote>\n";
+                    content += DfmBlockQuoteTokenRender(renderer, splitToken);
                 }
```
vicancy: wrap these to a virtual method RenderSplitToken so that you can override it?

src/Microsoft.DocAsCode.Dfm/SplitToken.cs
```diff
@@ -14,6 +14,33 @@ public SplitToken(IMarkdownToken token)
...
+            Context = token.Context;
+            SourceInfo = token.SourceInfo;
         }
+
+        public IMarkdownRule Rule { get; }
+        public IMarkdownContext Context { get; }
+        public SourceInfo SourceInfo { get; }
+    }
+
+    public class DfmSectionBlockSplitToken : SplitToken
```
vicancy: >DfmSectionBlockSplitToken [](start = 17, length = 25)

per class per file

fenxu:
:shipit:

superyyrrzz:
:shipit:

vicancy:
:shipit:

PR #1938, opened 2017-08-09
better error message for invalid toc yaml file.

vwxyzh:


src/Microsoft.DocAsCode.Build.TableOfContents/TocHelper.cs
```diff
@@ -86,14 +88,25 @@ public static TocItemViewModel LoadYamlToc(string file)
...
-                    Metadata = tocWithMetadata.Metadata
-                };
+                    var tocWithMetadata = YamlUtility.Deserialize<TocRootViewModel>(file);
+                    return new TocItemViewModel
+                    {
+                        Items = tocWithMetadata.Items,
+                        Metadata = tocWithMetadata.Metadata
+                    };
+                }
+                catch (YamlException exForObject)
```
vicancy: Could this be a general deserialize helper method with fallbacks?

vwxyzh: good idea, to do later.

PR #1940, opened 2017-08-09
Add documentation for incremental post processing

hellosnow:


superyyrrzz:
I have a question: Can I get the change list in post processor? If a file is removed, the removed file doesn't exist in incremental manifest items, so I will load its info from ContextInfo I saved in last build, but actually I should know it's removed and no info should be fetched in this build.

hellosnow:
Changes list haven't exposed yet, I will implement it.

---
In reply to: [321712635](https://github.com/dotnet/docfx/pull/1940#issuecomment-321712635) [](ancestors = 321712635)

PR #1942, opened 2017-08-09
add yaml deserializer with fallback.

vwxyzh:


superyyrrzz:
:shipit:

vicancy:
Can I provide multiple fallbacks?

vwxyzh:
yes.

---
In reply to: [321495245](https://github.com/dotnet/docfx/pull/1942#issuecomment-321495245) [](ancestors = 321495245)

hellosnow:
:shipit:

Issue #1955, opened 2017-08-15
Error when serializing config file with dictionary valued items


killa1218:
**DocFX Version Used**: v2.22.1

**Template used**: `default`

**Steps to Reproduce**:

1. Deserialize a docfx.json file with complex data structure to an object
2. Serialize the object.

**Code snippet**:
```C#
public class DocFXConfig
{
    [JsonProperty(PropertyName = "build")]
    public BuildJsonConfig Build { get; set; }

    [JsonProperty(PropertyName = "metadata")]
    public MetadataJsonConfig Metadata { get; set; }
}
```
```C#
DocFXConfig docfxConfig = JsonConvert.DeserializeObject<DocFXConfig>(File.ReadAllText("docfx.json"));

using (FileStream fs = File.Open(targetConfigPath, FileMode.OpenOrCreate))
using (StreamWriter sw = new StreamWriter(fs))
using (JsonTextWriter jw = new JsonTextWriter(sw))
{
    JsonSerializer serializer = new JsonSerializer();

    serializer.NullValueHandling = NullValueHandling.Ignore;
    jw.Formatting = Formatting.Indented;
    jw.IndentChar = ' ';
    jw.Indentation = 2;
    serializer.Serialize(jw, docfxConfig);
}
```

**Expected Behavior**: Generate a new config file

**Actual Behavior**: Throw exception.

Fixed in PR #1956

vicancy:
Also add an UT in https://github.com/dotnet/docfx/blob/dev/test/docfx.Tests/JsonConverterTest.cs to reflect the issue?

killa1218:
Fixed.

PR #1956, opened 2017-08-15
Fix error when serializing config file with dictionary valued items

killa1218:
Use serializer instead of JsonWriter

Issue number: #1955


vicancy:
Also add an UT in https://github.com/dotnet/docfx/blob/dev/test/docfx.Tests/JsonConverterTest.cs to reflect the issue?

vicancy:
Thanks, cool 👍

PR #1968, opened 2017-08-16
fix relateive path for url

yishengjin1413:


src/Microsoft.DocAsCode.Common/Path/RelativePath.cs
```diff
@@ -40,6 +41,10 @@ private RelativePath(bool isFromWorkingFolder, int parentDirectoryCount, string[
         #endregion
 
         #region Public Members
+        public static RelativePath FromUrl(string path)
+        {
+            return (RelativePath)Uri.UnescapeDataString(path);
```
vicancy: should it be UrlDecode instead of UnescapeDataString?

vicancy: I don't think this method needs to be inside `RelativePath`, the one calling `RelativePath` knows how to pre-process `path`, `FromUrl` sounds obscure to me what the input it is expecting.

src/Microsoft.DocAsCode.Common/Path/RelativePath.cs
```diff
@@ -236,6 +241,12 @@ public RelativePath UrlDecode()
             {
                 parts[i] = Uri.UnescapeDataString(_parts[i]);
             }
+
+            if (_parts.Length > 0 && parts[0] == WorkingFolderString)
+            {
+                return new RelativePath(true, _parentDirectoryCount, parts.Skip(1).ToArray());
```
vicancy: why is it needed?

src/Microsoft.DocAsCode.Build.Engine/MarkupResultUtility.cs
```diff
@@ -104,8 +104,8 @@ private static MarkupResult ParseCore(MarkupResult markupResult, string file, Im
                 }
                 if (RelativePath.IsRelativePath(linkFile))
                 {
-                    var path = (RelativePath)currentFile + (RelativePath)linkFile;
-                    var file = path.GetPathFromWorkingFolder().UrlDecode();
+                    var path = (RelativePath)currentFile + RelativePath.FromUrl(linkFile);
```
vicancy: Add some ut?

src/Microsoft.DocAsCode.Common/Path/RelativePath.cs
yishengjin1413: This change is in UrlDecode method. Since after decode the path, 7E% may become ~, so we need to skip the first part(~) and set _isFromWorkingFolder to true.

src/Microsoft.DocAsCode.Build.Engine/MarkupResultUtility.cs
yishengjin1413: Added

src/Microsoft.DocAsCode.Common/Path/RelativePath.cs
yishengjin1413: As we synced offline, just leave this method here.

PR #1978, opened 2017-08-18
Corrected "docfxsvc" typo to "docfxvsc"

bitbonk:


hellosnow:
@bitbonk `docfxsvc` is the short name of `docfx service`, and it's an internal account to publish the extension...

PR #1984, opened 2017-08-21
publish extension under Microsoft

928PJY:
* use new icon
* publish under Microsoft


bitbonk:
I still cannot see the extensions under the new publisher: https://marketplace.visualstudio.com/search?term=publisher%3A%22ms-docfx%22&target=VSCode&category=All%20categories&sortBy=Relevance

Issue #1982, opened 2017-08-19
c# 7.1 feature `default` is not correctly handled


jskeet:
**DocFX Version Used**: 22.2.3

(And .NET Core SDK 2.0, on Windows.)

**Template used**: default

**Steps to Reproduce**:

Project file:

```xml
<Project Sdk="Microsoft.NET.Sdk">
  <PropertyGroup>
    <TargetFramework>netstandard1.4</TargetFramework>
    <LangVersion>latest</LangVersion>
  </PropertyGroup>
</Project>
```

docfx.json:

```json
{
  "metadata": [{
    "src": [{ "files": ["Project.csproj"] }],
    "dest": "api",
  }]
}
```

Sole class (Class1.cs):

```csharp
using System.Threading.Tasks;

namespace Project
{
    public class Class1
    {
        /// <param name="x">A parameter</param>
        public Task Foo(int x = default) => (Task) null;
    }
}
```

**Expected Behavior**:

Running `docfx metadata -f` should not cause any errors.

**Actual Behavior**:

The following error is generated:

> Error:Error extracting metadata for c:/Users/jon/Test/Projects/docfxbug/Project/Project.csproj: Value cannot be null.  
> Parameter name: name

Note that changing the code even slightly removes the error:

- Using `default(int)` instead of `default`
- Giving the method a regular body
- Removing the XML documentation for the parameter
- Changing the return value to just `null` instead of the cast

No idea what's going on or what's generating the error... I'm hoping this is sufficiently self-contained that it should be easy for the team to diagnose.

vicancy:
Fixed in 2.23.1

jskeet:
Firstly, thanks for jumping on this so quickly - it's much appreciated.

While this now produced metadata without an error, it's not producing the *right* metadata.

Generating the API from my sample code above using 2.23.1, I get a child with a UID of `Project.Class1.Foo(System.Int32,System.)`, and a declaration syntax of:

```csharp
public Task Foo(int x = null, )
```

The trailing comma is incorrect, as is the null. (It should be 0.) I'd expect the output from docfx to be exactly the same, regardless of whether I use `= default` or `= default(int)` in the parameter declaration.

In my real work projects, I'm seeing lots of members just not being generated at all, which is very odd. I haven't been able to reproduce that in a minimal test project yet, but please let me know if you'd like me to investigate that aspect further.

vicancy:
[...]

PR #1987, opened 2017-08-21
Fix https://github.com/dotnet/docfx/issues/1982

vicancy:


hellosnow:
>         public string GetTypeParameter(string name)

add some case?

---
Refers to: src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs:149 in 30afa9c. [](commit_id = 30afa9c4fad09a6752d77cef07412e501df9f18e, deletion_comment = False)

vicancy:
>         public string GetTypeParameter(string name)

the code uses c# 7.1 features and depends on env, not easy to add ut, will consider add a test repo for such tests

---
In reply to: [323694609](https://github.com/dotnet/docfx/pull/1987#issuecomment-323694609) [](ancestors = 323694609)

---
Refers to: src/Microsoft.DocAsCode.Metadata.ManagedReference/Parsers/TripleSlashCommentModel.cs:149 in 30afa9c. [](commit_id = 30afa9c4fad09a6752d77cef07412e501df9f18e, deletion_comment = False)

PR #1988, opened 2017-08-22
Remove unneccesory references

vicancy:
There is no need to include all the packageReferences in the new version of csproj as long as the referenced project contains the packageReference


superyyrrzz:
CI failed

PR #1993, opened 2017-08-22
support glyphicon icons in generated pdfs

cedar-ave:
I don't know that this is the right solution, but glyphicons didn't show up in PDFs until I tried this.


vicancy:
Hi @cedar-ave however the pdf.default template does not contain glyphicons? Are you using your custom template which uses glyphicon fonts?

bitbonk:
I do have the same problem. The glyphicon do not show up in the rendered PDF.

cedar-ave:
Hi @vicancy,
Yes, I made a custom template off pdf.default and added the Glyphicon fonts to it. Doing the same thing I also added support for FontAwesome in both the HTML and PDF outputs. But looking at my pull request, I don't think my solution is very good - maybe it is better if I document how I fixed this issue for myself and shared that instead, and then that document could be removed when someone who knows the better way to fix this issue in the code has a chance to.

PR #1997, opened 2017-08-24
insert isDependency metadata to fileModel

olprod:
insert isDependency metadata to fileModel

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
@@ -61,11 +61,14 @@ namespace Microsoft.DocAsCode.Build.Engine
         {
             using (new LoggerFileScope(file.File))
             {
+                // isDependency always is false for full build
+                bool isDependency = false;
                 if (CanProcessorIncremental(processor))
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Build.Engine/HostServiceCreatorWithIncremental.cs
```diff
@@ -60,29 +60,37 @@ public override bool CanProcessorIncremental(IDocumentProcessor processor)
...
+                            {
+                                Logger.LogDiagnostic($"Processor {processor.Name}, File {file.FullPath}: Skip build by incremental.");
+                                return null;
+                            }
+                            Logger.LogDiagnostic($"Processor {processor.Name}, File {file.FullPath}: Incremental not available.");
+                        }
                     }
                 }
+
+                var fmetadata = metadata == null ? new Dictionary<string, object>() : new Dictionary<string, object>(metadata);
```
vicancy: >fmetadata [](start = 20, length = 9)

this will be in final output, how about set into Properties for the FileModel in PostCreate

vwxyzh: or manifest property?

PR #1999, opened 2017-08-24
add group, obsolete version

superyyrrzz:


src/docfx/Models/VersionConfig.cs
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.DocAsCode
 
     using Newtonsoft.Json;
 
+    [Obsolete]
     [Serializable]
     public class VersionConfig
```
vicancy: rename to GroupConfig is enough?

src/docfx/Models/GroupConfig.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        [JsonProperty("dest")]
+        public string Destination { get; set; }
+
+        /// <summary>
+        /// The Root TOC Path used for navbar in current group, relative to output root.
+        /// If not set, will use the toc in output root in current group if exists.
+        /// </summary>
+        [JsonProperty("rootTocPath")]
+        public string RootTocPath { get; set; }
```
vicancy: does not sound like a general property for a group?

src/docfx/SubCommands/DocumentBuilderWrapper.cs
```diff
@@ -307,7 +307,21 @@ private static IEnumerable<DocumentBuildParameters> ConfigToParameter(BuildJsonC
             foreach (var pair in fileMappingParametersDictionary)
             {
                 var p = parameters.Clone();
-                if (config.Versions != null && config.Versions.TryGetValue(pair.Key, out VersionConfig vi))
+                if (config.Groups != null)
```
vicancy: move the fallback logic to the model instead when serializing or deserializing

PR #2000, opened 2017-08-24
Add metadata support in sdp

vicancy:


src/Microsoft.DocAsCode.Build.SchemaDriven/Models/JsonPointer.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            if (_isRoot)
+            {
+                return rootSchema;
+            }
+
+            BaseSchema schema = rootSchema;
+            foreach(var part in _parts)
+            {
+                schema = GetChildSchema(rootSchema, part);
```
superyyrrzz: >rootSchema [](start = 40, length = 10)

schema?

PR #2009, opened 2017-08-29
add group to IHostService

superyyrrzz:


src/Microsoft.DocAsCode.Plugins/GroupInfoModel.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Plugins
+{
+    using System.Collections.Generic;
+
+    public class GroupInfoModel
```
vicancy: naming: GroupInfo?

src/Microsoft.DocAsCode.Plugins/IHostService.cs
```diff
@@ -3,22 +3,27 @@
...
     {
         ImmutableList<TreeItemRestructure> TableOfContentRestructions { get; set; }
 
+        [Obsolete("use GroupInfo")]
         /// <summary>
         /// current version's name, String.Empty for default version
         /// </summary>
         string VersionName { get; }
 
+        [Obsolete("use GroupInfo")]
```
vicancy: should we place the attribute below the comment section...

superyyrrzz: yes...should

PR #2011, opened 2017-08-30
Changing video embed behavior

DuncanmaMSFT:
Change YouTube into youtube-nocookie.com and Channel 9 to append ?nocookie=true. Both are for EU Cookie Compliance

### DIFF
--- a/src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
+++ b/src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
@@ -131,6 +131,30 @@ namespace Microsoft.DocAsCode.Dfm
             StringBuffer content = string.Empty;
 
             var videoToken = splitToken.Token as DfmVideoBlockToken;
+            //TODO: fix video link
+            //if youtube.com, change to youtube-nocookie.com
+            //if channel9.msdn.com, add ?nocookie=true at end
+            if (videoToken != null && !String.IsNullOrWhiteSpace(videoToken.Link))
+            {
+                string link = videoToken.Link;
+
+                if (link.IndexOf("channel9.msdn.com", StringComparison.InvariantCultureIgnoreCase) > -1)
+                {
+                    if (link.Contains("?"))
+                    {
+                        link = link + "&nocookie=true";
+                    }
+                    else
+                    {
+                        link = link + "?nocookie=true";
+                    }
+                } else if (link.IndexOf("youtube.com", StringComparison.InvariantCultureIgnoreCase) > -1)
+                {
+                    link = link.Replace("youtube.com", "youtube-nocookie.com");
+                }
+                videoToken.Link = link;
+            }
+
             content += "<div class=\"embeddedvideo\"><iframe src=\"";
             content += videoToken.Link;
             content += "\" frameborder=\"0\" allowfullscreen=\"true\"";
--- a/src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
+++ b/src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
@@ -137,22 +137,28 @@ namespace Microsoft.DocAsCode.Dfm
             if (videoToken != null && !String.IsNullOrWhiteSpace(videoToken.Link))
             {
                 string link = videoToken.Link;
+                Uri videoLink;
+                if (Uri.TryCreate(link, UriKind.Absolute, out videoLink))
+                {
+                    string host = videoLink.Host;
 
-                if (link.IndexOf("channel9.msdn.com", StringComparison.InvariantCultureIgnoreCase) > -1)
-                {
-                    if (link.Contains("?"))
+                    if (host.Equals("channel9.msdn.com", StringComparison.InvariantCultureIgnoreCase))
                     {
-                        link = link + "&nocookie=true";
+                        if (link.Contains("?"))
+                        {
+                            link = link + "&nocookie=true";
+                        }
+                        else
+                        {
+                            link = link + "?nocookie=true";
+                        }
                     }
-                    else
+                    else if (host.IndexOf("youtube.com", StringComparison.InvariantCultureIgnoreCase) > -1)
                     {
-                        link = link + "?nocookie=true";
+                        link = link.Replace("youtube.com", "youtube-nocookie.com");
                     }
-                } else if (link.IndexOf("youtube.com", StringComparison.InvariantCultureIgnoreCase) > -1)
-                {
-                    link = link.Replace("youtube.com", "youtube-nocookie.com");
+                    videoToken.Link = link;
                 }
-                videoToken.Link = link;
             }
 
             content += "<div class=\"embeddedvideo\"><iframe src=\"";
--- a/src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
+++ b/src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
@@ -134,7 +134,7 @@ namespace Microsoft.DocAsCode.Dfm
             //TODO: fix video link
             //if youtube.com, change to youtube-nocookie.com
             //if channel9.msdn.com, add ?nocookie=true at end
-            if (videoToken != null && !String.IsNullOrWhiteSpace(videoToken.Link))
+            if (videoToken != null && !string.IsNullOrWhiteSpace(videoToken.Link))
             {
                 string link = videoToken.Link;
                 Uri videoLink;
--- a/test/Microsoft.DocAsCode.Dfm.Tests/DocfxFlavoredMarkdownTest.cs
+++ b/test/Microsoft.DocAsCode.Dfm.Tests/DocfxFlavoredMarkdownTest.cs
@@ -379,6 +379,7 @@ Inline [!include[ref3](ref3.md ""This is root"")]
             var marked = DocfxFlavoredMarked.Markup(root);
             Assert.Equal(expected.Replace("\r\n", "\n"), marked);
         }
+
         [Fact]
         [Trait("Related", "DfmMarkdown")]
         public void TestDfmVideo_YouTubeVideosUseNoCookie()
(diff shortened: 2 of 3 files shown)


src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
```diff
@@ -131,6 +131,30 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmVideoBlockSpli
...
 
             var videoToken = splitToken.Token as DfmVideoBlockToken;
+            //TODO: fix video link
+            //if youtube.com, change to youtube-nocookie.com
+            //if channel9.msdn.com, add ?nocookie=true at end
+            if (videoToken != null && !String.IsNullOrWhiteSpace(videoToken.Link))
+            {
+                string link = videoToken.Link;
+
+                if (link.IndexOf("channel9.msdn.com", StringComparison.InvariantCultureIgnoreCase) > -1)
```
vicancy: Use Uri to parse first? it can hit such case "channel9.msdn.com.anotherwebsite.com"

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
```diff
@@ -131,6 +131,30 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmVideoBlockSpli
             StringBuffer content = string.Empty;
 
             var videoToken = splitToken.Token as DfmVideoBlockToken;
+            //TODO: fix video link
+            //if youtube.com, change to youtube-nocookie.com
+            //if channel9.msdn.com, add ?nocookie=true at end
+            if (videoToken != null && !String.IsNullOrWhiteSpace(videoToken.Link))
```
928PJY: >String [](start = 39, length = 6)

use string?

test/Microsoft.DocAsCode.Dfm.Tests/DocfxFlavoredMarkdownTest.cs
```diff
@@ -369,11 +369,26 @@ public void TestDfmVideo_Video()
...
 
             var expected = @"<p>The following is video.</p>
-<div class=""embeddedvideo""><iframe src=""https://sec.ch9.ms/ch9/4393/7d7c7df7-3f15-4a65-a2f7-3e4d0bea4393/Episode208_mid.mp4"" frameborder=""0"" allowfullscreen=""true""></iframe></div>
+<div class=""embeddedvideo""><iframe src=""https://channel9.msdn.com/Shows/This+Week+On+Channel+9/TWC9-Mad-about-Mads-and-C-80-HoloLens-Wins-Award-3D-Printed-Warthog-and-more/player?nocookie=true"" frameborder=""0"" allowfullscreen=""true""></iframe></div>
+";
+
+            var marked = DocfxFlavoredMarked.Markup(root);
+            Assert.Equal(expected.Replace("\r\n", "\n"), marked);
+        }
+        [Fact]
```
928PJY: new line?

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
DuncanmaMSFT: not sure what you are suggesting? string vs. String ? Given that I'm using a static method of the framework class this seems appropriate vs. string, but I can change if it matches your desired coding guidelines

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
DuncanmaMSFT: Updated to use URI

test/Microsoft.DocAsCode.Dfm.Tests/DocfxFlavoredMarkdownTest.cs
DuncanmaMSFT: This is the same code being used in the other video related tests. It seems to just be a difference between how the expected string is constructed and the real markdown output gets created.

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
```diff
@@ -131,6 +131,36 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmVideoBlockSpli
...
+                            link = link + "&nocookie=true";
+                        }
+                        else
+                        {
+                            link = link + "?nocookie=true";
+                        }
+                    }
+                    else if (host.IndexOf("youtube.com", StringComparison.InvariantCultureIgnoreCase) > -1)
+                    {
+                        link = link.Replace("youtube.com", "youtube-nocookie.com");
```
superyyrrzz: >link.Replace [](start = 31, length = 12)

use Regex.Replace for ignore case?

test/Microsoft.DocAsCode.Dfm.Tests/DocfxFlavoredMarkdownTest.cs
superyyrrzz: I think Jiayin suggests to add a empty line between 2 methods. Minor style issue:)

---
In reply to: [136177047](https://github.com/dotnet/docfx/pull/2011#discussion_r136177047) [](ancestors = 136177047)

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
superyyrrzz: Yes, DocFX also preferes string in static methods.

---
In reply to: [136176198](https://github.com/dotnet/docfx/pull/2011#discussion_r136176198) [](ancestors = 136176198)

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
DuncanmaMSFT: Instantiating a Regex object seems heavier than IndexOf performance wise, should I run a perf test between the two paths?

DuncanmaMSFT: Updated the code based on the other pieces of feedback (line break between methods, string instead of String)

superyyrrzz: Yes, performance should be an issue. My concern is that the ignore case `IndexOf` check makes no sense if the `Replace` is case sensitive...

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
```diff
@@ -131,6 +131,36 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmVideoBlockSpli
...
+            //if channel9.msdn.com, add ?nocookie=true at end
+            if (videoToken != null && !string.IsNullOrWhiteSpace(videoToken.Link))
+            {
+                string link = videoToken.Link;
+                Uri videoLink;
+                if (Uri.TryCreate(link, UriKind.Absolute, out videoLink))
+                {
+                    string host = videoLink.Host;
+
+                    if (host.Equals("channel9.msdn.com", StringComparison.InvariantCultureIgnoreCase))
```
vicancy: use `OrdinalIgnoreCase`?

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
```diff
@@ -131,6 +131,36 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmVideoBlockSpli
...
+                    {
+                        if (link.Contains("?"))
+                        {
+                            link = link + "&nocookie=true";
```
vicancy: what if the link contains hashtag `#`?

src/Microsoft.DocAsCode.Dfm/DfmRenderer.cs
```diff
@@ -131,6 +131,36 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmVideoBlockSpli
...
+                        }
+                        else
+                        {
+                            link = link + "?nocookie=true";
+                        }
+                    }
+                    else if (host.IndexOf("youtube.com", StringComparison.InvariantCultureIgnoreCase) > -1)
```
vicancy: also use `equals` here? host can be `youtube.com.another.com`

vwxyzh:
@DuncanmaMSFT i think this logic should [customize DFM renderer](http://dotnet.github.io/docfx/tutorial/howto_customize_docfx_flavored_markdown.html#customize-renderer).

vicancy:
I think it is OK to keep the code logic here as "youtube", "channel9" are large video platforms @vwxyzh Agree that customized renderer can be a clearer approach

DuncanmaMSFT:
@vwxyzh and @vicancy ... to your point about a customized renderer, does OPS already contain a set of code where I could move my 'forked' version of the [!video] markdown extension and have it replace this one when OPS runs? From my point of view, this needs to be live *for docs* as soon as possible, and if DocFX is not the right place to do this work, just let me know exactly where it should be

vwxyzh:
OPS markdown extension code is [here](https://mseng.visualstudio.com/VSChina/Open%20Publishing%20Build/_git/OpenPublishing.Build?path=%2FLibraries%2FDocAsCode.Plugins%2FMarkdownExtensions).

vwxyzh:
@herohua should we help @DuncanmaMSFT add the customized renderer in OPS?

DuncanmaMSFT:
I can add it to OpDfmRenderer.cs easily enough, but is the process to build and test the customized renderer documented?

vwxyzh:
@DuncanmaMSFT *OpDfmRenderer* is a special renderer for special repo, I think it is better to refer to *MonikerZone* or *InteractiveCode*.
[Here](https://mseng.visualstudio.com/VSChina/Open%20Publishing%20Build/_git/OpenPublishing.Build?_a=contents&path=%2FUnitTests%2FDocAsCode.Plugins.UnitTest%2FMonikerZone.UnitTest.cs&version=GBdevelop) is unit test for *MonikerZone*.
[Here](https://mseng.visualstudio.com/VSChina/Open%20Publishing%20Build/_git/OpenPublishing.Build?_a=contents&path=%2FUnitTests%2FDocAsCode.Plugins.UnitTest%2FInteractiveCodeRenderer.UnitTest.cs&version=GBdevelop) is unit test for *InteractiveCode*.
It will be applied to document build in the next day after check-in the source code in sandbox environment (i.e. PPE).

PR #2028, opened 2017-09-04
fix syntax for c# 7.1 feature default.

vwxyzh:
#1982

### DIFF
--- a/test/Microsoft.DocAsCode.Metadata.ManagedReference.Tests/GenerateMetadataFromCSUnitTest.cs
+++ b/test/Microsoft.DocAsCode.Metadata.ManagedReference.Tests/GenerateMetadataFromCSUnitTest.cs
@@ -2660,6 +2660,31 @@ namespace Test1
             Assert.Equal("Test1.I1`1.M1(`0)", r1.Definition);
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


test/Microsoft.DocAsCode.Metadata.ManagedReference.Tests/GenerateMetadataFromCSUnitTest.cs
```diff
@@ -2660,6 +2660,31 @@ public interface I2<T> : I1<string>, I1<T> {}
...
+            Assert.Equal(1, output.Items.Count);
+            var ns = output.Items[0];
+            Assert.NotNull(ns);
+            var foo = ns.Items[0];
+            Assert.NotNull(foo);
+            Assert.Equal("Test1.Foo", foo.Name);
+            Assert.Equal(1, foo.Items.Count);
+            var bar = foo.Items[0];
+            Assert.Equal("Test1.Foo.Bar(System.Int32)", bar.Name);
+            Assert.Equal("public int Bar(int x = default (int))", bar.Syntax.Content[SyntaxLanguage.CSharp]);
```
superyyrrzz: >default (int) [](start = 49, length = 13)

Can this be `default` to be same as in input code?

PR #2031, opened 2017-09-05
fix xref query deserialize error.

vwxyzh:


src/Microsoft.DocAsCode.Build.Engine/DocumentBuildContext.cs
```diff
@@ -252,8 +252,19 @@ private List<string> ResolveByExternalReferencePackages(List<string> uidList, Co
...
                 {
-                    IList<XRefSpec> xsList = JsonUtility.Deserialize<List<XRefSpec>>(sr);
-                    return xsList;
+                    var xsList = JsonUtility.Deserialize<List<Dictionary<string, object>>>(sr);
+                    return xsList.ConvertAll(item =>
+                    {
+                        var spec = new XRefSpec();
+                        foreach (var pair in item)
+                        {
+                            if (pair.Value is string s)
```
vicancy: LogVerbose for ignored ones?

vwxyzh: I think too many logs will be produced.

PR #2037, opened 2017-09-06
fix yaml mime warning.

vwxyzh:


src/Microsoft.DocAsCode.Build.ManagedReference/ManagedReferenceDocumentProcessor.cs
```diff
@@ -79,6 +79,11 @@ public ManagedReferenceDocumentProcessor()
 
         protected override FileModel LoadArticle(FileAndType file, ImmutableDictionary<string, object> metadata)
         {
+            if (YamlMime.ReadMime(file.File) == null)
+            {
+                Logger.LogWarning("Please add yamlmime.", file: file.File);
```
vicancy: "Please add `### YamlMime:ManagedReference` if this is a managed reference metadata file."

PR #2052, opened 2017-09-12
support getting deleted file content from git

OsmondJiang:


vicancy:
>                 // ignored

LogWarning instead of ignore?

---
Refers to: src/Microsoft.DocAsCode.Common/Git/GitUtility.cs:291 in 2e4afcd. [](commit_id = 2e4afcd651fb54be22811e44e9d6856a826bd788, deletion_comment = False)

PR #2054, opened 2017-09-13
Move validation config to plugin project

qinezh:


vwxyzh:
What about namespace?

PR #2056, opened 2017-09-13
Support fallback token/code snippet to GIT HISTORY

OsmondJiang:


OsmondJiang:
close this PR as we are going another way

PR #2065, opened 2017-09-18
use wbr instead of zero width space

superyyrrzz:
Copy a string containing ZERO-WIDTH SPACE from Firefox may be devastating

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -38,9 +38,7 @@ $(function () {
     $(".xref").addClass("text-break");
     var texts = $(".text-break");
     texts.each(function () {
-      $(this).html(function (index, oldHtml) {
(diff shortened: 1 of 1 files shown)


src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -38,8 +38,8 @@ $(function () {
     $(".xref").addClass("text-break");
     var texts = $(".text-break");
     texts.each(function () {
-      $(this).text(function (index, text) {
-        return util.breakText(text);
+      $(this).html(function (index, oldHtml) {
+        return util.breakText(oldHtml);
```
vwxyzh: does it work for html tags?

superyyrrzz: No. The input should not contain html tags, otherwise it may be inserted into a tag and break it.

PR #2077, opened 2017-09-25
add uri template.

vwxyzh:


src/Microsoft.DocAsCode.Common/UriTemplates/UriTemplate.cs
```diff
@@ -0,0 +1,96 @@
...
+        private readonly string _template;
+        private readonly Func<string, T> _func;
+        private readonly IUriTemplatePipeline<T>[] _pipeline;
+        private readonly string[][] _parameters;
+
+        /// <summary>
+        /// step 1 (macro):
+        /// * env var:      {%abc%}                         -> expand env var abc.
+        /// Step 2 (pipeline):
+        /// * pipeline:     |> function parameters          -> "", and append a pipeline.
```
superyyrrzz: >function parameters [](start = 31, length = 19)

seems function name is not used. What about move parameters into UriTemplatePipeline's ctor? Then step2 isn't needed in template string.

src/Microsoft.DocAsCode.Common/UriTemplates/UriTemplate.cs
```diff
@@ -0,0 +1,96 @@
...
+            var t = _marcoRegex.Replace(
+                template,
+                m => Environment.GetEnvironmentVariable(m.Groups[1].Value));
+            List<IUriTemplatePipeline<T>> pipeline = new List<IUriTemplatePipeline<T>>();
+            List<string[]> parameters = new List<string[]>();
+            t = _pipelineRegex.Replace(
+                t,
+                m =>
+                {
+                    pipeline.Add(pipelineProvider(m.Groups[1].Value));
```
superyyrrzz: >pipeline.Add [](start = 20, length = 12)

seems the list will contain one item at most. No need to use a list?

src/Microsoft.DocAsCode.Common/UriTemplates/UriTemplate.cs
vwxyzh: Function name is the contract name of IUriTemplatePipeline\<T\>.
If move to parameters into UriTemplatePipeline's ctor, i need one more interface: `IUriTemplatePipelineFactory<T>`

src/Microsoft.DocAsCode.Common/UriTemplates/UriTemplate.cs
vwxyzh: no, you can write multi-pipeline in one uri template, e.g.: `a|>x|>y`

vwxyzh: @superyyrrzz , see new test case.

src/Microsoft.DocAsCode.Common/UriTemplates/UriTemplate.cs
vwxyzh: @superyyrrzz , see the create method added in unit test.

src/Microsoft.DocAsCode.Common/UriTemplates/UriTemplate.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Text.RegularExpressions;
+
+    using Microsoft.DocAsCode.Plugins;
+
+    public class UriTemplate<T>
+    {
+        private static readonly Regex _marcoRegex = new Regex(@"{%\s*([\S]+?)\s*%}");
```
vicancy: Compiled?

src/Microsoft.DocAsCode.Common/UriTemplates/UriTemplate.cs
```diff
@@ -0,0 +1,96 @@
...
+            _template = template;
+            _func = func;
+            _pipeline = pipeline;
+            _parameters = parameters;
+        }
+
+        /// <summary>
+        /// Evaluate uri template with variables
+        /// </summary>
+        public T Evaluate(IDictionary<string, string> variables)
```
vicancy: >variables [](start = 54, length = 9)

null check?

vicancy: >string [](start = 46, length = 6)

can be object?

vwxyzh: no, string only.

PR #2078, opened 2017-09-25
pass token but not properties of token

928PJY:


src/Microsoft.DocAsCode.Dfm/DfmCodeRenderer.cs
```diff
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
```
vicancy: move into namespace scope

Issue #2085, opened 2017-09-27
Git Repo information not generated in Raw Model 


senglin:
**DocFX Version Used**: 2.24

**Template used**: `default`

**Steps to Reproduce**:

Git repository information is used by DocFX to generate git repository links in the generated docs.

Surprisingly, this works on the Mac OS X mono framework but *not on linux (mono docker container (Debian) or an Ubuntu machine)*

This scenario also works on Windows machine.

1.  Please see below

**Expected Behavior**: (when run on Mac OS X installation)
```
$ git clone https://github.com/dotnet/docfx.git
$ cd docfx/Documentation/
$ mono /usr/bin/docfx/docfx.exe build --exportRawModel
...
$ jq '.documentation' _site/index.raw.json
{
  "remote": {
    "path": "Documentation/index.md",
    "branch": "dev",
    "repo": "https://github.com/dotnet/docfx.git"
  },
  "startLine": 0,
  "endLine": 0,
  "isExternal": false
}
```

**Actual Behavior**: (when run on Ubuntu machine / Docker mono)
```
$ git clone https://github.com/dotnet/docfx.git
$ cd docfx/Documentation/
$ mono /usr/bin/docfx/docfx.exe build --exportRawModel
...
$ jq '.documentation' _site/index.raw.json
{
  "startLine": 0,
  "endLine": 0,
  "isExternal": false
}
```

# Tested Environments

## Mono Docker Container

```
# uname -a; echo; cat /etc/os-release; echo; mono --version
Linux 6cf83763b2f3 4.4.0-92-generic #115-Ubuntu SMP Thu Aug 10 09:04:33 UTC 2017 x86_64 GNU/Linux

PRETTY_NAME="Debian GNU/Linux 8 (jessie)"
NAME="Debian GNU/Linux"
VERSION_ID="8"
VERSION="8 (jessie)"
ID=debian
HOME_URL="http://www.debian.org/"
SUPPORT_URL="http://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

Mono JIT compiler version 5.2.0.215 (tarball Mon Aug 14 15:50:54 UTC 2017)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
        TLS:           __thread
        SIGSEGV:       altstack
        Notifications: epoll
        Architecture:  amd64
        Disabled:      none
        Misc:          softdebug 
        LLVM:          supported, not enabled.
[...]

Issue #2023, opened 2017-09-02
No warning when `git` not found


saulsdu:
When building project and `git` is not found on path, docfx does not produce a warning in the logs.

We noticed this because the contribution links were not being generated and in trying to troubleshoot realized neither source nor documentation repo info was in the raw models (using `--exportRawModel`). 

Inspecting `Microsoft.DocAsCode.Common.Git.GitUtility` reveals that try-catch methods appear to ignore/swallow exceptions rather than log output. See method `TryGetFileDetail` for example.

Agree that docfx should not exit if git is not found, but it should provide info necessary for troubleshooting in this case.

**DocFX Version Used**: 2.23.1.0

**Template used**: `default`

**Steps to Reproduce**:

1. On system without Git installed or `git` command not found...
2. In root of docfx project (has .git subdirectory)
3. Run `docfx build` 

**Expected Behavior**:

Docfx logs warning message if `git` not found.

**Actual Behavior**:

When `git` is not found, no warning or error is logged as output.

PR #2086, opened 2017-09-27
Log Git Command Exceptions and Utilize Bash Executable

senglin:
This pull request addresses both issues #2023 and #2085 

After resolving issue #2023  by enabling logging when exceptions are found when running the Git command, it was found that `type` command was not found.

```
$ git clone https://github.com/dotnet/docfx.git
$ cd docfx/Documentation/
$ mono /usr/bin/docfx.verboselogs/docfx.exe build --exportRawModel | grep index.md
[17-09-26 06:27:39.978]Warning:[BuildCore.Build Document.Load.ConceptualDocumentProcessor](index.md)ApplicationName='type', CommandLine='git', CurrentDirectory='', Native error= Cannot find the specified file
[17-09-26 06:27:46.039]Warning:[BuildCore.Build Document.Load.ConceptualDocumentProcessor](index.md)ApplicationName='type', CommandLine='git', CurrentDirectory='', Native error= Cannot find the specified file
$ jq '.documentation' _site/index.raw.json
{
  "startLine": 0,
  "endLine": 0,
  "isExternal": false
}
```

`type` is actually a built-in bash command, rather than an application found on the environment path.

On why this works differently from a OS X enviroment; my hunch is that the implementation of mono on OS X is different.

We could have just set `UseShellExecute` to `false`, however this would remove certain features such as redirected errors and outputs. [[reference](https://stackoverflow.com/questions/5255086/when-do-we-need-to-set-useshellexecute-to-true)]

Anyway, the most appropriate way to run a bash command is via bash itself. [[Reference1](https://raspberrypi.stackexchange.com/questions/9474/how-can-i-run-a-terminal-command-bash-from-c-mono), [Reference2](https://loune.net/2017/06/running-shell-bash-commands-in-net-core/)]

We should not assume that the spawned process uses a Bash environment.

Switching from calling `type` directly to making it `type` an argument to `bash`; e.g. `bash -c “type git”`

```
$ git clone https://github.com/dotnet/docfx.git
$ cd docfx/Documentation/
$ mono /usr/bin/docfx/docfx.exe build --exportRawModel --exportRawModel | grep index.md
$ jq '.documentation' _site/index.raw.json
{
  "remote": {
    "path": "Documentation/index.md",
    "branch": "dev",
    "repo": "https://github.com/dotnet/docfx.git"
  },
  "startLine": 0,
  "endLine": 0,
  "isExternal": false
}
```

This resolves issue #2085 .


src/Microsoft.DocAsCode.Common/CommandUtility.cs
```diff
@@ -100,6 +100,7 @@ public static int RunCommand(CommandInfo commandInfo, StreamWriter stdoutWriter
                     }
                     else
                     {
+                        Logger.LogWarning($"Timeout ({timeoutInMilliseconds}ms) exceeded. Killing process.");
```
vicancy: >. [](start = 105, length = 1)

Killing process {process.Name}

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -46,9 +46,9 @@ public static GitDetail TryGetFileDetail(string filePath)
             {
                 detail = GetFileDetail(filePath);
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                // ignored
+                Logger.LogWarning(ex.ToString());
```
vicancy: >LogWarning [](start = 23, length = 10)

too much warning if git does not exists?

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -272,9 +272,9 @@ private static string TryRunGitCommand(string repoPath, string arguments)
             {
                 RunGitCommand(repoPath, arguments, output => content.AppendLine(output));
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                // ignored
+                Logger.LogWarning(ex.ToString());
```
vicancy: >LogWarning [](start = 23, length = 10)

too much warning noises

src/Microsoft.DocAsCode.Common/CommandUtility.cs
```diff
@@ -116,21 +117,35 @@ public static int RunCommand(CommandInfo commandInfo, StreamWriter stdoutWriter
...
+                        Name = "where",
+                        Arguments = commandName
+                    }, outputWriter, errorWriter, timeoutInMilliseconds: 1000);
+                }
+
+                String resultError = System.Text.Encoding.UTF8.GetString(errorStream.ToArray(), 0, (int)errorStream.Length);
+                String resultOutput = System.Text.Encoding.UTF8.GetString(outputStream.ToArray(), 0, (int)outputStream.Length);
+                if (!string.IsNullOrEmpty(resultError) || !string.IsNullOrEmpty(resultOutput))
+                {
+                    Logger.LogWarning($"RunCommand exitCode: {exitCode}, output: \"{resultOutput}\", error: \"{exitCode}\"");
```
vicancy: >LogWarning [](start = 27, length = 10)

warning looks too heavy for checking command existence

src/Microsoft.DocAsCode.Common/CommandUtility.cs
senglin: Thanks, I will make that change.

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
senglin: OK, I was in a dilemma about this. During debugging, ex.Message wasn't enough to narrow down where the fault was.  Adding a stack trace proved critical in finding faults.

What do you think?

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
senglin: Same comment as above.

src/Microsoft.DocAsCode.Common/CommandUtility.cs
senglin: Same comment as above.

senglin: I will have to fix the last `exitCode` reference to  `resultError`.

This warning message would only appear if there's an error / warning when the command is run. 

I think it is important to know what the fault is, if the program is able to tell us so.

Comments?

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
senglin: If you look at the example above, the exception results in a warning:
`[17-09-26 06:27:39.978]Warning:[BuildCore.Build Document.Load.ConceptualDocumentProcessor](index.md)ApplicationName='type', CommandLine='git', CurrentDirectory='', Native error= Cannot find the specified file`

Notice that this tells you that this command fails, but doesn't let you know exactly at which point it fails.

After thinking about this, do you think it would be better to have
```
Logger.LogWarning(ex.Message);
Logger.LogDiagnostic(ex.ToString());
```

src/Microsoft.DocAsCode.Common/CommandUtility.cs
senglin: Similar to above, would it be better to rewrite it as
```
Logger.LogDiagnostic($"RunCommand exitCode: {exitCode}, output: \"{resultOutput}\", error: \"{resultError}\"");
```

vicancy: For example, When `git` does not exist, `where git` prints to `resultError`. however, I don't think docfx should print warning, as `git` not exist is also an acceptable scenario. `LogDiagnostic` sounds better.

vicancy: I make a change in https://github.com/dotnet/docfx/pull/2090 so that it does not throw exception every time when git does not exist, after the change, my concern about `warning too much` is gone and it looks cool to me to use `LogWarning`

senglin: Looking forward to #2090 being merged. I will retain LogWarning for this change.

senglin:
@vicancy I used `LogVerbose` for parts that had you concern about long warning messages. `LogDiagnostic` doesn't seem to work when I tried `-loglevel Diagnostic`. It seems that it's not supported.

Thanks again for your review.

src/Microsoft.DocAsCode.Common/CommandUtility.cs
```diff
@@ -116,21 +117,35 @@ public static int RunCommand(CommandInfo commandInfo, StreamWriter stdoutWriter
...
+                        Name = "where",
+                        Arguments = commandName
+                    }, outputWriter, errorWriter, timeoutInMilliseconds: 1000);
+                }
+
+                String resultError = System.Text.Encoding.UTF8.GetString(errorStream.ToArray(), 0, (int)errorStream.Length);
+                String resultOutput = System.Text.Encoding.UTF8.GetString(outputStream.ToArray(), 0, (int)outputStream.Length);
+                if (!string.IsNullOrEmpty(resultError) || !string.IsNullOrEmpty(resultOutput))
+                {
+                    Logger.LogWarning($"RunCommand exitCode: {exitCode}, output: \"{resultOutput}\", error: \"{resultError}\"");
```
vicancy: Every time this method gets called, it always log a warning as either error or output stream triggers a warning? How about letting the caller decide whether or not logging anything? For this method, it returns the error and output string..

vicancy: https://github.com/dotnet/docfx/pull/2092 to log in GitUtility when command not found

senglin:
@vicancy I have removed the logging logic in `ExistCommand` and allow callers to provide delegates.
I have also made changes to previous changes. It seems that StreamWriters have to not be closed but flushed before the MemoryStream can be read.
Reference: https://stackoverflow.com/questions/1232443/writing-to-then-reading-from-a-memorystream

src/Microsoft.DocAsCode.Common/CommandUtility.cs
```diff
@@ -113,24 +114,49 @@ public static int RunCommand(CommandInfo commandInfo, StreamWriter stdoutWriter
...
+                    }
+
+                    // writer streams have to be flushed before reading from memory streams
+                    // make sure that streamwriter is not closed before reading from memory stream
+                    outputStreamWriter.Flush();
+                    errorStreamWriter.Flush();
+
+                    outputStream.Position = 0;
+                    errorStream.Position = 0;
+                    var outputString = System.Text.Encoding.UTF8.GetString(outputStream.GetBuffer(), 0, (int)outputStream.Length);
```
vicancy: >outputStream.GetBuffer(), 0, (int)outputStream.Length [](start = 75, length = 53)

use `outputStream.ToArray())`?

vicancy:
Hi @senglin Sorry for the late response, I was on vacation and back to internet now, the code looks good to me except one minor issue as the comment shows. Again, thanks for your patience and contribution to docfx. 😄

senglin:
@vicancy No worries! Thanks for the helpful reviews. Glad to contribute to this project.

I have made the recommended changes and retested on my side.

src/Microsoft.DocAsCode.Common/CommandUtility.cs
```diff
@@ -113,24 +114,47 @@ public static int RunCommand(CommandInfo commandInfo, StreamWriter stdoutWriter
...
+                    errorStreamWriter.Flush();
+
+                    var outputString = System.Text.Encoding.UTF8.GetString(outputStream.ToArray());
+                    var errorString = System.Text.Encoding.UTF8.GetString(errorStream.ToArray());
+
+                    // Allow caller to decide what to do with the output and error logs 
+                    processOutput?.Invoke(outputString);
+                    processError?.Invoke(errorString);
+                }
+
```
vicancy: remove extra lines

vicancy:
@senglin Thanks for reminder, I will merge it. 😄

senglin:
@vicancy Thank you!

PR #2087, opened 2017-09-27
implment resolve uid by xref service.

vwxyzh:


src/Microsoft.DocAsCode.Build.Engine/XrefServiceResolver.cs
```diff
@@ -33,55 +40,56 @@ public XrefServiceResolver(ImmutableArray<string> xrefServiceUrls)
...
                 }
             }
             return unresolvedUidList;
         }
 
-        private async Task<XRefSpec> ResolveAsync(string uid)
+        public async Task<XRefSpec> ResolveAsync(string uid)
         {
-            var d = new Dictionary<string, string> { ["uid"] = uid };
+            var d = new Dictionary<string, string> { ["uid"] = Uri.EscapeDataString(uid) };
```
vicancy: >EscapeDataString [](start = 67, length = 16)

Evaluate to responsible for var's uri escape?

PR #2090, opened 2017-09-28
Share git existence status

vicancy:


src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -55,45 +57,15 @@ public static GitDetail TryGetFileDetail(string filePath)
 
         public static GitDetail GetFileDetail(string filePath)
         {
-            if (string.IsNullOrEmpty(filePath))
+            if (string.IsNullOrEmpty(filePath) || !ExistGitCommand())
```
qinezh: >ExistGitCommand [](start = 51, length = 15)

Do we need leave a message to tell user reason why the git information doesn't be generated?

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -55,45 +57,15 @@ public static GitDetail TryGetFileDetail(string filePath)
...
             {
                 return null;
             }
 
-            if (!Path.IsPathRooted(filePath))
-            {
-                throw new GitException($"{nameof(filePath)} should be an absolute path");
-            }
+            var path = Path.Combine(EnvironmentContext.BaseDirectory, filePath).ToNormalizedPath();
```
qinezh: As a generic method, it would be better to avoid using EnvironmentContext from Plugins?

vicancy: It's Environment.CurrentDirectory as default value, I think it is general enough

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
vicancy: Yes, https://github.com/dotnet/docfx/pull/2086 does

PR #2092, opened 2017-09-29
Add LogInfo when git not found

vicancy:


src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -305,9 +305,14 @@ private static bool ExistGitCommand()
                     if (GitCommandExists == null)
                     {
                         GitCommandExists = CommandUtility.ExistCommand(CommandName);
+                        if (GitCommandExists != true)
+                        {
+                            Logger.LogInfo("Looks like Git is not installed globally. We depends on Git to extract repository information for source code and files.");
```
qinezh: depends  => depend?

PR #2093, opened 2017-09-29
update manifest spec

superyyrrzz:


Documentation/spec/docfx_build_manifest_file.md
```diff
@@ -9,13 +9,14 @@ For backward compatibility, in version 1.8, *DocFX* generates both Array model t
 
 In version 1.9, manifest file in Array model will be **removed**.
```
herohua: >will be [](start = 45, length = 7)

is

herohua:
since manifest is in JSON format, would it be better we just provide a JSON schema? :)

superyyrrzz:
@herohua Not sure whether someone will consume the JSON schema... Will OPS use it?

PR #2103, opened 2017-10-09
quick fix for sdp with overwrite markdown.

vwxyzh:


herohua:
@vicancy , @hellosnow , shall we consider temporarily disabling overwrite document loading for SDP?

PR #2106, opened 2017-10-09
add some cases for markdown em.

vwxyzh:


superyyrrzz:
should consider `a_b\_c`?

vwxyzh:
@superyyrrzz `_` is another story, let me fix it later.

PR #2124, opened 2017-10-13
load overwrite files before all other processors and share the result…

vicancy:
… cross them


vwxyzh:
It seems break build in some cases.

vicancy:
Close this PR and move the new overwrite design to SDP

PR #2140, opened 2017-10-20
Update docfx_getting_started.md

mxplusb:
`cinst` is now `choco` as it should be. :)

### DIFF
--- a/Documentation/tutorial/docfx_getting_started.md
+++ b/Documentation/tutorial/docfx_getting_started.md
@@ -26,7 +26,7 @@ For detailed description about DFM, please refer to [DFM](../spec/docfx_flavored
 -----------------------
 
 *Step1.* DocFX ships as a [chocolatey package](https://chocolatey.org/packages/docfx).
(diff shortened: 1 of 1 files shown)


Documentation/tutorial/docfx_getting_started.md
```diff
@@ -26,7 +26,7 @@ For detailed description about DFM, please refer to [DFM](../spec/docfx_flavored
 -----------------------
 
 *Step1.* DocFX ships as a [chocolatey package](https://chocolatey.org/packages/docfx).
-Install docfx through [Chocolatey](https://chocolatey.org/install) by calling `cinst docfx -y`.
+Install docfx through [Chocolatey](https://chocolatey.org/install) by calling `choco docfx -y`.
```
superyyrrzz: choco **install** docfx -y?

mxplusb: I think without the flag is better since it let's users read the terms and agreements if that's something they are interested in doing.

PR #2145, opened 2017-10-24
Use input fal for build.

vwxyzh:


src/docfx/Models/BuildCommandOptions.cs
```diff
@@ -120,5 +120,8 @@ internal class BuildCommandOptions : LogOptions, ICanPrintHelpMessage
 
         [Option("cleanupCacheHistory", HelpText = "If set to true, docfx create a new intermediate folder for cache files, historical cache data will be cleaned up")]
         public bool CleanupCacheHistory { get; set; }
+
+        [Option("inputFALBuilderName", HelpText = "Set the name of input file abstract layer builder.")]
+        public string InputFALBuilderName { get; set; }
```
vicancy: naming: falName?

OsmondJiang:
how about the fallback helper in docfx? do we still use it?

vwxyzh:
@OsmondJiang not in this pr.

PR #2151, opened 2017-10-26
remove DfmFallbackHelper.

vwxyzh:


src/Microsoft.DocAsCode.Dfm/DfmCodeRenderer.cs
```diff
@@ -94,7 +95,14 @@ public virtual StringBuffer RenderFencesFromCodeContent(string codeContent, DfmF
 
         public virtual string FindFile(DfmFencesToken token, IMarkdownContext context)
         {
-            return DfmFallbackHelper.GetFilePathWithFallback(token.Path, context).Item1;
```
OsmondJiang: Do we support backward compatibility? otherwise build will fail until we complete the integration in OPS side?

PR #2166, opened 2017-11-01
update dfm fallbak helper.

vwxyzh:
@OsmondJiang


src/Microsoft.DocAsCode.Dfm/DfmFallbackHelper.cs
```diff
@@ -34,20 +34,24 @@ public static class DfmFallbackHelper
...
                 parentFileDirectoryToDocset = Path.GetDirectoryName(Path.Combine(context.GetBaseFolder(), parent));
             }
 
-            var originalFilePath = Path.Combine(context.GetBaseFolder(), filePathToDocset);
-            var actualFilePath = originalFilePath;
+            var originalFullPath = Path.Combine(context.GetBaseFolder(), filePathToDocset);
+            string actualFilePath = null;
             bool hitFallback = false;
-            if (!EnvironmentContext.FileAbstractLayer.Exists(originalFilePath))
+            if (EnvironmentContext.FileAbstractLayer.Exists(filePathToDocset))
```
OsmondJiang: path to docset?  can we change it to path to repo root? otherwise i we use it in plugins? in plugins we should not have docset logic

vwxyzh: docfx do not know path to repo.

PR #2175, opened 2017-11-07
refactor ValidateAndPrepare

superyyrrzz:


src/Microsoft.DocAsCode.Dfm/CodeRenderer/DfmCodeExtractorException.cs
```diff
@@ -0,0 +1,23 @@
...
+        }
+
+        public DfmCodeExtractorException(string message) : base(message)
+        {
+        }
+
+        public DfmCodeExtractorException(string message, Exception inner) : base(message, inner)
+        {
+        }
+    }
```
vwxyzh: add serialize constructor

PR #2185, opened 2017-11-10
increase default parallelism for http.

vwxyzh:


src/docfx/SubCommands/DocumentBuilderWrapper.cs
```diff
@@ -283,7 +283,7 @@ private static IEnumerable<DocumentBuildParameters> ConfigToParameter(BuildJsonC
                 }
             }
 
-            parameters.MaxHttpParallelism = Math.Max(64, parameters.MaxParallelism * 2);
+            parameters.MaxHttpParallelism = Math.Max(64, parameters.MaxParallelism * 4);
```
superyyrrzz: make it configurable?

PR #2202, opened 2017-11-16
Add the developer dependency attribute to package

paulomorgado:


src/nuspec/docfx.msbuild/docfx.msbuild.nuspec
```diff
@@ -9,6 +10,7 @@
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <licenseUrl>https://github.com/dotnet/docfx/blob/dev/LICENSE</licenseUrl>
     <projectUrl>https://github.com/dotnet/docfx</projectUrl>
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
```
vicancy: Looks like this change does not reflect the purpose described in the title **developer dependency attribute**?

vicancy:
Hey @paulomorgado Thanks for the PR. the `docfx.msbuild` package is actually obsoleted and we are publishing `docfx.console` instead. (Should've removed it to avoid confusion).

paulomorgado:
@vacancy, the [readme instructions](https://github.com/dotnet/docfx) are a bit scarce on that.

What's the recommend way to use docfx with that package?

Either way, it is also missing the developer dependency attribute.

vicancy:
@paulomorgado Totally agree, could you help to also update `docfx.console`? 👍

paulomorgado:
@vicancy, I thought I had, but apparently I didn't follow through all the way to the end.

But now, I'm sure I did: https://github.com/dotnet/docfx/pull/2217

vicancy:
Close this PR as `docfx.msbuild` is now cleaned up

Issue #2203, opened 2017-11-16
"Spacers" in toc.md break filter search in TOC


crazyfx1:
**DocFX Version Used**: 2.26.3

**Template used**: default

**Steps to Reproduce**:

In toc.md I use empty links as spacers to separate the list:

```
# Types
## [Common types](types/commontypes.md)
## ---
## [Type1](types/type1.md)
## [Type2](types/type2.md)
## [Type3](types/type3.md)
...
```

**Expected Behavior**:
Filtering still works.

**Actual Behavior**:
Filtering throws JavaScript error.

PR #2204, opened 2017-11-16
fix: #2203 empty navitems break toc filter

crazyfx1:
Added null check for the name parameter in filterNavItem in docfx.js

#2203

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -470,6 +470,7 @@ $(function () {
 
         function filterNavItem(name, text) {
           if (!text) return true;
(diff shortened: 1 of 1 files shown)


src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -470,7 +470,7 @@ $(function () {
 
         function filterNavItem(name, text) {
           if (!text) return true;
-          if (name.toLowerCase().indexOf(text.toLowerCase()) > -1) return true;
+          if (name != null && name.toLowerCase().indexOf(text.toLowerCase()) > -1) return true;
```
vicancy: `if (name && name....)` ?

PR #2224, opened 2017-11-23
Add --disableGitFeatures flag so that for perf concern we are able to…

vicancy:
… disable git related features


superyyrrzz:
When disableGitFeatures changes from `true` to `false`, metadata command still can't generate git info for incremental issue. Seems it's not a good experience... What about provide only `disableBuildGitFeatures` for now?

vicancy:
@superyyrrzz nice catch! Fix metadata incremental is next step, I am working on that.. In general I'd like to keep the option applied to all commands

superyyrrzz:
Sounds good! Agree that using one option is better.

PR #2225, opened 2017-11-23
package-lock.json should be checked in

qinezh:
`package-lock.json` is intended to be checked into source control. refer to: https://docs.npmjs.com/files/package-lock.json

### DIFF
--- a/src/docfx.website.themes/default/package-lock.json
+++ b/src/docfx.website.themes/default/package-lock.json
@@ -1,0 +1,2825 @@
+{
+  "name": "default",
+  "version": "1.0.0",
+  "lockfileVersion": 1,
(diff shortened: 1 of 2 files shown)


vicancy:
Personally I would say "hold this change" and ignore the file instead for now. This file is to solve the version incompatibility issue. 
Some issues with this file: https://github.com/npm/npm/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20package-lock

I don't see a benefit for us when committing the file for now...

qinezh:
Then how about adding it in `.gitignore` ?

vicancy:
@qinezh Sounds cool to me 😸

PR #2231, opened 2017-11-24
Rewrite to encode inline link

qinezh:


src/Microsoft.DocAsCode.MarkdigMarkdownRewriters/MarkdigMarkdownRenderer.cs
```diff
@@ -28,22 +28,47 @@ public virtual StringBuffer Render(IMarkdownRenderer render, DfmXrefInlineToken
...
+                    {
+                        var mailText = RenderInlineTokens(token.Content, render, context);
 
-            return base.Render(render, token, context);
+                        return $"<{mailText}>";
+                    }
+                    goto default;
+
+                case MarkdownLinkType.NormalLink:
+                    var href = Uri.EscapeUriString(token.Href);
```
superyyrrzz: >Uri.EscapeUriString(token.Href); [](start = 31, length = 32)

will it be double escaped if Href is already escaped?

Issue #2220, opened 2017-11-23
Consolidating languages support to UniversalDocumentProcessor


vicancy:
We are consolidating different languages into going through one UniversalDocumentProcessor, including TypeScript(https://github.com/docascode/type2docfx), JavaScript(https://github.com/docascode/Node2DocFX), and Python(https://github.com/docascode/sphinx-docfx-yaml)... 
* [x] 1. A template for UniversalDocumentProcessor
* [ ] 2. Intuitive tutorial/guidance on how to enable the feature
  * [x] JavaScript
  * [x] TypeScript
  * [ ] Python

Related issues: #2218, #157, #915

crazyfx1:
When is this going to be released? I'm eager to try out the typescript documentation ;)

superyyrrzz:
@crazyfx1 2.29 will contains JavaScript support. We will consider add TypeScript support in 2.30.

crazyfx1:
What is missing for typescript support? I thought I just can use the type2docfx tool to generate yml files and use them the same way as the javascript yml files.

superyyrrzz:
I'm afraid the current template isn't general enough for type2docfx YAMLs. Still need some test to confirm it works.

leopignataro:
@superyyrrzz does 2.29+ really support JavaScript? If so, were can we find documentation on how to use it? A few questions already come to mind:

1. On the metadata/src/files array of our docfx.json, we have the path for csproj files. What will be the case for JS files?
2. What is the syntax for the comments on the JS code?

We're excited with this update, we've been wanting this for quite a while!

crazyfx1:
@leopignataro The documentation can be found here, but I don't know why it's not visible from the TOC ;)
https://dotnet.github.io/docfx/tutorial/universalreference/gen_doc_for_js.html

superyyrrzz:
@leopignataro As in JavaScript, the comments should use [JSDoc](http://usejsdoc.org/) sytnax
@crazyfx1 My fault... forgot changing the TOC...

ovasquez:
Is there any way to help speeding up the support for typescript?

superyyrrzz:
@ovasquez We still have some template issue to handle generic types of typescript, which will lead to missing link for them in html. We are now lack of resource to support this as typescript support is more complicate than expected.

superyyrrzz:
v2.33 supports TypeScript: http://dotnet.github.io/docfx/tutorial/universalreference/gen_doc_for_ts.html

[...]

PR #2234, opened 2017-11-27
add templates for udp

superyyrrzz:
#2220

sample: https://superyyrrzz.github.io/DOCFX-JSREF/api/BatchResult.html


plugins/Microsoft.DocAsCode.Build.UniversalReference/uref.nuspec
```diff
@@ -9,7 +9,7 @@
     <requireLicenseAcceptance>true</requireLicenseAcceptance>
     <licenseUrl>https://github.com/dotnet/docfx/blob/dev/LICENSE</licenseUrl>
     <projectUrl>https://github.com/dotnet/docfx</projectUrl>
-    <description>This is a plugin for DocFX to generate API documentation for projects in various languages. This package contains both plugin DLLs and templates.</description>
```
vicancy: Move into docfx instead?

src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -38,6 +38,7 @@ $(function () {
 
   function breakText() {
     $(".xref").addClass("text-break");
+    $("td em").addClass("text-break");
```
vicancy: sounds not safe, use a special class to indicate `text-break`, for example, `type-name`?

plugins/Microsoft.DocAsCode.Build.UniversalReference/uref.nuspec
superyyrrzz: will do this in another PR

---
In reply to: [153388968](https://github.com/dotnet/docfx/pull/2234#discussion_r153388968) [](ancestors = 153388968)

PR #2237, opened 2017-11-28
Update the logic to rewrite DfmVideoToken

qinezh:


src/Microsoft.DocAsCode.MarkdigMarkdownRewriters/MarkdigMarkdownRenderer.cs
```diff
@@ -46,13 +46,44 @@ public override StringBuffer Render(IMarkdownRenderer render, MarkdownLinkInline
...
+            {
+                if (content == StringBuffer.Empty)
+                {
+                    content += BlockQuoteStartString;
+                    content += item;
+                }
+                else
+                {
+                    content += BlockQuoteJoinString;
+                    content += item;
```
superyyrrzz: >content += item; [](start = 20, length = 16)

can move out of if? Can use String.Join?

qinezh: the content is StringBuffer instead of String?

superyyrrzz: Can convet to StringBuffer finally. Just for simplicity.

PR #2239, opened 2017-11-28
Retry to resolve uid in case of httprequestexception

qinezh:


src/Microsoft.DocAsCode.MarkdigMarkdownRewriters/MarkdigMarkdownRenderer.cs
```diff
@@ -71,21 +73,51 @@ private StringBuffer RenderInlineTokens(ImmutableArray<IMarkdownToken> tokens, I
...
         }
 
-        private async Task<bool> TryResolveUidAsync(string uid)
+        private async Task<bool> CanResolveUidWithRetryAsync(string uid)
+        {
+            var retryCount = 3;
+            var delay = TimeSpan.FromSeconds(3);
+
+            var count = 1;
+            for (; ; )
```
vwxyzh: `while(true)`?

src/Microsoft.DocAsCode.MarkdigMarkdownRewriters/MarkdigMarkdownRenderer.cs
```diff
@@ -71,21 +73,51 @@ private StringBuffer RenderInlineTokens(ImmutableArray<IMarkdownToken> tokens, I
...
+                    count++;
+                }
+
+                await Task.Delay(delay);
+            }
+        }
+
+        private async Task<bool> CanResolveUidAsync(string uid)
         {
             var page = $"https://xref.docs.microsoft.com/query?uid={uid}";
```
vwxyzh: uri data escape.

qinezh: nice catch!

Issue #1969, opened 2017-08-16
docfx metadata fails after installing Visual Studio 2017 Update 15.3


bitbonk:
I just installed  Visual Studio 2017 15.3 and now I cannot run `docfx metadata` anymore because I get the following error:

> [17-08-16 12:22:20.437]Warning:[ExtractMetadata]Workspace failed with: [Failure] Msbuild failed when processing the file 'C:\foo\bar\bar.csproj' with message: The tools version "15.0" is unrecognized. Available tools versions are "14.0", "2.0", "3.5", "4.0".  C:\foo\bar\bar.csproj

The interseting part here is that my csproj file does not even speficy version 15:

```xml
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
```

If I change it to `ToolsVersion="14.0` it works, but I can't change the tools version for now. And if I were to change it, I would like to change it to 15, wich does not work due to the above error.

**DocFX Version Used**:  2.22.2.0

bitbonk:
He seems to have the same probem: http://www.michalkomorowski.com/2017/04/why-i-hate-roslyn-even-more.html

vicancy:
I think I should revert https://github.com/dotnet/docfx/pull/1735, It is not a good idea to contain a compiler inside docfx... otherwise every time VS updates, docfx has to update the compiler accordingly.

vicancy:
Hi @bitbonk does the error only happen to **empty** project or every project?

Sunzhuokai:
@vicancy every project

vicancy:
@Sunzhuokai Thanks, only `Visual Studio 2017 15.3` installed or other versions are also installed?

Sunzhuokai:
`14.0` is available if  `Visual Studio 2015`  installed，if only `Visual Studio 2017 15.3` installed,the avaliable values are ` "2.0", "3.5", "4.0"`. may be it is related to this https://docs.microsoft.com/en-us/visualstudio/msbuild/what-s-new-in-msbuild-15-0

vicancy:
Hi @Sunzhuokai Could you try 2.22.3 to see if it solves the issue, I updated Compiler to 2.3.1

Sunzhuokai:
@vicancy  not work

vicancy:
[...]

PR #2242, opened 2017-11-29
Fix 1969 by manually set vs 2017 path

vicancy:
Fix based on:
https://github.com/dotnet/docfx/issues/1969#issuecomment-323235324
https://github.com/dotnet/roslyn/issues/21799#issuecomment-343695700


vicancy:
Fix https://github.com/dotnet/docfx/issues/1969

PR #2246, opened 2017-11-30
Update the logic to migrate normal link

qinezh:


test/Microsoft.DocAsCode.MarkdownRewriters.Tests/MarkdigMarkdownRewritersTests.cs
```diff
@@ -103,6 +103,20 @@ public void TestMarkdigMarkdownRewriters_MailTo()
...
+
+        [Fact]
+        [Trait("Related", "MarkdigMarkdownRewriters")]
+        public void TestMarkdigMarkdownRewriters_MailTo2()
+        {
+            var source = "<Mailto:docs@microsoft.com>";
+            var expected = "<docs@microsoft.com>\n\n";
+
             var result = Rewrite(source, "topic.md");
             Assert.Equal(expected.Replace("\r\n", "\n"), result);
```
superyyrrzz: Does the above case already cover this one?

qinezh: nice catch!

PR #2250, opened 2017-12-01
Migrate Microsoft.DocAsCode.Common

vicancy:
PR with @tintoy 's change


src/Microsoft.DocAsCode.Common/LogicalCallContext.cs
```diff
@@ -1,8 +1,11 @@
-﻿using System.Collections.Concurrent;
-using System.Threading;
+﻿// Copyright (c) Microsoft. All rights reserved.
```
tintoy: Oops! Sorry, I should have caught that 😊

PR #2261, opened 2017-12-04
Remove extra > in blockquote

qinezh:


src/Microsoft.DocAsCode.MarkdownLite/MarkdownRenderer.cs
```diff
@@ -350,7 +350,7 @@ public virtual StringBuffer Render(IMarkdownRenderer render, MarkdownBlockquoteB
             {
                 content += render.Render(t);
             }
-            var contents = content.ToString().Split('\n');
+            var contents = content.ToString().TrimEnd().Split('\n');
```
vwxyzh: trim '\n' only.

PR #2274, opened 2017-12-07
try use lock for TocMap to improve perf

superyyrrzz:


andreycha:
@vicancy will 2.28.3 appear in Nuget?

superyyrrzz:
@andreycha We are pushing packages to Nuget, but meet some issue. Please wait a moment.

BTW, 2.28.3 only fixed some performance issue when the build containing huge amount of TOCs. Is it what you need in 2.28.3?

andreycha:
@superyyrrzz I have tried it out and noticed perf improvements on my projects, so yes. Anyway, it's already in Nuget. Thanks!

PR #2275, opened 2017-12-07
Override heading markdown renderer

qinezh:
To avoid unnecessary change


src/Microsoft.DocAsCode.MarkdigMarkdownRewriters/MarkdigMarkdownRenderer.cs
```diff
@@ -125,6 +127,30 @@ public override StringBuffer Render(IMarkdownRenderer render, MarkdownHtmlBlockT
             return result;
         }
 
+        public override StringBuffer Render(IMarkdownRenderer render, MarkdownHeadingBlockToken token, MarkdownBlockContext context)
```
vwxyzh: lheading

PR #2282, opened 2017-12-08
remove eii dependency

vicancy:


vwxyzh:
@vicancy update test case

Issue #2286, opened 2017-12-10
Merge section in docfx.json does not support more than one merge operation


lextm:
**DocFX Version Used**: 2.28.2

**Steps to Reproduce**:

I am combining two C# projects into a single web site, so I use the following `docfx.json`,
```
{
  "metadata": [
    {
      "src": [
        {
          "files": [
            "sharpsnmplib/SharpSnmpLib/*.csproj"
          ],
          "exclude": [
            "**/bin/**",
            "**/obj/**"
          ],
          "cwd": "."
        }
      ],
      "properties": {
        "TargetFramework": "net452"
      },
      "dest": "obj/lib/10.0.2_net452/api",
      "shouldSkipMarkup": true
    },
    {
      "src": [
        {
          "files": [
            "sharpsnmplib/SharpSnmpLib/*.csproj"
          ],
          "exclude": [
            "**/bin/**",
            "**/obj/**"
          ],
          "cwd": "."
        }
      ],
      "properties": {
        "TargetFramework": "netstandard1.3"
      },
      "dest": "obj/lib/10.0.2_netstandard1.3/api",
      "shouldSkipMarkup": true
    },
    {
      "src": [
        {
          "files": [
            "sharpmibsuite/SharpSnmpPro.Mib/*.csproj"
          ],
          "exclude": [
            "**/bin/**",
            "**/obj/**"
          ],
          "cwd": "."
        }
      ],
      "properties": {
        "TargetFramework": "net452"
      },
      "dest": "obj/pro/2.0.0_net452/api",
      "shouldSkipMarkup": true,
      "filter": "filterConfig.yml"
    },
    {
      "src": [
        {
          "files": [
            "sharpmibsuite/SharpSnmpPro.Mib/*.csproj"
          ],
          "exclude": [
            "**/bin/**",
            "**/obj/**"
          ],
          "cwd": "."
        }
      ],
      "properties": {
        "TargetFramework": "netstandard2.0"
      },
      "dest": "obj/pro/2.0.0_netstandard2.0/api",
      "shouldSkipMarkup": true,
      "filter": "filterConfig.yml"
    }
  ],
...
```
Then I can use the following to merge the documents for sharpsnmplib,
```
"merge": {
    "content": [
      {
        "files": "*.yml",
        "src": "obj/lib/10.0.2_netstandard1.3/api"
      },
      {
        "files": "*.yml",
        "src": "obj/lib/10.0.2_net452/api"
      }
    ],
    "fileMetadata": {
[...]

PR #2287, opened 2017-12-11
Added multiple merge operation support

lextm:
As stated in #2286 I think multiple merge operations should be allowed.

However, this would be a breaking change to the JSON schema.

### DIFF
--- a/src/docfx/Models/MergeJsonConfig.cs
+++ b/src/docfx/Models/MergeJsonConfig.cs
@@ -3,43 +3,14 @@
 
 namespace Microsoft.DocAsCode
 {
-    using System;
     using System.Collections.Generic;
 
-    using Microsoft.DocAsCode.Common;
-
-    using Newtonsoft.Json;
-
-    [Serializable]
-    public class MergeJsonConfig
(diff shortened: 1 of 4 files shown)


vicancy:
This is a breaking change, we can use a custom JsonConverter to support both, as similar to https://github.com/dotnet/docfx/blob/dev/src/docfx/Models/FileMappingConverter.cs

lextm:
@vicancy I just added the converter as you described. It seems to work. Any idea how I should add `MergeCommandTest` in `docfx.Tests` project?

lextm:
Sounds like GitHub does not sync that commit, https://github.com/lextm/docfx/commit/9cf7d13b9c282fad9b16a2e69f14d1cf60524ceb

vicancy:
Hi @lextm , how about a serialization case in https://github.com/dotnet/docfx/blob/dev/test/docfx.Tests/JsonConverterTest.cs and a Merge case in https://github.com/dotnet/docfx/blob/dev/test/docfx.Tests/CompositeCommandTest.cs ?

vicancy:
Hi @lextm I merged this change, and you can open another PR for test cases. Thanks for your contributions to it.

PR #2307, opened 2017-12-13
[WIP] fix error when link containing encoded invalide char for filename

superyyrrzz:


src/Microsoft.DocAsCode.Build.Engine/MarkupResultUtility.cs
```diff
@@ -102,9 +102,10 @@ private static MarkupResult ParseCore(MarkupResult markupResult, string file, Im
                     anchor = linkFile.Substring(index);
                     linkFile = linkFile.Remove(index);
                 }
+                linkFile = Uri.UnescapeDataString(linkFile);
                 if (RelativePath.IsRelativePath(linkFile))
                 {
-                    var path = (RelativePath)currentFile + RelativePath.FromUrl(linkFile);
+                    var path = (RelativePath)currentFile + (RelativePath)linkFile;
```
vwxyzh: I don't think this is a right fix.
What about a file with name: a%2fb.md
`UnescapeDataString` will decode `%2f` to `/`, and you cannot find `a/b.md`

superyyrrzz: agree that an error should be logged or corrected. Will update the logic of FromUrl.

vwxyzh:
#2315

superyyrrzz:
fixed in #2315

PR #2314, opened 2017-12-15
add group info to flie link info

superyyrrzz:


src/Microsoft.DocAsCode.Plugins/IFileLinkInfo.cs
```diff
@@ -46,8 +46,8 @@ public interface IFileLinkInfo
         bool IsResolved { get; }
 
         /// <summary>
-        /// The group that current link belongs to
+        /// The group information that current link belongs to
         /// </summary>
-        string Group { get; }
```
vwxyzh: breaking change?

fenxu: no worry, I'll make the change with it at the same time.

PR #2350, opened 2018-01-10
support naming convention for overwrite segments

superyyrrzz:


src/Microsoft.DocAsCode.Build.ConceptualDocuments/ConceptualDocumentProcessor.cs
```diff
@@ -62,6 +62,17 @@ public override ProcessingPriority GetProcessingPriority(FileAndType file)
...
             {
+                var filenameWithoutExtension = Path.ChangeExtension(file.File, null);
+
+                // exclude overwrite markdown segments
+                var subExtension = Path.GetExtension(filenameWithoutExtension);
+                if ((".yml".Equals(subExtension, StringComparison.OrdinalIgnoreCase) ||
+                    ".yaml".Equals(subExtension, StringComparison.OrdinalIgnoreCase))
+                    && EnvironmentContext.FileAbstractLayer.Exists(filenameWithoutExtension))
+                {
+                    return ProcessingPriority.NotSupported;
```
vicancy: use `Low`?

src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
```diff
@@ -197,7 +197,13 @@ private void Cleanup(HostService hostService)
                          group file by p).ToList();
 
             var toHandleItems = files.Where(s => s.Key != null);
-            var notToHandleItems = files.Where(s => s.Key == null).SelectMany(s => s).Where(s => s.Type != DocumentType.Overwrite).ToList();
+            var notToHandleItems = files
+                .Where(s => s.Key == null)
+                .SelectMany(s => s)
+                .Where(s => s.Type != DocumentType.Overwrite &&
+                    !s.File.EndsWith(".yaml.md", StringComparison.OrdinalIgnoreCase) &&
```
vicancy: is this a breaking change?

src/Microsoft.DocAsCode.Build.ConceptualDocuments/ConceptualDocumentProcessor.cs
superyyrrzz: There's no dedicated processor for new overwrite, so `Low` can't work.

src/Microsoft.DocAsCode.Build.Engine/SingleDocumentBuilder.cs
superyyrrzz: Yes. If a markdown ends with ".y(a)ml.md", and the company ".y(a)ml" exists, it will not be processed. It's a fairly rare case.

vicancy: How about reporting some warning when markdown engine is not `markdig` and files with `.y(a)ml.md` are detected?

superyyrrzz: Sounds good. Will add it in another PR, after qinezh's markdig extension is merged into docfx code base.

PR #2382, opened 2018-01-26
save overwrite markdown as MarkdownDocument instead of List<Block>

928PJY:


src/Microsoft.DocAsCode.Build.OverwriteDocuments/MarkdownFragmentsCreater.cs
```diff
@@ -68,12 +68,13 @@ private List<MarkdownPropertyModel> PropertySection()
...
                     PropertyName = key,
                     PropertyNameSource = Next(),
-                    PropertyValue = new List<Block>(),
+                    PropertyValue = new MarkdownDocument()
                 };
                 Block block;
                 while ((block = Peek()) != null && !_inlineCodeHeadingRule.Parse(block, out var _))
                 {
-                    item.PropertyValue.Add(Next());
+                    block.Parent?.Remove(block);
```
vicancy: is Remove behavior safe? is there any case when some Rules `Parse` relies on context, for example, does YamlHeader's parse check if its previous sibling is Header?

928PJY: For now, it is safe, but it is better to move this convert to the end of Markdown fragments create.

928PJY:
move this convert to Step2(MarkdownFragmentsCreater)

PR #2386, opened 2018-01-26
Fix heading

pascalberger:
Make Step0 a heading


superyyrrzz:
Thank you @pascalberger !

PR #2436, opened 2018-02-11
put mdfrag componets together

superyyrrzz:


src/Microsoft.DocAsCode.Build.OverwriteDocuments/OverwriteDocumentModelCreater.cs
```diff
@@ -122,20 +129,22 @@ private static void FindOrCreateObject(Dictionary<string, object> currentObject,
...
+                    var listObject = childObject as List<object>;
                     if (listObject != null)
                     {
                         object value;
                         var goodItems = (from item in listObject
-                            where item.TryGetValue(segment.Key, out value) && (value as string).Equals(segment.Value)
-                            select item).ToList();
+                                         where item is Dictionary<string, object> 
+                                            && ((Dictionary<string, object>)item).TryGetValue(segment.Key, out value) 
+                                            && (value as string).Equals(segment.Value)
```
vwxyzh: `(value as string)` -> `(string)value`

PR #2462, opened 2018-02-26
support incremental for fragments

superyyrrzz:


test/Microsoft.DocAsCode.Build.SchemaDriven.Tests/MergeMarkdownFragmentsTest.cs
```diff
@@ -267,6 +267,42 @@ markdown content
             Assert.Null(rawModel["summary"]);
         }
 
+        [Fact]
+        public void TestFragmentsWithIncremental()
+        {
+            // first build
```
vicancy: use test log listener to validate logs when incremental build

src/Microsoft.DocAsCode.Build.SchemaDriven/SchemaDrivenDocumentProcessor.cs
```diff
@@ -141,28 +141,25 @@ public override FileModel Load(FileAndType file, ImmutableDictionary<string, obj
                         {
                             LocalPathFromRoot = localPathFromRoot,
                         };
+                        fm.MarkdownFragmentsModel = new FileModel(
```
vicancy: why we still need create FragmentsModel when the content is null?

superyyrrzz: This model is used to report a dependency. Even md doesn't exist, DocFx should think yaml has a dependency on it.

test/Microsoft.DocAsCode.Build.SchemaDriven.Tests/MergeMarkdownFragmentsTest.cs
```diff
@@ -267,6 +267,78 @@ markdown content
...
+                {
+                    BuildDocument(_files);
+                }
+                Assert.True(File.Exists(_rawModelFilePath));
+                rawModel = JsonUtility.Deserialize<JObject>(_rawModelFilePath);
+                Assert.NotNull(rawModel["summary"]);
+                Assert.Contains("I update a summary", rawModel["summary"].ToString());
+                Assert.NotNull(listener.Items.FirstOrDefault(s => s.Message.StartsWith("There is no template processing document type(s): RESTMixedTest")));
+                Assert.NotNull(listener.Items.FirstOrDefault(s => s.Message.StartsWith("Can't find") && s.Message.EndsWith("/invalid.md.")));
+                listener.Items.Clear();
```
vicancy: Add a check to compare logs from each round. They should be the same.

PR #2469, opened 2018-02-27
Added delegate to resolve a cref in TripleSlashCommentModel.

surban:
- Introduced ResolveCRef delegate in ITripleSlashCommentParserContext.
- It is called when a cref attribute from an XML comment is to be
  resolved.
- The delegate returns the name of a reference to the appropriate target.
- Necessary for F# support, since the F# compiler does not resolve
  the cref attribute to comment ids in triple slash comments.


surban:
Rebased to dev branch as requested.

PR #2472, opened 2018-02-28
add reader with redirection folder

superyyrrzz:


src/Microsoft.DocAsCode.Common/FileAbstractLayer/FolderRedirectionManager.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace Microsoft.DocAsCode.Common
+{
+    using System;
+    using System.Collections.Generic;
+
+    public class FolderRedirectionManager
+    {
+        private Dictionary<string, RelativePath> _redirections = new Dictionary<string, RelativePath>();
+
+        public void AddFolderRedirectionRule(string from, string to)
```
vicancy: >AddFolderRedirectionRule [](start = 20, length = 24)

Use .ctor to construct the redirection rules instead of exposing `Add` method, so that the redirection rules are immutable and thread safety issue can be simplified?

src/Microsoft.DocAsCode.Common/FileAbstractLayer/FolderRedirectionManager.cs
```diff
@@ -0,0 +1,55 @@
...
+            }
+            _redirections[normalizedFrom] = toRel;
+        }
+
+        public RelativePath GetRedirectedPath(RelativePath file)
+        {
+            var fileRel = file.ToString();
+            foreach (var redirection in _redirections)
+            {
+                if (fileRel.StartsWith(redirection.Key, StringComparison.OrdinalIgnoreCase))
```
vicancy: >StartsWith [](start = 28, length = 10)

use RelativePath in Redirection instead?

test/Microsoft.DocAsCode.Common.Tests/RelativePathTest.cs
```diff
@@ -415,5 +415,22 @@ public void TestUrlDecode(string path, string expected)
...
         }
+
+        [Theory]
+        [InlineData("a/b/c", "a/b/", true)]
+        [InlineData("a/b", "a/b", true)]
+        [InlineData("a/b/", "a/b", true)]
+        [InlineData("a/b/c", "a/b/c", true)]
+        [InlineData("a/b/c", "a/b/c/d", false)]
+        [InlineData("a/b/c", "a/b/d", false)]
+        [InlineData("../a/b", "../a", false)]
```
vicancy: >"../a/b", "../a" [](start = 20, length = 16)

what is the result for `a/../b/c` and `b`?

superyyrrzz: It's true. `../` will pop out `a/` in parts stack when constructing RelativePath.

test/Microsoft.DocAsCode.Common.Tests/RelativePathTest.cs
```diff
@@ -415,5 +415,23 @@ public void TestUrlDecode(string path, string expected)
         {
             Assert.Equal(expected, ((RelativePath)path).UrlDecode());
         }
+
+        [Theory]
+        [InlineData("a/b/c", "a/b/", true)]
+        [InlineData("a/b", "a/b", true)]
+        [InlineData("a/b/", "a/b", true)]
```
vwxyzh: not make sense.

test/Microsoft.DocAsCode.Common.Tests/RelativePathTest.cs
```diff
@@ -415,5 +415,28 @@ public void TestUrlDecode(string path, string expected)
...
+        [InlineData("a/b/", "a/b/", true)]
+        [InlineData("a/b/c", "a/b/c", false)]
+        [InlineData("a/b/c", "a/b/c/d", false)]
+        [InlineData("a/b/c", "a/b/d", false)]
+        [InlineData("a/../b/c", "b/", true)]
+        [InlineData("../a/b", "../a", false)]
+        [InlineData("../a/b", "../", false)]
+        [InlineData("../a/b", "../../a", false)]
+        [InlineData("../../", "../", false)]
+        [InlineData("../", "../../", false)]
```
vwxyzh: what about "~/../"?

superyyrrzz: It starts with `../`, which always return false.

src/Microsoft.DocAsCode.Common/Path/RelativePath.cs
```diff
@@ -312,6 +312,48 @@ public bool Equals(RelativePath other)
...
+            }
+            if (_parts.Length < value._parts.Length)
+            {
+                return false;
+            }
+
+            int i;
+            for (i = 0; i < value._parts.Length; i++)
+            {
+                if (value._parts[i] == "")
```
vicancy: >"" [](start = 39, length = 2)

use string.Empty

src/Microsoft.DocAsCode.Common/Path/RelativePath.cs
```diff
@@ -312,6 +312,48 @@ public bool Equals(RelativePath other)
...
+            }
+            if (value._parts[value._parts.Length - 1] != "")
+            {
+                return false;
+            }
+            if (_isFromWorkingFolder ^ value._isFromWorkingFolder)
+            {
+                return false;
+            }
+            if (_parentDirectoryCount > 0 || value._parentDirectoryCount > 0)
```
vicancy: >_parentDirectoryCount [](start = 16, length = 21)

is `../a/b` in directory `../a/`?

superyyrrzz: It's hard to determine whether `../` is in `../../`, so I return false id either path starts with `../`

PR #2477, opened 2018-03-01
add markdownValidatorContext data model

928PJY:


seshenoy:
The IMarkdownValidatorContext must have 2 properties and not just one. 
1. ) Global Rule set file path
2.) Repo specific rule set file path

seshenoy:
looks good.

Issue #2358, opened 2018-01-16
metadata broken with mono and linux


Godgifu:
**DocFX Version Used**:  Latest release with mono 5.4.1.6

**Template used**: Default

**Steps to Reproduce**:

1.  mono docfx.exe metadata

**Expected Behavior**:

It works.

**Actual Behavior**:

```bash
mono ../../Downloads/docfx.exe metadata                                                                                                                                                                                                       ✓  1123  23:36:43 
[18-01-16 10:36:59.052]Info:Config file docfx.json found, start generating metadata...
[18-01-16 10:37:05.369]Info:[ExtractMetadata]Loading projects...
[18-01-16 10:37:05.820]Warning:[ExtractMetadata](/home/mgysin/RiderProjects/lamada/lamada/lamada.csproj)Workspace failed with: [Failure] Msbuild failed when processing the file '/home/mgysin/RiderProjects/lamada/lamada/lamada.csproj' with message: Operation is not supported on this platform.
[18-01-16 10:37:05.873]Warning:[ExtractMetadata]Project '/home/mgysin/RiderProjects/lamada/lamada/lamada.csproj' does not contain any documents.
[18-01-16 10:37:05.891]Info:[ExtractMetadata]Cache for /home/mgysin/RiderProjects/lamada/lamada/lamada.csproj in /home/mgysin/RiderProjects/lamada/lamada/obj/xdoc/cache/final/741176259 is corrupted, rebuild...
[18-01-16 10:37:05.891]Info:[ExtractMetadata]Generating metadata for each project...
[18-01-16 10:37:05.991]Info:[ExtractMetadata]'/home/mgysin/RiderProjects/lamada/lamada/lamada.csproj' keep up-to-date since '1/16/2018 8:59:24 PM', cached intermediate result '/home/mgysin/RiderProjects/lamada/lamada/obj/xdoc/cache/obj/yigx11l9.c61' is used.
[18-01-16 10:37:06.039]Warning:[ExtractMetadata]No metadata is generated for lamada.
[18-01-16 10:37:06.054]Info:Completed in 6894.4386 milliseconds


Build succeeded with warning.
[18-01-16 10:37:06.056]Warning:[ExtractMetadata](/home/mgysin/RiderProjects/lamada/lamada/lamada.csproj)Workspace failed with: [Failure] Msbuild failed when processing the file '/home/mgysin/RiderProjects/lamada/lamada/lamada.csproj' with message: Operation is not supported on this platform.
[...]

PR #2482, opened 2018-03-02
Add net461 build target for common libararies

vicancy:
Fix https://github.com/dotnet/docfx/issues/2358. As with netstandard libarary referenced, new version of *System.Runtime.InteropServices.RuntimeInformation* is introduced by build system with no way to turn around

### DIFF
--- a/src/Microsoft.DocAsCode.Common/Microsoft.DocAsCode.Common.csproj
+++ b/src/Microsoft.DocAsCode.Common/Microsoft.DocAsCode.Common.csproj
@@ -1,8 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="../Shared/base.netstandard.props" />
-  
   <ItemGroup>
     <ProjectReference Include="..\Microsoft.DocAsCode.Plugins\Microsoft.DocAsCode.Plugins.csproj" />
     <ProjectReference Include="..\Microsoft.DocAsCode.YamlSerialization\Microsoft.DocAsCode.YamlSerialization.csproj" />
+  </ItemGroup>
+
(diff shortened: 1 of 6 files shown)


tzachshabtay:
Hi @vicancy , can you explain the fix? 
I have the same "MSBuildNameIgnoreCaseComparer" exception in my own application (unrelated to DocFX), and I'm trying to understand the issue and the fix, but nothing in your changes seem to actually mention "System.Runtime.InteropServices.RuntimeInformation".
How is adding net461 solves the problem?

vicancy:
Hi @tzachshabtay, in our case, it is that when net461 project A.csproj references netstandard2.0 project B.csproj, msbuild uses a newer version of `System.Runtime.InteropServices.RuntimeInformation` inside netcore2.0 sdk even if I specify a lower version in my A.csproj because this assembly is a MUST have assembly for netstandard project. Looks like this newer version of `System.Runtime.InteropServices.RuntimeInformation` is not supported by Mono, so as a workaround, I uses net461 all the way down.

tzachshabtay:
Thanks for the explanation @vicancy , is there an issue on Mono for this? Is this the issue? https://github.com/mono/mono/issues/7011

vicancy:
@tzachshabtay yes that is! didn't find that one before, thanks for pointing that out.

vicancy:
Are you meeting the same issue @tzachshabtay ? does the workaround apply for you?

tzachshabtay:
@vicancy yes, I'm pretty sure I'm meeting the same issue. I haven't tried the workaround. As mono supposedly fixed the issue I think I'm going to wait for a version with the fix and try that first.

PR #2488, opened 2018-03-05
Make GetMulitpleCrefInfo honor refId attributes like GetMultipleLinkInfo

surban:
- Checks the refId attribute also for `<exception>` tags. 
- This makes the changes introduced by #2469 for resolution of F# symbol names also applicable to exceptions.


surban:
I am working on some test cases. But it will take some time, since I need the restructure the code a bit to be able to call methods from outside for testing.

PR #2489, opened 2018-03-05
Remove projects that could not be loaded from project cache.

surban:
- This fixes "Object reference not set to an instance of an object"
  if a project failed to load for some reason or is of unsupported
  project type.

This is not specific to F# support, but was found while testing project load error handling. I am not sure if this is the optimal way to do it.


vicancy:
Hi @surban revisit this PR, and I think when project.Value is `null`, we can also set it to cache and no need to clear it for a second try. It is the F# loader to do the null check before using the `AbstractProject`

superyyrrzz:
@surban Is there any update on this?

superyyrrzz:
Close this as existing for a long time. Feel free to reopen this if needed.

PR #2501, opened 2018-03-08
extract interfaces for validation from MarkdigEngine.Extensions to Pl…

928PJY:
…ugins


vwxyzh:
please move to a new package.

PR #2510, opened 2018-03-09
Optimise getHierarchy

### LINKED ISSUE #2509: API doc pages for large classes perform very badly on the front end
**DocFX Version Used**: 2.32.2, tested on OSX in Chrome 64 and Firefox 57

**Template used**: `default`

**Steps to Reproduce**:

1. Generate API documentation for a large class (the one I used has around 2500 methods; its generated HTML is around 6.5MB).
2. Try to look at the documentation in a browser

**Expected Behavior**:
Can view the docs relatively quickly

**Actual Behavior**:
The page takes a very long time to load (around two minutes for my example) and does not respond in the mean time.

I did a little profiling in the Chrome inspector. It looks like the HTML itself renders in around 10 seconds on my machine, which is slow but I think acceptable for such a large page. The vast majority of the time is spent in JS, in the `getHierarchy` function. It looks like this function repeatedly traverses the DOM, so I suspect that's what the problem is.

I've tested an alternative algorithm which simply grabs all the `h` tags and then traverses that list once to build the hierarchy. It runs much faster - like, 1000 times faster - and I think the code is simpler too. Would you accept a PR with this improved implementation?

benjamin-hodgson:
This fixes #2509.

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -555,59 +555,65 @@ $(function () {
 
     function getHierarchy() {
       // supported headers are h1, h2, h3, and h4
-      // The topest header is ignored
-      var selector = ".article article";
-      var affixSelector = "#affix";
-      var headers = ['h4', 'h3', 'h2', 'h1'];
-      var hierarchy = [];
-      var toppestIndex = -1;
-      var startIndex = -1;
-      // 1. get header hierarchy
-      for (var i = headers.length - 1; i >= 0; i--) {
-        var header = $(selector + " " + headers[i]);
-        var length = header.length;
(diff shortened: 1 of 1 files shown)


src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -555,59 +555,65 @@ $(function () {
...
+        var parent = parentFrame.siblings[parentFrame.siblings.length - 1];
+        $.each(childrenToAttach.siblings, function(i, child) {
+          parent.items.push(child);
+        });
+      }
+
+      var topLevel = stack.pop().siblings;
+      if (topLevel.length === 1) {  // if there's only one topmost header, dump it
+        return topLevel[0].items;
+      }
```
superyyrrzz: No need to check this? A stand alone root item can also be there to keep completeness.

superyyrrzz:
Thank you for contribution @benjamin-hodgson

PR #2520, opened 2018-03-13
fix anchor hover weirdness by implementing fixed header workaround in JS

### LINKED ISSUE #2508: h*::before CSS produces weird hovering behaviour
**DocFX Version Used**: 2.32.2, tested in Chrome 64 on OSX

**Template used**: `default`

**Steps to Reproduce**:

1. Generate some API docs and open them in a browser
2. Hover over some of the section headings
3. Look at the heading tags in the inspector

**Expected Behavior**:
Heading anchor links should appear when hovering over the heading.
Headings should not overlap each other.

**Actual Behavior**:
Anchor links appear when hovering somewhere above the heading.
Headings are much taller than they should be and overlap each other.

gif

This appears to be due to [this chunk of CSS](https://github.com/dotnet/docfx/blob/dev/src/docfx.website.themes/default/styles/docfx.css#L26-L38), which gives all the `h` tags a `before` with a large height, artificially blowing up the height of the `h` itself. Do you know what it's for? The comment is a little bit cryptic, and removing it doesn't appear to have any ill effects.

Would you accept a pull request fixing this?

benjamin-hodgson:
This fixes #2508.

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -8,6 +8,7 @@ $(function () {
   var hide = 'hide';
   var util = new utility();
 
+  workAroundFixedHeaderForAnchors();
   highlight();
   enableSearch();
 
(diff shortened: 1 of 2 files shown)


src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -8,6 +8,7 @@ $(function () {
   var hide = 'hide';
   var util = new utility();
 
+  workAroundFixedHeaderForAnchors();
```
vicancy: naming: lowercase `Around`'s `a` as `workaround` is a word

benjamin-hodgson: "workaround" is a noun but when it's a verb it's "work around".

PR #2526, opened 2018-03-16
enable xref shortcut in markdig

yishengjin1413:


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Xref/XrefInlineShortParser.cs
```diff
@@ -16,7 +16,8 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
 
     class XrefInlineShortParser : InlineParser
     {
-        private const string Punctuation = ".,;:!?`~";
+        private const string ContinuableCharacters = ".,;:!?~";
+        private const string StopCharacters = @"""'<>";
```
vwxyzh: also add `[]|`?
`[]` defend for markdown links.
`|` defend for table cell.

yishengjin1413: I see. Have update.

PR #2540, opened 2018-03-22
Fix tabbed content bug in default js (#2538)

### LINKED ISSUE #2538: Tabbed content switch doesn't work if Tab name contains not plain text
**DocFX Version Used**:  2.31

**Template used**: default

**Steps to Reproduce**:

1.  Add a markdown document with text like this in the project
```
# [**Tab Name**](#tab/tab-id-1)
Content1 ...
# [**Tab Name**](#tab/tab-id-2)
Content2 ...
---
```

> Note that "Tab Name" is bold

2. Build the docfx project
3. Open result document in a browser

**Expected Behavior**: Page with tabbed content that can be switched

**Actual Behavior**: Tabbed content can't be switched

Click event handling doesn't work if you click something other than HTMLAnchorElement

vicancy:
Fix #2538

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -37,6 +37,8 @@ $(function () {
     renderTabs();
   }
 
+  // window.addEventListener('content-update', contentUpdate);
+
   function breakText() {
     $(".xref").addClass("text-break");
     var texts = $(".text-break");
@@ -181,7 +183,7 @@ $(function () {
           }
           searchData = JSON.parse(this.responseText);
           for (var prop in searchData) {
-            if (searchData.hasOwnProperty(prop)){
+            if (searchData.hasOwnProperty(prop)) {
               lunrIndex.add(searchData[prop]);
             }
           }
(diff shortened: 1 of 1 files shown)


src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -37,6 +37,8 @@ $(function () {
     renderTabs();
   }
 
+  // window.addEventListener('content-update', contentUpdate);
```
superyyrrzz: > // window.addEventListener('content-update', contentUpdate); [](start = 1, length = 61)

remove?

src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -719,227 +721,269 @@ $(function () {
...
+            return true;
+          }
+        }
+      }
+      return false;
+    }
+
+    function notifyContentUpdated() {
+      // window.dispatchEvent(new CustomEvent('content-update'));
+    }
```
superyyrrzz: remove?

src/docfx.website.themes/default/styles/docfx.js
vicancy: Keep and add a comment

src/docfx.website.themes/default/styles/docfx.js
vicancy: Keep and add a comment

Issue #2491, opened 2018-03-06
MSBuild errors when generating metadata for SDK based net461 project and VS 2017 15.6


GeorgDangl:
**DocFX Version Used**: 2.32.0

**Template used**: `default` with an additional custom css file

**Steps to Reproduce**:

1. [Checkout the repro project](https://github.com/GeorgDangl/Nuke.WebDeploy)  
  The project uses `net461` as target framework.
2. Have at least Visual Studio 2017 15.6 installed (same issue with 15.6 Preview 7.0)
3.  Run `powershell ./build.ps1 BuildDocFxMetadata` at the solution root

**Expected Behavior**:

Generate no warnings.

**Actual Behavior**:

Unclear MSBuild warnings are present in the output:

    > C:\Users\Georg\.nuget\packages\docfx.console\2.32.0\tools\docfx.exe metadata "D:\Visual Studio Projects\Nuke.WebDeploy\docfx.json" --logLevel Verbose
    [18-03-06 04:54:21.659]Verbose:Config file "D:\Visual Studio Projects\Nuke.WebDeploy\docfx.json" is used.
    [18-03-06 04:54:21.697]Info:[ExtractMetadata]Using msbuild E:\Visual Studio 2017\MSBuild\15.0\Bin as inner compiler.
    [18-03-06 04:54:21.782]Info:[ExtractMetadata]Loading projects...
    [18-03-06 04:54:21.782]Verbose:[ExtractMetadata](D:/Visual Studio Projects/Nuke.WebDeploy/src/Nuke.WebDeploy/Nuke.WebDeploy.csproj)Loading project...
    [18-03-06 04:54:22.792]Warning:[ExtractMetadata](D:/Visual Studio Projects/Nuke.WebDeploy/src/Nuke.WebDeploy/Nuke.WebDeploy.csproj)Workspace failed with: [Failure] Msbuild failed when processing the file 'D:\Visual Studio Projects\Nuke.WebDeploy\src\Nuke.WebDeploy\Nuke.WebDeploy.csproj' with message: MSB0001: Internal MSBuild Error: Missing resource ''
    [18-03-06 04:54:22.903]Verbose:[ExtractMetadata](D:/Visual Studio Projects/Nuke.WebDeploy/src/Nuke.WebDeploy/Nuke.WebDeploy.csproj)Project D:\Visual Studio Projects\Nuke.WebDeploy\src\Nuke.WebDeploy\Nuke.WebDeploy.csproj loaded.
[...]

PR #2550, opened 2018-03-26
WIP: Avoid shipping conflicting MSBuild version

rainersigwald:
The Microsoft.Build.Locator package handles finding MSBuild. It will
find the installed version, avoiding problems like #2491 that result
from mismatched MSBuild assembly versions.

This change relies on Microsft.Build.Locator to find MSBuild and avoids
shipping MSBuild assemblies next to docfx.exe that could cause version
conflicts.

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference/MsBuildEnvironmentScope.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference/MsBuildEnvironmentScope.cs
@@ -9,7 +9,7 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
     using System.Linq;
     using System.Text.RegularExpressions;
 
-    using Microsoft.Build.MSBuildLocator;
+    using Microsoft.Build.Locator;
     using Microsoft.DocAsCode.Common;
     using Microsoft.DocAsCode.Exceptions;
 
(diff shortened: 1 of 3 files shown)


rainersigwald:
@vicancy This is some work to move toward a more robust consumption of MSBuild. It should be immune to problems like #2491 and keep y'all from having to continuously update the MSBuild references as we release.

This isn't ready to go in yet, though: it breaks the no-VS-only-dotnet-SDK scenario, which was relying on having a docfx-private version of the MSBuild engine available. To get that back, we'll need to either implement https://github.com/Microsoft/MSBuildLocator/issues/21 or work around it here.

Thoughts on this approach? See something I'm missing?

vicancy:
This approach sounds terrific! Thanks @rainersigwald for saving me from fixing those keep-to-latest-msbuild bugs...

Correct me if I misunderstand anything: current PR can resolve the issue with VS installed case. But still WIP because no-VS-only-dotnet-SDK and mono case still needs docfx-private version of the MSBuild engine. And Microsoft/MSBuildLocator#21 can solve both no-VS-only-dotnet-SDK and mono case perfectly. (Cheers!)

Then I am very much looking forward to Microsoft/MSBuildLocator#21 😄

src/Microsoft.DocAsCode.Metadata.ManagedReference/MsBuildEnvironmentScope.cs
```diff
@@ -29,21 +29,6 @@ public MSBuildEnvironmentScope()
 
         private EnvironmentScope GetScope()
         {
-            var vsInstallDirEnv = Environment.GetEnvironmentVariable(VSInstallDirKey);
```
vicancy: why remove these lines? as when user specifies ENV, we don't even need to go through the following check...

superyyrrzz:
@rainersigwald  Is there any update on this?

superyyrrzz:
Close this as existing for a long time. Feel free to reopen this if needed.

PR #2551, opened 2018-03-27
Add Roadmap for v3

vicancy:


superyyrrzz:
Find a missing item from https://github.com/docascode/docfx-feature-proposals/blob/master/wishlist_for_v3.md

Target to NetStantard 2.0.

vicancy:
@superyyrrzz we will always target netcore and x-platforms. But I think that is part of the build script instead of a single item of Roadmap

superyyrrzz:
trace and progress?

PR #2552, opened 2018-03-27
markdown is editable in fragments by default

superyyrrzz:


src/Microsoft.DocAsCode.Build.SchemaDriven/SchemaExtensions.cs
```diff
@@ -13,21 +13,19 @@ public static class SchemaExtensions
...
         /// <summary>
         /// Return if a property is legal to appear in markdown fragmetns
-        /// </summary>
-        public static bool IsLegalInFragments(this BaseSchema schema)
+        /// </summary
+        public static bool IsLegalInFragments(this BaseSchema schema) => IsEditable(schema);
+
+        private static bool IsEditable(this BaseSchema schema)
         {
+            if (schema.ContentType == ContentType.Markdown)
```
928PJY: >schema [](start = 16, length = 6)

do we need check null?

superyyrrzz: resolved

PR #2563, opened 2018-03-29
Add TagParameters config and pass it down to host service

vicancy:


superyyrrzz:
No comments on this PR. A comments related to spec: don't use `-` as separator like `docsetname-uid`, as `-` is a common char in docset name

PR #2572, opened 2018-03-30
MonikerRange infinite loop bug

daxianji007:


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MonikerRange/MonikerRangeParser.cs
```diff
@@ -61,7 +61,7 @@ public override BlockState TryOpen(BlockProcessor processor)
             var range = StringBuilderCache.Local();
             c = slice.CurrentChar;
 
-            while (c != '"')
+            while (c != '\0' && c != '"')
```
vwxyzh: and `\n`?

daxianji007: Markdig split the content into slice by '\n'. So every slice does not have '\n' but only '\0' as ending

PR #2587, opened 2018-04-02
log a warning when there is a markdown content in YAML section of fragments

928PJY:
log a warning when there is a markdown content in YAML section of fragments


src/Microsoft.DocAsCode.Build.SchemaDriven/Processors/FragmentsValidationInterpreter.cs
```diff
@@ -39,6 +39,14 @@ public object Interpret(BaseSchema schema, object value, IProcessContext context
                 return value;
             }
 
+            if (markdownDocument == null && schema.ContentType == ContentType.Markdown)
+            {
+                Logger.LogWarning(
+                $"There is an invalid yaml item: `{path.Trim('/')}`: the contentType of this property in schema cannot be `markdown`, please move it to content section",
```
vicancy: >There is an invalid yaml item: `{path.Trim('/')}`: the contentType of this property in schema cannot be `markdown`, please move it to content section [](start = 18, length = 149)

refine message: `markdown` property `{path}` is not allowed inside a YAML code block

Issue #2597, opened 2018-04-06
Using the search box before the search index is ready shows no result


mmaitre314:
**DocFX Version Used**: 2.33.2

**Template used**: `default`

**Steps to Reproduce**:

1. Open http://dotnet.github.io/docfx/tutorial/docfx_getting_started.html
2. Right away enter 'started' in the search box

**Expected Behavior**: after the index is loaded, shows a list of search results

**Actual Behavior**: the message 'Search Results for "started"' is displayed and nothing else happens.

There is a race condition here in docfx.js:

```
      indexReady.promise().done(function () {
        $("body").bind("queryReady", function () {
          worker.postMessage({ q: query });
        });
      });
```

when the promise completes and `query` has a value already, it needs to post it to the worker. Otherwise the query never runs.

Proposed fix (PR getting prepared):

```
      indexReady.promise().done(function () {
        $("body").bind("queryReady", function () {
          worker.postMessage({ q: query });
        });
        if (query && (query.length >= 3)) {
          worker.postMessage({ q: query });
        }
      });
```

PR #2598, opened 2018-04-06
Fix a race condition when loading the search index

mmaitre314:
There is a race condition here in docfx.js:

```
      indexReady.promise().done(function () {
        $("body").bind("queryReady", function () {
          worker.postMessage({ q: query });
        });
      });
```

when the promise completes and `query` already has a value, the code needs to post `query` to the worker. Otherwise the query never runs.

The PR also displays a spinner while the search index is loaded as index loading takes some time on larger docs.

#2597

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.css
+++ b/src/docfx.website.themes/default/styles/docfx.css
@@ -354,6 +354,28 @@ article section {
   text-align: center;
 }
 
+#search-results p .index-loading {
+  animation: index-loading 1.5s infinite linear;
+  -webkit-animation: index-loading 1.5s infinite linear;
+  -o-animation: index-loading 1.5s infinite linear;
+  font-size: 2.5rem;
+}
+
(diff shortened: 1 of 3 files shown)


src/docfx.website.themes/default/styles/docfx.css
```diff
@@ -354,6 +354,28 @@ article section {
...
+}
+
+@-webkit-keyframes index-loading {
+    from { -webkit-transform: rotate(0deg);}
+    to { -webkit-transform: rotate(360deg);}
+}
+
+@-o-keyframes index-loading {
+    from { -webkit-transform: rotate(0deg);}
+    to { -webkit-transform: rotate(360deg);}
```
superyyrrzz: `-webkit-transform` => `-o-transform` ?

mmaitre314: Fixed the typo

PR #2605, opened 2018-04-09
add warning for moniker range

daxianji007:


vwxyzh:
@daxianji007 question, what about following content:
```md
::: moniker ....
some text
```

PR #2609, opened 2018-04-10
add json utility

928PJY:


src/docfx/lib/JsonUtililty.cs
```diff
@@ -0,0 +1,104 @@
...
+using Newtonsoft.Json.Serialization;
+
+namespace Microsoft.Docs
+{
+    /// <summary>
+    /// Provide Utilities of Json
+    /// </summary>
+    public static class JsonUtililty
+    {
+        private static readonly ThreadLocal<JsonSerializer> s_defaultSerializer = new ThreadLocal<JsonSerializer>(
```
yufeih: This does not need to be ThreadLocal. JsonSerializer is thread safe here.

src/docfx/lib/JsonUtililty.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        /// <summary>
+        /// Serialize an object to TextWriter
+        /// </summary>
+        public static void Serialize(TextWriter writer, object graph, bool isStable = false, Formatting formatting = Formatting.None, JsonSerializer serializer = null)
+        {
+            var localSerializer = serializer ?? s_defaultSerializer.Value;
+            if (isStable)
+            {
+                localSerializer.Formatting = Formatting.Indented;
```
yufeih: Do not mutate the state of localSerializer. you can use `JsonConvert.SerializeObject`

928PJY: good point, but `JsonConvert.SerializeObject` can not a take `TextWriter` but only `Object`.
will reset `localSerializer.Formatting` after using it

---
In reply to: [180296892](https://github.com/dotnet/docfx/pull/2609#discussion_r180296892) [](ancestors = 180296892)

src/docfx/lib/JsonUtililty.cs
```diff
@@ -0,0 +1,104 @@
...
+
+        private static readonly JsonMergeSettings s_jsonMergeSettings = new JsonMergeSettings
+        {
+            MergeArrayHandling = MergeArrayHandling.Replace,
+        };
+
+        /// <summary>
+        /// Serialize an object to TextWriter
+        /// </summary>
+        public static void Serialize(TextWriter writer, object graph, bool isStable = false, Formatting formatting = Formatting.None, JsonSerializer serializer = null)
```
vicancy: >isStable [](start = 75, length = 8)

what does "stable" mean? do we need such concept?

928PJY: stable means stable(ordered) output, which is a config item in ops-cli,

---
In reply to: [180306534](https://github.com/dotnet/docfx/pull/2609#discussion_r180306534) [](ancestors = 180306534)

vicancy: Is this really a need? or should we always keep it stable?

---
In reply to: [180314569](https://github.com/dotnet/docfx/pull/2609#discussion_r180314569) [](ancestors = 180314569,180306534)

src/docfx/lib/JsonUtililty.cs
```diff
@@ -0,0 +1,109 @@
...
+                NullValueHandling = NullValueHandling.Ignore,
+                ReferenceLoopHandling = ReferenceLoopHandling.Serialize,
+                Converters =
+                {
+                    new StringEnumConverter { CamelCaseText = true },
+                },
+                ContractResolver = new CamelCasePropertyNamesContractResolver(),
+            };
+
+        private static readonly JsonMergeSettings s_jsonMergeSettings = new JsonMergeSettings
```
vicancy: >s_jsonMergeSettings [](start = 50, length = 19)

not used?

src/docfx/lib/JsonUtililty.cs
928PJY: stable is just for diff tools, always keep is stable will  bring additional cost


---
In reply to: [180316156](https://github.com/dotnet/docfx/pull/2609#discussion_r180316156) [](ancestors = 180316156,180314569,180306534)

src/docfx/lib/JsonUtililty.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (isStable)
+            {
+                localSerializer.Formatting = Formatting.Indented;
+                try
+                {
+                    localSerializer.Serialize(writer, Stablize(JToken.FromObject(graph, localSerializer)));
+                }
+                finally
+                {
+                    localSerializer.Formatting = Formatting.None;
```
OsmondJiang: >Formatting = Formatting.None [](start = 36, length = 28)

how do you know the original formatting is None?

Issue #2582, opened 2018-04-02
YamlUtility


vicancy:
One idea is to stream with `YamlSerialization` and parse with `Newtonsoft.Json`

PR #2626, opened 2018-04-12
add yaml utility

928PJY:
#2582


src/docfx/lib/YamlUtility.cs
```diff
@@ -0,0 +1,162 @@
...
+                    {
+                        arr.Add(ToJson(item));
+                    }
+                    return arr;
+                }
+                throw new NotSupportedException($"Unknown yaml node type {node.GetType()}");
+            }
+        }
+
+        class JsonContractResolver : DefaultContractResolver
```
yufeih: This ContractResolver was to support deserializing into readonly fields and properties in the form of:

- public readonly string Field
- public string GetOnlyProperty { get; }

Since our style does not encourage *readonly fields*, lets drop *readonly fields* and add the following property tests:

1. Can deserialize into get only property
2. When deserialize into get only property with default value, `null` and `undefined` does not override default value. E.g. when loading `public string A { get; } = "default"` from `{ "a": null }` or `{ }`, the result object should be `{ "a": "default" }`

vicancy: share the same logic with JsonUtility that both support readonly properties or both not...

---
In reply to: [180948737](https://github.com/dotnet/docfx/pull/2626#discussion_r180948737) [](ancestors = 180948737)

vicancy: >class [](start = 8, length = 5)

use `private sealed` class

src/docfx/lib/YamlUtility.cs
```diff
@@ -0,0 +1,162 @@
...
+                        }
+                    }
+                    return new JValue(scalar.Value);
+                }
+                if (node is YamlMappingNode map)
+                {
+                    var obj = new JObject();
+                    foreach (var (key, value) in map)
+                    {
+                        obj[key.ToString()] = ToJson(value);
```
vicancy: >key.ToString() [](start = 28, length = 14)

shall we throw when it is `string` or primitive type

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -0,0 +1,213 @@
...
+
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+using Xunit;
+
+namespace Microsoft.Docs
+{
+    public class YamlUtilityTest
```
vicancy: >YamlUtilityTest [](start = 17, length = 15)

add test cases covered in docfx\test\Microsoft.DocAsCode.Common.Tests\YamlSerializationTest.cs, such as
1. big integer
2. anchors support

src/docfx/lib/YamlUtility.cs
```diff
@@ -0,0 +1,162 @@
...
+            var stream = new YamlStream();
+
+            stream.Load(reader);
+            if (stream.Documents.Count != 1)
+            {
+                throw new NotSupportedException("Does not support mutiple YAML documents");
+            }
+            return ToJson(stream.Documents[0].RootNode);
+
+            JToken ToJson(YamlNode node)
```
vicancy: >ToJson [](start = 19, length = 6)

let's move this out to a private method :blush:

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -0,0 +1,213 @@
...
+            Assert.Equal("ItemA", value.ValueList[0]);
+            Assert.Equal("True", value.ValueList[1]);
+            Assert.Equal("3", value.ValueList[2]);
+            Assert.Equal("ItemB", value.ValueList[3]);
+            Assert.Equal((long)2, value.ValueBasic.B);
+            Assert.Equal("Good3!", value.ValueBasic.C);
+            Assert.Equal(false, value.ValueBasic.D);
+        }
+
+        public class BasicClass
```
OsmondJiang: > public class BasicClass [](start = 7, length = 24)

please notice the build warnings, u can disable it or follow it

 CA1034

src/docfx/lib/YamlUtility.cs
928PJY: there is not a way to deserialize a get-only property, because there is not a setter for this property, value cannot be set to this property even if we set `Writable` to true.

---
In reply to: [180956068](https://github.com/dotnet/docfx/pull/2626#discussion_r180956068) [](ancestors = 180956068,180948737)

PR #2627, opened 2018-04-12
add file glob

928PJY:
#2584

### DIFF
--- a/src/docfx/lib/FileGlob.cs
+++ b/src/docfx/lib/FileGlob.cs
@@ -1,0 +1,101 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.Docs
+{
+    public class FileGlob
+    {
+        public static IEnumerable<string> GetFiles(string cwd, IEnumerable<string> patterns, IEnumerable<string> excludePatterns, GlobMatcherOptions options = GlobMatcher.DefaultOptions)
+        {
+            // If there is no pattern, nothing will be included
+            if (patterns == null) return Enumerable.Empty<string>();
+            if (string.IsNullOrEmpty(cwd)) cwd = Directory.GetCurrentDirectory();
+
+            IEnumerable<GlobMatcher> globList = patterns.Select(s => new GlobMatcher(s, options));
+            IEnumerable<GlobMatcher> excludeGlobList = Enumerable.Empty<GlobMatcher>();
+            if (excludePatterns != null)
+                excludeGlobList = excludePatterns.Select(s => new GlobMatcher(s, options));
+            return GetFilesCore(cwd, globList, excludeGlobList);
+        }
+
+        private static IEnumerable<string> GetFilesCore(string cwd, IEnumerable<GlobMatcher> globs, IEnumerable<GlobMatcher> excludeGlobs)
+        {
+            if (!Directory.Exists(cwd)) yield break;
+            foreach (var file in GetFilesFromSubfolder(cwd, cwd, globs, excludeGlobs))
+            {
+                yield return NormalizeToFullPath(file);
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


herohua:
@928PJY , can you highlight the difference from v2? It would help people review more efficiently. Thanks!

src/docfx/lib/FileGlob.cs
```diff
@@ -0,0 +1,101 @@
...
+            // 1. root folder, e.g. E:\ or /
+            // 2. sub folder, e.g. a or a/ or a\
+            return subpath.TrimStart('\\', '/');
+        }
+
+        private static string GetRelativeDirectoryPath(string parentDirectory, string directory)
+        {
+            var relativeDirectory = GetRelativeFilePath(parentDirectory, directory);
+            return relativeDirectory.TrimEnd('\\', '/') + "/";
+        }
```
OsmondJiang: why not just use Path.GetRelativePath()

src/docfx/lib/FileGlob.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        private static string NormalizeToFullPath(string path)
+        {
+            return Path.GetFullPath(path).Replace('\\', '/');
+        }
```
OsmondJiang: better to use existing path utility

src/docfx/lib/GlobMatcher.cs
```diff
@@ -0,0 +1,837 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Docs
+{
+    [Serializable]
```
OsmondJiang: >[Serializable] [](start = 4, length = 14)

what's that for?

src/docfx/lib/GlobMatcher.cs
```diff
@@ -0,0 +1,837 @@
...
+    public class GlobMatcher
+    {
+        #region Private fields
```
OsmondJiang: >#region [](start = 8, length = 7)

better to avoid region

src/docfx/lib/FileGlob.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        private static bool IsMatch(string path, IEnumerable<GlobMatcher> globs, IEnumerable<GlobMatcher> excludeGlobs, bool partial)
+        {
+            foreach (var exclude in excludeGlobs)
+            {
+                if (exclude.Match(path, false)) return false;
+            }
+            foreach (var glob in globs)
+            {
+                if (glob.Match(path, partial)) return true;
```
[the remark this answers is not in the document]
superyyrrzz: I think this is by design as `!` only applies to the current pattern, not the entire include group. The usage is a bit confusing. What about stop supporting negate `!` in glob pattern? `exclude` is enough. We also doesn't mention this usage in v2 document.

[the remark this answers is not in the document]
yufeih: Yes. negate + exclude adds confusion. Let's just drop `!`

src/docfx/lib/FileGlob.cs
928PJY: there is a bug in this function:
if I call `Path.GetRelativePath("test", "test/**.md")`, it returns `test/**.md` but not `**.md`

---
In reply to: [181018217](https://github.com/dotnet/docfx/pull/2627#discussion_r181018217) [](ancestors = 181018217)

src/docfx/lib/GlobMatcher.cs
```diff
@@ -0,0 +1,845 @@
...
+{
+    [Flags]
+    public enum GlobMatcherOptions
+    {
+        None = 0x0,
+        IgnoreCase = 0x1,
+        AllowNegate = 0x2,
+        AllowExpand = 0x4,
+        AllowEscape = 0x8,
+        AllowGlobStar = 0x10,
```
OsmondJiang: >0x10 [](start = 24, length = 4)

just a ignorable suggest, 1， 1<<1, 1<<2, 1<<3 would be better for reading

PR #2637, opened 2018-04-13
Enable extra emphasis syntax to keep consistent with GFM

qinezh:
Here's [the spec](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/Specs/EmphasisExtraSpecs.md) for extra emphasis.


vicancy:
Cool... and do we want a frontend support for default template...?

qinezh:
This PR is a bug fix actually, because DFM supports Strikethrough syntax as well as GitHub. I think default template has already supported it.

As for the syntax of Superscript/Subscript/Inserted/Marked, you're right, we need support it in default template before enable the syntax.

qinezh:
Update PR to support Strikethrough syntax only. And the other syntax will be supported until the template work is ready.

Issue #2648, opened 2018-04-17
Is Microsoft documentation docfx now CommonMark compliant?


ghost:
I am coming from this way:

https://github.com/jgm/pandoc/issues/4557 => https://github.com/dotnet/cli/pull/9081 => https://github.com/dotnet/docs

So CLI has a handwritten python script https://github.com/dotnet/cli/blob/20cd0f04b28c009838274b4ea741dfa132b07802/Documentation/manpages/tool/man-pandoc-filter.py, to remove docfx extensions from dotnet-cli tools docs, which is used during the pandoc processing / conversion to manpage.

I have been asked to encourage you guys to embrace CommonMark and treat the rest of additions as extensions to the original thing: https://github.com/jgm/pandoc/issues/4557#issuecomment-382099471. Is it already the case? :)

vicancy:
We do support CommonMark, but not set as the default markdown engine. `docfx.config` to set `markdownEngineName` as `markdig` to enable CommonMark syntax:
```json
{
  "build": {
    "markdownEngineName": "markdig",
...
```

herohua:
@kasper3 Have you tried the solution above to enable CommonMark in DocFX markdown engine?

kasper3:
@herohua, I proposed https://github.com/dotnet/docs/pull/4963, but it was pushed to later this year.
Tools like pandoc to convert one text format to another, which is markdown to Unix manpages in case of dotnet/cli, will have to then wait later this year.

xoofx:
@kasper3 docfx is switching to markdig which is CommonMark compliant and is going to be default soon (if it is not already the case). There is really no value of using Pandoc for converting markdown to HTML in docfx case (not C# based, impossible to develop specific C# extensions, Pandoc is not CommonMark compliant, does not support and won't likely never support DFM.. .etc.).

kasper3:
>  Pandoc is not CommonMark compliant

@xoofx, if you follow this thread closely, this issue is way tangential to what you have commented :)

Pandoc actually is CommonMark complaint. Pandoc maintainer has sent me here (first comment) to request docfx to be CommonMark complaint. I was then introduced to markdig (thanks a lot btw!) engine. My interest in this engine is to see dotnet/docs repo start using markdig and all existing `.md` files in that repo are converted to CommonMark: https://github.com/dotnet/docs/pull/4963.

Why?

[...]

Issue #2647, opened 2018-04-17
Official GFM link


kasper3:
https://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html states 

>  It supports all GitHub Flavored Markdown syntax with 2 exceptions when resolving list.

Where GitHub Flavored Markdown points to https://guides.github.com/features/mastering-markdown/

I think it should link to official spec https://github.github.com/gfm/, since GFM is rebased on CommandMark.

PR #2650, opened 2018-04-18
Enable markdig engine in documentation

qinezh:


vicancy:
Hi @qinezh Do you think it is ready for docfx to switch to "markdig" as the default engine? 
Related issues: https://github.com/dotnet/docfx/issues/2648 https://github.com/dotnet/docfx/issues/2647

qinezh:
@vicancy I'm afraid not at the moment. There're so many repositories generate documentation without setting the markdown engine explicitly in `docfx.json`. For these repositories, switch to "markdig" as default engine in DocFX would be a breaking change if the documentation aren't written in a CommonMark style.

kasper3:
Instead of coining yet another engine ..
Can `dfm-latest` be updated to base on CommonMark which new GFM is using, instead of old-GFM?
I thought there was no significant difference except for:

`###heading` - used to be allowed, but not anymore

vs.

`### heading` - with one space, works before and after

GitHub guys could also have created a new engine instead of updating GFM, but they decided to take a one-time breaking change. They have many times more documents to support than docfx does.

kasper3:
With the current approach, imo it would take huge amount of time to convince everyone to update to `markdig` engine, e.g. https://github.com/dotnet/docs/blob/e299873564d36e3349ad158d5a98a9c716c75d59/docfx.json#L3

qinezh:
@kasper3 The repo you mentioned will be migrated to use `markdig` before too long. And yes we have to take a one-time breaking change like GitHub, but the timing not right at this moment.

vicancy:
@qinezh Understand that. How about adding that to our DFM documentation, after all, we claimed we are compatible with GFM by default, but actually that is not the case now.

PR #2652, opened 2018-04-18
support deserializing readonly field in jsonUtility

928PJY:


src/docfx/lib/YamlUtility.cs
```diff
@@ -127,7 +122,14 @@ private static JToken ToJson(YamlNode node)
...
                 foreach (var (key, value) in map)
                 {
-                    obj[key.ToString()] = ToJson(value);
+                    if (key is YamlScalarNode scalarKey)
+                    {
+                        obj[scalarKey.Value] = ToJson(value);
+                    }
+                    else
+                    {
+                        throw new NotSupportedException($"Not Supported: {key.ToString()} is not a primitive type");
```
yufeih: .ToString() not needed: `{key}` is just {key.ToString()}

src/docfx/lib/JsonUtililty.cs
```diff
@@ -17,14 +18,16 @@ namespace Microsoft.Docs.Build
     /// </summary>
     public static class JsonUtililty
     {
+        public static readonly JsonSerializer DefaultNoneFormatSerializer = JsonSerializer.Create(s_noneFormatJsonSerializerSettings);
```
yufeih: s_xxx ?

PR #2655, opened 2018-04-18
And Homebrew package mention

ChrisChinchilla:
Added a mention of the Homebrew package for Macs as that's very useful.


ChrisChinchilla:
And I wasn't sure what branch to commit against…

superyyrrzz:
It should target dev branch. Never mind, I will handle it. Thanks for contribution!

Issue #2131, opened 2017-10-18
Code examples in XML documentation


cesarsouza:
Can DocFx handle ```<code source="someclass.cs" region="some_region" />``` tags in XML documentation?

In Sandcastle and a few other tools it is possible to specify that the contents of a ```<code></code>``` tag should be fetched from a ```#region``` from inside an arbitrary file in the project, e.g.:

```csharp
namespace SomeNamespace
{
    /// <summary>
    /// Some class
    /// </summary>
    /// <example>
    /// <code source="unit tests/class1test.cs" region="some_region" />
    /// </example>
    /// 
    public class SomeClass 
    {
    }
}
```

My question is, can DocFx handle this as well? 

I am trying to replace Sancastle by DocFx in a project, but got stuck trying to get this working.

superyyrrzz:
http://dotnet.github.io/docfx/spec/docfx_flavored_markdown.html#code-snippet

vicancy:
Yes, it is supported as described by https://github.com/dotnet/docfx/issues/910

cesarsouza:
Many thanks for the quick answers!

@superyyrrzz thanks, I had seen that page, but understood it applied only to markdown. There is no mention about how to accomplish the same as the ```[!code-<language>[<name>](<codepath><queryoption><queryoptionvalue> "<title>")]``` tag using triple slash XML comments.

@vicancy thanks, but I couldn't get it working yet. How do I tell DocFx where the root directory where the files with the ```#region``` tags is located? 

For example, let's say my repository has the following structure:

```markdown
- Root
  - Unit Tests
    - UnitTestsProject1
       - SomeClassTest.cs (this file contains the code snippets in #region tags) 
  - Sources
     - Project1
        - Namespace1
           - SomeClass.cs (this file contains XML documentation referring to SomeClassTests.cs)
     - Docs
        - DocFx
           - docfx.json
```

The SomeClass.cs file looks like this:
```csharp
namespace SomeNamespace
{
    /// <summary>
    /// Some class
    /// </summary>
    /// <example>
    /// <code source="Unit Tests/UnitTestsProject1/SomeClassTest.cs" region="some_region" />
    /// </example>
    /// 
    public class SomeClass 
    {
    }
}
```

[...]

PR #2663, opened 2018-04-20
allow setting the base path for code sources

vicancy:
https://github.com/dotnet/docfx/issues/2131


src/docfx/SubCommands/MetadataCommand.cs
```diff
@@ -157,6 +157,7 @@ private ExtractMetadataInputModel ConvertToInputModel(MetadataJsonItemConfig con
                 UseCompatibilityFileName = configModel?.UseCompatibilityFileName ?? false,
                 MSBuildProperties = configModel?.MSBuildProperties,
                 OutputFolder = outputFolder,
+                CodeSourceBasePath = configModel.CodeSourceBasePath
```
herohua: `configModel` could be null?

PR #2669, opened 2018-04-23
Fix tabbed content bug

vicancy:
that complex anchor element can also be handled


Documentation/spec/docfx_flavored_markdown.md
```diff
@@ -583,7 +583,7 @@ a: b
 ---
 ```
 
-In GFM, it would be rendered as `<hr>a: b<hr>`.
```
superyyrrzz: reserve space?

Documentation/spec/docfx_flavored_markdown.md
```diff
@@ -611,5 +611,5 @@ For example:
 >     This is code.
 ```
 
-In GFM, it will be rendered as a paragraph with content `[!NOTE] This is code.` in blockquote.
```
superyyrrzz: reserve space?

Issue #2686, opened 2018-04-25
Markdown Integration


yufeih:
Integrate markdig and docfx extensions in v3. 

The current markdig extension implementation has a strong dependency on docfx v2, need to first decouple it from v2:

- [x] Define `MarkdownContext` to be the abstraction of markdig extensions
- [x] Decouple Logger
- [x] Decouple CodeSnippet, Include from FileSystem and RelativePath
- [x] Decouple link resolve
- [x] Enable custom markdig pipeline by isolating `InclusionContext` from `MarkdownContext`
- [x] Decouple markdown validation
- [x] Decouple YamlSerialization and other helper functions.

To integrate markdig extensions in v3.

- [x] Integrate markdig and docfx extensions in v3
- ~~Add markdown build test cases in v3 (long-term)~~
- ~~Move markdig extensions to v3 branch (long-term)~~

/cc @qinezh @vwxyzh @herohua @OsmondJiang @vicancy

yufeih:
These behaviors needs clarification:

- `YamlHeaderRenderer` dumps raw yaml as text to `<yamlheader>` tag. Is this the expected in v3.
- Lots of errors and warnings are not written to Logger, but are written to html using a comment tag. Examples: include resolve failure, circular dependency. Is this expected in v3?

PR #2701, opened 2018-04-27
Decouple markdig extensions context from docfx.

yufeih:
Markdig extensions now does not depend on `IMarkdownEngine` and `MarkdownServiceParameters`. 

/cc #2686


superyyrrzz:
@qinezh Could you help review this?

PR #2708, opened 2018-05-02
Expand each section in yaml to be a separate test case.

yufeih:
Now you can select an individual YAML section in VS test explorer for debugging.

/cc @superyyrrzz @OsmondJiang


src/docfx/lib/YamlUtility.cs
```diff
@@ -24,21 +24,25 @@ public static class YamlUtility
         /// </summary>
         public static string ReadMime(TextReader reader)
         {
-            if (reader == null)
+            var content = ReadHeader(reader);
+            if (!content.StartsWith(YamlMimePrefix, StringComparison.OrdinalIgnoreCase))
             {
-                throw new ArgumentNullException(nameof(reader));
```
superyyrrzz: >hrow new ArgumentNullException(nameof(reader)); [](start = 17, length = 47)

Replace with Debug.Assert?

PR #2716, opened 2018-05-03
improve warning for no content

superyyrrzz:


src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -215,6 +224,14 @@ public void Build(IList<DocumentBuildParameters> parameters, string outputDirect
                         }
                     }
                 }
+                if (noContentFound)
+                {
+                    Logger.LogWarning($"No content file found. Please make sure the content section of docfx.json is correctly configured.");
```
herohua: can we add a LogCode so that OPS can decide whether to filter it or not?

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -215,6 +224,14 @@ public void Build(IList<DocumentBuildParameters> parameters, string outputDirect
...
+                }
+                else if (emptyContentGroups.Count > 0)
```
herohua: should this `if` to be embedded in above `if`?

superyyrrzz: Above `if` is for no content totally, this `if` is for content is available overall but some group has no content.

---
In reply to: [185978481](https://github.com/dotnet/docfx/pull/2716#discussion_r185978481) [](ancestors = 185978481)

src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -147,6 +147,9 @@ public void Build(IList<DocumentBuildParameters> parameters, string outputDirect
                         $"No file found, nothing will be generated. Please make sure docfx.json is correctly configured.",
                         code: WarningCodes.Build.EmptyInputFiles);
                 }
+
+                var noContentFound = true;
```
vicancy: naming: contentNotFound?

PR #2718, opened 2018-05-03
add logger and reporter

vicancy:


src/docfx/lib/log/IReporter.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Docs.Build
+{
+    internal interface IReporter
```
yufeih: removing ILog and adding it to Context?

src/docfx/lib/log/IReporter.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        void Error(string message, string code, string file = "", int line = 0, int column = 0);
```
yufeih: putting `code` before `message`? `message` tend to be very long.

PR #2728, opened 2018-05-04
fix yml parser bug of empty string and split test cases

OsmondJiang:
fix yml parser bug of empty string and split test cases


src/docfx/lib/YamlUtility.cs
```diff
@@ -94,6 +94,12 @@ public static JToken Deserialize(TextReader reader)
             var stream = new YamlStream();
 
             stream.Load(reader);
```
yufeih: Add a test case in YamlUtilityTest

PR #2732, opened 2018-05-07
Decouple markdown validation in Markdig

928PJY:
#2686


test/Microsoft.DocAsCode.MarkdigEngine.Tests/InclusionTest.cs
```diff
@@ -535,9 +536,9 @@ public void TestFallback_InclusionWithCodeFences()
 			{
 				BasePath = "."
 			};
-			var service = new MarkdigMarkdownService(parameter);
-			//var rootMarked = service.Markup(Path.Combine(Directory.GetCurrentDirectory(), $"{uniqueFolderName}/root_folder"), root, fallbackFolders, "root.md");
-			var rootMarked = service.Markup("place", "holder");
+			var service = new MarkdigMarkdownServiceCreator().CreateMarkdigMarkdownService(parameter);
```
yufeih: Is it possible to NOT change the existing interface?

OsmondJiang:
would you please add v3 tags so we can easily pick up PR only for V3?

src/docfx/SubCommands/DocumentBuilderWrapper.cs
```diff
@@ -156,7 +156,7 @@ private static IEnumerable<Assembly> LoadPluginAssemblies(string pluginDirectory
                 typeof(TocDocumentProcessor).Assembly,
                 typeof(SchemaDrivenDocumentProcessor).Assembly,
                 typeof(UniversalReferenceDocumentProcessor).Assembly,
-                typeof(MarkdigServiceProvider).Assembly
+                typeof(MarkdigServiceProvider).Assembly,
```
qinezh: >, [](start = 55, length = 1)

why add `,` here?

928PJY: following stylecop `SA1413UseTrailingCommasInMultiLineInitializers`

---
In reply to: [186641787](https://github.com/dotnet/docfx/pull/2732#discussion_r186641787) [](ancestors = 186641787)

PR #2733, opened 2018-05-07
add config for global/file metadata

superyyrrzz:


test/docfx.Test/config/ConfigTest.cs
```diff
@@ -0,0 +1,91 @@
...
+using Newtonsoft.Json.Linq;
+using System.IO;
+using Xunit;
+
+namespace Microsoft.Docs.Build
+{
+    public static class ConfigTest
+    {
+        [Fact]
+        public static void TestConigWithGlobalMetadata()
```
yufeih: Can these test be written as yaml specs? You can implement the basic metadata in build pipeline that merges file and global metadata and output them to a json file.

src/docfx/config/Config.cs
```diff
@@ -96,9 +102,63 @@ private static Config LoadCore(string configPath, CommandLineOptions options)
...
+
+        private static JArray ToGlobConfigs(JObject obj)
+        {
+            var result = new JArray();
+
+            foreach (var (key, value) in obj)
+            {
+                result.Add(new JObject
+                {
+                    [ConfigConstants.Include] = ToGlob(key),
```
yufeih: Can we avoid turning folders into glob patterns? It is a lot faster to check if a file inside a folder using starts with. Glob matchers are slower.

test/docfx.Test/config/ConfigTest.cs
superyyrrzz: want to keep here to avoid too large PR. the spec is also prone to change for now as they are not final output.

src/docfx/config/Config.cs
superyyrrzz: resolved

PR #2744, opened 2018-05-09
Move event hooks to markdig extension setup method

yufeih:
This allows extensions to be applied to `MarkdownPipelineBuilder` without calling `UseXXX`.


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/LineNumber/LineNumberExtension.cs
```diff
@@ -3,23 +3,34 @@
...
             {
-                AddSourceInfoInDataEntry(document, filePath);
+                AddSourceInfoInDataEntry(document, _context.GetFilePath(_context.File));
             };
         }
 
+        public void Setup(MarkdownPipeline pipeline, IMarkdownRenderer renderer)
+        {
+
+        }
```
OsmondJiang: what's that for?

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TabGroup/TabGroupExtension.cs
```diff
@@ -10,6 +10,21 @@ public class TabGroupExtension : IMarkdownExtension
...
         public void Setup(MarkdownPipelineBuilder pipeline)
         {
+            var tabGroupAggregator = new TabGroupAggregator();
+            var aggregateVisitor = new MarkdownDocumentAggregatorVisitor(tabGroupAggregator);
+
+            var tagGroupIdRewriter = new TabGroupIdRewriter();
+            var tagGroupIdVisitor = new MarkdownDocumentVisitor(tagGroupIdRewriter);
+
+            var activeAndVisibleRewriter = new ActiveAndVisibleRewriter();
+            var activeAndVisibleVisitor = new MarkdownDocumentVisitor(activeAndVisibleRewriter);
```
OsmondJiang: >var activeAndVisibleVisitor = new MarkdownDocumentVisitor(activeAndVisibleRewriter); [](start = 12, length = 84)

var activeAndVisibleVisitor = new MarkdownDocumentVisitor(new ActiveAndVisibleRewriter());

yufeih: This is a copy of old code. I do not intend to change existing styles.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/LineNumber/LineNumberExtension.cs
yufeih: in v2, filePath is a `RelativePath`, `RelativePath.ToString()` starts with '~/' if the path is from working directory. The expected source info should not contain '~/'

PR #2745, opened 2018-05-09
filter out referenced toc from output

OsmondJiang:
1. filter out referenced toc from output
2. optimize test case


src/docfx/build/Build.cs
```diff
@@ -96,6 +96,12 @@ private static bool ShouldBuildFile(Document itemToBuild, ConcurrentDictionary<D
                 return false;
             }
 
+            if (itemToBuild.ContentType == ContentType.TableOfContents &&
```
yufeih: Can this be done in BuildToc instead of here?

OsmondJiang: I think we should control the build item in one place

---
In reply to: [186925294](https://github.com/dotnet/docfx/pull/2745#discussion_r186925294) [](ancestors = 186925294)

src/docfx/build/toc/TableOfContentsMap.cs
```diff
@@ -11,16 +11,23 @@ namespace Microsoft.Docs.Build
...
     {
-        private readonly List<Document> _tocs;
+        private readonly HashSet<Document> _tocs;
 
         private readonly Dictionary<Document, HashSet<Document>> _documentToTocs;
 
         public TableOfContentsMap(List<Document> tocs, Dictionary<Document, HashSet<Document>> documentToTocs)
         {
-            _tocs = tocs ?? throw new ArgumentNullException(nameof(tocs));
+            _tocs = new HashSet<Document>(tocs ?? throw new ArgumentNullException(nameof(tocs)));
```
vwxyzh: debug.assert instead of throw?

vwxyzh: also assert `_toc.Count == tocs.Count`

OsmondJiang: sorry, missing your comment, will change it later

PR #2746, opened 2018-05-09
Remove InclusionContext for other uses.

yufeih:
`InclusionContext` will be used to store per document information like *file path* as a thread static class


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Inclusion/InclusionBlock/InclusionBlock.cs
```diff
@@ -8,7 +8,11 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
 
     public class InclusionBlock : LeafBlock
     {
-        public InclusionContext Context { get; set; }
+        public string Title { get; set; }
+
+        public string IncludedFilePath { get; set; }
+
+        public string GetRawToken() => $"[!include[{Title}]({IncludedFilePath})]";
```
vwxyzh: What about escape charactor?
e.g.:
source markdown:
```md
[!include[\*Title\*](IncludedFilePath)]
```
What the title will be?

yufeih: The code is copied as is. It's current usage is only for error cases, when an include failed to resolve, this raw token is written to output.

Judging from the implementation, it'll be `[include[*Title*](IncludeFilePath)]`.

PR #2749, opened 2018-05-09
implement global/file metadata

superyyrrzz:


docs/specs/build/config/basics.yml
```diff
@@ -71,4 +71,94 @@ inputs:
...
+        key: value
+      docs/index.md:
+        key: another value
+  docs/a.md:
+  docs/b.md:
+  docs/index.md:
+outputs:
+  docs/a.json: |
+    { "content": "" }
+  docs/a.mta.json: |
```
yufeih: if we are producing current ops `*.page.json` format, can `a.mta.json` be merged into `a.json`:

{ "content": "", "metadata": { key: value } }

src/docfx/config/GlobConfig.cs
```diff
@@ -41,5 +42,35 @@ public GlobConfig(string[] include, string[] exclude, T value, bool isGlob = tru
...
+                return false;
+
+            if (IsGlob)
+            {
+                var glob = new GlobMatcher(pattern);
+                return glob.Match(filePath);
+            }
+            else
+            {
+                return pattern.EndsWith('/') ?
```
yufeih: do we allow users to specify directories without trailing `/`?

superyyrrzz: As filename is also supported here, `/` is needed to distinguish.

yufeih: can we look ask the file system if it is a directory or a file?

superyyrrzz: Still not want to do this. It will be a bit confused when we support features like FAL. Another option is to drop support of file, which needs more investigation of existing usage...

PR #2753, opened 2018-05-10
decouple logger in markdigEngine.extensions

928PJY:


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MarkdownContext.cs
```diff
@@ -41,14 +46,36 @@ public class MarkdownContext
...
+        /// <summary>
+        /// Set logger scope function
+        /// </summary>
+        public Func<string, IDisposable> SetLoggerScope { get; }
+
         public MarkdownContext(
             IReadOnlyDictionary<string, string> tokens,
+            LogActionDelegate logWarning,
+            LogActionDelegate logError,
+            Func<string, IDisposable> setLoggerScope,
```
yufeih: can we remove setLoggerScope?

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MarkdownContext.cs
```diff
@@ -10,6 +10,11 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
 
     public class MarkdownContext
     {
+        /// <summary>
+        /// Log delegate
+        /// </summary>
+        public delegate void LogActionDelegate(string message, string phase = null, string file = null, string line = null, string code = null);
```
yufeih: To align with v3 logging, this could be `(string code, string message, string file = null, int line)`. `code` is mandatory.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MarkdownContext.cs
928PJY: why? scope will make log easier to locate

---
In reply to: [187253456](https://github.com/dotnet/docfx/pull/2753#discussion_r187253456) [](ancestors = 187253456)

PR #2758, opened 2018-05-10
Implement report

yufeih:
- Add `docs/errors-and-warnings` to track our errors and troubleshooting guides
- Update test framework to support examine build reports
- Implement basic error handling, close several TODOs related to error handling


src/docfx/ErrorCodes.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Docs.Build
+{
+    internal static class ErrorCodes
+    {
+        public const string ConfigNotFound = "config-not-found";
```
OsmondJiang: >ConfigNotFound [](start = 28, length = 14)

should the error code should have corresponding action or message formate? otherwise the error message is really hard to maintain and can be anything...

src/docfx/lib/log/Reporter.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        private readonly object _lock = new object();
+        private Lazy<TextWriter> _output;
+        private bool _stable;
+
+        public void OutputToFile(string outputFilePath, bool stable)
+        {
+            Debug.Assert(_output == null, "Cannot change report output path");
+
+            Directory.CreateDirectory(Path.GetDirectoryName(outputFilePath));
```
vicancy: directory can be empty

src/docfx/lib/log/Reporter.cs
```diff
@@ -0,0 +1,96 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    internal sealed class Reporter : IDisposable
+    {
+        private readonly object _lock = new object();
+        private Lazy<TextWriter> _output;
+        private bool _stable;
+
+        public void OutputToFile(string outputFilePath, bool stable)
```
vicancy: not thread safe?

OsmondJiang: agree

---
In reply to: [187504049](https://github.com/dotnet/docfx/pull/2758#discussion_r187504049) [](ancestors = 187504049)

OsmondJiang: can it be a constructor?

---
In reply to: [187504180](https://github.com/dotnet/docfx/pull/2758#discussion_r187504180) [](ancestors = 187504180,187504049)

docs/errors-and-warnings.md
```diff
@@ -0,0 +1,35 @@
...
+
+> ❌ This is an error  
+> ⚠️ This is a warning  
+> 💡 Guide to help you troubleshoot  
+
+## General Errors
+
+### circular-reference
+
+❌ A series of document references creates a closed loop.
```
superyyrrzz: Does this mean reference including tokens? This can be warning, as it like a document error with limited influence.

src/docfx/ErrorCodes.cs
superyyrrzz: >config-not-found [](start = 46, length = 16)

use camel case as in v2? to make it easy to integrate

[https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Common/Loggers/WarningCodes.cs](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Common/Loggers/WarningCodes.cs)

src/docfx/cli/Program.cs
```diff
@@ -50,5 +85,48 @@ private static (string command, string docset, CommandLineOptions options) Parse
...
+            // windows command line does not have good emoji support
+            var showEmoji = !RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
+            if (showEmoji)
+                Console.Write("🚘💥🚗 ");
+            Console.Write("docfx has crashed");
+            if (showEmoji)
+                Console.Write(" 🚘💥🚗");
+
+            Console.WriteLine();
+            Console.WriteLine("Help us improve by creating an an issue at https://github.com/dotnet/docfx with the following content:");
```
superyyrrzz: >an  [](start = 62, length = 3)

remove

src/docfx/ErrorCodes.cs
yufeih: v2 is Pascal. The idea is that we can have a webpage like: [https://docfx.github.io/errors-and-warnings#config-not-exist](https://docfx.github.io/errors-and-warnings#config-not-exist), having a *slug* like format is more url friendly. but open for any options.

---
In reply to: [187515915](https://github.com/dotnet/docfx/pull/2758#discussion_r187515915) [](ancestors = 187515915)

src/docfx/lib/log/Reporter.cs
yufeih: This method is intended to be only be called once on a single instance, thus thread safe, see `Debug.Assert`.

It cannot be put into constructor because, before config is loaded, we don't know the output location so errors are written to console only, after that, errors will be logged to both console and file.

---
In reply to: [187504215](https://github.com/dotnet/docfx/pull/2758#discussion_r187504215) [](ancestors = 187504215,187504180,187504049)

yufeih: We could buffer logs before an output is specified and write them all to file once we have the output, provided we don't hold too much logs in memory.

---
In reply to: [187523738](https://github.com/dotnet/docfx/pull/2758#discussion_r187523738) [](ancestors = 187523738,187504215,187504180,187504049)

src/docfx/config/Config.cs
```diff
@@ -90,12 +88,16 @@ private static Config LoadCore(string configPath, CommandLineOptions options)
             try
             {
                 var configObject = Expand(YamlUtility.Deserialize<JObject>(File.ReadAllText(configPath)) ?? new JObject());
+                configObject.Merge(options.ToJObject(), JsonUtility.DefaultMergeSettings);
```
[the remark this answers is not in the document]
yufeih: yeah, but that one produces a new object rather than merging into an existing object.

---
In reply to: [187517363](https://github.com/dotnet/docfx/pull/2758#discussion_r187517363) [](ancestors = 187517363)

src/docfx/ErrorCodes.cs
yufeih: I am more towards [eslint](https://eslint.org/docs/user-guide/configuring) style config. config itself is camel case, the **rules** are *slugs*

PR #2762, opened 2018-05-14
Prevent docfx build when building for Live Unit Testing

paulomorgado:
docfx.console.targets builds documentation even on Live Unit Testing builds.

This PR prevents that from happening.


superyyrrzz:
implemnted in #2763

PR #2763, opened 2018-05-14
Conditional build

paulomorgado:
Build only if `BuildDocFx` is `true` with a default value of `true`. `BuildDocFx` is always false for building for live unit testing (`'$(BuildingForLiveUnitTesting)' == 'true'`).

### DIFF
--- a/src/nuspec/docfx.console/build/docfx.console.targets
+++ b/src/nuspec/docfx.console/build/docfx.console.targets
@@ -13,28 +13,28 @@
     <CopyToOutput>False</CopyToOutput>
     <RebuildDoc Condition=" '$(RebuildDoc)' == '' ">False</RebuildDoc>
   </PropertyGroup>
-  <Target Name="DocRebuild">
+  <Target Name="DocRebuild" Condition="'$(BuildingForLiveUnitTesting)' != 'true'">
     <CallTarget Targets="DocClean"/>
     <CallTarget Targets="DocBuild"/>
   </Target>
-  <Target Name="DocBuild" AfterTargets="Build">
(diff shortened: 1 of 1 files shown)


superyyrrzz:
Looks like this PR is more general. I guess #2762 is not needed if this one is merged?

src/nuspec/docfx.console/build/docfx.console.targets
```diff
@@ -8,41 +8,43 @@
     <MetadataOutputFolder Condition=" '$(MetadataOutputFolder)' == '' ">$(OutputFolder)</MetadataOutputFolder>
     <LogFile Condition=" '$(LogFile)' == '' ">log.txt</LogFile>
     <LogLevel Condition=" '$(LogLevel)' == '' ">Verbose</LogLevel>
+    <BuildDocFx Condition=" '$(LogLevel)' == '' ">true</BuildDocFx>
```
superyyrrzz: `BuildDocFx` => `BuildDocfx`, `LogLevel` => `BuildDocfx`. I think you can swap this line with below, so that it will always respect the value intentionally set.

paulomorgado:
Yes, this one builds up on #2762 

OOPS! Copy+paste is evil!


I also think that the property names used are very dangerously generic. Would you consider making these changes?


* BuildDocToolPath -> DocFxToolPath
* PreviewPort -> DocFxPreviewPort
* IsServing -> DocFxIsServing
* MetadataOutputFolder -> DocFxMetadataOutputFolder
* LogFile -> DocFxLogFile
* LogLevel -> DocFxLogLevel
* CopyToOutput -> DocFxCopyToOutput
* RebuildDoc -> DocFxRebuildDoc

superyyrrzz:
Thanks for contribution!

PR #2766, opened 2018-05-15
fix test bug

OsmondJiang:


OsmondJiang:
there would be concurrency issue to create all test docsets in one folder, split them to different folder

PR #2768, opened 2018-05-15
Add code to markdown logs

yufeih:
These codes serves as a placeholder, we will review all error and warning codes in v3 later

Depend on PR: #2767


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Row/RowParser.cs
```diff
@@ -92,7 +92,7 @@ public override BlockState TryContinue(BlockProcessor processor, Block block)
 
             if (!c.IsZero())
             {
-                _context.LogWarning($"Row has some invalid chars in the ending.");
+                _context.LogWarning("invalid-row", $"Row has some invalid chars in the ending.");
```
OsmondJiang: >invalid-row [](start = 37, length = 11)

should we make it as a const string?

yufeih: yes, but I'd like to do it later. The current focus is to get the public interface ready and stable

PR #2773, opened 2018-05-16
support resolve from dependent docset

OsmondJiang:
support to resolve href from dependent docset
todo: add test if the markdig has been integrated into the pipeline

### DIFF
--- a/src/docfx/docset/Document.cs
+++ b/src/docfx/docset/Document.cs
@@ -96,8 +96,7 @@ namespace Microsoft.Docs.Build
         public override int GetHashCode()
         {
             // todo: add docset for calculation
-            // todo: case senstive or not?
-            return StringComparer.OrdinalIgnoreCase.GetHashCode(FilePath);
+            return StringComparer.Ordinal.GetHashCode(FilePath);
         }
 
         public bool Equals(Document other)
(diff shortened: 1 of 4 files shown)


test/docfx.Test/build/TocTest.cs
```diff
@@ -37,11 +37,11 @@ public static class TocTest
         public static void FindTocRelativePath(string[] tocFiles, string file, string expectedTocPath)
```
yufeih: shall we use spec test so we don't need to change test case when refactor source code 😆

OsmondJiang: ><, spec test is not always suitable, especially for very special purpose.

---
In reply to: [188491520](https://github.com/dotnet/docfx/pull/2773#discussion_r188491520) [](ancestors = 188491520)

src/docfx/docset/Document.cs
```diff
@@ -218,14 +217,36 @@ internal static Document TryResolveFromPathToDocset(Docset docset, string path)
...
             {
                 return new Document(docset, path);
             }
 
             // todo: localization fallback logic
             // todo: redirection files
-            // todo: resolve from dependencies
+
+            // resolve from dependent docsets
+            foreach (var (dependencyName, url) in docset.Config.Dependencies)
```
yufeih: ordered

PR #2777, opened 2018-05-16
support extend in config

superyyrrzz:


docs/designs/config.md
```diff
@@ -59,7 +59,7 @@
 
 ``` yml
 name: Azure.azure-documents
-default: msdocs.yml
+extend: msdocs.yml
```
yufeih: where do you think we put 'msdocs.yml'?

superyyrrzz: It can be in a separate repo, restored in current config.

superyyrrzz:
the behavior of `.eslintrc.js`'s `extends`:
1. current config takes precedence over `extends` config
2. latter config takes precedence over former config

will change accordingly

src/docfx/config/Config.cs
```diff
@@ -107,6 +102,52 @@ private static Config LoadCore(string configPath, CommandLineOptions options = n
...
         }
 
+        private static JObject LoadOriginalConfigObject(string configPath, List<string> parents)
+        {
+            var result = YamlUtility.Deserialize<JObject>(File.ReadAllText(configPath)) ?? new JObject();
+            if (!result.TryGetValue(ConfigConstants.Extend, out var objExtend))
+                return result;
+
+            if (parents.Contains(configPath))
+                throw Errors.CircularReference(configPath, parents);
```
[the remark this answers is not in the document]
superyyrrzz: Agree. Removed it.

docs/designs/config.md
```diff
@@ -118,7 +118,10 @@ Configs is grouped by scenario:
 * `contribution`: to config whether to enable contribution, and where to config.
 * `contrtibutors`: to config which contributors to include or exclude.
 * `gitDependencies`: to config CRRs.
-Extension config is supported to share default config for all OPS repos, by `defaults: msdocs.yml`. Current config has higher priority than extension config when conflicts.
+Extension config is supported to share default config for all OPS repos, by `extend: msdocs.yml`. Current config has higher priority than extension config when conflicts. Extension config also supports array, whose latter item has higher priority.
+> [!NOTE]
+>
+> `dependencies` doesn't support `extend`, as we may import extending configs from dependent repos.
```
yufeih: having dependency repos in a shared location is a valid scenario, we may want to put localization config for all repos in a central repo,

superyyrrzz: Yes, this should be supported. Changed the wording to make it clear...

PR #2778, opened 2018-05-16
Integrate markdig and extensions

yufeih:


src/docfx/lib/MarkdownUtility.cs
```diff
@@ -0,0 +1,181 @@
...
+    /// <summary>
+    /// Converts markdown to html
+    /// </summary>
+    internal static class MarkdownUtility
+    {
+        // In docfx 2, a localized text is prepended to quotes beginning with
+        // [!NOTE], [!TIP], [!WARNING], [!IMPORTANT], [!CAUTION].
+        //
+        // Docfx 2 reads localized tokens from template repo. In docfx3, build (excluding static page generation)
+        // does not depend on template, thus these tokens are managed by us.
```
superyyrrzz: Can we put loc token into some dependency or resoruce for better extensibility?

src/docfx/build/markdown/BuildMarkdown.cs
```diff
@@ -11,20 +10,19 @@ internal static class BuildMarkdown
         public static Task Build(Context context, Document file, TableOfContentsMap tocMap)
         {
             var markdown = file.ReadText();
-            var model = new PageModel<string> { Content = Markup(markdown) };
+            var (html, yamlHeader) = MarkdownUtility.Markup(markdown, file, context);
+            var content = !string.IsNullOrEmpty(html) ? $"<div>{html.Trim()}</div>" : "";
```
superyyrrzz: > "" [](start = 85, length = 3)

We should unity to "" or string.Empty later...

src/docfx/lib/MarkdownUtility.cs
yufeih: I'd like to push these to the frontend. We should not care what the **visual** looks like. Build just add `class="warning"`, frontend can determine the localized visual, either a text or icons using css/js

src/docfx/build/markdown/BuildMarkdown.cs
```diff
@@ -11,20 +10,19 @@ internal static class BuildMarkdown
...
+            var model = new PageModel<string> { Content = content };
 
             context.WriteJson(model, file.OutputPath);
 
             var metadata = Metadata.FetchFromConfig(file);
```
OsmondJiang: >FetchFromConfig [](start = 36, length = 15)

would we cache that? otherwise it will fetch from config for every file

yufeih: *fetch* is misleading here.

OsmondJiang: we can cache the merge metadata from config via file metadata key

---
In reply to: [188828823](https://github.com/dotnet/docfx/pull/2778#discussion_r188828823) [](ancestors = 188828823)

yufeih: Agreed, @superyyrrzz

PR #2784, opened 2018-05-18
output sccm toc backward compatility model

OsmondJiang:


src/docfx/cli/Program.cs
```diff
@@ -83,12 +80,10 @@ private static (string command, string docset, CommandLineOptions options) Parse
                 syntax.DefineOption("o|out", ref options.Output, "output folder");
                 syntax.DefineOption("l|log", ref options.Log, "path to log file");
                 syntax.DefineOption("stable", ref options.Stable, "produces stable output for comparison in a diff tool");
+                syntax.DefineOption("legacy", ref options.OutputLegacyModel, "output legacy model for backward compatility");
                 syntax.DefineParameter("docset", ref docset, "docset path that contains docfx.yml");
             });
 
-            // Don't show it in help
-            options.OutputLegacyModel = args.Contains("--legacy");
```
superyyrrzz: why move this above?

OsmondJiang: otherwise the command parser will fail when meet unknow syntax

---
In reply to: [189169291](https://github.com/dotnet/docfx/pull/2784#discussion_r189169291) [](ancestors = 189169291)

Issue #2765, opened 2018-05-15
Design build output format for dynamic rendering


yufeih:
This is to track the desired output file formats for dynamic rendering in v3.

- Can we change the filename, directory layout of output files
- Do we still need `.raw.json`, `.page.json` and `html` as 3 separate files
- Need to lookup build output from url for local rendering, what is the output path when locale, moniker range, branch is involved (aka, where to put these context info)

fenxu:
We have a [documentation](https://github.com/MicrosoftDocs/openpublishing-docs/blob/master/openpublishing/docs/engdocs/open-publish-build-tools-output.md) describe the output things. @yufeih

superyyrrzz:
> Need to lookup build output from url for local rendering, what is the output path when local, moniker range, branch is involved (aka, where to put these context info)

This [spec](https://github.com/dotnet/docfx/blob/v3/docs/designs/versioning.md#532-local-implementation) covers this. Details maybe be further discussed.

yufeih:
We could have a dedicated output mode for compatibility with ops, preferably through an *undocumented* command line switch `--ops`

Running `docfx build --ops` produces the exact same output as ops
Running `docfx build` produces an ideal format that:
  - contains all the information to generate ops output
  - is the format our spec tests expects
  - ideally maps directly to what is stored in hosting

We can have a dedicated procedure to convert *ideal format* to *ops format* after `docfx build` when `--ops` is supplied.

PR #2785, opened 2018-05-18
Add build output spec

yufeih:
#2765 

@vwxyzh @vicancy @fenxu @OsmondJiang @superyyrrzz @herohua

### DIFF
--- a/docs/designs/build-output.md
+++ b/docs/designs/build-output.md
@@ -1,0 +1,127 @@
+# Build Output
+
+This document specifies docfx output file layout. It is designed to satisfy these requirements:
+
+- **Backward compatible with legacy system**: the output files should contain all the information needed to transform to whatever the *legacy system* desires. With an *undocumented* command line switch `--legacy`, `docfx` converts the output files to the legacy format.
+
+- **Dynamic build output maps directly to hosting server**: the build output for dynamic rendering should map directly to what the hosting layer expects, without permutation.
+  > *Hosting server* here is a fictional server using `documentdb`
+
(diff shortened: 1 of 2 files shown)


docs/designs/build-output.md
```diff
@@ -0,0 +1,127 @@
...
+- **Static build output xcopy deployable**: should just xcopy static build output to any static hosting server.
+
+
+## URL Schema
+
+An **URL** is an universal identifier that confirms to [URL Standard](https://url.spec.whatwg.org/).
+
+For *dynamic rendering*, `docfx` produces an output for lookup by the following URL schema:
+
+`https://{host}/{locale?}/{site-url}/?view={moniker?}`
```
OsmondJiang: why static and dynamic outputs are different?

yufeih: cannot have query string for static rendering. dynamic render has to follow docs.

docs/designs/build-output.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+> `?` means optional.
+> When a docset is not localized, we will not create the `{locale}` folder.
+> This allows static websites with minimal url when localized and verioning is not needed.
+
+## File Outputs
+
+Each input file transforms to zero or more output files depending on its content type. `docfx` places output files in `output-path` relative to output directory. Regardless of static rendering or dynamic rendering, output path shares the same schema:
+
```
OsmondJiang: this is important, mark it strong

docs/designs/build-output.md
```diff
@@ -0,0 +1,127 @@
...
+`{output-dir}/{locale?}/{moniker?}/{site-path}`
+
+```
+      locale    moniker                 site-path
+      |-^-| |------^------| |----------------^----------------|
+_site/en-us/netstandard-2.0/dotnet/api/system.string/index.html
```
superyyrrzz: >/index.html [](start = 52, length = 11)

make this as an option? keep the default output style as in v2

Issue #2782, opened 2018-05-17
Build `edge-developer` with expected ops build outputs


yufeih:
Track remaining works based on PR1159 on impact repo (will no list changes to be accepted here).
- [x] [**ACCEPT CHANGE**] ready
- [x] [**compare tool**] remove null value metadata from `.mta.json` in impact tool
- [x] [**v2 template**] ~~improve logic to generate `_op_displayDate` in v2 template~~ accept this first as we don't want to depend on template deployment circle
- [x] [**legacy metadata**] redirect file (including explicitly add `redirect_url` in YAML header) doesn't need some metadata. remove in LegacyMetadata.

Track the work items needed to produce the same build outputs for `edge-developer`

 - [x] HTML (@yufeih)
 - [x] missing metadata (@928PJY) #2813
 - [x] apply Jint template (@928PJY) #2813 
 - [x] git metadata + user profile log file (@superyyrrzz) #2813
 - [x] output files (@OsmondJiang), see below
 - [x] log file: .dependency-map.json (@OsmondJiang)
 - [x] redirection #2822 (@OsmondJiang)
 - [x] document_id #2821 (@OsmondJiang)
 - [x] build report (@yufeih)
 - [x] metadata coming from config (@928PJY)
 - [x] raw title (@yufeih)
 - [x] asset id (@928PJY)
 - [x] remove redirect_document_id (@OsmondJiang)
 - [x] _op_allContributorsStr (@928PJY)
 - [x] title, titleSuffix, _op_ogTitle (#2992)

OsmondJiang:
#2784

OsmondJiang:
compatible output from ops build package: 
  - op_aggregated_file_map_info.json, used for localization
  - filemap.json, contains file mappings
  - .manifest.json
  - xrefmap.yml, xref service will use it
  - op_crr_info.json, preview service will use it

OsmondJiang:
#2787

yufeih:
`sccmdocs` is not using markdig. We need to switch to a markdig repo.

yufeih:
Let's use https://github.com/MicrosoftDocs/edge-developer as our initial target

PR #2786, opened 2018-05-18
Adjust output according to spec

yufeih:
#2765 #2782

### DIFF
--- a/docs/specs/build/toc/corners.yml
+++ b/docs/specs/build/toc/corners.yml
@@ -11,8 +11,7 @@ inputs:
     # [Topic2](https://github.com/docfx) #
   docs/index.md:
 outputs:
-  docs/index.json: |
-    { "content": "" }
+  docs/index.json:
   docs/toc.json: |
     {  
       "items":[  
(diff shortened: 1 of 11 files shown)


docs/specs/build/toc/corners.yml
```diff
@@ -69,14 +68,7 @@ inputs:
...
   build.log: |
-    [
-      {
-        "message": "Found circular reference: docs/toc.md --> docs/f1/toc.md --> docs/f2/toc.md --> docs/toc.md",
-        "message_severity": "Error",
-        "type": "circular-reference",
-        "file": "docs/toc.md"
-      }
-    ]
+    ["error","circular-reference","Found circular reference: docs/toc.md --> docs/f1/toc.md --> docs/f2/toc.md --> docs/toc.md","docs/toc.md"]
```
OsmondJiang: what's that format?

yufeih: JSON

OsmondJiang: we need this log to generate report, so modeled build.log would be better

PR #2795, opened 2018-05-21
Add Document.SitePath

yufeih:
`Document.OutputPath` should only be used to indicate where the output file is placed. Use `SitePath` to calculate relative path.

`OutputPath` may contain locale, moniker and other properties.

### DIFF
--- a/docs/specs/build/config/basics.yml
+++ b/docs/specs/build/config/basics.yml
@@ -147,7 +147,7 @@ inputs:
   docs/a.md:
   docs/b/b.md:
 outputs:
-  output/a.json:
+  output.json:
   output/b.json:
 ---
 # latter routes rule takes precedence - 1
@@ -158,7 +158,7 @@ inputs:
       docs/a/b.md: outputB
   docs/a/b.md:
 outputs:
-  outputB/b.json:
+  outputB.json:
 ---
 # latter routes rule takes precedence - 2
 inputs:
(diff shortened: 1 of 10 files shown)


docs/specs/build/config/basics.yml
```diff
@@ -147,7 +147,7 @@ inputs:
   docs/a.md:
   docs/b/b.md:
 outputs:
-  output/a.json:
+  output.json:
```
superyyrrzz: tend to make the value only contains only output folder. URL is confusing when considering pretty URL or extension.

docs/specs/build/config/basics.yml
```diff
@@ -158,7 +158,7 @@ inputs:
       docs/a/b.md: outputB
   docs/a/b.md:
 outputs:
-  outputB/b.json:
+  outputB.json:
```
OsmondJiang: >outputB.json: [](start = 2, length = 13)

seems strange?

src/docfx/docset/Document.cs
```diff
@@ -154,22 +159,9 @@ internal static ContentType GetContentType(string path, string docsetPath)
...
-                default:
-                    return path;
-            }
-        }
-
-        internal static string GetSiteUrl(string path, ContentType contentType)
+        internal static string GetSiteUrl(string path, ContentType contentType, Config config)
         {
-            path = '/' + path;
+            path = '/' + ApplyRoutes(path, config.Routes);
```
OsmondJiang: >path = '/' + ApplyRoutes(path, config.Routes); [](start = 12, length = 46)

there would be a problem when calculating the relative site url in Resolve

yufeih: we'll handle it in url resolve.

PR #2798, opened 2018-05-22
Update output.md to include ugly url

yufeih:


docs/designs/output.md
```diff
@@ -26,17 +26,27 @@ For *dynamic rendering*, `docfx` produces an output for lookup by the following
...
+#### Static rendering with pretty URL
+
 ```
              host          locale    moniker            site-url
         |------^---------| |-^-| |------^------||----------^-------------|
 https://docs.microsoft.com/en-us/netstandard-2.0/dotnet/api/system.string/#Instantiation
 ```
 
-> This document uses pretty url format that ends with `/`. We also supports ugly url format that ends with `.html`.
+#### Static rendering with ugly URL
```
superyyrrzz: Rename `ugly URL` to `normal URL`? Just don't want to name the default behavior "ugly"...

yufeih: It's what hogo uses.

superyyrrzz: got it.

PR #2804, opened 2018-05-22
fix .targets file

sxul:
property name missing `()`


superyyrrzz:
@sxul Thank you for the contribution!

PR #2806, opened 2018-05-22
Jeahyoun/21139 context zones

jeahyoun:
This PR adds support for the Zones plugin, which allows content authors to use delimiters to show or hide content based on the target zone (e.g. chromeless-only content, PDf-only content).

See https://ceapex.visualstudio.com/Engineering/_workitems/edit/21139.

### DIFF
--- a/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MarkdownExtensions.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MarkdownExtensions.cs
@@ -14,26 +14,26 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
     {
         public static MarkdownPipelineBuilder UseDocfxExtensions(this MarkdownPipelineBuilder pipeline, MarkdownContext context)
         {
-			return pipeline
-				//.UseMathematics()
-				.UseEmphasisExtras(EmphasisExtraOptions.Strikethrough)
-				.UseAutoIdentifiers(AutoIdentifierOptions.GitHub)
-				.UseMediaLinks()
-				.UsePipeTables()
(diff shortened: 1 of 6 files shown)


yufeih:
Please format and align the code.

superyyrrzz:
Please use 4 spaces instead of tab, sign cla or join ms oss.

A question: the syntax seems like moniker range. Can `UseMonikerRange` and `UseRenderZone` be merged?

yufeih:
@superyyrrzz I believe they are for different purposes. The split makes sense.

herohua:
@qinen is added to the review. #Closed

herohua:
but is it possible to consolidate the code? I guess some of them can be shared.

---
In reply to: [391221062](https://github.com/dotnet/docfx/pull/2806#issuecomment-391221062) [](ancestors = 391221062)

jeahyoun:
@yufeih @superyyrrzz I've updated the formatting for this PR.

@herohua The zone plugin is based on the functionality of the moniker plugin, but they are slightly different. It may be possible to pull out shared functionality into a utility class or a mixin but I don't know how this Markdown extension will change in the future. The bulk of the logic that could be pulled out is in the parser. It may be possible to make a generic handler for any construct like the following:

```
::: keyword attribute1="value" attribute2="value, value2"

::: keyword-end
```

adkinn:
@yufeih @superyyrrzz thanks for merging this PR. When will this be deployed to PPE and Prod?  

/cc @jeahyoun

superyyrrzz:
@adkinn @jeahyoun  It is already on PPE. 6/11 on prod.

adkinn:
@superyyrrzz thanks for the info!

Issue #2805, opened 2018-05-22
docfx download throws exception


bitbonk:
**DocFX Version Used**: 
docfx 2.35.4.0

**Steps to Reproduce**:

1. run `docfx download -x https://docs.microsoft.com/dotnet/xrefmap.yml`

**Expected Behavior**:
The xrefmap is downloaded so I can build my API documentation even when the internet is broken

**Actual Behavior**:
```
[18-05-22 09:39:30.415]Error:System.AggregateException: One or more errors occurred. ---> System.ArgumentException: Phase name cannot be null or white space.
Parameter name: fileName
   at Microsoft.DocAsCode.Common.LoggerFileScope..ctor(String fileName)
   at Microsoft.DocAsCode.Build.Engine.XRefArchiveBuilder.<DownloadAsync>d__3.MoveNext()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at Microsoft.DocAsCode.SubCommands.DownloadCommand.Exec(SubCommandRunningContext context)
   at Microsoft.DocAsCode.Program.ExecSubCommand(String[] args)
---> (Inner Exception #0) System.ArgumentException: Phase name cannot be null or white space.
Parameter name: fileName
   at Microsoft.DocAsCode.Common.LoggerFileScope..ctor(String fileName)
   at Microsoft.DocAsCode.Build.Engine.XRefArchiveBuilder.<DownloadAsync>d__3.MoveNext()<---

[18-05-22 09:39:30.424]Info:Completed in 46,1122 milliseconds


Build failed.
[18-05-22 09:39:30.431]Error:System.AggregateException: One or more errors occurred. ---> System.ArgumentException: Phase name cannot be null or white space.
Parameter name: fileName
   at Microsoft.DocAsCode.Common.LoggerFileScope..ctor(String fileName)
   at Microsoft.DocAsCode.Build.Engine.XRefArchiveBuilder.<DownloadAsync>d__3.MoveNext()
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
[...]

PR #2815, opened 2018-05-28
Ensure output file provided.

vwxyzh:
#2805

### DIFF
--- a/src/docfx/SubCommands/DownloadCommand.cs
+++ b/src/docfx/SubCommands/DownloadCommand.cs
@@ -25,6 +25,11 @@ namespace Microsoft.DocAsCode.SubCommands
 
         public void Exec(SubCommandRunningContext context)
         {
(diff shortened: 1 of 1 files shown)


src/docfx/SubCommands/DownloadCommand.cs
```diff
@@ -25,6 +25,11 @@ public DownloadCommand(DownloadCommandOptions options)
 
         public void Exec(SubCommandRunningContext context)
         {
+            if (string.IsNullOrWhiteSpace(_options.ArchiveFile))
+            {
+                Logger.LogError("Please provide output file.");
```
vicancy: Change options.ArchiveFile to be a **Required** parameter in the CommandLine settings?

vwxyzh: There is no **Required** in `ValueOption`

Issue #2816, opened 2018-05-28
Define and generate dependency map for docfxV3


OsmondJiang:
- Dependency type
  - Ideal output
  - Impact files

PR #2817, opened 2018-05-28
add dependency map to manifest

OsmondJiang:


OsmondJiang:
#2816

docs/designs/output.md
```diff
@@ -126,9 +128,9 @@ These are called *browser navigation properties* and are stored in manifest.
 {
     "files": [{
         "url": "/dotnet/api/system.string",
-        "path": "en-us/netstandard-2.0/dotnet/api/system.string.json",
+        "site-path": "en-us/netstandard-2.0/dotnet/api/system.string.json",
         "locale": "en-us",
-        "moniker": "netstandard-2.0",
+        "moniker": "netstandard-2.0", 
 
         // other browser navigation properties...
```
yufeih: How about:
```json
"files": [{
    "source": "system.string.yml",
    "siteUrl": "/dotnet/api/system.string",
    "outputPath": "en-us/netstandard-2.0/dotnet/api/system.string.json",
    "locale": "en-us",
    "moniker": "netstandard-2.0",
    "dependencies": [{
        "type": "link",
        "source": "a.md"
    }]
}]
```

OsmondJiang: it would be a little big,  for difference purpose, i suggest to put them in two sections

OsmondJiang: and dependencies has no relationship with moniker?

PR #2818, opened 2018-05-29
link resolve and warnings for outof scope

OsmondJiang:


OsmondJiang:
#2819

src/docfx/lib/markdown/MarkdownUtility.cs
```diff
@@ -32,7 +32,7 @@ internal static class MarkdownUtility
             { "Caution", "Caution" },
         };
 
-        public static (string html, MarkupResult result) Markup(string markdown, Document file, Context context)
+        public static (string html, MarkupResult result) Markup(string markdown, Document file, Context context, ResolveHref resolveHref)
```
yufeih: This could be simplified to `Markup(md, file, context, Action<Document> buildChild)`

src/docfx/build/Build.cs
```diff
@@ -92,26 +93,36 @@ private static Task BuildAsset(Context context, Document file)
             return Task.CompletedTask;
         }
 
-        private static bool ShouldBuildFile(Document itemToBuild, ConcurrentDictionary<Document, byte> set, TableOfContentsMap tocMap)
+        /// <summary>
+        /// All children will be built through this gate
+        /// We control all the inclusion login here:
```
yufeih: nit: logic

src/docfx/build/Build.cs
```diff
@@ -47,19 +47,20 @@ private static List<Document> GlobFiles(Context context, Docset docset)
         {
             var manifest = new ConcurrentDictionary<Document, byte>();
             var references = new ConcurrentDictionary<Document, byte>();
+            var globbelFiles = new HashSet<Document>(files);
```
yufeih: nit: globbelFiles -> buildScope

docs/specs/build/toc/basics.yml
```diff
@@ -265,6 +265,9 @@ outputs:
        ]
     }
   build.manifest:
+  build.log: |
+    ["warning","referenced-content-outof-scope","the referenced content file docs/f1/n2.md is outof config's scope"]
```
yufeih: nit: "reference-out-of-scope" "*outof* -> *out* *of*"

docs/specs/build/toc/basics.yml
```diff
@@ -309,3 +310,7 @@ outputs:
        ]
     }
   build.manifest:
+  build.log: |
+    ["warning","referenced-content-outof-scope","the referenced content file docs/f1/f11/n1.md is outof config's scope"]
```
yufeih: nit: surround file path with '' in error/warning message in case there are spaces in file names

src/docfx/build/Build.cs
```diff
@@ -47,19 +47,20 @@ private static List<Document> GlobFiles(Context context, Docset docset)
...
-                    if (!ShouldBuildFile(file, manifest, tocMap))
+                    if (!ShouldBuildFile(context, file, tocMap, globbelFiles) || !manifest.TryAdd(file, 0))
                     {
                         return Task.CompletedTask;
                     }
 
                     return BuildOneFile(context, file, tocMap, item =>
                     {
-                        if (ShouldBuildFile(item, references, tocMap))
+                        if (references.TryAdd(item, 0) && ShouldBuildFile(context, item, tocMap, globbelFiles))
```
yufeih: you don't need the additional `references` dictionary. It could be:

```
if (!ShouldBuildFile(file)) return
if (!manifest.TryAdd(file)) return

return BuildOneFile(item => if (ShouldBuildFile()) { xxx })
```

Where `ShouldBuildFile` has:

```
if (manifest.Contains(file)) return false
```

src/docfx/lib/markdown/MarkdownUtility.cs
OsmondJiang: we don't want to introduce the build action to lower level component? 

---
In reply to: [191296672](https://github.com/dotnet/docfx/pull/2818#discussion_r191296672) [](ancestors = 191296672)

yufeih: For markdown, the lower level component is markdig and extensions. This is just a helper and it already depend on `Document`

src/docfx/build/Build.cs
OsmondJiang: there are a lot of files referenced by others more than 1 times, so the whole build pipeline will full of duplicated entities waiting to be queued, so we need filter them out from the beginning.

---
In reply to: [191297328](https://github.com/dotnet/docfx/pull/2818#discussion_r191297328) [](ancestors = 191297328)

src/docfx/lib/markdown/MarkdownUtility.cs
OsmondJiang: build action should belongs to build, it's strange to have a build action in markup method, right? the markup is to render markdown to html?

---
In reply to: [191297651](https://github.com/dotnet/docfx/pull/2818#discussion_r191297651) [](ancestors = 191297651)

src/docfx/build/markdown/BuildMarkdown.cs
```diff
@@ -1,17 +1,32 @@
...
+                context,
+                (a, b, c) =>
+                {
+                    // resolve href link
+                    var (link, buildItem) = Resolve.TryResolveHref(a, b, c);
+                    if (buildItem != null)
+                    {
+                        buildChild(buildItem);
+                    }
+
```
superyyrrzz: Is this for single file build? Seems like a separate story we will support later, not needed for normal build.

OsmondJiang: it's for single file build including it's dependency

---
In reply to: [191305652](https://github.com/dotnet/docfx/pull/2818#discussion_r191305652) [](ancestors = 191305652)

Issue #2699, opened 2018-04-27
Handle href resolution


OsmondJiang:
- [x] resolve basic markdown links
- [x] resolve html links inside markdown
- [ ] resolve TOC links
- [x] show warnings when resolve failed
- [x] handle escape
- [x] Handle dependencies
- [x] Handle output URL conflict
- [x] Handle output file path conflict
- [x] Handle files outside initial build scope

yufeih:
One of the goal is to switch between platforms without friction, here is the proposal:

- Require path to be case sensitive
- Support both forward slash and back slash
- Do not allow two files with same name but difference casing

OsmondJiang:
just add one item to cover, resolve on dependencies

OsmondJiang:
#2752

OsmondJiang:
one open question need to be discussed:

What's the resolved link in the published page? is it relative path or absolute path?

@yufeih , @herohua

herohua:
article
 in docset: [build] relative
            [dynamic rendering] append moniker
			[frontend js] continue context? (what in context)
 out docset (uid): [build] absolute (no locale, no host), or with host
            [dynamic rendering] don't append moniker
resource
 in docset: [build] absolute with locale (no host) -> TODO: confirm with hosting why?
            [dynamic rendering] append moniker

OsmondJiang:
i will follow up and give the update later.

OsmondJiang:
Original | Data link type | Build | Template without moniker | Template with moniker
-- | -- | -- | -- | --
relative path | relative Path | no change | no change | append moniker query string
absolute path start with https/http | external | no change | no change | no change
absolute path start with / | absolute path | appends locale if missing | no change | no change
UID(self contains) | relative path | resolve to relative path | no change | append moniker query string
UID(XREF Query) | absolute path | resolve to absolute path, got from Xref service | no change | no change

yufeih:
Ideally, *appends locale if missing* could be done in template just like *moniker*, because they are all for browser navigation right?

OsmondJiang:
yes, just list the existing behaviors here, will give out ideal behaviors for docfx v3 later

PR #2824, opened 2018-05-29
Add data-linktype to html on legacy mode

yufeih:
Put this logic behind `--legacy` because they can all be done in razor.

cc #2782 #2699


src/docfx/lib/HtmlUtility.cs
```diff
@@ -0,0 +1,75 @@
...
+                {
+                    node.SetAttributeValue("data-linktype", "self-bookmark");
+                    continue;
+                }
+                if (href.Contains(":"))
+                {
+                    node.SetAttributeValue("data-linktype", "external");
+                    continue;
+                }
+                if (href[0] == '/')
```
superyyrrzz: `|| href[0] == '\'`? like line 57

src/docfx/lib/HtmlUtility.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            try
+            {
+                var pos = href.IndexOfAny(new[] { '/', '\\' }, 1);
+                if (pos >= 1)
+                {
+                    var urlLocale = href.Substring(1, pos - 1);
+                    if (urlLocale.Contains("-"))
+                    {
+                        CultureInfo.GetCultureInfo(href.Substring(1, pos - 1));
```
928PJY: >CultureInfo.GetCultureInfo(href.Substring(1, pos - 1)); [](start = 24, length = 55)

this result never have been used?

yufeih: Yeah, it is just to check if the string is a valid locale, otherwise it will throw

Issue #2303, opened 2017-12-12
docfx metadata crashes with "The given key was not present in the dictionary"


evil-shrike:
**DocFX Version Used**: 
2.28.3

**Template used**: (`default` or `statictoc` or contain custom template)
`default`

Run
`docfx metadata ./docfx.json`

Getting "The given key was not present in the dictionary":
```
docfx metadata ./docfx.json -l Verbose
[17-12-12 11:14:03.708]Info:[ExtractMetadata]Using msbuild Z:\Prog\Microsoft Visual Studio\2017\Enterprise\MSBuild\15.0\Bin as inner compiler.
[17-12-12 11:14:04.076]Info:[ExtractMetadata]Loading projects...
[17-12-12 11:14:15.720]Warning:[ExtractMetadata]Project 'D:\Work\R-n-D\XFW3\Src\Transfer.Common\Transfer.Common.csproj' does not contain any documents.
[17-12-12 11:14:15.733]Warning:[ExtractMetadata]Project 'D:\Work\R-n-D\XFW3\Src\ReportService\ReportService.csproj' does not contain any documents.
[17-12-12 11:14:15.736]Warning:[ExtractMetadata]Project 'D:\Work\R-n-D\XFW3\Src\Transfer\Transfer.csproj' does not contain any documents.
[17-12-12 11:14:18.156]Info:[ExtractMetadata]'D:/Work/R-n-D/XFW3/Src/Configuration/Configuration.csproj' keep up-to-date since '12.12.2017 11:01:12', cached intermediate result 'D:/Work/R-n-D/XFW3/Src/Configuration/obj/xdoc/cache/obj\3waufjop.pbh' is used.
[17-12-12 11:14:18.326]Info:[ExtractMetadata]'D:/Work/R-n-D/XFW3/Src/Utils/Utils.csproj' keep up-to-date since '12.12.2017 11:01:13', cached intermediate result 'D:/Work/R-n-D/XFW3/Src/Utils/obj/xdoc/cache/obj\wfcni3xu.zzc' is used.
[17-12-12 11:14:18.333]Info:[ExtractMetadata]'D:/Work/R-n-D/XFW3/Src/Interop/Interop.csproj' keep up-to-date since '12.12.2017 11:01:13', cached intermediate result 'D:/Work/R-n-D/XFW3/Src/Interop/obj/xdoc/cache/obj\vgtfa01t.uok' is used.
[...]

PR #2825, opened 2018-05-29
log: always print out detailed error stack when extracting metadata fails

herohua:
@vicancy @superyyrrzz @vwxyzh @928PJY 

There are many reportings on metadata failure without detailed information, e.g. #2303. It would help troubleshooting if we have the entire error stack.


herohua:
@vicancy , what do you mean by unexpected exceptions?

PR #2832, opened 2018-06-04
Strip out unwanted html tags in markdown

yufeih:
#2782


src/docfx/lib/HtmlUtility.cs
```diff
@@ -71,5 +81,29 @@ private static string AddLocaleIfMissing(string href, string locale)
                 return '/' + locale + href;
             }
         }
+
+        public static void StripTags(HtmlNode html)
```
OsmondJiang: >StripTags [](start = 27, length = 9)

seems this is not only for legacy?

PR #2839, opened 2018-06-05
Add basic link resolve spec

yufeih:
#2699 #2782

PR #2846, opened 2018-06-06
Resolve html links embeded in markdown

yufeih:
#2782

This depend on #2839

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -4,27 +4,28 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Text;
 using HtmlAgilityPack;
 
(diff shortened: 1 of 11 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -118,6 +120,61 @@ public static void StripTags(HtmlNode html)
...
                 node.Remove();
             }
+            return html;
+        }
+
+        public static string TransformLinks(this string html, Func<string, string> transform)
+        {
+            // Fast pass it does not have <a> tag or <img> tag
+            if (!((html.Contains("<a") && html.Contains("href")) ||
+                  (html.Contains("<img") && html.Contains("src"))))
```
superyyrrzz: ignore case?

yufeih: yes

PR #2847, opened 2018-06-06
Output directory can only be specified in cmd args

yufeih:
Otherwise we don't know where to log config parsing errors.

This PR also add `docfx --version` and make `docfx` show help text.


src/docfx/cli/CommandLineOptions.cs
```diff
@@ -7,17 +7,10 @@ namespace Microsoft.Docs.Build
 {
     internal class CommandLineOptions
     {
-        public string Output;
-        public string Log;
-        public bool OutputLegacyModel;
+        public string Output = "_site";
```
OsmondJiang: >Output [](start = 22, length = 6)

you may want to change the config document to remove the output field

PR #2849, opened 2018-06-06
Disable CA1307 as it generates too much false positives

yufeih:
image


superyyrrzz:
@OsmondJiang docfx impact failed

OsmondJiang:
@superyyrrzz , the impact failure caused by the concurrency running trigger by PR and V3 updated, since the PR created from forked repo will never trigger the impact, i will disable PR trigger.

PR #2873, opened 2018-06-12
support legacy aggregated file map info

OsmondJiang:
used by OL

### DIFF
--- a/src/docfx/build/legacy/LegacyAggregatedFileMap.cs
+++ b/src/docfx/build/legacy/LegacyAggregatedFileMap.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Docs.Build
                 var aggregatedFileMapItem = new
                 {
                     aggregated_monikers = Array.Empty<string>(), // todo
-                    docset_names = new[] { docset.Name },
+                    docset_names = new[] { docset.Config.Name },
                     has_non_moniker_url = true, // todo
                     type = fileMapItem.Type,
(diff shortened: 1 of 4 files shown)


OsmondJiang:
#2782

src/docfx/docset/Docset.cs
```diff
@@ -27,6 +27,11 @@ internal class Docset
         /// </summary>
         public Repository Repository { get; }
 
+        /// <summary>
+        /// Gets the docset name
+        /// </summary>
+        public string Name => Config.Name;
```
yufeih: why not just use `docset.config.name`. people may forward every other property to `Docset`. I intend `Docset` to only contain properties that need further processing.

src/docfx/build/legacy/LegacyAggregatedFileMap.cs
```diff
@@ -0,0 +1,49 @@
...
+                    has_non_moniker_url = true, // todo
+                    type = fileMapItem.Type,
+                };
+
+                aggregatedFileMapItems.Add(PathUtility.NormalizeFile(Path.Combine(docset.Config.SourceBasePath, legacyFilePathRelativeToBaseFolder)), aggregatedFileMapItem);
+            }
+
+            context.WriteJson(
+                new
+                {
```
yufeih: 💛 👍 let's use anonymous types in legacy

PR #2885, opened 2018-06-13
Support json TOC

yufeih:
We should treat json just like yaml.


src/docfx/docset/Document.cs
```diff
@@ -182,30 +182,32 @@ public static Document TryCreate(Docset docset, string path)
...
+                path.EndsWith(".json", StringComparison.OrdinalIgnoreCase))
+            {
+                if (name.Equals("toc", StringComparison.OrdinalIgnoreCase))
+                {
+                    return ContentType.TableOfContents;
+                }
+                if (name.Equals("docfx", StringComparison.OrdinalIgnoreCase))
+                {
+                    return ContentType.Unknown;
+                }
```
OsmondJiang: consider the extensions

PR #2887, opened 2018-06-14
Resolve unknown xref as fallback

yufeih:
*edge-developer* content contains `@font-face` that are incorrectly parsed as xref. This PR lefts unknown xrefs as is and report an `xref-not-found` warning.

#2782

### DIFF
--- a/docs/specs/build/xref.yml
+++ b/docs/specs/build/xref.yml
@@ -2,10 +2,21 @@
 # Unknown xrefs are output as is with a warning
 inputs:
   docfx.yml:
-  docs/a.md: Link to @xref
+  docs/a.md: Link to <xref:a>
 outputs:
   docs/a.json: |
-    { "content": "<p>Link to @xref</p>" }
+    { "content": "<p>Link to &lt;xref:a&gt;</p>" }
   build.log: |
-    ["warning","xref-not-found","Cannot find uid 'xref' using xref '@xref'","docs/a.md"]
+    ["warning","uid-not-found","Cannot find uid 'a' using xref '<xref:a>'","docs/a.md"]
+  build.manifest:
+---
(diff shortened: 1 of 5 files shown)


docs/specs/build/xref.yml
```diff
@@ -0,0 +1,11 @@
+---
+# Unknown xrefs are output as is with a warning
+inputs:
+  docfx.yml:
+  docs/a.md: Link to @xref
+outputs:
+  docs/a.json: |
+    { "content": "<p>Link to @xref</p>" }
+  build.log: |
+    ["warning","xref-not-found","Cannot find uid 'xref' using xref '@xref'","docs/a.md"]
```
superyyrrzz: @xref needn't warning as it's prone to false positive

yufeih: Does <xref> syntax show warnings?

src/docfx/lib/markdown/ResolveXref.cs
```diff
@@ -0,0 +1,41 @@
...
+            {
+                document.Replace(node =>
+                {
+                    if (node is XrefInline xref)
+                    {
+                        var href = resolveXref(xref.Href);
+                        if (string.IsNullOrEmpty(href))
+                        {
+                            var raw = xref.GetAttributes().Properties.First(p => p.Key == "data-raw-source").Value;
+                            var error = raw.StartsWith("@")
```
superyyrrzz: >var error = raw.StartsWith("@") [](start = 28, length = 31)

can judge from "data-throw-if-not-resolved"="true" to log warning

yufeih: didn't know that. I think looking up html attributes for both `raw` and `should-throw `is a hack. I'll add them to markdig ast as part of the reference work.

PR #2889, opened 2018-06-14
Fix all warnings and treat warning as error on release builds

yufeih:
Treat warning as errors only in release build. This allows us to catch them in CI, but does not interfere with our dev work.


src/docfx/docfx.csproj
```diff
@@ -7,6 +7,7 @@
     <CodeAnalysisRuleSet>../../stylecop.ruleset</CodeAnalysisRuleSet>
     <RootNamespace>Microsoft.Docs.Build</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <TreatWarningsAsErrors Condition="'$(Configuration)'=='Release'">true</TreatWarningsAsErrors>
```
OsmondJiang: >'$(Configuration)'=='Release' [](start = 38, length = 29)

suggest to enable for all, the local build can fix these problem before pushing to remote

PR #2891, opened 2018-06-15
Html: fix self link bug, redirection content is empty

yufeih:
With this PR, html in `edge-developer` produces the exact same result except the following known differences:

|  | v2 | v3 |
|---|---|---|
|Links to redirected files  | file path | redirected url|
|Links to self | filename itself | `#`|

### DIFF
--- a/docs/specs/build/resolve.yml
+++ b/docs/specs/build/resolve.yml
@@ -26,9 +26,13 @@ inputs:
   docs/a.md: |
     ## Hello World
     Link to [bookmark](#hello-world)
+  docs/b.md: |
+    Link to [self](./b.md)
 outputs:
   docs/a.json: |
     { "content": "<h2 id=\"hello-world\">Hello World</h2><p>Link to <a href=\"#hello-world\">bookmark</a></p>" }
+  docs/b.json: |
+    { "content": "<p>Link to <a href=\"#\">self</a></p>" }
   build.manifest:
 ---
 # Support both / and \ as directory seperator
@@ -161,7 +165,7 @@ outputs:
   docs/b&/c~.json:
   build.manifest:
 ---
-# Reference to a redirection file(absolute)
+# Absolute link to a redirection file
 inputs:
   docfx.yml: |
     redirections:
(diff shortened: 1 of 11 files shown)


src/docfx/build/resolve/Resolve.cs
```diff
@@ -21,30 +21,43 @@ public static (DocfxException error, string content, Document file) TryResolveCo
             return file != null ? (error, file.ReadText(), file) : default;
         }
 
-        public static (DocfxException error, string href, string fragment, Document file) TryResolveHref(this Document relativeTo, string href, Document resultRelativeTo)
+        public static (DocfxException error, string href, Document file, bool hasBookmark) TryResolveHref(this Document relativeTo, string href, Document resultRelativeTo)
```
OsmondJiang: >bool hasBookmark [](start = 73, length = 16)

fragment doesn't means bookmark like #row=4?  it's better to returen the fragment, otherwise you need return more flags if we need

docs/specs/build/resolve.yml
```diff
@@ -26,9 +26,13 @@ inputs:
   docs/a.md: |
     ## Hello World
     Link to [bookmark](#hello-world)
+  docs/b.md: |
+    Link to [self](./b.md)
 outputs:
   docs/a.json: |
     { "content": "<h2 id=\"hello-world\">Hello World</h2><p>Link to <a href=\"#hello-world\">bookmark</a></p>" }
+  docs/b.json: |
+    { "content": "<p>Link to <a href=\"#\">self</a></p>" }
```
OsmondJiang: >"#\" [](start = 38, length = 4)

why add this? there is a standard for that?

src/docfx/build/resolve/Resolve.cs
```diff
@@ -34,12 +34,21 @@ public static (DocfxException error, string href, string fragment, Document file
...
                 return (error, href, fragment, file);
             }
 
+            // Self reference, leave href as is
+            if (file == relativeTo)
+            {
+                if (string.IsNullOrEmpty(fragment))
+                {
+                    fragment = "#";
+                }
```
OsmondJiang: why add this? is there a standard for that?

PR #2896, opened 2018-06-19
missing metadata

928PJY:


src/docfx/build/legacy/files/LegacyMarkdown.cs
```diff
@@ -64,11 +77,17 @@ internal static class LegacyMarkdown
 
         private static JObject GenerateLegacyRawMetadata(PageModel pageModel, Docset docset, Document file, GitRepoInfoProvider repo)
         {
+            var depotName = docset.Config.Name;
+            var items = depotName.Split('.');
```
superyyrrzz: use `string.IndexOf`, as docsetName may contain `.`

src/docfx/build/legacy/files/LegacyMarkdown.cs
```diff
@@ -90,6 +113,20 @@ private static JObject GenerateLegacyRawMetadata(PageModel pageModel, Docset doc
                 rawMetadata["original_content_git_url"] = repoInfo.GetGitLink(relPath);
             }
 
+            return AddOpPrefix(rawMetadata);
+        }
+
+        private static JObject AddOpPrefix(JObject rawMetadata)
```
superyyrrzz: can prioritize Jint work, which already contains this logic.

src/docfx/build/legacy/files/LegacyMarkdown.cs
yufeih: this can be merged with @OsmondJiang 's PR

PR #2898, opened 2018-06-19
fix redirection type to unblock impact

OsmondJiang:


src/docfx/docset/Document.cs
```diff
@@ -61,38 +61,30 @@ internal class Document
...
-        /// </summary>
-        public bool IsRedirection { get; }
+        public bool IsMasterContent { get; }
 
         /// <summary>
         /// Gets the document id and version independent id
         /// </summary>
         public (string docId, string versionIndependentId) Id => _id.Value;
 
+        private readonly ContentType _originType;
```
yufeih: where is this used?

PR #2911, opened 2018-06-20
Jeahyoun/23660 chromeless form

jeahyoun:
Create markdown extension to allow the creation of Chromeless Forms.

See the work item here: https://ceapex.visualstudio.com/Engineering/_workitems/edit/23660

Notes:

1. This extension pulls in the System.ComponentModel.DataAnnotations DLL as a dependency. If this will cause issues, it can be removed and we can write custom validation code.
2. This extension uses the same markdown syntax as the moniker plugin and the chromeless zones plugin. It does not refactor shared functionality into a base class, but this can be done.


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/ChromelessForms/ChromelessFormRender.cs
```diff
@@ -0,0 +1,38 @@
...
+	public class ChromelessFormRender : HtmlObjectRenderer<ChromelessFormBlock>
+	{
+		protected override void Write(HtmlRenderer renderer, ChromelessFormBlock obj)
+		{
+			obj.HtmlAttributes.AddClass("chromeless-form"); ;
+			renderer.Write("<form")
+				.WriteAttributes(obj)
+				.Write(@" class=""chromeless-form""")
+				.Write($"{(String.IsNullOrEmpty(obj.Model) ? "" : $" data-model=\"{obj.Model}\"")}")
+				.Write($"{(String.IsNullOrEmpty(obj.Action) ? "" : $" data-action=\"{obj.Action}\"")}")
```
yufeih: escape `obj.Model`/`obj.Action` or put it to `obj.HtmlAttributes`

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/ChromelessForms/ChromelessFormRender.cs
```diff
@@ -0,0 +1,38 @@
...
+				.WriteLine(">");
+
+			if (String.IsNullOrEmpty(obj.Model))
+			{
+				renderer.WriteLine($"<button type=\"submit\">{obj.SubmitText}</button>");
```
yufeih: html escape

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Microsoft.DocAsCode.MarkdigEngine.Extensions.csproj
```diff
@@ -10,4 +10,10 @@
     <ProjectReference Include="..\Microsoft.DocAsCode.MarkdigEngine.Validators\Microsoft.DocAsCode.MarkdigEngine.Validators.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="System.ComponentModel.DataAnnotations">
+      <HintPath>..\..\..\..\..\..\..\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.6.1\System.ComponentModel.DataAnnotations.dll</HintPath>
```
yufeih: Suggest to simply validate using code. <Reference> like this won't work in .net core.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MarkdownExtensions.cs
```diff
@@ -29,6 +29,7 @@ public static MarkdownPipelineBuilder UseDocfxExtensions(this MarkdownPipelineBu
                 .UseXref()
                 .UseEmojiAndSmiley(false)
                 .UseTabGroup(context)
+				.UseChromelessForm(context)
```
yufeih: nit: is this because template prefer to uses tabs and docfx prefer to use spaces?

herohua:
@qinezh @wh-alice

qinezh:
@yishengjin1413

jdanyow:
Might be an opportunity to consolidate with #2929 - creating a shell extension for the common `::: xyz` cases with support for multiple attributes.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/ChromelessForms/ChromelessFormRender.cs
jeahyoun: @yufeih I actually attempted this, but I couldn't get the output to work using the `WriteAttributes(obj)` call, which is why I added this afterwards (along with the `class` declaration, which also should be handled through `HtmlAttributes`. I looked at the Markdig codebase on Github and it appears that `WriteAttributes` uses reflection to get the `HtmlAttributes.Properties` list of KeyValuePairs and write it to the output. (You may be able to see this in the initial commit, actually.) I wasn't seeing the output at all. Is there something that I may have missed?

jdanyow: Markdig doesn't create the HtmlAttributes instance by default. I believe they do this to avoid the unnecessary memory overhead.

In triple-colon, I allow the extension implementations to return an `HtmlAttributes` collection via `TryProcessAttributes`. If an instances is returned it is associated with the block like this:
https://github.com/dotnet/docfx/pull/2929/files#diff-ef494d7b545157d836bf679679191860R64

jeahyoun: @jdanyow  Calling `AddClass` or `AddProperty` adds the collection if it doesn't exist. See https://github.com/lunet-io/markdig/blob/master/src/Markdig/Renderers/Html/HtmlAttributes.cs. I tried calling this in the renderer and in the parser, but I never saw the results in the HTML output.

jdanyow: you have to create an HtmlAttributes instance and associate it with your block.
```cs
var htmlAttributes = new HtmlAttributes();
block.SetData(typeof(HtmlAttributes), htmlAttributes);

// set an attribute
htmlAttributes.AddProperty("data-zone", target);
```

If it makes sense, consider rebasing on #2929

Then you'd just need to implement this interface: 
https://github.com/jdanyow/docfx/blob/ec4f54f169a85f153db6889caf2ba426a85a491f/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonExtension.cs#L50-L56

In `TryProcessAttributes` you'd validate the markdown attributes and transforms them to html attributes or class names as appropriate. You may need to extend the interface if you have custom rendering beyond html attributes and classes.

jeahyoun: @jdanyow See https://github.com/dotnet/docfx/pull/2911/files#diff-94c4df4127ff307a120f542c615c487bR20. Regardless, I'll rebase this to clean up the code.

jeahyoun:
@yufeih @jdanyow This has been refactored and is ready for a new review.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonExtension.cs
```diff
@@ -49,11 +50,12 @@ public void Setup(MarkdownPipeline pipeline, IMarkdownRenderer renderer)
...
-    {
-        string Name { get; }
+	public interface ITripleColonExtensionInfo
+	{
+		string Name { get; }
+		bool SelfClosing { get; }
         bool TryProcessAttributes(IDictionary<string, string> attributes, out HtmlAttributes htmlAttributes, Action<string> logError);
         bool TryValidateAncestry(ContainerBlock container, Action<string> logError);
-        // todo: "Render" function as-needed.
+		Func<HtmlRenderer, TripleColonBlock, bool> Render { get; }
```
yufeih: Use `bool Render(HtmlRenderer, TripleColonBlock)` over delegate for this interface

test/Microsoft.DocAsCode.MarkdigEngine.Tests/ChromelessFormsTest.cs
```diff
@@ -0,0 +1,93 @@
...
+				TestUtility.MarkupWithoutSourceInfo(content);
+			}
+			Logger.UnregisterListener(listener);
+
+			// Listener should have an error message and not output.
+			Assert.NotEmpty(listener.Items.Where(x => x.Code == "invalid-form"));
+		}
+
+		[Fact]
+		public void ChromelessFormsTestActionRequired()
```
yufeih: add a test case where `submitText` is `"`

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonExtension.cs
jeahyoun: Added.

test/Microsoft.DocAsCode.MarkdigEngine.Tests/ChromelessFormsTest.cs
jeahyoun: Added. Nothing breaks, although in practice I believe the content authors would have to escape the `"` for it to register properly.

yufeih: Sorry, I was trying to say `>` and see if it is properly escaped: `<button type=\"submit\">{submitText}</button>` --> `<button type=\"submit\">{HtmlUtility.HtmlEncode(submitText)}</button>`

jeahyoun: Ah, good catch! I will add this.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonExtension.cs
```diff
@@ -49,11 +50,12 @@ public void Setup(MarkdownPipeline pipeline, IMarkdownRenderer renderer)
...
-    {
-        string Name { get; }
+	public interface ITripleColonExtensionInfo
+	{
+		string Name { get; }
+		bool SelfClosing { get; }
         bool TryProcessAttributes(IDictionary<string, string> attributes, out HtmlAttributes htmlAttributes, Action<string> logError);
         bool TryValidateAncestry(ContainerBlock container, Action<string> logError);
-        // todo: "Render" function as-needed.
+		bool Render(HtmlRenderer renderer, TripleColonBlock block);
```
jdanyow: 👍

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonParser.cs
```diff
@@ -108,7 +119,12 @@ public override BlockState TryContinue(BlockProcessor processor, Block block)
 
         public override bool Close(BlockProcessor processor, Block block)
         {
-            var extensionName = ((TripleColonBlock)block).Extension.Name;
+			if ((block as TripleColonBlock).Extension.SelfClosing)
```
jdanyow: Little rusty in C# but I think `(TripleColonBlock)block` style cast will lead to better error message- invalid cast exception instead of null reference exception.

In any case, the nit pick I came for was to suggest casting once and using the reference in this if-statement as well as the `var extensionName = ...` line that follows.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonParser.cs
```diff
@@ -165,13 +181,13 @@ private bool TryMatchAttributeValue(ref StringSlice slice, out string value, str
...
 
-        private bool TryMatchAttributes(ref StringSlice slice, out IDictionary<string, string> attributes, string extensionName, Action<string> logError)
+        private bool TryMatchAttributes(ref StringSlice slice, out IDictionary<string, string> attributes, string extensionName, bool selfClosing, Action<string> logError)
         {
             attributes = EmptyAttributes;
             while (true)
             {
                 ExtensionsHelper.SkipSpaces(ref slice);
-                if (slice.CurrentChar.IsZero())
+                if (slice.CurrentChar.IsZero() || (selfClosing && ExtensionsHelper.MatchStart(ref slice, ":::")))
```
jdanyow: nice 🔥

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonRenderer.cs
```diff
@@ -9,6 +9,11 @@ public class TripleColonRenderer : HtmlObjectRenderer<TripleColonBlock>
     {
         protected override void Write(HtmlRenderer renderer, TripleColonBlock b)
         {
+			if (b.Extension.Render(renderer, b))
+			{
+				return;
+			}
+
```
jdanyow: can you restore the 2 space indents in all the modified files? We'll have strange diffs going forward if we don't keep the whitespace rules consistent.

jeahyoun: Yeah, I keep forgetting this.

test/Microsoft.DocAsCode.MarkdigEngine.Tests/ChromelessFormsTest.cs
jeahyoun: The `submitText` variable is now encoded using System.Net.WebUtility, and there is a test for the output.

PR #2917, opened 2018-06-22
Split Error from DocfxException

yufeih:
This refactor create an `Error` class to represent errors.

### DIFF
--- a/src/docfx/lib/log/ErrorLevel.cs
+++ b/src/docfx/lib/log/ErrorLevel.cs
@@ -3,8 +3,9 @@
 
 namespace Microsoft.Docs.Build
 {
-    internal enum ReportLevel
+    internal enum ErrorLevel
     {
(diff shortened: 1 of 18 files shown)


src/docfx/lib/log/ErrorLevel.cs
```diff
@@ -3,8 +3,9 @@
 
 namespace Microsoft.Docs.Build
 {
-    internal enum ReportLevel
+    internal enum ErrorLevel
     {
+        Off,
```
superyyrrzz: >Off [](start = 8, length = 3)

What does this mean?

yufeih: Turn errors and warnings off, to support a config entry like:
```
rules:
    file-not-found: warning
    invalid-yaml-header: off
```

PR #2920, opened 2018-06-22
fix legacy toc model

OsmondJiang:


OsmondJiang:
#2782

src/docfx/build/legacy/files/LegacyTableOfContents.cs
```diff
@@ -17,20 +17,42 @@ internal static class LegacyTableOfContents
...
             }
 
             context.Delete(doc.OutputPath);
             context.WriteJson(toc, legacyManifestOutput.TocOutput.ToLegacyOutputPath(docset));
             context.WriteJson(new { }, legacyManifestOutput.MetadataOutput.ToLegacyOutputPath(docset));
         }
 
+        private static void ConvertLegacyHref(IEnumerable<TableOfContentsItem> items)
+        {
+            if (items == null)
```
yufeih: For collections, suggest to always initialize it to an empty collection, to avoid null checks everywhere.

OsmondJiang: it's view model required, in the view model, if no children, the children field will be null

Issue #2838, opened 2018-06-05
YAML and JSON validation


superyyrrzz:
Some invalid inputs to consider:
- [x] If a dict has duplicate keys: log warning, continue/fail?
  - log error to output and continue
    - [x] YAML
    - ~~Json~~
- [x] If an array contains null element: log warning, remove null element
  - [x] YAML
  - [x] Json
- ~~If a key is not found or type mismatch in c# model, log warning~~

@OsmondJiang @yufeih

yufeih:
We could warning and de-null all null inputs, yaml parse result should never contain null value.

yufeih:
To encourage the use of non-null reference, the strategy here is:

- On **de-serialization**, remove nulls from arrays and property values: 
  `[null, 1, null]` --> `[1]`, `{a: null}` --> `{}`
- Models properties are initialized with a non-null default value
- On **serialization**, do not output empty strings/arrays/objects to reduce size
  `{a: ""}` --> `{}`, `{a: []}` --> `{}`, `{a: {}} ` --> `{}`

yufeih:
[yaml spec](http://yaml.org/spec/1.2/spec.html) does allow duplicated keys, if we follow that, it should be a warning and de-serialization should continue.

But I **DO** want to support just the [strict subset of YAML](http://hitchdev.com/strictyaml/), which we already did like dropping object keys, multiple documents, custom tags... One thing we support now but is cut in that spec is anchors and references.

If we use strict yaml, we should just error out on duplicated keys.

PR #2927, opened 2018-06-22
Add unit test for yaml deserializer with duplicated keys

live1206:


test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -269,6 +270,17 @@ public void TestStringEmpty()
...
         }
 
+        [Fact]
+        public void TestDuplicatedKeys()
+        {
+            var yaml = @"
+Key1: 0
+Key1: 0
+";
+            var exception = Assert.Throws<YamlException>(() => YamlUtility.Deserialize(new StringReader(yaml)));
```
OsmondJiang: >new StringReader(yaml) [](start = 87, length = 22)

just yaml? we have overload method to accept text

live1206: fixed

test/docfx.Test/docfx.Test.csproj
```diff
@@ -11,6 +11,7 @@
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.5.0" />
     <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
+    <PackageReference Include="YamlDotNet" Version="4.3.1" />
```
yufeih: unnecessary, `PackageReference` is transitive

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -269,6 +270,17 @@ public void TestStringEmpty()
             Assert.Null(value);
         }
 
+        [Fact]
+        public void TestDuplicatedKeys()
```
yufeih: Since you start working on this, can you take a look at #2838 ? You can ignore the `null` part of that issue. We should report an error if the input yaml contains duplicated keys.

live1206: sure thing

PR #2934, opened 2018-06-25
Preserve query and fragment for links to files

yufeih:
Per URL spec, query come before fragment, adjust SplitHref to respect that

### DIFF
--- a/src/docfx/build/toc/parser/TableOfContentsParser.cs
+++ b/src/docfx/build/toc/parser/TableOfContentsParser.cs
@@ -213,12 +213,6 @@ namespace Microsoft.Docs.Build
                     return (tocHref, default, default);
                 }
 
-                var (hrefPath, fragment, query) = HrefUtility.SplitHref(tocHref);
-                if (!string.IsNullOrEmpty(fragment) || !string.IsNullOrEmpty(query))
-                {
-                    // '#' and '?' is not allowed when referencing toc file
-                    tocHref = hrefPath;
-                }
                 var (referencedTocContent, referenceTocFilePath) = ResolveTocHrefContent(tocHrefType, tocHref, filePath, resolveContent);
                 if (referencedTocContent != null)
                 {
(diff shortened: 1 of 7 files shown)


test/docfx.Test/lib/HrefUtilityTest.cs
```diff
@@ -11,17 +11,18 @@ public static class HrefUtilityTest
...
-        [InlineData("a#b", "a", "#b", "")]
-        [InlineData("a#b<", "a", "#b<", "")]
-        [InlineData("a?c", "a", "", "?c")]
-        [InlineData("a?b#c", "a", "#c", "?b")]
-        public static void SplitHref(string href, string path, string fragment, string query)
+        [InlineData("a#b", "a", "", "#b")]
+        [InlineData("a#b<", "a", "", "#b<")]
+        [InlineData("a?c", "a", "?c", "")]
+        [InlineData("a?b#c", "a", "?b", "#c")]
+        [InlineData("a#b?c=d", "a", "", "#b?c=d")]
```
OsmondJiang: >"a#b?c=d" [](start = 20, length = 9)

add one test case: a?b#c?d=e -> a + ?b + #c?d=e

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -213,12 +213,6 @@ string GetTopicHref(TableOfContentsInputItem tocInputModel)
                     return (tocHref, default, default);
                 }
 
-                var (hrefPath, fragment, query) = HrefUtility.SplitHref(tocHref);
-                if (!string.IsNullOrEmpty(fragment) || !string.IsNullOrEmpty(query))
-                {
-                    // '#' and '?' is not allowed when referencing toc file
-                    tocHref = hrefPath;
-                }
```
OsmondJiang: why remove this logic?

yufeih: ResolveContent will split href internally, so this is unnecessary

OsmondJiang: what if the href is like 'a/b/?branch=master'

---
In reply to: [197993285](https://github.com/dotnet/docfx/pull/2934#discussion_r197993285) [](ancestors = 197993285)

yufeih: You are right 👍

PR #2940, opened 2018-06-26
add __global

superyyrrzz:
* can generate "_op_allContributorsStr"
* can fix "__global" null error

### DIFF
--- a/src/docfx/build/legacy/metadata/LegacyMetadata.cs
+++ b/src/docfx/build/legacy/metadata/LegacyMetadata.cs
@@ -46,6 +46,11 @@ namespace Microsoft.Docs.Build
             newMetadata["site_name"] = "Docs";
             newMetadata["version"] = 0;
 
+            newMetadata["__global"] = new JObject
+            {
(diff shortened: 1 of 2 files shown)


src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -28,7 +28,7 @@ internal static class LegacyMetadata
             "document_version_independent_id", "experiment_id", "experimental", "gitcommit", "keywords",
             "layout", "locale", "ms.assetid", "ms.author", "ms.date", "ms.prod", "ms.topic", "open_to_public_contributors", "original_content_git_url",
             "page_type", "pdf_url_template", "search.ms_docsetname", "search.ms_product", "search.ms_sitename", "site_name", "title", "titleSuffix", "toc_asset_id",
-            "toc_rel", "uhfHeaderId", "updated_at", "version", "word_count", "redirect_url",
```
928PJY: >__global [](start = 25, length = 8)

move to `GenerateLegacyRawMetadata`, this is not a common metadata

PR #2943, opened 2018-06-26
Feature/yaml error handling

live1206:
Pass along Yaml syntax error for TOC loading and add e2e test for it. Throw exception for Config loading.

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -810,4 +810,4 @@ outputs:
   docs/f1/b.json:
   build.manifest:
   build.log: |
-    ["error","yaml-syntax-error","(Line: 3, Col: 3, Idx: 63) - (Line: 3, Col: 12, Idx: 72): Duplicate key","docs/toc.yml"]
+    ["error","yaml-syntax-error","(Line: 3, Col: 3, Idx: 47) - (Line: 3, Col: 12, Idx: 56): Duplicate key","docs/toc.yml"]
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -797,7 +797,7 @@ outputs:
 inputs:
   docfx.yml:
   docs/toc.yml: |
-    - name: topic href has highest priority
+    - name: Duplicated keys
       topicHref: f1/a.md
       topicHref: f1/a.md
       tocHref: f1/
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -69,6 +69,17 @@ outputs:
     }
   build.manifest:
 ---
+# TOC json file is invalid
+inputs:
+  docfx.yml:
+  docs/toc.json: |
+    [{ "name": "title" ]
+outputs:
+  docs/toc.json:
+  build.manifest:
+  build.log: |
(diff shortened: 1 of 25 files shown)


src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -72,11 +66,11 @@ private static List<TableOfContentsInputItem> LoadTocModel(string content, strin
                 {
                     throw new NotSupportedException($"{filePath} is not a valid TOC file, detail: {ex.Message}.", ex);
```
OsmondJiang: >throw new NotSupportedException($"{filePath} is not a valid TOC file, detail: {ex.Message}.", ex); [](start = 20, length = 98)

also move this to error?

src/docfx/config/Config.cs
```diff
@@ -150,7 +151,12 @@ private static Config LoadCore(string configPath, CommandLineOptions options = n
         private static JObject LoadOriginalConfigObject(string configPath, List<string> parents, bool expand)
         {
             // TODO: support URL
-            var (_, config) = YamlUtility.Deserialize<JObject>(File.ReadAllText(configPath));
+            var (errors, config) = YamlUtility.Deserialize<JObject>(File.ReadAllText(configPath));
+            if (errors.Any())
```
OsmondJiang: >errors.Any() [](start = 16, length = 12)

will it be null?

docs/specs/toc.yml
```diff
@@ -792,4 +792,22 @@ outputs:
   build.log: |
     ["error","invalid-toc-href","The toc href 'f1/a.md' can only reference to a local TOC file, folder or absolute path","docs/toc.yml"]
     ["error","invalid-toc-href","The toc href 'f1/a.md' can only reference to a local TOC file, folder or absolute path","docs/toc.yml"]
-
+---
+# Toc syntax error with duplicated keys
+inputs:
+  docfx.yml:
+  docs/toc.yml: |
+    - name: topic href has highest priority
```
OsmondJiang: >topic href has highest priority [](start = 12, length = 31)

change the example name

docs/specs/toc.yml
```diff
@@ -792,4 +792,22 @@ outputs:
...
+      topicHref: f1/a.md
+      topicHref: f1/a.md
+      tocHref: f1/
+      href: f1/b.md
+  docs/f1/a.md:
+  docs/f1/b.md:
+outputs:
+  docs/toc.json:
```
OsmondJiang: >json [](start = 11, length = 4)

what's the output toc looks like?

src/docfx/config/Config.cs
live1206: nope, will  return an empty list if no error

docs/specs/toc.yml
live1206: the actual output toc is an empty list: 
{"items":[]}

src/docfx/build/toc/parser/TableOfContentsParser.cs
OsmondJiang: ?

---
In reply to: [198069924](https://github.com/dotnet/docfx/pull/2943#discussion_r198069924) [](ancestors = 198069924)

Issue #2813, opened 2018-05-25
generate required metadata


superyyrrzz:
|    Metadata      |   Assignee  |    Comment    |
|---------------   |-------------|---------------|
|_op_allContributorsStr | | #2940  |
|~_op_canonicalUrl~ | | Generated by template |
|~_op_canonicalUrlPrefix~ | | Input |
|~_op_displayDate~ | | Generated by template |
|_op_feedback_github_repo | | Template |
|_op_feedback_product_url | | Template |
|_op_feedback_system | | Template |
| ~_op_gitCommitHistory~ | @superyyrrzz | for intermediate use
| ~_op_gitContributionCommitHistory~ | @superyyrrzz | for intermediate use
|_op_gitContributorInformation | @superyyrrzz | #2912 
|_op_ogTitle | | **Investigate** |
|_op_openToPublicContributors | @superyyrrzz
|_op_pdfUrlPrefixTemplate | | Used by Pdf tool |
|_op_rawTitle | @928PJY |
|_op_searchScope | @fenxu | **Investigate** where add \_op_ prefix |
|_op_wordCount | @fenxu | **Investigate** where add \_op_ prefix |
| author           | @superyyrrzz| #2951 handle user profile cache #2812 |
| ~canonical_url~    | @928PJY  | Generated by template |
| content_git_url  | @superyyrrzz| #2957
| depot_name       | @928PJY |
| document_id      | | #2821 |
| document_version_independent_id | | #2821 |
| gitcommit        | @superyyrrzz| #2876 #2944 |
| is_dynamic_rendering | @928PJY |
| layout           | @928PJY |
| locale           | @928PJY |
| open_to_public_contributors | @superyyrrzz  | #2948
| original_content_git_url | @superyyrrzz | #2876 |
| ~page_type~        | @928PJY  | Generated by template | 
| pdf_url_template | | Generate by Pdf tool |
| redirect_document_id | | #2822 | 
| redirect_url     | | #2822 | 
| search.ms_docsetname | @fenxu | **Investigate** |
| search.ms_product | @fenxu | **Investigate** |
| search.ms_sitename | @fenxu | **Investigate** |
| site_name        | @928PJY |
| ~toc_asset_id~     | | Can be removed |
| toc_rel          | @928PJY |
| updated_at       | @superyyrrzz | #2962 #2969
| version          | @928PJY |
| word_count       | @928PJY  |

OsmondJiang:
#2882

PR #2948, opened 2018-06-27
add config for contribution

superyyrrzz:
#2813


src/docfx/config/Config.cs
```diff
@@ -77,10 +77,15 @@ internal class Config
         public readonly RouteConfig[] Routes = Array.Empty<RouteConfig>();
 
         /// <summary>
-        /// Gets the configuration about contributors.
+        /// Gets the configuration about contributors display in page.
         /// </summary>
         public readonly ContributorConfig Contributor = new ContributorConfig();
```
yufeih: can we merge these 2 into 1 section `Contribution`?

superyyrrzz: Sure. Contribution.Enabled to control whether to show "edit button", and use other config to control whether to display author/contributors.

src/docfx/models/PageModel.cs
```diff
@@ -30,6 +30,8 @@ internal class PageModel
 
         public DateTime UpdatedAt { get; set; }
 
+        public bool EnableContribution { get; set; }
```
OsmondJiang: >EnableContribution [](start = 20, length = 18)

Also add ContributionRepo?

Issue #2941, opened 2018-06-26
"Improve this doc" contribution URL for GitLab wiki repository


yulia-m:
I am using Gitlab wiki repositories as submodules in the Docfx project.  I am trying to get the "Improve this doc" button to work, but the URL pattern of GitLab wiki repositories is different from normal GitLab/GitHub repositories :

**GitHub/GitLab:**  .../{repository}**/blob/master/**{document}
**GitLab wiki:** .../{repository}**/wikis/**{document})

Is it possible to generate the correct contribution URL to GItLab wiki by changing _gitUrlPattern or _gitContribute path?

yufeih:
@superyyrrzz docs also need a git url pattern for public/private repo pairs

superyyrrzz:
@yulia-m This is a pattern other than `github` and `vso` that we already supported. We can add a new pattern like `github.wiki` to support this. Looks like GitHub and GitLab also shares the same URL pattern for wiki.

superyyrrzz:
@yufeih V2 already supports git url pattern, with logic in default template. V3 can put these logic in core.

PR #2957, opened 2018-06-27
generate edit link

superyyrrzz:


src/docfx/lib/git/GitRepoInfoProvider.cs
```diff
@@ -84,6 +84,26 @@ public GitRepoInfoProvider(Docset docset)
             return (errors, ToGitUserInfo(authorInfo), contributors.Select(ToGitUserInfo).ToArray(), updatedAt);
         }
 
+        public string GetEditLink(Document document)
+        {
+            Debug.Assert(document != null);
```
yufeih: How about using a string template for edit link? https://github.com/dotnet/docfx/issues/2941

src/docfx/models/PageModel.cs
```diff
@@ -34,6 +34,8 @@ internal class PageModel
 
         public bool EnableContribution { get; set; }
 
+        public string EditLink { get; set; }
```
yufeih: update spec to include `EditLink`

src/docfx/lib/git/GitRepoInfoProvider.cs
superyyrrzz: will do some refactor when supporting git host other than GitHub

PR #2962, opened 2018-06-28
can read update_date

superyyrrzz:


src/docfx/build/markdown/BuildMarkdown.cs
```diff
@@ -36,7 +36,10 @@ internal static class BuildMarkdown
             var (id, versionIndependentId) = file.Docset.Redirections.TryGetDocumentId(file, out var docId) ? docId : file.Id;
 
             // TODO: add check before to avoid case failure
-            var (repoErrors, author, contributors, updatedAt) = repo.GetContributorInfo(file, metadata.Value<string>("author"));
+            var (repoErrors, author, contributors, updatedAt) = repo.GetContributorInfo(
+                file,
+                metadata.Value<string>("author"),
+                metadata.Value<string>("update_date"));
```
yufeih: `metadata.Value<DateTime>("update_date")`

docs/specs/git.yml
```diff
@@ -23,6 +23,7 @@ outputs:
           "id": "19990166"
         }
       ],
+      "updatedAt": "2018-06-28T10:00:00",
```
yufeih: please add more test cases to cover `updatedAt` and `author`, a lot of code branches are not covered.

src/docfx/lib/git/GitRepoInfoProvider.cs
```diff
@@ -40,48 +40,63 @@ public GitRepoInfoProvider(Docset docset)
...
+                updateDateTime = userDateTime;
+            }
+            else if (commits?.Count > 0)
+            {
+                // TODO: support read build history
+                updateDateTime = commits[0].Time.DateTime;
+            }
+            else
+            {
+                updateDateTime = DateTime.Now;
```
yufeih: If we are here, the file is not managed by git or it does not have a commit. Suggest to use file last modified time as update time.

PR #2963, opened 2018-06-28
fix author and add _op_openToPublicContributors

928PJY:


src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -88,7 +88,7 @@ public static JObject GenerataCommonMetadata(JObject metadata, Docset docset)
                     : null,
                 ["update_at"] = pageModel.UpdatedAt.ToString(culture.DateTimeFormat.ShortDatePattern, culture),
             };
-            rawMetadata["author"] = pageModel.Author?.ProfileUrl.Substring(pageModel.Author.ProfileUrl.LastIndexOf('/'));
+            rawMetadata["author"] = pageModel.Author?.ProfileUrl.Substring(pageModel.Author.ProfileUrl.LastIndexOf('/') + 1);
```
yufeih: can we have a name field in GitUser? ProfileUser does not always ends with author name.

928PJY: really?
interesting!

yufeih: We should not depend on the schema of profile URL.

PR #2967, opened 2018-06-28
Handle gIt url fallback in case of detached head

yufeih:
#2782


src/docfx/lib/git/GitRepoInfoProvider.cs
```diff
@@ -114,7 +114,7 @@ public string GetEditLink(Document document)
                 ? $"https://github.com/{repoInfo.Account}/{repoInfo.Name}"
                 : document.Docset.Config.Contribution.Repository;
             var branch = string.IsNullOrEmpty(document.Docset.Config.Contribution.Branch)
-                ? repoInfo.Branch
+                ? repoInfo.Branch ?? "master"
```
superyyrrzz: can fall back to commit sha

Issue #2960, opened 2018-06-28
Fix error when open solution


superyyrrzz:
Fix error: Internal error: Critical capabilities changes were detected without any change made to the project 'Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests.fsproj', and forces the project to be reloaded. Potential capabilities involved: 'FSharp'.

Fix: https://github.com/dotnet/project-system/issues/3673#issuecomment-400625814

PR #2971, opened 2018-06-29
#2960 Try to fix fsharp project loading

live1206:


src/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.fsproj
```diff
@@ -1,38 +1,32 @@
...
+    <ProjectReference Include="..\Microsoft.DocAsCode.YamlSerialization\Microsoft.DocAsCode.YamlSerialization.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="Dotnet.ProjInfo" version="0.9.0" />
-    <PackageReference Include="FSharp.Compiler.Service" version="17.0.1" />
-    <PackageReference Include="FSharp.Compiler.Service.MSBuild.v12" version="17.0.1" />
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'net461' OR '$(TargetFramework)' == 'net46' OR '$(TargetFramework)' == 'net452'">
```
yufeih: It seems this section is the root cause, how about just removing this `<ItemGroup>` section?

live1206: target framework section was removed, since we have them in the shared part.
And package reference section was moved in front of fsharp files section, maybe this is the reason...

Issue #2972, opened 2018-06-29
Performance optimization


yufeih:
Build _azure-docs-pr_ currently takes more than 6 min, investigate and optimize build time.

- [ ] Do not load html twice to calculate word count
- [ ] CPU usage is around 20%, it should be higher

yufeih:
Creating a new `MarkdownPipeline` allocates lots of memory, especially in the [EmojiParser](https://github.com/lunet-io/markdig/blob/master/src/Markdig/Extensions/Emoji/EmojiParser.cs#L67). These tiny allocations literally turns our parallel processing into serial processing.

**Concurrency Visualizer** shows most threads are just sit there waiting (red part)

![image](https://user-images.githubusercontent.com/511355/42299430-b03b8f3c-803d-11e8-9ffc-a9856eba4276.png)

**Resource contention call tree** shows EmojiParser is heavily contended
![image](https://user-images.githubusercontent.com/511355/42299284-f644ab18-803c-11e8-9e8a-577db3676fb7.png)


**Functions view** shows the most contended functions are dictionary insersion and allocation (*JIT_New*)
![image](https://user-images.githubusercontent.com/511355/42299401-882d0a20-803d-11e8-841c-e407bc658502.png)
![image](https://user-images.githubusercontent.com/511355/42299511-24ced412-803e-11e8-874d-78ce59dfbbac.png)

The fix is reusing a _single_, _global_  `MarkdownPipeline` instance for all contents.

@herohua @qinezh @yishengjin1413 @OsmondJiang @superyyrrzz @live1206 @928PJY 

cc @xoofx

xoofx:
Yes, a MarkdownPipeline should be cached, it can be heavy to initialized and was actually built immutable so that multiple concurrent parsers can work with the same pipeline. Though, the TextMatchHelper class will require an optimization pass at some point (there is a TODO in the code about this memory consumption problem)

yufeih:
@xoofx , some pipelines are not thread safe. [this](https://github.com/lunet-io/markdig/blob/master/src/Markdig/Extensions/AutoIdentifiers/AutoIdentifierExtension.cs#L35) is an example I found so far, `headingWriter` is shared in `HeadingBlock_ProcessInlinesEnd`

xoofx:
> @xoofx , some pipelines are not thread safe. this is an example I found so far, headingWriter is shared in HeadingBlock_ProcessInlinesEnd

[...]

PR #2974, opened 2018-06-29
Optimize word count

yufeih:
`LoadHtml` has huge performance impact, this PR optimizes word count by: 

- not load a mutated html
- not split strings

Time to build *azure-docs-pr* (without git contribution info) drops roughly from 6 mins to 3 mins.

**Note**:

This implementation is slightly different from the original one, but is more correct:

- Do not special case `<title>` tag because title, it is already removed in markup phase
- Count both starting tags and ending tags as word separators, the old version only treat ending tags as separator (using string replace hack)

*This impacts legacy tests*, we could consider removing `word_count` in as impact test post process.

#2972


src/docfx/lib/HtmlUtility.cs
```diff
@@ -215,13 +199,35 @@ private static string AddLocaleIfMissing(string href, string locale)
...
+                    {
+                        word = false;
+                        total++;
+                    }
+                }
+                else if (
+                    ch != '.' && ch != '?' && ch != '!' &&
+                    ch != ';' && ch != ':' && ch != ',' &&
+                    ch != '(' && ch != ')' && ch != '[' &&
+                    ch != ']')
```
live1206: can we put special chars into HashSet<char> hashset, then call set.Contains(ch) to get a better performance?

yufeih: it depends

PR #2979, opened 2018-07-02
change all IO operation in legacy convert to context operation

928PJY:


src/docfx/docset/Context.cs
```diff
@@ -47,6 +47,11 @@ public void Report(Error error)
             _report.Write(error);
         }
 
+        public string ConvertToAbsolutePath(string relativePath)
```
herohua: >ConvertToAbsolutePath [](start = 22, length = 21)

add a comment

src/docfx/docset/Context.cs
```diff
@@ -47,6 +47,15 @@ public void Report(Error error)
             _report.Write(error);
         }
 
+        /// <summary>
+        /// Convert relative path to absolute path
+        /// </summary>
+        /// <param name="relativePath">Relatived to _outputPath</param>
+        public string ConvertToAbsolutePath(string relativePath)
```
OsmondJiang: >ConvertToAbsolutePath [](start = 22, length = 21)

doesn't make sense to put this method to Context, would we put it to Config.Output?

yufeih: or just use Path.Combine in legacy

yufeih:
👏

src/docfx/build/legacy/LegacyExtensions.cs
```diff
@@ -39,5 +39,10 @@ public static string ToLegacyOutputPath(this LegacyManifestOutputItem legacyMani
         {
             return Path.Combine(docset.Config.SiteBasePath ?? string.Empty, legacyManifestOutputItem.OutputPathRelativeToSiteBasePath);
         }
+
+        public static string ToAbsoluteOutputPath(this string relativePath, Docset docset)
```
OsmondJiang: >relativePath [](start = 62, length = 12)

all string will has this extension

928PJY: any advice?

---
In reply to: [199389546](https://github.com/dotnet/docfx/pull/2979#discussion_r199389546) [](ancestors = 199389546)

PR #2980, opened 2018-07-02
#2838 If yaml contains null value in mapping or array, log warning an…

live1206:
If Yaml contains null value in mapping or array, during deserialization log warning and remove null items.


superyyrrzz:
test failed

PR #2985, opened 2018-07-02
Optimize file glob to run match in parallel

yufeih:
This PR makes file glob first enumerate all files then run matches in parallel, this reduces `azure-docs-pr` glob file from 10s to 2.5s.

Traversing directories recursively is significantly slower than calling `EnumerateFiles(AllDirectories)`. 

#2972


src/docfx/lib/FileGlob.cs
```diff
@@ -1,122 +1,60 @@
...
 
-            foreach (var file in GetFilesFromSubfolder(cwd, cwd, globs, excludeGlobs))
-            {
-                yield return Path.GetFullPath(file);
-            }
-        }
+            var includeGlobs = patterns.Select(s => new GlobMatcher(s)).ToList();
+            var excludeGlobs = excludePatterns.Select(s => new GlobMatcher(s)).ToList();
+            var result = new ConcurrentBag<string>();
+            Parallel.ForEach(Directory.EnumerateFiles(cwd, "*.*", SearchOption.AllDirectories), MatchFile);
```
OsmondJiang: >ForEach [](start = 21, length = 7)

赞

Issue #2987, opened 2018-07-03
support restore lock file


OsmondJiang:
- [x]  support to restore to different place
- [x]  avoid site effects caused by another repo
- [x]  clean up unused restore work trees
- [x] clarify using `RestoreLockFile` and `FileMutex`
- [ ]  ~pass auth token instead of using git auth manager~

cc @fenxu  @superyyrrzz @yufeih

fenxu:
@OsmondJiang One proposal way is like current server logic: copy dependent repositories to repository root and pull there.

yufeih:
After enabling parallel e2e, restore may fail due to parallel execution: https://travis-ci.org/dotnet/docfx/builds/399431858?utm_source=github_status&utm_medium=notification

OsmondJiang:
@yufeih , i will take a look

OsmondJiang:
@yufeih , i can reproduce it locally, seems FileStream's Share option doesn't work on linux, i will try more and update later.

OsmondJiang:
https://bugzilla.xamarin.com/show_bug.cgi?id=38201

OsmondJiang:
http://0pointer.de/blog/projects/locking.html

OsmondJiang:
@fenxu , now docfx v3 support restore to a different place and using work trees to avoid noisy neighbor issue. You can set "DOCFX_APPDATA_PATH" environment to reset the restore root dir. 

The document will be added later.

PR #2993, opened 2018-07-03
fix process lock doesn't work on linux

OsmondJiang:
use `lock file` instead of `file lock` to handle process lock, mentioned at https://github.com/dotnet/corefx/issues/5964

### DIFF
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -126,14 +126,7 @@ namespace Microsoft.Docs.Build
             Directory.CreateDirectory(Path.GetDirectoryName(lockPath));
             using (var lockFile = await AcquireFileStreamLock(lockPath, retry < 0 ? 0 : retry, retryTimeSpanInterval ?? TimeSpan.FromSeconds(1)))
             {
-                try
-                {
-                    await action();
-                }
-                finally
-                {
-                    File.Delete(lockPath);
-                }
+                await action();
             }
         }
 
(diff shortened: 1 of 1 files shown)


OsmondJiang:
#2987

src/docfx/lib/ProcessUtility.cs
```diff
@@ -145,7 +144,7 @@ public static Task<string> Execute(string fileName, string commandLineArgs, stri
             {
                 try
                 {
-                    return new FileStream(lockPath, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.Delete);
+                    return new FileStream(lockPath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.Delete);
```
yufeih: can [fileoptions.deleteonclose](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileoptions?view=netframework-4.7.1#System_IO_FileOptions_DeleteOnClose) (aka [FILE_FLAG_DELETE_ON_CLOSE](https://docs.microsoft.com/en-us/windows/desktop/api/FileAPI/nf-fileapi-createfilea)) handle process crashes in between the using statement?

src/docfx/lib/ProcessUtility.cs
```diff
@@ -145,7 +137,7 @@ public static Task<string> Execute(string fileName, string commandLineArgs, stri
             {
                 try
                 {
-                    return new FileStream(lockPath, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.Delete);
+                    return new FileStream(lockPath, FileMode.CreateNew, FileAccess.ReadWrite, FileShare.None, 1, FileOptions.DeleteOnClose);
```
yufeih: since we don't read or write to the file, 1 is enough 👍

PR #2994, opened 2018-07-04
Feature/yaml null validation

live1206:
Validate if toc.yaml or toc.json contains null values, log warning and remove the null elements.
#2838


src/docfx/lib/YamlUtility.cs
```diff
@@ -151,25 +161,40 @@ private static JToken ToJson(YamlNode node)
...
                 }
+                SetMappings(mappings, node, arr);
                 return arr;
             }
             throw new NotSupportedException($"Unknown yaml node type {node.GetType()}");
         }
+
+        private static void SetMappings(Dictionary<JToken, List<(int, int)>> mappings, YamlNode scalar, JToken value)
+        {
+            if (mappings.ContainsKey(value))
```
yufeih: >ContainsKey [](start = 25, length = 11)

when will this be equal, if we use reference comparison, it should be a one one map

live1206: Some previous e2e tests are failing while setting the mapping during parsing as below:
Message: System.ArgumentException : An item with the same key has already been added. Key: { "metadata": { "key": "value" } }

src/docfx/Errors.cs
```diff
@@ -95,5 +95,11 @@ public static Error RedirectionDocumentIdConflict(IEnumerable<Document> redirect
 
         public static Error GitShadowClone(string repoPath)
             => new Error(ErrorLevel.Error, "git-shadow-clone", $"Does not support git shallow clone: '{repoPath}'");
+
+        public static Error NullPropertyValue(int startLine = 0, int startColumn = 0)
+            => new Error(ErrorLevel.Warning, "null-property", $"(Line: {startLine}, Column: {startColumn}) Property value is null");
```
herohua: >value [](start = 116, length = 5)

is it possible to print out the property name?

docs/specs/toc.yml
```diff
@@ -822,3 +822,88 @@ outputs:
   build.manifest:
   build.log: |
     ["error","yaml-syntax-error","(Line: 3, Col: 3, Idx: 47) - (Line: 3, Col: 12, Idx: 56): Duplicate key","docs/toc.yml"]
+---
```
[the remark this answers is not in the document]
live1206: if we put validatoin into YamlUtility/JsonUtility, yes

src/docfx/Errors.cs
live1206: yes, it can be added

src/docfx/Errors.cs
```diff
@@ -95,5 +95,11 @@ public static Error RedirectionDocumentIdConflict(IEnumerable<Document> redirect
 
         public static Error GitShadowClone(string repoPath)
             => new Error(ErrorLevel.Error, "git-shadow-clone", $"Does not support git shallow clone: '{repoPath}'");
+
+        public static Error NullPropertyValue(int startLine = 0, int startColumn = 0)
```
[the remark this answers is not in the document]
live1206: if the error message provide list name or property name, maybe it is enough saying the name contains null value

src/docfx/lib/LineInfo.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    internal class LineInfo
+    {
+        public LineInfo(int lineNumber, int linePosition)
+        {
+            LineNumber = lineNumber;
+            LinePosition = linePosition;
+        }
+
+        internal int LineNumber { get; set; }
```
herohua: >internal [](start = 8, length = 8)

this can be `public`

src/docfx/lib/JsonUtility.cs
```diff
@@ -125,6 +125,17 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
...
 
+        public static bool IsNullOrEmpty(this JToken token)
+        {
+            return
+                (token == null) ||
+                (token.Type == JTokenType.Array && !token.HasValues) ||
+                (token.Type == JTokenType.Object && !token.HasValues) ||
+                (token.Type == JTokenType.String && string.IsNullOrEmpty(token.ToString())) ||
+                (token.Type == JTokenType.Null) ||
+                (token.Type == JTokenType.Undefined);
```
herohua: >JTokenType.Undefined [](start = 31, length = 20)

just curious: what does this mean? any example?

src/docfx/lib/MappingKey.cs
```diff
@@ -0,0 +1,31 @@
...
+        public bool Equals(MappingKey other)
+        {
+            if (other == null)
+                return false;
+            return Key == other.Key;
+        }
+
+        public override int GetHashCode()
+        {
+            return Key.GetHashCode();
```
herohua: >Key [](start = 19, length = 3)

`Key` could be `null`

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -65,24 +68,73 @@ private static (List<Error> errors, List<TableOfContentsInputItem>) LoadTocModel
...
+            void LogWarningForNullValue(JToken item, Func<int, int, Error> log)
+            {
+                if (fromJsonFile)
+                {
+                    var lineInfo = item as IJsonLineInfo;
+                    errors.Add(log(lineInfo.LineNumber, lineInfo.LinePosition));
+                }
+                else
+                {
+                    if (mappings.TryGetValue(new MappingKey { Key = item }, out LineInfo value))
```
herohua: >new MappingKey { Key = item }, out LineInfo value [](start = 45, length = 49)

can we actually get out the value?

src/docfx/lib/Range.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Docs.Build
+{
+    internal struct Range
+    {
+        public Range(int startLine, int startCharacter, int endLine = 0, int endCharacter = 0)
```
yufeih: >Range [](start = 15, length = 5)

since you created this constructor, why not make Range readonly?

`internal readonly struct Range
{
     public readonly int StartLine;
}`

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -57,7 +57,7 @@ public static class E2ETest
         private static TheoryData<string, int> FindTestSpecs()
         {
             var result = new TheoryData<string, int>();
-            foreach (var file in Directory.EnumerateFiles("specs", "*.yml", SearchOption.AllDirectories))
+            foreach (var file in Directory.EnumerateFiles("specs", "toc.yml", SearchOption.AllDirectories))
```
yufeih: >toc [](start = 68, length = 3)

don't forget to remove this

src/docfx/lib/Range.cs
live1206: fixed

test/docfx.Test/e2e/E2ETest.cs
live1206: fixed

src/docfx/lib/JsonUtility.cs
```diff
@@ -94,6 +92,15 @@ public static T Deserialize<T>(string json)
             return Deserialize<T>(new StringReader(json));
         }
 
+        /// <summary>
+        /// Parse a string to JToken.
+        /// Validate null value during the process.
+        /// </summary>
+        public static (List<Error>, JToken) Parse(string json)
```
yufeih: >Parse [](start = 44, length = 5)

Instead of having a new `Parse` method, modify the above 2 `Deserialize` methods to return `(errors, T, sourceMap)`, the idea is for JSON and YAML to share the same API signature.

You could implement JSON source map later (by using JsonReader API)

src/docfx/lib/JsonUtility.cs
```diff
@@ -125,6 +132,91 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
             return result;
         }
 
+        public static bool IsNullOrEmpty(this JToken token)
```
yufeih: >IsNullOrEmpty [](start = 27, length = 13)

IsNullOrUndefined ?

src/docfx/lib/YamlUtility.cs
```diff
@@ -64,33 +64,34 @@ public static string ReadHeader(string yaml)
...
-            return (errors, json.ToObject<T>(JsonUtility.DefaultDeserializer));
+            var (errors, json, mappings) = Deserialize(reader);
+            return (errors, json.ToObject<T>(JsonUtility.DefaultDeserializer), mappings);
         }
 
         /// <summary>
         /// Deserialize to JToken From string
         /// </summary>
-        public static (List<Error> errors, JToken) Deserialize(string input)
+        public static (List<Error> errors, JToken jtoken, Dictionary<MappingKey, LineInfo> mappings) Deserialize(string input)
```
superyyrrzz: >jtoken [](start = 50, length = 6)

minor: token

src/docfx/lib/JsonUtility.cs
live1206: fixed

src/docfx/lib/JsonUtility.cs
live1206: Unit tests for these have been added to JsonUtilityTest and YamlUtilityTest

src/docfx/lib/JsonUtility.cs
```diff
@@ -125,6 +121,103 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
...
+        /// </summary>
+        private static (List<Error>, JToken) Deserialize(TextReader reader)
+        {
+            using (JsonReader json = new JsonTextReader(reader))
+            {
+                return DefaultDeserializer.Deserialize<JToken>(json).ValidateNullValue();
+            }
+        }
+
+        private static void Traverse(this JToken token, List<Error> errors, JTokenSourceMap mappings, List<JToken> nullNodes)
```
yufeih: this can be a single function, because we only care about 3 `JToken` types: `JObject`, `JArray` and `JValue`, the skeleton can be:

```csharp
void traverse(JToken token)
{
    if (token is JArray arr) {}
    else if (token is JObject obj) {}
    else if (token is JValue v) {}
}
```

src/docfx/lib/JsonUtility.cs
```diff
@@ -125,6 +121,103 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
...
+                else if (prop.Value is JArray)
+                {
+                    prop.Value.Traverse(errors, mappings, nullNodes);
+                }
+            }
+        }
+
+        private static void LogWarningForNullValue(JToken item, List<Error> errors, JTokenSourceMap mappings)
+        {
+            if (mappings == null)
```
yufeih: This should never be null ?

src/docfx/lib/JsonUtility.cs
```diff
@@ -125,6 +121,103 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
...
+            {
+                var lineInfo = item as IJsonLineInfo;
+                errors.Add(Errors.NullValue(new Range(lineInfo.LineNumber, lineInfo.LinePosition)));
+            }
+            else
+            {
+                if (mappings.TryGetValue(item, out Range value))
```
yufeih: item should always be in sourcemap? `Debug.Assert(mappings.Contains(item))`

src/docfx/lib/Range.cs
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Docs.Build
+{
+    internal readonly struct Range
```
yufeih: Make `Range` a readonly struct with readonly fields

src/docfx/lib/JsonUtility.cs
live1206: This is not null for Yaml parsing, null for Json parsing

---
In reply to: [200284005](https://github.com/dotnet/docfx/pull/2994#discussion_r200284005) [](ancestors = 200284005)

src/docfx/lib/YamlUtility.cs
```diff
@@ -151,25 +163,33 @@ private static JToken ToJson(YamlNode node)
...
-                    arr.Add(ToJson(item));
+                    arr.Add(ToJson(item, mappings));
                 }
+                SetMappings(mappings, node, arr);
                 return arr;
             }
             throw new NotSupportedException($"Unknown yaml node type {node.GetType()}");
         }
+
+        private static void SetMappings(JTokenSourceMap mappings, YamlNode scalar, JToken value)
```
yufeih: you can make this method returns `JToken`, so the above return statements can remain single line:

```csharp
return SetMapping(mappings, scalar, new JValue(d))
```

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -58,9 +59,11 @@ private static (List<Error> errors, List<TableOfContentsInputItem>) LoadTocModel
...
             {
                 JToken tocToken;
+                List<Error> nullErrors;
                 try
                 {
-                    tocToken = JToken.Parse(content);
+                    (nullErrors, tocToken) = JsonUtility.Deserialize<JToken>(content);
+                    errors.AddRange(nullErrors);
                 }
                 catch (Exception ex)
```
yufeih: remove the catch here because errors are reported using return values

src/docfx/lib/JsonUtility.cs
live1206: Right, but we only care about JArray and JObject, since if the value of JObject is null, we will log warning and remove this JObject.

---
In reply to: [200283844](https://github.com/dotnet/docfx/pull/2994#discussion_r200283844) [](ancestors = 200283844)

src/docfx/build/toc/parser/TableOfContentsParser.cs
live1206: cannot just remove, since catch will take care of json syntax errors, will hide them in JsonUtility instead then

---
In reply to: [200285659](https://github.com/dotnet/docfx/pull/2994#discussion_r200285659) [](ancestors = 200285659)

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -124,10 +124,15 @@ public void TestBasicClassWithNullCharactor()
 ";
```
herohua: this is invalid?

live1206: During Yaml de-serialization, we treat scalar value ~ as null, therefore it will be removed during the null value validation

---
In reply to: [200322112](https://github.com/dotnet/docfx/pull/2994#discussion_r200322112) [](ancestors = 200322112)

src/docfx/lib/JsonUtility.cs
```diff
@@ -125,6 +123,101 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
...
+        {
+            if (mappings == null)
+            {
+                var lineInfo = item as IJsonLineInfo;
+                errors.Add(Errors.NullValue(new Range(lineInfo.LineNumber, lineInfo.LinePosition)));
+            }
+            else
+            {
+                Debug.Assert(mappings.ContainsKey(item));
+                if (mappings.TryGetValue(item, out Range value))
```
yufeih: nit: Given the assertion above, use `var value = mappings[item]` to throw an exception if something goes terribly wrong,

PR #2996, opened 2018-07-04
change restore dir

OsmondJiang:


OsmondJiang:
#2987

src/docfx/restore/Restore.cs
```diff
@@ -52,6 +52,22 @@ public static (string restoreDir, string url, string refSpec) GetGitRestoreInfo(
...
         }
 
+        /// <summary>
+        /// Get the restore root dir, default is the user proflie dir.
+        /// User can set the DOCFX_APPDATA_PATH environment to change the root
+        /// </summary>
+        public static string GetRestoreRoot()
+        {
+            // TODO: document this environment variable and show it in welcome message
+            var docfxAppData = Environment.GetEnvironmentVariable("DOCFX_APPDATA_PATH");
```
yufeih: this also affects lock, user profile cache. add a static `AppDataPath` to manage them all.

src/docfx/config/AppData.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        private static readonly string s_appDataPath = GetAppDataPath();
+
+        /// <summary>
+        /// Get the restore root dir, default is the user proflie dir.
+        /// User can set the DOCFX_APPDATA_PATH environment to change the root
+        /// </summary>
+        private static string GetAppDataPath()
+        {
+            // TODO: document this environment variable and show it in welcome message
```
yufeih: we could document this, but don't show it in welcome :)

src/docfx/config/AppData.cs
```diff
@@ -0,0 +1,35 @@
...
+            var docfxAppData = Environment.GetEnvironmentVariable("DOCFX_APPDATA_PATH");
+            if (!string.IsNullOrEmpty(docfxAppData))
+            {
+                docfxAppData = Path.GetFullPath(docfxAppData);
+            }
+
+            return docfxAppData ?? Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
```
yufeih: seem like empty string leaks through, how about
```csharp
string.IsNullOrEmpty(docfxAppData) ? UserProfile : GetFullPath(appdata)
```

OsmondJiang: unless use set a environment value is ""...

Issue #2966, opened 2018-06-28
Case sensitive resolving causes broken links for docs repos


OsmondJiang:
Right now the resolving file link logic is case sensitive which cause some files can't be resolved in docs repo.

cc @yufeih  @herohua

PR #3004, opened 2018-07-05
Use OS case sensitivity settings

yufeih:
This PR added `PathComparer`, `PathComparison` and `IsCaseSensitive` to `PathUtility`, and modifies existing comparison to use os path comparison.

Special files like `docfx.yml`, `TOC.md`, `index.md` are ~~always treated as case insensitive regardless of os sensitivity~~ **also treated using os path comparison**.

#2966 #2782


src/docfx/build/legacy/LegacyExtensions.cs
```diff
@@ -27,12 +27,12 @@ public static string ToLegacyOutputPathRelativeToBaseSitePath(this Document doc,
         public static string ToLegacySiteUrlRelativeToBaseSitePath(this Document doc, Docset docset)
         {
             var legacyOutputFilePathRelativeToSiteBasePath = doc.SiteUrl;
-            if (legacyOutputFilePathRelativeToSiteBasePath.StartsWith($"/{docset.Config.SiteBasePath}", StringComparison.Ordinal))
+            if (legacyOutputFilePathRelativeToSiteBasePath.StartsWith($"/{docset.Config.SiteBasePath}", PathUtility.PathComparison))
             {
                 legacyOutputFilePathRelativeToSiteBasePath = Path.GetRelativePath(docset.Config.SiteBasePath, doc.SiteUrl.Substring(1));
             }
 
-            return PathUtility.NormalizeFile(doc.FilePath.EndsWith("index.md", StringComparison.OrdinalIgnoreCase) ? $"{legacyOutputFilePathRelativeToSiteBasePath}index" : legacyOutputFilePathRelativeToSiteBasePath);
```
herohua: >OrdinalIgnoreCase [](start = 96, length = 17)

this one should always ignore case?

yufeih: we need a strategy for handling special file names like _TOC_, _index_, _docfx_. This PR always uses case insensitive comparison for these special files, but we could use os specific comparison.

yufeih: I'm leaning towards using os specifc comparison for these special file names. Sometimes we need to probe files like TOC.md, index.md, this is problematic on linux if it is case insensitive.

herohua: makes sense. let's go with OS specific

---
In reply to: [200537310](https://github.com/dotnet/docfx/pull/3004#discussion_r200537310) [](ancestors = 200537310)

PR #3005, opened 2018-07-05
Add details about metadata src

sandorfr:
Adding a sample to reference project located outside of docfx directory.


superyyrrzz:
Thank you for the contribution! I wonder can this information be some note of step 2? It seems not an essential step after the first 5 steps.

sandorfr:
more like this ? https://github.com/dotnet/docfx/compare/dev...sandorfr:patch-1?expand=1

PR #3007, opened 2018-07-06
fix legacy title error

928PJY:


docs/specs/markdown.yml
```diff
@@ -17,22 +17,47 @@ outputs:
...
+      "metadata": { "title": "title from yaml header" }
     }
   build.manifest:
 ---
+# Yaml header title in upper-case
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    ---
+    Title: Title from yaml header
```
yufeih: if we want to enforce casing here, we should also enforce camel casing during `JToken -> T`:

Deserialize `{ AB: 1 }` into `class { int AB }` should produce `{ AB: 0 }`
Deserialize `{ aB: 1 }` into `class { int AB }` should produce `{ AB: 1 }`

@live1206

docs/specs/markdown.yml
```diff
@@ -17,22 +17,47 @@ outputs:
...
-# Yaml header
+# Yaml header title in lower-case
 inputs:
   docfx.yml:
   docs/a.md: |
     ---
-    title: me
+    title: title from yaml header
     ---
+    # Title from H1
```
yufeih: don't modify this test, it covers the cases where there is a title in the header but no title in markdown

PR #3009, opened 2018-07-06
Improve error message for YAML syntax error by adding inner error mes…

live1206:
Improve error message for YAML syntax error by adding inner error message

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Docs.Build
             => new Error(ErrorLevel.Warning, "yaml-header-not-object", $"Expect yaml header to be an object, but got {(isArray ? "an array" : "a scalar")}", filePath.ToString());
 
         public static Error YamlSyntaxError(Exception ex)
-            => new Error(ErrorLevel.Error, "yaml-syntax-error", ex.Message);
+            => new Error(ErrorLevel.Error, "yaml-syntax-error", $"{ex.Message}. {ex.InnerException?.Message}");
 
         public static Error InvalidYamlHeader(Document file, Exception ex)
             => new Error(ErrorLevel.Warning, "invalid-yaml-header", ex.Message, file.ToString());
--- a/test/docfx.Test/lib/YamlUtilityTest.cs
+++ b/test/docfx.Test/lib/YamlUtilityTest.cs
@@ -299,7 +299,7 @@ Key1: 0
             {
                 Assert.Equal(ErrorLevel.Error, error.Level);
                 Assert.Equal("yaml-syntax-error", error.Code);
-                Assert.Contains("Duplicate key", error.Message);
+                Assert.Contains("An item with the same key has already been added. Key: Key1", error.Message);
             });
         }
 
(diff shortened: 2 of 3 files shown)


test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -299,7 +299,7 @@ public void TestDuplicatedKeys()
             {
                 Assert.Equal(ErrorLevel.Error, error.Level);
                 Assert.Equal("yaml-syntax-error", error.Code);
-                Assert.Contains("Duplicate key", error.Message);
+                Assert.Contains("An item with the same key has already been added. Key: Key1", error.Message);
```
herohua: >An item with the same key has already been added. Key: Key1 [](start = 33, length = 59)

I don't think this message is clear enough for users to take action. How about `Key 'Key1' is already defined. Please remove the duplicate key.`

yufeih: [Per MS style guide, avoid please](https://github.com/dotnet/docfx/commit/3ea6ddb104647ec432900f2361f6e1540c57b1ba)

---
In reply to: [200588479](https://github.com/dotnet/docfx/pull/3009#discussion_r200588479) [](ancestors = 200588479)

src/docfx/Errors.cs
```diff
@@ -52,7 +52,10 @@ public static Error YamlHeaderNotObject(object filePath, bool isArray)
             => new Error(ErrorLevel.Warning, "yaml-header-not-object", $"Expect yaml header to be an object, but got {(isArray ? "an array" : "a scalar")}", filePath.ToString());
 
         public static Error YamlSyntaxError(Exception ex)
-            => new Error(ErrorLevel.Error, "yaml-syntax-error", ex.Message);
+            => new Error(ErrorLevel.Error, "yaml-syntax-error", $"{ex.Message}. {ex.InnerException?.Message}");
```
yufeih: >Message [](start = 70, length = 7)

can we extract line column info from YamlSyntaxException and put them to Error ?

live1206: Yes, we can. But it wouldn't make any difference for this one, would simplify refinement for the duplicate key

---
In reply to: [200893825](https://github.com/dotnet/docfx/pull/3009#discussion_r200893825) [](ancestors = 200893825)

herohua: so the *original* error message with syntax error is clear regarding line/column info?

---
In reply to: [200897131](https://github.com/dotnet/docfx/pull/3009#discussion_r200897131) [](ancestors = 200897131,200893825)

live1206: We only redefined the error message of duplicate key for now, keep the rest as they are. If we want to redefine more, we can do it accordingly.

---
In reply to: [200916391](https://github.com/dotnet/docfx/pull/3009#discussion_r200916391) [](ancestors = 200916391,200897131,200893825)

live1206: misunderstood above, yes, the inner error message contains clear line/column info

---
In reply to: [200916391](https://github.com/dotnet/docfx/pull/3009#discussion_r200916391) [](ancestors = 200916391,200897131,200893825)

PR #3012, opened 2018-07-06
Fix for parsing null json

live1206:


test/docfx.Test/lib/JsonUtilityTest.cs
```diff
@@ -230,6 +230,16 @@ public void TestListItemWithNullValue()
             Assert.Equal("{\"name\":\"title\",\"items\":[{\"displayName\":\"1\"}]}", resultJsonString);
         }
 
+        [Fact]
+        public void TestNullJson()
+        {
+            var json = "{}";
```
OsmondJiang: change it to ""

PR #3019, opened 2018-07-09
check if merge key exists when splitting yaml

TianqiZhang:


src/Microsoft.DocAsCode.Build.SchemaDriven/SchemaFragmentsIterator.cs
```diff
@@ -102,8 +102,15 @@ public SchemaFragmentsIterator(ISchemaFragmentsHandler handler)
...
-                            var opath = string.Format("{0}[{1}=\"{2}\"]", parentOPath, mergeKey, mapNode.Children[mergeKey].ToString());
-                            TraverseCore(item, fragments, schema.Items, opath, uid);
+                            if (mapNode.Children.ContainsKey(mergeKey))
+                            {
+                                var opath = string.Format("{0}[{1}=\"{2}\"]", parentOPath, mergeKey, mapNode.Children[mergeKey].ToString());
+                                TraverseCore(item, fragments, schema.Items, opath, uid);
+                            }
+                            else
+                            {
+                                Logger.LogError($"Cannot find merge key {mergeKey} in {mapNode.ToString()}");
```
superyyrrzz: Do you need return here to terminate the method?

herohua: >.ToString() [](start = 94, length = 11)

no need

herohua: >LogError [](start = 39, length = 8)

@superyyrrzz@gmail.com 
When there is error, will the caller of SchemaFragmentsIterator terminate?

superyyrrzz: No, only an error is logged.

herohua:
add unit test

src/Microsoft.DocAsCode.Build.SchemaDriven/SchemaFragmentsIterator.cs
herohua: is it expected?

---
In reply to: [200884075](https://github.com/dotnet/docfx/pull/3019#discussion_r200884075) [](ancestors = 200884075)

superyyrrzz: Seems SchemaFragmentsIterator should throw, while the call should catch and handle it. I'm not sure whether exception is needed here. Will talk to Tianqi about this.

TianqiZhang:
thanks guys for reviewing this, as the PR's purpose is just to expose more information when the certain error occurs, UT will be added later.

Issue #2863, opened 2018-06-08
Improve error and logging


yufeih:
## Add missing error code to:

- [ ] Config
- [ ] Toc build
- [ ] Markdown build
- [ ] Link resolve
- [ ] Yaml deserializer
- [ ] Metadata
- [ ] Restore

## Add detailed line/column/range into to:

- [ ] Yaml
- [ ] Markdown
- [ ] TOC

@herohua @superyyrrzz @OsmondJiang @928PJY

PR #3022, opened 2018-07-09
Show warning when git is not installed

yufeih:
#2863


src/docfx/lib/git/GitUtility.cs
```diff
@@ -128,9 +129,14 @@ private static Task<string> ExecuteQuery(string cwd, string commandLineArgs, Tim
...
             Debug.Assert(!string.IsNullOrEmpty(cwd));
 
-            // todo: check git exist or not
-            var response = await ProcessUtility.Execute("git", commandLineArgs, cwd, timeout, outputHandler);
-            return parser(response);
+            try
+            {
+                return parser(await ProcessUtility.Execute("git", commandLineArgs, cwd, timeout, outputHandler));
+            }
+            catch (Win32Exception ex) when (ProcessUtility.IsNotFound(ex))
```
herohua: >Win32Exception [](start = 19, length = 14)

this is not cross platform?

yufeih: this **is** xplat, linux test passed. 

.Net core cannot make breaking changes like this (renaming exception), so it'll still throw Win32Exception on linux and mac.

src/docfx/Errors.cs
```diff
@@ -96,6 +96,9 @@ public static Error RedirectionDocumentIdConflict(IEnumerable<Document> redirect
         public static Error GitShadowClone(string repoPath)
             => new Error(ErrorLevel.Error, "git-shadow-clone", $"Does not support git shallow clone: '{repoPath}'");
 
+        public static Error GitNotFound()
+            => new Error(ErrorLevel.Error, "git-not-found", $"Cannot find git, install git https://git-scm.com/");
```
superyyrrzz: >Error [](start = 36, length = 5)

Warning? Some users may not need git features.

yufeih: when this is executed, it sure is invoking git exe and the user is using git features that depend on git exe

src/docfx/lib/git/GitUtility.cs
herohua: looks good!

---
In reply to: [200887867](https://github.com/dotnet/docfx/pull/3022#discussion_r200887867) [](ancestors = 200887867)

src/docfx/lib/git/GitUtility.cs
```diff
@@ -128,9 +129,14 @@ private static Task<string> ExecuteQuery(string cwd, string commandLineArgs, Tim
...
+            {
+                throw Errors.GitNotFound().ToException();
```
herohua: >ToException [](start = 43, length = 11)

put ex as inner exception?

PR #3027, opened 2018-07-10
Fix exception when redirect url only differs by case

yufeih:


src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -53,7 +53,7 @@ public static (List<Error> errors, RedirectionMap map) Create(Docset docset)
             AddRedirections(docset.Config.Redirections);
 
             var redirectionsByRedirectionUrl = redirections
-                .GroupBy(file => file.RedirectionUrl)
+                .GroupBy(file => file.RedirectionUrl, PathUtility.PathComparer)
```
OsmondJiang: >RedirectionUrl [](start = 38, length = 14)

the redirection url should be always case insenstive?

yufeih: URL is also using os case sensitivity. We are a static site generator, that means a URL maps directly to a file, including case sensitivity.

OsmondJiang: what if the url is absolute path?

---
In reply to: [201203096](https://github.com/dotnet/docfx/pull/3027#discussion_r201203096) [](ancestors = 201203096)

OsmondJiang: and most of the browser will lower case url

---
In reply to: [201205938](https://github.com/dotnet/docfx/pull/3027#discussion_r201205938) [](ancestors = 201205938,201203096)

yufeih: browser will not adjust URL casing.

OsmondJiang: the host name will be lower case

---
In reply to: [201206103](https://github.com/dotnet/docfx/pull/3027#discussion_r201206103) [](ancestors = 201206103)

PR #3030, opened 2018-07-10
fix missing warning in yaml header

superyyrrzz:


docs/specs/markdown.yml
```diff
@@ -17,6 +17,24 @@ outputs:
...
+# Report warning when Yaml header has null value
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    ---
+    title:
+    ---
+    hello
+outputs:
+  docs/a.json: |
```
yufeih: how about leave `docs/a.json: ` as empty, to just check build.log

superyyrrzz: sounds good

PR #3032, opened 2018-07-10
Errors from one file should not affect another file

yufeih:
If one file has errors, it should not prevent other files from publish.

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Docs.Build
             var context = new Context(report, outputPath);
             var docset = new Docset(context, docsetPath, config, options);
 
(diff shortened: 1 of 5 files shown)


src/docfx/build/Build.cs
```diff
@@ -68,27 +68,35 @@ bool ShouldBuildFile(Document file)
...
+                        return await BuildSchemaDocument.Build();
+                    case ContentType.TableOfContents:
+                        return await BuildTableOfContents.Build(context, file, tocMap, buildChild);
+                    case ContentType.Redirection:
+                        return await BuildRedirectionItem(context, file);
+                    default:
+                        return DependencyMap.Empty;
+                }
+            }
+            catch (DocfxException ex)
```
OsmondJiang: >DocfxException [](start = 19, length = 14)

which ex should be thrown and which ex whould be handled in their own pipeline? should we handle all errors in their own pipeline?

PR #3034, opened 2018-07-10
Add copyResources option to suppress resource copying

yufeih:
For docs publish, we don't need to copy the resources. This PR:

- renames _asset_ to _resource_
- add _copyResources_ config
- add _sourcePath_ to build.manifest

#2972

### DIFF
--- a/docs/designs/output.md
+++ b/docs/designs/output.md
@@ -128,10 +128,10 @@ These are called *browser navigation properties* and are stored in manifest.
 {
     "files": [
         {
-        "siteUrl": "/dotnet/api/system.string",
-        "outputPath": "en-us/netstandard-2.0/dotnet/api/system.string.json",
-        "locale": "en-us",
-        "moniker": "netstandard-2.0", 
+            "siteUrl": "/dotnet/api/system.string",
+            "outputPath": "en-us/netstandard-2.0/dotnet/api/system.string.json",
+            "locale": "en-us",
+            "moniker": "netstandard-2.0", 
 
             // other browser navigation properties...
             // other properties needed for backward compatibility
--- a/docs/designs/output.md
+++ b/docs/designs/output.md
@@ -139,7 +139,7 @@ These are called *browser navigation properties* and are stored in manifest.
         {
             "siteUrl": "/dotnet/api/system.string",
 
-            // when `outputPath` does not exists,
+            // when `outputPath` does not exist,
             // use `sourcePath` relative to source docset folder to locate the file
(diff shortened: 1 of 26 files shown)


docs/designs/output.md
```diff
@@ -127,14 +127,19 @@ These are called *browser navigation properties* and are stored in manifest.
...
         "outputPath": "en-us/netstandard-2.0/dotnet/api/system.string.json",
         "locale": "en-us",
         "moniker": "netstandard-2.0", 
 
         // other browser navigation properties...
         // other properties needed for backward compatibility
+    }, {
+        "siteUrl": "/dotnet/api/system.string",
+
+        // when `outputPath` does not exists,
```
superyyrrzz: >exists [](start = 38, length = 6)

exist

src/docfx/build/Build.cs
```diff
@@ -63,7 +63,13 @@ internal static class Build
 
                 bool ShouldBuildFile(Document file)
                 {
-                    return file.ContentType != ContentType.Unknown && fileListBuilder.TryAdd(file);
+                    if (file.ContentType == ContentType.Unknown)
+                        return false;
+                    if (!fileListBuilder.TryAdd(file))
+                        return false;
+                    if (file.ContentType == ContentType.Resource && !file.Docset.Config.Output.CopyResources)
+                        return false;
```
superyyrrzz: Will we build metadata for resource? If so, resource *should* be built when copyResources=true.

OsmondJiang: Move this logic to BuildResource?

docs/designs/output.md
```diff
@@ -127,14 +127,19 @@ These are called *browser navigation properties* and are stored in manifest.
...
     "files": [{
-        "source": "dotnet/api/system.string",
         "siteUrl": "/dotnet/api/system.string",
         "outputPath": "en-us/netstandard-2.0/dotnet/api/system.string.json",
         "locale": "en-us",
         "moniker": "netstandard-2.0", 
 
         // other browser navigation properties...
         // other properties needed for backward compatibility
+    }, {
```
OsmondJiang: >}, { [](start = 4, length = 4)

format?

src/docfx/build/Build.cs
yufeih: yeah, the only downside is progress, it will immediately jump from 30% to 100%

yufeih: we don't build resource metadata, it is just a copy, but it is a good point though, putting it here will cause other problems.

PR #3035, opened 2018-07-10
convert boolean value in page metadata to lower case

928PJY:


src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -122,14 +122,26 @@ public static JObject GenerataCommonMetadata(JObject metadata, Docset docset)
...
-                        ? string.Join(",", item.Value)
-                        : item.Value.ToString();
+                    string content;
+                    if (item.Value is JArray)
+                    {
+                        content = string.Join(",", item.Value);
+                    }
+                    else if (item.Value.Type == JTokenType.Boolean)
+                    {
+                        content = item.Value.ToString().ToLower(culture);
```
yufeih: item.Value<bool>() ? "true" : "false", this is culture independent

src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -122,14 +122,26 @@ public static JObject GenerataCommonMetadata(JObject metadata, Docset docset)
         public static string GenerateLegacyPageMetadata(JObject rawMetadata)
         {
             StringBuilder pageMetadataOutput = new StringBuilder(string.Empty);
+            var locale = rawMetadata["locale"].ToString();
+            var culture = new CultureInfo(string.IsNullOrEmpty(locale) ? "en-us" : locale);
```
superyyrrzz: >string.IsNullOrEmpty(locale) [](start = 42, length = 28)

This check is not necessary, as it's already ensured in when loading config.

Issue #2586, opened 2018-04-02
Logging, progress and telemetry


vicancy:
We need:

- [x] Report to show errors and warnings to end users (output to `stderr`)
- [x] Log to help us diagnostics issues (output to `stdout`)
- [x] Progress to show build status
- [ ] ~~Telemetry connected to application insights to track usage~~

@OsmondJiang @herohua

yufeih:
We need:

- [x] Report to show errors and warnings to end uers
- [ ] Log to help us diagnostics issues
- [ ] Progress to show build status
- [ ] Telemetry connected to application insights to track usage

PR #3036, opened 2018-07-10
Pipe git exe output to current process console.

yufeih:
#2586 #2987

### DIFF
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -16,15 +16,9 @@ namespace Microsoft.Docs.Build
     internal static class ProcessUtility
     {
         /// <summary>
-        /// Execute process with args
+        /// Start a new process and wait for its execution asynchroniously
         /// </summary>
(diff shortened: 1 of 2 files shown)


src/docfx/lib/ProcessUtility.cs
```diff
@@ -16,46 +16,41 @@ namespace Microsoft.Docs.Build
...
+            {
+                FileName = fileName,
+                WorkingDirectory = cwd,
+                Arguments = commandLineArgs,
+            };
+
+            if (redirectOutput)
+            {
+                psi.UseShellExecute = false;
+                psi.CreateNoWindow = true;
```
OsmondJiang: >psi.CreateNoWindow = true [](start = 16, length = 25)

should we always not create new windows ?

yufeih: it should always set UseShellExecute = false. CreateNoWindow = false will suppress the output

---
In reply to: [201551877](https://github.com/dotnet/docfx/pull/3036#discussion_r201551877) [](ancestors = 201551877)

PR #3037, opened 2018-07-10
Upgrade markdig

yufeih:
#2972


superyyrrzz:
@fenxu to aware of this.

PR #3043, opened 2018-07-11
Share a single MarkdownPipeline

yufeih:
Time to build _azure-docs-pr_ without _loading commits_, _run legacy_, _copy resource_ drops to 2 minutes.

#2972

### DIFF
--- a/src/docfx/build/markdown/Markup.cs
+++ b/src/docfx/build/markdown/Markup.cs
@@ -4,16 +4,11 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Runtime.CompilerServices;
 using Markdig;
 using Microsoft.DocAsCode.MarkdigEngine.Extensions;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
-    /// <summary>
-    /// Converts markdown to html
-    /// </summary>
     internal static class Markup
     {
         // In docfx 2, a localized text is prepended to quotes beginning with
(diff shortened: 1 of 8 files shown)


src/docfx/lib/markdown/ExtractTitle.cs
```diff
@@ -13,32 +12,27 @@ namespace Microsoft.Docs.Build
...
         private static string RenderTitle(HeadingBlock h1)
         {
-            using (var writer = new StringWriter())
-            {
-                var renderer = new HtmlRenderer(writer);
-                var pipeline = new MarkdownPipelineBuilder().Build();
-                pipeline.Setup(renderer);
-                renderer.Render(h1);
-                writer.Flush();
+            var writer = new StringWriter();
```
herohua: >StringWriter [](start = 29, length = 12)

dispose?

src/docfx/build/markdown/Markup.cs
```diff
@@ -32,91 +27,101 @@ internal static class Markup
             { "Caution", "<p>Caution</p>" },
         };
 
+        private static readonly MarkdownPipeline s_markdownPipeline = CreateMarkdownPipeline();
+
+        [ThreadStatic]
+        private static (MarkupResult result, DependencyMapBuilder dependencyMap, Action<Document> buildChild) t_context;
```
herohua: >t_context [](start = 110, length = 9)

do we have a better name? `t_context` and `Context` look confusing to me.

src/docfx/build/markdown/Markup.cs
```diff
@@ -32,91 +27,101 @@ internal static class Markup
...
+
+        public static ref MarkupResult Context => ref t_context.result;
```
live1206: >ref [](start = 22, length = 3)

just out of curiosity, why do we need "ref" here?

src/docfx/build/markdown/Markup.cs
yufeih: `t_` is the prefix for thread static fields as described [here](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md#c-coding-style)

src/docfx/lib/markdown/ExtractTitle.cs
yufeih: StringWriter dispose is an no-op

---
In reply to: [201658646](https://github.com/dotnet/docfx/pull/3043#discussion_r201658646) [](ancestors = 201658646)

PR #3052, opened 2018-07-11
Feature/json schema validate

live1206:
Make schema validation in place ready to use(2838)

### DIFF
--- a/src/docfx/lib/JsonUtility.cs
+++ b/src/docfx/lib/JsonUtility.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
@@ -28,7 +29,7 @@ namespace Microsoft.Docs.Build
             ContractResolver = new JsonContractResolver(),
         };
 
-        private static readonly IDictionary<Type, JSchema> s_typeSchemas = new Dictionary<Type, JSchema>();
+        private static readonly ConcurrentDictionary<Type, JSchema> s_typeSchemas = new ConcurrentDictionary<Type, JSchema>();
 
         private static readonly JsonMergeSettings s_defaultMergeSettings = new JsonMergeSettings
         {
@@ -164,7 +165,7 @@ namespace Microsoft.Docs.Build
             {
                 var generator = new JSchemaGenerator() { ContractResolver = DefaultDeserializer.ContractResolver };
                 var schema = generator.Generate(type, true);
-                s_typeSchemas.Add(type, schema);
+                s_typeSchemas.TryAdd(type, schema);
                 return schema;
             }
             return value;


src/docfx/Errors.cs
```diff
@@ -109,6 +109,12 @@ public static Error GitNotFound()
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} {name} contains null value");
 
+        public static Error JsonSchemaLicenseIssue(Exception ex)
+            => new Error(ErrorLevel.Warning, "json-schema-license", $"Encountered issue registering licenes for NewtonsoftJson.Schema: {ex.Message}");
```
herohua: >licenes [](start = 100, length = 7)

typo: license

src/docfx/lib/JsonUtility.cs
```diff
@@ -213,6 +237,25 @@ private static void LogWarningForNullValue(JToken item, List<Error> errors, JTok
             }
         }
 
+        private static List<Error> RegisterLicense(string license)
```
herohua: >List<Error> [](start = 23, length = 11)

just Error?

src/docfx/lib/JsonUtility.cs
```diff
@@ -135,6 +140,25 @@ public static (List<Error>, JToken) ValidateNullValue(this JToken token, JTokenS
             return (errors, token);
         }
 
+        public static List<Error> ValidateSchema(this JToken token, Type type, JTokenSourceMap mappings = null)
+        {
+            // TODO: Get the license somewhere
+            var registerErrors = RegisterLicense(string.Empty);
+            if (registerErrors.Any())
```
herohua: >.Any() [](start = 30, length = 6)

.Count > 0

src/docfx/lib/JsonUtility.cs
```diff
@@ -135,6 +140,25 @@ public static (List<Error>, JToken) ValidateNullValue(this JToken token, JTokenS
             return (errors, token);
         }
 
+        public static List<Error> ValidateSchema(this JToken token, Type type, JTokenSourceMap mappings = null)
```
herohua: >ValidateSchema [](start = 34, length = 14)

rename to `ValidateAgainstType`?

src/docfx/lib/JsonUtility.cs
```diff
@@ -135,6 +140,25 @@ public static (List<Error>, JToken) ValidateNullValue(this JToken token, JTokenS
...
+            var registerErrors = RegisterLicense(string.Empty);
+            if (registerErrors.Any())
+                return registerErrors;
+
+            var generator = new JSchemaGenerator() { ContractResolver = new CamelCasePropertyNamesContractResolver() };
+            var schema = generator.Generate(type, true);
+            var schemaString = schema.ToString();
+            if (!token.IsValid(schema, out IList<ValidationError> errors))
+            {
+                var error = errors.First();
```
herohua: >errors [](start = 28, length = 6)

output all errors?

src/docfx/lib/YamlUtility.cs
```diff
@@ -64,35 +64,37 @@ public static string ReadHeader(string yaml)
...
-            return Deserialize(new StringReader(input), nullValidation);
+            var (errors, _, token) = Deserialize(new StringReader(input), schemaValidation, nullValidation);
+            return (errors, token);
         }
 
         /// <summary>
         /// Deserialize to JToken from TextReader
         /// </summary>
-        public static (List<Error>, JToken) Deserialize(TextReader reader, bool nullValidation = true)
+        public static (List<Error>, JTokenSourceMap, JToken) Deserialize(TextReader reader, bool schemaValidation = false, bool nullValidation = true, Type type = null)
```
herohua: >schemaValidation [](start = 97, length = 16)

put this logic into a separate API

src/docfx/lib/JsonUtility.cs
live1206: Basically this method is used to call License.RegisterLicense, it could be void if we just throw the exception

---
In reply to: [201662287](https://github.com/dotnet/docfx/pull/3052#discussion_r201662287) [](ancestors = 201662287)

src/docfx/lib/JsonUtility.cs
live1206: I thought we could just throw the first error to use to fix for further information. You are right, we should output all the errors for later usage.

---
In reply to: [201662860](https://github.com/dotnet/docfx/pull/3052#discussion_r201662860) [](ancestors = 201662860)

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -332,6 +332,21 @@ public void TestListWithNullItem()
             });
         }
 
+        // TODO: this test should only be run with newton json schema license
+//        [Fact]
```
OsmondJiang: >Fact [](start = 11, length = 4)

[Fact(Skip = "this test should only be run with newton json schema license")]

OsmondJiang:
>         public Range(int startLine, int startCharacter, int endLine = 0, int endCharacter = 0)

-1? sometimes we may be some unexpected char at the 0,0

---
Refers to: src/docfx/models/Range.cs:13 in fd800aa. [](commit_id = fd800aa16367935a0b13f96385847c241e5a0649, deletion_comment = False)

src/docfx/lib/JsonUtility.cs
```diff
@@ -135,6 +140,25 @@ public static (List<Error>, JToken) ValidateNullValue(this JToken token, JTokenS
...
 
+        public static List<Error> ValidateSchema(this JToken token, Type type, JTokenSourceMap mappings = null)
+        {
+            // TODO: Get the license somewhere
+            var registerErrors = RegisterLicense(string.Empty);
+            if (registerErrors.Any())
+                return registerErrors;
+
+            var generator = new JSchemaGenerator() { ContractResolver = new CamelCasePropertyNamesContractResolver() };
+            var schema = generator.Generate(type, true);
```
OsmondJiang: >Generate [](start = 35, length = 8)

is this used reflection? can we cache the schema for the same type?

src/docfx/build/toc/parser/TableOfContentsInputItem.cs
```diff
@@ -4,11 +4,13 @@
...
 using System.Linq;
 using System.Web;
+using Newtonsoft.Json;
 
 namespace Microsoft.Docs.Build
 {
     internal class TableOfContentsInputItem
     {
+        [JsonProperty(Required = Required.DisallowNull)]
         public string Name { get; set; }
```
OsmondJiang: >Name [](start = 22, length = 4)

add a test?

live1206:
>         public Range(int startLine, int startCharacter, int endLine = 0, int endCharacter = 0)

we only have end line info from Yaml exceptions, not from Json

---
In reply to: [404368683](https://github.com/dotnet/docfx/pull/3052#issuecomment-404368683) [](ancestors = 404368683)

---
Refers to: src/docfx/lib/Range.cs:13 in fd800aa. [](commit_id = fd800aa16367935a0b13f96385847c241e5a0649, deletion_comment = False)

src/docfx/build/toc/parser/TableOfContentsInputItem.cs
live1206: The related test was the one you suggest to be skipped

---
In reply to: [201892945](https://github.com/dotnet/docfx/pull/3052#discussion_r201892945) [](ancestors = 201892945)

src/docfx/lib/JsonUtility.cs
```diff
@@ -213,6 +237,25 @@ private static void LogWarningForNullValue(JToken item, List<Error> errors, JTok
             }
         }
 
+        private static List<Error> RegisterLicense(string license)
+        {
+            var errors = new List<Error>();
```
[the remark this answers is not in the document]
live1206: True, it can only be one, got used to return a collection of errors everywhere

---
In reply to: [201892445](https://github.com/dotnet/docfx/pull/3052#discussion_r201892445) [](ancestors = 201892445)

src/docfx/lib/JsonUtility.cs
```diff
@@ -135,6 +140,25 @@ public static (List<Error>, JToken) ValidateNullValue(this JToken token, JTokenS
...
         }
 
+        public static List<Error> ValidateSchema(this JToken token, Type type, JTokenSourceMap mappings = null)
+        {
+            // TODO: Get the license somewhere
+            var registerErrors = RegisterLicense(string.Empty);
+            if (registerErrors.Any())
+                return registerErrors;
+
+            var generator = new JSchemaGenerator() { ContractResolver = new CamelCasePropertyNamesContractResolver() };
```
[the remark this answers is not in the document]
live1206: we only need ContactResolver here, not JsonSerializerSettings

---
In reply to: [201892620](https://github.com/dotnet/docfx/pull/3052#discussion_r201892620) [](ancestors = 201892620)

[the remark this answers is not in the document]
OsmondJiang: I mean why we always use CamelCase?

---
In reply to: [201894512](https://github.com/dotnet/docfx/pull/3052#discussion_r201894512) [](ancestors = 201894512,201892620)

src/docfx/lib/JsonUtility.cs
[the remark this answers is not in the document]
OsmondJiang: No need, it's based on your implementation and requirement.

---
In reply to: [201894021](https://github.com/dotnet/docfx/pull/3052#discussion_r201894021) [](ancestors = 201894021,201892445)

src/docfx/build/toc/parser/TableOfContentsInputItem.cs
OsmondJiang: ho I see, can it be a spec test?

---
In reply to: [201893698](https://github.com/dotnet/docfx/pull/3052#discussion_r201893698) [](ancestors = 201893698,201892945)

src/docfx/lib/JsonUtility.cs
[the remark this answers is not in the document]
live1206: based on the use cases?  

---
In reply to: [201895022](https://github.com/dotnet/docfx/pull/3052#discussion_r201895022) [](ancestors = 201895022,201894512,201892620)

src/docfx/lib/JsonUtility.cs
live1206: Actually the more common way is to load the schema from template file. But sure, we can cache the schemas for types.

---
In reply to: [201892746](https://github.com/dotnet/docfx/pull/3052#discussion_r201892746) [](ancestors = 201892746)

src/docfx/build/toc/parser/TableOfContentsInputItem.cs
live1206: do you mean a e2e test? but how could we skip a e2e test within .yml file?

---
In reply to: [201895244](https://github.com/dotnet/docfx/pull/3052#discussion_r201895244) [](ancestors = 201895244,201893698,201892945)

src/docfx/lib/JsonUtility.cs
```diff
@@ -135,6 +143,34 @@ public static (List<Error>, JToken) ValidateNullValue(this JToken token, JTokenS
             return (errors, token);
         }
 
+        public static IEnumerable<Error> ValidateSchemaAgainstType(this JToken token, Type type)
```
superyyrrzz: >ValidateSchemaAgainstType [](start = 41, length = 25)

Add a TODO to cache type?

We can add another method to accept JSON schema string later. e.g. for config, the user input and strong type model is different, as there's a conversion in between. Type doesn't work in this case.

live1206:
I had a brief look of the schema we have in v2, most of the schema validation features we could achieve with https://msdn.microsoft.com/en-us/library/system.componentmodel.dataannotations(v=vs.110).aspx

Not to mention type for properties, we can also use "RegularExpressionAttribute" for pattern matching. Beside those, if we want to define something like "minItems" of array, we could implement a custom validation attribute implementing "ValidationAttribute". Maybe this could cover "anyOf" as well, or we do not use it that much? 

Some concerns:
- Do we have a way to validate if a field content is markdown?
- Additional properties, if we want to disallow additional properties, seems not possible to do it with DataAnnotations? How about validating during traversing JToken like we did for null value check?

superyyrrzz:
> Do we have a way to validate if a field content is markdown

It enough to ensure it's string type.

live1206:
We will not use "Newtonsoft.Json.Schema" to validate Json schema

PR #3055, opened 2018-07-12
change process lock to file mutex to avoid conflicts with resource lock

OsmondJiang:


OsmondJiang:
#2987

src/docfx/config/AppData.cs
```diff
@@ -14,7 +14,7 @@ internal static class AppData
 
         public static string RestoreLockDir => Path.Combine(AppDataDir, "restore_lock");
 
-        public static string ProcessLockDir => Path.Combine(AppDataDir, "process_lock");
+        public static string FileMutexDir => Path.Combine(AppDataDir, "file_mutex");
```
yufeih: suggest to use `-` over `_` for file names to align with docs style

OsmondJiang: OK, i will also change the restore lock

---
In reply to: [201902110](https://github.com/dotnet/docfx/pull/3055#discussion_r201902110) [](ancestors = 201902110)

OsmondJiang: that's the folder name

---
In reply to: [201902554](https://github.com/dotnet/docfx/pull/3055#discussion_r201902554) [](ancestors = 201902554,201902110)

PR #3057, opened 2018-07-12
fix exception when move file in legacy

superyyrrzz:


src/docfx/build/legacy/LegacyUtility.cs
```diff
@@ -0,0 +1,19 @@
+using System.Diagnostics;
+using System.IO;
+
+namespace Microsoft.Docs.Build
+{
+    internal static class LegacyUtility
```
yufeih: we already have an `LegacyExtensions`, move the methods there to `LegacyUtility` and maybe make them non-extension methods?

superyyrrzz: Merged these into LegacyUtility

Issue #3060, opened 2018-07-12
remove build failed file from manifest


928PJY:
when the file build failed,  should not add it to manifest

PR #3064, opened 2018-07-12
Do not write output in cases of error

yufeih:
This PR:

- files with errors does not have output
- files with errors does not appear in manifest
- make build markdown pure by returning the model and errors

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -50,25 +50,21 @@ namespace Microsoft.Docs.Build
 
                 await ParallelUtility.ForEach(buildScope, BuildOneFile, ShouldBuildFile, Progress.Update);
 
-                var files = filesBuilder.Build(context).Except(filesWithErrors).OrderBy(file => file.FilePath).ToList();
+                var files = filesBuilder.Build(context).OrderBy(file => file.FilePath).Except(filesWithErrors).ToList();
                 var allDependencies = sourceDependencies.OrderBy(d => d.Key.FilePath).ToDictionary(k => k.Key, v => v.Value);
 
                 return (files, new DependencyMap(allDependencies));
 
                 async Task BuildOneFile(Document file, Action<Document> buildChild)
                 {
-                    var dependencyMap = await BuildFile(context, file, tocMap, contribution, buildChild);
-
-                    if (dependencyMap == null)
+                    var (hasError, dependencyMap) = await BuildFile(context, file, tocMap, contribution, buildChild);
+                    if (hasError)
                     {
                         filesWithErrors.Add(file);
                     }
-                    else
+                    foreach (var (source, dependencies) in dependencyMap)
                     {
-                        foreach (var (source, dependencies) in dependencyMap)
-                        {
-                            sourceDependencies.TryAdd(source, dependencies);
-                        }
+                        sourceDependencies.TryAdd(source, dependencies);
                     }
                 }
 
@@ -79,7 +75,7 @@ namespace Microsoft.Docs.Build
             }
         }
 
-        private static async Task<DependencyMap> BuildFile(
+        private static async Task<(bool hasError, DependencyMap)> BuildFile(
             Context context,
             Document file,
             TableOfContentsMap tocMap,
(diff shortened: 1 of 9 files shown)


src/docfx/build/Build.cs
```diff
@@ -45,25 +45,36 @@ internal static class Build
             using (Progress.Start("Building files"))
             {
                 var sourceDependencies = new ConcurrentDictionary<Document, List<DependencyItem>>();
-                var fileListBuilder = new DocumentListBuilder();
+                var filesBuilder = new DocumentListBuilder();
+                var filesWithErrors = new ConcurrentBag<Document>();
```
live1206: >ConcurrentBag [](start = 42, length = 13)

How about changing it to ConcurrentDictionary? ConcurrentBag might cause a memory leak here, since we only add items into it and use it for except, and never took them out afterwards. 
Please refer to [https://stackoverflow.com/questions/9256735/is-concurrentbag-cause-of-memory-leak](https://stackoverflow.com/questions/9256735/is-concurrentbag-cause-of-memory-leak)

src/docfx/build/Build.cs
```diff
@@ -77,56 +88,67 @@ bool ShouldBuildFile(Document file)
...
+                    return dependencies;
                 }
+
+                return null;
             }
             catch (DocfxException ex)
             {
                 context.Report(file.ToString(), ex.Error);
-                return DependencyMap.Empty;
+                return null;
```
928PJY: >return null; [](start = 16, length = 12)

here, you use `null` to represent a special status of `BuildFile` function, please add some return value description incase someone edit this in the future

yufeih: right, i should use a tuple instead.

src/docfx/build/Build.cs
```diff
@@ -77,56 +88,67 @@ bool ShouldBuildFile(Document file)
         {
             try
             {
+                var model = (object)null;
```
live1206: >var model = (object)null; [](start = 16, length = 25)

how about "PageModel model = null;"?

yufeih: is may not be a `PageModel`, build TOC returns a `TocModel`

src/docfx/build/Build.cs
928PJY: that would be better

---
In reply to: [202225993](https://github.com/dotnet/docfx/pull/3064#discussion_r202225993) [](ancestors = 202225993)

src/docfx/build/Build.cs
yufeih: `(object)null` is just to keep `var` visually aligned

live1206: I see

---
In reply to: [202227129](https://github.com/dotnet/docfx/pull/3064#discussion_r202227129) [](ancestors = 202227129)

src/docfx/build/Build.cs
```diff
@@ -45,30 +45,37 @@ internal static class Build
...
             {
                 var sourceDependencies = new ConcurrentDictionary<Document, List<DependencyItem>>();
-                var fileListBuilder = new DocumentListBuilder();
+                var filesBuilder = new DocumentListBuilder();
+                var filesWithErrors = new ConcurrentBag<Document>();
 
                 await ParallelUtility.ForEach(buildScope, BuildOneFile, ShouldBuildFile, Progress.Update);
 
-                return (fileListBuilder.Build(context), new DependencyMap(sourceDependencies.OrderBy(d => d.Key.FilePath).ToDictionary(k => k.Key, v => v.Value)));
+                var files = filesBuilder.Build(context).OrderBy(file => file.FilePath).Except(filesWithErrors).ToList();
```
928PJY: >Except(filesWithErrors) [](start = 87, length = 23)

what's the difference between those two

src/docfx/build/Build.cs
yufeih: Interesting. I believe it should be collected unless the bag references itself directly or indirectly: [https://github.com/dotnet/corefx/issues/23068#issuecomment-321315063](https://github.com/dotnet/corefx/issues/23068#issuecomment-321315063)

Enumerating a ConcurrentBag does have a tiny performance penalty because internally it creates a snapshot for each enumeration: [https://github.com/dotnet/corefx/issues/2335](https://github.com/dotnet/corefx/issues/2335), we've already used linq here so the performance does not matter that much.


---
In reply to: [202224709](https://github.com/dotnet/docfx/pull/3064#discussion_r202224709) [](ancestors = 202224709)

928PJY:
#3060

PR #3072, opened 2018-07-13
Make JsonContractResolver inherit CamelCasePropertyNamesContractResolver

live1206:
Although newton json deserialization is case insensitive, make JsonContractResolver inherit CamelCasePropertyNamesContractResolver to reuse them for serialization setting


yufeih:
If it's hard to be case sensitive, let's leave it for now.

PR #3075, opened 2018-07-14
Update intro_overwrite_files.md

superyyrrzz:


Documentation/tutorial/intro_overwrite_files.md
```diff
@@ -58,6 +58,10 @@ footer: <p>Footer for <code>microsoft.com/docfx/Contacts</code></p>
 
 `uid` for an *Overwrite Model* stands for the Unique IDentifier of the *Model* it will overwrite. So it is allowed to have multiple *Overwrite Section*s with YAML Header containing the same `uid`. For one *Overwrite File*, the latter *Overwrite Section* overwrites the former one with the same `uid`. For different *Overwrite File*s, the order of overwrite is **Undetermined**. So it is suggested to have *Overwrite Sections* with the same `uid` in the same *Overwrite File*.
 
+> [!NOTE]
+>
+> Multiple *Overwrite Section*s in one file doesn't work in markdig markdown engine. You should remove `"markdownEngineName": "markdig",` from `docfx.json` to support this feature.
```
OsmondJiang: it would be more clear if we can throw a error message like that

superyyrrzz: Yes, it is. Will not add it now as extra specific parse logic is needed.

Issue #3050, opened 2018-07-11
port SDP to v3


herohua:
- [x] deserialize yaml/json into c# model with error handling
- [x] generate page model
- [x] model `LandingData`
- [x] ~~run jint and mustache in legacy mode~~
- [x] CI to generate JSON schema from c# model
- [x] support data annotation
- [x] support markdown, inline markdown
- [ ] support xref/uid
- [x] potential merge with markdown pipeline
- [ ] support external xref
- [x] filter out files referenced by config
- [x] export description to json schema
- [x] support JsonExtensionData and JToken in create json schema
- [x] determine output type (page or json) based on DataSchema or PageSchema ~~whether template (cshtml) exists~~

yufeih:
index.yml page is not using SDP, it also depend on Mustache

yufeih:
SDP in v3 is designed to work with razor templates, here is the transformation process:

```
        deserialize         razor
JSON/YAML ------> C# class  ------> Html

```

There are the goals for the de-serialization process:

- Report errors and warnings with precise line info
    - Report syntax errors with line info
    - Preserve line info from `JToken.Parse` and `YamlDocument.Load`
    - Report and remove null values with line info
    - Handle markdown embedded inside YAML
    - Handle YAML header inside markdown
    - Propagate line info into data validation library
    - Preserve line info when calling `JToken.ToObject` to validate:
           - Property type mismatches
           - JSON/YAML keys that does not exit in c# model
- ~~Support case sensitive member names~~

PR #3078, opened 2018-07-16
Rebuild yaml line info mapping based on the key of structure path

live1206:
Rebuild Yaml line info mapping based on the key of structure path, fix tests accordingly.

#3050


src/docfx/lib/YamlUtility.cs
```diff
@@ -166,35 +166,39 @@ private static JToken ToJson(YamlNode node, JTokenSourceMap mappings = null)
                 {
                     if (key is YamlScalarNode scalarKey)
                     {
-                        var jToken = ToJson(value, mappings);
+                        var path = string.Join('|', name, key.ToString());
+                        SetMappings(mappings, key, path);
```
yufeih: how do we put these line info to `JToken` if it is a separate dictionary?

live1206: For flat types, we can just use "|propertyName". For nested types during the custom validation we can build the structure separated with "|" as the mapping key

Issue #3063, opened 2018-07-12
dependency-repo-not-found is reported for all files


yufeih:
Run `docfx build` on a repo with dependency without run `docfx restore`

**Expected**

`dependency-repo-not-found` should only throw once and stop the build.

**Actual**

![image](https://user-images.githubusercontent.com/511355/42618946-1befba9a-85e9-11e8-8e1a-a0584104a39d.png)

OsmondJiang:
@yufeih , the original design is the docfx build will stop once met any error belongs to any file, if we continue to run the docfx, we need change some Lazy load to immediate load

PR #3080, opened 2018-07-16
directly load dependency repo and throw error to abort the whole proc…

OsmondJiang:
…ess if there are any wrong with the restores


OsmondJiang:
#3063

PR #3083, opened 2018-07-16
improve error msg when for empty uid in udp

superyyrrzz:


src/Microsoft.DocAsCode.Build.UniversalReference/UniversalReferenceDocumentProcessor.cs
```diff
@@ -36,6 +36,10 @@ protected override FileModel LoadArticle(FileAndType file, ImmutableDictionary<s
             {
                 throw new ArgumentException($"{nameof(ItemViewModel.SupportedLanguages)} must contain at least one language");
             }
+            if (page.Items.Any(i => string.IsNullOrEmpty(i.Uid)))
+            {
+                throw new ArgumentException($"{nameof(ItemViewModel.Uid)} must not be null or empty");
```
herohua: would it be better to give more info on what node is empty, such as the other properties in `page.Items` or line number?

PR #3084, opened 2018-07-17
disable validation when it is redirection file

928PJY:


src/Microsoft.DocAsCode.Build.Common/RedirectionFileHelper.cs
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Build.Common
+{
+    using System;
+    using System.Text.RegularExpressions;
+
+    public class RedirectionFileHelper
+    {
+        public static readonly Regex RedirectionRegex = new Regex(@"^\-{3}\s*(.+\:.*\s)*(redirect_url\:.*\s)(.+\:.*\s)*\s*\-{3}\s$", RegexOptions.Compiled | RegexOptions.Multiline, TimeSpan.FromSeconds(10));
```
superyyrrzz: >redirect_url [](start = 89, length = 12)

do not put redirect_url into docfx?

src/Microsoft.DocAsCode.Build.Common/RedirectionFileHelper.cs
```diff
@@ -0,0 +1,19 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.DocAsCode.Build.Common
+{
+    using System;
+    using System.Text.RegularExpressions;
+
+    public class RedirectionFileHelper
```
herohua: This is not a correct implementation. As we discussed, please talk to validation library and see whether they can embed this logic.

PR #3085, opened 2018-07-17
accept token for docfx restore

OsmondJiang:


src/docfx/lib/git/GitUtility.cs
```diff
@@ -35,6 +35,42 @@ public static (string url, string branch) GetGitRemoteInfo(string remoteHref)
...
+            await action(remoteUrlWithToken);
+
+            // reset url back without token
+            await SetRemoteGitUrl(cwd, remoteUrl);
+        }
+
+        /// <summary>
+        /// Append token to git remote href
+        /// </summary>
+        public static string AppendToken(string remoteHref, string token)
```
superyyrrzz: >AppendToken [](start = 29, length = 11)

EmbedToken?

src/docfx/lib/git/GitUtility.cs
```diff
@@ -35,6 +35,42 @@ public static (string url, string branch) GetGitRemoteInfo(string remoteHref)
...
+                return;
+            }
+
+            // add token to remote href
+            var remoteUrlWithToken = AppendToken(remoteUrl, token);
+
+            await action(remoteUrlWithToken);
+
+            // reset url back without token
+            await SetRemoteGitUrl(cwd, remoteUrl);
```
superyyrrzz: >await SetRemoteGitUrl(cwd, remoteUrl); [](start = 12, length = 38)

Is this only needed by `clone`? I see `fetch` doesn't use this. If so, can we put this into `GitUtility.Clone`?

OsmondJiang: pull also need, it's just a wrapper for any action and i don't want to git clone itself

---
In reply to: [202904586](https://github.com/dotnet/docfx/pull/3085#discussion_r202904586) [](ancestors = 202904586)

src/docfx/lib/git/GitUtility.cs
OsmondJiang: good suggestion

---
In reply to: [202904142](https://github.com/dotnet/docfx/pull/3085#discussion_r202904142) [](ancestors = 202904142)

src/docfx/cli/Program.cs
```diff
@@ -84,8 +83,9 @@ private static (string command, string docset, CommandLineOptions options) Parse
             ArgumentSyntax.Parse(args, syntax =>
             {
                 // Restore command
-                // usage: docfx restore [docset]
+                // usage: docfx restore [docset] [-t/--token token]
                 syntax.DefineCommand("restore", ref command, "Restores dependencies before build.");
+                syntax.DefineOption("t|token", ref options.Token, "The token used to restore dependency repositories");
```
yufeih: consider `--gitToken`, we could have other token types, also remove `t|`, it does not deserve a short hand.

src/docfx/lib/git/GitUtility.cs
superyyrrzz: Got it. Personally, I prefer to hide the logic inside clone/pull, with an extra optional parameter passing the token.

yufeih: second @superyyrrzz

OsmondJiang: With token logic will have an extra action to reset the url back, my purpose is to let the user choice, not a default logic hided inside clone/pull

---
In reply to: [202929623](https://github.com/dotnet/docfx/pull/3085#discussion_r202929623) [](ancestors = 202929623)

src/docfx/lib/git/GitUtility.cs
```diff
@@ -97,40 +133,8 @@ public static Task Clone(string cwd, string remote, string path, string branch =
...
-        /// <param name="branch">The remote branch name</param>
-        /// <returns>Task status</returns>
-        public static Task Reset(string cwd, string branch)
-        {
-            Debug.Assert(!string.IsNullOrEmpty(branch));
-
-            return ExecuteNonQuery(cwd, $"reset --hard origin/{branch}");
-        }
+        public static Task Fetch(string cwd, string remoteWitToken)
+            => ExecuteNonQuery(cwd, $"fetch {remoteWitToken} '*:*'");
```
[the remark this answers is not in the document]
OsmondJiang: the fetch with remoteWithToken will not change the repo's remote url, so no need to execute the reset remote url action

---
In reply to: [202917515](https://github.com/dotnet/docfx/pull/3085#discussion_r202917515) [](ancestors = 202917515)

src/docfx/lib/git/GitUtility.cs
superyyrrzz: For now the action == reset_url, so I think the clone/pull method itself already knows what action to take, thus to reduce user's effort to choose an action.

fenxu:
@OsmondJiang Before we use Razor, we need a token to clone themes repo which is on GitHub. However, if the repo is Vso, you can't use same token. Shall we consider to use a token list or just separate them? Otherwise, one token can't meet current situation for Vso.

yufeih:
It is essentially a map of URL to token as [described here](https://git-scm.com/docs/gitcredentials#_credential_contexts). Can we use [$XDG_CONFIG_HOME](https://git-scm.com/docs/git-credential-store#git-credential-store-XDGCONFIGHOMEgitcredentials) to specify the config location?

OsmondJiang:
Right now we don't have a case that the crr repos are mixed github repo and vso repo, the restore token actually is the token of restored repo, we can do that later until we have that case?

PR #3092, opened 2018-07-17
apply routes to resource

928PJY:
#3091


src/docfx/build/Build.cs
```diff
@@ -129,7 +129,7 @@ private static void BuildResource(Context context, Document file)
 
             if (file.Docset.Config.Output.CopyResources)
             {
-                context.Copy(file, file.FilePath);
+                context.Copy(file, Document.ApplyRoutes(file.FilePath, file.Docset.Config.Routes));
```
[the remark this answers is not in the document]
yufeih: also can we have a test case in config.yml

docs/specs/config.yml
```diff
@@ -167,20 +173,28 @@ inputs:
...
   outputB/b.json:
+  outputC/c.gif:
   build.manifest:
 ---
 # latter routes rule takes precedence - 2
 inputs:
   docfx.yml: |
     routes:
       docs/a/b.md: outputB
+      docs/a/c.gif: outputC
```
yufeih: remove this line, to check `docs/a` should include `docs/a/c.gif`

src/docfx/config/RouteConfig.cs
```diff
@@ -28,6 +28,11 @@ public string GetOutputPath(string path)
         {
             Debug.Assert(!string.IsNullOrEmpty(path));
 
+            if (Source == "./")
```
yufeih: is this covered in test?

docs/specs/config.yml
928PJY: this line doesn't influent the check, but can also check the routes priority in resource

---
In reply to: [203285427](https://github.com/dotnet/docfx/pull/3092#discussion_r203285427) [](ancestors = 203285427)

PR #3094, opened 2018-07-18
Do not mess up with global appdata during test

yufeih:
- Set DOCFX_APPDATA_PATH to a local folder during test
- Do not append .docfx is DOCFX_APPDATA_PATH is set from environment variable
- Rename some test output directorys

### DIFF
--- a/test/docfx.Test/common/TestFramework.cs
+++ b/test/docfx.Test/common/TestFramework.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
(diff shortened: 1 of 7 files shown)


test/docfx.Test/common/TestFramework.cs
```diff
@@ -25,6 +26,7 @@ public TestFramework(IMessageSink messageSink)
 
         protected override ITestFrameworkExecutor CreateExecutor(AssemblyName assemblyName)
         {
+            Environment.SetEnvironmentVariable("DOCFX_APPDATA_PATH", Path.GetFullPath("appdata"));
```
superyyrrzz: >SetEnvironmentVariable [](start = 24, length = 22)

What about fetch the original value first, and set it back when all tests finish?

yufeih: This method sets the process wide environment variable, it does not affect your machine and other processes.

---
In reply to: [203579305](https://github.com/dotnet/docfx/pull/3094#discussion_r203579305) [](ancestors = 203579305)

Issue #3077, opened 2018-07-16
Config should not require a trailing / to indicate folders


yufeih:
This config does not work for azure because _articles_ does not have a trailing `/`. This is not intuitive for most people. We should not have this constraint.

Don't work:
```yaml
routes:
   articles: azure
```

Works

```yaml
routes:
   articles/: azure
```

superyyrrzz:
Agree it's not intuitive to config... while I also don't want to make the meaning of the config dependent on the folder structure of the repo. I'd like to know what the config means by only looking at it.

I think we can also support 2 forms:

- `articles: azure`: this targets a file by default. It can fallback to a folder if file doesn't exist and folder exists.
- `articles: azure/`: this targets a folder for sure.

Issue #2821, opened 2018-05-29
generate document_id, document_version_independent_id 


herohua:

OsmondJiang:
- [ ] Ideal Output with document id
- [ ] Legacy Output with document id

PR #3095, opened 2018-07-18
support depot mapping and directory mapping for document id generation

OsmondJiang:


OsmondJiang:
#2821

docs/specs/metadata.yml
```diff
@@ -99,3 +99,31 @@ outputs:
...
     ["warning","redirected-id-conflict","Multiple documents redirected to '/sccm/mdm/understand/hybrid-mobile-device-management' with document id: 'sccm/mdm/index.md', 'sccm/mdm/index2.md'","docfx.yml"]
+---
+# generate document id with depot mapping and directory mapping
+# id = hash({mapped_depot_name}|{mapped_file_in_docset})
+# version_independent_id = hash({mapped_depot_name}|{file_path})
+inputs:
+  docfx.yml: |
+    documentId:
+      siteBasePath: azure
+      depotMappings:
```
yufeih: can we keep the original `_op_documentIdPathDepotMapping` syntax? `file -> path + depot`

src/docfx/config/DocumentIdConfig.cs
```diff
@@ -1,10 +1,17 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
 namespace Microsoft.Docs.Build
 {
     internal class DocumentIdConfig
     {
+        private readonly Lazy<Dictionary<string, string>> _reversedDepotMappings;
+        private readonly Lazy<Dictionary<string, string>> _reversedDirectoryMappings;
```
yufeih: nit: put these computed values to `Docset` to align with other computed properties like _Redirections_, _RestoreMap_

src/docfx/config/DocumentIdConfig.cs
```diff
@@ -14,5 +21,71 @@ internal class DocumentIdConfig
...
+                : normalizedFilePathToSourceBasePath;
+
+            return (depotName, mappedPathRelativeToSourceBasePath);
+        }
+
+        private static (string value, string matchedDirectory) GetReversedMapping(Dictionary<string, string> mappings, string normalizedFilePathToSourceBasePath)
+        {
+            foreach (var (path, value) in mappings)
+            {
+                if (string.Equals(path, normalizedFilePathToSourceBasePath, PathUtility.PathComparison))
```
yufeih: we probably need a `PathUtility.PathMatches` method to detect if a path is inside a folder or is the same path. `routes` also need this.

yufeih: Also, it is proved problematic to indicate folders using trailing slash, https://github.com/dotnet/docfx/issues/3077.

docs/specs/metadata.yml
OsmondJiang: the original syntax is hard to understand and a little confused. and also it contains a lot of duplications. I prefer to splite them and reverse them

---
In reply to: [203581517](https://github.com/dotnet/docfx/pull/3095#discussion_r203581517) [](ancestors = 203581517)

src/docfx/config/DocumentIdConfig.cs
OsmondJiang: I followed the RouteConfig which only want to provide two methods for document id config.

---
In reply to: [203581997](https://github.com/dotnet/docfx/pull/3095#discussion_r203581997) [](ancestors = 203581997)

src/docfx/config/DocumentIdConfig.cs
OsmondJiang: OK, i will add one.

---
In reply to: [203582744](https://github.com/dotnet/docfx/pull/3095#discussion_r203582744) [](ancestors = 203582744)

OsmondJiang: Let's resolve this issue together, for this PR, i follow the original rules, folder need end with '/'

---
In reply to: [203583022](https://github.com/dotnet/docfx/pull/3095#discussion_r203583022) [](ancestors = 203583022)

src/docfx/config/RouteConfig.cs
```diff
@@ -28,20 +27,19 @@ public string GetOutputPath(string path)
         {
             Debug.Assert(!string.IsNullOrEmpty(path));
 
-            if (Source == "./")
+            if (PathUtility.Match(path, Source))
             {
-                return Path.Combine(Destination, path);
-            }
+                if (Source == "./")
```
yufeih: put "./" check under `PathUtility.Match`

add UT

`PathUtility.Match` can return `(bool matches, bool isFile, string remainingPath)` like:

- `Match("a", "a") -> (true, true, "")`
- `Match("a/b", "a/") -> (true, false, "b")`

PR #3099, opened 2018-07-19
Include environment variables starting with DOCFX_ in crash report

yufeih:
Now that docfx depend on environment variables


src/docfx/cli/Program.cs
```diff
@@ -156,9 +158,31 @@ private static void PrintFatalErrorMessage(Exception exception)
             Console.ResetColor();
         }
 
+        private static string GetDocfxEnvironmentVariables()
+        {
+            try
+            {
+                return string.Concat(
+                from entry in Environment.GetEnvironmentVariables().Cast<DictionaryEntry>()
```
superyyrrzz: indent these lines?

PR #3100, opened 2018-07-19
Feature/yaml line info investigation

live1206:
- Populate output JToken with Yaml deserialization line info
- Propagate the line info above into null value error
- Add unit tests for line info


src/docfx/lib/JsonUtility.cs
```diff
@@ -187,30 +186,21 @@ private static void Traverse(this JToken token, List<Error> errors, JTokenSource
...
                     {
-                        prop.Value.Traverse(errors, mappings, nullNodes, prop.Name);
+                        prop.Value.Traverse(errors, nullNodes, prop.Name);
                     }
                 }
             }
         }
 
-        private static void LogWarningForNullValue(JToken item, List<Error> errors, JTokenSourceMap mappings, string name)
+        private static void LogInfoForNullValue(JToken item, List<Error> errors, string name)
```
yufeih: >JToken [](start = 48, length = 6)

use `IJsonLineInfo`, the `as` line below can be removed

src/docfx/lib/YamlUtility.cs
```diff
@@ -103,18 +105,19 @@ public static (List<Error>, JToken) Deserialize(TextReader reader, bool nullVali
...
-                var (nullErrors, token) = ToJson(stream.Documents[0].RootNode, mappings).ValidateNullValue(mappings);
+                var root = stream.Documents[0].RootNode;
+                var (nullErrors, token) = PopulateLineInfoToJToken(ToJson(root), root).ValidateNullValue();
                 errors.AddRange(nullErrors);
                 return (errors, token);
             }
             else
             {
-                return (errors, ToJson(stream.Documents[0].RootNode));
+                var token = ToJson(stream.Documents[0].RootNode);
```
yufeih: populate line info even without null validation?

src/docfx/lib/YamlUtility.cs
```diff
@@ -150,35 +153,38 @@ private static JToken ToJson(YamlNode node, JTokenSourceMap mappings = null)
...
 
-        private static JToken SetMappings(JTokenSourceMap mappings, YamlNode scalar, JToken value)
+        private static JToken PopulateLineInfoToJToken(JToken token, YamlNode node)
         {
-            if (mappings == null)
-                return value;
-
-            mappings.Add(value, new Range(scalar.Start.Line, scalar.Start.Column));
-            return value;
+            var m = typeof(JToken).GetMethod("SetLineInfo", BindingFlags.NonPublic | BindingFlags.Instance, null, new[] { typeof(int), typeof(int) }, null);
```
yufeih: s_setLineInfo can be a static memeber

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -328,10 +330,34 @@ public void TestListWithNullItem()
             {
                 Assert.Equal(ErrorLevel.Info, error.Level);
                 Assert.Equal("null-value", error.Code);
```
yufeih: can we have a Theory based test for both JSON and YAML? It should cover both value/array/and object

src/docfx/lib/YamlUtility.cs
```diff
@@ -130,18 +133,18 @@ private static JToken ToJson(YamlNode node, JTokenSourceMap mappings = null)
...
                     }
                     if (double.TryParse(scalar.Value, out var d))
                     {
-                        return SetMappings(mappings, scalar, new JValue(d));
+                        return new JValue(d);
                     }
                     if (bool.TryParse(scalar.Value, out var b))
                     {
-                        return SetMappings(mappings, scalar, new JValue(b));
+                        return new JValue(b);
```
yufeih: lineinfo should be set to every token including JValue

src/docfx/lib/YamlUtility.cs
```diff
@@ -150,35 +153,38 @@ private static JToken ToJson(YamlNode node, JTokenSourceMap mappings = null)
                 {
                     if (key is YamlScalarNode scalarKey)
                     {
-                        var jToken = ToJson(value, mappings);
-                        obj[scalarKey.Value] = jToken;
+                        var token = ToJson(value);
+                        if (token != null)
+                        {
+                            token = PopulateLineInfoToJToken(token, value);
```
yufeih: nit: you could put null check inside `PopulateLineInfo`

src/docfx/lib/YamlUtility.cs
live1206: Because we are gonna need the line info for null value error, so we should prepare line info before that

---
In reply to: [203930545](https://github.com/dotnet/docfx/pull/3100#discussion_r203930545) [](ancestors = 203930545)

src/docfx/lib/YamlUtility.cs
live1206: JValue is taken care of, when it returns to JObject call

---
In reply to: [203930758](https://github.com/dotnet/docfx/pull/3100#discussion_r203930758) [](ancestors = 203930758)

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -1,13 +1,13 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Newtonsoft.Json;
```
superyyrrzz: put 3rd party ns after system ns

live1206:
#3050

Issue #3049, opened 2018-07-11
incremental publish improvement: keep HASH unchanged between branches


herohua:
- ~~[ ] docfx put repo/branch to manifest~~
- ~~[x] docfx outputs git url templates~~
- ~~[ ] build service uploads document metadata~~
- ~~[ ] hosting exposes these metadata to template~~
- ~~[ ] template expand git url templates~~

yufeih:
This seems to be [solved in v2](https://github.com/dotnet/docfx/commit/89fae9485cd856bcf9c8e7d4a34b04e091b5a7ea) using the `DOCFX_SOURCE_BRANCH_NAME` environment variable.

@fenxu , can we set an environment variable like `DOCFX_SOURCE_BRANCH_NAME` to the target branch name during PR build?

fenxu:
@yufeih I prefer to use a command parameter other than environment variable.

fenxu:
Besides, it should be a full source repository url. Branch name is not enough.

yufeih:
Here is the solution we came up with from offline discussion:
- move branch, repo from output JSON file to manifest
- publish every document with branch, repo embeded as DHS document metadata
- publish every output JSON file using incremental publish
- turn editUrl, gitContentUrl, gitCommitUrl into string template

yufeih:
This needs further discussion. right now I'm putting repo/branch back to git urls, but does not allow overrides. This means for PR, the edit link always points to the PR target branch.

yufeih:
If we suppress loading git commits on review, `update_at` will fallback to file LastModified date which is un-deterministic.

OsmondJiang:
@yufeih , do we need this issue anymore?

yufeih:
yes

yufeih:
After #3780, we could leverage the input change list to filter determine edit button target repo.

yufeih:
Edit button is tracked by #3791 , all the other outputs produces consistent hash, so close this one.

PR #3103, opened 2018-07-20
change git urls back to include repo/branch

yufeih:
Let's only override repo/branch for changed files. This PR revert git urls to include repo/branch.

#3049

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -96,41 +96,25 @@ outputs:
   build.manifest:
   dep.json:
 ---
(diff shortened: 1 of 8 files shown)


docs/specs/git.yml
```diff
@@ -96,50 +96,34 @@ outputs:
   build.manifest:
   dep.json:
 ---
-# Show editUrl, contentUrl and commitUrl as template, manifest contains repo info
```
OsmondJiang: let's just revert the change but hold on the `change list` idea. :)

PR #3105, opened 2018-07-20
replace is-absolute-href with is-http-href

OsmondJiang:


src/docfx/restore/RestoreMap.cs
```diff
@@ -53,7 +53,7 @@ public string GetUrlRestorePath(string docsetPath, string path)
         {
             Debug.Assert(!string.IsNullOrEmpty(path));
 
-            if (!HrefUtility.IsAbsoluteHref(path))
+            if (!HrefUtility.IsHttpHref(path))
```
yufeih: nit: IsHttp?

src/docfx/lib/HrefUtility.cs
```diff
@@ -66,6 +66,10 @@ public static bool IsAbsoluteHref(string str)
                 || Uri.TryCreate(str, UriKind.Absolute, out _);
         }
 
+        public static bool IsHttpHref(string str)
+            => !string.IsNullOrEmpty(str) && Uri.TryCreate(str, UriKind.Absolute, out var uriResult)
```
yufeih: nit: `Startswith("http://")  || StartsWith("https://")` ?

is "https://///a.com" considered http?

src/docfx/restore/RestoreMap.cs
OsmondJiang: consistent with IsAbsoluteHref?

---
In reply to: [203931119](https://github.com/dotnet/docfx/pull/3105#discussion_r203931119) [](ancestors = 203931119)

src/docfx/lib/HrefUtility.cs
OsmondJiang: StartWith will not work for your case. https://///a.com is not http

---
In reply to: [203931273](https://github.com/dotnet/docfx/pull/3105#discussion_r203931273) [](ancestors = 203931273)

OsmondJiang: will add test for your case

---
In reply to: [203931759](https://github.com/dotnet/docfx/pull/3105#discussion_r203931759) [](ancestors = 203931759,203931273)

PR #3113, opened 2018-07-26
consolidate sdp + markdown into BuildPage

yufeih:
This PR merges sdp and markdown into a single `BuildPage` pipeline:

- `isMasterContent` -> `ContentType.Page`
- `titleHtml` -> `htmlTitle`
- `tocRelativePath` -> `toc` // pending spec test

#3050


src/docfx/docset/Document.cs
```diff
@@ -241,40 +233,31 @@ public static Document TryCreateFromFile(Docset docset, string path)
 
         internal static ContentType GetContentType(string path)
         {
-            var name = Path.GetFileNameWithoutExtension(path);
-
-            if (path.EndsWith(".md", PathUtility.PathComparison))
+            if (!path.EndsWith(".md", PathUtility.PathComparison) &&
+                !path.EndsWith(".json", PathUtility.PathComparison) &&
+                !path.EndsWith(".yml", PathUtility.PathComparison))
             {
```
OsmondJiang: exclude config file and extend file from here?

yufeih: that is a another workitem tracked in #3050

src/docfx/build/page/BuildPage.cs
```diff
@@ -0,0 +1,148 @@
...
+            var metadata = JsonUtility.Merge(Metadata.GetFromConfig(file), fileMetadata);
+            var (id, versionIndependentId) = file.Docset.Redirections.TryGetDocumentId(file, out var docId) ? docId : file.Id;
+
+            // TODO: add check before to avoid case failure
+            var (repoErrors, author, contributors, updatedAt) = contribution.GetContributorInfo(
+                file,
+                metadata.Value<string>("author"),
+                metadata.Value<DateTime?>("update_date"));
+
+            var (editUrl, contentUrl, commitUrl) = contribution.GetGitUrls(file);
```
superyyrrzz: >contribution.GetGitUrls(file); [](start = 51, length = 30)

the logic to generate these metadata is different between conceptual and sdp. See [https://ceapex.visualstudio.com/Engineering/_workitems/edit/29187](https://ceapex.visualstudio.com/Engineering/_workitems/edit/29187) There will be more diversity like this, e.g. Markdown fragement. Still keep md and sdp separate?

yufeih: for legacy format they are different, in v3 we could consolidate them

PR #3114, opened 2018-07-26
Add showContributors config to suppress loading git commits

yufeih:
#2972 

renames `enableContribution` / `contribution.enable` to `showEdit` to align with `showContributors`.

@superyyrrzz , we were using `update_date` to populate `UpdatedAt`, but most of articles on docs uses `ms.date`, which should be use? It seems `ms.date` has no effect even though it is spread out everywhere.


docs/specs/git.yml
```diff
@@ -129,6 +129,24 @@ outputs:
...
+      showContributors: false
+  docs/a.md: |
+    ---
+    ms.date: 07/03/2018
+    ---
+outputs:
+  docs/a.json: |
+    {
+      "contributors": [],
+      "!commitUrl": "",
```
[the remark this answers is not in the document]
yufeih: this is, the intention is to not loading git commits entirely, it affects anything that depend on git commit. the switch name is a little confusing, I am thinking about using `useGitCommit: false`

src/docfx/build/metadata/ContributionInfo.cs
```diff
@@ -115,14 +108,15 @@ public static ContributionInfo Load(Docset docset)
...
+            var editUrl = document.Docset.Config.Contribution.ShowEdit
                 ? $"https://github.com/{editRepo}/blob/{editBranch}/{pathToRepo}"
                 : null;
 
-            var commitUrl = _commitsByFile.TryGetValue(document.FilePath, out var commits) && commits.Count > 0
-                ? $"https://github.com/{repo.Name}/blob/{commits[0].Sha}/{pathToRepo}"
-                : null;
+            var commit = _commitsByFile.TryGetValue(document.FilePath, out var commits) && commits.Count > 0
+                ? commits[0].Sha
+                : repo.Commit;
```
superyyrrzz: >: repo.Commit [](start = 16, length = 13)

Is this for uncommitted files? Prefer leave it null, as the link will be 404 in this case.

yufeih: This is for review.docs when git commits is diabled, we cannot get the last changed commit, but still can fallback to repo commit to help us diagnose

---
In reply to: [205657741](https://github.com/dotnet/docfx/pull/3114#discussion_r205657741) [](ancestors = 205657741)

superyyrrzz: Got your ideas. I feel it like a side effect to let `showContributors` control `LoadCommits`. What about let template to consume `showContributors` and `showEdit`, and make they don't affect build logic?

---
In reply to: [205660235](https://github.com/dotnet/docfx/pull/3114#discussion_r205660235) [](ancestors = 205660235,205657741)

yufeih: The idea is to have a switch to turn off git commit calculation during build to speed things up on review. This PR uses `showContributors` as the switch, which is the most obvious reasons to load git commits. Unfortunately it might also affects `update_at` and `git_url`. 

We could use another name for this purpose if there is a strong disagreement

PR #3118, opened 2018-07-27
fix: refine regex for tables and add timeout

herohua:
@vwxyzh @superyyrrzz


superyyrrzz:
add a test to dfmTest.cs?

PR #3120, opened 2018-07-27
make path default value . instead of String.Empty

928PJY:
[System.IO.Path.GetRelativePath](https://github.com/dotnet/corefx/blob/2c55db90d622fa6279184e6243f0470a3755d13c/src/Common/src/CoreLib/System/IO/Path.cs#L699)

### DIFF
--- a/src/docfx/build/legacy/LegacyUtility.cs
+++ b/src/docfx/build/legacy/LegacyUtility.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Docs.Build
 
         public static string ToLegacyPathRelativeToBasePath(this Document doc, Docset docset)
         {
-            return PathUtility.NormalizeFile(Path.GetRelativePath(docset.Config.SourceBasePath ?? string.Empty, doc.FilePath));
+            return PathUtility.NormalizeFile(Path.GetRelativePath(docset.Config.SourceBasePath ?? ".", doc.FilePath));
         }
 
         public static string ToLegacyOutputPathRelativeToBaseSitePath(this Document doc, Docset docset)
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -49,12 +49,12 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Just for backward compatibility, the output site path prefix
         /// </summary>
-        public readonly string SiteBasePath = string.Empty;
+        public readonly string SiteBasePath = ".";
 
         /// <summary>
         /// Just for backward compatibility, the source path prefix
         /// </summary>
-        public readonly string SourceBasePath = string.Empty;
+        public readonly string SourceBasePath = ".";
 
         /// <summary>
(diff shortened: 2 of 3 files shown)


src/docfx/config/Config.cs
```diff
@@ -49,12 +49,12 @@ internal class Config
...
         /// Just for backward compatibility, the output site path prefix
         /// </summary>
-        public readonly string SiteBasePath = string.Empty;
+        public readonly string SiteBasePath = ".";
 
         /// <summary>
         /// Just for backward compatibility, the source path prefix
         /// </summary>
-        public readonly string SourceBasePath = string.Empty;
+        public readonly string SourceBasePath = ".";
```
OsmondJiang: Would you please check the code if there are any places using string.IsNullorEmpty() for these two field?

src/docfx/build/legacy/LegacyUtility.cs
```diff
@@ -24,7 +24,7 @@ public static void MoveFileSafe(string sourceFileName, string destFileName)
 
         public static string ToLegacyPathRelativeToBasePath(this Document doc, Docset docset)
         {
-            return PathUtility.NormalizeFile(Path.GetRelativePath(docset.Config.SourceBasePath ?? string.Empty, doc.FilePath));
+            return PathUtility.NormalizeFile(Path.GetRelativePath(docset.Config.SourceBasePath ?? ".", doc.FilePath));
```
superyyrrzz: use `string.IsNullOrEmpty(docset.Config.SourceBasePath)`

src/docfx/docset/Document.cs
```diff
@@ -341,9 +341,9 @@ private static bool IsValidRelativePath(string path)
 
         private (string docId, string versionIndependentId) LoadDocumentId()
         {
-            var sourcePath = string.IsNullOrEmpty(Docset.Config.DocumentId.SourceBasePath)
-                ? FilePath
-                : PathUtility.NormalizeFile(Path.GetRelativePath(Docset.Config.DocumentId.SourceBasePath, FilePath));
+            var sourcePath = PathUtility.NormalizeFile(Path.GetRelativePath(
+                    string.IsNullOrEmpty(Docset.Config.DocumentId.SourceBasePath) ? "." : Docset.Config.DocumentId.SourceBasePath,
```
OsmondJiang: >string.IsNullOrEmpty(Docset.Config.DocumentId.SourceBasePath) [](start = 20, length = 61)

why do we need that?

src/docfx/config/Config.cs
live1206: >readonly [](start = 15, length = 8)

how about const instead of readonly and we do not need to check empty afterwards? it needed to be readonly since it was string.Empty. Same for "SiteBasePath"

928PJY: it is not a const, just a default value

---
In reply to: [206002546](https://github.com/dotnet/docfx/pull/3120#discussion_r206002546) [](ancestors = 206002546)

PR #3130, opened 2018-07-30
Use data annotation validation for json schema and add tests for it

live1206:
Added tests:
- RegularExpression
- MinLength/MaxLength for string

#3050

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -112,8 +112,8 @@ namespace Microsoft.Docs.Build
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} '{name}' contains null value", line: range.StartLine, column: range.StartCharacter);
 
-        public static Error InvalidSchema(Range range, string message)
-            => new Error(ErrorLevel.Error, "invalid-schema", $"{range} {message}", line: range.StartLine, column: range.StartCharacter);
+        public static Error ViolateSchema(Range range, string message)
+            => new Error(ErrorLevel.Error, "violate-schema", $"{range} {message}", line: range.StartLine, column: range.StartCharacter);
 
         public static Error SchemaNotFound(string schema)
             => new Error(ErrorLevel.Error, "schema-not-found", $"Unknown schema '{schema}'");
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -112,7 +112,7 @@ namespace Microsoft.Docs.Build
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} '{name}' contains null value", line: range.StartLine, column: range.StartCharacter);
 
-        public static Error InValidSchema(Range range, string message)
+        public static Error InvalidSchema(Range range, string message)
             => new Error(ErrorLevel.Error, "invalid-schema", $"{range} {message}", line: range.StartLine, column: range.StartCharacter);
 
         public static Error SchemaNotFound(string schema)
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -112,6 +112,9 @@ namespace Microsoft.Docs.Build
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} '{name}' contains null value", line: range.StartLine, column: range.StartCharacter);
 
+        public static Error InValidSchema(Range range, string message)
+            => new Error(ErrorLevel.Error, "invalid-schema", $"{range} {message}", line: range.StartLine, column: range.StartCharacter);
+
         public static Error SchemaNotFound(string schema)
             => new Error(ErrorLevel.Error, "schema-not-found", $"Unknown schema '{schema}'");
 
--- a/src/docfx/lib/JsonUtility.cs
+++ b/src/docfx/lib/JsonUtility.cs
@@ -225,8 +225,8 @@ namespace Microsoft.Docs.Build
 
             private static SchemaValidationConverter GetConverter(MemberInfo member)
             {
-                var validators = member.GetCustomAttributes<ValidationAttribute>(false);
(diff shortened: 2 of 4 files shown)


src/docfx/lib/JsonUtility.cs
```diff
@@ -205,8 +208,58 @@ protected override JsonProperty CreateProperty(MemberInfo member, MemberSerializ
...
+                        prop.Converter = converter;
+                    }
+                }
                 return prop;
             }
+
+            private static SchemaValidationConverter GetConverter(MemberInfo member)
+            {
+                var validators = member.GetCustomAttributes<ValidationAttribute>();
+                return !validators.Any() ? null : new SchemaValidationConverter(validators);
```
herohua: >validators [](start = 24, length = 10)

validators.Length == 0

src/docfx/lib/JsonUtility.cs
```diff
@@ -205,8 +208,58 @@ protected override JsonProperty CreateProperty(MemberInfo member, MemberSerializ
...
+                    {
+                        prop.Converter = converter;
+                    }
+                }
                 return prop;
             }
+
+            private static SchemaValidationConverter GetConverter(MemberInfo member)
+            {
+                var validators = member.GetCustomAttributes<ValidationAttribute>();
```
herohua: >GetCustomAttributes [](start = 40, length = 19)

this is reflection, will it impact perf greatly?

src/docfx/lib/JsonUtility.cs
live1206: The type of validators is IEnumerable<T>, Length is not available here, Any() would just iterate once with the first element

---
In reply to: [206377491](https://github.com/dotnet/docfx/pull/3130#discussion_r206377491) [](ancestors = 206377491)

src/docfx/lib/JsonUtility.cs
live1206: Tried to cache validators using MemberInfo as key, seems like equality of MemberInfo is not a trivial operation either. Update the reflection to not look into inherit types, will get back to this if we have perf issue later

---
In reply to: [206377654](https://github.com/dotnet/docfx/pull/3130#discussion_r206377654) [](ancestors = 206377654)

src/docfx/lib/JsonUtility.cs
superyyrrzz: can `ToList()` first. Using IEnumerable twice will call `member.GetCustomAttributes<ValidationAttribute>(false);` twice.

---
In reply to: [206379098](https://github.com/dotnet/docfx/pull/3130#discussion_r206379098) [](ancestors = 206379098,206377491)

src/docfx/Errors.cs
```diff
@@ -112,6 +112,9 @@ public static Error GitNotFound()
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} '{name}' contains null value", line: range.StartLine, column: range.StartCharacter);
 
+        public static Error InvalidSchema(Range range, string message)
```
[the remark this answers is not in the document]
live1206: True :)

---
In reply to: [206406195](https://github.com/dotnet/docfx/pull/3130#discussion_r206406195) [](ancestors = 206406195)

src/docfx/lib/JsonUtility.cs
live1206: I suppose GetCustomAttributes should have yield return inside, since the return type is IEnumerable<T>, .Any() will add one more iteration, not the entire collection :)  Anyway, ToList will  avoid this additional iteration, I will add it.

---
In reply to: [206410955](https://github.com/dotnet/docfx/pull/3130#discussion_r206410955) [](ancestors = 206410955,206379098,206377491)

herohua: seems no need to `ToList()`?
[https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.getcustomattributes?view=netframework-4.7.2#System_Reflection_MemberInfo_GetCustomAttributes_System_Boolean](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.memberinfo.getcustomattributes?view=netframework-4.7.2#System_Reflection_MemberInfo_GetCustomAttributes_System_Boolean)_

---
In reply to: [206421770](https://github.com/dotnet/docfx/pull/3130#discussion_r206421770) [](ancestors = 206421770,206410955,206379098,206377491)

src/docfx/lib/JsonUtility.cs
herohua: sure, that's good enough to know at this moment. :)

---
In reply to: [206384045](https://github.com/dotnet/docfx/pull/3130#discussion_r206384045) [](ancestors = 206384045,206377654)

src/docfx/lib/JsonUtility.cs
live1206: We want to use the extension method with generic type here, because we want to use ValidationAttribute afterwards without casting.
[https://msdn.microsoft.com/en-us/library/hh138362(v=vs.110).aspx](https://msdn.microsoft.com/en-us/library/hh138362(v=vs.110).aspx)
Basically with ToList(), we can have Count in place without iteration, but we created a new list. For IEnumerable<T>, while invoke .Any(), we will need 1 additional iteration to check if any element inside.
[https://stackoverflow.com/questions/305092/which-method-performs-better-any-vs-count-0](https://stackoverflow.com/questions/305092/which-method-performs-better-any-vs-count-0)

---
In reply to: [206512597](https://github.com/dotnet/docfx/pull/3130#discussion_r206512597) [](ancestors = 206512597,206421770,206410955,206379098,206377491)

herohua: sure, in this case, I also prefer use Any(). thanks for detailed explanation!

---
In reply to: [206736462](https://github.com/dotnet/docfx/pull/3130#discussion_r206736462) [](ancestors = 206736462,206512597,206421770,206410955,206379098,206377491)

PR #3135, opened 2018-07-31
Implementation for yaml/json field type not found in target object type 

live1206:
- Add mismatching type validation for Yaml and Json with line info
- Add tests for Yaml and Json
    - Field not found in existing type(e.g. A), warning should be added
    - Field in list item not found in existing type(e.g. List\<A\>), warning should be added
    - Field in nested list item not found in existing type(e.g. List<List\<A\>>), warning should be added
    - Object type contains JsonExtensionData attribute, no warning should be added
    - Nested object type contains JsonExtensionData attibute, no warning should be added for the nested object type
    - Mismatching primitive type(e.g. List\<int\>), error should be added
#3050

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -112,8 +112,8 @@ namespace Microsoft.Docs.Build
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} '{name}' contains null value", line: range.StartLine, column: range.StartCharacter);
 
-        public static Error UnknownFieldType(Range range, string propName, string typeName)
-            => new Error(ErrorLevel.Warning, "unknown-field-type", $"{range} Could not find member '{propName}' on object of type '{typeName}'", line: range.StartLine, column: range.StartCharacter);
+        public static Error UnknownField(Range range, string propName, string typeName, string path)
+            => new Error(ErrorLevel.Warning, "unknown-field", $"{range} Path:{path} Could not find member '{propName}' on object of type '{typeName}'", line: range.StartLine, column: range.StartCharacter);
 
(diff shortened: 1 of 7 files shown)


src/docfx/Errors.cs
```diff
@@ -112,6 +112,12 @@ public static Error GitNotFound()
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} '{name}' contains null value", line: range.StartLine, column: range.StartCharacter);
 
+        public static Error UnknownFieldType(Range range, string propName, string typeName)
+            => new Error(ErrorLevel.Warning, "unknown-field-type", $"{range} Could not find member '{propName}' on object of type '{typeName}'", line: range.StartLine, column: range.StartCharacter);
```
herohua: >object [](start = 115, length = 6)

is it possible to print the name of the object, or even the path of the object relative to root?

src/docfx/Errors.cs
```diff
@@ -112,6 +112,12 @@ public static Error GitNotFound()
         public static Error NullValue(Range range, string name)
             => new Error(ErrorLevel.Info, "null-value", $"{range} '{name}' contains null value", line: range.StartLine, column: range.StartCharacter);
 
+        public static Error UnknownFieldType(Range range, string propName, string typeName)
```
superyyrrzz: >UnknownFieldType [](start = 28, length = 16)

UnknownField?

src/docfx/Errors.cs
live1206: Good point, we can build the path and put it into warning.
The name of the object, do you mean the Yaml/json file name?

superyyrrzz:
CI failed in stylecop

PR #3137, opened 2018-07-31
fix bare fetch bug

OsmondJiang:
I have manually tested it.

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -133,8 +133,8 @@ namespace Microsoft.Docs.Build
         /// </summary>
         /// <param name="cwd">The current working directory</param>
         /// <returns>Task status</returns>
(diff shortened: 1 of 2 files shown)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -133,8 +133,8 @@ public static Task Clone(string cwd, string remote, string path, string branch =
         /// </summary>
         /// <param name="cwd">The current working directory</param>
         /// <returns>Task status</returns>
-        public static Task Fetch(string cwd, string remoteWitToken)
-            => ExecuteNonQuery(cwd, $"fetch {remoteWitToken} '*:*'");
+        public static Task Fetch(string cwd, string remoteWitToken, string refSpec)
```
herohua: >refSpec [](start = 75, length = 7)

add documentation above

Issue #3139, opened 2018-07-31
TOC metadata support


OsmondJiang:
- [ ] Markdown TOC
- [ ] YAML TOC
- [ ] JSON TOC

PR #3142, opened 2018-08-01
support markdown toc metadata

OsmondJiang:


OsmondJiang:
#3139

src/docfx/build/toc/parser/TableOfContentsInputModel.cs
```diff
@@ -8,7 +8,7 @@ namespace Microsoft.Docs.Build
 {
     internal class TableOfContentsInputModel
     {
-        public JObject Metadata { get; set; } = new JObject();
```
superyyrrzz: >= new JObject(); [](start = 46, length = 16)

which change this?

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -40,12 +41,35 @@ public static TableOfContentsInputModel LoadMdTocModel(string tocContent, string
...
+                var yamlHeaderRegex = new Regex(@"^\-{3}(?:\s*?)\n([\s\S]+?)(?:\s*?)\n\-{3}(?:\s*?)(?:\n|$)", RegexOptions.Compiled | RegexOptions.Singleline, TimeSpan.FromSeconds(10));
+                var match = yamlHeaderRegex.Match(content);
+                if (match.Success)
+                {
+                    var yamlContent = content.Substring(match.Groups[1].Index, match.Groups[1].Length);
+                    content = content.Substring(match.Length);
+                    lineNumber += match.Value.Count(ch => ch == '\n');
+                    var (errors, headMetadata) = ExtractYamlHeader.Extract(yamlContent);
+
+                    // todo: return errors once change to markdig
```
superyyrrzz: It should be too late... Can we return them now?

src/docfx/build/toc/parser/TableOfContentsInputModel.cs
OsmondJiang: this is to prevent the output model always has e empty object, it's added by my previous PR

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -40,12 +41,35 @@ public static TableOfContentsInputModel LoadMdTocModel(string tocContent, string
...
+                    if (errors.Any())
```
live1206: >errors.Any() [](start = 24, length = 12)

we should check ErrorLevel.Error here, because we added null value as ErrorLevel.Info into errors during deserialization

docs/specs/toc.yml
```diff
@@ -850,6 +850,29 @@ outputs:
   docs/f1/TOC.experimental.json: |
     {"items":[{"toc_title":"Index Reference","href":"../"}]}
 ---
+# Paired experimental TOCs' metadata
+inputs:
```
herohua: also add case for yaml format?

src/docfx/build/toc/parser/TableOfContentsInputModel.cs
herohua: +1

---
In reply to: [206783267](https://github.com/dotnet/docfx/pull/3142#discussion_r206783267) [](ancestors = 206783267)

superyyrrzz: Sure. It should align with the ideal output schema. We can discuss it when finalize the schema.

---
In reply to: [206786785](https://github.com/dotnet/docfx/pull/3142#discussion_r206786785) [](ancestors = 206786785)

docs/specs/toc.yml
OsmondJiang: you mean the experimental metadata for yaml? sure. I thought the yaml/json toc spec test has already covered that

PR #3148, opened 2018-08-02
Make cache thread safe

live1206:


src/docfx/lib/JsonUtility.cs
```diff
@@ -267,14 +267,11 @@ private static string BuildPath(string path, Type type)
...
         {
-            if (!s_cache.TryGetValue(type, out var containsJsonExtensionData))
-            {
-                containsJsonExtensionData = type.GetProperties().Any(prop => prop.GetCustomAttribute<JsonExtensionDataAttribute>() != null);
-                s_cache.TryAdd(type, containsJsonExtensionData);
-            }
-            return containsJsonExtensionData;
+            return s_cacheTypeContainsJsonExtensionData.GetOrAdd(
+                type,
+                new Lazy<bool>(() => type.GetProperties().Any(prop => prop.GetCustomAttribute<JsonExtensionDataAttribute>() != null))).Value;
```
OsmondJiang: > Lazy<bool> [](start = 19, length = 11)

no need lazy?

OsmondJiang: i know, but it's ok to two twice as it's stateless

live1206: I think we need Lazy here to avoid multiple expensive instantiation of the value with the same key
[https://stackoverflow.com/questions/47489211/concurrentdictionary-lazy-would-instantiation-happen-only-once](https://stackoverflow.com/questions/47489211/concurrentdictionary-lazy-would-instantiation-happen-only-once)

---
In reply to: [207413221](https://github.com/dotnet/docfx/pull/3148#discussion_r207413221) [](ancestors = 207413221)

OsmondJiang: we know that, just wandering whether the getting attribute is worth it

Issue #2812, opened 2018-05-25
GitHub user profile cache


yufeih:
Tracks how we cache GitHub user profile.

## Assumption
- User profiles are global and can be share across repos

## Goals
- Run `docfx build` many times should reuse the cache
- The cache could be shared across machines using `docfx restore` and `docfx publish`

## Draft spec:
- User profile cache has a config entry in `docfx.yml`, it points to either a file or an `URL`
- `docfx restore` download and place them in *%USERPROFILE%/.docfx/downloads/{url-sha2-hash}* if it is an URL.
- `docfx build` maintain a cache at *%USERPROFILE%/.docfx/cache/user-profile.json* from run to run.
  - Call GitHub API to fetch user profile, if missing or expired in the existing cache
  - **TBD** use soft expiration time to avoid spike API call to GitHub
- `docfx publish` POST updated `user-profiles.json` to the URL specified in config **only if** new contributors are found. *Etag* could be used for concurrency control.

### TODO
- [x] Query user profile using user name #3144 
- [x] Query user profile using commit info #3176
- [x] Use token to call GitHub API #3163 
- [x] Update user profile file
### Backlog
- [x] restore from URL
- [x] publish to URL #3673
- [x] log user profile update time for expiration use

@fenxu @superyyrrzz @OsmondJiang @herohua

superyyrrzz:
>  `docfx build` creates `_site/user-profiles.json` **only if** there are new contributors

Is this needed? docfx can directly publish it from local cache, without a copy.

yufeih:
@superyyrrzz you are right, I was under the impression that services wants it under output. It could be from the cache. Updated.

herohua:
@yufeih @superyyrrzz will DocFX handle concurrent issue when multiple instances try writing to the same `user-profile.json` file?

@fenxu as FYI

yufeih:
@herohua , yes, we could read the file on build start and write it once when build ends. Check *last modified time* before write and merge result in case some other process update the file during the build.

herohua:
@yufeih 👍

yufeih:
ETag hasn't been implemented yet, should we keep this issue open or we could live without etag handling?

superyyrrzz:
@yufeih I will create another one tracking it.

PR #3163, opened 2018-08-07
support git-token to call github api

superyyrrzz:
#2812 

remove GitHubAccessor's static as its constructor need parameter

### DIFF
--- a/src/docfx/lib/github/GitHubAccessor.cs
+++ b/src/docfx/lib/github/GitHubAccessor.cs
@@ -3,23 +3,29 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Threading;
 using System.Threading.Tasks;
 using Octokit;
 
 namespace Microsoft.Docs.Build
 {
-    internal static class GitHubAccessor
+    internal class GitHubAccessor
     {
-        private static readonly GitHubClient _client = new GitHubClient(new ProductHeaderValue("DocFXv3"));
+        private readonly GitHubClient _client;
 
-        private static bool _isRateLimitExceeded = false;
+        private bool _isRateLimitExceeded = false;
+
+        public GitHubAccessor(string gitToken = null)
+        {
+            _client = new GitHubClient(new ProductHeaderValue("DocFXv3"));
+            if (!string.IsNullOrEmpty(gitToken))
+                _client.Credentials = new Credentials(gitToken);
+        }
 
         /// <summary>
         /// Get user profile by user name from GitHub API
         /// </summary>
         /// <exception cref="DocfxException">Thrown when user doesn't exist or GitHub rate limit exceeded</exception>
-        public static async Task<UserProfile> GetUserProfileByName(string name)
+        public async Task<UserProfile> GetUserProfileByName(string name)
         {
             Debug.Assert(!string.IsNullOrEmpty(name));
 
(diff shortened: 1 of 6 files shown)


src/docfx/lib/github/UserProfileCache.cs
```diff
@@ -63,9 +66,11 @@ public UserProfile GetByUserEmail(string userEmail)
         /// Create an instance of <see cref="UserProfileCache"/> from local cache
         /// </summary>
         /// <param name="cachePath">the path of the cache file</param>
-        public static UserProfileCache Create(string cachePath)
+        /// <param name="gitToken">the GitHubAccessor to fetch information when missing in cache</param>
+        public static UserProfileCache Create(string cachePath, GitHubAccessor github)
```
live1206: >github [](start = 79, length = 6)

Do we need to pass in github again? since we already have it from the constructor.

superyyrrzz: This is a static method to help create the instance, it calls constructor internally, which need it.

live1206: Right, did not notice that

---
In reply to: [208160192](https://github.com/dotnet/docfx/pull/3163#discussion_r208160192) [](ancestors = 208160192)

src/docfx/lib/github/UserProfileCache.cs
```diff
@@ -16,11 +16,13 @@ internal class UserProfileCache
         private readonly ConcurrentDictionary<string, UserProfile> _cacheByName;
         private readonly ConcurrentDictionary<string, UserProfile> _cacheByEmail;
         private readonly string _cachePath;
+        private readonly GitHubAccessor _github;
 
-        public UserProfileCache(IDictionary<string, UserProfile> cache, string path)
+        public UserProfileCache(IDictionary<string, UserProfile> cache, string path, GitHubAccessor github)
```
live1206: >public [](start = 8, length = 6)

should this be private?

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -46,18 +52,23 @@ public static async Task<UserProfile> GetUserProfileByName(string name)
...
                 ProfileUrl = user.HtmlUrl,
                 DisplayName = user.Name,
                 Name = user.Login,
                 Id = user.Id.ToString(),
+
+                // EmailAddress (public) can only be obtained with OAuth token (not required to be current user's token)
+                EmailAddress = user.Email,
+
+                // UserEmails can only be obtained with current user's OAuth token, so we only get the user's public email
+                UserEmails = user.Email,
```
herohua: >UserEmails [](start = 16, length = 10)

this is emails, not email?

superyyrrzz: This filed is intended to be email**s**, but usually the passed token can only obtain other user's one public email. Private emails can also be filled into this field from commit info.

src/docfx/lib/github/UserProfileCache.cs
superyyrrzz: yes~

---
In reply to: [208164576](https://github.com/dotnet/docfx/pull/3163#discussion_r208164576) [](ancestors = 208164576)

PR #3164, opened 2018-08-07
 Add GetJsonSchemaFromType to JsonUtility, and add tests for it

live1206:
No entry for CLI yet, add the functionality for now
#3050

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -127,11 +127,11 @@ namespace Microsoft.Docs.Build
         public static Error GitHubUserNotFound()
             => new Error(ErrorLevel.Warning, "github-user-not-found", $"User not found on GitHub");
 
-        public static Error NewtonJsonSchemaLimitExceeded(string message)
-            => new Error(ErrorLevel.Warning, "newton-json-schema-limit-exceeded", message);
+        public static Error NewtonsoftJsonSchemaLimitExceeded(string message)
+            => new Error(ErrorLevel.Warning, "newtonsoft-json-schema-limit-exceeded", message);
 
-        public static Error NewtonJsonSchemaLicenseRegistrationFailed(string message)
-            => new Error(ErrorLevel.Warning, "newton-json-schema-registraion-failed", $"Encountered issue while register license for Newtonsoft.Json.Schema: {message}");
+        public static Error NewtonsoftJsonSchemaLicenseRegistrationFailed(string message)
+            => new Error(ErrorLevel.Warning, "newtonsoft-json-schema-registraion-failed", $"Encountered issue while register license for Newtonsoft.Json.Schema: {message}");
 
         private static Range ParseRangeFromYamlSyntaxException(YamlException ex)
         {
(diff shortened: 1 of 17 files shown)


src/docfx/Errors.cs
```diff
@@ -127,6 +127,12 @@ public static Error ExceedGitHubRateLimit()
         public static Error GitHubUserNotFound()
             => new Error(ErrorLevel.Warning, "github-user-not-found", $"User not found on GitHub");
 
+        public static Error NewtonJsonSchemaLimitExceeded(string message)
```
superyyrrzz: NewtonsoftJsonSchemaLimitExceeded?

src/docfx/lib/JsonUtility.cs
```diff
@@ -162,6 +168,26 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
             return result;
         }
 
+        /// <summary>
+        /// Generate Json schema from object type
+        /// </summary>
+        public static (List<Error>, JSchema) GetJsonSchemaFromType(Type type, string license = null)
+        {
+            var errors = new List<Error>();
+            var licenseRegistrationError = RegisterNewtonJsonSchemaLicense(license);
```
superyyrrzz: Can we call register json schema only once?

src/docfx/lib/JsonUtility.cs
```diff
@@ -181,6 +207,51 @@ internal static List<Error> ValidateMismatchingFieldType(this JToken token, Type
...
+                return new List<Error>() { Errors.NewtonJsonSchemaLicenseRegistrationFailed(ex.Message) };
+            }
+        }
+
+        private static (List<Error>, JSchema) GenerateJSchema(Type type)
+        {
+            try
+            {
+                if (_schemaGenerationEnabled != 1)
+                    return (new List<Error>(), null);
```
superyyrrzz: also return limit exceeded error here?

src/docfx/lib/JsonUtility.cs
```diff
@@ -181,6 +207,51 @@ internal static List<Error> ValidateMismatchingFieldType(this JToken token, Type
...
+            }
+
+            try
+            {
+                License.RegisterLicense(license);
+                return new List<Error>();
+            }
+            catch (Exception ex)
+            {
+                return new List<Error>() { Errors.NewtonJsonSchemaLicenseRegistrationFailed(ex.Message) };
```
herohua: >() [](start = 38, length = 2)

no need

src/docfx/Errors.cs
herohua: +1

---
In reply to: [208163143](https://github.com/dotnet/docfx/pull/3164#discussion_r208163143) [](ancestors = 208163143)

src/docfx/lib/JsonUtility.cs
live1206: You are right, I will make register method public for the consumer to invoke just once.

---
In reply to: [208164623](https://github.com/dotnet/docfx/pull/3164#discussion_r208164623) [](ancestors = 208164623)

src/docfx/lib/JsonUtility.cs
live1206: I am think about only logging the limit exceeded warning once, for instance if we generate 20 types during the run, it would be too many warnings in the log. In v2, we handled it the same way.

---
In reply to: [208167542](https://github.com/dotnet/docfx/pull/3164#discussion_r208167542) [](ancestors = 208167542)

PR #3167, opened 2018-08-07
Fix closure bug in render delegate.

jeahyoun:
See https://ceapex.visualstudio.com/Engineering/_workitems/edit/33877. This fixes a closure-based bug in the rendering portion of the chromeless forms plugin.

### DIFF
--- a/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ChromelessFormExtension.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ChromelessFormExtension.cs
@@ -25,8 +25,8 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
         public bool TryProcessAttributes(IDictionary<string, string> attributes, out HtmlAttributes htmlAttributes, out IDictionary<string, string> renderProperties, Action<string> logError)
         {
             htmlAttributes = null;
-			renderProperties = new Dictionary<string, string>();
-			var model = string.Empty;
+            renderProperties = new Dictionary<string, string>();
+            var model = string.Empty;
             var action = string.Empty;
             var submitText = string.Empty;
             foreach (var attribute in attributes)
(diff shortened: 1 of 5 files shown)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ChromelessFormExtension.cs
```diff
@@ -15,6 +15,8 @@ public class ChromelessFormExtension : ITripleColonExtensionInfo
         public bool SelfClosing => true;
         public Func<HtmlRenderer, TripleColonBlock, bool> RenderDelegate { get; private set; }
 
+		private Queue<string> submitTextArray = new Queue<string>();
```
herohua: align with line 16

herohua:
@yishengjin1413 @qinezh @wh-alice to take a look

jeahyoun:
@herohua Whitespace has been updated.

herohua:
thank you!

---
In reply to: [411869135](https://github.com/dotnet/docfx/pull/3167#issuecomment-411869135) [](ancestors = 411869135)

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ChromelessFormExtension.cs
```diff
@@ -68,11 +70,15 @@ public bool TryProcessAttributes(IDictionary<string, string> attributes, out Htm
...
+            submitTextArray.Enqueue(submitText);
+
             RenderDelegate = (renderer, obj) =>
             {
+                var s = submitTextArray.Dequeue();
+
                 renderer.Write("<form").WriteAttributes(obj).WriteLine(">");
                 renderer.WriteLine("<div></div>");
-                renderer.WriteLine($"<button disabled=\"disabled\" type=\"submit\">{submitText}</button>");
+                renderer.WriteLine($"<button class=\"button is-primary\" disabled=\"disabled\" type=\"submit\">{s}</button>");
```
yishengjin1413: Why don't you add "submitText" to RenderProperties in obj? So the obj can contain all the information you need for rendering.

jeahyoun: The problem here is that inside of the "TryProcessAttributes" function, we don't have access to the root RenderProperties object. (We gain access to it in the delegate, but by then it's too late.) This was by design. The only way to hold state here for multiple iterations is at the class level by using multiple variables, or at the class level by using a single collection.

jeahyoun: I'd like to preserve Jeremy's API, as it seems simple and extensible (I don't want to add methods that are only needed for my extension). However, I think there may be a way to do this by creating a new HTML child block.

yishengjin1413: Sure. That would be great. It will more friendly for others to maintain the code.

jeahyoun:
This is ready for another review - I've updated tabs to spaces again (sorry!). Thank you!

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ChromelessFormExtension.cs
yishengjin1413: Thanks @jeahyoun

Issue #2544, opened 2018-03-22
Main nav bar does not collapse when items are added to the dropdown


cedar-ave:
**DocFX Version Used**: 2.32.2

**Template used**: (`default` or `statictoc` or contain custom template) default + custom

**Steps to Reproduce**:

1. Add dropdown items to the main nav bar
```yml
- name: Guides
  href: articles/
- name: API
  items:
  - name: REST API
    href: restapi/
  - name: Plugins
    href: obj/api/
```
**Expected Behavior**: In this example, the nav bar would expand when you hover over "API"

**Actual Behavior**: The "dropdown" items are permanently stacked in an open position

![2018-03-22 16_43_46-microsoft edge](https://user-images.githubusercontent.com/15255009/37802666-0d5f5d82-2df1-11e8-8832-409249241d44.png)

Thank you.

vicancy:
Yes, currently only 1 level navbar is supported....... Need css/js/html update to support deeper levels.

cedar-ave:
Thank you @vicancy

cblackuk:
Is there an eta for fixing it please? :-)

ollie-nye:
I've written a basic fix [here](https://github.com/ollie-nye/docfx-dropdown-fix). The Readme provides the steps to fix the issue in a blank project, so should be portable to other projects as it only modifies a copy of the default DocFX template.

cblackuk:
@ollie-nye  Thanks a lot for that! Outstanding job!

superyyrrzz:
@ollie-nye Looks great! Would mind creating a PR to DocFX's main source base? Thank you.

superyyrrzz:
@cedar-ave @ollie-nye Thank you for contribution. The feature is released in v2.38.

cedar-ave:
Thank you very much @ollie-nye !

mohaOujda:
It's possible to add a dropdownlist to the TOC like in MSDN?

![image](https://user-images.githubusercontent.com/35834224/56206464-c5926f00-604c-11e9-870a-5acda402294f.png)

Thank you,

superyyrrzz:
@WanderWang See [this](https://dotnet.github.io/docfx/tutorial/howto_create_custom_template.html#merge-template-with-default-template), and modify `docfx.css`.

@mohaOujda The theme for docs.microsoft.com is not open source. There is no easy way to achieve this for now.

mohaOujda:
DocFX Version Used: .2.42.3

[...]

Issue #1914, opened 2017-07-27
Flawed Navbar with long titles


xxcuratorxx:
**DocFX Version Used**: 2.21.2

**Template used**: statictoc

**Steps to Reproduce**:
Create website with docfx.json

```
{  
  "build": {
    "content": [
      {
        "files": [ "*.{md,yml}"]
      }
    ],
    "dest": "_site",
    "template": [
      "statictoc"
    ]
  }
}
```
test.md
```
Hello world
```

and 

toc.yml
```
- name: VeryLongVeryLongEntry1
  href: test.md
- name: VeryLongVeryLongEntry2
  href: test.md
- name: VeryLongVeryLongEntry3
  href: test.md
- name: VeryLongVeryLongEntry4
  href: test.md
```
All in the same path.

**Expected Behavior**: Nice Navbar in each responsive design

**Actual Behavior**: Flawed navbar in medium design (emptry line)

Benzelona:
push.

xxcuratorxx:
Hi again,
are there any updates planed for this bug? The same occurs when there are some more entries, where their names are not very long.

That makes the html unusable, since the empty line overwrites the navigation bar of programming references

xxcuratorxx:
However, since this seems not to be fixed that early: Are there any plans to have a 2nd level navbar?

xxcuratorxx:
Any Update on the?

yufeih:
The navbar experience has been improved in the new modern template.

PR #3168, opened 2018-08-07
Fix dropdown issue

ollie-nye:
- Allows dropdowns in the navbar
- Testing steps
  - `docfx init -q`
  - Add: 
    ```
    - name: More
      dropdown: true
      items:
      - name: Item 1
        topicHref: /abc
      - name: Item 2
        topicHref: /def
      - name: Item 3
        topicHref: /ghi
    ```
  to toc.yml
  - `docfx docfx.json --serve

Dropdown should appear alongside 'Api Documentation'

Relating to issue #2544

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -391,7 +391,9 @@ $(function () {
               }
             }
             if (isActive) {
-              $(e).addClass(active);
+              if ((window.location.pathname != '/index.html') && (window.location.pathname != '/')) {
+                $(e).addClass(active);
+              }
             }
           }
         });
--- a/src/docfx.website.themes/default/partials/dd-li.tmpl.partial
+++ b/src/docfx.website.themes/default/partials/dd-li.tmpl.partial
@@ -1,0 +1,3 @@
+{{#items}}
+  <li><a href="{{topicHref}}">{{name}}</a></li>
+{{/items}}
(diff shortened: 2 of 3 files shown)


src/docfx.website.themes/default/styles/docfx.js
```diff
@@ -391,7 +391,9 @@ $(function () {
               }
             }
             if (isActive) {
-              $(e).addClass(active);
+              if ((window.location.pathname != '/index.html') && (window.location.pathname != '/')) {
+                $(e).addClass(active);
```
superyyrrzz: The e2e test in CI fails because the breadcrumb is missing for index.html, which should be related to this piece of code.

I think the proper fix should modify L389 `if (util.getAbsolutePath(href) === currentAbsPath)`. This line checks whether the node in nav bar links to the current page. In the drop down menu scenario, it happens `getAbsolutePath("#") === "index.html"`, which makes `isActive = true` wrongly. 

What about complement this check to make sure the current node isn't a drop down placeholder? This may be done by check whether the node has the attribute `data-toggle="dropdown"`.

ollie-nye: This has been moved in 2c68a28

xxcuratorxx:
Hi,

nice feature. However, it does not work with template statictoc for local use. If that would be possible, I could workaround #1914. Any chances to extend the dropdown for statictoc, too?

superyyrrzz:
@xxcuratorxx It need some code change. I created an issue to track this: #3361

ollie-nye:
I'll port the change over to statictoc in the next couple of days hopefully

xxcuratorxx:
Sorry I have to came back to this.

We are planning to release our product early with an extended navigation. Is there any time schedule available or is a local hotfix possible?

In addition: even with default template (--serve), the breadCrump bar seams to miss the first level of navigation

ollie-nye:
Hi @xxcuratorxx 
Sorry about the delay, PR open for this now [here](https://github.com/dotnet/docfx/pull/3536)

PR #3172, opened 2018-08-09
Fix for check JsonExtensionDataAttribute for array type

live1206:


superyyrrzz:
>         private static bool CheckTypeContainsJsonExtensionData(Type type)

This logic should only apply to object type. If it's array, we should check each of its item.

---
Refers to: src/docfx/lib/JsonUtility.cs:341 in 6f3e231. [](commit_id = 6f3e2319518ade26859f79e1cecca475adfbd306, deletion_comment = False)

superyyrrzz:
>                 return;

This check will ignore nested items.

---
Refers to: src/docfx/lib/JsonUtility.cs:310 in 6f3e231. [](commit_id = 6f3e2319518ade26859f79e1cecca475adfbd306, deletion_comment = False)

superyyrrzz:
>                     schema.AllowAdditionalProperties = false;

This should be added to object type, not array type.

---
Refers to: src/docfx/lib/JsonUtility.cs:241 in 6f3e231. [](commit_id = 6f3e2319518ade26859f79e1cecca475adfbd306, deletion_comment = False)

live1206:
>         private static bool CheckTypeContainsJsonExtensionData(Type type)

Good catch!

---
In reply to: [411951317](https://github.com/dotnet/docfx/pull/3172#issuecomment-411951317) [](ancestors = 411951317)

---
Refers to: src/docfx/lib/JsonUtility.cs:341 in 6f3e231. [](commit_id = 6f3e2319518ade26859f79e1cecca475adfbd306, deletion_comment = False)

live1206:
>                     schema.AllowAdditionalProperties = false;

We use Newtonsoft.Json.Schema to generate json schema from object type, it sets AllowAddtionalProperties to false while the object type is a sealed class. We need to post set it after the generation, seems tricky to deal with for the nested type, will create a new PR for this.

---
In reply to: [411951551](https://github.com/dotnet/docfx/pull/3172#issuecomment-411951551) [](ancestors = 411951551)

---
Refers to: src/docfx/lib/JsonUtility.cs:241 in 6f3e231. [](commit_id = 6f3e2319518ade26859f79e1cecca475adfbd306, deletion_comment = False)

src/docfx/lib/JsonUtility.cs
```diff
@@ -342,10 +345,33 @@ private static bool CheckTypeContainsJsonExtensionData(Type type)
...
+            else if (contract is JsonArrayContract arrayContract)
+            {
+                var itemType = arrayContract.CollectionItemType;
+                if (itemType is null)
+                {
+                    return type;
+                }
+                else
+                {
+                    return GetCollectionItemTypeFromArrayType(itemType);
```
superyyrrzz: >return GetCollectionItemTypeFromArrayType(itemType); [](start = 20, length = 52)

As the outer `TraverseForUnknownFieldType` will be called recurtively, I think here only need to get the type of item of current array, which doens't need recursion to inspect deeper.

superyyrrzz:
>                     schema.AllowAdditionalProperties = false;

sounds nice

---
In reply to: [411959224](https://github.com/dotnet/docfx/pull/3172#issuecomment-411959224) [](ancestors = 411959224,411951551)

---
Refers to: src/docfx/lib/JsonUtility.cs:241 in 6f3e231. [](commit_id = 6f3e2319518ade26859f79e1cecca475adfbd306, deletion_comment = False)

PR #3175, opened 2018-08-10
Feature/json deserialize multiple error

live1206:
Return all the schema violation errors, not throw exception to block the flow.

### DIFF
--- a/src/docfx/lib/JsonUtility.cs
+++ b/src/docfx/lib/JsonUtility.cs
@@ -64,9 +64,6 @@ namespace Microsoft.Docs.Build
         // 1 for schema generation enabled, 0 for disabled
         private static int _schemaGenerationEnabled = 1;
 
-        [ThreadStatic]
-        private static List<Error> _schemaViolationErrors;
-
         /// <summary>
         /// Serialize an object to TextWriter
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/docfx/lib/JsonUtility.cs
```diff
@@ -108,14 +111,32 @@ public static (List<Error>, T) ToObject<T>(JToken token)
...
         public static (List<Error>, object) ToObject(JToken token, Type type)
         {
+            ThreadLocal<List<Error>> errors = new ThreadLocal<List<Error>>(() => new List<Error>());
             try
             {
-                return (token.ValidateMismatchingFieldType(type), token.ToObject(type, DefaultDeserializer));
+                var mismatchingErrors = token.ValidateMismatchingFieldType(type);
+                errors.Value.AddRange(mismatchingErrors);
+                DefaultDeserializer.Error += HandleError;
+                _schemaViolationErrors = new List<Error>();
```
superyyrrzz: >_schemaViolationErrors [](start = 16, length = 22)

what about initialize deserializer to pass the errors

live1206: We should not use the static DefaultDeserializer here, since it might pollute the other methods calls, since we are handling the deserialization errors in a different way.
Since we need to create a new Deserializer for each call, we could pass along the errors

---
In reply to: [209191678](https://github.com/dotnet/docfx/pull/3175#discussion_r209191678) [](ancestors = 209191678)

src/docfx/lib/JsonUtility.cs
```diff
@@ -108,14 +108,29 @@ public static (List<Error>, T) ToObject<T>(JToken token)
 
         public static (List<Error>, object) ToObject(JToken token, Type type)
         {
-            try
+            List<Error> errors = new List<Error>();
+            var mismatchingErrors = token.ValidateMismatchingFieldType(type);
+            errors.AddRange(mismatchingErrors);
+            DefaultDeserializer.Error += HandleError;
```
superyyrrzz: >DefaultDeserializer.Error += HandleError; [](start = 12, length = 41)

remove?

src/docfx/lib/JsonUtility.cs
```diff
@@ -108,14 +108,29 @@ public static (List<Error>, T) ToObject<T>(JToken token)
...
             {
-                return (token.ValidateMismatchingFieldType(type), token.ToObject(type, DefaultDeserializer));
-            }
-            catch (JsonException ex) when (ex is JsonSerializationException || ex is JsonReaderException)
+                NullValueHandling = NullValueHandling.Ignore,
+                ContractResolver = new JsonContractResolver(errors),
+            };
+            serializer.Error += HandleError;
+            var value = token.ToObject(type, serializer);
+            DefaultDeserializer.Error -= HandleError;
```
superyyrrzz: >DefaultDeserializer.Error -= HandleError; [](start = 12, length = 41)

remove?

src/docfx/lib/JsonUtility.cs
```diff
@@ -395,18 +410,17 @@ private static void LogInfoForNullValue(IJsonLineInfo token, List<Error> errors,
 
         private sealed class JsonContractResolver : DefaultContractResolver
         {
+            private readonly List<Error> _errors;
+
+            public JsonContractResolver(List<Error> errors = null)
```
superyyrrzz: >null [](start = 61, length = 4)

I see this is null in `DefaultDeserializer`. Will it cause ArgumentNullException when call `converter.Validate()`?

live1206: And also we might want to invoke our ToObject<T> instead of the one from Json.Net to execute schema validation, like config and toc

---
In reply to: [209227254](https://github.com/dotnet/docfx/pull/3175#discussion_r209227254) [](ancestors = 209227254,209203462)

PR #3182, opened 2018-08-13
fix legacy metaddata

928PJY:


src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -67,7 +69,7 @@ public static JObject GenerataCommonMetadata(JObject metadata, Docset docset)
 
             rawMetadata["layout"] = rawMetadata.TryGetValue("layout", out JToken layout) ? layout : "Conceptual";
 
-            rawMetadata["_path"] = PathUtility.NormalizeFile(file.ToLegacyPathRelativeToBasePath(docset));
+            rawMetadata["_path"] = PathUtility.NormalizeFile(Path.GetRelativePath(file.Docset.Config.SiteBasePath, file.OutputPath));
```
superyyrrzz: >Path.GetRelativePath(file.Docset.Config.SiteBasePath, file.OutputPath) [](start = 61, length = 70)

can we use `legacyManifestOutput.PageOutput.OutputPathRelativeToSiteBasePath`?

928PJY: that including file extension

---
In reply to: [209518015](https://github.com/dotnet/docfx/pull/3182#discussion_r209518015) [](ancestors = 209518015)

PR #3189, opened 2018-08-13
Update legacy TOC metadata

mingwli:
Move metadata from first child to toc directly.

### DIFF
--- a/src/docfx/build/legacy/files/LegacyTableOfContents.cs
+++ b/src/docfx/build/legacy/files/LegacyTableOfContents.cs
@@ -22,12 +22,17 @@ namespace Microsoft.Docs.Build
             var firstItem = toc?.Items?.FirstOrDefault();
             if (firstItem != null)
             {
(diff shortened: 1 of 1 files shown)


src/docfx/build/legacy/files/LegacyTableOfContents.cs
```diff
@@ -22,13 +22,18 @@ internal static class LegacyTableOfContents
             var firstItem = toc?.Items?.FirstOrDefault();
             if (firstItem != null)
             {
-                firstItem.PdfAbsolutePath = PathUtility.NormalizeFile(
```
928PJY: >PdfAbsolutePath [](start = 26, length = 15)

remove those two  properties

928PJY:
>         private class LegacyTableOfContentsItem : TableOfContentsItem

do not need this class any more

---
Refers to: src/docfx/build/legacy/files/LegacyTableOfContents.cs:61 in 07b43af. [](commit_id = 07b43af56b4626f3b9c0b75b871739c90e3adbd2, deletion_comment = False)

PR #3193, opened 2018-08-14
Use v3 toc item model in legacy conversion

mingwli:
Not inheriting v3 toc item model in case that pdf-related metadata get deserialized


928PJY:
#3194

PR #3199, opened 2018-08-14
Json/Yaml deserialize split

live1206:
Split json/yaml deserialization into normal one and the one with schema validation.


live1206:
Performance issue solved  with #3202

Issue #3197, opened 2018-08-14
Converting legacy markdown/redirection took 3-5 minutes


OsmondJiang:
That takes too much times, even the converting resource file only takes 1-2 minutes.

OsmondJiang:
I suspect the JINT takes too much times

OsmondJiang:
The newly impact test shows that the difference is almost 1 minutes on my machine:

**using schema validation:**
Convert Legacy Markdown/Redirection Files done in **00:02:21**
Convert Legacy Resource Files done in 00:01:40

**without schema validation:**
Convert Legacy Markdown/Redirection Files done in **00:01:34**
Convert Legacy Resource Files done in 00:01:43

@live1206

PR #3202, opened 2018-08-15
The footprint of passing along errors list was too big, it caused per…

live1206:
…f issues.

#3197 

Before fix:
Convert Legacy Markdown/Redirection Files done in 00:01:42

After fix:
Convert Legacy Markdown/Redirection Files done in 00:01:01

### DIFF
--- a/src/docfx/lib/JsonUtility.cs
+++ b/src/docfx/lib/JsonUtility.cs
@@ -64,6 +64,9 @@ namespace Microsoft.Docs.Build
         // 1 for schema generation enabled, 0 for disabled
         private static int _schemaGenerationEnabled = 1;
 
+        [ThreadStatic]
+        private static List<Error> s_schemaViolationErrors;
+
         /// <summary>
         /// Serialize an object to TextWriter
         /// </summary>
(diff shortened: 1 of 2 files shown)


live1206:
With the reminding of @herohua, I had a second look at this performance issue.

The fix in this PR is still valid: 
```
Convert Legacy Markdown/Redirection Files done in 00:01:04
```

And I found the real root cause was located at:
[root cause](https://github.com/dotnet/docfx/blob/21e264dddb99aed67b62981e175fe88a4eb9ac14/src/docfx/lib/JsonUtility.cs#L122)

I was able to reproduce the performance issue with changing this line to:
```csharp
ContractResolver = new JsonContractResolver(),
```

Then the result would be:
```
Convert Legacy Markdown/Redirection Files done in 00:01:45
```

So if we create a new contract resolver for each call, we will face a performance problem.
And the reason for that is within the constructor of `DefaultContractResolver` from `Newtonsoft.Json`, a contract cache will be built for each instantiation, which is quite expensive:
[contract cache](https://github.com/JamesNK/Newtonsoft.Json/blob/ad246c775a6efc6fb42b81311ab35524064d5ec1/Src/Newtonsoft.Json/Serialization/DefaultContractResolver.cs#L186)

FYI @yufeih @OsmondJiang @928PJY @superyyrrzz @mingwli @herohua

PR #3204, opened 2018-08-15
Write errors to stderr, set exit code to 1 on errors

yufeih:


src/docfx/lib/log/Report.cs
```diff
@@ -89,10 +104,11 @@ private static void ConsoleLog(ErrorLevel level, Error error)
             lock (Console.Out)
```
live1206: >Console.Out [](start = 18, length = 11)

since output could be Console.Error  or Console.Out, should we lock output instead?

PR #3196, opened 2018-08-14
report error when there is no H1 or there are some non-comments block between yaml-header and H1

928PJY:
#3195

PR #3206, opened 2018-08-16
Exclude config external references from build

yufeih:
#3050 #3196

### DIFF
--- a/src/docfx/docset/Document.cs
+++ b/src/docfx/docset/Document.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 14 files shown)


src/docfx/docset/Document.cs
```diff
@@ -188,7 +189,7 @@ public static (Error error, Document doc) TryCreate(Docset docset, string path,
             Debug.Assert(!Path.IsPathRooted(path));
 
             var filePath = PathUtility.NormalizeFile(path);
-            var type = GetContentType(filePath);
+            var type = docset.Config.Extend.Contains(filePath, PathUtility.PathComparer) ? ContentType.Unknown : GetContentType(filePath);
```
OsmondJiang: should we also exclude other configuration files like github user info and build info?

PR #3208, opened 2018-08-16
remove most of metadata for redirection file

OsmondJiang:


src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -40,13 +40,20 @@ public static JObject GenerataCommonMetadata(JObject metadata, Docset docset)
             return newMetadata;
         }
 
+        public static JObject GenerateLegacyRedirectionRawMetadata(Docset docset, PageModel pageModel)
+            => new JObject
+            {
+                ["redirect_url"] = pageModel.RedirectionUrl,
+                ["locale"] = docset.Config.Locale,
```
928PJY: >["locale"] = docset.Config.Locale [](start = 16, length = 33)

why we need `locale`

OsmondJiang: for publishing usage

---
In reply to: [210468903](https://github.com/dotnet/docfx/pull/3208#discussion_r210468903) [](ancestors = 210468903)

yufeih:
>             return new PageModel

Instead of returning a PageModel, can we return a `Redirection` that only contains these two properties?

---
Refers to: src/docfx/build/Build.cs:152 in 90b37f8. [](commit_id = 90b37f853b9b5ed1372b4d599f7336aa9863d5de, deletion_comment = False)

PR #3215, opened 2018-08-16
Improve how to verify log, so that we could see the expected and actu…

live1206:
…al values.

In #3169, wildcard "*" was involved for the build log, but the build error would not display the build log content any more.


test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -174,8 +175,8 @@ private static void VerifyFile(string file, string content)
                     break;
                 case ".log":
                     var expected = content.Split(new char[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries).OrderBy(_ => _).ToList();
-                    var actual = File.ReadAllLines(file).OrderBy(_ => _).ToList();
-                    TestHelper.VerifyLogEquals(expected, actual);
+                    var actual = File.ReadAllLines(file).OrderBy(_ => _);
+                    Assert.Matches("^" + Regex.Escape(string.Join("\n", expected)).Replace("\\*", ".*").Replace("\\?", ".") + "$", string.Join("\n", actual));
```
yufeih: what's wrong with `Assert.Equals(join("\n", expected), join("\n", actual))` ?

live1206: Updated the PR description, wildcard was involved in #3169

---
In reply to: [210510655](https://github.com/dotnet/docfx/pull/3215#discussion_r210510655) [](ancestors = 210510655)

live1206: yes, it was due to github rate limit. Hmm, I have another PR CI failing with a  test I could not reproduce locally. So I want to fix the build error log to unblock my PR, can we get this in firstly? I will add a TODO to get rid of wildcard

---
In reply to: [210512951](https://github.com/dotnet/docfx/pull/3215#discussion_r210512951) [](ancestors = 210512951)

yufeih: yes, sure

PR #3216, opened 2018-08-16
create and validate json schema in CI

yufeih:
#3050


schemas/Config.json
```diff
@@ -0,0 +1,332 @@
+{
+  "definitions": {
```
live1206: >"definitions": { [](start = 2, length = 16)

Where does this file come from?

yufeih: generated by CreateJsonSchema

live1206: I mean I can see like LandingData.json generated from LandingData.cs, was this generated from Config.cs?

---
In reply to: [210524771](https://github.com/dotnet/docfx/pull/3216#discussion_r210524771) [](ancestors = 210524771)

yufeih: the idea is that if you make any changes that changes JSON schema, CI can detect that and force you regenerate a new JSON schema by running build.ps1 locally. This ensures JSON schema is always up to date with code changes.

tools/CreateJsonSchema/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+using Newtonsoft.Json;
+using Newtonsoft.Json.Schema;
+using Newtonsoft.Json.Schema.Generation;
+using Newtonsoft.Json.Serialization;
+
+class Program
+{
+    static int Main(string[] args)
+    {
+        var license = Environment.GetEnvironmentVariable("JSON_SCHEMA_LICENSE");
```
live1206: >license [](start = 12, length = 7)

So we have set the license on CI machine already? Otherwise, the limit is 10 generation per hour, which we have not reached yet.

schemas/Config.json
yufeih: yes, it is hard coded as `CreateJsonSchema(typeof(Config))`

tools/CreateJsonSchema/Program.cs
yufeih: not yet, I don't have the keys :(

tools/CreateJsonSchema/Program.cs
```diff
@@ -0,0 +1,79 @@
...
+            License.RegisterLicense(license);
+        }
+
+        if (Directory.Exists("schemas"))
+        {
+            Directory.Delete("schemas");
+        }
+        Directory.CreateDirectory("schemas");
+
+        GenerateJSchema(typeof(Config));
```
live1206: >GenerateJSchema(typeof(Config)); [](start = 8, length = 32)

OK, I see it

live1206:
CI build is failing

schemas/docfx.json
```diff
@@ -0,0 +1,332 @@
...
+        "showContributors": {
+          "type": "boolean"
+        },
+        "showEdit": {
+          "type": "boolean"
+        },
+        "repository": {
+          "type": [
+            "string",
+            "null"
```
OsmondJiang: no need null? string can be null?

schemas/LandingData.json
```diff
@@ -0,0 +1,109 @@
...
+      ],
+      "properties": {
+        "type": {
+          "type": "integer",
+          "enum": [
+            0,
+            1,
+            2,
+            3,
+            4
```
OsmondJiang: really?

tools/CreateJsonSchema/Program.cs
```diff
@@ -0,0 +1,80 @@
...
+            License.RegisterLicense(license);
+        }
+
+        if (Directory.Exists("schemas"))
+        {
+            Directory.Delete("schemas", recursive: true);
+        }
+        Directory.CreateDirectory("schemas");
+
+        GenerateJSchema(typeof(Config), "docfx");
```
OsmondJiang: so this tool is not using passed-in path?

PR #3222, opened 2018-08-17
add retry and friendly errors to restore url

yufeih:
#2863

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -78,3 +78,11 @@ restores:
     {"git":{"https://github.com/docascode/docfx-test-dependencies#child":"5736374c0a142252063bbedafe5d98db495e263d"}}
 outputs:
   build.manifest:
+---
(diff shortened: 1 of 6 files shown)


docs/specs/restore.yml
```diff
@@ -78,3 +78,12 @@ restores:
...
   build.manifest:
+---
+# Show error when download failed
+os: windows
+inputs:
+  docfx.yml: |
+    extend: https://this-url-does-not-exit.com/this-is-a-bad-url
+outputs:
+  build.log: |
+    ["error","download-failed","Download 'https://this-url-does-not-exit.com/this-is-a-bad-url' failed: No such host is known"]
```
live1206: The error message should be "No such device or address", according to the failing test.

yufeih: yeah, it is different based on os, I'm trying to exclude the test from linux/mac

PR #3223, opened 2018-08-17
Show build.log content if actual output has errors or warnings.

yufeih:


test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -44,28 +44,34 @@ private static async Task RunCore(string docsetPath, E2ESpec spec)
...
+            }
+
             Assert.Equal(spec.Outputs.Keys.OrderBy(_ => _), outputFileNames.OrderBy(_ => _));
 
             foreach (var (filename, content) in spec.Outputs)
             {
                 VerifyFile(Path.GetFullPath(Path.Combine(docsetOutputPath, filename)), content);
             }
+
+            // Verify restored files
```
OsmondJiang: why move to last？ If the restoring failed, the output file may also be wrong, then you can't easily found the root cause

Issue #2512, opened 2018-03-09
ValueTuples don't show up properly


benjamin-hodgson:
**DocFX Version Used**: 2.32.2

**Template used**: `default`

**Steps to Reproduce**:

Generate API docs for a method which features a `ValueTuple` in its type. I happened to spot it with an implicit conversion:
```csharp
public static implicit operator Attr((string name, string value) nameValue) { /* ... */ }
```

**Expected Behavior**:

The correct tuple type shows up in the generated HTML

**Actual Behavior**:

The type is missing:

![](https://i.imgur.com/ISL8WOp.jpg)

leandroltavares:
It also happens in return types

`public static IEnumerable<(string, Match)> GetFilesByRegex(string path, string pattern, SearchOption searchOption)`

is shown as 

`public static IEnumerable<> GetFilesByRegex(string path, string pattern, SearchOption searchOption)`

werwolfby:
I’m working on this, and I have few promissing results: [#117](
https://github.com/nuke-build/nuke/issues/117)

werwolfby:
There is some small issue in Roslyn, that do not add right spaces in some cases:
![image](https://user-images.githubusercontent.com/705754/44307907-a9eaa200-a3b3-11e8-97cd-85ac0c8f1f30.png)

Beside that everything looks ok.

werwolfby:
@superyyrrzz I think we can close this issue, because the fix (#3228) was already merged. 
It wasn't auto closed because there is not reference in commit message

Issue #3211, opened 2018-08-16
Params parameter with tuple type


matkoch:
Method declarations like this:

```
public static string XmlPeekSingle(string path, string xpath, params (string prefix, string uri)[] namespaces)
```

cause docfx.console.exe to throw NullReferenceException here:

https://github.com/dotnet/docfx/blob/275e0c337ed9ea9145c54d49612c34950e687337/src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/NameVisitor.cs#L128

Symbol is of type `UnsupportedMetadataTypeSymbol`, namespace is null.

matkoch:
Might be that this is a Roslyn bug?

matkoch:
@herohua  Please reopen and mark as bug.

superyyrrzz:
I thought @werwolfby had fixed this. I just tested it on 2.37.1 and it can work:
![image](https://user-images.githubusercontent.com/3831744/54282903-ee4fc080-45d7-11e9-8fda-47334d2e9b3f.png)

I also tried to put it in a class with no namespace. It will log warning saying no metadata generated, instead of error. Could you reproduce the error now? @matkoch

matkoch:
I don't remember exactly, just got the notice of the issue being closed. However, the image doesn't look right exactly? `params[] namespaces` and `System.[]` is not really what I would expect.

superyyrrzz:
Just notice this... This is indeed a bug.

matkoch:
Thank you!

werwolfby:
I was sure that everythimg were fixed, I’ll recheck it next week.

yufeih:
This is fixed in the latest version:

![image](https://user-images.githubusercontent.com/511355/215404934-f395f5a4-df9f-4b72-9c06-4c22348a826e.png)

werwolfby:
Hmm...

PR #3228, opened 2018-08-19
Support for value tuples in documentation

werwolfby:
Add processing `ValueTuple` in `NameVisitor` and `ReferenceItemVisitor` to display it correctly in documentation:
- When we use `Qualified` name it will represent tuple as `System.ValueTuple<T1, T2, ...>` generic class
- In other cases it will represent tuple as C# tuple syntax: `(string prefix, string uri)`, depending on `Options`
- Implemented for C# and VB

It fixes issues: #2512 and #3211.

I didn't find any tests for this visitors. But I think we should add it (maybe in separate PR).
There is still one issue in Rolsyn: dotnet/roslyn#29390 that doesn't have proper workaround for now.
I'll try to fix the issue in Roslyn as well.


werwolfby:
It is required for nuke-build/nuke#117

src/docfx/Models/FileMappingItem.cs
```diff
@@ -30,12 +30,24 @@ public class FileMappingItem
         [JsonProperty("files")]
         public FileItems Files { get; set; }
 
+        /// <summary>
+        /// The reference glob pattern collection, with path relative to property `src`/`cwd` is value is set
+        /// </summary>
+        [JsonProperty("references")]
+        public FileItems References { get; set; }
```
superyyrrzz: Question: is this only needed when generating document from assembly? We don't need this when generating from source code?

test/Microsoft.DocAsCode.Metadata.ManagedReference.Tests/GenerateMetadataFromAssemblyTest.cs
```diff
@@ -47,5 +61,38 @@ public void TestGenerateMetadataFromAssembly()
                 Assert.Equal(new[] { "System.Object", "CatLibrary.Cat{CatLibrary.Dog{System.String},System.Int32}" }, @class.Inheritance);
             }
         }
+
+        [Fact]
+        public void TestGenerateMetadataFromAssemblyWithReferences()
```
superyyrrzz: >TestGenerateMetadataFromAssemblyWithReferences [](start = 20, length = 46)

Could you also add some test cases in `GenerateMetadataFromCSUnitTest.cs` and `GenerateMetadataFromVBUnitTest.cs`?

src/docfx/Models/FileMappingItem.cs
werwolfby: It can be possible to use it from sources as well. But the main reason is to use it for generating from assembly.

test/Microsoft.DocAsCode.Metadata.ManagedReference.Tests/GenerateMetadataFromAssemblyTest.cs
werwolfby: Sure, will do it

werwolfby:
@superyyrrzz Tests was added, there was issue that was covered by tests

werwolfby:
Branch was rebased to the latest dev

werwolfby:
@superyyrrzz Any additional comments?

src/docfx/Models/FileMappingItem.cs
superyyrrzz: Got your ideas. Looks good.

I think it may be better to move `references` from `metadata/src` to `metadata`. As you can see `metadata/src` is an array. If it has multiple objects, users may think `references` is needed for every object, as it looks like only applies to the current object in array. (Of course, I know it applies to the whole metadata section from your implementation). I'd expect the config like:
```
  "metadata": [
    {
      "src": [
        {
          "src": "../.tmp",
          "files": [
            "SomeLibrary.Common/lib/net462/*.dll"
          ],
        }
      ],
      "dest": "api/SomeLibrary.Common",
      "references": [
            "System.ValueTuple/lib/netstandard1.0/System.ValueTuple.dll"
          ]
    }
  ]
```
After this change `FileMappingItem`, we can stop glob support for it. User need to list all the required DLLs here. It will no be too long as far as we can see.

superyyrrzz: @werwolfby Sounds nice. It makes sense to keep glob support after moving `references` out. 👍

werwolfby:
@superyyrrzz, I've analyzed logic of `metadata` section in `docfx.json` file. 

And I was **completely** wrong when put `references` into `src`. 😞 

Now it is part of `metadata` itself and reuse `FileMapping` class to work with `references` as we work with `src`. In most simplest cases it allow us to specify list of references as regular array with globbing support:

```json
"references": [
  "System.ValueTuple/lib/netstandard1.0/System.ValueTuple.dll",
  "SomeOtherLib/lib/net462/*.dll"
]
```

But this do not allow us to use `./../../` in path. And we can workaround it by `src` property:

```json
"references": [
  {
    "src": "../packages",
    "files": [
      "System.ValueTuple/lib/netstandard1.0/*.dll",
      "SomeOtherLib/lib/net462/*.dll"
    ]
  }
]
```

What do you think about this? It allow me to reuse already implemented code. Allow be consistent with `src` property, and allow to simplify documentation.

werwolfby:
Or just as a single object:

```json
"references": {
    "src": "../.tmp",
    "files": [
      "System.ValueTuple/lib/netstandard1.0/*.dll"
    ]
}
```

Some documentation available [here](https://github.com/dotnet/docfx/blob/dev/src/docfx/Models/FileMapping.cs#L12-L30)

matkoch:
Any estimation when this gets released?

Issue #3229, opened 2018-08-19
Warnings as errors should be properly supported


lextm:
To extend the comment I left in #1991, docfx should support "warnings as errors" like other tools, so that CI builds can easily fail when severe warnings are detected.

It would be very subjective to discuss which warnings should really be errors ("No metadata is generated" in issues like #3225 is much more than annoying), so at least this option gives docfx users a chance to fail the build in extreme cases.

And more importantly it only requires less than 10 lines of changes.

PR #3230, opened 2018-08-19
Implemented the simplest warnings as errors.

lextm:
Added a new option as documented in #3229.

By adding `--warningsAsErrors true` to the command line, any warning would trigger a return code of `1` which can be used to fail CI builds or any other scripts relying on the return code.


yufeih:
Thank you @lextm  LGTM 👍

PR #3233, opened 2018-08-20
support warnings as errors

yufeih:
https://github.com/dotnet/docfx/issues/3229


perlun:
(Note to self and others: only available in v3, that's why it's not present in the published schema at https://json.schemastore.org/docfx)

PR #3234, opened 2018-08-20
Add schema validation to Toc parser and fix loading array during de-serialization

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/17355?src=WorkItemMention&%3Bsrc-action=artifact_link

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -977,5 +977,5 @@ outputs:
   build.log: |
     ["info","null-value","(Line: 2, Character: 3) 'tocHref' contains null value","docs/TOC.yml",2,3]
     ["error","violate-schema","(Line: 1, Character: 3) Required property 'Name' not found in JSON","docs/TOC.yml",1,3]
-    ["error","violate-schema","(Line: 5, Character: 10) The field JArray must be a string or array type with a minimum length of '1'.","docs/TOC.yml",5,10]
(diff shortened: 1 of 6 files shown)


docs/specs/toc.yml
```diff
@@ -959,3 +959,23 @@ outputs:
...
+  docs/f1/a.md:
+  docs/f1/b.md:
+outputs:
+  docs/f1/a.json:
+  docs/f1/b.json:
+  build.manifest:
+  build.log: |
+    ["info","null-value","(Line: 2, Character: 3) 'tocHref' contains null value","docs/TOC.yml",2,3]
+    ["error","violate-schema","(Line: 1, Character: 3) Required property 'Name' not found in JSON","docs/TOC.yml",1,3]
+    ["error","violate-schema","(Line: 5, Character: 10) The field JArray must be a string or array type with a minimum length of '1'.","docs/TOC.yml",5,10]
```
herohua: >The field JArray must be a string or array type with a minimum length of '1' [](start = 56, length = 76)

this error message is not so readable, shall we refine it to:
`Array {propertyName} should contain at least 1 element.`

yufeih: This message comes from data annotation library which is used by a wide range of c# model validation scenarios. I'm happy with the message template, we should just change `JArray` to `{propertyName}`

PR #3235, opened 2018-08-20
[wip] add --config command line switch

yufeih:
This PR:

- add a `--config` command line switch to allow overriding configurations using command line args
- you can specify multiple configs in command line using `--config A --config B` 
- `--config` switch take precedence over `docfx.yml` and extend in `docfx.yml`
- fixed a bug where merge happened before expand config object

This unblocks #3196

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,2 +1,5 @@
 build_script:
   - ps: ./build.ps1
+environment:
+  JSON_SCHEMA_LICENSE:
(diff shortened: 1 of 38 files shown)


yufeih:
Seems like this is different from `--config` in [hogo](https://gohugo.io/getting-started/configuration/)

yufeih:
This change needs more design considerations.

PR #3236, opened 2018-08-20
make userProfileCache name case-insensitive

928PJY:


superyyrrzz:
add a test case?

docs/specs/git.yml
```diff
@@ -100,6 +100,34 @@ outputs:
   build.manifest:
   dep.json:
 ---
+### Author is case-insensitive
```
yufeih: is there an existing test we can modify? git tests are slower to run, I'm trying to reduce the total e2e test time.

PR #3243, opened 2018-08-21
Config schema validation

live1206:
- Add schema validation for Config
- Log Config errors
- Fix some Config schema errors

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -325,7 +325,7 @@ inputs:
   docfx.yml: |
     redirections:
       docs/a.md: /absolute/path1
-    redirectionsWithoutDocumentId:
+    redirectionsWithoutId:
       docs/a.md: /absolute/path2
 outputs:
   docs/a.json:
@@ -339,7 +339,7 @@ inputs:
     redirections:
       docs/TOC.md: /absolute/path
       docs/a.md: /absolute/path1
-    redirectionsWithoutDocumentId:
+    redirectionsWithoutId:
       docs/a.md: /absolute/path2
 outputs:
   docs/a.json:
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -332,3 +332,19 @@ outputs:
   build.log: |
     ["error","redirection-conflict","The 'docs/a.md' appears twice or more in the redirection mappings","docfx.yml"]
   build.manifest:
+---
+# Multiple errors
+inputs:
+  docfx.yml: |
+    redirections:
+      docs/TOC.md: /absolute/path
+      docs/a.md: /absolute/path1
+    redirectionsWithoutDocumentId:
+      docs/a.md: /absolute/path2
+outputs:
+  docs/a.json:
+  build.log: |
+    ["error","invalid-redirection","The 'docs/TOC.md' shouldn't belong to redirections since it's a TableOfContents","docfx.yml"]
+    ["error","redirection-conflict","The 'docs/a.md' appears twice or more in the redirection mappings","docfx.yml"]
+  build.manifest:
+---
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -274,6 +274,22 @@ inputs:
 outputs:
   build.manifest:
 ---
+# expand config before merge
+inputs:
+  docfx.yml: |
+    content: '**/*.md'
+    extend:
+      extend.yml
+  extend.yml: |
+    content: 
+      include: a.md
+      exclude: b.md
+  a.md:
+  b.md:
+outputs:
+  a.json:
+  build.manifest:
+---
 # `/` and `\` in path will be normalized
 inputs:
   docfx.yml: |
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -299,6 +299,17 @@ outputs:
     { "content": "<p><a href=\"b.md\">b</a></p>" }
   build.manifest:
 ---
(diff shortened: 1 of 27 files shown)


docs/specs/config.yml
```diff
@@ -325,10 +325,26 @@ inputs:
   docfx.yml: |
     redirections:
       docs/a.md: /absolute/path1
-    redirectionsWithoutId:
+    redirectionsWithoutDocumentId:
```
yufeih: let's stick with `redirectionsWithoutId`, this feels like a breaking change.

src/docfx/config/Config.cs
```diff
@@ -146,97 +146,111 @@ public static Config Load(string docsetPath, CommandLineOptions options, bool ex
         /// Load the config if it exists under <paramref name="docsetPath"/>
         /// </summary>
         /// <returns>Whether config exists under <paramref name="docsetPath"/></returns>
-        public static bool LoadIfExists(string docsetPath, CommandLineOptions options, out Config config, bool extend = true, RestoreMap restoreMap = null)
+        public static bool LoadIfExists(List<Error> errors, string docsetPath, CommandLineOptions options, out Config config, bool extend = true, RestoreMap restoreMap = null)
```
yufeih: How about `(errors, config) LoadIfExists(...)`? when it does not exist, return `(errors, null)`

src/docfx/config/Config.cs
```diff
@@ -146,97 +146,111 @@ public static Config Load(string docsetPath, CommandLineOptions options, bool ex
...
-            ValidateLocale(config);
-            ValidateContributorConfig(config.Contribution, docsetPath);
+            var errors = new List<Error>();
+            errors.AddRange(ValidateLocale(config));
+            errors.AddRange(ValidateContributorConfig(config.Contribution, docsetPath));
+            return errors;
         }
 
-        private static void ValidateLocale(Config config)
+        private static List<Error> ValidateLocale(Config config)
```
yufeih: can this be implemented as an DataAnnotation extension, like `LocaleAttribute` ?

src/docfx/config/Config.cs
```diff
@@ -146,97 +146,111 @@ public static Config Load(string docsetPath, CommandLineOptions options, bool ex
...
             catch (CultureNotFoundException e)
             {
-                throw Errors.InvalidLocale(config.Locale).ToException(e);
+                errors.Add(Errors.InvalidLocale(config.Locale));
             }
+            return errors;
         }
 
-        private static void ValidateContributorConfig(ContributionConfig config, string docsetPath)
+        private static List<Error> ValidateContributorConfig(ContributionConfig config, string docsetPath)
```
yufeih: can this be implemented as `ExternalFileAttribute` ?

src/docfx/restore/Restore.cs
```diff
@@ -17,22 +17,24 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
         {
             // Restore has to use Config directly, it cannot depend on Docset,
             // because Docset assumes the repo to physically exist on disk.
+            var errors = new List<Error>();
```
yufeih: this is not thread safe if `RestoreDocset` runs in parallel.

src/docfx/config/Config.cs
live1206: Thought about that as well, but we do use the bool return value to check if config file exists to control diff flows. So seems better to keep the signature and pass in errors list

---
In reply to: [211462329](https://github.com/dotnet/docfx/pull/3243#discussion_r211462329) [](ancestors = 211462329)

src/docfx/restore/Restore.cs
live1206: Right, will update it to ConcurrentBag

---
In reply to: [211463029](https://github.com/dotnet/docfx/pull/3243#discussion_r211463029) [](ancestors = 211463029)

src/docfx/config/Config.cs
yufeih: or if can be `bool TryLoad(out var (errors, config))`, just don't let it populate an errors list in an API boundary like here, it is easy to introduce mutli threading problems like below.

---
In reply to: [211463455](https://github.com/dotnet/docfx/pull/3243#discussion_r211463455) [](ancestors = 211463455,211462329)

live1206: ok, sounds better

---
In reply to: [211463951](https://github.com/dotnet/docfx/pull/3243#discussion_r211463951) [](ancestors = 211463951,211463455,211462329)

src/docfx/config/Config.cs
live1206: Will create a new PR for this

---
In reply to: [211462828](https://github.com/dotnet/docfx/pull/3243#discussion_r211462828) [](ancestors = 211462828)

src/docfx/config/Config.cs
live1206: Will create a separate PR for this

---
In reply to: [211462516](https://github.com/dotnet/docfx/pull/3243#discussion_r211462516) [](ancestors = 211462516)

src/docfx/restore/Restore.cs
live1206: Pass Report in RestoreDocset instead

---
In reply to: [211463509](https://github.com/dotnet/docfx/pull/3243#discussion_r211463509) [](ancestors = 211463509,211463029)

Issue #3251, opened 2018-08-22
azure-docs-pr issue: remove null value in raw-metadata


928PJY:
articles/azure-maps/map-get-information-from-coordinate.md
![image](https://user-images.githubusercontent.com/17958546/44438044-eefb1800-a5ef-11e8-9faa-813c00829e94.png)

PR #3252, opened 2018-08-22
remove null value in raw-metadatag

928PJY:
#3251

### DIFF
--- a/src/docfx/lib/JsonUtility.cs
+++ b/src/docfx/lib/JsonUtility.cs
@@ -61,6 +61,8 @@ namespace Microsoft.Docs.Build
         private static List<Error> t_schemaViolationErrors;
         [ThreadStatic]
         private static Func<DataTypeAttribute, JsonReader, object> t_transform;
+        [ThreadStatic]
+        private static string t_docsetPath;
 
         /// <summary>
         /// Serialize an object to TextWriter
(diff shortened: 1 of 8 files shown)


src/docfx/lib/JsonUtility.cs
```diff
@@ -184,6 +184,12 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
             return result;
         }
 
+        public static JObject ValidateNullValue(this JObject graph)
```
yufeih: `JToken RemoveNulls(JToken)`

src/docfx/lib/JsonUtility.cs
```diff
@@ -184,6 +184,12 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
...
+        {
+            var (_, jtoken) = ((JToken)graph).ValidateNullValue();
```
superyyrrzz: >ValidateNullValue [](start = 46, length = 17)

We need to remove null silently in LegacyMarkdown. Will this method log error if finding null?

src/docfx/lib/JsonUtility.cs
live1206: +1

---
In reply to: [211813607](https://github.com/dotnet/docfx/pull/3252#discussion_r211813607) [](ancestors = 211813607)

src/docfx/lib/JsonUtility.cs
OsmondJiang: >ValidateNullValue [](start = 46, length = 17)

why ’validate‘  changed the input? chaneg to validateAndRemoveNullValue()?

test/docfx.Test/lib/JsonUtilityTest.cs
```diff
@@ -104,6 +104,15 @@ public void TestListOfBasicClass()
             }
         }
 
+        [Fact]
+        public void TestValidateNullValue()
```
yufeih: This should have been tested, remove?

PR #3256, opened 2018-08-22
Remove duplicated line info from error message since it is presented i…

live1206:
…n Error's properties already


src/docfx/lib/log/Report.cs
```diff
@@ -94,7 +94,7 @@ private static string LegacyReport(Error error, ErrorLevel level)
             var code = error.Code;
             var message = error.Message;
             var file = error.File;
-            var line = error.Line;
+            var line = error.Range?.StartLine;
```
928PJY: >StartLine [](start = 36, length = 9)

why don't use error.line?

Issue #3259, opened 2018-08-22
azure-docs-pr toc_rel is different


OsmondJiang:
- [x] toc_rel is different when there are two tocs referencing the same file
- [x] toc_rel is not null when the file are not referenced by any other toc files

The first one is by design, we always return the first one ordered by sit path when mutiple toc files have the same distances. 

@mingwli  is looking at the second one.

@mingwli  please add some spec tests for toc_rel cases.

OsmondJiang:
the second scenario is caused by different distance logic between v2 and v3(v2 is file -> toc and v3 is toc -> file), now the behavior is consistent after consist the distance logic.  @mingwli  @928PJY @herohua @yufeih

PR #3260, opened 2018-08-22
Fix toc rel

mingwli:
1.  Fix toc_rel in metadata

    1. calculate relative path from doc-file=>toc-file
    2. when doc isn't referenced, only look up in parent folders
    3. don't export toc_rel when toc is null
2.  Add toc e2e test cases

    1. orphan file only look up in parent folders
    2. file referenced by multiple files sorted alphabetically

FYI V2 logic:
1. [GetNearestToc](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Build.Engine/SystemMetadataGenerator.cs#LC138)
2. [GetDefaultToc](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Build.Engine/SystemMetadataGenerator.cs#LC124)
3. ['-'](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Common/Path/RelativePath.cs#LC528) & [MakeRelativeTo](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Common/Path/RelativePath.cs#LC126)

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -899,7 +899,7 @@ inputs:
   docs/f1/TOC.experimental.md: |
     # [Index Reference](../index.md)
     # [File Reference](../a.md)
-  docs/f1/TOC.md:
+  docs/f1/TOC.md: |
     # [Index Reference](../index.md)
   docs/index.md:
   docs/a.md:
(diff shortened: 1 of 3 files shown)


docs/specs/toc.yml
```diff
@@ -979,3 +979,61 @@ outputs:
...
+outputs:
+  docs/dir1/a/TOC.json: |
+    {
+      "items": [
+        {
+          "toc_title": "TOC Ref",
+          "href": "../c/file1"
+        }
+      ]
+    }
```
OsmondJiang: don't need show toc's output

mingwli:
#3259

PR #3273, opened 2018-08-24
Add InlineMarkdown attribute

live1206:


src/docfx/build/page/Markup.cs
```diff
@@ -80,6 +88,18 @@ internal static class Markup
             }
         }
 
+        private static string ToHtml(string markdown, MarkdownPipelineType pipelineType)
```
yufeih: make it `MarkdownPipeline GetPipeline(pipelineType)` to be shareable with `Parse`

Issue #3263, opened 2018-08-23
path too long issue during restore using git


OsmondJiang:
git config core.longpaths true

PR #3282, opened 2018-08-24
Restore git path too long

mingwli:
1. turn on core.longpaths on clone;
2. turn on core.longpaths before fetch;
[git-config bool case-insensitive](https://git-scm.com/docs/git-config#git-config-boolean)

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -132,8 +132,6 @@ namespace Microsoft.Docs.Build
             return ExecuteNonQuery(cwd, cmd);
         }
 
-
-
         /// <summary>
         /// Fetch update from remote
         /// </summary>
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
+using System.Text;
 using System.Threading.Tasks;
 
 namespace Microsoft.Docs.Build
@@ -118,15 +119,20 @@ namespace Microsoft.Docs.Build
         public static Task Clone(string cwd, string remote, string path, string branch = null, bool bare = false)
         {
             Directory.CreateDirectory(cwd);
+
+            // clone with configuration core.longpaths turned-on
+            // https://stackoverflow.com/questions/22575662/filename-too-long-in-git-for-windows#answer-40909460
             var cmd = string.IsNullOrEmpty(branch)
-                ? $"clone {remote} \"{path.Replace("\\", "/")}\""
-                : $"clone -b {branch} --single-branch {remote} \"{path.Replace("\\", "/")}\"";
+                ? $"clone -c core.longpaths=true {remote} \"{path.Replace("\\", "/")}\""
+                : $"clone -c core.longpaths=true -b {branch} --single-branch {remote} \"{path.Replace("\\", "/")}\"";
 
             if (bare)
                 cmd += " --bare";
 
             return ExecuteNonQuery(cwd, cmd);
         }
+
+
 
         /// <summary>
         /// Fetch update from remote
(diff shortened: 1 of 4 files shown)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -118,16 +119,21 @@ public static bool IsRepo(string path)
...
+                ? $"clone -c core.longpaths=true {remote} \"{path.Replace("\\", "/")}\""
+                : $"clone -c core.longpaths=true -b {branch} --single-branch {remote} \"{path.Replace("\\", "/")}\"";
 
             if (bare)
                 cmd += " --bare";
 
             return ExecuteNonQuery(cwd, cmd);
         }
 
+
```
OsmondJiang: remove two empty lines

src/docfx/lib/git/GitUtility.cs
```diff
@@ -179,6 +185,13 @@ public static Task AddWorkTree(string cwd, string commitHash, string path)
         public static Task PruneWorkTrees(string cwd)
             => ExecuteNonQuery(cwd, $"worktree prune");
 
+        /// <summary>
+        /// Turn core.longpaths to true in current git repository.
```
OsmondJiang: the description is not matched with the method

mingwli:
#3263

src/docfx/lib/git/GitUtility.cs
```diff
@@ -118,9 +119,12 @@ public static bool IsRepo(string path)
         public static Task Clone(string cwd, string remote, string path, string branch = null, bool bare = false)
         {
             Directory.CreateDirectory(cwd);
+
+            // clone with configuration core.longpaths turned-on
+            // https://stackoverflow.com/questions/22575662/filename-too-long-in-git-for-windows#answer-40909460
             var cmd = string.IsNullOrEmpty(branch)
-                ? $"clone {remote} \"{path.Replace("\\", "/")}\""
-                : $"clone -b {branch} --single-branch {remote} \"{path.Replace("\\", "/")}\"";
+                ? $"clone -c core.longpaths=true {remote} \"{path.Replace("\\", "/")}\""
```
OsmondJiang: can we add a spec test for that?

src/docfx/restore/RestoreGit.cs
```diff
@@ -93,8 +93,11 @@ async Task FetchOrCloneRepo()
             {
                 if (GitUtility.IsRepo(restoreDir))
                 {
+                    // turn on core.longpaths before fetch
+                    await GitUtility.SetCoreLongPaths(restorePath, true);
```
yufeih: >SetCoreLongPaths [](start = 37, length = 16)

if the repo is cloned using `-c core.longpaths=true`, do we still need to configure it on fetch?

src/docfx/lib/git/GitUtility.cs
yufeih: +1

---
In reply to: [212568157](https://github.com/dotnet/docfx/pull/3282#discussion_r212568157) [](ancestors = 212568157)

src/docfx/lib/git/GitUtility.cs
```diff
@@ -6,6 +6,7 @@
 using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
+using System.Text;
```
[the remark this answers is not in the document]
mingwli: it's not used, will remove it.

src/docfx/restore/RestoreGit.cs
mingwli: If the repo is migrated from a `core.longpaths=false` state, it needs to be switched on. 
Do you recommend check `core.longpaths` before fetch and decide if it needs to be turned on?

yufeih: no, I was thinking we may safely assume this repo is totally managed by *us*, so it won't have a `core.longpaths=false` state given we haven't shipped v3 yet.

OsmondJiang: for migration? some repo were already cloned by us using core.longpaths=false

PR #3285, opened 2018-08-27
support expanded and maintainContext

OsmondJiang:


src/Microsoft.Docs.Template/Models/TableOfContents.cs
```diff
@@ -24,6 +24,12 @@ public class TableOfContentsItem
         [JsonProperty(PropertyName = "href")]
         public string Href { get; set; }
 
+        [JsonProperty(PropertyName = "expanded")]
+        public bool? Expanded { get; set; }
```
yufeih: just `bool` with a default value

live1206:
https://review.docs.microsoft.com/en-us/new-hope/specs/ops/toc_schema_validation?branch=master

src/Microsoft.Docs.Template/Models/TableOfContents.cs
OsmondJiang: it will pollute the normal toc output

PR #3291, opened 2018-08-27
[DataSchema] produces JSON, [PageSchema] produces HTML

yufeih:
Some of our SDP only produces JSON model as output, e.g. ContextObject. This PR let schema express the desired output `JSON` vs `HTML` using `DataSchemaAttribute` vs `PageSchemaAttribute`.

This change requires that mime headers are read upfront, when constructing `Document`, a specialized ReadMime method is created for JSON and Yaml, it takes a `TextReader`, does not load everything into memory, so the performance should be OK.

#3050


src/docfx/docset/Document.cs
```diff
@@ -20,6 +20,16 @@ internal class Document
...
 
+        /// <summary>
+        /// Gets the MIME type specifed in YAML header or JSON $schema.
+        /// </summary>
+        public string Mime { get; }
+
+        /// <summary>
+        /// Gets the schema identified by <see cref="Mime"/>.
+        /// </summary>
+        public DataSchema Schema { get; }
```
OsmondJiang: >DataSchema [](start = 15, length = 10)

it's confusing me for data schema and data schema attribute....

src/docfx/docset/Document.cs
```diff
@@ -20,6 +20,16 @@ internal class Document
         /// </summary>
         public ContentType ContentType { get; }
 
+        /// <summary>
+        /// Gets the MIME type specifed in YAML header or JSON $schema.
+        /// </summary>
+        public string Mime { get; }
```
OsmondJiang: >Mime [](start = 22, length = 4)

if it's equals to Schema, we don't need it?

src/docfx/docset/Document.cs
OsmondJiang: rename to Schema?

---
In reply to: [212891103](https://github.com/dotnet/docfx/pull/3291#discussion_r212891103) [](ancestors = 212891103)

docs/specs/schema.yml
```diff
@@ -133,3 +133,20 @@ outputs:
...
+  docs/a.md: |
+    link to [data.yml]
+    link to [page.yml]
+  docs/data.yml: |
+    #YamlMime:TestData
+  docs/page.yml: |
+    #YamlMime:TestPage
+outputs:
+  docs/a.json: |
+    { "content": { "" } }
```
OsmondJiang: > "content": { "" } [](start = 5, length = 18)

you didn't validate the link?

yufeih: wip

src/docfx/docset/Document.cs
live1206: Agree, seems part of Schema?

---
In reply to: [212894056](https://github.com/dotnet/docfx/pull/3291#discussion_r212894056) [](ancestors = 212894056)

yufeih: There is a temporary hack to fallback to `documentType` when YamlMime is `YamlDocument`. The error reporting needs to know the correct mime type, we cannot parse YAML upfront as it is too heavy, I'll leave a TODO

PR #3292, opened 2018-08-27
Add Razor template for LandingData

live1206:
#3050


src/docfx/build/Build.cs
```diff
@@ -29,7 +29,7 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
             var (contributionErrors, contribution) = await ContributionInfo.Load(docset, options.GitToken);
             errors.AddRange(contributionErrors);
 
-            var (files, sourceDependencies) = await BuildFiles(context, docset.BuildScope, tocMap, contribution);
+            var (files, sourceDependencies) = await BuildFiles(context, docset.BuildScope, tocMap, contribution, options.Legacy);
```
yufeih: Docset already has a `Legacy` property

src/Microsoft.Docs.Template/Views/Template/LandingData.cshtml
```diff
@@ -0,0 +1,143 @@
+@using Microsoft.Docs.Build;
```
yufeih: copyright headers in cshtml as well?

live1206: sure

Issue #3150, opened 2018-08-03
azure-docs-pr issue: document_version_independent_id, document_id error


928PJY:

OsmondJiang:
https://ceapex.visualstudio.com/Engineering/All/_git/Docs.DocFX.Impact/pullrequest/2258?_a=files&path=%2Fsource%2Fazure-docs-pr.docfx.yml

OsmondJiang:
https://ceapex.visualstudio.com/Engineering/All/_git/Docs.DocFX.Impact/pullrequest/2290?path=%2F&_a=files

928PJY:
there are still some diff

OsmondJiang:
they are expected for SDP files, we always use source path to calculate the document id, @live1206  and @herohua do you agree?

OsmondJiang:
https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2745?path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Factive-directory%2Fsaas-apps%2Fworkfront-tutorial.mta.json&_a=files

OsmondJiang:
all are different under: https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2745?path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Factive-directory%2Fsaas-apps&_a=files

OsmondJiang:
the root cause for above two case is: a -> b -> c scenario is not considered in v3 logic

yufeih:
`documentId` is used by a lot of other services including GitHub based commenting. I suspect we might need to keep it as is including bugs caused by previous versions. I have two options in mind:

1. implement the buggy old behavior (not pretty)
2. add a config entry that let us selectively override `documentId` for particular documents

OsmondJiang:
there are still some difference caused by unknown reason:

https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2816?_a=files&path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Fcosmos-db%2Fquery-cheat-sheet.mta.json

https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2816?_a=files&path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Fcosmos-db%2Flocal-emulator-export-ssl-certificates.mta.json

[...]

PR #3296, opened 2018-08-27
fix document id for redirection to redirection

OsmondJiang:


OsmondJiang:
#3150

src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -32,15 +32,14 @@ public bool TryGetRedirectionUrl(string sourcePath, out string redirectionUrl)
             return false;
         }
 
-        public bool TryGetDocumentId(Document file, out (string id, string versionIndependentId) id)
+        public bool TryGetDocumentId(Document file, ref (string id, string versionIndependentId) id)
```
yufeih: use `out` to indicate it is an out parameter. `ref` are typically used to pass large structs by reference.

Issue #3096, opened 2018-07-18
Filter out duplicated errors


yufeih:
If two errors has the same code, message, file, line and column, report just one error.

This depend on our errors to have line / column information

Issue #2874, opened 2018-06-12
Add detailed line column range info to error messages


yufeih:
- [x] YAML
- [x] Markdown (https://ceapex.visualstudio.com/Engineering/_workitems/edit/72242/)
- [x] TOC

yufeih:
Diagnostic API for VSCode: https://code.visualstudio.com/docs/extensionAPI/vscode-api#Diagnostic

yufeih:
> If the start and end positions are the same, VS Code will underline with a squiggle the word at that position.
> If you want to underline with a squiggle until the end of the line, then set the character of the end position to Number.MAX_VALUE.

PR #3298, opened 2018-08-28
dedup errors, add max error count

yufeih:
Implements #3096 

- Dedup errors by error content.
- Determine max errors by DOCFX_MAX_ERRORS environment variable (should it be a config?). It controls max errors, max warnings and max infos.
- Some TOC errors do not have line info, tracked by #2874


src/docfx/lib/log/Report.cs
```diff
@@ -11,15 +11,22 @@ namespace Microsoft.Docs.Build
 {
     internal sealed class Report : IDisposable
     {
+        public static readonly int MaxErrors = int.TryParse(Environment.GetEnvironmentVariable("DOCFX_MAX_ERRORS"), out var n) ? n : 1000;
```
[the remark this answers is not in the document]
OsmondJiang: what's the point? could we add a message to tell use that the other errors are suppressed? otherwise it causes confusion

src/docfx/lib/log/Report.cs
```diff
@@ -54,27 +63,59 @@ public bool Write(Error error)
                 level = ErrorLevel.Error;
             }
 
-            if (_output != null)
+            if (ReachMaxErrors())
```
superyyrrzz: >ReachMaxErrors() [](start = 16, length = 16)

`&& _maxExceeded == 0`? to avoid call `Interlocked.Exchange` after max exceeded.

yufeih: Performance wise it I guess there aren't that much difference. If you check `_maxExceeded == 0`, then `_maxExceeded` should be marked as `volatile`. `Interlocked.Exchange` is a single CPU instruction.

---
In reply to: [213230267](https://github.com/dotnet/docfx/pull/3298#discussion_r213230267) [](ancestors = 213230267)

PR #3300, opened 2018-08-28
don't need to make href to lowercase

928PJY:


superyyrrzz:
test failed

928PJY:
fixed

---
In reply to: [416459660](https://github.com/dotnet/docfx/pull/3300#issuecomment-416459660) [](ancestors = 416459660)

PR #3303, opened 2018-08-28
store splitted user emails in memory

superyyrrzz:
To avoid duplicate split+join.

Another minor fix: if the docfx calls GitHub api for commit info, and find the commit author already contained in user profile cache, it needn't to call GitHub for user info again.

### DIFF
--- a/src/docfx/lib/github/UserProfile.cs
+++ b/src/docfx/lib/github/UserProfile.cs
@@ -1,9 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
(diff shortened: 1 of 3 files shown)


src/docfx/lib/github/UserProfile.cs
```diff
@@ -26,32 +24,6 @@ internal class UserProfile
...
-                var result = new HashSet<string>(existingEmails)
-                {
-                    newEmail,
-                };
-
-                // TODO: avoid split+join multiple times
-                return string.Join(';', result);
-            }
-        }
+        public List<string> UserEmails { get; set; } = new List<string>();
```
yufeih: we also need to save this value,

Issue #3306, opened 2018-08-29
azure-docs-pr: TOC experimental metadata is missing


OsmondJiang:
![image](https://user-images.githubusercontent.com/19990166/44759976-5294c080-ab6f-11e8-8ac3-36adeb6f04ff.png)

![image](https://user-images.githubusercontent.com/19990166/44759983-57f20b00-ab6f-11e8-9d2f-2ca91379ac51.png)

https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2816?_a=files&path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Fcosmos-db%2FTOC.experimental.mta.json

https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2816?_a=files&path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Fcosmos-db%2FTOC.json

PR #3309, opened 2018-08-29
fix bug of keeping toc metadata

OsmondJiang:


OsmondJiang:
#3306

src/docfx/build/legacy/files/LegacyTableOfContents.cs
```diff
@@ -74,6 +67,9 @@ private sealed class LegacyTableOfContentsMetadata
 
             [JsonProperty(PropertyName = "pdf_name")]
             public string PdfName { get; set; }
+
+            [JsonExtensionData]
+            public JObject Metadata { get; set; }
```
yufeih: Is `experimental` and `experimental_id` needed for non-legacy output ?

OsmondJiang: yes, it's used for AB testing

OsmondJiang: and we just pass through that, don't need to parse the metadata, it's wrote by writer

PR #3310, opened 2018-08-29
Treat //url as external url

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2816?_a=files&path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Fcognitive-services%2Ftext-analytics%2Fquickstarts%2Fcsharp.page.html

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -171,7 +171,7 @@ namespace Microsoft.Docs.Build
                     node.SetAttributeValue("data-linktype", "self-bookmark");
                     continue;
                 }
-                if (href[0] == '/' || href[0] == '\\')
+                if ((href[0] == '/' || href[0] == '\\') && !href.StartsWith("//"))
                 {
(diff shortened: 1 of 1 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -171,7 +171,7 @@ private static void AddLinkType(this HtmlNode html, string tag, string attribute
                     node.SetAttributeValue("data-linktype", "self-bookmark");
                     continue;
                 }
-                if (href[0] == '/' || href[0] == '\\')
+                if ((href[0] == '/' || href[0] == '\\') && !href.StartsWith("//"))
```
928PJY: >!href.StartsWith("//") [](start = 59, length = 22)

put `external` check before this and remove this?

yufeih: IsUrl treats `/asdf` as a valid url

Issue #3277, opened 2018-08-24
azure-docs-pr: a lot of GitHubAPI Rate Limit error


OsmondJiang:
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=10593&_a=summary&view=logs

```
2018-08-24T08:17:55.1708599Z resolve-commit-failed (0,0): Resolve commit 'a3470d3acb0f699eccdfc747b9b3658e265c017b' of repository 'MicrosoftDocs/azure-docs-pr' from GitHub failed: GitHub API rate limit exceeded
2018-08-24T08:17:55.1709033Z resolve-commit-failed (0,0): Resolve commit '421476b3e2f72a1ea5bdf7b087ab458963cf842d' of repository 'MicrosoftDocs/azure-docs-pr' from GitHub failed: GitHub API rate limit exceeded
2018-08-24T08:17:55.1709417Z resolve-commit-failed (0,0): Resolve commit '6e3f7468b5f6db6fecafc6eb06610208d04ccd27' of repository 'MicrosoftDocs/azure-docs-pr' from GitHub failed: GitHub API rate limit exceeded
2018-08-24T08:17:55.1709801Z resolve-commit-failed (0,0): Resolve commit 'c82172764ed9c1e9cf715efcbff69ddb0e620e4c' of repository 'MicrosoftDocs/azure-docs-pr' from GitHub failed: GitHub API rate limit exceeded
2018-08-24T08:17:55.1710185Z resolve-commit-failed (0,0): Resolve commit 'ec95b8b50d6f7fc737c96a05824e54af76bfb017' of repository 'MicrosoftDocs/azure-docs-pr' from GitHub failed: GitHub API rate limit exceeded
2018-08-24T08:17:55.1710575Z resolve-commit-failed (0,0): Resolve commit 'd26dfa72bdb054dcde8a8b432cdaf31635121602' of repository 'MicrosoftDocs/azure-docs-pr' from GitHub failed: GitHub API rate limit exceeded
2018-08-24T08:17:55.1710962Z resolve-commit-failed (0,0): Resolve commit '98965835c0b5dbc6bff2ada35f0b6b859b51c965' of repository 'MicrosoftDocs/azure-docs-pr' from GitHub failed: GitHub API rate limit exceeded
[...]

PR #3315, opened 2018-08-29
cache emails unbound to github users

superyyrrzz:
#3277 

azure-docs-pr `resolve-commit-failed` count reduced from 958 to 277

### DIFF
--- a/src/docfx/lib/github/UserProfileCache.cs
+++ b/src/docfx/lib/github/UserProfileCache.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Docs.Build
     {
         private readonly ConcurrentDictionary<string, UserProfile> _cacheByName;
         private readonly ConcurrentDictionary<string, UserProfile> _cacheByEmail;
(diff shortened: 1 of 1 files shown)


src/docfx/lib/github/UserProfileCache.cs
```diff
@@ -16,6 +16,7 @@ internal class UserProfileCache
     {
         private readonly ConcurrentDictionary<string, UserProfile> _cacheByName;
         private readonly ConcurrentDictionary<string, UserProfile> _cacheByEmail;
+        private readonly HashSet<string> _unboundEmails = new HashSet<string>();
```
yufeih: HashSet is not thread safe. How about reusing `_cacheByEmail` and use `email` -> `null` to indicate the email does not have a GitHub account

superyyrrzz: As offline discussed, this can be addressed when implementing updating cache.

PR #3316, opened 2018-08-29
schema determines properties included in output page model

yufeih:
Some schemas like `LandingData` does not show contributors, we should exclude it from output, and may save some time by not calculating contributors.

### DIFF
--- a/src/Microsoft.Docs.Template/Models/Attributes.cs
+++ b/src/Microsoft.Docs.Template/Models/Attributes.cs
@@ -9,35 +9,47 @@ namespace Microsoft.Docs.Build
     /// Exports a type to be processed by build pipeline into a JSON model.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class)]
-    public class DataSchemaAttribute : Attribute{ }
+    public class DataSchemaAttribute : Attribute { }
 
     /// <summary>
     /// Exports a type to be processed by build pipeline into an HTML page.
     /// </summary>
     [AttributeUsage(AttributeTargets.Class)]
-    public class PageSchemaAttribute : DataSchemaAttribute { }
+    public class PageSchemaAttribute : DataSchemaAttribute
+    {
+        public bool Contributors { get; }
+
+        public bool GitUrl { get; }
+
+        public bool DocumentId { get; }
+
+        public bool Toc { get; }
+
+        public PageSchemaAttribute(
+            bool contributors = true,
+            bool gitUrl = true,
+            bool documentId = true,
+            bool toc = true)
+        {
+            Contributors = contributors;
+            GitUrl = gitUrl;
+            DocumentId = documentId;
+            Toc = toc;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Class)]
+    public abstract class SchemaFeatureAttribute : Attribute { }
 
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
     public abstract class DataTypeAttribute : Attribute
     {
-        public Type RequiredType { get; set; }
+        public virtual Type TargetType => typeof(string);
     }
 
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
-    public class HrefAttribute : DataTypeAttribute
-    {
-        public HrefAttribute() => RequiredType = typeof(string);
-    }
+    public class HrefAttribute : DataTypeAttribute { }
 
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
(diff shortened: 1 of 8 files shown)


src/docfx/build/page/BuildPage.cs
```diff
@@ -26,49 +26,60 @@ internal static class BuildPage
...
+            model.ShowEdit = file.Docset.Config.Contribution.ShowEdit;
 
-            // TODO: add check before to avoid case failure
-            var (repoError, author, contributors, updatedAt) = await contribution.GetContributorInfo(file, metadata.Value<string>("author"));
-            if (repoError != null)
-                errors.Add(repoError);
-            var (editUrl, contentUrl, commitUrl) = contribution.GetGitUrls(file);
+            if (schema.Attribute is PageSchemaAttribute pageSchema)
+            {
+                if (pageSchema.DocumentId)
```
live1206: >if (pageSchema.DocumentId) [](start = 16, length = 26)

How about putting this block  into a SetupModel() internal method for readability?

src/Microsoft.Docs.Template/Models/LandingData.cs
```diff
@@ -5,7 +5,7 @@
 
 namespace Microsoft.Docs.Build
 {
-    [PageSchema]
+    [PageSchema(contributors: false)]
```
OsmondJiang: >false [](start = 30, length = 5)

they are not controlled by configuration?

src/Microsoft.Docs.Template/Models/Conceptual.cs
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.Docs.Build
     //   Conceptual model is more than just an html string, it also contain other properties.
     //   We currently bake these other properties into PageModel to make the output flat,
     //   but it makes conceptual kinda special. We may consider lift them outside PageModel.
+    [PageSchema]
```
OsmondJiang: >PageSchema [](start = 5, length = 10)

what if the contributor is false here but the configuration says we need contributor?

src/Microsoft.Docs.Template/Models/LandingData.cs
yufeih: by both.

---
In reply to: [213583452](https://github.com/dotnet/docfx/pull/3316#discussion_r213583452) [](ancestors = 213583452)

OsmondJiang: hmmm, i don't like two source of truth


---
In reply to: [213583681](https://github.com/dotnet/docfx/pull/3316#discussion_r213583681) [](ancestors = 213583681,213583452)

src/Microsoft.Docs.Template/Models/Conceptual.cs
yufeih: It'll not have contributors.

---
In reply to: [213583619](https://github.com/dotnet/docfx/pull/3316#discussion_r213583619) [](ancestors = 213583619)

src/docfx/build/page/BuildPage.cs
OsmondJiang: actually it's not pretty and common, what if TOC also need this logic to control the output?

---
In reply to: [213583003](https://github.com/dotnet/docfx/pull/3316#discussion_r213583003) [](ancestors = 213583003)

src/Microsoft.Docs.Template/Models/Attributes.cs
```diff
@@ -9,35 +9,47 @@ namespace Microsoft.Docs.Build
...
-        public MarkdownAttribute() => RequiredType = typeof(string);
+        public bool DocumentId { get; }
+
+        public bool Toc { get; }
+
+        public PageSchemaAttribute(
+            bool contributors = true,
+            bool gitUrl = true,
+            bool documentId = true,
+            bool toc = true)
```
superyyrrzz: change names to more like bool? maybe "enableXXX"...

yufeih: yeah

---
In reply to: [213596931](https://github.com/dotnet/docfx/pull/3316#discussion_r213596931) [](ancestors = 213596931)

src/Microsoft.Docs.Template/Models/LandingData.cs
OsmondJiang: I am thinking maybe this can more clear for output defination:

PageSchema(Ignore: output=> output.Contributors)

---
In reply to: [213584156](https://github.com/dotnet/docfx/pull/3316#discussion_r213584156) [](ancestors = 213584156,213583681,213583452)

PR #3329, opened 2018-08-30
Rawtitle

yufeih:


src/docfx/build/page/BuildPage.cs
```diff
@@ -148,17 +134,26 @@ private static (List<Error> errors, string pageType, object content, JObject met
...
 
             var metadata = obj?.Value<JObject>("metadata") ?? new JObject();
+            var title = metadata.Value<string>("title");
+
+            var model = new PageModel
+            {
+                Content = content,
+                Metadata = metadata,
+                Title = title,
+                HtmlTitle = file.Docset.Legacy ? $"<h1>{obj?.Value<string>("title")}</h1>" : null,
```
live1206: >file.Docset.Legacy [](start = 28, length = 18)

Since we have more legacy logic into BuildPage, shall we add TODO to split them out to LegacyOutput afterwards?

PR #3330, opened 2018-08-30
tolerate user name only differs in case in profile cache

superyyrrzz:


src/docfx/lib/github/UserProfileCache.cs
```diff
@@ -147,13 +147,23 @@ private UserProfileCache(IDictionary<string, UserProfile> cache, string path, Gi
             Debug.Assert(github != null);
 
             _cachePath = path;
-            _cacheByName = new ConcurrentDictionary<string, UserProfile>(cache, StringComparer.OrdinalIgnoreCase);
+            _cacheByName = ToIgnoreCaseConcurrentDictionary(cache);
```
yufeih: >ToIgnoreCaseConcurrentDictionary [](start = 27, length = 32)

use `JsonConvert.PopulateObject`

herohua: +1

---
In reply to: [213911812](https://github.com/dotnet/docfx/pull/3330#discussion_r213911812) [](ancestors = 213911812)

superyyrrzz: `JsonConvert.PopuldateObject` can only accept a JSON string. We will have no chance to normalize this JObject if we use this...

PR #3338, opened 2018-08-31
fix the bug of redirection document id

OsmondJiang:


src/docfx/cli/Program.cs
```diff
@@ -95,7 +95,7 @@ private static (string command, string docset, CommandLineOptions options) Parse
                 syntax.DefineCommand("build", ref command, "Builds a docset.");
                 syntax.DefineOption("o|output", ref options.Output, "Output directory in which to place built artifacts.");
                 syntax.DefineOption("legacy", ref options.Legacy, "Enable legacy output for backward compatibility.");
-                syntax.DefineOption("github-token", ref options.GitHubToken, "The git token used to get contribution information from GitHub API");
+                syntax.DefineOption("github-token", ref options.GitHubToken, "The github token used to get contribution information from GitHub API");
```
yufeih: GitHub

docs/specs/metadata.yml
```diff
@@ -57,6 +57,26 @@ outputs:
     { "redirectionUrl":"/sccm/mdm/understand/hybrid-mobile-device-management" }
   build.manifest:
 ---
+# generate same document id and version id with 'redirection with document id' evem it's out of build scope
```
yufeih: how about merging the other `redirection-out-of-scope` test in here? feels like this covers the other test

yufeih: evem -> even

I think we really need an automated spellchecker that integrate with GH PR

live1206: how about auto check extension within VS?

---
In reply to: [214230742](https://github.com/dotnet/docfx/pull/3338#discussion_r214230742) [](ancestors = 214230742)

PR #3346, opened 2018-08-31
Use sealed to disallow additional properties in JSON schema

yufeih:
Rename `Metadata` to `ExtensionData` to avoid confusion with the docfx metadata concept @OsmondJiang 

#3050

### DIFF
--- a/test/docfx.Test/lib/YamlUtilityTest.cs
+++ b/test/docfx.Test/lib/YamlUtilityTest.cs
@@ -451,7 +451,7 @@ Data:
     B: 1
     C: c
     E: e", typeof(ClassWithNestedTypeContainsJsonExtensionData))]
-        public void TestObjectTypeWithJsonExtensionData(string json, Type type)
+        public void TestSealedObjectTypeWithJsonExtensionData(string json, Type type)
         {
             var (_, token) = YamlUtility.Deserialize(json);
             var (errors, value) = JsonUtility.ToObject(token, type);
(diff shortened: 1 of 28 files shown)


live1206:
We need to update the unit tests accordingly

tools/CreateJsonSchema/Program.cs
```diff
@@ -63,7 +67,53 @@ static IEnumerable<(Type type, string name)> GetSchemas()
...
+    static void GenerateJSchema(JSchemaGenerator generator, Type type, string name)
+    {
+        var schema = generator.Generate(type, rootSchemaNullable: false);
+
+        VerifyContract(generator.ContractResolver, type);
+
+        File.WriteAllText(Path.Combine("schemas", name + ".json"), schema.ToString());
+    }
+
+    private static void VerifyContract(IContractResolver resolver, Type type, HashSet<Type> set = null)
```
live1206: >null [](start = 98, length = 4)

Set default value as new HashSet<Type>()?

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -546,13 +546,13 @@ public class ClassWithMoreMembers : BasicClass
             public string ValueRequired { get; set; }
         }
 
-        public class ClassWithJsonExtensionData : BasicClass
+        public sealed class ClassWithJsonExtensionData : BasicClass
         {
             [JsonExtensionData]
```
live1206: >JsonExtensionData [](start = 13, length = 17)

Disallowed with sealed?

Issue #3322, opened 2018-08-29
Improve toc_rel mapping when doc referenced by multiple TOCs


mingwli:
is lexicographical sort the best solution?  (for discussion)
[PR](https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/2868?_a=files)

yufeih:
lexicographical sort always picks the first TOC, in azure's case, it is almost always _active_directory_ ...

OsmondJiang:
@yufeih , no, it's also related to the toc folder

yufeih:
git is using [levenshtein distance](https://github.com/git/git/blob/53f9a3e157dbbc901a02ac2c73346d375e24978c/levenshtein.c) to caculate similarities between words.

#3201

mingwli:
[Latest changes preview](https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/3394?_a=files)

mingwli:
Output the TOC map file and check if word-level levenshtein is reasonable.  

Consider switch on/off configuration.

mingwli:
azure-docs-pr, Documents that have more than one TOCs referencing to itself.
[toc_map.json.txt](https://github.com/dotnet/docfx/files/2387323/toc_map.json.txt)

mingwli:
Files that have more than one candidates after ordering by parent/sub directory count.  
[to_discuss.toc_map.json.txt](https://github.com/dotnet/docfx/files/2387343/to_discuss.toc_map.json.txt)

PR #3354, opened 2018-09-03
Add levenshtein distance and UTs

mingwli:
[wiki levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) (there's DP implementation pseudo code in the middle)  
#3322

### DIFF
--- a/src/docfx/lib/StringUtility.cs
+++ b/src/docfx/lib/StringUtility.cs
@@ -1,0 +1,65 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Linq;
+
+namespace Microsoft.Docs.Build
+{
+    internal static class StringUtility
+    {
+        /// <summary>
+        /// Calculate the Levenshtein Distance from src to target.
+        /// </summary>
+        /// <param name="src"> The source string </param>
+        /// <param name="target">The target string </param>
+        /// <returns>Levenshtein Distance</returns>
+        public static int LevenshteinDistance(this string src, string target)
+        {
+            // for all i and j, matrix[i,j] will hold the Levenshtein distance between
+            // the first i characters of source and the first j characters of target
+            // note that matrix has (src + 1) * (target + 1) values
+            int srcLength = string.IsNullOrEmpty(src) ? 0 : src.Length;
+            int targetLength = string.IsNullOrEmpty(target) ? 0 : target.Length;
+
+            if (srcLength == 0)
+            {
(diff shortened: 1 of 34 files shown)


src/docfx/lib/StringUtility.cs
```diff
@@ -0,0 +1,65 @@
...
+            if (srcLength == 0)
+            {
+                return targetLength;
+            }
+            if (targetLength == 0)
+            {
+                return srcLength;
+            }
+
+            int[,] matrix = new int[srcLength + 1, targetLength + 1];
```
yufeih: can we reuse this buffer using [System.Buffers.ArrayPool<T>](https://github.com/dotnet/corefx/blob/master/src/System.Buffers/src/System/Buffers/ArrayPool.cs) ?

src/docfx/lib/StringUtility.cs
```diff
@@ -0,0 +1,65 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Linq;
+
+namespace Microsoft.Docs.Build
+{
+    internal static class StringUtility
```
yufeih: consider calling this `Levenshtein.GetLevenshteinDistance` and does not make it an extension method

src/docfx/lib/StringUtility.cs
```diff
@@ -0,0 +1,65 @@
...
+                {
+                    int cost = src[i - 1] == target[j - 1] ? 0 : 1;
+                    matrix[i, j] = new int[]
+                    {
+                        matrix[i - 1, j] + 1,
+                        matrix[i, j - 1] + 1,
+                        matrix[i - 1, j - 1] + cost,
+                    }.Min();
+                }
+            }
```
superyyrrzz: What about use recursion? We can choose `matrix[i - 1, j - 1] + cost` for sure when `cost==0`. Thus, in best case (src=target), complexity can be reduced from quadratic to linear.

src/docfx/lib/StringUtility.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                matrix[0, j] = j;
+            }
+
+            for (int j = 1; j <= targetLength; j++)
+            {
+                for (int i = 1; i <= srcLength; i++)
+                {
+                    int cost = src[i - 1] == target[j - 1] ? 0 : 1;
+                    matrix[i, j] = new int[]
```
yufeih: don't alloc a new `int[]` here, you could use a tuple

PR #3355, opened 2018-09-03
 HtmlAttribute should return a string

yufeih:
Fixes `index.yml` build errors found in impact test


live1206:
Need to update json schema locally to pass CI

PR #3358, opened 2018-09-03
skip schema check

OsmondJiang:


src/docfx/build/legacy/manifest/LegacyManifest.cs
```diff
@@ -71,6 +71,7 @@ public static List<(LegacyManifestItem manifestItem, Document doc)> Convert(Docs
                             Type = GetType(document.ContentType),
                             Output = output,
                             SkipNormalization = !(document.ContentType == ContentType.Resource),
+                            SkipSchemaCheck = !(document.ContentType == ContentType.Resource),
```
live1206: >!(document.ContentType == ContentType.Resource) [](start = 46, length = 47)

why not just != instead of !(==) ?

OsmondJiang: personal preference

PR #3363, opened 2018-09-04
Improve json/yaml error line info

yufeih:
- Add [`in`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/in-generic-modifier) before `Range` to speed up passing large struct
- Move error line column info to `Error.Range` property
- Improve line/column/key detection

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using YamlDotNet.Core;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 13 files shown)


src/docfx/Errors.cs
```diff
@@ -64,20 +63,17 @@ public static Error DownloadFailed(string url, string message)
...
-        }
+        public static Error YamlDuplicateKey(in Range range, string key)
+            => new Error(ErrorLevel.Error, "yaml-duplicate-key", $"Key '{key}' is already defined, remove the duplicate key.", range: range);
 
         public static Error InvalidYamlHeader(Document file, Exception ex)
             => new Error(ErrorLevel.Warning, "invalid-yaml-header", ex.Message, file.ToString());
 
-        public static Error JsonSyntaxError(Exception ex)
-            => new Error(ErrorLevel.Error, "json-syntax-error", ex.Message);
+        public static Error JsonSyntaxError(string message, string path, int line, int column)
```
live1206: >int line, int column [](start = 73, length = 20)

send in Range as well?

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -180,11 +180,18 @@ private static void VerifyFile(string file, string content)
...
                     // TODO: Configure github token in CI to get rid of github rate limit,
                     // then we could remove the wildcard match
-                    Assert.Matches("^" + Regex.Escape(string.Join("\n", expected)).Replace("\\*", ".*") + "$", string.Join("\n", actual));
+                    if (expected.Contains("*"))
+                    {
+                        Assert.Matches("^" + Regex.Escape(string.Join("\n", expected)).Replace("\\*", ".*") + "$", string.Join("\n", actual));
+                    }
+                    else
+                    {
+                        Assert.Equal(expected, actual);
```
live1206: >Equal [](start = 31, length = 5)

It seems like collection equal error message would truncate the list content:
```
Assert.Equal() Failure
Expected: Int32[] [1, 2, 3, 4, 5, ...]
Actual:   Int32[] [1, 2, 3, 4, 5, ...]
```

PR #3368, opened 2018-09-04
fix manifest items

OsmondJiang:


OsmondJiang:
@live1206

src/docfx/build/legacy/manifest/LegacyManifest.cs
```diff
@@ -25,7 +25,9 @@ public static List<(LegacyManifestItem manifestItem, Document doc)> Convert(Docs
 
                         var output = new LegacyManifestOutput
                         {
-                            MetadataOutput = new LegacyManifestOutputItem
+                            MetadataOutput = document.IsSchemaData
+                            ? null
```
live1206: >null [](start = 30, length = 4)

If I understand correctly, we are missing a metadata output for "iot-hub/rc/rc.yml", which is a schema data document. And here we set Metadata output to null for schema data document?

PR #3370, opened 2018-09-04
Add error for git-clone-failure, consolidate need-restore error

yufeih:
#2863


src/docfx/restore/RestoreMap.cs
```diff
@@ -11,7 +11,7 @@ internal class RestoreMap
         private readonly RestoreLock _restoreLock;
 
         public RestoreMap(string docsetPath)
-            : this(RestoreLocker.Load(docsetPath).Result)
+            : this(RestoreLocker.Load(docsetPath).GetAwaiter().GetResult())
```
live1206: >RestoreLocker.Load(docsetPath).GetAwaiter().GetResult() [](start = 19, length = 55)

Is it possible to call "await" at where we call this constructor and then put the result into the other constructor instead?

yufeih: This particular call stack has too many dependencies, used by `Config.Load`, so it currently sync waits. 

This change is to ensure Exception type are the same. Task.Result will wrap exceptions.

---
In reply to: [215170582](https://github.com/dotnet/docfx/pull/3370#discussion_r215170582) [](ancestors = 215170582)

PR #3371, opened 2018-09-04
Fix for array field with nonconvertible nested type

live1206:
Test cases and fix for nonconvertible nested type


live1206:
https://github.com/JamesNK/Newtonsoft.Json/issues/1816

PR #3377, opened 2018-09-06
Disallow additional properties if the model class is sealed during de-serialization

live1206:


tools/CreateJsonSchema/Program.cs
```diff
@@ -87,21 +87,6 @@ private static void VerifyContract(IContractResolver resolver, Type type, HashSe
         var contract = resolver.ResolveContract(type);
         if (contract is JsonObjectContract objectContract)
         {
-            if (!type.IsSealed && objectContract.ExtensionDataGetter == null)
```
yufeih: If we don't have a member with [JsonExtensionData] on a non-sealed type, unknown members will be discarded after `JSON -> Object -> JSON` roundtrip. This check helps avoid that.

live1206: right, adding it back

---
In reply to: [215483605](https://github.com/dotnet/docfx/pull/3377#discussion_r215483605) [](ancestors = 215483605)

PR #3384, opened 2018-09-07
Allow mutex name to be any unique string

yufeih:
This PR relax interprocess mutex name to be any string, by hashing the input name. We will be using file mutex to sync access to other resources outside of restore. The original implementation left a lots of empty folders under `appdata/file-mutex` directory.

This PR also implement wait notification and retry duration back off. A warning message is print to the console every 30 seconds when waiting for a mutex.

### DIFF
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -94,19 +94,16 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Create a file mutex to lock a resource/action
         /// </summary>
-        /// <param name="mutexFileRelativePath">The mutex file relative path</param>
+        /// <param name="mutexName">A globbaly unique mutext name</param>
         /// <param name="action">The action/resource you want to lock</param>
-        /// <param name="retry">The retry count, default is 600 times</param>
-        /// <param name="retryTimeSpanInterval">The retry interval, default is 1 seconds</param>
-        /// <returns>The task status</returns>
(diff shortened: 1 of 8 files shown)


src/docfx/lib/ProcessUtility.cs
```diff
@@ -121,20 +119,34 @@ public static bool IsNotFound(Win32Exception ex)
...
         }
 
-        private static async Task<FileStream> AcquireFileMutex(string lockPath, int retry, TimeSpan retryTimeSpanInterval)
+        private static async Task<FileStream> AcquireFileMutex(string mutexName, string lockPath)
         {
-            var retryCount = 0;
+            var retryDelay = 100;
+            var lastWait = DateTime.UtcNow;
+
             while (true)
```
OsmondJiang: >true [](start = 19, length = 4)

retry forever?

yufeih: Yes. I tend to let build server having a total `docfx` timeout like an hour. It helps with all the other cases where we might stuck, like a code bug that causes deadlock. 

When you use other lock mechanism, like `lock (obj)`, there is no timeout by default.

---
In reply to: [216057373](https://github.com/dotnet/docfx/pull/3384#discussion_r216057373) [](ancestors = 216057373)

superyyrrzz: What about keep timeout as a parameter (default to forever) in utility layer, and let business layer to decide the time out? 

---
In reply to: [216180485](https://github.com/dotnet/docfx/pull/3384#discussion_r216180485) [](ancestors = 216180485,216057373)

PR #3395, opened 2018-09-10
Consolidate JsonUtility

yufeih:
- Make `SchemaValidationAndTransformConverter` operate on serialization result value (don't parse JArray twice)
- Consolidate error handling. Add `Path` to all serialization messages. Use Json.NET error message for all type conversion errors.
- Attach source info to yaml null value nodes.
- Some minor renames

Some UT are note updated yet


superyyrrzz:
ut failed

src/docfx/lib/JsonUtility.cs
```diff
@@ -201,11 +201,14 @@ public static (List<Error>, T) ToObject<T>(JToken token, string docsetPath = nul
 
             void HandleError(object sender, Newtonsoft.Json.Serialization.ErrorEventArgs args)
             {
-                if (args.CurrentObject == args.ErrorContext.OriginalObject
```
live1206: >args.CurrentObject == args.ErrorContext.OriginalObject [](start = 20, length = 54)

This is needed to handle the error once
[https://www.newtonsoft.com/json/help/html/SerializationErrorHandling.htm](https://www.newtonsoft.com/json/help/html/SerializationErrorHandling.htm)

src/docfx/lib/JsonUtility.cs
```diff
@@ -251,51 +258,36 @@ public static JObject Merge(JObject first, IEnumerable<JObject> objs)
             return result;
         }
 
-        internal static (List<Error>, JToken) ValidateNullValue(this JToken token)
+        public static (List<Error>, JToken) RemoveNulls(this JToken token)
```
live1206: >RemoveNulls [](start = 44, length = 11)

how about LogAndRemoveNulls?

docs/specs/toc.yml
```diff
@@ -952,7 +952,7 @@ outputs:
   docs/TOC.json:
   build.manifest:
   build.log: |
-    ["info","null-value","'tocHref' contains null value","docs/TOC.yml",3,3]
+    ["info","null-value","'[0].tocHref' contains null value","docs/TOC.yml",3,3]
```
live1206: >[0].tocHref [](start = 27, length = 11)

Seems better to put just "tockHref" here, and add a path info like the others

PR #3410, opened 2018-09-12
Take "Path" away from error message, set up a property within Error.cs instead

live1206:
- Add JsonPath to Error
- Remove duplicated tests from YamlUtilityTest since they have been tested in JsonUtilityTest


live1206:
Now we use ExtensionDataSetter to handle unknown field types, there is one issue needs more investigation:
While there is a field mismatch with the contract type, Json.Net would treat it as extension data, which will cause additional errors.

Issue #3409, opened 2018-09-12
azure-docs-pr: Experimental TOC pdf_name


mingwli:
In OPS:  
when it's an experimental TOC, file goes like 'xx/xx/toc.experimental.json', it's not well treated.  
[src](https://ceapex.visualstudio.com/Engineering/All/_git/OpenPublishing.Build?path=%2FLibraries%2FPdfUtility%2FPdfUrlUtility.cs&version=GBdevelop)
![image](https://user-images.githubusercontent.com/42199316/45401380-f2c80a80-b682-11e8-9e23-18fd13bca4c5.png)
![image](https://user-images.githubusercontent.com/42199316/45402965-ec3d9100-b68a-11e8-844d-ed3897a07fde.png)
  
The template doesn't modify the pdf-related property in TOC metadata, it only uses it for rendering.

yufeih:
remove `pdf_name`. we only use `pdf_absolute_path` for download

mingwli:
in `templates.docs.msft`
![image](https://user-images.githubusercontent.com/42199316/45472901-22047780-b768-11e8-91fc-b03cf026c516.png)
![image](https://user-images.githubusercontent.com/42199316/45472945-3fd1dc80-b768-11e8-8b82-b614b10aa31f.png)
![image](https://user-images.githubusercontent.com/42199316/45473011-655ee600-b768-11e8-8471-9e43c965d74c.png)  
so don't generate `pdf_name` in legacy conversion.

mingwli:
[Impact PR](https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/3562?_a=files)  
pdf_name added to toc metadata blacklist.

PR #3414, opened 2018-09-12
experimental pdf name same as v2

mingwli:
#3409


OsmondJiang:
Can we firstly confirmed the v2 is correct behavior?

Issue #3412, opened 2018-09-12
TableOfContentsInputItem Expanded nullable


mingwli:
bool with `DefaultValueHandling = DefaultValueHandling.Ignore` will lead to missing expanded field on serialization.  
change it to nullable.  
https://review.docs.microsoft.com/en-us/new-hope/specs/ops/toc_schema_validation?branch=master  
![image](https://user-images.githubusercontent.com/42199316/45414258-80205480-b6ad-11e8-8d3b-f405e0d53b75.png)

yufeih:
This should be as design. `{ expanded: false }` should be equivalent to `{ }`

PR #3415, opened 2018-09-12
make expanded property nullable

mingwli:
#3412


src/Microsoft.Docs.Template/Models/TableOfContents.cs
```diff
@@ -28,7 +28,7 @@ public class TableOfContentsItem
         public string TocHref { get; set; }
 
         [JsonProperty(PropertyName = "expanded", DefaultValueHandling = DefaultValueHandling.Ignore)]
-        public bool Expanded { get; set; }
```
herohua: why do we need to make it nullable?

mingwli: in v2, toc file with `expanded` property explicitly given as `false`, the `expanded: false` property will be built into the result. in v3, if we use `DefaultValueHandling = DefaultValueHandling.Ignore` attribute, the false property won't be serialized, which leads to a diff.  
Though both `null` and `false` should behave the same.

PR #3422, opened 2018-09-14
Add test case for 'ms.date' path becomes '['ms.date']'

yufeih:
https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/3566?_a=files


live1206:
Handled in #3410

Issue #3201, opened 2018-08-15
Show suggestions in error messages


yufeih:
We could show suggestions in error messages. E.g. when a bookmark is not found due to misspelling, we can give writers a hint like _did you mean another-bookmark?_ 

This issue tracks all possible errors messages where we can show suggestions.

- [x] bookmark hint
- [x] ~~file name hint~~

#3115

yufeih:
Removed filename hint

Issue #3115, opened 2018-07-26
Bookmark validation


yufeih:
- [x] warn if a link with bookmark is not found on the target markdown document
- [ ] show suggestions

PR #3423, opened 2018-09-14
Bookmark validation suggestion

mingwli:


docs/specs/bookmark.yml
```diff
@@ -0,0 +1,20 @@
+---
```
yufeih: this could be inside markdown.yml, there are already bookmark test there

src/docfx/lib/Levenshtein.cs
```diff
@@ -1,26 +1,53 @@
...
+
+        /// <summary>
+        /// Get levenshtein distance of two generic type list.
+        /// </summary>
+        /// <typeparam name="T"> Generic type, e.g. string</typeparam>
+        /// <param name="src">The source type list</param>
+        /// <param name="target">The target type list</param>
+        /// <param name="comparer">Implementation of IEqualityComparer for given type</param>
+        /// <returns>Levenshtein distance of the two given list</returns>
+        public static int GetLevenshteinDistance<T>(IList<T> src, IList<T> target, IComparer<T> comparer)
```
yufeih: when is T not a string?

src/docfx/build/page/BookmarkValidator.cs
```diff
@@ -45,7 +46,12 @@ public List<(Error error, Document file)> Validate()
                     continue;
                 }
 
-                result.Add((Errors.BookmarkNotFound(file, reference, bookmark), file));
+                var suggestedBookmark = bookmarks != null ?
```
yufeih: There should be a threshold here, like <= 3 edits. 

Consider have a `string Levenshtein.GetMinDistance(string value, IEnumerable<string> candicates, int threshold = 3)` method.

mingwli:
#3115 #3201

src/docfx/Errors.cs
```diff
@@ -120,8 +120,8 @@ public static Error GitShadowClone(string repoPath)
         public static Error GitNotFound()
             => new Error(ErrorLevel.Error, "git-not-found", $"Cannot find git, install git https://git-scm.com/");
 
-        public static Error BookmarkNotFound(Document relativeTo, Document reference, string bookmark)
-            => new Error(ErrorLevel.Warning, "bookmark-not-found", $"Cannot find bookmark '#{bookmark}' in '{reference}'", relativeTo.ToString());
+        public static Error BookmarkNotFound(Document relativeTo, Document reference, string bookmark, string suggestedBookmark = null)
+            => new Error(ErrorLevel.Warning, "bookmark-not-found", $"Cannot find bookmark '#{bookmark}' in '{reference}'{(!string.IsNullOrEmpty(suggestedBookmark) ? $", did you mean '{suggestedBookmark}'" : null)}", relativeTo.ToString());
```
yufeih: nit: consider having a `Hint` helper here:

`$"Cannot find a bookmark in '{reference}'{Hint(bookmark, candidates)}."`

src/docfx/lib/Levenshtein.cs
mingwli: maybe a list of integers? but there's no such use case yet.

src/docfx/build/page/BookmarkValidator.cs
```diff
@@ -45,7 +46,14 @@ public List<(Error error, Document file)> Validate()
...
```
yufeih: nit: `bookmarks` can never be `null`, put `where` because `orderby`

src/docfx/build/page/BookmarkValidator.cs
```diff
@@ -45,7 +46,12 @@ public List<(Error error, Document file)> Validate()
                     continue;
                 }
 
-                result.Add((Errors.BookmarkNotFound(file, reference, bookmark), file));
+                var suggestedBookmark = (from bookmarkCandidate in bookmarks.ToList()
```
superyyrrzz: >.ToList() [](start = 76, length = 9)

No need?

Issue #3417, opened 2018-09-13
Support docs template


yufeih:
This is to enable static rendering, local preview using docs ui template:

- [ ] Decouple docs template from docfx/ops
- [ ] Add liquid template engine
- [ ] Add mustache template engine
- [ ] Drive SDP using JSON schema
- [ ] Apply Jint/Mustache/Liquid

#3283

PR #3424, opened 2018-09-14
Add basic liquid template engine

yufeih:
#3417 

customized liquid tags TBD


src/docfx/template/LiquidTemplate.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        private static Template LoadTemplate(string fullPath)
+        {
+            var template = Template.Parse(File.ReadAllText(fullPath));
+            template.MakeThreadSafe();
+            return template;
+        }
+
+        private static object ToHash(JToken token)
```
live1206: >ToHash [](start = 30, length = 6)

Add more tests for different scenarios? for instance, the null token

superyyrrzz:
Is this for ideal or legacy? If for legacy, it not required as liquid rendering happens on the dynamic rendering side. DHS only need the HTML applied with Mustache template.

PR #3425, opened 2018-09-14
add spec about handle credentials

superyyrrzz:


docs/designs/credential.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Request header
+
+Request header can also contains credential in `Authorization` header. e.g.
+```
+Authorization: Bearer {token}
+```
+The value of secrets object can be an object to provide more details of the request header. It can be configured in `docfx.yml` like:
+```yml
+secrets:
```
yufeih: there will be other secrets that are not determined by base URL, like GitHub access token, app insight credential, etc.

docs/designs/credential.md
```diff
@@ -0,0 +1,64 @@
...
+  "https://contoso.com/path/file":
+    token: "?sig={token}"
```
yufeih: query? `token` sounds too general

docs/designs/credential.md
```diff
@@ -27,11 +27,12 @@ It can be configured in `docfx.yml`:
 ```yml
 referenceToAResource: "https://contoso.com/path/file"
 secrets:
-  "https://contoso.com/path/file": "?sig={token}"
+  http:
```
yufeih: 👍

docs/designs/credential.md
```diff
@@ -42,14 +43,31 @@ Authorization: Bearer {token}
 The value of secrets object can be an object to provide more details of the request header. It can be configured in `docfx.yml` like:
 ```yml
 secrets:
-  "https://contoso.com/path/file":
-    token: "?sig={token}"
+  http:
+  - location: "https://contoso.com/path/file":
```
yufeih: is this _base URL_ or _exact URL_, using location feels like it is an exact URL, but we want to just qualify base URL, right?

docs/designs/credential.md
yufeih: do we auto-detect if it is auth header or query by checking if it starts with "?"

docs/designs/credential.md
```diff
@@ -0,0 +1,83 @@
...
+
+### Request header
+
+Request header can also contains credential in `Authorization` header. e.g.
+```
+Authorization: Bearer {token}
+```
+The value of secrets object can be an object to provide more details of the request header. It can be configured in `docfx.yml` like:
+```yml
+secrets:
```
yufeih:
```yml
http:
  secrets:
    token.example.com:
        query:
        headers:

git:
    token:

github:
    url:
    token:
    userCache:

appinsight:
    url:
    token:
```

docs/designs/credential.md
superyyrrzz: Changed to `baseUrl`. We can also use it for *exact URL*

docs/designs/credential.md
superyyrrzz: These short form only take it as query string. Seems hard to keep key-value pair info in headers into a single string here.

PR #3430, opened 2018-09-17
Generate CanonicalUrl

yufeih:
Add missing metadata from ideal output to be able to use docs template #3417


src/docfx/build/page/BuildPage.cs
```diff
@@ -52,6 +53,25 @@ internal static class BuildPage
             return (errors, output, dependencies.Build());
         }
 
+        private static string GetCanonicalUrl(Document file)
```
928PJY: >GetCanonicalUrl [](start = 30, length = 15)

do we still need the `_op_canonicalUrlPrefix` generating logic in  legacy?

yufeih: no. that will be a separate PR.

PR #3434, opened 2018-09-19
Apply liquid template on build if _theme is specified

yufeih:
This PR integrate liquid template into main build pipeline. Static rendering using docs liquid template is still a WIP due to missing custom tags

#3417

### DIFF
--- a/src/docfx/build/legacy/metadata/LegacyMetadata.cs
+++ b/src/docfx/build/legacy/metadata/LegacyMetadata.cs
@@ -14,8 +14,6 @@ namespace Microsoft.Docs.Build
 {
     internal static class LegacyMetadata
     {
-        private static readonly string[] s_pageMetadataBlackList = { "_op_", "absolutePath", "canonical_url", "content_git_url", "open_to_public_contributors", "fileRelativePath", "layout", "title", "redirect_url", "contributors_to_exclude", "f1_keywords" };
-
         private static readonly string[] s_metadataBlackList = { "_op_", "fileRelativePath" };
 
         public static JObject GenerataCommonMetadata(JObject metadata, Docset docset)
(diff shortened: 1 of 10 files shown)


src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -120,46 +118,9 @@ public static JObject GenerateLegacyRedirectionRawMetadata(Docset docset, PageMo
...
-                    if (key.Equals("robots", StringComparison.OrdinalIgnoreCase))
-                    {
-                        // Add to the first line
-                        key = "ROBOTS";
-                        pageMetadataOutput.Insert(0, $"<meta name=\"{key}\" content=\"{HttpUtility.HtmlEncode(content)}\" />{Environment.NewLine}");
-                    }
-                    else
-                    {
-                        pageMetadataOutput.AppendLine($"<meta name=\"{key}\" content=\"{HttpUtility.HtmlEncode(content)}\" />");
-                    }
```
superyyrrzz: Missed this part after move?

yufeih: this is intentional, the order and casing does not matter in html <meta> tags. If it is for comparison, it should be done in patch tool.

superyyrrzz: Another request is to make the "robots" upper case.

yufeih: sitemap requires it to be either `robots` or `ROBOTS`. Treat all metadata keys case insensitive. And make 'robots' as a known metadata for build.

src/docfx/template/Template.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly MustacheTemplate _mustache;
+        private readonly JavaScript _js;
+
+        public Template(string templateDir)
+        {
+            var contentTemplateDir = Path.Combine(templateDir, "ContentTemplate");
+
+            _liquid = new LiquidTemplate(templateDir);
+            _js = new JavaScript(contentTemplateDir);
+            _mustache = new MustacheTemplate(contentTemplateDir);
```
live1206: >_mustache [](start = 12, length = 9)

never used?

yufeih: yes, for now

Issue #3376, opened 2018-09-06
Support static rendering


yufeih:
The current output is a bunch of JSON models. Apply a razor template and produce a static site.

#2579  #2865

PR #3441, opened 2018-09-21
Implement missing features to support docs template

yufeih:
#3376 #3417


OsmondJiang:
> // Copyright (c) Microsoft. All rights reserved.

do we want to whole liquid to move to legacy folder?

---
Refers to: src/docfx/template/liquid/LiquidTemplate.cs:1 in 6a05550. [](commit_id = 6a0555053a85b07da89a6497fe5ebf80c4fcb506, deletion_comment = False)

src/docfx/build/Build.cs
```diff
@@ -40,6 +40,11 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
                 Legacy.ConvertToLegacyModel(docset, context, files, sourceDependencies, tocMap);
             }
 
+            if (!docset.Config.Output.Json && docset.Config.Dependencies.ContainsKey("_themes"))
```
OsmondJiang: >docset [](start = 17, length = 6)

options.Legacy &&

src/docfx/template/Template.cs
```diff
@@ -25,35 +29,55 @@ internal class Template
...
             var fileMetadata = JObject.FromObject(model.Metadata);
 
             var obj = JObject.FromObject(model);
+            obj["_op_gitContributorInformation"] = new JObject
+            {
+                ["contributors"] = obj["contributors"],
+                ["update_at"] = model.UpdatedAt.ToString(culture.DateTimeFormat.ShortDatePattern, culture),
+                ["updated_at_date_time"] = model.UpdatedAt,
+            };
+            obj["__global"] = _global;
```
OsmondJiang: don't need do that again, we already did in legacy converter

src/docfx/build/Build.cs
yufeih: Right, this could be merged with LegacyMetadata

src/docfx/template/Template.cs
```diff
@@ -93,5 +139,11 @@ public static string CreateHtmlMetaTags(JObject metadata)
 
             return result.ToString();
         }
+
+        private JObject LoadGlobalTokens(string templateDir)
+        {
+            var path = Path.Combine(templateDir, "LocalizedTokens/docs(en-us).html/tokens.json");
```
superyyrrzz: >en-us [](start = 71, length = 5)

question: this varies with loc templates. How to support this?

yufeih: just change it to `/docs({locale}).html` when we support loc

yufeih:
Everything seems good locally:

image

PR #3444, opened 2018-09-21
Fix crash when locale is invalid

yufeih:
https://github.com/dotnet/docfx/pull/3441

### DIFF
--- a/src/Microsoft.Docs.Template/Models/Attributes.cs
+++ b/src/Microsoft.Docs.Template/Models/Attributes.cs
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.ComponentModel.DataAnnotations;
-using System.Globalization;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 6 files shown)


src/docfx/config/Config.cs
```diff
@@ -26,8 +26,8 @@ internal sealed class Config
 
         /// <summary>
         /// Gets the default locale of this docset.
+        /// TODO: lower case user inputs.
         /// </summary>
-        [Locale]
```
OsmondJiang: >Locale [](start = 9, length = 6)

Some pipeline can be run without docset, how can we prevent these errors

src/Microsoft.Docs.Template/Models/Attributes.cs
```diff
@@ -36,21 +34,4 @@ public class HtmlAttribute : DataTypeAttribute { }
...
-        protected override ValidationResult IsValid(object value, ValidationContext validationContext)
-        {
-            try
-            {
-                var culture = new CultureInfo(value.ToString());
-                return null;
-            }
-            catch (CultureNotFoundException)
-            {
-                return new ValidationResult($"Locale '{value.ToString()}' is not supported.", null);
```
OsmondJiang: >ValidationResult [](start = 27, length = 16)

why not just throw DocfxException here?

live1206: +1

---
In reply to: [219429736](https://github.com/dotnet/docfx/pull/3444#discussion_r219429736) [](ancestors = 219429736)

yufeih: Data annotation is primarily used for SDP as an extension mechanism where all the logic is self-contained inside the implementation of a particular attribute.

yufeih:
According to https://github.com/dotnet/corefx/issues/6374, `CultureInfo` depend on OS, and there is no predefined list of **all valid locales**

Issue #3427, opened 2018-09-17
Detect and warn merge conflict markers


yufeih:
Merge conflict markers can leak into published pages. docfx should be able to detect that and leave a warning.

PR #3455, opened 2018-09-27
Show error when file contains merge conflict

yufeih:
#3427

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -165,7 +165,9 @@ namespace Microsoft.Docs.Build
 
         public static void CheckMergeConflictMarker(string content, string file)
         {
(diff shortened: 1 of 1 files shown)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -163,6 +163,14 @@ public static Task PruneWorkTrees(string cwd)
         public static Task<string> Revision(string cwd, string branch = "HEAD")
            => ExecuteQuery(cwd, $"rev-parse {branch}");
 
+        public static void CheckMergeConflictMarker(string content, string file)
+        {
+            if ((content.StartsWith("<<<<<<<") || content.Contains("\n<<<<<<<")) && content.Contains("\n>>>>>>>"))
```
superyyrrzz: `"<<<<<<<"` => `"<<<<<<< "` and also check `=======`? to avoid false positive as much as possible

PR #3456, opened 2018-09-27
delegate config expand errors to JsonUtility

yufeih:
This ensures we have a consistent JSON deserialization error handling experience using `JsonUtility`

#2863


src/docfx/config/Config.cs
```diff
@@ -336,8 +335,6 @@ private static JArray ToGlobConfigs(JObject obj)
 
             foreach (var (key, value) in obj)
             {
-                if (key.Contains("*"))
```
live1206: Will JsonUtility handle this error case?

yufeih: no, this check is inconsistent (we don't check every place where folders are misused as glob pattern) and incomplete (glob pattern has many other syntax like `?`, `,`, `|` )

PR #3466, opened 2018-09-30
fix IsExpired logic

superyyrrzz:


src/docfx/lib/github/GitHubUser.cs
```diff
@@ -19,7 +19,7 @@ internal class GitHubUser
 
         public bool IsValid() => Id != 0;
 
-        public bool IsExpired() => Expiry > DateTime.UtcNow;
+        public bool IsExpired() => Expiry < DateTime.UtcNow;
```
live1206: For the failing test, Expiry is not set, the default value is "1/1/0001 12:00:00 AM", which is expired. 
In this case, could we say this data should never expire since this is for test purpose?

PR #3470, opened 2018-10-02
Suppress `done in 00:00:00` message if the step runs faster than 2s

yufeih:
For most projects, we just build the whole thing in a fraction of seconds.


mingwli:
do we need to add some test cases?

yufeih:
> do we need to add some test cases?

for this simple UI change, no

PR #3475, opened 2018-10-08
Run a single e2e test with [ONLY]

yufeih:
In debug mode, if you attribute an e2e spec with `only: true`, then it will only run these tests with this attribute. Useful for debugging on a mac.


live1206:
Possible to add a flag for YAML file level as well? For instance, only want to run config.yml

yufeih:
> Possible to add a flag for YAML file level as well? For instance, only want to run config.yml

What would the syntax look like?

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -154,21 +164,28 @@ private static void DeleteDirectory(string targetDir)
...
                 var header = YamlUtility.ReadHeader(yaml) ?? "";
 
                 foreach (var ch in Path.GetInvalidPathChars())
                 {
                     header = header.Replace(ch, ' ');
                 }
-
-                yield return header.Replace('/', ' ').Replace('\\', ' ');
+#if DEBUG
+                if (yaml.Contains("only: true"))
```
live1206: How about putting the flag just in the header? Like [only] in the beginning to avoid going through the entire YAML content

test/docfx.Test/e2e/E2ESpec.cs
```diff
@@ -12,6 +12,8 @@ public class E2ESpec
 
         public readonly string Repo;
 
+        public readonly bool Only;
```
OsmondJiang: >Only [](start = 29, length = 4)

can we just use mac VS/VSCODE to support that?

yufeih: vscode does not support it.

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -154,21 +164,27 @@ private static void DeleteDirectory(string targetDir)
...
+            var sections = content.Split("\n---", StringSplitOptions.RemoveEmptyEntries);
 
             foreach (var section in sections)
             {
+                var only = false;
                 var yaml = section.Trim('\r', '\n', '-');
                 var header = YamlUtility.ReadHeader(yaml) ?? "";
-
+#if DEBUG
+                if (header.Contains("ONLY"))
```
live1206: What if "only" is needed in the header text? Can we make it more strict as in the beginning, using `header.StartsWith("[only]", StringComparison.OrdinalIgnoreCase)`?

OsmondJiang: we already have that case. check # Orphan file(not explicitly referenced in toc file)'s toc, only look up parent folder toc files

Issue #3356, opened 2018-09-03
Support uid and xref


yufeih:
Split from #3050 

- Support internal xref resolve from files with uids within the docset.
  - [x] Pre-build internal uid map for conceptual
  - [x] Pre-build internal uid map for SDP
    - [ ] ~~Define multiple Uid within one file~~
    - [x] Support XrefProperty, whose content is markdown
  - [x] Add cache for file loading to improve performance
- Support external xref resolve from xref dependency files.
  - [ ] ~~Expose xref map JSON file via public URL (for long-term, the file should be merged from multiple repositories for the same category, for instance, dotnet) #3487~~
  - Restore the previous JSON files and use them to resolve xref
    -  [x] Restore xref for conceptual
    -  [ ] ~~Restore xref for SDP(#3497)~~

PR #3479, opened 2018-10-08
Support multiple DataTypeAttribute

live1206:
#3356

### DIFF
--- a/src/Microsoft.Docs.Template/Models/Attributes.cs
+++ b/src/Microsoft.Docs.Template/Models/Attributes.cs
@@ -17,10 +17,26 @@ namespace Microsoft.Docs.Build
     [AttributeUsage(AttributeTargets.Class)]
     public class PageSchemaAttribute : DataSchemaAttribute { }
 
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = true)]
     public abstract class DataTypeAttribute : Attribute
     {
+        private int _order = -1;
+
         public virtual Type TargetType => typeof(string);
+
+        public int Order
+        {
+            get => _order;
+            set
+            {
+                if (value < 0)
+                {
+                    throw new ArgumentException(nameof(Order));
+                }
+
+                _order = value;
+            }
+        }
     }
 
     public class HrefAttribute : DataTypeAttribute { }
--- a/src/Microsoft.Docs.Template/Models/Attributes.cs
+++ b/src/Microsoft.Docs.Template/Models/Attributes.cs
@@ -20,23 +20,7 @@ namespace Microsoft.Docs.Build
     [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = true)]
     public abstract class DataTypeAttribute : Attribute
     {
-        private int _order = -1;
-
         public virtual Type TargetType => typeof(string);
-
-        public int Order
-        {
-            get => _order;
-            set
-            {
-                if (value < 0)
-                {
-                    throw new ArgumentException(nameof(Order));
-                }
-
-                _order = value;
-            }
-        }
     }
 
     public class HrefAttribute : DataTypeAttribute { }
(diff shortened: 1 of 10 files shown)


src/Microsoft.Docs.Template/Models/Attributes.cs
```diff
@@ -17,10 +17,26 @@ public class DataSchemaAttribute : Attribute { }
...
 
-    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property)]
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property, AllowMultiple = true)]
     public abstract class DataTypeAttribute : Attribute
     {
+        private int _order = -1;
+
         public virtual Type TargetType => typeof(string);
+
+        public int Order
```
mingwli: is RangeAttribute applicable here?

live1206: What is RangeAttribute used for?

---
In reply to: [223546032](https://github.com/dotnet/docfx/pull/3479#discussion_r223546032) [](ancestors = 223546032)

live1206: The idea is from [https://github.com/dotnet/corefx/blob/f5ba053ec0ff8a3a74917f47a56b2927de27473e/src/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/Schema/ColumnAttribute.cs#L47](https://github.com/dotnet/corefx/blob/f5ba053ec0ff8a3a74917f47a56b2927de27473e/src/System.ComponentModel.Annotations/src/System/ComponentModel/DataAnnotations/Schema/ColumnAttribute.cs#L47)
We need to sort the attributes based on Order, to apply them accordingly.

---
In reply to: [223546194](https://github.com/dotnet/docfx/pull/3479#discussion_r223546194) [](ancestors = 223546194,223546032)

yufeih: `Order` should be fixed (get only), markdown is always smaller than xrefproperty

src/docfx/build/page/BuildPage.cs
```diff
@@ -172,8 +172,9 @@ private static (List<Error> errors, Schema schema, PageModel model, JObject meta
 
             return (errors, schema, model, metadata);
 
-            object TransformContent(DataTypeAttribute attribute, object value, string jsonPath)
+            object TransformContent(IEnumerable<DataTypeAttribute> attributes, object value, string jsonPath)
             {
+                var attribute = attributes.OrderBy(attr => attr.Order).First();
                 if (attribute is HrefAttribute)
```
yufeih: how about using `if (attributes.Any(attr => attr is HrefAttribute)` just like in `BuildXrefMap`? This way you don't even need the `Order` property

OsmondJiang: how about split the XrefProperty and other attributes, add validation to oter attributes

---
In reply to: [223548820](https://github.com/dotnet/docfx/pull/3479#discussion_r223548820) [](ancestors = 223548820)

src/docfx/build/xref/XrefMap.cs
```diff
@@ -56,6 +57,20 @@ public void OutputXrefMap(Context context)
...
+        public static DataTypeAttribute GetNonXrefPropertyAttribute(IEnumerable<DataTypeAttribute> attributes, List<Error> errors, string jsonPath)
+        {
+            try
+            {
+                var otherAttributes = attributes.Where(attr => !(attr is XrefPropertyAttribute));
+                return otherAttributes.SingleOrDefault(attr => !(attr is XrefPropertyAttribute));
+            }
+            catch (InvalidOperationException)
+            {
+                errors.Add(Errors.InvalidMultipleAttributes(jsonPath));
```
yufeih: we only add error code for end users (writers). for errors introduced by us (developers), just let it throw and make docfx crash.

live1206: Good to know!

OsmondJiang: that means the schema is owned by developer?

---
In reply to: [223567812](https://github.com/dotnet/docfx/pull/3479#discussion_r223567812) [](ancestors = 223567812)

yufeih: yes, there is currently no way for people to extend schemas other than us.

yufeih: In the future we might add the ability to extend docfx with custom schemas, that is a different scenario and requires extending templates.

Issue #803, opened 2016-10-20
Multilanguage support


xen2:
Is there any plan for multilanguage support?

That's probably something that could work per document, for example, MyPage.md and MyPage.jp.md
(It could use default version as a fallback if no translation present)

xen2:
Note: if nothing is planned, we'll probably just build non-english doc by taking english one, overwrite the few files that are different and regenerate a full doc under a different URL (i.e. /jp instead of /en).
That way, it will keep english doc for everything not translated.

FSharpCSharp:
In my opinion, the topic of "multilingualism" is essential. Is there any ideas in doxFX how to best represent this? We have a similar requirement here that we need to provide help in different languages. Or is there Microsoft "Best Practices", which should be considered?

zwcloud:
@FSharpCSharp 
The Microsoft docs uses multiple docfx sites for multilingualism, such as [docs(en-us)](https://github.com/dotnet/docs), [docs.zh-cn](https://github.com/dotnet/docs.zh-cn) and so on.
https://github.com/dotnet?q=docs

FSharpCSharp:
@zwcloud 
That's right! But apparently Microsoft also has a special docfx template to switch the languages easily. This is unfortunately not included in the standard template, which makes it very difficult to use from my point of view. I agree that it would of course be best to benefit from the templates of Microsoft docs.

fleed:
@FSharpCSharp there are many advanced features from Microsoft docs that should be added to the standard template, or released as advanced template. Hopefully something will hear us..

superyyrrzz:
@fleed docs.msft have some extra feature on hosting layer to make multi-lang happen. However, it might not be the same implementation for a static site generator like DocFX. It still needs some design work.

fleed:
@superyyrrzz I'm also implementing a .NET core web app with a few middlewares to wrap the statically generated website.. but I feel like "reinventing the wheel"

superyyrrzz:
@FSharpCSharp @fleed We haven't plan to support it in v2. I'm afraid it will not come soon. 

Now we are designing v3. I removed it from board as it is not the top priority for now, but I have added it to the [commnity asks](https://github.com/dotnet/docfx/projects/1#card-8543119). This is a valid user scenario we aim to support.

It is also welcome to share your expected input/output structure to help us better understand the request. Our best practice for docs.ms.com may not apply for other open source repos, as it heavily relies on the system behind DocFX, which is hard to open source.

[...]

PR #3482, opened 2018-10-10
restore loc repository before building

OsmondJiang:
- support 4 types of loc mappings, only "Repository" and "RepositoryAndFolder" need restore loc repo before building
- find loc repo and edit url share the same logic
- only restore root docset loc repo


src/docfx/restore/RestoreGit.cs
```diff
@@ -42,162 +45,66 @@ public static async Task<IEnumerable<(string href, string workTreeHead)>> Restor
...
-                    {
-                        leftWorkTrees.Add(workTree);
-                    }
+                    return default;
                 }
-                await GitUtility.PruneWorkTrees(restorePath);
 
-                return leftWorkTrees;
+                var locRepoFullName = LocConfigConvention.GetLocRepository(config.LocMappingType, repo.FullName, locale, config.DefaultLocale);
+                var locRepo = repo.With(repo.Owner, locRepoFullName.Split(new[] { '\\', '/' }, StringSplitOptions.RemoveEmptyEntries).Last());
```
live1206: >repo.Owner [](start = 40, length = 10)

Updating with the same owner? Seems only need to set Name here, do we need a With method?

src/docfx/loc/LocConfigurationConvention.cs
```diff
@@ -10,17 +10,24 @@ internal static class LocConfigConvention
     {
         private static readonly Regex s_repoNameWithLocale = new Regex(@"^.+?(\.[a-z]{2,4}-[a-z]{2,4}(-[a-z]{2,4})?)?$", RegexOptions.IgnoreCase);
 
+        public static string GetEditRepository(this Config config, string locale)
```
live1206: >GetEditRepository [](start = 29, length = 17)

What is the difference between GetEditRepository and GetLocRepository? If we put repository to input parameters, we can just expose this extension method?

docs/specs/config.yml
```diff
@@ -508,11 +508,11 @@ commands:
 inputs:
```
yufeih: how about a dedicated `localization.yml` spec ?

docs/specs/config.yml
```diff
@@ -508,11 +508,11 @@ commands:
...
   docfx.yml: |
     contribution:
-      repository: github.com/docfx/docfx
+      repository: dotnet/docfx
   docs/a.md:
 outputs:
   docs/a.json: |
-    { "content_git_url": "*github.com/docfx/docfx.zh-cn*"}
+    { "content_git_url": "*dotnet/docfx.zh-cn*"}
```
yufeih: can we remove the wildcard to make the test result more explicit? we only use wildcard only if there are no other choice.

src/docfx/loc/LocMappingType.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Docs.Build
+{
+    public enum LocMappingType
+    {
+        Repository, // loc files are stored in an independent repository like dotnet/docfx.zh-cn per locale
```
yufeih: use `/// <summary>` comments. 

The examples in the comments does not state clearly which part is _repo_, which part is _folder_.

For the **Folder** convention, is it
```
/zh-cn/articles/a.md
/zh-tw/articles/a.md
/articles/a.md
```
or
```
/localization/zh-cn/articles/a.md
/localization/zh-tw/articles/a.md
articles/a.md
```
The latter feels better when locale list is large.

Suppose we have **Folder** convention, do we still need **SideBySide** convention? We can just support **Folder** for initial release

docs/specs/config.yml
OsmondJiang: sounds good, changed

---
In reply to: [223994665](https://github.com/dotnet/docfx/pull/3482#discussion_r223994665) [](ancestors = 223994665)

docs/specs/config.yml
OsmondJiang: we do not have other choice here, the content_git_url for spec test contains rooted path which is different for different machine and platform

---
In reply to: [223995143](https://github.com/dotnet/docfx/pull/3482#discussion_r223995143) [](ancestors = 223995143)

src/docfx/loc/LocMappingType.cs
OsmondJiang: thanks for the suggestion.

For SideBySide, I believe the ideal is from you, where did you get this idea?

---
In reply to: [223997053](https://github.com/dotnet/docfx/pull/3482#discussion_r223997053) [](ancestors = 223997053)

yufeih: It's from here: https://github.com/dotnet/docfx/issues/803

OsmondJiang: we can just support Folder for initial release, like you said

---
In reply to: [224298010](https://github.com/dotnet/docfx/pull/3482#discussion_r224298010) [](ancestors = 224298010)

src/docfx/restore/RestoreGit.cs
```diff
@@ -42,162 +35,80 @@ public static async Task<IEnumerable<(string href, string workTreeHead)>> Restor
...
             {
-                if (GitUtility.IsRepo(restoreDir))
-                {
-                    // already exists, just pull the new updates from remote
-                    // fetch bare repo: https://stackoverflow.com/questions/3382679/how-do-i-update-my-bare-repo
-                    return GitUtility.Fetch(restorePath, url, "+refs/heads/*:refs/heads/*", token);
-                }
-                else
+                // restore dependency repositories
+                var restoreItems = config.Dependencies.Values.GroupBy(d => GetRestoreRootDir(d), PathUtility.PathComparer).Select(g => (g.Key, g.Distinct().ToList())).ToList();
```
yufeih: can we have a method that returns all git dependencies: `IEnumerable<(string? name, string remote)> GetGitDependencies(Config config, string locale)`

OsmondJiang: already have, RestoreGit.GetRestoreItems()

PR #3483, opened 2018-10-10
Fix null error

live1206:


src/docfx/build/legacy/LegacyXrefMap.cs
```diff
@@ -11,7 +11,10 @@ internal static class LegacyXrefMap
         public static void Convert(Docset docset, Context context, XrefMap xrefMap)
         {
             var map = new XrefMapModel();
-            map.References.AddRange(xrefMap.References);
+            if (xrefMap.References != null)
```
yufeih: instead of fixing it here, can you make `XrefMap.References` always non-nullable? all of our collection types should be non-nullable to avoid checking arrays against null everywhere. 

Use an empty dictionary [here](https://github.com/dotnet/docfx/blob/v3/src/docfx/build/xref/XrefMap.cs#L47) if `BuildInternalXrefMap` is false

Also `XrefMap.References` -> `XrefMap.InternalReferences` to be more explicit.

live1206: sounds better, will do

src/docfx/build/xref/XrefMap.cs
```diff
@@ -17,7 +17,7 @@ internal class XrefMap
         private readonly IReadOnlyDictionary<string, XrefSpec> _internalXrefMap;
         private readonly IReadOnlyDictionary<string, XrefSpec> _externalXrefMap;
 
-        public IEnumerable<XrefSpec> References => _internalXrefMap?.Values;
+        public IEnumerable<XrefSpec> InternalReferences => _internalXrefMap?.Values;
```
yufeih: other null checks against _internalXrefMap can also be removed

live1206: yes, they should be removed as well

PR #3498, opened 2018-10-16
Accept JSON config file

live1206:
While docfx.yml and docfx.json both exist, prefer docfx.yml.

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -1,10 +1,14 @@
 ---
-# Specify include files
+# Specify include files, prefer docfx.yml to docfx.json
 inputs:
   docfx.yml: |
     content:
       include:
         - docs/**/*.md
+  docfx.json: |
+    content:
+      include:
+        - ignore.md
   ignore.md:
   docs/index.md:
 outputs:
@@ -244,12 +248,12 @@ outputs:
   output/c.gif:
   build.manifest:
 ---
-# relative path extend
+# relative path extend with json file
 inputs:
   docfx.yml: |
-    extend: _shared/extend.yml
-  _shared/extend.yml: |
-    content: a.md
+    extend: _shared/extend.json
+  _shared/extend.json: |
+    {"content": "a.md"}
   a.md:
 outputs:
   a.json:
@@ -321,11 +325,15 @@ inputs:
   docfx.yml: |
     content: '**/*.md'
     extend:
-      extend.yml
-  extend.yml: |
-    content: 
-      include: a.md
-      exclude: b.md
+      extend.json
+  extend.json: |
+    {
+      "content":
+        {
+          "include": "a.md",
+          "exclude": "b.md"
+        }
+    }
   a.md:
   b.md:
 outputs:
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -6,9 +6,12 @@ inputs:
       include:
         - docs/**/*.md
   docfx.json: |
-    content:
-      include:
-        - ignore.md
+    {
+      "content":
+        {
+          "include": "ignore.md"
+        }
+    }  
   ignore.md:
   docs/index.md:
 outputs:
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -152,13 +152,14 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Load the config under <paramref name="docsetPath"/>
         /// </summary>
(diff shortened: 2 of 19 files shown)


src/docfx/config/Config.cs
```diff
@@ -217,7 +231,16 @@ string GetBranch()
 
         private static (List<Error>, JObject) LoadConfigObject(string filePath)
         {
-            var (errors, config) = YamlUtility.Deserialize<JObject>(File.ReadAllText(filePath));
+            var errors = new List<Error>();
+            JObject config = null;
+            if (filePath.Contains(".yml", StringComparison.OrdinalIgnoreCase))
```
yufeih: EndsWith

src/docfx/config/Config.cs
```diff
@@ -217,7 +231,16 @@ string GetBranch()
...
+            {
+                (errors, config) = YamlUtility.Deserialize<JObject>(File.ReadAllText(filePath));
+            }
+            else if (filePath.Contains(".json", StringComparison.OrdinalIgnoreCase))
+            {
+                (errors, config) = JsonUtility.Deserialize<JObject>(File.ReadAllText(filePath));
```
yufeih: Some of the config error file are hard coding `docfx.yml`

src/docfx/config/Config.cs
yufeih: interesting, maybe we should be using the name before calling GetRestorePath?

src/docfx/docset/Docset.cs
```diff
@@ -59,6 +59,11 @@ internal class Docset
         /// </summary>
         public HashSet<Document> BuildScope => _buildScope.Value;
 
+        /// <summary>
+        /// Gets the config file name.
+        /// </summary>
+        public string ConfigFile { get; }
```
OsmondJiang: >ConfigFile [](start = 22, length = 10)

ConfigFileName?

yufeih: Other places uses `XXXPath`, so `ConfigPath`?

OsmondJiang: ConfigPath like a relative path, but it;s OK

---
In reply to: [225480012](https://github.com/dotnet/docfx/pull/3498#discussion_r225480012) [](ancestors = 225480012)

src/docfx/config/Config.cs
```diff
@@ -152,14 +152,14 @@ public IEnumerable<string> GetExternalReferences()
         /// <summary>
         /// Load the config under <paramref name="docsetPath"/>
         /// </summary>
-        public static (List<Error>, Config) Load(string docsetPath, CommandLineOptions options, bool extend = true, RestoreMap restoreMap = null)
+        public static (List<Error>, Config, string) Load(string docsetPath, CommandLineOptions options, bool extend = true, RestoreMap restoreMap = null)
```
OsmondJiang: >List<Error>, Config, string [](start = 23, length = 27)

(List<Error> errors, Config config, string configPath/ConfigName)?

docs/specs/config.yml
```diff
@@ -1,10 +1,14 @@
...
+# Specify include files, prefer docfx.yml to docfx.json
 inputs:
   docfx.yml: |
     content:
       include:
         - docs/**/*.md
+  docfx.json: |
+    content:
+      include:
+        - ignore.md
```
OsmondJiang: why json file is in yml formate ...

src/docfx/docset/Docset.cs
live1206: ConfigPath is something else, would take ConfigFilaName here as below:
```csharp
public static (List<Error> errors, Config config, string configFileName) Load(string docsetPath, CommandLineOptions options, bool extend = true, RestoreMap restoreMap = null)
```

src/docfx/docset/Docset.cs
```diff
@@ -67,11 +72,12 @@ internal class Docset
         private readonly Lazy<RedirectionMap> _redirections;
         private readonly Lazy<LegacyTemplate> _legacyTemplate;
 
-        public Docset(Context context, string docsetPath, Config config, CommandLineOptions options)
+        public Docset(Context context, string docsetPath, Config config, CommandLineOptions options, string configFile)
```
OsmondJiang: >Config config, CommandLineOptions options, string configFile [](start = 58, length = 60)

that seems weired when the parameters have both config and its name, if the configurartion is always load from Config.Load, why config doesn't know it's name?

PR #3501, opened 2018-10-17
use fallback reader in FAL

yishengjin1413:
use fallback reader in FAL.


yufeih:
can you add a test for this token fallback scenario?

yishengjin1413:
I've added the markdig fallback test cases back.

PR #3506, opened 2018-10-18
 Author should be the last contributor from commit history

### LINKED ISSUE #3500: Contributors calculation logic
The logic to extract author from contributors is wrong from impact test. e.g. For commit history:
- userA
- userB
- userA

The contributors is [userA, userB]. The expect author should be userA, but the actual author is userB. The current logic to fetch the last contributor as author is wrong: https://github.com/dotnet/docfx/blob/v3/src/docfx/build/metadata/ContributionInfo.cs#L98

@yufeih Can you help fix this? I can contribute a [test case](https://github.com/dotnet/docfx/commit/9aca06029500cb4a66c65a7de59cc426460a6444) but cannot find a elegant way to fix it on current logic...

yufeih:
Fixes #3500


src/docfx/build/metadata/ContributionInfo.cs
```diff
@@ -57,50 +56,56 @@ public ContributionInfo(Docset docset, GitHubUserCache gitHubUserCache)
...
+                        var (error, result) = await _gitHubUserCache.GetByLogin(authorName);
+                        if (error != null)
+                            errors.Add(error);
+                        return result?.ToContributor();
+                    }
+                }
+                else if (contributors.Count > 0)
+                {
+                    // When author name is not specified, last contributor is author
+                    return await GetContributor(commits[commits.Count - 1]);
```
superyyrrzz: >commits[commits.Count - 1] [](start = 48, length = 26)

If the last contributor is null, should use the second last contributor as author

src/docfx/build/metadata/ContributionInfo.cs
```diff
@@ -57,50 +56,56 @@ public ContributionInfo(Docset docset, GitHubUserCache gitHubUserCache)
...
-                    if (author != null)
-                        contributors.RemoveAll(c => c.Id == author.Id.ToString());
+            return (errors, author, contributors, updatedDateTime);
 
-                    return (errors, author?.ToContributor(), contributors, updatedDateTime);
+            async Task<Contributor> GetContributor(GitCommit commit)
+            {
+                if (!resolveGitHubUsers)
+                {
+                    return new Contributor { DisplayName = commit.AuthorName, Id = commit.AuthorEmail };
```
superyyrrzz: > Id = commit.AuthorEmail [](start = 77, length = 24)

Will template consume Id? but actually it's a email.

yufeih: no, it won't

PR #3509, opened 2018-10-18
Yaml header cache

live1206:
- Add the cache for extracting metadata from markdown
- Remove the step from markdig pipeline


src/docfx/docset/Context.cs
```diff
@@ -142,6 +146,14 @@ private sealed class Cache
...
 
+            public (List<Error> errors, JObject metadata) ExtractMetadata(Document file)
+                => _metadataCache.GetOrAdd(GetKeyFromFile(file), new Lazy<(List<Error>, JObject)>(() =>
+                {
+                    var content = file.ReadText();
+                    GitUtility.CheckMergeConflictMarker(content, file.FilePath);
+                    return ExtractYamlHeader.Extract(content);
+                })).Value;
+
             private string GetKeyFromFile(Document file)
```
OsmondJiang: >GetKeyFromFile [](start = 27, length = 14)

do we need consider the path Comparison case sensitive or not?

src/docfx/lib/markdown/ExtractYamlHeader.cs
```diff
@@ -1,56 +1,37 @@
...
-using Microsoft.DocAsCode.MarkdigEngine.Extensions;
+using System.Text.RegularExpressions;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal static class ExtractYamlHeader
     {
-        public static MarkdownPipelineBuilder UseExtractYamlHeader(this MarkdownPipelineBuilder builder)
+        public static (List<Error> errors, JObject metadata) Extract(string content)
```
OsmondJiang: >Extract [](start = 61, length = 7)

add unit test for this method

src/docfx/docset/Context.cs
live1206: I think it should be case-sensitive here since we just want to identify if the file has been parsed

Issue #3508, opened 2018-10-18
fix vulnerabilities with YamlDotNet.Signed library


herohua:
https://github.com/dotnet/docfx/network/alerts

FYI @superyyrrzz @fenxu

superyyrrzz:
# Introduction
The related vulnerability is [CVE-2018-1000210](https://nvd.nist.gov/vuln/detail/CVE-2018-1000210):
> YamlDotNet version 4.3.2 and earlier contains a Insecure Direct Object Reference vulnerability in The default behavior of Deserializer.Deserialize() will deserialize user-controlled types in the line "currentType = Type.GetType(nodeEvent.Tag.Substring(1), throwOnError: false);" and blindly instantiates them. that can result in Code execution in the context of the running process. This attack appear to be exploitable via Victim must parse a specially-crafted YAML file. This vulnerability appears to have been fixed in 5.0.0.

# Impact
The code in question is [`TypeNameInTagNodeTypeResolver `](https://github.com/aaubry/YamlDotNet/blob/f96b7cc40a0498f8bafdeb49df3aa23aa2c60993/YamlDotNet/Serialization/NodeTypeResolvers/TypeNameInTagNodeTypeResolver.cs#L35), which is added in [YamlDotNet's default Deserializer](https://github.com/aaubry/YamlDotNet/blob/f96b7cc40a0498f8bafdeb49df3aa23aa2c60993/YamlDotNet/Serialization/DeserializerBuilder.cs#L87), also in DocFX's [YamlUtility](https://github.com/dotnet/docfx/blob/9fbc59c16d5939acd21fc7ddcae018b9ed53a900/src/Microsoft.DocAsCode.YamlSerialization/YamlDeserializer.cs#L110). So DocFX is affected by this vulnerability, which can result in arbitrary code execution on client or serve, with a evil YAML input.

# Fix
YamlDotNet does not add `TypeNameInTagNodeTypeResolver` since 5.0.0. However, in DocFx, simply updating YamlDotNet will not resolve this issue as DocFX constructs the Deserializer and explicitly adds it back. So the right fix for DocFX is:
1. Do not add `TypeNameInTagNodeTypeResolver` in to DocFX's Deserializer.
2. Upgrade YamlDotNet to 5.0.0. (This is not mandatory, as all the code in DocFX use DocFX's Deserializer instead of YamlDotNet's default Deserializer)

PR #3511, opened 2018-10-18
avoid vulnerability in YamlDotNet

superyyrrzz:
#3508


yufeih:
Does this suppresses the vulnerabilities alert without upgrading YamlDotNet?

superyyrrzz:
I think GitHub only scans for the YamlDotNet version, so it will not automatically. However, We can manually mark alerts as "vulnerable code is not actually used" and close it.

yufeih:
ok

PR #3514, opened 2018-10-19
ParallelUtility rethrow OperationCanceledException

### LINKED ISSUE #3510: Parallel Foreach eats TaskCancelled Exception
https://ci.appveyor.com/project/DocFXService/docfx/builds/19606644

yufeih:
Fixes #3510


src/docfx/lib/ParallelUtility.cs
```diff
@@ -51,19 +60,38 @@ public static Task ForEach<T>(IEnumerable<T> source, Func<T, Task> action, Actio
...
+            }
 
             async Task Run(T item)
             {
-                await action(item);
+                try
+                {
+                    await action(item);
+                }
+                catch (TaskCanceledException tce)
```
OsmondJiang: do we have any document can reference?

src/docfx/lib/ParallelUtility.cs
```diff
@@ -134,5 +173,11 @@ private static void DebugAssertPostedOrFaulted<T>(bool posted, ActionBlock<T> qu
                 Debug.Assert(queue.Completion.IsFaulted);
             }
         }
+
+        private class WrapException : Exception
```
OsmondJiang: >WrapException [](start = 22, length = 13)

add link or comment to here, explain why do you do that?

src/docfx/lib/ParallelUtility.cs
```diff
@@ -51,19 +60,34 @@ public static Task ForEach<T>(IEnumerable<T> source, Func<T, Task> action, Actio
...
+            }
 
             async Task Run(T item)
             {
-                await action(item);
+                try
+                {
+                    await action(item);
+                }
+                catch (OperationCanceledException oce)
```
live1206: >OperationCanceledException [](start = 23, length = 26)

Shall we try a few retry here when we get OperationCanceledException? Especially while we get it with restoring timeout, maybe we do not need to fail the entire run

yufeih: Retry should be handled in the inner action.

live1206: true

---
In reply to: [226557534](https://github.com/dotnet/docfx/pull/3514#discussion_r226557534) [](ancestors = 226557534)

PR #3517, opened 2018-10-19
support 'folder' mapping for loc docset build

OsmondJiang:


src/docfx/docset/Docset.cs
```diff
@@ -157,7 +157,9 @@ private HashSet<Document> CreateBuildScope(IEnumerable<Document> redirections)
         {
             using (Progress.Start("Globbing files"))
             {
-                var fileGlob = new FileGlob(Config.Content.Include, Config.Content.Exclude);
+                // exclude the localization folder if the current localization mapping is "folder"
+                var exclude = Config.LocalizationMapping == LocalizationMapping.Folder ? Config.Content.Exclude.Concat(new[] { "localization/**/*" }).ToArray() : Config.Content.Exclude;
```
yufeih: just add to `Config.s_defaultContentExclude` ?

PR #3519, opened 2018-10-22
output folder only has loc built files

OsmondJiang:


src/docfx/build/Build.cs
```diff
@@ -105,7 +106,8 @@ async Task BuildOneFile(Document file, Action<Document> buildChild)
 
                 bool ShouldBuildFile(Document file)
                 {
-                    return file.ContentType != ContentType.Unknown && filesBuilder.TryAdd(file);
+                    return file.ContentType != ContentType.Unknown && filesBuilder.TryAdd(file) &&
+                        (docset.FallbackDocset == null/*source docset*/ || file.Docset.FallbackDocset != null /*loc file*/);
```
live1206: In which case, the file to be built is neither from source docset nor a loc file? And we want to exclude it from build scope

OsmondJiang: ？

---
In reply to: [227236163](https://github.com/dotnet/docfx/pull/3519#discussion_r227236163) [](ancestors = 227236163)

src/docfx/build/Build.cs
```diff
@@ -69,6 +69,7 @@ Docset GetBuildDocset()
 
         private static async Task<(List<Document> files, DependencyMap sourceDependencies)> BuildFiles(
             Context context,
+            Docset docset,
             HashSet<Document> buildScope,
```
superyyrrzz: >HashSet<Document> buildScope, [](start = 12, length = 29)

this parameter is no longer needed as it's included in `docset`?

src/docfx/build/Build.cs
```diff
@@ -105,7 +105,10 @@ async Task BuildOneFile(Document file, Action<Document> buildChild)
 
                 bool ShouldBuildFile(Document file)
                 {
-                    return file.ContentType != ContentType.Unknown && filesBuilder.TryAdd(file);
+                    return file.ContentType != ContentType.Unknown &&
+                        filesBuilder.TryAdd(file) &&
+                        !(docset.FallbackDocset != null &&
```
yufeih: can you refactor this a little bit? there are too much conditions here:
```csharp
// English content in a localized build
if (docset.FallbackDocset != null && file.Docset.LocalizationDocset != null)
{
    return false;
}
return file.ContentType != Unknown && TryAdd(xxx)
```

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -24,11 +24,6 @@ internal static class BuildTableOfContents
             var dependencyMapBuilder = new DependencyMapBuilder();
             var (errors, tocModel, tocMetadata, refArticles, refTocs) = Load(context, file, dependencyMapBuilder);
```
yufeih: what if a link in TOC points to an PNG or config? we could have a TODO here.

OsmondJiang: toc can't reference to a link or config right now, but i can add a TODO here.

---
In reply to: [227639083](https://github.com/dotnet/docfx/pull/3519#discussion_r227639083) [](ancestors = 227639083)

PR #3521, opened 2018-10-22
Switch to kthompson/glob for better performance

yufeih:
This has a significant impact on azure. The existing docfx glob is too slow. It reduces file build time from 45s to 23s, reduces legacy resource time from 1.2min to 35s. Future optimization could be made by #3518 

Core glob features are preserved:

`**, *, {}, []`

https://github.com/kthompson/glob/#supported-pattern-expressions

There are some notifiable changes from the switch:
- throw an error for bad glob patterns like `[`
- drop negate pattern (we have exclude)


src/docfx/config/GlobConfig.cs
```diff
@@ -47,32 +50,30 @@ public bool Match(string filePath)
...
-                return false;
-
             if (IsGlob)
             {
-                // TODO: optimize this
-                var options = PathUtility.IsCaseSensitive ? GlobMatcher.DefaultCaseSensitiveOptions : GlobMatcher.DefaultOptions;
-                var glob = new GlobMatcher(pattern, options);
-                return glob.Match(filePath);
+                return LazyInitializer.EnsureInitialized(
+                    ref _glob,
```
live1206: >_glob [](start = 24, length = 5)

can it just be a local variable?

yufeih: This ensure `_glob` is only initialized once. the `CreateGlobMatcher` method is more expensive compare to the `match` function itself

live1206: Thanks for the explanation

Issue #3497, opened 2018-10-15
Resolve xref for SDP


live1206:
Only set Href while resolving xref for SDP for now, it needs to set additional xref properties as well

herohua:
this is now part of https://dev.azure.com/ceapex/Engineering/_workitems/edit/102269

PR #3523, opened 2018-10-23
For SDP, reoslve uid to html according to defined display property

live1206:
#3497


src/docfx/build/xref/XrefMap.cs
```diff
@@ -32,6 +32,12 @@ public XrefSpec Resolve(string uid)
             return null;
         }
 
+        public string ResolveToHtml(string uid, string displayProperty)
+        {
+            XrefSpec xrefSpec = Resolve(uid);
+            return $"<a href=\"{xrefSpec.Href}\">{xrefSpec.GetXrefProperty(displayProperty.ToLowerInvariant())}</a>";
```
superyyrrzz: >xrefSpec.GetXrefProperty(displayProperty.ToLowerInvariant()) [](start = 50, length = 60)

Need some logic to judge whether to html encode this or not.

live1206:
https://github.com/Microsoft/templates.docs.msft/blob/8f2bb41a8b6537bfe6fda3482577b67dd77d0170/ContentTemplate/partials/dotnet/clsCompliantAlternative.tmpl.partial
Template would use xref spec as a object, so there is no way for docfx to render html during SDP xref resolving.

PR #3530, opened 2018-10-25
Cache and reuse git commit history

yufeih:
Replaces the existing bulk git commit history computation logic with a cache friendly, watch friendly `GitCommitProvider`. `GitCommitProvider` can be used to query commit history for a file at any time. It internally manages an in memory cache and a disk cache.

Performance data for loading commits on azure:
- First time (without cache): **1:55s**
- Second time (with cache): **3s**
- First time peak memory size: ~5GB
- Second time peak memory size: ~1GB
- Cache size on disk: ~2MB per run

The disk cache uses a simple LRU algorithm to only store last (**N** = 10) runs.


yufeih:
Test failed due to competing changes to the cache file.

src/docfx/build/metadata/ContributionInfo.cs
```diff
@@ -226,23 +231,27 @@ private Repository GetRepository(string path)
...
 
                 foreach (var group in filesByRepo)
                 {
-                    var pathToDocset = group.Select(pair => pair.file.FilePath).ToList();
-                    var pathToRepo = group.Select(pair => pair.pathToRepo).ToList();
                     var repo = group.Key;
                     var repoPath = repo.Path;
-                    var (commitsByFile, allCommits) = GitUtility.GetCommits(repoPath, pathToRepo);
-                    for (var i = 0; i < pathToDocset.Count; i++)
+                    var commitCachePath = Path.Combine(AppData.CacheDir, "commits", HashUtility.GetMd5Hash(repo.Remote));
```
OsmondJiang: >var commitCachePath = Path.Combine(AppData.CacheDir, "commits", HashUtility.GetMd5Hash(repo.Remote)); [](start = 20, length = 101)

can we put all cache file in one place to manage? like commits or github user cache files?

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -0,0 +1,393 @@
...
+
+            var x = _commits.Value;
+        }
+
+        public static async Task<GitCommitProvider> Create(string repoPath, string cacheFilePath = null)
+        {
+            return new GitCommitProvider(repoPath, cacheFilePath, await LoadCommitCache(cacheFilePath));
+        }
+
+        public unsafe List<GitCommit> GetCommitHistory(string file)
```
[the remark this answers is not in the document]
yufeih: It'll be in another PR, this one is already too big

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -0,0 +1,393 @@
...
+                        if (commitCache.TryGetValue((commit.Sha.A, blob), out var cachedValue))
+                        {
+                            updateCache = result.Count != 0;
+
+                            var (cachedCommitHistory, lruOrder) = cachedValue;
+                            foreach (var cachedCommit in cachedCommitHistory)
+                            {
+                                result.Add(commitsBySha[cachedCommit]);
+                            }
+                            commitCache[(commit.Sha.A, blob)] = (cachedCommitHistory, _nextLruOrder--);
```
OsmondJiang: >commitCache[(commit.Sha.A, blob)] = (cachedCommitHistory, _nextLruOrder--); [](start = 28, length = 75)

do we need set _cacheUpdated here?

PR #3539, opened 2018-10-29
fix loc restore map bug

OsmondJiang:
1. loc docset should share the source docset' restore map since it already shares the source docset's config
2. fix legacy template locale issue
3. dependency docsets don't need find fallback docset


src/docfx/restore/RestoreMap.cs
```diff
@@ -38,15 +41,15 @@ public string GetGitRestorePath(string remote)
             throw Errors.NeedRestore(remote).ToException();
         }
 
-        public string GetUrlRestorePath(string docsetPath, string path)
+        public string GetUrlRestorePath(string path)
```
yufeih: 👍 I was about to do the same thing

src/docfx/docset/Docset.cs
```diff
@@ -85,28 +85,28 @@ internal class Docset
...
         }
 
-        private Docset(Context context, string docsetPath, Config config, string locale, CommandLineOptions options, Docset fallbackDocset)
+        private Docset(Context context, string docsetPath, Config config, string locale, CommandLineOptions options, Docset fallbackDocset, RestoreMap restoreMap)
         {
             _options = options;
             _context = context;
             Config = config;
             DocsetPath = PathUtility.NormalizeFolder(Path.GetFullPath(docsetPath));
             Locale = locale;
```
yufeih: nit: maybe `ToLower` here

src/docfx/docset/Docset.cs
```diff
@@ -85,28 +85,28 @@ internal class Docset
         private readonly Lazy<RedirectionMap> _redirections;
         private readonly Lazy<LegacyTemplate> _legacyTemplate;
 
-        public Docset(Context context, string docsetPath, Config config, CommandLineOptions options)
-            : this(context, docsetPath, config, !string.IsNullOrEmpty(options.Locale) ? options.Locale.ToLowerInvariant() : config.Localization.DefaultLocale, options, null)
+        public Docset(Context context, string docsetPath, Config config, CommandLineOptions options, bool fallback = true)
```
yufeih: nit: maybe `fallback` -> `isDependency` to help express that we don't resolve loc for dependency docsets?

src/docfx/docset/Docset.cs
```diff
@@ -150,7 +150,7 @@ private CultureInfo CreateCultureInfo(string locale)
                 // todo: what parent config should be pass on its children
                 Config.LoadIfExists(dir, _options, out var loadErrors, out var subConfig);
                 errors.AddRange(loadErrors);
-                result.TryAdd(PathUtility.NormalizeFolder(name), new Docset(_context, dir, subConfig, _options));
+                result.TryAdd(PathUtility.NormalizeFolder(name), new Docset(_context, dir, subConfig, _options, false));
```
yufeih: nit: `fallback: true`

docs/specs/localization.yml
```diff
@@ -26,11 +26,61 @@ inputs:
       mapping: RepositoryAndFolder
 restores:
   git/github.com/docascode/docfx-test-dependencies.localization/18d3d65979413ce8342fd709dbce1c78ab45ffeb/dep.md: loc-DEP
-  restore-lock/*Restore loc repo, all loc files are in one repo-*-lock.json: |
+  restore-lock/*restore loc repo, all loc files are in one repo-*-lock.json: |
     {"git":{"https://github.com/docascode/docfx-test-dependencies.localization#master":"18d3d65979413ce8342fd709dbce1c78ab45ffeb"}}
 outputs:
   build.manifest:
 ---
+# loc docset share the source docset's restore map
```
superyyrrzz: Also add `restore` section to check there is only one restore map?

PR #3542, opened 2018-10-29
Replace some process mutex with ProcessUtility.ReadFile/WriteFile

yufeih:
- Add `ProcessUtility.ReadFile/WriteFile` for better inter-process single file read write. It uses `FileShare` to control access, it does not use a separate file lock, provide concurrent read on the same file.
- Replace git commits time with the new methods
- Will replace git commit history cache read/write when https://github.com/dotnet/docfx/pull/3530 is in.

### DIFF
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -188,7 +188,7 @@ namespace Microsoft.Docs.Build
         /// </summary>
         public static bool IsFileUsedByAnotherProcessException(IOException ex)
         {
-            return ex.HResult == 35 /*|| ex.HResult == 17*/;
+            return ex.HResult == 35 || ex.HResult == -2147024864;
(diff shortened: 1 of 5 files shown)


src/docfx/lib/ProcessUtility.cs
```diff
@@ -91,6 +91,62 @@ public static Task<string> Execute(string fileName, string commandLineArgs, stri
...
+            {
+                using (var fs = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, 1024, FileOptions.SequentialScan))
+                using (var reader = new StreamReader(fs))
+                {
+                    return reader.ReadToEnd();
+                }
+            });
+        }
+
+        public static Task<T> ReadFile<T>(string path, Func<Stream, T> read)
```
live1206: never used?

PR #3546, opened 2018-10-30
Support get contributor from non-sxs branch

OsmondJiang:


src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -221,15 +224,36 @@ public void Dispose()
...
+                {
+                    if (GitBranchLookup(out var refBranch, _repo, pLocaleBranchName, 1 /*locale branch*/) == 0 ||
+                        GitBranchLookup(out refBranch, _repo, pRemoteBranchName, 2 /*remote branch*/) == 0)
+                    {
+                        var commit = GitReferenceTarget(refBranch);
+                        GitRevwalkPush(walk, commit);
+                    }
+                    else
+                    {
+                        throw new ArgumentException($"{branchName} can't not be resolved");
```
yufeih: we should have an user error if the loc repo or branch does not exists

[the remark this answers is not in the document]
OsmondJiang: ok, but why below Errors.GitShadowClone doesn't release walk before throw?

---
In reply to: [229592511](https://github.com/dotnet/docfx/pull/3546#discussion_r229592511) [](ancestors = 229592511)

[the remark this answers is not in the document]
OsmondJiang: and also the other invalidOperationException()?

---
In reply to: [229653306](https://github.com/dotnet/docfx/pull/3546#discussion_r229653306) [](ancestors = 229653306,229592511)

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -223,15 +226,36 @@ public void Dispose()
...
+            {
+                GitRevwalkPushHead(walk);
+            }
+            else
+            {
+                fixed (byte* pLocaleBranchName = ToUtf8Native($"{branchName}"))
+                fixed (byte* pRemoteBranchName = ToUtf8Native($"origin/{branchName}"))
+                {
+                    if (GitBranchLookup(out var refBranch, _repo, pLocaleBranchName, 1 /*locale branch*/) == 0 ||
+                        GitBranchLookup(out refBranch, _repo, pRemoteBranchName, 2 /*remote branch*/) == 0)
```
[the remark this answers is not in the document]
OsmondJiang: if the local branch has the new commit, we should use the local branch firstly,like testing

---
In reply to: [229591569](https://github.com/dotnet/docfx/pull/3546#discussion_r229591569) [](ancestors = 229591569)

OsmondJiang:
#3547

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -249,14 +249,15 @@ private async Task LoadCommits(Docset docset)
                 {
                     var repo = group.Key;
                     var repoPath = repo.Path;
+                    var repoBranch = bilingual && repo.Branch.EndsWith("-sxs") ? repo.Branch.Substring(0, repo.Branch.Length - 4) : null;
```
live1206: when bilingual is true, and input repo branch not ending with "-sxs", should we log some warning to inform user?

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -61,13 +61,16 @@ public static async Task<GitCommitProvider> Create(string repoPath, string cache
             return new GitCommitProvider(repoPath, cacheFilePath, await LoadCommitCache(cacheFilePath));
         }
 
-        public List<GitCommit> GetCommitHistory(string file)
+        public List<GitCommit> GetCommitHistory(string file, string branch = null)
```
live1206: default branch as `string.Empty`?

src/docfx/build/metadata/ContributionProvider.cs
OsmondJiang: u are right, but we should not log here, we should log warnings at the begining of building, let's check it in another PR

---
In reply to: [229918463](https://github.com/dotnet/docfx/pull/3546#discussion_r229918463) [](ancestors = 229918463)

src/docfx/lib/git/GitCommitProvider.cs
OsmondJiang: I believe both are OK, even the default branch is string.empty, user still gets chance to pass branch 'null'

---
In reply to: [229918800](https://github.com/dotnet/docfx/pull/3546#discussion_r229918800) [](ancestors = 229918800)

src/docfx/restore/RestoreWorkTree.cs
```diff
@@ -75,7 +75,9 @@ async Task AddWorkTrees()
                     var workTreePath = GetRestoreWorkTreeDir(restoreDir, workTreeHead);
                     if (existingWorkTrees.TryAdd(workTreePath, 0))
                     {
-                        await GitUtility.AddWorkTree(restorePath, workTreeHead, workTreePath);
+                        // use branch name instead of commit hash
+                        // https://git-scm.com/docs/git-worktree#_commands
+                        await GitUtility.AddWorkTree(restorePath, rev, workTreePath);
```
yufeih: There is a tiny problem with this approach if this `branch` is shared globally and accessed at the same time by different builds.

OsmondJiang: ok

OsmondJiang: although we use the branch, but the worktree path is different for different build.

PR #3549, opened 2018-10-31
add cache to fix swagger parser perf issue

partychen:
1. when return type SwaggerLoopReferenceObject, we should also return the extension data such Token.
2. add _resolvedObjectCache, when the object is resolved we don't need to resolve again.


src/Microsoft.DocAsCode.Build.RestApi/Swagger/Internals/SwaggerJsonBuilder.cs
```diff
@@ -18,13 +18,17 @@ namespace Microsoft.DocAsCode.Build.RestApi.Swagger.Internals
     internal class SwaggerJsonBuilder
```
herohua: >SwaggerJsonBuilder [](start = 19, length = 18)

is there any unit test around this?

partychen: SwaggerJsonParserTest#ParseSwaggerJsonWithLoopReferenceShouldSucceed() with hit the cache

PR #3550, opened 2018-10-31
 Respect libgit2 naming convension for pinvoke

yufeih:
Makes libgit2 easier to work with.
Subsequent change after https://github.com/dotnet/docfx/pull/3548


src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -244,84 +248,84 @@ private unsafe (List<Commit>, Dictionary<long, Commit>) LoadCommits()
                 if (error != 0)
                     throw new InvalidOperationException($"Unknown error calling git_revwalk_next: {error}");
 
-                GitObjectLookup(out var commit, _repo, &commitId, GitObjectType.Commit);
-                var author = GitCommitAuthor(commit);
-                var committer = GitCommitCommitter(commit);
-                var parentCount = GitCommitParentcount(commit);
-                var parents = new GitOid[parentCount];
+                git_object_lookup(out var commit, _repo, &commitId, 1 /* GIT_OBJ_COMMIT */);
```
live1206: why not using enum GitObjectType any more?

src/docfx/lib/git/GitUtility.cs
```diff
@@ -76,44 +77,31 @@ public static bool IsRepo(string path)
         public static unsafe (string remote, string branch, string commit) GetRepoInfo(string repoPath)
         {
             var (remote, branch, commit) = default((string, string, string));
-            var pRepo = OpenRepo(repoPath);
 
-            // TODO: marshal strings
-            fixed (byte* pRemoteName = LibGit2.ToUtf8Native("origin"))
+            if (LibGit2.git_repository_open(out var pRepo, repoPath) != 0)
```
live1206:
```csharp
using static Microsoft.Docs.Build.LibGit2;
```
To get rid of `LibGit2.`?

yufeih: 👍 , I did that in another PR.

src/docfx/lib/git/GitCommitProvider.cs
yufeih: We only use 2 of them, there are various other enums we use that are not defined, like `GIT_ITEROVER`, I just use this style to be consistent.

Issue #3513, opened 2018-10-19
Config, config service and config UI


yufeih:
Tracks config related problems:

- docs wants a centralized place to manage configs with an UI
    - Config expansion complicates schema driven UI
    - A flat list of simple configs makes schema driven UI a lot easier.
- docs prefers convention over config, if there is a convention we could apply all across docs, don't make it configurable. This makes UI a lot easier to use:
    - Validation rules are the same for all docs content
    - master/live publish workflow
    - Localization repositories follows a couple of conventions
    - ~~Can we enforce directory names to match url (no routes #2864) ?~~
    - ~~Can we use directory name to indicate moniker (#3507) ?~~

PR #3552, opened 2018-10-31
Make fileMetadata config same as v2

yufeih:
`anyof` object or array creates problems with configuration service and configuration UI. 

- Revert file metadata to be the same as v2.
- Make routes only take object form

Corresponding changes https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/5641?_a=files

#3513


yufeih:
One test case updated due to https://github.com/kthompson/glob/issues/41

PR #3554, opened 2018-10-31
Implement mock repos in e2e test

yufeih:
Implements #2919 

Depending on PR https://github.com/dotnet/docfx/pull/3548 https://github.com/dotnet/docfx/pull/3550

Added a `repos` property to e2e spec to mock repos. It is a list of repos with commit history. The first one will be used as build entry point.

> Mocked repos does not support GitHub API, so the recommended mock URL is https://docs.com/XXX


src/docfx/lib/git/GitUtility.cs
```diff
@@ -202,6 +187,43 @@ public static void CheckMergeConflictMarker(string content, string file)
...
+            }
+
+            git_repository_free(repo);
+
+            // Fetch from local file system if we are using a mock repo for testing
+            var mockedRepos = MockedRepos.Value;
+            if (mockedRepos != null && mockedRepos.TryGetValue(url, out var mockedLocation))
+            {
+                url = mockedLocation;
+            }
```
OsmondJiang: can we avoid to mix the test code and prod code, i think we can prepare the NockedRepos before test

src/docfx/restore/RestoreGit.cs
```diff
@@ -68,53 +68,58 @@ public static async Task GC(string docsetPath, Config config, Func<string, Task>
...
             if (repo == null)
             {
-                return default;
+                yield break;
+            }
+
+            if (config.Localization.Bilingual)
+            {
+                // Bilingual repos also depend on non bilingual branch
+                yield return ToHref(LocalizationConvention.GetLocalizationRepo(
```
OsmondJiang: is that too much heavy to calculate non-sxs branch?

yufeih: GetLocalizationRepo should be compute only and runs very fast right? I want a single place to handle `-sxs` logic.

PR #3556, opened 2018-10-31
Remove BuildInternalXrefMap switch

live1206:
- remove `BuildInternalXrefMap` from config
- skip `xrefmap.json` and `build.manifest` if spec output does not contain them but the actual output does

### DIFF
--- a/docs/specs/localization.yml
+++ b/docs/specs/localization.yml
@@ -12,7 +12,6 @@ restores:
   restore-lock/1dc2455be4a78b4fd29b5641e6fb8d687f37464a-*-lock.json: |
     {"git":{"https://github.com/docascode/docfx-test-dependencies#child":"5736374c0a142252063bbedafe5d98db495e263d"}}
 outputs:
-  build.manifest:
 ---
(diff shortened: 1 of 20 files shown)


test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -74,6 +74,17 @@ private static async Task RunCore(string docsetPath, E2ESpec spec)
                 VerifyFile(restoredFile, content);
             }
 
+            // These files output mostly contains empty content which e2e tests are not intrested in
+            // we can just skip the verification for them
+            var skippableOutput = new string[] { "xrefmap.json", "build.manifest" };
```
OsmondJiang: put them in E2ESpec.cs

docs/specs/localization.yml
```diff
@@ -375,8 +375,7 @@ commands:
   - build --locale zh-cn
 repo: https://github.com/docascode/docfx-test-dependencies#loc-xrefmap
 inputs:
```
yufeih: also remove empty `build.manifest:`

Issue #3559, opened 2018-11-01
System.IO.EndOfStreamException : Unable to read beyond the end of the stream.


yufeih:
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=21795&view=logs
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=21958&view=logs
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=22258&view=logs
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=22365&view=logs
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=22630&view=logs
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=22739&view=logs

yufeih:
Feels like cache file is empty

PR #3563, opened 2018-11-01
Fix git commit cache data curruption

yufeih:
Fixes https://github.com/dotnet/docfx/issues/3559 due to `FileShare` and `FileMode` are not atomic.

### DIFF
--- a/src/docfx/lib/git/GitCommitProvider.cs
+++ b/src/docfx/lib/git/GitCommitProvider.cs
@@ -192,8 +192,11 @@ namespace Microsoft.Docs.Build
             {
                 using (var writer = new BinaryWriter(stream))
                 {
-                    writer.Write(_commitCache.Count);
-                    foreach (var (file, value) in _commitCache)
+                    // Create a snapshot of commit cache to ensure count and items matches.
+                    var commitCache = _commitCache.ToList();
+
+                    writer.Write(commitCache.Count);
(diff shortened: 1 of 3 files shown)


src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -189,8 +189,11 @@ public Task SaveCache()
             {
                 using (var writer = new BinaryWriter(stream))
                 {
-                    writer.Write(_commitCache.Count);
-                    foreach (var (file, value) in _commitCache)
+                    // Create a snapshot of the commit cache to avoid writing corrupected data
+                    var commitCache = _commitCache.ToList();
```
OsmondJiang: any reason here why ToList can resolve this problem?

yufeih: after `writer.Write(_commitCache.Count)` and before `foreach (xxx in _commitCache)` is completed, there is a tiny chance that a new entry is added to `_commitCache`, causing the count on disk to not match the items count.

OsmondJiang: so we should lock _commitCache right?

OsmondJiang: I looked the code, didn't fond the tiny chance to change _commitCache during save unless we have some bugs, so we should take a look at the root cause

PR #3564, opened 2018-11-01
Fix run-inside-mutex-hang issue

OsmondJiang:
its' caused by deadlock between restore and GC. I have simplified the GC, resolve the deadlock, it's depends on PR #3567

### DIFF
--- a/src/docfx/restore/RestoreUrl.cs
+++ b/src/docfx/restore/RestoreUrl.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Docs.Build
 
                     foreach (var existingVersionPath in existingVersionPaths)
                     {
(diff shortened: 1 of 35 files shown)


src/docfx/restore/RestoreUrl.cs
```diff
@@ -13,7 +13,7 @@ namespace Microsoft.Docs.Build
 {
     internal static class RestoreUrl
     {
-        private const int MaxVersionCount = 5;
+        public const int MaxKeepingDays = 10;
```
live1206: shall we keep this as private if this is only used for testing? I mean we can even set the `LastAccessTime` to 20 days ago.
Same for `RestoreWorkTree`

OsmondJiang: ok, sounds good

---
In reply to: [230249986](https://github.com/dotnet/docfx/pull/3564#discussion_r230249986) [](ancestors = 230249986)

PR #3567, opened 2018-11-01
fix restore worktree path bug

OsmondJiang:
if the master and live has the same HEAD, then they will be checked out to same worktree path, we want to avoid that


src/docfx/restore/RestoreWorkTree.cs
```diff
@@ -71,7 +71,7 @@ async Task AddWorkTrees()
                 await ParallelUtility.ForEach(hrefs, async href =>
                 {
                     var (_, rev) = GitUtility.GetGitRemoteInfo(href);
-                    var workTreeHead = await GitUtility.Revision(restorePath, rev);
+                    var workTreeHead = $"{await GitUtility.Revision(restorePath, rev)}-{rev}";
```
superyyrrzz: Will this fail when `rev` contains InvalidFileChars, or contains `/` that makes path contains sub folder?

PR #3569, opened 2018-11-02
Xref conceptual display property

live1206:
Support 
```
<xref:uid?displayProperty=fullName>
```

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -112,6 +112,25 @@ outputs:
   xrefmap.json: | 
     {"references":[{"uid":"a","href":"docs/a.json"}]}
 ---
+# Resolve xref with non-existing display property, display name
+inputs:
+  docfx.yml:
+  docs/a.json: |
+    {
+      "$schema": "https://raw.githubusercontent.com/dotnet/docfx/v3/schemas/TestData.json",
+      "uid": "a",
+      "name": "test",
+      "fullName": "test.test",
+      "description": "a.description"
+    }
+  docs/c.md: Link to <xref:a?displayProperty=notExisting>
+outputs:
+  docs/c.json: | 
+    {"content":"<p>Link to <a href=\"docs/a.json\">test</a></p>\n"}
+  docs/a.json:
+  xrefmap.json: | 
+    {"references":[{"uid":"a","href":"docs/a.json"}]}
+---
 # Define duplicate uid in conceptual and SDP 
 # should output warning and resolve it with the first one ordered by href
 inputs:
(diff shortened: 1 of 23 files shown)


src/docfx/lib/markdown/ResolveXref.cs
```diff
@@ -15,31 +16,49 @@ internal static class ResolveXref
...
-            });
+                         string display;
+                         var content = string.IsNullOrEmpty(xrefSpec.GetName()) ? xrefSpec.Uid : xrefSpec.GetName();
+                         if (!string.IsNullOrEmpty(query))
+                         {
+                             var queries = HttpUtility.ParseQueryString(query.Substring(1));
+                             var displayProperty = queries["displayProperty"];
+                             if (!string.IsNullOrEmpty(displayProperty))
+                             {
+                                 display = xrefSpec.GetXrefPropertyValue(displayProperty);
```
yufeih: display could be null if the property does not exists, suggest to use the following fallback:

```csharp

propertyName = query?.displayName ?? "name"
display = xrefspec.ExtensionData[propertyName] ?? uid

```

docs/specs/xref.yml
```diff
@@ -93,6 +93,44 @@ outputs:
...
+      "name": "test",
+      "fullName": "test.test",
+      "description": "a.description"
+    }
+  docs/c.md: Link to <xref:a?displayProperty=fullName>
+outputs:
+  docs/c.json: | 
+    {"content":"<p>Link to <a href=\"docs/a.json\">test.test</a></p>\n"}
+  docs/a.json:
+  xrefmap.json: |
```
yufeih: maybe merge these 2 test cases to save some test time

src/docfx/build/xref/XRefSpec.cs
```diff
@@ -15,6 +15,9 @@ public class XrefSpec
         [JsonExtensionData]
         public JObject ExtensionData { get; } = new JObject();
 
-        public string GetName() => ExtensionData.TryGetValue("name", out var name) && name is JValue v && v.Value is string str ? str : null;
+        public string GetName() => GetXrefPropertyValue("name");
+
+        public string GetXrefPropertyValue(string property)
+            => ExtensionData.TryGetValue(property, out var prop) && prop is JValue v && v.Value is string str ? str : null;
```
OsmondJiang: >ExtensionData.TryGetValue(property, out var prop) && prop is JValue  [](start = 15, length = 68)

we have a JosnUtility.TryGetValue<>();

live1206: ok, good to know

Issue #3568, opened 2018-11-02
Remove restore lock file


OsmondJiang:
Currently `docfx restore` uses a `xxx.lock` file to track the mapping between dependency urls and real restored file/folder path, but with this file, every change/read to this file need apply process lock to this file, which caused a lot of performance issue and also deadlock issue. We want to remove lock file, always use the latest restored version of file/git.

PR #3570, opened 2018-11-02
optimize restore lock

OsmondJiang:
remove restore lock file, simplify the restore process.


src/docfx/restore/RestoreWorkTree.cs
```diff
@@ -80,6 +80,9 @@ async Task AddWorkTrees()
                         await GitUtility.AddWorkTree(restorePath, rev, workTreePath);
                     }
 
+                    // update the last access time
+                    Directory.SetLastAccessTimeUtc(workTreePath, DateTime.UtcNow);
```
live1206: why do we need to update the last access time manually?

OsmondJiang: depends on different platform, it's supported differently, so we manually handle that:
https://superuser.com/questions/464290/why-is-cat-not-changing-the-access-time

live1206: ok, thanks for the explanation

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -65,15 +65,6 @@ private static async Task RunCore(string docsetPath, E2ESpec spec)
                 Assert.True(false, File.ReadAllText(Path.Combine(docsetOutputPath, "build.log")));
             }
 
-            // Verify restored files
-            foreach (var (file, content) in spec.Restores)
```
yufeih: There are also tests to check git clone location, do you still want those?

src/docfx/restore/RestoreMap.cs
```diff
@@ -1,69 +1,82 @@
...
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal class RestoreMap
     {
-        private readonly RestoreLock _restoreLock;
+        private static readonly ConcurrentDictionary<string, string> s_mappings = new ConcurrentDictionary<string, string>();
```
yufeih: Since everything is static and readonly across a single process, can we expose 2 static methods to avoid passing objects around? like:

```csharp
Restore.GetGitRepositoryPath(string remote, string branch)
Restore.GetFileDownloadPath(string docsetPath, string pathOrUrl)
```

live1206: Should it be as below to make each value created once?
```csharp
ConcurrentDictionary<string, Lazy<string>>
```

OsmondJiang: not everything is static, we need keep the docset path context in every restore map, share it with source and loc docset

---
In reply to: [230308044](https://github.com/dotnet/docfx/pull/3570#discussion_r230308044) [](ancestors = 230308044)

OsmondJiang: good point, will change

---
In reply to: [230617031](https://github.com/dotnet/docfx/pull/3570#discussion_r230617031) [](ancestors = 230617031)

test/docfx.Test/e2e/E2ETest.cs
OsmondJiang: yes, i just remove the lock file content check, but still want to make sure the dependencies are restored correctly

---
In reply to: [230305694](https://github.com/dotnet/docfx/pull/3570#discussion_r230305694) [](ancestors = 230305694)

live1206:
#3568

src/docfx/lib/ProcessUtility.cs
```diff
@@ -160,25 +160,34 @@ public static async Task RunInsideMutex(string mutexName, Func<Task> action)
             Debug.Assert(!string.IsNullOrEmpty(mutexName));
 
             // avoid the RunInsideMutex to be nested used
+            // doesn't support to require a lock before releasing a lock
+            // which will cause deadlock
             Debug.Assert(t_innerCall.Value == null);
```
yufeih: throw? Debug.Assert only runs in DEBUG mode, test will fail in release mode.

OsmondJiang: zan

PR #3573, opened 2018-11-02
Fix link to Homebrew package

franklinyu:
The link to unstable http://brewformulas.org is replaced with official https://formulae.brew.sh. Currently the former is down for me.


yufeih:
Thank you Franklin!

PR #3580, opened 2018-11-06
wrap directory.createdirectory

mingwli:
only create directory when path is not null or empty

### DIFF
--- a/src/docfx/lib/PathUtility.cs
+++ b/src/docfx/lib/PathUtility.cs
@@ -126,6 +126,21 @@ namespace Microsoft.Docs.Build
             return Normalize(path);
         }
 
+        /// <summary>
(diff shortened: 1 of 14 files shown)


src/docfx/lib/PathUtility.cs
```diff
@@ -126,6 +126,21 @@ public static string NormalizeFile(string path)
...
 
+        /// <summary>
+        /// Create a new directory. Return true if creation succeeds or directory already exists.
+        /// </summary>
+        /// <param name="path">The path to create</param>
+        /// <returns>True if creation succeeds or directory already exists</returns>
+        public static bool CreateDirectoryIfNotEmpty(string path)
+        {
+            if (string.IsNullOrEmpty(path))
+                return true;
```
herohua: looks weird that an invalid path returns true

PR #3581, opened 2018-11-07
Create directory from filepath

mingwli:


src/docfx/lib/git/GitUtility.cs
```diff
@@ -156,7 +155,7 @@ public static Task<List<string>> ListWorkTrees(string cwd, bool includeMain)
         /// Create a work tree for a given repo
         /// </summary>
         /// <param name="cwd">The current working directory</param>
-        /// <param name="commitHash">The commit hash or branch name you want to use to create a work tree</param>
+        /// <param name="commitIsh">The commit hash or branch name you want to use to create a work tree</param>
```
live1206: commitHash?

PR #3583, opened 2018-11-07
Make git operations support committish

yufeih:
This PR makes all git operations support committish: branch, tags, short hash, etc... Unblocks deployment due to impact test uses commit hash in dependency configuration.

- Make GetCommitHistory, Fetch support committish (fetch will first attempt to fetch single branch, then fallback to fetch all tags and branches)
- Split stdout and stderr from process exec, to detect fetch errors while preserving fetch progress
- Adjust a test case to use short hash

### DIFF
--- a/src/docfx/lib/git/GitCommitProvider.cs
+++ b/src/docfx/lib/git/GitCommitProvider.cs
@@ -255,14 +255,16 @@ namespace Microsoft.Docs.Build
             {
                 var error = git_revwalk_next(out var commitId, walk);
                 if (error == -31 /* GIT_ITEROVER */)
+                {
                     break;
+                }
 
                 // https://github.com/libgit2/libgit2sharp/issues/1351
-                if (error == -3 /* GIT_ENOTFOUND */)
-                    throw Errors.GitShadowClone(_repoPath).ToException();
-
-                if (error != 0)
-                    throw new InvalidOperationException($"Unknown error calling git_revwalk_next: {error}");
+                if (error != 0 /* GIT_ENOTFOUND */)
+                {
+                    git_revwalk_free(walk);
+                    throw Errors.GitLogError(_repoPath, error).ToException();
+                }
 
                 git_object_lookup(out var commit, _repo, &commitId, 1 /* GIT_OBJ_COMMIT */);
                 var author = git_commit_author(commit);
--- a/src/docfx/lib/git/GitCommitProvider.cs
+++ b/src/docfx/lib/git/GitCommitProvider.cs
@@ -65,14 +65,14 @@ namespace Microsoft.Docs.Build
             return new GitCommitProvider(repoPath, cacheFilePath, await LoadCommitCache(cacheFilePath));
         }
 
-        public List<GitCommit> GetCommitHistory(string file, string branch = null)
+        public List<GitCommit> GetCommitHistory(string file, string committish = null)
         {
             Debug.Assert(!file.Contains('\\'));
 
             const int MaxParentBlob = 32;
 
             var (commits, commitsBySha) = _commits.GetOrAdd(
-                branch ?? "",
+                committish ?? "",
                 key => new Lazy<(List<Commit>, Dictionary<long, Commit>)>(() => LoadCommits(key))).Value;
 
             if (commits.Count <= 0)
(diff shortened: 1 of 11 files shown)


src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -228,45 +228,43 @@ public void Dispose()
...
-            }
+            git_revwalk_push(walk, git_object_id(headCommit));
+            git_object_free(headCommit);
 
             while (true)
             {
                 var error = git_revwalk_next(out var commitId, walk);
                 if (error == -31 /* GIT_ITEROVER */)
+                {
                     break;
```
live1206: should `walk` be freed while exiting as well?

src/docfx/lib/git/GitUtility.cs
```diff
@@ -174,10 +174,30 @@ public static Task PruneWorkTrees(string cwd)
...
+            string result = null;
+
+            if (string.IsNullOrEmpty(committish))
+            {
+                committish = "HEAD";
+            }
+
+            if (git_repository_open(out var repo, repoPath) != 0)
+            {
+                throw new InvalidOperationException($"Not a git repo {repoPath}");
```
live1206: should `repo` be freed before throwing?

src/docfx/lib/git/GitUtility.cs
```diff
@@ -174,10 +174,30 @@ public static Task PruneWorkTrees(string cwd)
...
+            }
+
+            if (git_revparse_single(out var reference, repo, committish) == 0)
+            {
+                result = git_object_id(reference)->ToString();
+                git_object_free(reference);
```
live1206: free `reference` outside `if` block?

src/docfx/lib/git/GitCommitProvider.cs
yufeih: it is freed after while loop

src/docfx/lib/git/GitUtility.cs
yufeih: when `git_revparse_single` fails (returns non zero), reference is an invalid point, you can only free a valid pointer.

src/docfx/lib/git/GitUtility.cs
yufeih: nope

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -231,45 +228,43 @@ public void Dispose()
             }
         }
 
-        private unsafe (List<Commit>, Dictionary<long, Commit>) LoadCommits(string branchName = null)
+        private unsafe (List<Commit>, Dictionary<long, Commit>) LoadCommits(string committish = null)
         {
```
OsmondJiang: >committish [](start = 63, length = 10)

how about the local and remote?

test/docfx.Test/lib/GitUtilityTest.cs
```diff
@@ -59,16 +59,6 @@ public static async Task GetCommitsSameAsGitExe(string file)
             }
         }
 
-        [Fact]
-        public static async Task GitCommandConcurreny()
-        {
-            var cwd = GitUtility.FindRepo(Path.GetFullPath("README.md"));
```
OsmondJiang: >var cwd = GitUtility.FindRepo(Path.GetFullPath("README.md")); [](start = 12, length = 61)

why remove this test?

yufeih: moved to ProcessUtilityTest

test/docfx.Test/lib/GitUtilityTest.cs
```diff
@@ -48,8 +48,8 @@ public static async Task GetCommitsSameAsGitExe(string file)
                     string.Join("\n", lib.Select(c => $"{c.Sha}|{c.Time.ToString("s")}{c.Time.ToString("zzz")}|{c.AuthorName}|{c.AuthorEmail}")));
 
                 // another branch
-                exe = Exec("git", $"--no-pager log --format=\"%H|%cI|%an|%ae\" origin/test -- \"{pathToRepo}\"", repo);
-                lib = commitsProvider.GetCommitHistory(pathToRepo, "test");
+                exe = Exec("git", $"--no-pager log --format=\"%H|%cI|%an|%ae\" a050eaf -- \"{pathToRepo}\"", repo);
```
OsmondJiang: >a050eaf [](start = 79, length = 7)

also add a normal other branch name test?

yufeih: this test fails if your local branch does not have the same commit as server, or if origin is a forked repo that does not have a test branch (like me), a commit hash will always have a the same result.

PR #3584, opened 2018-11-07
skip e2eTest temporarily

928PJY:


superyyrrzz:
The version we need is still on mozilla site. We can add it back later.

https://ftp.mozilla.org/pub/firefox/releases/46.0.1/

PR #3586, opened 2018-11-07
Split content to files and exclude

yufeih:
#3513 

Split `content` to `files` and `exclude`, after this change the only supported `oneof` syntax is _token_ or _token array_

cc @LucasDvp @mingwli


src/docfx/config/Config.cs
```diff
@@ -23,9 +26,14 @@ internal sealed class Config
...
-        /// Gets the contents that are managed by this docset.
+        /// Gets the file glob patterns included by the docset.
+        /// </summary>
+        public readonly string[] Content = s_defaultContentInclude;
+
+        /// <summary>
+        /// Gets the file glob patterns excluded from this docset.
         /// </summary>
-        public readonly FileConfig Content = new FileConfig();
+        public readonly string[] ContentExclude = s_defaultContentExclude;
```
superyyrrzz: Naming: what about `files` and `exclude` as in v2? Jekyll also has a `exclude` at config's top level.

yufeih: sure

src/docfx/config/Config.cs
```diff
@@ -23,9 +25,14 @@ internal sealed class Config
         public readonly string Name = string.Empty;
 
         /// <summary>
-        /// Gets the contents that are managed by this docset.
+        /// Gets the file glob patterns included by the docset.
+        /// </summary>
+        public readonly string[] Files = s_defaultInclude;
```
OsmondJiang: how about
```
content:
  include:
  exclude:
```

OsmondJiang: or use `!` to represent exclude, only `files`
```
files:
  - a.md
  - !b.md
```

yufeih:
Test failed due to configs in dependent repos, @superyyrrzz @OsmondJiang , I'll continue this PR when you guys finished migrating test cases.

src/docfx/config/Config.cs
yufeih: The final schema is `files` and `exclude`.

PR #3588, opened 2018-11-08
add bilingual metadata to legacy model

OsmondJiang:
template need know whether this page is bilingual or not, so here we add a flag to our output model, and also add legacy model metadata


mingwli:
is there any spec reference?

OsmondJiang:
this is a legacy behavior, at least we need the template knows that the page is bilingual

---
In reply to: [436856622](https://github.com/dotnet/docfx/pull/3588#issuecomment-436856622) [](ancestors = 436856622)

live1206:
Maybe some description would help :)

OsmondJiang:
you are right, just added, you guys too fast...

---
In reply to: [436857453](https://github.com/dotnet/docfx/pull/3588#issuecomment-436857453) [](ancestors = 436857453)

src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -111,6 +111,11 @@ public static JObject GenerateLegacyRedirectionRawMetadata(Docset docset, PageMo
             if (pageModel.UpdatedAt != default)
                 rawMetadata["updated_at"] = pageModel.UpdatedAt.ToString("yyyy-MM-dd hh:mm tt", docset.Culture);
 
+            if (pageModel.Bilingual)
+            {
+                rawMetadata["bilingual_type"] = "hover over";
```
yufeih: https://ceapex.visualstudio.com/Engineering/_workitems/edit/20449

Do we want to be consistent with template and loc team and change the term `bilingual` to `enable_loc_sxs` ?

OsmondJiang: i don't want to it right now since it causes big impact changes right now, i will log a issue for that

OsmondJiang: [https://github.com/dotnet/docfx/issues/3590](https://github.com/dotnet/docfx/issues/3590)

---
In reply to: [231770644](https://github.com/dotnet/docfx/pull/3588#discussion_r231770644) [](ancestors = 231770644)

PR #3599, opened 2018-11-09
add supported localized tokens to markup

OsmondJiang:
remove `todo` in markup, support localized token in markup.


src/docfx/build/page/Markup.cs
```diff
@@ -104,9 +85,29 @@ public static (MarkdownDocument ast, MarkupResult result) Parse(string content)
             }
         }
 
-        private static MarkdownPipeline CreateConceptualMarkdownPipeline()
+        private static MarkdownPipeline GetPipeline(MarkdownPipelineType pipelineType, string locale)
+        {
+            return s_pipelineMapping.GetOrAdd(pipelineType, key => new Lazy<MarkdownPipeline>(() =>
```
yufeih: This is wrong if we build multiple locales in a single process
Keep the existing pipeline map. put `locale` into the thread static `Status` object.

docfx.sln
```diff
@@ -16,6 +16,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "tools", "tools", "{7C85AA4C
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "CreateJsonSchema", "tools\CreateJsonSchema\CreateJsonSchema.csproj", "{A899DBCD-C7A3-4372-AA0C-5FA542759BBB}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ImportTemplateResource", "tools\ImportTemplateResource\ImportTemplateResource.csproj", "{B52B4C8B-DFF2-43CF-B7B5-3DA058CCD6F0}"
```
yufeih: since this is mostly an ad-hoc tool, can we put it somewhere else like a personal github repo?

OsmondJiang: i'd like to keep it in tools in case we need run it again and again before new template is implemented, and also we can added it to CI job

OsmondJiang:
the impact test shows the localized tokens has been applied but the spec test still have some problem to apply the resx file, i am working on it, but it;s not blocked, we can start the review. ;) @live1206  @yufeih

superyyrrzz:
>             { "Caution", "<p>Caution</p>" },

For Markup, only these 5 tokens are needed. All the remaining are template tokens instead of Markup tokens. What about adding a white list to make the resx smaller?

---
Refers to: src/docfx/build/page/Markup.cs:36 in 74e11a0. [](commit_id = 74e11a0b85e7a8c539c25beed081de17389bce5a, deletion_comment = True)

src/Microsoft.Docs.Template/resources/tokens.ar-SA.resx
```diff
@@ -0,0 +1,135 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!--
```
yufeih: nit: can we get rid of these large section of comments?

OsmondJiang: its' a default schema template, which can help to integrate the resx with vs

src/docfx/properties/AssemblyInfo.cs
```diff
@@ -5,3 +5,4 @@
 
 [assembly: InternalsVisibleTo("docfx.Test")]
 [assembly: InternalsVisibleTo("CreateJsonSchema")]
+[assembly: InternalsVisibleTo("ImportTemplateResource")]
```
yufeih: revert

src/docfx/build/page/Markup.cs
```diff
@@ -50,13 +42,14 @@ internal static class Markup
...
 
-        public static (MarkdownDocument ast, MarkupResult result) Parse(string content)
+        public static (MarkdownDocument ast, MarkupResult result) Parse(string content, CultureInfo culture)
         {
             try
             {
                 var status = new Status
                 {
                     Result = new MarkupResult(),
+                    Culture = culture,
```
yufeih: It does not affect `Parse`, tokens are only used in the `ToHtml` part.

OsmondJiang: yes, reverting

src/docfx/docfx.csproj
```diff
@@ -30,6 +30,7 @@
     <PackageReference Include="Stubble.Core" Version="1.1.2" />
     <PackageReference Include="StyleCop.Analyzers" Version="1.1.0-beta006" PrivateAssets="All" />
     <PackageReference Include="System.CommandLine" Version="0.1.0-e171109-2" />
+    <PackageReference Include="System.Resources.ResourceManager" Version="4.3.0" />
```
yufeih: this is already referenced implicitly I think.

PR #3600, opened 2018-11-09
Move output path to manifest

yufeih:
Moves `OutputPath` outside `Document` to `Manifest` because we don't know the output path until build is complete (in case of moniker).

One notable behavior change is that when `copyResources` is false, instead of not output `OutputPath` in manifest, it set `OutputPath` to source relative path, to make it easier to consume by publish.

This PR changes `Build` produces a `Manifest`, but due to dependencies on `Document` in legacy, `Build` currently also output `Document[]`. It'll be cleaned up in upcoming PR.


src/docfx/build/Build.cs
```diff
@@ -162,33 +162,65 @@ void ValidateBookmarks()
...
-                if (model != null && !hasErrors)
+                if (hasErrors || model == null)
                 {
-                    if (model is string str)
-                        context.WriteText(str, file.OutputPath);
-                    else
-                        context.WriteJson(model, file.OutputPath);
-
-                    return (false, dependencies);
+                    manifestBuilder.MarkError(file);
```
live1206: how about extracting manifest related logic into an inline method for readability?

src/docfx/docset/Document.cs
```diff
@@ -100,6 +87,12 @@ internal class Document
 
         private readonly Lazy<(string docId, string versionIndependentId)> _id;
 
+        // TODO:
+        // This is a temporary property just so that legacy can access OutputPath,
+        // I'll slowly converge legacy into main build and remove this property eventually.
+        // Do not use this property in main build.
+        internal string OutputPath;
```
928PJY: >OutputPath [](start = 24, length = 10)

so in the future, legacy should get the `outputPath` from manifest, right?

yufeih: yes

src/docfx/build/Build.cs
yufeih: yeah, the refactor hasn't been completed yet, especially around the legacy part, it'll be in another PR.

PR #3604, opened 2018-11-12
Make ListWorkTreePath use --porcelain output

yufeih:
`--porcelain` is stable across git versions, the plan output is not.

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -127,28 +127,38 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// List work trees for a given repo
         /// </summary>
-        /// <param name="cwd">The current working directory</param>
-        public static Task<List<string>> ListWorkTrees(string cwd, bool includeMain)
+        public static Task<List<string>> ListWorkTree(string repoPath)
         {
-            return Execute(cwd, $"worktree list", ParseWorkTreeList);
+            return Execute(repoPath, $"worktree list --porcelain", ParseWorkTreeList);
 
             List<string> ParseWorkTreeList(string stdout, string stderr)
             {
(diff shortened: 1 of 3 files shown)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -127,28 +127,38 @@ public static Task CloneOrUpdateBare(string path, string url, IEnumerable<string
...
                 Debug.Assert(stdout != null);
-                var worktreeLines = stdout.Split(s_newline, StringSplitOptions.RemoveEmptyEntries);
-                var workTreePaths = new List<string>();
 
-                var i = 0;
-                foreach (var workTreeLine in worktreeLines)
+                // https://git-scm.com/docs/git-worktree#_porcelain_format
+                var result = new List<string>();
+                var isMain = true;
+                foreach (var property in stdout.Split("\n", StringSplitOptions.RemoveEmptyEntries))
```
live1206: not using `s_newline` anymore?

src/docfx/lib/git/GitUtility.cs
```diff
@@ -127,28 +127,38 @@ public static Task CloneOrUpdateBare(string path, string url, IEnumerable<string
...
                 {
-                    if (i++ > 0 || includeMain)
+                    var i = property.IndexOf(' ');
```
live1206: `s_newlineTab`

src/docfx/lib/git/GitUtility.cs
yufeih: yes, used to be `string.Split` only has a `split(params char[])` overload, passing `'\n'` incurs a tiny allocation, so it is extracted as a global variable. Now dotnet core added an overload `split(char)`, passing `'\n'` here is as efficient as before.

PR #3605, opened 2018-11-12
Refactor appdata path logic

yufeih:
Consolidates all appdata path calculation logic to Appdata.cs,

_fixes_:
- git repository path should escape url
- file download path should respect query string

_renames_:
- RestoreUrl -> RestoreFile
- GetUrlRestorePath -> GetFileRestorePath

### DIFF
--- a/src/docfx/config/AppData.cs
+++ b/src/docfx/config/AppData.cs
@@ -8,34 +8,34 @@ namespace Microsoft.Docs.Build
 {
     internal static class AppData
     {
-        public static readonly string AppDataDir = GetAppDataDir();
+        private static readonly string s_root = GetAppDataRoot();
 
-        public static string GitDir => Path.Combine(AppDataDir, "git");
+        public static string GitRoot => Path.Combine(s_root, "git");
 
-        public static string DownloadsDir => Path.Combine(AppDataDir, "downloads");
+        public static string DownloadsRoot => Path.Combine(s_root, "downloads");
 
-        public static string MutexDir => Path.Combine(AppDataDir, "mutex");
+        public static string MutexRoot => Path.Combine(s_root, "mutex");
 
-        public static string CacheDir => Path.Combine(AppDataDir, "cache");
+        public static string CacheRoot => Path.Combine(s_root, "cache");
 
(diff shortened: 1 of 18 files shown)


src/docfx/config/AppData.cs
```diff
@@ -10,16 +10,48 @@ internal static class AppData
...
 
         public static string MutexDir => Path.Combine(AppDataDir, "mutex");
 
         public static string CacheDir => Path.Combine(AppDataDir, "cache");
 
         public static string GlobalConfigPath => GetGlobalConfigPath();
 
+        public static string GitHubUserCachePath => Path.Combine(CacheDir, "github-users.json");
+
+        public static string GetGitDir(string url)
```
OsmondJiang: >GetGitDir [](start = 29, length = 9)

GetGitDir and GitDir may cause confusion

src/docfx/config/AppData.cs
```diff
@@ -10,16 +10,48 @@ internal static class AppData
     {
         public static readonly string AppDataDir = GetAppDataDir();
 
-        public static string GitRestoreDir => Path.Combine(AppDataDir, "git");
+        public static string GitDir => Path.Combine(AppDataDir, "git");
```
OsmondJiang: GitRootDir to avoid confusion?

src/docfx/config/AppData.cs
```diff
@@ -8,33 +8,65 @@ namespace Microsoft.Docs.Build
...
+
+        public static string GetFileDownloadDir(string url)
+        {
+            // URL to a resource is case sensitive, query string matters, so hash the download path
+            return PathUtility.NormalizeFolder(Path.Combine(DownloadsRoot, UrlToPath(url) + "-" + url.Trim().GetMd5HashShort()));
+        }
+
+        public static string GetCommitCachePath(string remote)
+        {
+            return Path.Combine(AppData.CacheRoot, "commits", HashUtility.GetMd5Hash(remote));
```
live1206: >AppData [](start = 32, length = 7)

can be removed

yufeih: yes

Issue #3607, opened 2018-11-13
conceptual versioning in v3


928PJY:
[Dev spec](https://github.com/dotnet/docfx/blob/v3/docs/designs/conceptual-versioning.md)
# Phase 1 Tasks:
- [x] Support setting `monikerRange` and `monikerDefinitionUrl` in config. (@928PJY )
- [x] Support setting `monikerRange` in yaml header. (@live1206 )
- [x] Support `moniker zone` syntax in conceptual markdown file (@928PJY )
- [x] Internal Xref map with moniker info (@live1206 )
- [x] Input/output mapping (@928PJY )
- [x] File output content
    - [x] Legacy output(@live1206 )
    - [x] Conceptual file
    - [x] Redirection file (@928PJY )
    - [x] Toc file(@928PJY )
- [x] Build manifest file(@live1206 )
- [x] ~~Blank page support ?~~
- [x] API to get moniker definition file (@928PJY )
    - Convert the output of current API `/v2/monikertrees/allfamiliesproductsmonikers` to our expected model
    - Add two config items to global config in `Docs.Build`:
    ```
    monikerDefinition: https://op-build-sandbox2.azurewebsites.net/v2/monikertrees/allfamiliesproductsmonikers
    http:
        https://op-build-sandbox2.azurewebsites.net/v2/monikertrees/allfamiliesproductsmonikers:
            headers:
                X-OP-BuildUserToken: xxxx-xxxxx
    ```
- [ ] Impact repo test & migration tool support (@928PJY )
![pipeline](https://user-images.githubusercontent.com/17958546/48385749-af455780-e72a-11e8-9372-ab69794b1c59.jpg)

# Dependencies
- [ ] Support publishing with overlapping monikerRange to DHS.
- [x] Publish with the new manifest file(depend on the new manifest design & implementation).

# Remaining work after phase1:
- [ ] Reference validation:
    - [x] refenrece to internal uid with specific moniker 
    - [ ] refenrece to external uid with specific moniker 
    - [ ] refenrece to another file with specific moniker
- [ ] Output xref map with moniker info.
- [ ] Static site
    - [ ] implement fallback logic on build side.

928PJY:
Repository contains versioning:
- https://github.com/MicrosoftDocs/sql-docs-pr
- https://github.com/MicrosoftDocs/OfficeDocs-O365ITPro-pr
- https://github.com/aspnet/Docs
- https://github.com/MicrosoftDocs/CordovaDocs
[...]

PR #3610, opened 2018-11-13
Apply yaml header moniker range to page building.

live1206:
#3607 
- input null/empty/whitespace string, `MonikerRangeParser` output empty string list
- apply YAML header moniker range to page building

### DIFF
--- a/docs/designs/conceptual-versioning.md
+++ b/docs/designs/conceptual-versioning.md
@@ -1,0 +1,571 @@
+# Versioning Dev Spec
+
+This document specifies Docfx vnext versioning dev design.
+
+## Conceptual markdown versioning - Phase 1
+
+### 1 Scope
+
+#### 1.1 In scope
+
+##### For content writer
+
+1. Support expressing version constrains as moniker range.
+
+    In order to ease version configuration and content maintenance, [moniker range](./versioning.md#moniker-ranges) enable the writer to associate more than one moniker with content, but without needing to list monikers.
+
+2. Support moniker zone markdown syntax.
+
+3. Support publishing multiple files with same *sitePath* but different monikerRange(mutually exclusive with others), where the appropriate file is selected by URL with query `view={moniker}`.
+
+4. All files will be built in one round, which will bring a better performance.
+
+##### For content reader
+
+1. Support viewing page with specific version by query `?view={moniker}`. If the version is not existed, they should be redirected to some other version, but not get a `404` page.
+
+2. Support selecting Toc version in available versions.
+
+3. Prevent user to visit a [blank page](#42-blank-page) in different version.
+
+4. Ensure TOC always only show available (non-empty) page node under specific version.
+
+#### 1.2 Out of scope
+
+1. Not support version validation on reference, including link and xref.
+
+    In phase 1, we will not check whether the version is existed on the referenced file.
+
+    1. If user reference another file by link/Uid without query string `?view={moniker}`, we will not check whether the referenced file have the same monikerRange with current file. So if the referenced file have different monikerRange, the final view page is not guaranteed.
+
+        Sample:
+    
+        ```txt
+        |- articles/
+        |    |- folder1/
+        |    |   |- a.md
+        |    |- folder2/
+        |    |   |- b.md
+        |    |- folder3/
+        |        |- b.md
+        |- config.yml
+        ```
+    
+        the content of `config.yml` is
+        ```yml
+        name: dotnet
+        content: "articles/**/*.md"
+        monikerRange:
+            "articles/folder1/**/*.md": "netcore-1.0"
+            "articles/folder2/**/*.md": "netcore-1.0"
+            "articles/folder3/**/*.md": "netcore-2.0"
+        routing:
+            "articles/folder1/": "articles/"
+            "articles/folder2/": "articles/"
+            "articles/folder3/": "articles/"
+        monikerDefinitionUrl: "https://api.docs.com/monikers/"
+        ```
+
+        If content writer want to reference `folder3/b.md` in `folder1/a.md` by link `'[B](../folder3/b.md)'`, when the reader click the link [B]() on page `{host}/{docset-base-path}/articles/a?view=netcore-1.0`, they will jump to page `{host}/{docset-base-path}/articles/b?view=netcore-1.0`, which is the output of file `folder2/b.md`.
+
+    2. If user reference another file by link/Uid with query string `?view={moniker}`, we will not check whether the referenced file contains this `moniker`, the query string will be preserved to the resolve result, but the final view page is not guaranteed.
+
+        Sample:
+
+        ```txt
+        |- articles/
+        |    |- folder1/
+        |    |   |- a.md
+        |    |- folder2/
+        |    |   |- b.md
+        |    |- folder3/
+        |        |- b.md
+        |- config.yml
+        ```
+
+        the content of `config.yml` is
+        ```yml
+        name: dotnet
+        content: "articles/**/*.md"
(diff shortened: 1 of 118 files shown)


src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -20,17 +21,25 @@ public MonikersProvider(Config config, MonikerRangeParser monikerRangeParser)
             _rules.Reverse();
         }
 
-        public List<string> GetMonikers(Document file)
+        public List<string> GetMonikers(Document file, JObject yamlHeader = null)
```
yufeih: Add `MonikerRange` to `FileMetadata` and use it as the input

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -20,17 +21,25 @@ public MonikersProvider(Config config, MonikerRangeParser monikerRangeParser)
...
+                    result = monikers;
                 }
             }
-            return new List<string>();
+
+            if (yamlHeader != null && yamlHeader.HasValues)
+            {
+                var monikerRange = yamlHeader.Value<string>("monikerRange");
+                var yamlHeaderMonikers = file.Docset.MonikerRangeParser.Parse(monikerRange);
+                return result.Intersect(yamlHeaderMonikers).ToList();
```
928PJY: >result.Intersect(yamlHeaderMonikers).ToList(); [](start = 23, length = 46)

report a warning when the intersection is empty?

src/docfx/build/moniker/MonikersProvider.cs
928PJY: where do we remove it? or it will  be output to .raw.page.json？

---
In reply to: [232901220](https://github.com/dotnet/docfx/pull/3610#discussion_r232901220) [](ancestors = 232901220)

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -20,17 +21,25 @@ public MonikersProvider(Config config, MonikerRangeParser monikerRangeParser)
...
-            // TODO: merge with the monikers from yaml header
+            var result = new List<string>();
             foreach (var (glob, monikers) in _rules)
             {
                 if (glob(file.FilePath))
                 {
-                    return monikers;
+                    result = monikers;
                 }
             }
```
928PJY: report  a warning if `monikerRange` hasn't been set in the config but has been set in the yaml header

test/docfx.Test/lib/MonikerRangeParserTest.cs
```diff
@@ -85,7 +88,7 @@ public MonikerRangeParserTest()
         public void TestEvaluateMonikerRange(string rangeString, string expectedMonikers)
         {
             var result = _monikerRangeParser.Parse(rangeString);
-            Assert.True(Enumerable.SequenceEqual(expectedMonikers.Split(' ', StringSplitOptions.RemoveEmptyEntries), result));
+            Assert.Equal(expectedMonikers, string.Join(' ', result));
```
928PJY: >Assert.Equal(expectedMonikers, string.Join(' ', result)); [](start = 12, length = 57)

:thumbsup:

docs/specs/moniker.yml
```diff
@@ -41,6 +41,54 @@ outputs:
       "monikers": []
     }
 ---
+# Apply moniker range setting in the config and yaml header to markdown file's output path and output content
```
OsmondJiang: >Apply moniker r [](start = 2, length = 15)

add test case that monikers is not defined in config but in files

OsmondJiang: and also moniker range in filer is bigger than config

---
In reply to: [232915502](https://github.com/dotnet/docfx/pull/3610#discussion_r232915502) [](ancestors = 232915502)

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -9,28 +9,54 @@ namespace Microsoft.Docs.Build
...
             foreach (var (key, monikerRange) in config.MonikerRange)
             {
-                _rules.Add((GlobUtility.CreateGlobMatcher(key), monikerRangeParser.Parse(monikerRange)));
+                _rules.Add((GlobUtility.CreateGlobMatcher(key), (monikerRange, monikerRangeParser.Parse(monikerRange))));
             }
             _rules.Reverse();
         }
 
-        public List<string> GetMonikers(Document file)
+        public (Error, List<string>) GetMonikers(Document file, string fileMonikerRange = null)
```
928PJY: >Error [](start = 16, length = 5)

why not keep the consistency to return `List<Error>`?

live1206: Because it's just one error, think we should not use a list if it is just one

src/docfx/build/page/BuildPage.cs
```diff
@@ -21,7 +21,8 @@ internal static class BuildPage
             Debug.Assert(file.ContentType == ContentType.Page);
 
             var (errors, schema, model, yamlHeader) = await Load(context, file, callback);
-            var (metaErrors, metadata) = JsonUtility.ToObject<FileMetadata>(file.Docset.Metadata.GetMetadata(file, yamlHeader));
+            var temp = file.Docset.Metadata.GetMetadata(file, yamlHeader);
```
yufeih: maybe `GetMetadata` returns `FileMetadata`?

src/docfx/Errors.cs
```diff
@@ -153,6 +153,15 @@ public static Error MonikerNameConflict(string monikerName)
...
             => new Error(ErrorLevel.Error, "invalid-moniker-range", $"MonikerRange `{monikerRange}` is invalid: {message}");
 
+        public static Error MonikerConfigMissing()
+            => new Error(ErrorLevel.Warning, "moniker-config-missing", "Moniker range missing in docfx.yml/docfx.json, user should not define it in file metadata.");
+
+        public static Error NoMonikersIntersection(string configMonikerRange, List<string> configMonikers, string fileMonikerRange, List<string> fileMonikers)
+            => new Error(
+                ErrorLevel.Warning,
+                "no-moniker-intersection",
+                $"No moniker intersection between docfx.yml/docfx.json and file metadata. Config moniker range {configMonikerRange} is {string.Join(',', configMonikers)}, while file moniker range {fileMonikerRange} is {string.Join(',', fileMonikers)}");
```
yufeih: There is a `Join` helper method

src/docfx/build/page/BuildPage.cs
live1206: was using it for debugging, removed

src/docfx/Errors.cs
live1206: right

docs/designs/conceptual-versioning.md
```diff
@@ -202,7 +202,10 @@ monikerRange: >=netcore-1.0
 ```
 
 > [!NOTE]
-> The final file level moniker range is the intersection of moniker range from `config file` and moniker range from `YAML header`, if the intersection is empty, an error throws.
+> The final file level moniker range is the intersection of moniker range from `config file` and moniker range from `YAML header`, if the intersection is empty, a warning will be logged.
+
+> [!NOTE]
+> The `config file` level moniker range should be defined to enable versioning. If moniker range is not defined in `config file`, and `Yaml header` moniker range is defined, a warning will be logged.
```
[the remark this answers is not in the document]
live1206: Yes

---
In reply to: [233285213](https://github.com/dotnet/docfx/pull/3610#discussion_r233285213) [](ancestors = 233285213)

PR #3613, opened 2018-11-13
Refactor and simplify restore

yufeih:
This PR further simplifies restore logic:

- Remove return value from `RestoreFile` and `RestoreGit`, `RestoreMap` no longer need them
- Avoid multiple `GetGitRemoteInfo` calls by grouping git dependencies by remote
- Consolidate `RestoreWorkTree` to `RestoreGit` since there aren't too much logic left
- Some other minor refactors like extract method, naming consistency


src/docfx/gc/GarbageCollector.cs
```diff
@@ -72,12 +71,10 @@ private static int CollectUrls(int retentionDays)
                 return cleaned;
             }
 
-            var downloadedFiles = Directory.EnumerateFiles(AppData.DownloadsRoot, "*", SearchOption.AllDirectories);
-
             using (Progress.Start("Cleaning downloaded files"))
             {
                 ParallelUtility.ForEach(
-                    downloadedFiles,
+                    Directory.EnumerateFiles(AppData.DownloadsRoot, "*", SearchOption.AllDirectories),
```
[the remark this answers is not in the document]
yufeih: this is to ensure `EnumerateFiles` is called inside `Progress.Start`, I can keep the existing variable

src/docfx/restore/RestoreMap.cs
```diff
@@ -24,31 +26,30 @@ public string GetGitRestorePath(string url)
...
+
+                if (!Directory.Exists(restoreDir))
+                {
+                    throw Errors.NeedRestore(url).ToException();
+                }
+
+                return Directory.EnumerateDirectories(restoreDir, "*", SearchOption.TopDirectoryOnly)
+                    .Where(f => f.EndsWith(HrefUtility.EscapeUrlSegment(branch)))
+                    .OrderByDescending(f => new DirectoryInfo(f).LastWriteTimeUtc)
+                    .FirstOrDefault();
```
OsmondJiang: lock before using

yufeih: Let's do it in another PR. lock is async, I am tentative to change this method to an async method.

Issue #3609, opened 2018-11-13
Remaining issues with JsonUtility 


yufeih:
- [x] When loading data files, essential errors like type mismatch is ignored. Need to provide 2 separate APIs for loading editable files and data files.: `(List<Error> T) DeserializeWithSchemaValidation(string)` and `T Deserialize<T>(string)`.

- [ ] ~~Most of our non-breakable data format uses _snake_casing_ (like output to template), may want to be consistent with just _snake_casing_ and do it at `JsonUtility` level.~~

PR #3617, opened 2018-11-14
Throw exception when de-serializing data files

live1206:
#3609 
- Throw type mismatching and syntax errors while de-serializing not user input JSON/YAML files
- Add tests for them


src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -232,7 +232,7 @@ private async Task LoadCommitsTime(Docset docset)
                 var path = docset.RestoreMap.GetFileRestorePath(docset.Config.Contribution.GitCommitsTime);
                 var content = await ProcessUtility.ReadFile(path);
 
-                foreach (var commit in JsonUtility.Deserialize<GitCommitsTime>(content).Item2.Commits)
+                foreach (var commit in JsonUtility.DeserializeData<GitCommitsTime>(content).Commits)
                 {
```
OsmondJiang: why need schema validation here?

src/docfx/lib/JsonUtility.cs
```diff
@@ -145,7 +145,7 @@ public static string Serialize(object graph, Formatting formatting = Formatting.
         }
 
         /// <summary>
-        /// De-serialize a string to an object
+        /// De-serialize a user input string to an object, return error list at the same time
         /// </summary>
         public static (List<Error> errors, T model) Deserialize<T>(string json)
         {
```
OsmondJiang: better name the method with its functions, not usage...

src/docfx/lib/JsonUtility.cs
```diff
@@ -145,7 +145,7 @@ public static string Serialize(object graph, Formatting formatting = Formatting.
         }
 
         /// <summary>
-        /// De-serialize a string to an object
+        /// De-serialize a user input string to an object, return error list at the same time
         /// </summary>
         public static (List<Error> errors, T model) Deserialize<T>(string json)
```
OsmondJiang: what about add a flag to the parameter to indicate whether to validate schema?

OsmondJiang: default is true

src/docfx/lib/JsonUtility.cs
yufeih: Whatever name you choose, keep it in sync with `YamlUtility`

PR #3618, opened 2018-11-14
Ensure restore map is a valid git repo

yufeih:
- Do not checkout original branch when bilingual is on (only fetch)
- Use `HEAD` file to check if a worktree is valid
- `GetGitRemoteInfo` -> `SplitGitHref` to align with `SplitHref`

### DIFF
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Docs.Build
             async Task RestoreGitRepo(IGrouping<string, string> group)
             {
                 var remote = group.Key;
-                var branches = group.Distinct();
+                var branches = group.Distinct().ToArray();
                 var repoPath = Path.GetFullPath(Path.Combine(AppData.GetGitDir(remote), ".git"));
                 var childRepos = new List<string>();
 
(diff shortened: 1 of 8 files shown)


src/docfx/restore/RestoreGit.cs
```diff
@@ -54,9 +54,16 @@ async Task AddWorkTrees()
 
                     await ParallelUtility.ForEach(branches, async branch =>
                     {
+                        // Bilingual repos ({branch}-sxs) only depend on non bilingual branch for git commit history,
```
OsmondJiang: that makes the convention everywhere...

src/docfx/restore/RestoreMap.cs
```diff
@@ -37,18 +37,24 @@ public string GetGitRestorePath(string url)
...
-                    .OrderByDescending(f => new DirectoryInfo(f).LastWriteTimeUtc)
-                    .FirstOrDefault();
+                return (
+                    from path in Directory.EnumerateDirectories(restoreDir, "*", SearchOption.TopDirectoryOnly)
+                    let name = Path.GetFileName(path)
+                    where name.StartsWith(HrefUtility.EscapeUrlSegment(branch) + "-") &&
+
+                        // Adding the HEAD file is the last step in a worktree checkout, use it to
+                        // filter out worktrees that are still in progress.
+                        File.Exists(Path.Combine(restoreDir, ".git/worktrees", name, "HEAD"))
```
OsmondJiang: do we have a doc to reference?

src/docfx/restore/RestoreGit.cs
live1206: how about having a method in `LocalizationConvention.cs` to hide `-sxs`:
```csharp
public static string GetBilingualBranch(string branch) => $"{branch}-sxs";
```
Anywhere we need to access it, just use this one.

src/docfx/restore/RestoreMap.cs
yufeih: there is no doc about it, I digged into list worktree and add worktree implementation

live1206: Maybe we can comment the related code link to our code for reference

PR #3627, opened 2018-11-15
pdf metadata only shows when output pdf is enabled

OsmondJiang:


OsmondJiang:
https://docs.microsoft.com/en-us/azure/active-directory/develop/reference-v2-libraries, toc title and display name both can be searched , try search "MSAL" or "Authentication libraries"

OsmondJiang:
```
docs-ui code:
                               for (let i = 0; i < node.length; i++) {
				aNode = node[i];
				aCleanTitle = cleanText(aNode.toc_title);
				//if displayName exists on a TOC node, add it to the data-text attribute
				if (aNode.displayName && aNode.displayName.length) {
					displayName = cleanText(aNode.displayName);
				} else {
					displayName = '';
				}
```

Issue #3452, opened 2018-09-26
Make git related e2e test faster


yufeih:
git related e2e tests runs very slow, mainly because of restore and git clone operations.

yufeih:
Close this as test run time have been reduced to ~25s on average

PR #3633, opened 2018-11-16
docfx build triggers implicit restore

yufeih:
Similar to `dotnet build`, `docfx build` and `docfx watch` now triggers an implicit restore options. Implicit restores will not fetch git repositories or download files if one already exists on disk. This creates a better first time experience by saving the `docfx restore` command, and it does not impact the performance of `docfx build`.

The restored data on disk can be updated by doing an explicit restore using `docfx restore` command.
Similarly, run `docfx build --no-restore` to opt-out of implicit restore.

This resulted in a significant speed up for our own tests cases, now all of our tests cases runs within 10s if everything is available locally.

#3452


src/docfx/restore/RestoreMap.cs
```diff
@@ -9,40 +9,47 @@
...
+            result = s_gitPath.GetOrAdd((remote, branch), new Lazy<string>(FindLastModifiedGitRepository)).Value;
 
-            if (!Directory.Exists(gitRestorePath))
+            // GetOrAdd and TryRemove together are not atomic operation,
+            // but because we only call TryGetGitRestorePath in implicit restore,
+            // it is very unlikely to cause a problem.
+            if (result == null)
             {
-                throw Errors.NeedRestore(url).ToException();
+                s_gitPath.TryRemove((remote, branch), out _);
```
live1206: Why do we need to remove the null value? Since the return result is false, the required key `remote, branch` is not found.

yufeih: `s_gitPath` will cache `null`. If a folder does not exist and is later restored, the cache is still `null`.

src/docfx/restore/RestoreMap.cs
```diff
@@ -9,40 +9,47 @@
...
+            {
+                throw Errors.NeedRestore($"{remote}#{branch}").ToException();
+            }
+            return result;
+        }
 
-            var gitRestorePath = s_gitPath.GetOrAdd(url, new Lazy<string>(FindLastModifiedGitRepository)).Value;
+        public static bool TryGetGitRestorePath(string remote, string branch, out string result)
+        {
+            result = s_gitPath.GetOrAdd((remote, branch), new Lazy<string>(FindLastModifiedGitRepository)).Value;
```
OsmondJiang: >GetOrAdd [](start = 31, length = 8)

use AddOrUpdate

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -86,7 +85,7 @@ public static string GetLocalizationDocsetPath(string docsetPath, Config config,
                             repo.Branch,
                             locale,
                             config.Localization.DefaultLocale);
-                        var restorePath = restoreMap.GetGitRestorePath($"{locRemote}#{locBranch}");
+                        var restorePath = RestoreMap.GetGitRestorePath($"{locRemote}#{locBranch}");
```
OsmondJiang: >GetGitRestorePath [](start = 53, length = 17)

Why not just use GetGitRestorePath(remote, branch) ?

src/docfx/restore/RestoreMap.cs
yufeih: It does not solve the null cache problem

OsmondJiang: if it's null, you can use update function to restore again

PR #3641, opened 2018-11-19
add open issues to localization spec

OsmondJiang:
@yufeih  and @herohua , i believe the below to open issues need to be discussed during your trip


OsmondJiang:
cc @fenxu

Issue #3535, opened 2018-10-25
Markdown <a> post process wrong


mingwli:
[Example](https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/5455?path=%2Fsource%2Fazure-docs-pr.output&_a=files#path=%2Fsource%2Fazure-docs-pr.output%2Fazure%2Fcognitive-services%2FLUIS%2Fluis-how-to-manage-from-cloud-shell.page.html&_a=files)

mingwli:
The markdig engine converts the valid html in markdown without a problem.  
The real reason is during the impact.normalize-html, the `"` aren't properly escaped in htmlParser.

PR #3644, opened 2018-11-20
`', "` should be properly escaped html attributes in .md files if written in valid html

mingwli:
#3535


docs/specs/markdown.yml
```diff
@@ -351,3 +351,16 @@ outputs:
     {"document_id":"f481558d-b8fa-5acb-facb-d3bee58fbf3e","document_version_independent_id":"80168efe-9436-03b9-9e7c-eaf20a91d680"}
   build.manifest: |
     {"files":[{"siteUrl":"/docs/a.experimental","outputPath":"docs/a.experimental.json","sourcePath":"docs/a.experimental.md"},{"siteUrl":"/docs/a","outputPath":"docs/a.json","sourcePath":"docs/a.md"}],"dependencies":{}}
+---
+# `", '` should be properly escaped when written in html directly
+inputs:
+  docfx.yml:
+  docs/a.md: |
```
yufeih: can you add a case with `<a href='javascript:window.open("https://shell.azure.com", "_blank", "toolbar=no,scrollbars=yes,resizable=yes,menubar=no,location=no,status=no")'>` and see the result?

mingwli: the `"` is encoded as `&quot;` in TransformLinks

PR #3647, opened 2018-11-20
mock repo to test git

superyyrrzz:
so that yufeih/OsmondJiang/superyyrrzz can retire from `git.yml`

#3586

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -1,7 +1,7 @@
 ---
 # Show git contribution info when userProfile & commitsTime is set to URL
 repos:
-  https://docs.com/testowner/testrepo#master-url-cache:
+  https://github.com/testowner/testrepo#master-url-cache:
     - author: Charlie
       email: charlie@contoso.com
       files:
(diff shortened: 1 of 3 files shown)


src/docfx/lib/github/GithubUtility.cs
```diff
@@ -7,9 +7,10 @@ namespace Microsoft.Docs.Build
 {
     internal static class GitHubUtility
     {
+        // TODO: remove the hard coded "docs.com"
         private static readonly Regex s_gitHubRepoUrlRegex =
            new Regex(
-               @"^((https|http):\/\/github\.com)\/(?<account>[^\/\s]+)\/(?<repository>[A-Za-z0-9_.-]+)((\/)?|(#(?<branch>\S+))?)$",
+               @"^((https|http):\/\/(github|docs)\.com)\/(?<account>[^\/\s]+)\/(?<repository>[A-Za-z0-9_.-]+)((\/)?|(#(?<branch>\S+))?)$",
```
yufeih: if the test depend on `github.com`, why not use `github.com` in test cases?

docs/specs/git.yml
```diff
@@ -1,263 +1,341 @@
...
+      files:
+        docs/index.md: |
+          Alice creates this. Bob updates this.
+    - author: Alice 
+      email: alice@contoso.com
+      files:
+        docs/index.md: |
+          Alice creates this.
+inputs:
+  docfx.yml: |
```
yufeih: can this be put into repos?

superyyrrzz: The only one cannot be put into repos is `gitCommitsTime` file, otherwise, I cannot consolidate a stable commit id in spec. So I put all the non-content outside repo for consistency.

PR #3648, opened 2018-11-20
handle toc conflicts

OsmondJiang:


src/docfx/build/toc/TableOfContentsMapBuilder.cs
```diff
@@ -41,14 +43,31 @@ public void Add(Document tocFile, IEnumerable<Document> referencedDocuments, IEn
         /// Build toc map including all tocs and reversed toc mapping(document -> toc)
         /// </summary>
         /// <returns>The toc map</returns>
-        public TableOfContentsMap Build()
+        public TableOfContentsMap Build(Context context)
```
live1206: how about add `List<Error>` to return type, and report it in `Build.cs`?

src/docfx/build/toc/TableOfContentsMapBuilder.cs
```diff
@@ -41,14 +43,31 @@ public void Add(Document tocFile, IEnumerable<Document> referencedDocuments, IEn
...
         {
             var documentToTocs = new Dictionary<Document, HashSet<Document>>();
 
             // reverse the mapping between toc and documents
             // order by toc path
             var allTocs = new List<Document>();
             var experimentalTocs = new List<Document>();
+
+            // handle conflicts
+            var tocsGroupBySiteUrl = _tocToDocuments.Keys.GroupBy(k => k.SiteUrl);
```
live1206: Extract this logic to a method? As we discussed, this might be shared logic from the build output

src/docfx/build/toc/TableOfContentsMapBuilder.cs
```diff
@@ -41,14 +43,32 @@ public void Add(Document tocFile, IEnumerable<Document> referencedDocuments, IEn
...
+            var conflictedTocs = new HashSet<Document>();
+            foreach (var group in tocsGroupBySiteUrl)
+            {
+                var siteUrl = group.Key;
+                if (group.Count() > 1)
```
superyyrrzz: Does this need consider version?

live1206: @928PJY is working on TOC with versioning, maybe it can be handled in that PR.

Issue #2989, opened 2018-07-03
FSharp tests are failing


herohua:
Unit tests under "Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests" are failing under test explorer.

System.IO.DirectoryNotFoundException : Could not find a part of the path 'C:\current\docfx\test\Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests\bin\Debug\net461\TestData\NetCoreProject\obj\NetCoreProject.fsproj.proj-info.targets'.

It seems like the output folder was not generated during the test run.
And CI build did not run these tests.

superyyrrzz:
It works now on my machine.

PR #3658, opened 2018-11-21
Fix EntityMetadata for FSharp when parsing signature files

FraserWaters-GR:
Needed to add a check for FSharpEntity.IsOpaque as that is true for types that have their implementation hidden by a signature file.


superyyrrzz:
@FraserWaters-GR Could you please update some tests in `Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests` to cover this case?

FraserWaters-GR:
So the FSharp tests are currently broken (#2989), I've had a look at it's not clear how these are supposed to work in the first place.

For now could we be pragmatic and merge this fix, I'll add a watch to #2989 and am happy to come back and add a test for this case then. I'm running the metadata generator across 18 large F# projects and this fix hasn't caused any issues with them so I'd consider this one line fix a low risk.

superyyrrzz:
@FraserWaters-GR Sure. It will be helpful if you can also help on the unit test issue. Thank you for contribution!

FraserWaters-GR:
Thanks for accepting. We might have some more little F# fixes, if we do we'll take another look at getting the tests fixed as well.

superyyrrzz:
@FraserWaters-GR Great! This fix is released in 2.40.4: https://github.com/dotnet/docfx/releases/tag/v2.40.4

PR #3659, opened 2018-11-22
move theme from dependency repo

OsmondJiang:
make the dependency repo more clear and clean, add localization convention for theme config so loc user no need to overwrite this config for every locale

### DIFF
--- a/src/docfx/localization/LocalizationConvention.cs
+++ b/src/docfx/localization/LocalizationConvention.cs
@@ -11,8 +11,6 @@ namespace Microsoft.Docs.Build
 {
     internal static class LocalizationConvention
     {
-        private static readonly Regex s_repoNameWithLocale = new Regex(@"^.+?(\.[a-z]{2,4}-[a-z]{2,4}(-[a-z]{2,4})?|\.localization)?$", RegexOptions.IgnoreCase);
-
         /// <summary>
         /// The loc repo name follows below conventions:
         /// source remote                                           -->     loc remote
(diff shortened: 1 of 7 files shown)


src/docfx/localization/LocalizationConvention.cs
```diff
@@ -50,12 +48,15 @@ public static (string remote, string branch) GetLocalizationRepo(LocalizationMap
 
             var newLocale = mapping == LocalizationMapping.Repository ? $".{locale}" : ".localization";
             var newBranch = bilingual ? $"{branch}-sxs" : branch;
-            var repoName = remote.Split(new char[] { '/', '\\' }).Last();
-            var match = s_repoNameWithLocale.Match(repoName);
-            if (match.Success && match.Groups.Count >= 2 && !string.IsNullOrEmpty(match.Groups[1].Value))
+
+            if (remote.EndsWith($".{defaultLocale}", StringComparison.OrdinalIgnoreCase))
```
yufeih: we may always need a reverse lookup from loc -> English to support build, does this affect that design?

OsmondJiang: not yet, I would like to do that in another PR, this is only for source to loc

OsmondJiang:
@mingwli, maybe you need change your migrator logic

src/docfx/localization/LocalizationConvention.cs
live1206: extract this method to a method? duplicated in `GetLocalizationTheme`. Might be more duplication

PR #3660, opened 2018-11-22
move all localization logic to one place

OsmondJiang:
we want to avoid the localization logic being placed everywhere, need one place to manage these logic


src/docfx/localization/LocalizationConvention.cs
```diff
@@ -109,6 +110,82 @@ public static string GetLocalizationDocsetPath(string docsetPath, Config config,
             return localizationDocsetPath;
         }
 
+        public static bool TryToContributionBranch(string branch, out string contributionBranch)
```
live1206: >TryToContributionBranch [](start = 27, length = 23)

how about `TryGetContributionBranch`?

src/docfx/build/Build.cs
```diff
@@ -106,7 +100,7 @@ async Task BuildOneFile(Document file, Action<Document> buildChild)
                 bool ShouldBuildFile(Document file)
                 {
                     // source content in a localization docset
-                    if (docset.FallbackDocset != null && file.Docset.LocalizationDocset != null)
+                    if (docset.IsLocalized() && !file.Docset.IsLocalized())
```
superyyrrzz: >IsLocalized [](start = 31, length = 11)

What's the difference between FallbackDocset and LocalizationDocset? Why change from check FallbackDocset to check LocalizationDocset here?

OsmondJiang: it's used for different purpose, here is only a flag to see if the current docset is localized or not

Issue #3661, opened 2018-11-22
Fetch dependency repos using --depth 1


yufeih:
Significantly reduces template repo clone speed.
	
Command | Time
------|-----
git clone https://github.com/Microsoft/templates.docs.msft  | 1:16
git clone https://github.com/Microsoft/templates.docs.msft --depth 1 | 0:08

Need a flag to preserve history for loc repos.

PR #3668, opened 2018-11-23
Only clone first depth for themes and dependency repos

yufeih:
#3661 

Added to flags for clone: 
- `DepthOne` indicates it is using `--depth 1`, and it only applies to dependency and theme repos which we don't need git history for sure
- `NoCheck` indicates a dependent git repo does not need to be checked out. Non sxs branch for sxs content fits this case.


herohua:
We may need to resolve git history in dependency repo as well (if that repo contains markdown files), so I guess it is not safe to only clone one-level there (cc @fenxu). For template repo, yes, we may just clone one-level.

yufeih:
Themes and dependencies repos are right now safe to be cloned using `--depth 1`, we only read content from dependencies repos.

PR #3669, opened 2018-11-23
Not output legacy monikers if the list is empty

live1206:


src/docfx/build/legacy/metadata/LegacyMetadata.cs
```diff
@@ -80,7 +86,10 @@ public static JObject GenerateLegacyRedirectionRawMetadata(Docset docset, PageMo
 
             rawMetadata["_op_canonicalUrlPrefix"] = $"{docset.Config.BaseUrl}/{docset.Locale}/{docset.Config.DocumentId.SiteBasePath}/";
 
-            rawMetadata["monikers"] = new JArray(pageModel.Monikers);
+            if (pageModel.Monikers.Count > 0)
```
yufeih: When monikers is empty, is `group` still in the output?

live1206: nope, removed in the previous PR

PR #3676, opened 2018-11-26
Do not serialize to JSON if monikers list is empty

live1206:
- Use JSON .net to not serialize `Monikers` if empty
  https://www.newtonsoft.com/json/help/html/ConditionalProperties.htm
- Change `Monikers` to read-only and initialize it with an empty list
- Add `Monikers` to legacy TOC model

### DIFF
--- a/docs/specs/moniker.yml
+++ b/docs/specs/moniker.yml
@@ -37,8 +37,7 @@ outputs:
     }
   docs/a.json: |
     {
-      "content": "<p>No version</p>",
-      "monikers": []
+      "content": "<p>No version</p>"
     }
 ---
 # Apply moniker range setting in the config and yaml header to markdown file's output path and output content
@@ -75,8 +74,7 @@ inputs:
 outputs:
   docs/a.json: |
     {
-      "content": "<p>No version</p>",
-      "monikers": []
+      "content": "<p>No version</p>"
     }
   218c13d0/docs/a.json: |
     {
@@ -94,8 +92,7 @@ outputs:
           {
               "siteUrl": "/docs/a",
               "outputPath": "docs/a.json",
-              "sourcePath": "docs/a.md",
-              "monikers": []
+              "sourcePath": "docs/a.md"
           },
           {
               "siteUrl": "/docs/a",
@@ -403,13 +400,11 @@ inputs:
 outputs:
   docs/b.json: |
     {
-      "content": "<p>Link to <a href=\"/docs/a?view=netcore-1.1\">netcore-1.1</a></p>",
-      "monikers": []
+      "content": "<p>Link to <a href=\"/docs/a?view=netcore-1.1\">netcore-1.1</a></p>"
     }
   docs/a.json: |
     {
-      "content": "<p>No version</p>",
-      "monikers": []
+      "content": "<p>No version</p>"
     }
   218c13d0/docs/a.json: |
     {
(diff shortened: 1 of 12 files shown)


src/docfx/build/legacy/files/LegacyTableOfContents.cs
```diff
@@ -67,6 +67,9 @@ private sealed class LegacyTableOfContentsMetadata : LegacyTableOfContentsExperi
         {
             [JsonProperty(PropertyName = "pdf_absolute_path")]
             public string PdfAbsolutePath { get; set; }
+
+            [JsonProperty("monikers")]
+            public List<string> Monikers { get; set; }
```
928PJY: initialize to a empty list too?

src/docfx/build/toc/parser/TableOfContentsInputItem.cs
```diff
@@ -43,11 +44,11 @@ public static TableOfContentsItem ToTableOfContentsModel(TableOfContentsInputIte
             }
 
             var children = inputModel.Items?.Select(l => ToTableOfContentsModel(l, comparer));
-            var childrenMonikers = children?.SelectMany(child => child?.Monikers ?? new List<string>());
+            var childrenMonikers = children?.SelectMany(child => child?.Monikers);
 
             var monikers = (childrenMonikers == null ? inputModel.Monikers : childrenMonikers.Union(inputModel.Monikers)).Distinct(comparer).ToList();
```
928PJY: >(childrenMonikers == null ? inputModel.Monikers : childrenMonikers.Union(inputModel.Monikers)).Distinct(comparer).ToList() [](start = 27, length = 122)

`var monikers = inputModel.Monikers.Union(childrenMonikers).Distinct(comparer).ToList()`

docs/specs/moniker.yml
```diff
@@ -37,8 +37,7 @@ outputs:
     }
   docs/a.json: |
     {
-      "content": "<p>No version</p>",
-      "monikers": []
+      "content": "<p>No version</p>"
```
928PJY: >"content": "<p>No version</p>" [](start = 6, length = 30)

build result 
```
"content": "<p>No version</p>",
"monikers": []
```
can also pass,
should we use `"!monikers": null`

src/docfx/build/legacy/files/LegacyTableOfContents.cs
live1206: Since this is legacy, we always just set the value

src/docfx/build/Build.cs
```diff
@@ -196,9 +196,9 @@ void ValidateBookmarks()
                 {
                     SourcePath = file.FilePath,
                     SiteUrl = file.SiteUrl,
-                    Monikers = monikers,
                     OutputPath = GetOutputPath(file, monikers),
                 };
+                manifest.Monikers.AddRange(monikers);
```
OsmondJiang: > manifest.Monikers.AddRange(monikers); [](start = 15, length = 38)

why doing that?

live1206: Because I want to initialize a list to `Monikers`, then we do not need to check null anywhere

src/docfx/build/toc/parser/TableOfContentsInputItem.cs
```diff
@@ -57,8 +58,9 @@ public static TableOfContentsItem ToTableOfContentsModel(TableOfContentsInputIte
                 Expanded = inputModel.Expanded,
                 ExtensionData = inputModel.ExtensionData,
                 Children = children?.ToList(),
-                Monikers = monikers,
             };
+            tocItem.Monikers.AddRange(monikers);
```
superyyrrzz: >tocItem.Monikers.AddRange(monikers); [](start = 12, length = 36)

v3 seems to prefer assignment instead of modify a property in place. Seems we an keep `set` for it.

live1206: ok

Issue #3560, opened 2018-11-01
Fallback non-existing resource for loc docset build


OsmondJiang:
- [x] fallback to no-existing token/codesnippet
- [x] fallback to no-existing included toc
- ~[ ] fallback to no-existing image~

OsmondJiang:
in v2, we don't fallback to no-existing images since DHS will not delete any resource, i think we can keep that in v3. so no work for non-existing image.

PR #3677, opened 2018-11-26
try to resolve deleted inclusion(token) from git history

OsmondJiang:
todo: support code snippet
no need to do too much optimization since this case happened really rare


OsmondJiang:
#3560

OsmondJiang:
https://github.com/dotnet/docfx/blob/v3/docs/designs/conceptual-localization.md#lookup-no-existing-source-resourcestokencodesnippetimage

src/docfx/docset/Document.cs
```diff
@@ -6,11 +6,14 @@
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text;
 
 namespace Microsoft.Docs.Build
 {
     internal class Document
     {
+        private readonly string _content;
```
yufeih: Don't change `Document`. Just change `Resolve.TryResolveContent` it will cover both token and code snippet.

src/docfx/lib/git/GitUtility.cs
```diff
@@ -213,6 +213,47 @@ public static void CheckMergeConflictMarker(string content, string file)
...
+
+            var obj = git_tree_entry_id(entry);
+            if (git_blob_lookup(out var blob, repo, obj) != 0)
+            {
+                git_tree_entry_free(entry);
+                git_repository_free(repo);
+                return false;
+            }
+
+            content = git_blob_rawcontent(blob);
```
yufeih: This line will fail if the blob uses a different encoding or memory layout than `System.String`.

Read the string from the memory chunk defined by `git_blob_rawcontent` and `git_blob_rawsize`. `git_blob_rawcontent` should return `IntPtr` instead of `string`.

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -146,8 +150,59 @@ public static bool TryResolveDocset(this Docset docset, string file, out Docset
...
+                });
+
+                if (gitCommitProvider != null)
+                {
+                    var pathToRepo = PathUtility.NormalizeFile(Path.GetRelativePath(repoPath, Path.Combine(fallbackDocset.DocsetPath, file)));
+                    var commits = gitCommitProvider.GetCommitHistory(pathToRepo);
+                    if (commits.Count > 1)
+                    {
+                        // the latest commit would be deleting it from repo
+                        if (GitUtility.TryGetContentFromHistory(repoPath, pathToRepo, commits[1].Sha, out content))
```
yufeih: is this `commits[0]`?

src/docfx/lib/git/GitUtility.cs
```diff
@@ -213,6 +213,47 @@ public static void CheckMergeConflictMarker(string content, string file)
...
 
+        public static unsafe bool TryGetContentFromHistory(string repoPath, string filePath, string committish, out string content)
+        {
+            content = null;
+            if (git_repository_open(out var repo, repoPath) != 0)
+            {
+                return false;
+            }
+
+            if (git_revparse_single(out var commit, repo, committish) != 0)
```
yufeih: we only need to support commit hash for content retrieval, `revparse` might be slower/heavier then a simple lookup by git object id.

src/docfx/localization/LocalizationConvention.cs
OsmondJiang: no, it's [1] since [0] is deleting the file

---
In reply to: [236457954](https://github.com/dotnet/docfx/pull/3677#discussion_r236457954) [](ancestors = 236457954)

src/docfx/lib/git/GitUtility.cs
OsmondJiang: i want to make it consistent with others and truly, we don't need do too much optimization for this scenario

---
In reply to: [236458528](https://github.com/dotnet/docfx/pull/3677#discussion_r236458528) [](ancestors = 236458528)

OsmondJiang:
@yufeih , i don't think this PR has any relationship with your PR 'dependency clean up', you just hold on my PR until your pr is merged?

src/docfx/docset/Document.cs
OsmondJiang: The document has ReadContent and ReadText and we can not make sure the they are only used in TryResolveContent

---
In reply to: [236350421](https://github.com/dotnet/docfx/pull/3677#discussion_r236350421) [](ancestors = 236350421)

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -2,16 +2,18 @@
...
 using System.IO;
-using System.Linq;
-using System.Text.RegularExpressions;
+using System.Threading.Tasks;
 
 namespace Microsoft.Docs.Build
 {
     internal static class LocalizationConvention
     {
+        private static readonly ConcurrentDictionary<string, (GitCommitProvider gitProvider, string repoPath)> s_gitCommitProvider = new ConcurrentDictionary<string, (GitCommitProvider gitProvider, string repoPath)>();
```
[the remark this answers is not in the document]
OsmondJiang: I don't think all GitCommitProvider need a cache file, in your previous code, it's optional. And in that case we only need a few files' commits, so no need load all cache and save all cache

---
In reply to: [236351742](https://github.com/dotnet/docfx/pull/3677#discussion_r236351742) [](ancestors = 236351742)

src/docfx/lib/git/GitUtility.cs
OsmondJiang: ok, thanks for this, try to change it....

---
In reply to: [236448956](https://github.com/dotnet/docfx/pull/3677#discussion_r236448956) [](ancestors = 236448956)

src/docfx/lib/git/GitUtility.cs
```diff
@@ -213,6 +213,47 @@ public static void CheckMergeConflictMarker(string content, string file)
...
+
+            var obj = git_tree_entry_id(entry);
+            if (git_blob_lookup(out var blob, repo, obj) != 0)
+            {
+                git_tree_entry_free(entry);
+                git_repository_free(repo);
+                return false;
+            }
+
+            content = Marshal.PtrToStringUTF8(git_blob_rawcontent(blob));
```
yufeih: content may not be UTF8 encoded

OsmondJiang: what's your suggestion, i refers LibGit2Sharp, they also by default using UTF8?

---
In reply to: [236529712](https://github.com/dotnet/docfx/pull/3677#discussion_r236529712) [](ancestors = 236529712)

src/docfx/docset/Repository.cs
```diff
@@ -41,5 +44,26 @@ public static Repository CreateFromFolder(string path)
...
+        public static Repository GetRepository(string fullPath)
+        {
+            if (GitUtility.IsRepo(fullPath))
+                return CreateFromFolder(fullPath);
+
+            var parent = System.IO.Path.GetDirectoryName(fullPath);
+            return !string.IsNullOrEmpty(parent)
+                ? _repositoryByFolder.GetOrAdd(parent, GetRepository)
+                : null;
+        }
```
superyyrrzz: Move these static methods to `RepositoryUtility`? No need to include global _repositoryByFolder in a specific repository instance.

OsmondJiang: great idea

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -187,56 +181,23 @@ public List<GitCommit> GetCommitHistory(string file, string committish = null)
...
-                                }
-                            }
-                        }
-                    }
-                }
-            });
         }
 
-        public void Dispose()
+        public Task Update()
```
live1206: Can we still call this `SaveCache`? We just moved the implementation to `GitCommitCacheProvider.SaveCache`

OsmondJiang: ok

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -40,30 +38,34 @@ internal sealed class GitCommitProvider : IDisposable
         // Only the last N = MaxCommitCacheCountPerFile commit histories are cached for a file, they are selected by least recently used order (lruOrder).
         private readonly ConcurrentDictionary<string, Dictionary<(long commit, long blob), (long[] commitHistory, int lruOrder)>> _commitCache;
 
+        private static readonly ConcurrentDictionary<string, GitCommitProvider> s_gitCommitProvider = new ConcurrentDictionary<string, GitCommitProvider>();
```
yufeih: same as `RepositoryUtility`, static cache may contain stall data, use a `RepositoryProvider` instance whose lifetime is bound to a single build run.

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -122,6 +120,53 @@ public static bool TryGetContributionBranch(string branch, out string contributi
             return false;
         }
 
+        public static (Error error, string content, Document file) TryResolveFromHistory(Docset docset, string pathToDocset)
+        {
+            // try to resolve from source repo's git history
+            // todo: support code snippet
```
yufeih: code snippet should be supported out of the box because it is also calling ResolveContent, we just need a test.

src/docfx/lib/git/GitUtility.cs
```diff
@@ -213,6 +214,52 @@ public static void CheckMergeConflictMarker(string content, string file)
...
+
+            var obj = git_tree_entry_id(entry);
+            if (git_blob_lookup(out var blob, repo, obj) != 0)
+            {
+                git_tree_entry_free(entry);
+                git_repository_free(repo);
+                return false;
+            }
+
+            using (var stream = new UnmanagedMemoryStream((byte*)git_blob_rawcontent(blob).ToPointer(), git_blob_rawsize(blob)))
```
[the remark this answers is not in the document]
OsmondJiang: i am referening LibGit2Sharp

---
In reply to: [237185651](https://github.com/dotnet/docfx/pull/3677#discussion_r237185651) [](ancestors = 237185651)

src/docfx/lib/git/GitCommitProvider.cs
OsmondJiang: ok

---
In reply to: [237187074](https://github.com/dotnet/docfx/pull/3677#discussion_r237187074) [](ancestors = 237187074)

src/docfx/localization/LocalizationConvention.cs
OsmondJiang: it's related to the content type, i will do it in another pr

---
In reply to: [237189150](https://github.com/dotnet/docfx/pull/3677#discussion_r237189150) [](ancestors = 237189150)

src/docfx/lib/git/GitCommitProvider.cs
OsmondJiang: so every watch build will reload the git commits again, interesting

yufeih: It right now is. There are a lot of state that could be shared across watch builds, I haven't started on that because we didn't have watch yet.

src/docfx/lib/git/RepositoryProvider.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    internal class RepositoryProvider
+    {
+        private readonly ConcurrentDictionary<string, Repository> _repositoryByFolder = new ConcurrentDictionary<string, Repository>();
+
+        private readonly ConcurrentDictionary<string, GitCommitProvider> _gitCommitProvider = new ConcurrentDictionary<string, GitCommitProvider>();
+
+        public GitCommitProvider GetGitCommitProvider(string repoPath, string repoRemote = null, ConcurrentDictionary<string, Dictionary<(long commit, long blob), (long[] commitHistory, int lruOrder)>> commitCache = null)
```
yufeih: if I first call this method without a commit cache and then with a commit cache using the same repo, the second call will give back the an instance without cache, that is incorrect. Same is true if `repoRemote` is passed differently for the same repo.

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -122,6 +120,52 @@ public static bool TryGetContributionBranch(string branch, out string contributi
...
+            var fallbackDocset = GetFallbackDocset();
+            if (fallbackDocset != null && Document.GetContentType(pathToDocset) == ContentType.Page)
```
yufeih: To be consistent, this feature is about fallback to git history for all _content inclusions_, including TOC content inclusion, but right now TOC loader does not respect `content` returned from `ResolveContent`, it will trigger the `Debug.Assert(!IsFromHistory);` above: https://github.com/dotnet/docfx/blob/v3/src/docfx/build/toc/parser/TableOfContentsParser.cs#L198

Strictly speaking, `TocHrefType.RelativeFolder` is not a content inclusion.

src/docfx/lib/git/RepositoryProvider.cs
yufeih: We could share a global GitCommitProvider (always with cache) and expose it like:

```csharp
class GitCommitProvider
{
    private readonly Lazy<Task<Dictionary>> _commitCache;

    public GitCommitProvider(string repoPath);

    // This one uses cache and returns full history
    public Task<List<GitCommit>> GetCommitHistory(string file, string committish = null);

    // This one does not use cache and return only the TOP N history
    public List<GitCommit> GetCommitHistory(string file, int n, string committish = null);
}
```

src/docfx/lib/git/GitCommitProvider.cs
OsmondJiang: so, why did you comment on that if we haven't got the watch design yet?

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -122,6 +120,52 @@ public static bool TryGetContributionBranch(string branch, out string contributi
...
+            }
+
+            return default;
+
+            Docset GetFallbackDocset()
+            {
+                if (docset.LocalizationDocset != null)
+                {
+                    // source docset in loc build
+                    return docset;
```
yufeih: so English also fallback to git history if missing? It shouldn't, add a test for that.

OsmondJiang:
I am closing this PR since i have disagreement with @yufeih's comment related to the `watch` feature is a not implemented an designed yet. cc @herohua

PR #3680, opened 2018-11-26
Make dependency map global

yufeih:
This is one of a series of refactor on dependency resolve logic. Ultimately we only need one global `DependencyResolver` to handle the tasks that is currently splited in `PageCallback`, `Resolve`, etc.

It makes `dependencyMapBuilder` global, just like `manifestBuilder`, to allow `PageCallback` to be global.

#3677


src/docfx/build/Build.cs
```diff
@@ -91,40 +91,20 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
...
-                        contribution,
-                        map,
-                        null,
-                        manifestBuilder);
-
-                    foreach (var (source, dependencies) in dependencyMap)
-                    {
-                        sourceDependencies.TryAdd(source, dependencies);
-                    }
+                    var callback = new PageCallback(xrefMap, dependencyMapBuilder, bookmarkValidator, null);
```
OsmondJiang: >var callback = new PageCallback(xrefMap, dependencyMapBuilder, bookmarkValidator, null); [](start = 20, length = 88)

when build TOC, we don't need that callback, it's confused me that what should i put there, why not just leave callback to buildPage and just pass the dependencyBuilder to toc

live1206: Maybe we should expand `callback`, here to build TOC, we only need to send in the global `dependencyMapBuilder`

yufeih: This is the first iteration of the refactor to make changes small. `DependencyMapBuild`, `PageCallback`, `Resolve` will be consolidated to a single `DependencyResolver` that is responsible for resolve content, url, xref dependencies. TOC also need to resolve xref so it depend on `DependencyResolver` hence I am putting it here.

Issue #3683, opened 2018-11-27
Moniker E2E test failing randomly


live1206:
2018-11-26T09:09:17.3305550Z Failed   Microsoft.Docs.Build.E2ETest.Run(name: "moniker/03. Redirection file with monikers")
2018-11-26T09:09:17.3327670Z Error Message:
2018-11-26T09:09:17.3334420Z  System.InvalidOperationException : Collection was modified; enumeration operation may not execute.
2018-11-26T09:09:17.3338620Z Stack Trace:
2018-11-26T09:09:17.3339510Z    at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
2018-11-26T09:09:17.3340250Z    at System.String.JoinCore[T](Char* separator, Int32 separatorLength, IEnumerable`1 values)
2018-11-26T09:09:17.3386870Z    at Microsoft.Docs.Build.Build.GetOutputPath(Document file, List`1 monikers) in /Users/vsts/agent/2.142.1/work/1/s/src/docfx/build/Build.cs:line 245
2018-11-26T09:09:17.3387210Z    at Microsoft.Docs.Build.Build.BuildFile(Context context, Document file, TableOfContentsMap tocMap, ContributionProvider contribution, Dictionary`2 monikersMap, PageCallback callback, ManifestBuilder manifestBuilder) in /Users/vsts/agent/2.142.1/work/1/s/src/docfx/build/Build.cs:line 195
2018-11-26T09:09:17.3387710Z    at Microsoft.Docs.Build.Build.<>c__DisplayClass1_1.<<BuildFiles>g__BuildOneFile|7>d.MoveNext() in /Users/vsts/agent/2.142.1/work/1/s/src/docfx/build/Build.cs:line 103
2018-11-26T09:09:17.3389150Z --- End of stack trace from previous location where exception was thrown ---
2018-11-26T09:09:17.3389440Z    at Microsoft.Docs.Build.Build.<>c__DisplayClass1_1.<<BuildFiles>b__1>d.MoveNext() in /Users/vsts/agent/2.142.1/work/1/s/src/docfx/build/Build.cs:line 80
2018-11-26T09:09:17.3391240Z --- End of stack trace from previous location where exception was thrown ---
[...]

PR #3684, opened 2018-11-27
return a new list in moniker range parser

928PJY:
#3683 
 
return a new list in moniker range parser to avoid the cache be changed outside the parser.

### DIFF
--- a/src/docfx/lib/moniker/MonikerRangeParser.cs
+++ b/src/docfx/lib/moniker/MonikerRangeParser.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Docs.Build
 {
     internal class MonikerRangeParser
     {
-        private readonly ConcurrentDictionary<string, Lazy<List<string>>> _cache = new ConcurrentDictionary<string, Lazy<List<string>>>();
+        private readonly ConcurrentDictionary<string, Lazy<IEnumerable<string>>> _cache = new ConcurrentDictionary<string, Lazy<IEnumerable<string>>>();
         private readonly EvaluatorWithMonikersVisitor _monikersEvaluator;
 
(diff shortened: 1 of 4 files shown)


src/docfx/lib/moniker/MonikerRangeParser.cs
```diff
@@ -19,23 +19,28 @@ public MonikerRangeParser(MonikerDefinitionModel monikerDefinition)
         }
 
         public List<string> Parse(string rangeString)
```
superyyrrzz: return `IEnumerable<string>`?

src/docfx/lib/moniker/MonikerRangeParser.cs
```diff
@@ -10,18 +10,18 @@ namespace Microsoft.Docs.Build
 {
     internal class MonikerRangeParser
     {
-        private readonly ConcurrentDictionary<string, Lazy<List<string>>> _cache = new ConcurrentDictionary<string, Lazy<List<string>>>();
+        private readonly ConcurrentDictionary<string, Lazy<IEnumerable<string>>> _cache = new ConcurrentDictionary<string, Lazy<IEnumerable<string>>>();
```
superyyrrzz: This can remain List as a private field

928PJY: yes, but it don't have to be, right?

---
In reply to: [236547338](https://github.com/dotnet/docfx/pull/3684#discussion_r236547338) [](ancestors = 236547338)

live1206: +1, it is enough to keep `IEnumerable<string> Parse(string rangeString)`

OsmondJiang: IReadOnlyList would be better?

PR #3689, opened 2018-11-28
parse token to running branch sync command

928PJY:


tools/Deployment/gulpfile.js
```diff
@@ -10,6 +10,8 @@ let del = require("del");
 let glob = require("glob");
 let gulp = require("gulp");
 let nconf = require("nconf");
+let format = require("string-format");
+format.extend(String.prototype, {})
```
superyyrrzz: Use this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

tools/Deployment/gulpfile.js
```diff
@@ -203,18 +205,26 @@ gulp.task("publish:chocolatey", () => {
...
 
     if (Common.isThirdWeekInSprint()) {
         console.log("Ignore to sync in the third week of a sprint");
         process.exit(2);
     }
 
+    var repoInfo = {
+        "account": config.docfx.account,
+        "token": process.env.TOKEN
+    }
```
superyyrrzz: append `;`

tools/Deployment/gulpfile.js
928PJY: `Template_literals` requires to prepare the variables first, but I want to set this in the config.

Issue #3688, opened 2018-11-28
Support branch level localization repository


yufeih:
Lets starts with two conventions to lookup localized content: __locale per repo__ and __locale per branch__. The config is a simple enum stored in source (English) repo.

There might be need to group locales in multiple repos using a `locale` --> `repo+branch` map in the future, but we don't do it for now.

OsmondJiang:
@yufeih , here are some questions:
 - is `locale per branch`  in the same repo with source or another localization repo?
- if the loc content are in another localization repo, does it follow our conventions?
- what's the branch conventions, like the document using {locale}-{branch} like `zh-cn-live` or {locale}/{branch} like `zh-cn/live`?

yufeih:
naming convention is:

__Small repo, all langs in one repo__
repo: `{source-repo}.loc`, branch `{branch}.{locale}`

__Large repo, one lang per repo__
repo: `{source-repo}.{locale}`, branch `{branch}` 
> loc team could migrate existing repos that do not follow this convention

OsmondJiang:
based on @herohua  and @yufeih 's information, this issue has lower priority, right now we will focus on implementing the repo-mapping localization which is backward compatible with docfx/ops v2

herohua:
We'd like to revisit whether we should keep the code in code base that partially supports this scenario.

herohua:
We don't need this issue as we are not going to support branch level localization contents.

PR #3690, opened 2018-11-28
support branch mapping localization build

OsmondJiang:
all changes follows the doc's updates


docs/specs/localization.yml
```diff
@@ -150,10 +187,31 @@ repos:
         zh-cn/docs/a.md:
 outputs:
   docs/a.json: |
-    { "content_git_url": "*dotnet/docfx.localization/blob/master/zh-cn*"}
+    { "content_git_url": "*dotnet/docfx.localization/blob/master/zh-cn/*"}
+---
+# edit url convention with source locale 2
```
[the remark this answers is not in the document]
OsmondJiang: it makes the test case name really only, which really hard to find which case failed viatest explorer

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -63,6 +61,23 @@ public static (string remote, string branch) GetLocalizationRepo(LocalizationMap
             return ($"{remote}{newLocale}", newBranch);
         }
 
+        public static string GetLocalizationBranch(LocalizationMapping mapping, string sourceBranch, string locale)
```
mingwli: > public static string [](start = 7, length = 21)

this can be private or local?

src/docfx/localization/LocalizationConvention.cs
```diff
@@ -22,7 +20,7 @@ internal static class LocalizationConvention
         /// <returns>The loc remote url</returns>
```
mingwli: >/// <returns>The loc remote url</returns> [](start = 8, length = 41)

need some update?

OsmondJiang:
#3688

PR #3695, opened 2018-11-29
Move moniker related functionality to MonikerProvider

OsmondJiang:
move moniker range parer and compare to moniker provider

### DIFF
--- a/src/docfx/build/moniker/MonikersProvider.cs
+++ b/src/docfx/build/moniker/MonikersProvider.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 
 namespace Microsoft.Docs.Build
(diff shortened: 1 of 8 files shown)


src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -3,23 +3,44 @@
...
             _rules.Reverse();
+
+            var monikerDefinition = new MonikerDefinitionModel();
+            if (!string.IsNullOrEmpty(config.MonikerDefinition))
+            {
+                var path = docset.GetFileRestorePath(config.MonikerDefinition);
+                monikerDefinition = JsonUtility.Deserialize<MonikerDefinitionModel>(File.ReadAllText(path));
+            }
+            _rangeParser = new MonikerRangeParser(monikerDefinition);
+            _monikerAscendingComparer = new Lazy<MonikerComparer>(() => new MonikerComparer(monikerDefinition));
```
live1206: `Lazy<>` is not needed here, right? just initiate a new comparer

live1206: Also, we have `Lazy<MonikersProvider>` in the outer layer

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -3,23 +3,44 @@
...
             foreach (var (key, monikerRange) in config.MonikerRange)
             {
                 _rules.Add((GlobUtility.CreateGlobMatcher(key), monikerRange));
             }
             _rules.Reverse();
+
+            var monikerDefinition = new MonikerDefinitionModel();
+            if (!string.IsNullOrEmpty(config.MonikerDefinition))
+            {
+                var path = docset.GetFileRestorePath(config.MonikerDefinition);
```
live1206: It should be enough to just input path here, avoid exposing the entire Docset

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -3,23 +3,44 @@
...
     {
         private readonly List<(Func<string, bool> glob, string monikerRange)> _rules = new List<(Func<string, bool>, string)>();
+        private readonly MonikerRangeParser _rangeParser;
+        private readonly Lazy<MonikerComparer> _monikerAscendingComparer;
+        private readonly Lazy<MonikerComparer> _monikerDescendingComparer;
 
-        public MonikersProvider(Config config)
+        public MonikerComparer AscendingComparer => _monikerAscendingComparer.Value;
+
+        public MonikerComparer DescendingComparer => _monikerDescendingComparer.Value;
```
yufeih: a single `Comparer` is enough. There is no `StringDescendingComparer` for example.  The only use of `DescendingComparer` is in `XrefMap` which should be using `OrderByDescending` over with a comparer.

src/docfx/build/moniker/MonikersProvider.cs
yufeih: Taking a `Docset` as input is fine, `config` can be accessed from `docset` so you just need a `docset` as the input.

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -3,23 +3,44 @@
...
+            {
+                var path = docset.GetFileRestorePath(config.MonikerDefinition);
+                monikerDefinition = JsonUtility.Deserialize<MonikerDefinitionModel>(File.ReadAllText(path));
+            }
+            _rangeParser = new MonikerRangeParser(monikerDefinition);
+            _monikerAscendingComparer = new Lazy<MonikerComparer>(() => new MonikerComparer(monikerDefinition));
+            _monikerDescendingComparer = new Lazy<MonikerComparer>(() => new MonikerComparer(monikerDefinition, false));
         }
 
+        public IReadOnlyList<string> GetRangeMonikers(string rangeString)
```
yufeih: use `_monikerRangeParser.Parse()` inline is clear enough

src/docfx/build/moniker/MonikersProvider.cs
OsmondJiang: ok


---
In reply to: [237374624](https://github.com/dotnet/docfx/pull/3695#discussion_r237374624) [](ancestors = 237374624)

PR #3697, opened 2018-11-29
not report dependency if path is invalid

v-pegao:


src/Microsoft.DocAsCode.Dfm/CodeRenderer/DfmCodeRenderer.cs
```diff
@@ -39,27 +40,38 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmFencesToken to
 
             try
             {
-                // Always report original dependency
-                context.ReportDependency(token.Path);
```
superyyrrzz: What about keep reportDependency here, but wrap this with `isValidPath`?

src/Microsoft.DocAsCode.Dfm/CodeRenderer/DfmCodeRenderer.cs
```diff
@@ -39,27 +40,38 @@ public virtual StringBuffer Render(IMarkdownRenderer renderer, DfmFencesToken to
...
             {
+                if (isValidPath(token.Path))
+                {
+                    context.ReportDependency(token.Path);
+                }
                 return RenderReferenceNotFoundErrorMessage(renderer, token);
             }
         }
 
+        public bool isValidPath(string path) => !_invalidChars.Any(c => path.Contains(c));
```
superyyrrzz: Can we reuse `PathUtility.IsValidFilePath`?

src/Microsoft.DocAsCode.Dfm/CodeRenderer/DfmCodeRenderer.cs
superyyrrzz: // Always report original dependency **when path is valid**

src/Microsoft.DocAsCode.Dfm/CodeRenderer/DfmCodeRenderer.cs
v-pegao: As I tested, it's working well

superyyrrzz:
cc @TianqiZhang

PR #3702, opened 2018-11-30
apply moniker info to resource file

928PJY:
#3607 
support setting moniker range for resource files in config


yufeih:
Even if it dynamic scope, can we simply apply `MonikerRange` and `Routes` config to determine resource monikers?

docs/specs/moniker.yml
```diff
@@ -831,3 +831,110 @@ outputs:
...
+  build.manifest: |
+    {
+      "files": [
+          {"siteUrl": "/docs/a"},
+          {"siteUrl": "/docs/b"}
+      ],
+      "dependencies": { }
+    }
+  build.log: |
+    ["error","publish-url-conflict","Two or more documents of the same version('netcore-1.1') publish to the same url '/docs/c.png': 'docs/v1/c.png', 'docs/v2/c.png'"]
```
superyyrrzz: Add a test case? If v1.md and v2.md reference to the **same** resource file, resource monikers will be their union(v1.monikers, v2.monikers)

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -113,7 +114,7 @@ private static void BuildTocMap(Context context, Document fileToBuild, TableOfCo
                     {
                         referencedDocuments.Add(buildItem);
                         dependencyMapBuilder?.AddDependencyItem(file, buildItem, HrefUtility.FragmentToDependencyType(fragment));
-                        if (monikersMap == null || !monikersMap.TryGetValue(buildItem, out itemMonikers))
+                        if (!monikersMap.TryGetValue(buildItem, out itemMonikers))
```
live1206: `monikersMap` can still be null here

src/docfx/build/Build.cs
```diff
@@ -76,14 +76,21 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
 
                 await ParallelUtility.ForEach(
                     docset.BuildScope.Where(doc => doc.ContentType != ContentType.TableOfContents),
-                    async (file, buildChild) => { monikersMap.TryAdd(file, await BuildOneFile(file, buildChild)); },
-                    ShouldBuildFile,
+                    async (file, buildChild) => { monikersMap.TryAdd(file, await BuildOneFile(file, buildChild, null, null)); },
+                    (file) => { return file.ContentType != ContentType.Resource && ShouldBuildFile(file); },
```
live1206: >file.ContentType != ContentType.Resource [](start = 39, length = 40)

feels better to filter on the build scope above

src/docfx/build/resource/ResourceDependencyMap.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Microsoft.Docs.Build
+{
+    internal class ResourceDependencyMap : ReadOnlyDictionary<Document, List<Document>>
+    {
+        public static readonly ResourceDependencyMap Empty = new ResourceDependencyMap(new Dictionary<Document, List<Document>>());
```
live1206: never used?

docs/specs/moniker.yml
928PJY: the above test?

928PJY:
then we cannot reference to resource file in another group?

---
In reply to: [443069180](https://github.com/dotnet/docfx/pull/3702#issuecomment-443069180) [](ancestors = 443069180)

docs/specs/moniker.yml
928PJY: the above one?

---
In reply to: [237731090](https://github.com/dotnet/docfx/pull/3702#discussion_r237731090) [](ancestors = 237731090)

live1206:
> using System.Diagnostics;

can be removed

---
Refers to: src/docfx/build/Build.cs:7 in f4f599b. [](commit_id = f4f599b97083b558e83dae27e60f2aebf0003ca1, deletion_comment = False)

src/docfx/build/Build.cs
```diff
@@ -148,7 +154,8 @@ void ValidateBookmarks()
                 switch (file.ContentType)
                 {
                     case ContentType.Resource:
-                        model = BuildResource(file);
+                        model = BuildResource.Build(file, monikersMap, referencingFiles);
```
live1206: >Build [](start = 46, length = 5)

how about adding `monikers` to the return as `BuildPage`, then we can avoid casting `model`. Same for `BuildRedirection`

src/docfx/build/Build.cs
928PJY: No, the build scope will not contains resource file, this is for filtering the `buildChild` action

---
In reply to: [237732440](https://github.com/dotnet/docfx/pull/3702#discussion_r237732440) [](ancestors = 237732440)

docs/specs/moniker.yml
```diff
@@ -831,3 +831,110 @@ outputs:
...
+    ---
+    monikerRange: netcore-1.1 || netcore-1.2
+    ---
+    Moniker: netcore-1.1, netcore-1.2
+    ![image](c.png)
+  docs/a.md: |
+    No version
+    ![image](c.png)
+  docs/v1/c.png:
+  docs/c.png:
```
yufeih: should we support this? what's the behavior if there are 2 articles with the same url, one with version, the other does not?
```
docs/v1/a.md
docs/a.md
```

src/docfx/build/Build.cs
```diff
@@ -75,16 +74,29 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
...
+                    async (file, buildChild) => { monikersMap.TryAdd(file, await BuildOneFile(file, buildChild, null, null)); },
+                    (file) => { return ShouldBuildFile(file, new List<ContentType> { ContentType.Page, ContentType.Redirection }); },
+                    Progress.Update);
+
+                // Build resources: since the monikers of resource file is the collection of the monikers of the files referencing to this resource
+                var resourcesFileBuildScope = dependencyMapBuilder.GetResourceDependencyMap();
+                foreach (var file in docset.BuildScope.Where(doc => doc.ContentType == ContentType.Resource))
+                {
+                    resourcesFileBuildScope.TryAdd(file, new List<Document>());
+                }
```
OsmondJiang: the resource are not defined in our build scope?

src/docfx/build/Build.cs
```diff
@@ -94,27 +93,24 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
...
                 }
 
-                async Task BuildTocFile(Document file, Dictionary<Document, List<string>> map)
-                {
-                    var callback = new PageCallback(xrefMap, dependencyMapBuilder, bookmarkValidator, null);
-                    await BuildFile(context, file, tocMap, contribution, map, callback, manifestBuilder);
-                }
-
-                bool ShouldBuildFile(Document file)
+                bool ShouldBuildFile(Document file, IEnumerable<ContentType> shouldBuildContentTypes)
```
yufeih: nit: `params ContentType[]`

src/docfx/build/resource/BuildResource.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            Debug.Assert(file.ContentType == ContentType.Resource);
+
+            var errors = new List<Error>();
+            var (metaErrors, metadata) = JsonUtility.ToObjectWithSchemaValidation<FileMetadata>(file.Docset.Metadata.GetMetadata(file));
+            errors.AddRange(metaErrors);
+
+            var (monikerError, monikers) = file.Docset.Monikers.GetFileLevelMonikers(file, metadata.MonikerRange);
+            errors.AddIfNotNull(monikerError);
+
```
OsmondJiang: almost the same with build redirection?

src/docfx/build/Build.cs
```diff
@@ -226,11 +222,35 @@ private static string GetOutputPath(Document file, List<string> monikers)
             return outputPath;
         }
 
-        private static ResourceModel BuildResource(Document file)
+        private static (List<Error> errors, object model, List<string> monikers) BuildAssert(Document file, ContentType contentType)
```
live1206: how about changing this method into a generic one:
```csharp
private static (List<Error> errors, T model, List<string> monikers) BuildAssert<T>(Document file, ContentType contentType, Func<Document, List<string>, T> func)
```

yufeih: Can we try not to merge `Redirection` with `Resource` ? Yes they currently looks almost the same, but they could be very different like `Resource` needs to be copied to output etc.

live1206: The upper proposal is just to avoid the switch case based on type, which is the generic method invented for.
I am not sure how they will look later, although like a copying `Resource` to output can be handled outside this `BuildAsset` method

yufeih: The concept `Asset` (not Assert) here is misleading, is seems to be _Resource_ + _Redirection_, which used to refer to `Resource`. It is a new term (only used here) for people to grasp.

I really want to see methods like `BuildXXX` where `XXX` is the value defined in `ContentType` for readability. Inside these methods, there are numerous ways to share code, but please keep the build structure as is.

live1206: that's a valid point, keep them separated then

Issue #3492, opened 2018-10-12
Dependency map with detailed classification


superyyrrzz:
v2 has several dependency types in generated dependency map, see: https://dotnet.github.io/docfx/tutorial/incrementalbuild/advanced_report_dependency.html#reserved-dependency-types Some consumer of the dependency map need to know detailed dependency type. e.g. For `include`, we cannot tell whether it's `file inclusion` or `code snippet` for now.

v3 should consider a more detailed dependency type.

yufeih:
Consider supporting dependency map in v3 non-legacy

OsmondJiang:
Where will the non-legacy dependency map be used to? @superyyrrzz ?

superyyrrzz:
@OsmondJiang  A requirement from triple crown loc need to hand off/hand back files with dependencies at the same time. They don't want partially translated token or path/module.

PR preview should also want the dependency map, while it doesn't care about the classification.

yufeih:
For PR comments I'd like to have a dedicated contract other than dependency map. Something similar to publish reports. Like you mentioned, it does not care about classification.

For triple crown loc, It is still too early for v3 to have a design on that. But it should also be something that is specifically for loc, other than just giving them dependency map.

PR #3707, opened 2018-11-30
TocInclusion --> inclusion, uidInclusion --> uid

yufeih:
Turn TOC inclusion into plain inclusion. This cleans up dependency type to be consistent with resolve, so we could later introduce a `DependencyResolver` that handles dependency map internally. We only need 3 types of dependencies to resolve: `ResolveContent`, `ResolveLink` and `ResolveXref`. Bookmark is a special type of link.


superyyrrzz:
Does this coflict with #3492?

yufeih:
Close in favor of removing dependency map all together.

Issue #3703, opened 2018-11-30
Uid not found because it got encoded


live1206:
```
{"code":"uid-not-found","file":"aspnetcore/fundamentals/logging/index.md","line":0,"
message":"Cannot find uid 'Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A' using xref '<xref:Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A>'",
"message_severity":"warning"}
```
The uid exists as 
```json
{
            "uid": "Microsoft.AspNetCore.WebHost.CreateDefaultBuilder*",
            "href": "/dotnet/api/microsoft.aspnetcore.webhost.createdefaultbuilder",
            "name": "CreateDefaultBuilder",
            "commentId": "Overload:Microsoft.AspNetCore.WebHost.CreateDefaultBuilder",
            "fullName": "Microsoft.AspNetCore.WebHost.CreateDefaultBuilder",
            "nameWithType": "WebHost.CreateDefaultBuilder"
        }
```
In V2, we are getting xrefSpec from xref service, url will decode the uid as
```
https://xref.docs.microsoft.com/query?uid=Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A
```
Shall we decode the uid in V3 as well, or is this a content issue?

yufeih:
The source file is using `<xref:Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A>` to reference the method. It is not a content issue and we need to decode `Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A` into `Microsoft.AspNetCore.WebHost.CreateDefaultBuilder*` for `<xref:{encoded_uid}>` references.

The thing I want to know is the behavior for `@` references. I suspect we don't need to decode `@` references. Can you help try and confirm the behavior for:

`@"Microsoft.AspNetCore.WebHost.CreateDefaultBuilder*"`, `@"Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A"`,
`@Microsoft.AspNetCore.WebHost.CreateDefaultBuilder*`
`@Microsoft.AspNetCore.WebHost.CreateDefaultBuilder%2A`

live1206:
@yufeih thanks for the clarification

superyyrrzz:
[...]

PR #3711, opened 2018-12-03
Refine xref reference

live1206:
According to [v2](https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#shorthand-form), shorthand form `@uid` should work almost the same as `<xref uid>`
- both should support displayProperty and view
- both should support hashtag
- both should url decode uid since the uid could contain `#`
- shorthand form `@uid` should not log a warning if uid not found since `@` is widely used
- implement merging `href` with `query` and `fragment` covered with unit tests
#3703

### DIFF
--- a/docs/designs/xref.md
+++ b/docs/designs/xref.md
@@ -111,7 +111,7 @@ Link to <xref:a?displayProperty=fullName#bookmark>
 ```json
 {"content":"<p>Link to <a href=\"/aspnet/a#bookmark\">ASP.NET Full Documentation</a></p>\n"}
 ```
->> Since hashtag is supported in xref, that means if you have # in UID, it has to be encoded to %23.
+> Since hashtag is supported in xref, that means if you have # in UID, it has to be encoded to %23.
 And then the xref format should follow URI standard, so all reserved characters should be encoded.
 
 ### Using `@` to reference a uid in markdown
--- a/docs/designs/xref.md
+++ b/docs/designs/xref.md
@@ -94,14 +94,34 @@ xref:
 During `docfx restore`, all the JSON files will be restored and merged.
 
 ## Resolve xref
-### Using `@` to reference a uid in markdown
-Not supporting `displayProperty` for this case for now, we need to refine the validation pattern of uid firstly, could reference to [V2](https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#shorthand-form).
+### Using `xref` to reference a uid in markdown
+#### Support `displayProperty`
 ```
-Link to @a
+Link to <xref:a?displayProperty=fullName>
+```
+  And it will be resolved as below, notice that the display title is from `fullName` not `name`
+```json
+{"content":"<p>Link to <a href=\"aspnet/a.md\">ASP.NET Full Documentation</a></p>\n"}
+```
+#### Support hashtag
+```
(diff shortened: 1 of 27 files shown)


src/docfx/build/xref/XrefMap.cs
```diff
@@ -37,6 +37,9 @@ public IEnumerable<XrefSpec> InternalReferences
 
         public (XrefSpec, Document) Resolve(string uid, string moniker = null)
```
superyyrrzz: >uid [](start = 51, length = 3)

Will this uid contains `#` or `?`? If not, it may be clearer to decode this while transforming the raw URI form cross refernce into uid+bookmark+query. This method only accepts a raw UID.

live1206: Good point, a further question would be: do we want to decode the raw URI while splitting href for all cases, not only for uid?

docs/specs/xref.yml
```diff
@@ -269,3 +269,22 @@ outputs:
...
+# Refer to uid with encoded content, it should be decoded then resolved
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    ---
+    title: Title from yaml header a
+    uid: a*
+    ---
+  docs/b.md: | 
+    Link to <xref:a%2A>
```
yufeih: Add a case to ensure query and fragment is preserved?

docs/designs/xref.md
```diff
@@ -94,14 +94,34 @@ xref:
...
+```
+#### Support hashtag
+```
+Link to <xref:a?displayProperty=fullName#bookmark>
+```
+  And it will be resolved as below, notice that the display title is from `fullName` not `name`
+```json
+{"content":"<p>Link to <a href=\"/aspnet/a#bookmark\">ASP.NET Full Documentation</a></p>\n"}
+```
+>> Since hashtag is supported in xref, that means if you have # in UID, it has to be encoded to %23.
```
superyyrrzz: >>> [](start = 0, length = 2)

minor: `>`?

src/docfx/lib/markdown/ResolveXref.cs
```diff
@@ -48,6 +50,11 @@ public static MarkdownPipelineBuilder UseResolveXref(this MarkdownPipelineBuilde
                          var displayPropertyValue = xrefSpec.GetXrefPropertyValue(queries?["displayProperty"]);
                          string display = !string.IsNullOrEmpty(displayPropertyValue) ? displayPropertyValue : (!string.IsNullOrEmpty(name) ? name : uid);
                          var href = !string.IsNullOrEmpty(moniker) ? $"{xrefSpec.Href}?view={moniker}" : xrefSpec.Href;
+
+                         // append fragment if any
+                         if (!string.IsNullOrEmpty(fragment))
+                             href += fragment;
```
superyyrrzz: >href [](start = 29, length = 4)

href may also contains fragment

docs/designs/xref.md
live1206: right

src/docfx/lib/markdown/ResolveXref.cs
yufeih: There are other places where we need this merge functionality

live1206: Implemented with unit tests

PR #3715, opened 2018-12-04
Remove legacy xref

live1206:
Had an offline discussion with @yufeih:
- Compared to `JObject`, which only could be `JObject`, `JArray` or `JValue`, `Dictionary<string, object>` the value could be anything, revert the change
- Since OP is outputting `xrefmap.json`, legacy `xrefmap.yaml` is not needed for comparison
- Add TODO that xref service needs to consume `xrefmap.json` instead


src/docfx/build/legacy/Legacy.cs
```diff
@@ -27,7 +26,8 @@ internal static class Legacy
                 LegacyFileMap.Convert(docset, context, files);
                 LegacyDependencyMap.Convert(docset, context, files, dependencyMap, tocMap);
                 LegacyCrossRepoReferenceInfo.Convert(docset, context);
-                LegacyXrefMap.Convert(docset, context, xrefMap);
+
+                // TODO: removed xrefmap.yml legacy output, xref service needs to consume xrefmap.json
```
yufeih: there aren't any items on now side to support xrefmap.json right?

PR #3717, opened 2018-12-04
report warning if a toc file is referenced but it is not in build scope

928PJY:


docs/specs/schema.yml
```diff
@@ -65,40 +65,39 @@ outputs:
...
-  # TODO: why is there no output for this toc file?
-  TOC.yml:
+  docs/A/TOC.yml:
   docs/TOC.yml:
   docs/a.yml: |
     #YamlMime:ContextObject
     brand: azure
     breadcrumb_path: TOC.yml
-    toc_rel: ../TOC.yml
+    toc_rel: A/TOC.yml
```
OsmondJiang: what's that schema defination?

928PJY: https://github.com/dotnet/docfx/blob/v3/schemas/ContextObject.json#L25

PR #3719, opened 2018-12-04
show wanings on page when codesnippet not found

yishengjin1413:
Refer to this feature: https://ceapex.visualstudio.com/Engineering/_workitems/edit/20848/.
Spec is here: https://review.docs.microsoft.com/en-us/new-hope/specs/ops/missing-sample-build?branch=master#design

We want to show a warning on page when the codesnippt is not found.

### DIFF
--- a/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/CodeSnippet/HtmlCodeSnippetRenderer.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/CodeSnippet/HtmlCodeSnippetRenderer.cs
@@ -18,7 +18,7 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
         private readonly MarkdownContext _context;
         private const string tagPrefix = "snippet";
         private const string warningMessageId = "codeIncludeNotFound";
-        private const string defaultWarningMessage = "It looks like the sample you are looking for has moved!  Rest assured we are working on resolving this.";
+        private const string defaultWarningMessage = "It looks like the sample you are looking for does not exist.";
         private const string warningTitleId = "warning";
         private const string defaultWarningTitle = "<h5>WARNING</h5>";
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/CodeSnippet/HtmlCodeSnippetRenderer.cs
```diff
@@ -17,6 +17,10 @@ public class HtmlCodeSnippetRenderer : HtmlObjectRenderer<CodeSnippet>
     {
         private readonly MarkdownContext _context;
         private const string tagPrefix = "snippet";
+        private const string warningMessageId = "codeIncludeNotFound";
+        private const string defaultWarningMessage = "It looks like the sample you are looking for has moved!  Rest assured we are working on resolving this.";
```
yufeih: Are you going to modify template to include these messages?

yishengjin1413: Yes. My change have checked in and localized: https://github.com/Microsoft/templates.docs.msft.zh-cn/blob/master/ContentTemplate/token.json#L75

superyyrrzz: The wording is good for docs user, but not well applies to community user. What about "It looks like the sample you are looking for does not exist." as default message?

superyyrrzz:
Will you also change dfm engine?

yishengjin1413:
@superyyrrzz I think we don't need to update DFM since it will retired soon and I remember we have stop supporting DFM long time ago, right?  Just keep the DFM have no breaking change for it is ok.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/CodeSnippet/HtmlCodeSnippetRenderer.cs
yishengjin1413: Oh, so for en-us build, template will also pass the token to markdown engine? If that, I think it's make sense to use the message you mentioned as the default string.

superyyrrzz: Yes, it exists: https://github.com/Microsoft/templates.docs.msft/blob/master/ContentTemplate/token.json#L75

PR #3721, opened 2018-12-04
reorder errors

mingwli:


src/docfx/Errors.cs
```diff
@@ -4,12 +4,14 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 
 namespace Microsoft.Docs.Build
 {
     internal static class Errors
     {
+#pragma warning disable SA1124 // DoNotUseRegions
```
OsmondJiang: how about we create three files `Errors.cs, Warnings.cs and Info.cs` put them under one folder?

src/docfx/Errors.cs
```diff
@@ -4,12 +4,14 @@
...
+
+        #region Level Error
```
yufeih: will this impose additional maintenance burden? if an info is turned into warning we'll have to do the house keep work to move it to a difference location?

PR #3722, opened 2018-12-05
Remove isInclude and monikers from TOC resolve delegate

yufeih:
Update TOC resolve delegate to match markdown resolve delegate, so TOC resolve and markdown resolve shares the same signature.


docs/specs/toc.yml
```diff
@@ -600,7 +600,7 @@ outputs:
 # replace href finally
 inputs:
   docfx.yml: |
-    files: docs/f1/TOC.yml
+    files: docs/**/TOC.*
```
OsmondJiang: what if we keep the test unchanged, do we changed the behavior?

yufeih: in this test case, `TOC.md` is resolved using dynamic scope (which is unexpected). We only do dynamic scope for resources.

OsmondJiang: if it's belongs to inclusion, i think the behavior is just like token, right?

yufeih: This is a weird case, and feels more like a link.
It does not embed the content of the sub TOC, it just use the first link in sub TOC.

yufeih: Either way is OK. And if it is treated like inclusion, it will also handle dynamic scope, dependency map, git history fallback for loc, etc.

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -15,13 +15,14 @@ internal static class TableOfContentsParser
         private static readonly string[] s_tocFileNames = new[] { "TOC.md", "TOC.json", "TOC.yml" };
         private static readonly string[] s_experimentalTocFileNames = new[] { "TOC.experimental.md", "TOC.experimental.json", "TOC.experimental.yml" };
 
-        public delegate (string resolvedTopicHref, List<string> monikers) ResolveHref(Document relativeTo, string href, Document resultRelativeTo);
+        public delegate (Error error, string href, Document file) ResolveHref(Document relativeTo, string href, Document resultRelativeTo);
 
-        public delegate (string content, Document file) ResolveContent(Document relativeTo, string href, bool isInclusion);
+        public delegate (Error error, string content, Document file) ResolveContent(Document relativeTo, string href);
```
OsmondJiang: > public delegate (Error error, string content, Document file) ResolveContent(Document relativeTo, string href); [](start = 7, length = 111)

if you want to make it consistent with BuildPage, do you want move this delegate out? You are trying to clean the interface? but actually finally we all call TryResolveContent with some wrapper?

OsmondJiang: then if we always need the wrapper, we can be different right?

---
In reply to: [238950245](https://github.com/dotnet/docfx/pull/3722#discussion_r238950245) [](ancestors = 238950245)

yufeih:
Close this monikers addressed by #3782

PR #3723, opened 2018-12-05
Unshallow repos when fetch with --depth 1 failed.

### LINKED ISSUE #3710: can't add worktree when clone repo with depth 1
git-revparse "committash" and git worktree add don't work.

yufeih:
Fixes #3710

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -16,4 +16,4 @@ Remove-Item ./drop -Force -Recurse -ErrorAction Ignore
 exec "dotnet run -p tools/CreateJsonSchema"
 exec "dotnet test test\docfx.Test"
 exec "dotnet test test\docfx.Test -c Release"
-exec "dotnet pack src\docfx -c Release -o $PSScriptRoot\drop /p:Version=$version /p:InformationalVersion=$version /p:PackAsTool=true"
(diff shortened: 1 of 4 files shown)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -254,6 +254,7 @@ private static async Task CloneOrUpdate(string path, string url, IEnumerable<str
             catch (InvalidOperationException ex) when (committishes.Any(rev => ex.Message.Contains(rev)))
             {
                 // Fallback to fetch all branches and tags if the input committish is not supported by fetch
+                depth = depthOne ? "--unshallow" : "";
```
OsmondJiang: make it full clone/fetch again? so the depth==1 only works for refName and tags?

yufeih: `depth == 1` only works for branches and tags, you cannot fetch a single with commit hash with git AFAIK.

live1206:
When `CloneOrUpdate` failed in an empty repository, `--unshallow` will break:
```
git  fetch --tags --prune --progress --update-head-ok --unshallow "https://github.com/Azure-Samples/azure-iot-samples-node" +refs/heads/*:refs/heads/* +refs/tags/*:refs/tags/*
fatal: --unshallow on a complete repository does not make sense
```

PR #3724, opened 2018-12-05
get deleted inclusion(token/codesnippet/toc) from git commit history

OsmondJiang:
1. add get deleted file commits to file commit provider, it's not using cache
2. make the git commit provider more easy to read
3. pass down the git commit provider, not added to docset
4. token/codesnippet/toc are all supported
5. only works for loc docset build


src/docfx/docset/Document.cs
```diff
@@ -194,7 +205,7 @@ public override string ToString()
         /// </summary>
         /// <param name="docset">The current docset</param>
         /// <param name="path">The path relative to docset root</param>
-        public static (Error error, Document doc) TryCreate(Docset docset, string path, string redirectionUrl = null)
+        public static (Error error, Document doc) TryCreate(Docset docset, string path, string redirectionUrl = null, bool isFromHistory = false)
```
live1206: should this method be 
```csharp
public static bool TryCreate(Docset docset, string path, out Error error, out Document doc, string redirectionUrl = null, bool isFromHistory = false)
```
Then, we do not need to check error again at outer layer.

src/docfx/lib/git/GitCommitProvider.cs
```diff
@@ -11,70 +11,66 @@ namespace Microsoft.Docs.Build
     internal sealed class GitCommitProvider : IDisposable
     {
         private readonly ConcurrentDictionary<string, Repository> _repositoryByFolder = new ConcurrentDictionary<string, Repository>();
-        private readonly ConcurrentDictionary<string, Lazy<Task<(FileCommitProvider provider, string repoRemote)>>> _fileCommitProvidersByRepoPath = new ConcurrentDictionary<string, Lazy<Task<(FileCommitProvider provider, string repoRemote)>>>();
+        private readonly ConcurrentDictionary<string, (FileCommitProvider provider, string repoRemote)> _fileCommitProvidersByRepoPath = new ConcurrentDictionary<string, (FileCommitProvider provider, string repoRemote)>();
```
live1206: Shall we keep `lazy` here?
Let's say, `GetOrAdd` get invoked concurrently, only 1 `FileCommitProvider` been added to the dictionary, and the ones not been added have been instanced. How could we dispose them?

OsmondJiang: good point here, let me change the code

src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -168,13 +135,62 @@ public List<GitCommit> GetCommitHistory(string file, string committish = null)
             return result.Select(c => c.GitCommit).ToList();
         }
 
-        public IReadOnlyList<KeyValuePair<string, Dictionary<(long commit, long blob), (long[] commitHistory, int lruOrder)>>> BuildCache()
+        public List<GitCommit> GetDeletedFileCommitHistory(string file, int top = -1, string committish = null)
```
yufeih: instead of duplicating this complex algorithm, can we have one method that get commit history with top N and an optional cache?

docs/specs/localization.yml
```diff
@@ -788,3 +787,75 @@ outputs:
...
+  - build
+repos:
+  https://github.com/onlyloc/a:
+    - files:
+        docfx.yml:
+        docs/b/TOC.yml: |
+          - href: ../a/TOC.yml
+            name: reference toc a
+        docs/a/a.md:
+      author: a
```
yufeih: author should be optional

src/docfx/lib/git/FileCommitProvider.cs
OsmondJiang: i can't think a way to get ride of `async` for optional cache, and i don't want to use `.result`, so any suggestions here?

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -29,26 +29,32 @@ public static (List<Error> errors, List<TableOfContentsItem> items, JObject meta
...
+
+            if (file.IsFromHistory)
+            {
+                Debug.Assert(!string.IsNullOrEmpty(content));
+            }
+
             if (filePath.EndsWith(".yml", PathUtility.PathComparison))
             {
-                var (errors, tocToken) = YamlUtility.Deserialize(file, context);
+                var (errors, tocToken) = !file.IsFromHistory ? YamlUtility.Deserialize(file, context) : YamlUtility.Deserialize(content);
```
yufeih: nit: Instead of do the switch based on `IsFromHistory`, how about do the switch based on `content != null` ?

docs/specs/localization.yml
yufeih: include _do not fallback token/codesnippet for english_ test

src/docfx/lib/git/FileCommitProvider.cs
yufeih: If could use some refactor of `Cache`, basically we can have a single method that contains the algorithm:

```csharp
List<GitCommit> GetCommitHistory(string file, int top, string committish, Cache commitCache)
```

and then have the async method calling the sync method, something similar to:

```csharp
async Task<List<GitCommit>> GetCommitHistory(string file, string committish)
{
        var cache = await GetCache();
        var result = GetCommitHistory(file, committish, cache)
        UpdateCache()
        return result;
}
```

I personally do not want changes to GitCommitProvider to mix with other changes, because there aren't enough tests in PR to cover GitCommitProvider and it is performance sensitive. (The only safeguard for GitCommitProvider right now is impact test).

docs/specs/localization.yml
OsmondJiang: ok

src/docfx/build/toc/parser/TableOfContentsParser.cs
OsmondJiang: I prefer to use the IsFromHistory which is more clear to me why we need do that

src/docfx/lib/git/FileCommitProvider.cs
OsmondJiang: we can do the cache refactor later, current goal is to:
1. serving current purpose
2. reduce the duplicate code as much as possible
3. make the code more readable, don't change the logic at all

OsmondJiang: we don't have any other mix changes here, just re-structure the code to make it more readable.

OsmondJiang:
#3560

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -29,26 +29,32 @@ public static (List<Error> errors, List<TableOfContentsItem> items, JObject meta
...
+
+            if (file.IsFromHistory)
+            {
+                Debug.Assert(!string.IsNullOrEmpty(content));
+            }
+
             if (filePath.EndsWith(".yml", PathUtility.PathComparison))
             {
-                var (errors, tocToken) = YamlUtility.Deserialize(file, context);
+                var (errors, tocToken) = string.IsNullOrEmpty(content) ? YamlUtility.Deserialize(file, context) : YamlUtility.Deserialize(content);
```
yufeih: it should be `content == null`, content read from history might be empty ""

OsmondJiang: yes

Issue #3718, opened 2018-12-04
Update GitHub use cache aware of ETag


superyyrrzz:
Follow #2812, Implement #3712

- [x] Restore url with ETag recorded, using ETag to avoid duplicate download #3726 
- [x] Update cache in cache folder, instead of in restore folder #3740
- [x] Upload cache with ETag #3756
- [x] ~~**Optional**: `docfx gc` clean up restored cache~~

yufeih:
Let's not worry about gc cleanup as the cache is too small.

PR #3726, opened 2018-12-05
restore with etag

superyyrrzz:
#3718

### DIFF
--- a/src/docfx/lib/HttpClientUtility.cs
+++ b/src/docfx/lib/HttpClientUtility.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Diagnostics;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading.Tasks;
 
 namespace Microsoft.Docs.Build
@@ -14,7 +15,7 @@ namespace Microsoft.Docs.Build
         private const int RetryInterval = 1000;
         private static readonly HttpClient s_httpClient = new HttpClient();
 
-        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config)
+        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config, string etag = null)
         {
             HttpResponseMessage response;
             for (var i = 0; i < RetryCount; i++)
@@ -25,6 +26,10 @@ namespace Microsoft.Docs.Build
                     // "The request message was already sent. Cannot send the same request message multiple times."
                     var message = CreateHttpRequestMessage(requestUri, config);
                     message.Method = HttpMethod.Get;
+                    if (!string.IsNullOrEmpty(etag))
+                    {
+                        message.Headers.IfNoneMatch.Add(new EntityTagHeaderValue(etag));
+                    }
                     response = await s_httpClient.SendAsync(message);
                 }
                 catch (Exception ex) when (
--- a/src/docfx/lib/markdown/ResolveXref.cs
+++ b/src/docfx/lib/markdown/ResolveXref.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
 using System.Web;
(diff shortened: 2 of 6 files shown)


test/docfx.Test/restore/RestoreTest.cs
```diff
@@ -98,5 +98,23 @@ public static async Task RestoreUrls()
...
+        [InlineData("abc123", "", "abc123+etag+")]
+        [InlineData("abc123", null, "abc123")]
+        public static void GetRestoreFileName(string hash, string etag, string expected)
+        {
+            Assert.Equal(expected, RestoreFile.GetRestoreFileName(hash, etag));
+        }
+
+        [Theory]
+        [InlineData("abc123+etag+%220xdef456%22", "\"0xdef456\"")]
+        [InlineData("abc123+etag+", "")]
```
yufeih: what is the purpose of `+etag+` here? is `abc123-%220xdef456%22` enough?

src/docfx/lib/HttpClientUtility.cs
```diff
@@ -25,6 +26,10 @@ public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config
                     // "The request message was already sent. Cannot send the same request message multiple times."
                     var message = CreateHttpRequestMessage(requestUri, config);
                     message.Method = HttpMethod.Get;
+                    if (!string.IsNullOrEmpty(etag))
+                    {
+                        message.Headers.IfNoneMatch.Add(new EntityTagHeaderValue(etag));
```
yufeih: where is the logic to handle `304 Not Modified` if the etag did match.

test/docfx.Test/restore/RestoreTest.cs
superyyrrzz: `-` is not a safe separator as the encoded etag can also contain it. It can be `abc123+%220xdef456%22`, just insert `etag` to make it more readable.

src/docfx/lib/HttpClientUtility.cs
superyyrrzz: https://github.com/dotnet/docfx/pull/3726/files#diff-10735dd8751bdbb809d9a5452f423fadR96

src/docfx/restore/RestoreFile.cs
```diff
@@ -12,15 +15,24 @@ internal static class RestoreFile
     {
         public static async Task Restore(string url, Config config, bool @implict)
         {
-            if (implict && RestoreMap.TryGetFileRestorePath(url, out _))
+            if (RestoreMap.TryGetFileRestorePath(url, out var existingPath) && implict)
```
OsmondJiang: >implict [](start = 79, length = 7)

why change the order?

src/docfx/restore/RestoreFile.cs
```diff
@@ -43,15 +55,50 @@ Task MoveFile()
             }
         }
 
-        private static async Task<string> DownloadToTempFile(string url, Config config)
+        /// <summary>
+        /// Get restore file name from hash and ETag
+        /// </summary>
+        /// <param name="hash">SHA1 hash of file content</param>
+        /// <param name="etag">ETag of the resource, null if file is from local, empty if server doesn't specify ETag</param>
+        public static string GetRestoreFileName(string hash, string etag = null)
```
OsmondJiang: >GetRestoreFileName [](start = 29, length = 18)

when can this etag be null?

src/docfx/restore/RestoreFile.cs
superyyrrzz: Before, we needn't call `RestoreMap.TryGetFileRestorePath` when `implicit == false`, but it is needed now.

src/docfx/lib/markdown/ResolveXref.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
```
mingwli: is this used?

src/docfx/restore/RestoreFile.cs
OsmondJiang: we either don't need do that, just split it into two `if`

---
In reply to: [239327888](https://github.com/dotnet/docfx/pull/3726#discussion_r239327888) [](ancestors = 239327888)

src/docfx/lib/markdown/ResolveXref.cs
superyyrrzz: on removed..

PR #3728, opened 2018-12-05
Implement sync GetCommitHistory with top N commits

yufeih:
- Revert the split of FileCommitProvider commit cache because CommitCache shares too much detail with CommitProvider (hidden locks, data structures), in another word, they are highly coherent.

- Implement `GetCommitHistory` with top N commits synchronously that leveraged the commit cache if possible.

#3724


OsmondJiang:
>                                 result.Add(commitsBySha[cachedCommit]);

so if i hit cache, the topc doesn't work?

---
Refers to: src/docfx/lib/git/FileCommitProvider.cs:142 in 9b9534b. [](commit_id = 9b9534bc43f9ecfa0d9153601489421daaa0694a, deletion_comment = False)

src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -34,27 +38,48 @@ internal sealed class FileCommitProvider : IDisposable
...
+
+        public async Task<List<GitCommit>> GetCommitHistory(string file, string committish = null)
+        {
+            var commitCache = await _commitCache.Value;
+            var fileCommitCache = commitCache.GetOrAdd(file, _ => new Dictionary<(long, long), (long[], int)>());
+
+            return GetCommitHistory(file, int.MaxValue, committish, fileCommitCache);
+        }
+
+        public List<GitCommit> GetCommitHistory(string file, int top, string committish = null)
```
OsmondJiang: normal file should not call this method ever, so why not change it to GetDeletedCommitHistory

OsmondJiang: and no need leverage cache anyway

OsmondJiang:
@yufeih , would you please change your pr to focus on the refactor cache? my pr has cover the other parts

src/docfx/lib/git/FileCommitProvider.cs
OsmondJiang: as we talked offline, these two method should not have any relationship, so one is with cache, one is without cache

OsmondJiang:
> 

and make some refactor to this long code as we discussed offline?

---
Refers to: src/docfx/lib/git/FileCommitProvider.cs:190 in 1e45de0. [](commit_id = 1e45de0d3336cc02bf1c8b0a401635d3e445da09, deletion_comment = False)

src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -34,27 +38,64 @@ internal sealed class FileCommitProvider : IDisposable
...
```
OsmondJiang: >GetCommitHistory [](start = 31, length = 16)

WithoutCache?

PR #3729, opened 2018-12-05
Fix FSharp tests

Frassle:
This fixes the FSharp tests so that, they run and pass and are also included by build.ps1 to run in CI.

### DIFF
--- a/test/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests.fsproj
+++ b/test/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests.fsproj
@@ -9,9 +9,14 @@
   <Target Name="RestoreTestProjects" BeforeTargets="Build">
     <ItemGroup>
       <TestProjects Include="TestData/**/*.fsproj" />
-      <TestData Include="TestData/**/*" />
     </ItemGroup>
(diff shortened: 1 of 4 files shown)


superyyrrzz:
@Frassle test failed on CI:

> The test source file "D:\a\1\s\test\Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests\bin\Release\net462\TestData\NetCoreLibProject\bin\Release\netstandard2.0\NetCoreLibProject.dll" provided was not found.

Frassle:
Ah I think I know the issue, I'll fix it over the weekend. Just need to exclude the test data projects from the test list.

test/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests/Microsoft.DocAsCode.Metadata.ManagedReference.FSharp.Tests.fsproj
```diff
@@ -6,11 +6,21 @@
     <Compile Include="FSharpProjectTests.fs" />
   </ItemGroup>
 
-  <ItemGroup>
-    <None Include="TestData/*/*.fsproj;TestData/**/*.fs">
-      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
-    </None>
-  </ItemGroup>
+  <Target Name="RestoreTestProjects" BeforeTargets="Build">
```
Frassle: I'm not an MSBuild expert so there may be a better way of doing a restore then copy but I could only get this working reliably via two targets.

superyyrrzz:
Thank you @Frassle 
/cc @928PJY

PR #3740, opened 2018-12-07
update cache to cache folder instead of restore folder

superyyrrzz:
#3718 

Before:
* read cache: from restore folder
* writer cache: to restore folder

After:
* read cache: from restore folder and a local temp cache
* write cache: to local temp cache

### DIFF
--- a/src/docfx/lib/github/GitHubUserCache.cs
+++ b/src/docfx/lib/github/GitHubUserCache.cs
@@ -37,6 +37,8 @@ namespace Microsoft.Docs.Build
         private readonly ConcurrentDictionary<string, Lazy<Task<(Error, string)>>> _outgoingGetLoginByCommitRequests
                    = new ConcurrentDictionary<string, Lazy<Task<(Error, string)>>>(StringComparer.OrdinalIgnoreCase);
 
+        private readonly string _url = null;
+        private readonly string _restorePath = null;
         private readonly string _cachePath;
         private readonly double _expirationInHours;
         private bool _updated = false;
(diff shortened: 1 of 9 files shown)


src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -236,15 +252,21 @@ private DateTime NextExpiry()
 
         private async Task ReadCacheFile()
         {
-            if (File.Exists(_cachePath))
+            await Read(_cachePath);
+            await Read(_restorePath);
```
OsmondJiang: >_restorePath [](start = 23, length = 12)

first restorePath and then cachePath? it's overwrite right?

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -56,10 +58,24 @@ private GitHubUserCache(Docset docset, string token)
...
-                : docset.GetFileRestorePath(docset.Config.GitHub.UserCache);
             _expirationInHours = docset.Config.GitHub.UserCacheExpirationInHours;
+            var path = docset.Config.GitHub.UserCache;
+            if (string.IsNullOrEmpty(path))
+            {
+                _cachePath = AppData.DefaultGitHubUserCachePath;
+                return;
+            }
+            var restorePath = docset.GetFileRestorePath(path);
+            if (HrefUtility.IsHttpHref(path))
```
OsmondJiang: >path [](start = 39, length = 4)

even it's from local file, should be different with your cache file, then why we abandond the cache file?

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -142,22 +158,22 @@ public async Task<Error> SaveChanges(Config config)
                     Users = Users.ToArray(),
```
OsmondJiang: > Users.ToArray() [](start = 27, length = 16)

toArray already makesure it's an snipshoot, no need lock?

src/docfx/lib/github/GitHubUserCache.cs
superyyrrzz: This is more like merge. I think the order does not matter in most cases.

---
In reply to: [239724827](https://github.com/dotnet/docfx/pull/3740#discussion_r239724827) [](ancestors = 239724827)

src/docfx/lib/github/GitHubUserCache.cs
superyyrrzz: This also ensures thread safe for every `User` instance, as `UpdateUser` will modify its property.

---
In reply to: [239725394](https://github.com/dotnet/docfx/pull/3740#discussion_r239725394) [](ancestors = 239725394)

src/docfx/lib/github/GitHubUserCache.cs
superyyrrzz: If user explicitly set it a local file, I think mostly it means to read and write to this file.

---
In reply to: [239725092](https://github.com/dotnet/docfx/pull/3740#discussion_r239725092) [](ancestors = 239725092)

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -56,10 +58,24 @@ private GitHubUserCache(Docset docset, string token)
...
-                ? AppData.GitHubUserCachePath
-                : docset.GetFileRestorePath(docset.Config.GitHub.UserCache);
             _expirationInHours = docset.Config.GitHub.UserCacheExpirationInHours;
+            var path = docset.Config.GitHub.UserCache;
+            if (string.IsNullOrEmpty(path))
+            {
+                _cachePath = AppData.DefaultGitHubUserCachePath;
+                return;
+            }
+            var restorePath = docset.GetFileRestorePath(path);
```
mingwli: >GetFileRestorePath [](start = 37, length = 18)

Suggestion: this method may return (restorePath, isFromURl) to avoid invoking `HrefUtility.IsHttpHref` twice.

PR #3741, opened 2018-12-07
 throw errors when meets reserved metadata

OsmondJiang:
metadata includes yaml header, file metadata and global metadata


docs/specs/metadata.yml
```diff
@@ -312,3 +312,41 @@ outputs:
...
+  docfx.yml:
+  docs/a.md: |
+    ---
+    Redirect_url: https://docfx.com
+    ---
+outputs:
+  docs/a.json: |
+    { "metadata":{} }
+  build.log: |
+    ["warning","redirection-in-metadata","Redirection defined in metadata(yaml header, file metadata or global metadata) is not supported, docfx only supports redirections defined in docfx.yml","docs/a.md"]
```
live1206: >docfx.yml [](start = 183, length = 9)

`docfx.yml/docfx.json`, we also support JSON

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -33,6 +33,18 @@ public MonikersProvider(Docset docset)
             Comparer = new MonikerComparer(monikerDefinition);
         }
 
+        public (List<Error> errors, List<string> monikers) GetFileLevelMonikers(Document file, MetadataProvider metadataProvider)
```
live1206: >GetFileLevelMonikers [](start = 59, length = 20)

Can we consolidate these two into one method?
The usage of the one below seems the same as this one.

OsmondJiang: as we talked offline, the other parts' file metadata is used in other places, so two methods are ok.

live1206: we still can consolidate with inputting `metadata`, right?

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -48,7 +48,13 @@ public JObject GetMetadata(Document file, JObject yamlHeader = null)
             {
                 result.Merge(yamlHeader, JsonUtility.MergeSettings);
             }
-            return result;
+
+            if (result.TryGetValue("redirect_url", StringComparison.OrdinalIgnoreCase, out _))
+            {
+                return (Errors.RedirectionInMetadata(), result);
```
yufeih: This could be solved in a general way with a more general error code. There are more metadatas to block in addition to `redirect_url`, like `monikers`. I'd like to see a way to _attribute_ these blacklisted metadata.

src/docfx/build/xref/XrefMap.cs
```diff
@@ -244,7 +244,7 @@ private XrefMap(IReadOnlyDictionary<string, XrefSpec> externalXrefMap, IReadOnly
                 if (file.FilePath.EndsWith(".md", PathUtility.PathComparison))
                 {
                     var (yamlHeaderErrors, yamlHeader) = ExtractYamlHeader.Extract(file, context);
-                    var (metaErrors, metadata) = JsonUtility.ToObjectWithSchemaValidation<FileMetadata>(metadataProvider.GetMetadata(file, yamlHeader));
+                    var (metaErrors, metadata) = JsonUtility.ToObjectWithSchemaValidation<FileMetadata>(metadataProvider.GetMetadata(file, yamlHeader).metadata);
```
yufeih: errors are ignore here?

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -48,7 +49,47 @@ public JObject GetMetadata(Document file, JObject yamlHeader = null)
...
+
+            if (invalidNames.Count > 0)
+            {
+                return (Errors.ReservedMetadata(invalidNames), result);
+            }
+
+            return (null, result);
+        }
+
+        private static HashSet<string> GetReservedMetadata()
```
yufeih: ((JsonObjectContract)JsonUtility.DefaultSerializer.ContractResolver.ResolveContract(typeof(PageModel))).Properties

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -48,7 +50,52 @@ public JObject GetMetadata(Document file, JObject yamlHeader = null)
             {
                 result.Merge(yamlHeader, JsonUtility.MergeSettings);
             }
-            return result;
+
+            var invalidNames = new List<string>();
```
yufeih: Generate one message per key. This aligns with the warning messages we generate from `JsonUtility` validation for yaml header.

OsmondJiang: i would like to use one error message per file, which is more user-friendly.

Issue #3727, opened 2018-12-05
Support xrefmap.yml as input


superyyrrzz:
When community user migrates docset from v2 to v3, it's hard to ensure that the referenced `xrefmap.yml` already has been migrated to `xrefmap.json`. This can be a obstacle to adopt v3.

Come up with this as recently I find [a user case](https://github.com/richardschneider/net-ipfs-http-client/blob/master/doc/docfx.json#L69) using this to reference cross docset. [After some search](https://github.com/search?utf8=%E2%9C%93&q=%22xref%22+%22xrefmap.yml%22+filename%3Adocfx.json+-org%3AMicrosoftDocs+-org%3AOpenLocalizationTestOrg+-org%3Aaspnet+-org%3Akilla1218+-org%3Abetterliyu+-org%3Asuperyyrrzz+-org%3Ayannanwang1&type=), I find this is not a rare case among conmunity users. There should be more cases if users don't put their `docfx.json` on GitHub.

@live1206 @yufeih

yufeih:
Do you want to support zip files as well? `https://github.com/richardschneider/net-ipfs-http-client/blob/master/doc/docfx.json#L68`

superyyrrzz:
This is optional. It is kind of an old way to reference BCL, which can migrate to xrefService or `xrefmap.json` easily.

PR #3744, opened 2018-12-07
Consume xrefmap.yml as well

live1206:
#3727
V3 should consume `xrefmap.yml` as well

### DIFF
--- a/src/docfx/build/xref/XrefMap.cs
+++ b/src/docfx/build/xref/XrefMap.cs
@@ -94,9 +94,17 @@ namespace Microsoft.Docs.Build
             Dictionary<string, XrefSpec> map = new Dictionary<string, XrefSpec>();
             foreach (var url in docset.Config.Xref)
             {
-                var json = File.ReadAllText(docset.GetFileRestorePath(url));
-                var xRefMap = JsonUtility.Deserialize<XrefMapModel>(json);
-                foreach (var spec in xRefMap.References)
+                var content = File.ReadAllText(docset.GetFileRestorePath(url));
+                XrefMapModel xrefMap = null;
+                if (url.EndsWith(".yml", StringComparison.OrdinalIgnoreCase))
+                {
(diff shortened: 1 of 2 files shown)


src/docfx/build/xref/XrefMap.cs
```diff
@@ -94,9 +94,17 @@ public IEnumerable<XrefSpec> InternalReferences
...
+                var content = File.ReadAllText(docset.GetFileRestorePath(url));
+                XrefMapModel xrefMap = new XrefMapModel();
+                if (url.EndsWith(".yml", StringComparison.OrdinalIgnoreCase))
+                {
+                    xrefMap = YamlUtility.Deserialize<XrefMapModel>(content);
+                }
+                else if (url.EndsWith(".json", StringComparison.OrdinalIgnoreCase))
+                {
+                    xrefMap = JsonUtility.Deserialize<XrefMapModel>(content);
+                }
```
superyyrrzz: log error if neither JSON nor YAML?

live1206: Seems like we are not logging an error if file extension not matched?
https://github.com/dotnet/docfx/blob/ff26cee570ccbcb528321a4c564d2b4061a49c77/src/docfx/config/Config.cs#L255

superyyrrzz: The caller has a guard to handle this:
https://github.com/dotnet/docfx/blob/ff26cee570ccbcb528321a4c564d2b4061a49c77/src/docfx/config/Config.cs#L174

live1206: I think that's because we always need a config file, but we are not warning the user that some extension are not supported.
Like this one as well, we do not warn the user the page extension is not supported, just make sure it is what we want:
https://github.com/dotnet/docfx/blob/45a0c9c7951c3bb9c7a7be98ce6b1a2b449c0b58/src/docfx/build/page/BuildPage.cs#L87

PR #3747, opened 2018-12-10
fix typo and mior ajusting in README

superyyrrzz:


README.md
```diff
@@ -10,7 +10,7 @@ We are still at an early develop phase, see our [roadmap](https://github.com/dot
 >  ```powershell
 >  dotnet tool install -g docfx --version 3.0.0-* --add-source https://www.myget.org/F/docfx-v3/api/v2
 >  ```
-> Note that the current output is raw JSON without any template applied, and it does not support source code or DLL as input for API reference. Right now it foucses on Markdown input parity with https://docs.microsoft.com
+> Note that it foucses on Markdown input parity with https://docs.microsoft.com, while the current output is still raw JSON without any template applied. API reference for source code or DLL will be one of the next focuses.
```
mingwli: focuses

PR #3750, opened 2018-12-10
change legacy metadata fileRelativePath outputRootRelativePath resolve logic

928PJY:


src/docfx/build/legacy/LegacyUtility.cs
```diff
@@ -50,7 +50,7 @@ public static string ToLegacySiteUrlRelativeToBaseSitePath(this Document doc, Do
         }
 
         public static string ToLegacyOutputPath(this LegacyManifestOutputItem legacyManifestOutputItem, Docset docset, string groupId)
-            => Path.Combine($"{groupId}", docset.Config.DocumentId.SiteBasePath, legacyManifestOutputItem.RelativePath);
+            => Path.Combine(docset.Config.DocumentId.SiteBasePath, $"{groupId}", legacyManifestOutputItem.RelativePath);
```
superyyrrzz: >$"{groupId}" [](start = 67, length = 12)

directly `groupId`?

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -96,5 +96,17 @@ public List<string> GetZoneMonikers(string rangeString, List<string> fileLevelMo
             monikers.Sort(Comparer);
             return monikers;
         }
+
+        public string GetGroupIdFromMonikerList(IEnumerable<string> monikers)
```
superyyrrzz: `IEnumerable` => `List`?

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -96,5 +96,17 @@ public List<string> GetZoneMonikers(string rangeString, List<string> fileLevelMo
...
+        {
+            if (monikers == null || monikers.Count() == 0)
+            {
+                return null;
+            }
+            var sortedMonikers = monikers.ToList();
+            sortedMonikers.Sort(Comparer);
```
superyyrrzz: Seems there are many duplicates `Sort` here. Add some cache?

src/docfx/build/moniker/MonikersProvider.cs
superyyrrzz: What about `string HashUtility.GetMd5HashShortStable(List<string> items)`?

src/docfx/build/legacy/LegacyUtility.cs
928PJY: groupId maybe null

---
In reply to: [240446206](https://github.com/dotnet/docfx/pull/3750#discussion_r240446206) [](ancestors = 240446206)

src/docfx/build/moniker/MonikersProvider.cs
928PJY: yes, but this is a utility, we have to make sure the output is stable

---
In reply to: [240448390](https://github.com/dotnet/docfx/pull/3750#discussion_r240448390) [](ancestors = 240448390)

src/docfx/build/moniker/MonikersProvider.cs
928PJY: that requires a comparer to be parsed into?

---
In reply to: [240448635](https://github.com/dotnet/docfx/pull/3750#discussion_r240448635) [](ancestors = 240448635)

src/docfx/lib/HashUtility.cs
```diff
@@ -54,6 +55,16 @@ public static string GetFileSha1Hash(string fileName)
             }
         }
 
+        public static string GetSortedMonikerListHash(List<string> monikers)
```
superyyrrzz: naming: `public static string GetMd5HashShort(List<string> items)`. Feels it more suitable in `HashUtility`

src/docfx/lib/HashUtility.cs
```diff
@@ -54,6 +55,16 @@ public static string GetFileSha1Hash(string fileName)
...
+        {
+            if (monikers == null || monikers.Count == 0)
```
live1206: Why do we want to put a `moninker` specific method in `HashUtility`? And do we really need another method for the null check?

PR #3751, opened 2018-12-10
remove metadata start with underline for resource legacy output

928PJY:


src/docfx/build/legacy/files/LegacyResource.cs
```diff
@@ -19,11 +19,23 @@ internal static class LegacyResource
             var legacyManifestOutput = legacyManifestItem.Output;
             var metadata = metadataProvider.GetMetadata(doc);
             metadata = LegacyMetadata.GenerataCommonMetadata(metadata, docset);
+
+            var metadataNeedToBeRemove = new List<string> { "__global" };
+            foreach (var property in metadata)
```
live1206: use `for` loop to remove and iterate at the same time?

src/docfx/build/legacy/manifest/LegacyManifestItem.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
```
live1206: not in use

src/docfx/build/legacy/files/LegacyResource.cs
live1206: not related

928PJY: cannot access jobject key by index

---
In reply to: [240453204](https://github.com/dotnet/docfx/pull/3751#discussion_r240453204) [](ancestors = 240453204)

PR #3756, opened 2018-12-11
 update with etag

superyyrrzz:
#3718

### DIFF
--- a/src/docfx/lib/HttpClientUtility.cs
+++ b/src/docfx/lib/HttpClientUtility.cs
@@ -11,49 +11,42 @@ namespace Microsoft.Docs.Build
 {
     internal static class HttpClientUtility
     {
-        private const int RetryCount = 3;
-        private const int RetryInterval = 1000;
         private static readonly HttpClient s_httpClient = new HttpClient();
+        private static readonly Type[] s_exceptionsToRetry = new[] { typeof(HttpRequestException), typeof(TimeoutException), typeof(OperationCanceledException) };
 
-        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config, string etag = null)
+        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config, EntityTagHeaderValue etag = null)
         {
-            HttpResponseMessage response;
-            for (var i = 0; i < RetryCount; i++)
-            {
-                try
+            return await RetryUtility.Retry(
+                () =>
                 {
                     // Create new instance of HttpRequestMessage to avoid System.InvalidOperationException:
                     // "The request message was already sent. Cannot send the same request message multiple times."
                     var message = CreateHttpRequestMessage(requestUri, config);
                     message.Method = HttpMethod.Get;
-                    if (!string.IsNullOrEmpty(etag))
+                    if (etag != null)
                     {
-                        message.Headers.IfNoneMatch.Add(new EntityTagHeaderValue(etag));
+                        message.Headers.IfNoneMatch.Add(etag);
                     }
-                    response = await s_httpClient.SendAsync(message);
-                }
-                catch (Exception ex) when (
-                    i < RetryCount - 1 &&
-                    (ex is HttpRequestException ||
-                     ex is TimeoutException ||
-                     ex is OperationCanceledException))
-                {
-                    await Task.Delay(RetryInterval);
-                    continue;
-                }
-                return response;
-            }
-
-            Debug.Fail("should never reach here");
-            return null;
+                    return s_httpClient.SendAsync(message);
+                },
+                s_exceptionsToRetry);
         }
 
-        public static async Task<HttpResponseMessage> PutAsync(string requestUri, HttpContent content, Config config)
+        public static async Task<HttpResponseMessage> PutAsync(string requestUri, HttpContent content, Config config, EntityTagHeaderValue etag = null)
         {
-            var message = CreateHttpRequestMessage(requestUri, config);
-            message.Method = HttpMethod.Put;
-            message.Content = content;
-            return await s_httpClient.SendAsync(message);
+            return await RetryUtility.Retry(
+                () =>
+                {
+                    var message = CreateHttpRequestMessage(requestUri, config);
+                    message.Method = HttpMethod.Put;
+                    message.Content = content;
+                    if (etag != null)
+                    {
+                        message.Headers.IfMatch.Add(etag);
+                    }
+                    return s_httpClient.SendAsync(message);
+                },
+                s_exceptionsToRetry);
         }
 
         private static HttpRequestMessage CreateHttpRequestMessage(string requestUri, Config config)
--- a/src/docfx/lib/github/GitHubUserCache.cs
+++ b/src/docfx/lib/github/GitHubUserCache.cs
@@ -7,7 +7,9 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -39,6 +41,7 @@ namespace Microsoft.Docs.Build
 
         private readonly string _url = null;
         private readonly string _restorePath = null;
+        private readonly EntityTagHeaderValue _etag = null;
         private readonly string _cachePath;
         private readonly double _expirationInHours;
         private bool _updated = false;
@@ -70,6 +73,7 @@ namespace Microsoft.Docs.Build
             {
                 _url = path;
                 _restorePath = restorePath;
+                _etag = RestoreFile.GetEtag(restorePath);
                 _cachePath = AppData.GetGitHubUserCachePath(path);
             }
             else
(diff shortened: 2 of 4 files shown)


src/docfx/lib/HttpClientUtility.cs
```diff
@@ -11,49 +11,42 @@ namespace Microsoft.Docs.Build
...
+        private static readonly Type[] s_exceptionsToRetry = new[] { typeof(HttpRequestException), typeof(TimeoutException), typeof(OperationCanceledException) };
 
-        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config, string etag = null)
+        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config, EntityTagHeaderValue etag = null)
         {
-            HttpResponseMessage response;
-            for (var i = 0; i < RetryCount; i++)
-            {
-                try
+            return await RetryUtility.Retry(
```
yufeih: nit: `RetryUtility.Retry` can take a predicate instead of Type array: `Retry(Func<Task<T>> action, Func<Exception, bool> isExpectedException)`

yufeih: Is 5s retry interval too long?

src/docfx/lib/HttpClientUtility.cs
```diff
@@ -11,49 +11,42 @@ namespace Microsoft.Docs.Build
...
         }
 
-        public static async Task<HttpResponseMessage> PutAsync(string requestUri, HttpContent content, Config config)
+        public static async Task<HttpResponseMessage> PutAsync(string requestUri, HttpContent content, Config config, EntityTagHeaderValue etag = null)
         {
-            var message = CreateHttpRequestMessage(requestUri, config);
-            message.Method = HttpMethod.Put;
-            message.Content = content;
-            return await s_httpClient.SendAsync(message);
+            return await RetryUtility.Retry(
```
yufeih: for write operations, it is safer to not retry blindly in case of failure.

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -167,9 +176,16 @@ public async Task<Error> SaveChanges(Config config)
...
+                var response = await HttpClientUtility.PutAsync(_url, new StringContent(file), config, etag);
+                if (response.StatusCode != HttpStatusCode.PreconditionFailed)
+                {
+                    return null;
+                }
+
+                var (restorePath, newEtag) = await RestoreFile.Restore(_url, config);
+                Debug.Assert(etag != newEtag);
+                await ReadCacheFile(restorePath);
+                return await SaveChangesCore(config, etag);
```
yufeih: add a max retry count here? a loop may be simpler to read than recursion?

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -70,6 +73,7 @@ private GitHubUserCache(Docset docset, string token)
             {
                 _url = path;
                 _restorePath = restorePath;
+                _etag = RestoreFile.GetEtag(restorePath);
```
yufeih: Change `GetEtag` method to accept full path, right now it accepts file name.

src/docfx/lib/HttpClientUtility.cs
superyyrrzz: yes... mistook it in the last PR, and will change back to 1s as it is before

src/docfx/lib/RetryUtility.cs
```diff
@@ -22,8 +22,7 @@ public static async Task<T> Retry<T>(Func<Task<T>> action, IEnumerable<Type> exc
                 {
                     return await action();
                 }
-                catch (Exception ex)
-                when (exceptions == null || exceptions.Any(e => e.IsInstanceOfType(ex)))
+                catch (Exception ex) when (isExpectedException == null || isExpectedException(ex))
```
yufeih: nit: `isExpectedException` should always be non null, so there is no need to do the null check here.

src/docfx/restore/RestoreFile.cs
```diff
@@ -76,11 +76,11 @@ public static string GetRestoreFileName(string hash, EntityTagHeaderValue etag =
         /// Get ETag from restore file name
         /// </summary>
         /// <returns>ETag of the resource, null if server doesn't specify ETag</returns>
-        public static EntityTagHeaderValue GetEtag(string restoreFileName)
+        public static EntityTagHeaderValue GetEtag(string restorePath)
```
yufeih: `RestoreFile.Restore` calls `GetEtag`, you might want to update that as well: `var existingEtag = GetEtag(existingPath)`.

src/docfx/lib/HttpClientUtility.cs
```diff
@@ -11,49 +11,36 @@ namespace Microsoft.Docs.Build
 {
     internal static class HttpClientUtility
     {
-        private const int RetryCount = 3;
-        private const int RetryInterval = 1000;
         private static readonly HttpClient s_httpClient = new HttpClient();
 
-        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config, string etag = null)
+        public static async Task<HttpResponseMessage> GetAsync(string requestUri, Config config, EntityTagHeaderValue etag = null)
```
yufeih: why the change from `string` to `EntityTagHeaderValue`?

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -81,7 +85,7 @@ private GitHubUserCache(Docset docset, string token)
         public static async Task<GitHubUserCache> Create(Docset docset, string token)
         {
             var result = new GitHubUserCache(docset, token);
-            await result.ReadCacheFile();
+            await result.ReadCacheFiles();
             return result;
         }
```
yufeih: https://github.com/dotnet/docfx/blob/af768136d161398b3f6c909e368328df6569adad/src/docfx/lib/github/GitHubUserCache.cs#L71

We may treat url and file path the same, `GetFileRestorePath` does not need a output `(fromUrl)`

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -150,6 +154,19 @@ public async Task<Error> SaveChanges(Config config)
                 return null;
             }
 
+            var remainingRetries = 30;
+            var (error, response) = await SaveChangesCore(config, _etag);
+            while (error == null && remainingRetries-- > 0 && response.StatusCode == HttpStatusCode.PreconditionFailed)
+            {
+                var (restorePath, etag) = await RestoreFile.Restore(_url, config);
```
yufeih: Instead of changing restore, can this line call `Http.Get` directly and do the merge in memory?

src/docfx/lib/HttpClientUtility.cs
superyyrrzz: For `PutAsync`, I can avoid `EntityTagHeaderValue`=>`stirng`=>`EntityTagHeaderValue` conversion between Get and Put. To keep consistent, also changed `GetAsync`.

src/docfx/lib/github/GitHubUserCache.cs
superyyrrzz: Want to write to GitHubUserCache directly if it points to a local file, then keep `fromUrl` can save some duplicated call to `HrefUtility.IsHttpHref`.

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -150,6 +154,27 @@ public async Task<Error> SaveChanges(Config config)
...
 
+            var remainingRetries = 30;
+            var (error, response) = await SaveChangesCore(config, _etag);
+            while (error == null && remainingRetries-- > 0 && response.StatusCode == HttpStatusCode.PreconditionFailed)
+            {
+                try
+                {
+                    response = await HttpClientUtility.GetAsync(_url, config);
+                }
+                catch (HttpRequestException ex)
```
yufeih: can this and [here](https://github.com/dotnet/docfx/blob/v3/src/docfx/restore/RestoreFile.cs#L108) be `catch (Exception)` because `GetAsync` may throw other exception types as well.

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -150,6 +154,27 @@ public async Task<Error> SaveChanges(Config config)
                 return null;
             }
 
+            var remainingRetries = 30;
```
yufeih: Is `30` too big? Upload failure is not catastrophic, it does not affect output correctness. We typically only retry 3 times for get operations, maybe just retry 3 times here. Let's say when the server is buzy and each request timed out. Default HttpClient timeout is 100s, it will take at least 3000s (~1 hour) for this operation to finish.

src/docfx/lib/HttpClientUtility.cs
```diff
@@ -11,49 +11,36 @@ namespace Microsoft.Docs.Build
...
-                    continue;
-                }
-                return response;
-            }
-
-            Debug.Fail("should never reach here");
-            return null;
+                    return s_httpClient.SendAsync(message);
+                },
+                IsExceptionsToRetry);
```
live1206: Put it inline here since it is not long?  Also, exceptions are plural, naming is wrong

PR #3757, opened 2018-12-11
Remove faulty warning message.

jeahyoun:
Warning message for self-closing triple colon block was always generated due to the way that the slicing worked. Additionally, the line numbers were incorrect.

This PR removes the entire warning message.

A future readdition could peek at the next characters in the slice to see if they match the ending (similar to MatchStart, but without moving the slice pointer) and then this could be re-added.

References https://ceapex.visualstudio.com/Engineering/_workitems/edit/36637

I've also added a "Closed" property to the TripleColonBlock to bring it in line with the Moniker Range block in order to unblock validation.

References https://ceapex.visualstudio.com/Engineering/_workitems/edit/44737.


jeahyoun:
> @jeahyoun Could you replace the indentation with 4 spaces?

I always forget about this - sorry! This has been updated.

test/Microsoft.DocAsCode.MarkdigEngine.Tests/TripleColonTest.cs
```diff
@@ -25,5 +27,57 @@ public void TripleColonTestGeneral()
 
             TestUtility.AssertEqual(expected, source, TestUtility.MarkupWithoutSourceInfo);
         }
+
+        [Fact]
+        public void TripleColonTestNested()
```
superyyrrzz: This test have no `Assert`. Do we need this one for now? Or add it when the feature is ready?

yishengjin1413: @jeahyoun Please add some valid test cases for these two changes.

jeahyoun: I'll check into this - this code was commented out when I found it. I'll ask and see if it's necessary.

jeahyoun: I've removed the unused test; it appears that even if it did have a valid assertion, since the moniker range block does not use the triplecolon extension, the test would not prove that nesting triple colon blocks works correctly. We only have the zone extension (which does not allow nesting) and the Chromeless Forms extension (which is a self-closing block) so we currently do not have this use case.

I looked into a way to test the "Closed" properties of the blocks, but I don't think it's possible to access the block properties in the tests - it seem that we can only check input and output from Markdig, not the internal state of the processor. If there is a way to do this, I would be interested.

yishengjin1413: Would you try to get the  MarkdownDocument use this method? https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine/MarkdigMarkdownService.cs#L70

I think you can get the block and test "Closed" properties with MarkdownDocument .

jeahyoun: Added a test to check the block conditions for TripleColon.

yufeih:
@yishengjin1413 @jeahyoun what is the status of this PR?

PR #3758, opened 2018-12-12
Resolve xref within link

live1206:
As [V2](https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#different-syntax-of-cross-reference) describes, we should also support `xref` within the link.
```
[link_text](xref:uid_of_another_file)
```
This should be resolved to:
```
<a href="path_of_another_file">link_text</a>
```


docs/designs/xref.md
```diff
@@ -136,6 +136,17 @@ outputs:
 ```
   - User can also define which property to display for the referenced uid
 
+### Refer to uid within markdown link
+In markdown link url, the user can also refer to uid.
+```
+[link_text](xref:uid_of_another_file)
```
928PJY: what if the uid is not existed?

live1206: test added for uid not found

docs/designs/xref.md
```diff
@@ -136,6 +136,17 @@ outputs:
...
+```
+It should be resolved as:
+```
+<a href="path_of_another_file">link_text</a>
+```
+In this case, display text will be ignored unless `link_text` is empty.
```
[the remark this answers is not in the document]
live1206: Updated test that displayProperty will be ignored

docs/specs/xref.yml
```diff
@@ -321,3 +321,36 @@ outputs:
...
+    title: Title from yaml header a
+    fullName: full name
+    uid: a
+    ---
+  docs/b.md: | 
+    Link to [title](xref:a?displayProperty=fullName#bookmark)
+outputs:
+  docs/a.json:
+  docs/b.json: |
+    { "content": "<p>Link to <a href=\"/docs/a#bookmark\">title</a></p>" }
```
superyyrrzz: >/docs/a#bookmark [](start = 39, length = 16)

resolve to relative path for intermal xref?

src/docfx/build/dependency/Resolve.cs
```diff
@@ -79,24 +81,33 @@ public static (Error error, string content, Document file) TryResolveContent(thi
...
+            var errors = new List<Error>();
             var (error, file, redirectTo, query, fragment, isSelfBookmark, _) = TryResolveFile(relativeTo, href);
+            errors.AddIfNotNull(error);
+
+            if (href.StartsWith("xref:"))
+            {
+                var (uidError, uidHref, _) = TryResolveXref(href.Substring("xref:".Length), xrefMap, dependencyMapBuilder, relativeTo);
+                errors.AddIfNotNull(uidError);
+                return (errors, uidHref, fragment, file);
+            }
```
superyyrrzz: resolve file here? `else { TryResolveFile() }`

docs/specs/xref.yml
live1206: I will fix this in another PR

src/docfx/build/dependency/Resolve.cs
live1206: You are right, this file is not the one we should return while resolving xref

src/docfx/build/dependency/Resolve.cs
```diff
@@ -79,24 +81,33 @@ public static (Error error, string content, Document file) TryResolveContent(thi
...
 
-        public static (Error error, string href, string fragment, Document file) TryResolveHref(this Document relativeTo, string href, Document resultRelativeTo)
+        public static (List<Error> errors, string href, string fragment, Document file) TryResolveHref(this Document relativeTo, string href, Document resultRelativeTo, XrefMap xrefMap = null, DependencyMapBuilder dependencyMapBuilder = null)
         {
             Debug.Assert(resultRelativeTo != null);
+            var errors = new List<Error>();
+
+            if (href.StartsWith("xref:"))
+            {
+                var (uidError, uidHref, _) = TryResolveXref(href.Substring("xref:".Length), xrefMap, dependencyMapBuilder, relativeTo);
```
yufeih: is one `error` enough?

live1206: Here the only error we need to take care of is `uid-not-found`

src/Microsoft.Docs.Template/Microsoft.Docs.Template.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Razor">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.0</TargetFramework>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
yufeih: revert this

PR #3762, opened 2018-12-13
Resolve internal xref to relative URL

live1206:
While referring to uid within the same repository, we should use relative URL


src/docfx/build/xref/XrefMap.cs
```diff
@@ -28,61 +28,75 @@ public IEnumerable<XrefSpec> InternalReferences
...
+                var (internalSpec, referencedFile) = GetInternalSpec(uid, moniker, internalSpecs);
+                if (internalSpec is null)
                     return default;
 
-                if (!string.IsNullOrEmpty(moniker))
+                var relativePath = PathUtility.GetRelativePathToFile(file.SiteUrl, internalSpec.Href);
+                var relativeUrl = HrefUtility.EscapeUrl(Document.PathToRelativeUrl(
+                    relativePath, file.ContentType, file.Schema, file.Docset.Config.Output.Json));
+                var spec = new XrefSpec { Uid = uid, Href = relativeUrl, Monikers = internalSpec.Monikers };
+                spec.ExtensionData.Merge(internalSpec.ExtensionData);
```
superyyrrzz: What about move to `XrefSpec.Clone()`?

src/docfx/build/xref/XrefMap.cs
```diff
@@ -26,63 +26,73 @@ public IEnumerable<XrefSpec> InternalReferences
...
+            }
+
+            if (_externalXrefMap.TryGetValue(uid, out var externalSpec))
+            {
+                return (externalSpec, null);
+            }
+            return default;
+        }
+
+        public static string GetRelativeUrlForXrefReference(Document referencedFile, Document fileRelativeTo)
```
928PJY: >GetRelativeUrlForXrefReference [](start = 29, length = 30)

should this logic be a general method?

src/docfx/build/dependency/Resolve.cs
```diff
@@ -46,14 +46,15 @@ public static string GetLink(string path, object relativeTo, object resultRelati
...
 
-        public static XrefSpec ResolveXref(string uid, XrefMap xrefMap, Document file, DependencyMapBuilder dependencyMapBuilder, string moniker = null)
+        public static (XrefSpec spec, Document referencedFile) ResolveXref(string uid, XrefMap xrefMap, Document file, DependencyMapBuilder dependencyMapBuilder, string moniker = null)
         {
             if (xrefMap is null)
-                return null;
+                return default;
+
+            var (xrefSpec, referencedFile) = xrefMap.Resolve(uid, moniker);
+            dependencyMapBuilder.AddDependencyItem(file, referencedFile, DependencyType.UidInclusion);
```
928PJY: >AddDependencyItem [](start = 33, length = 17)

when the uid is not found, should we add it to the dependency map?

live1206: no, we should not

src/docfx/build/xref/XrefMap.cs
live1206: yes, it can be

src/docfx/build/dependency/Resolve.cs
```diff
@@ -181,10 +183,21 @@ public static (Error error, string href, string display) TryResolveXref(string h
...
-            href = HrefUtility.MergeHref(xrefSpec.Href, monikerQuery, fragment.Length == 0 ? "" : fragment.Substring(1));
+
+            href = HrefUtility.MergeHref(referencedFile != null ? GetRelativeUrl(file, referencedFile) : xrefSpec.Href, monikerQuery, fragment.Length == 0 ? "" : fragment.Substring(1));
             return (null, href, display);
         }
 
+        /// <summary>
+        /// Get relative url from file to the file relative to
+        /// </summary>
+        public static string GetRelativeUrl(Document fileRelativeTo, Document file)
```
yufeih: nit: private

src/docfx/build/dependency/Resolve.cs
```diff
@@ -46,14 +46,18 @@ public static string GetLink(string path, object relativeTo, object resultRelati
             return link;
         }
 
-        public static XrefSpec ResolveXref(string uid, XrefMap xrefMap, Document file, DependencyMapBuilder dependencyMapBuilder, string moniker = null)
+        public static (XrefSpec spec, Document referencedFile) ResolveXref(string uid, XrefMap xrefMap, Document file, DependencyMapBuilder dependencyMapBuilder, string moniker = null)
```
yufeih: Can this return the already transformed `XrefSpec` so caller don't have to base the link themselves?

live1206: One option is that we create a new instance of `XrefSpec` since we need to output the original one later. Another option is the consumer rebase the link themselves, we want to minimize the memory, so prefer to just return it here.

yufeih: The contract does not hold if [Xref] resolves to an `XrefSpec` instead of `href` in the future.

src/docfx/build/dependency/Resolve.cs
live1206: it is used within `AttributeTransformer` as well since we rebase the link in the end

PR #3764, opened 2018-12-13
Remove op.common.js, it has been moved to template

yufeih:


mingwli:
does [this Javascript.cs#L67](https://github.com/dotnet/docfx/blob/v3/src/docfx/build/legacy/template/Javascript.cs#L67) need to be changed?

Issue #3720, opened 2018-12-04
Big variables like GitCommitProvider are not released after build


OsmondJiang:
cc @live1206

yufeih:
This is not related to `ParallelUtility`, changing `Parallel` to sync does not affect GC result.

yufeih:
Should be related to https://github.com/dotnet/roslyn/issues/437, local variables are promoted to heap allocated state machine and are never eligible for garbage collection.

yufeih:
Close this because there is no leak, and it does not affect build performance at the time being,

PR #3767, opened 2018-12-14
Consolidate resolve logic to DependencyResolver

yufeih:
This PR changes the static `Resolve` class to an instance `DependencyResolver`. It exposes 3 methods: `ResolveContent`, `ResolveLink` and `ResolveXref`. Under the hood, it manages dependency map, git history fallback, bookmark validation, etc.


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -11,52 +10,75 @@
 
 namespace Microsoft.Docs.Build
 {
-    internal static class Resolve
+    internal class DependencyResolver
     {
-        public static (string content, object file) ReadFile(string path, object relativeTo, List<Error> errors, DependencyMapBuilder dependencyMapBuilder, GitCommitProvider gitCommitProvider)
-        {
-            Debug.Assert(relativeTo is Document);
+        public readonly BookmarkValidator BookmarkValidator = new BookmarkValidator();
```
live1206: Why do we prefer public read-only field to read-only property?

src/docfx/build/Build.cs
```diff
@@ -27,49 +27,56 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
             var docset = new Docset(context, docsetPath, config, options).GetBuildDocset();
             var monikersProvider = new MonikersProvider(docset);
 
-            // TODO: toc map and xref map should always use source docset?
-            var tocMap = BuildTableOfContents.BuildTocMap(context, docset, monikersProvider);
-            var xrefMap = XrefMap.Create(context, docset, metadataProvider, monikersProvider);
+            using (var gitCommitProvider = new GitCommitProvider())
```
live1206: Will this cause more memory consumption since the scope of `gitCommitProvider` is bigger? @OsmondJiang had this issue before

yufeih: That one is tracked by #3720

src/docfx/build/dependency/DependencyResolver.cs
yufeih: both are fine as long as they readonly.

src/docfx/build/Build.cs
OsmondJiang: no, actually original code, convert to legacy will not have any memory presses

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -48,7 +45,7 @@ internal static class BuildTableOfContents
             return (errors, model, metadata.Monikers);
         }
 
-        public static TableOfContentsMap BuildTocMap(Context context, Docset docset, MonikersProvider monikersProvider)
+        public static TableOfContentsMap BuildTocMap(Context context, Docset docset, MonikersProvider monikersProvider, DependencyResolver dependencyResolver)
```
OsmondJiang: for loc, we need excludes the source dependency map as we talked offline

PR #3771, opened 2018-12-17
Cyclic xref reference

live1206:
Throw when cyclic xref reference detected.


docs/specs/xref.yml
```diff
@@ -390,3 +390,135 @@ outputs:
...
+      "summary": "Link to @a?displayProperty=summary",
+    }
+outputs:
+  docs/b.json: |
+    { "content":{"name":"name for b","description":"description for b"} }
+  build.log: |
+    ["error","circular-reference","Found circular reference","docs/a.md"]
+    ["error","circular-reference","Found circular reference"]
+
+
```
OsmondJiang: remove empty lines

docs/specs/xref.yml
```diff
@@ -390,3 +390,135 @@ outputs:
...
+    {
+      "$schema": "https://raw.githubusercontent.com/dotnet/docfx/v3/schemas/TestData.json",
+      "uid": "b",
+      "name": "name for b",
+      "description": "description for b",
+      "summary": "Link to @a?displayProperty=summary",
+    }
+outputs:
+  docs/b.json: |
+    { "content":{"name":"name for b","description":"description for b"} }
```
OsmondJiang: >{ "content":{"name":"name for b","description":"description for b"} } [](start = 4, length = 69)

why b can be build successfully?

docs/specs/xref.yml
```diff
@@ -390,3 +390,135 @@ outputs:
...
+  build.log: |
+    ["error","circular-reference","Found circular reference","docs/a.md"]
+    ["error","circular-reference","Found circular reference"]
+
+
+---
+# SDP Cyclic xref reference 1
```
OsmondJiang: > SDP Cyclic xref reference  [](start = 1, length = 27)

add more info to description instead of 1/2/3, can add it in another line

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -48,7 +49,7 @@ public DependencyResolver(GitCommitProvider gitCommitProvider, Lazy<XrefMap> xre
             return (error, link, file);
         }
 
-        public (Error error, string href, string display, Document file) ResolveXref(string href, Document relativeTo)
+        public (Error error, string href, string display, Document file) ResolveXref(string href, Document relativeTo, List<Document> callStack)
```
OsmondJiang: >callStack [](start = 134, length = 9)

make it a default value `null` and assign it a value if it's not set, so the caller no need change code anymore?

docs/specs/xref.yml
live1206: Because while we build xref map, for `.md` YAML header we do not build all xref properties, just name and title.
https://github.com/dotnet/docfx/blob/258d0969ad3bd473b9536cc7ef73b07cf024bc14/src/docfx/build/xref/XrefMap.cs#L301

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -60,13 +61,20 @@ public DependencyResolver(GitCommitProvider gitCommitProvider, Lazy<XrefMap> xre
             }
 
             // need to url decode uid from input content
-            var (xrefSpec, referencedFile) = _xrefMap.Value.Resolve(HttpUtility.UrlDecode(uid), moniker);
+            var (xrefSpec, referencedFile, referencedCallStack) = _xrefMap.Value.Resolve(HttpUtility.UrlDecode(uid), relativeTo, moniker);
```
OsmondJiang: >Resolve [](start = 81, length = 7)

1. try add ref to callstack here and handle conflicts
2. then pass to resolve
3. pop current ref

live1206: The problem here is that, while we build xref map we will create a lazy value of `xrefSpec`, in this callback we will call `ResolveXref`, we do not have `callStack` ready yet. So I prepared `callStack` along with build file, and return another `callStack` when uid resolved.

docs/specs/xref.yml
```diff
@@ -390,3 +390,123 @@ outputs:
...
+    summary: Link to @b?displayProperty=summary
+    ---
+    Link to @b?displayProperty=summary
+  docs/b.json: | 
+    {
+      "$schema": "https://raw.githubusercontent.com/dotnet/docfx/v3/schemas/TestData.json",
+      "uid": "b",
+      "name": "name for b",
+      "description": "description for b",
+      "summary": "Link to @a?displayProperty=summary",
```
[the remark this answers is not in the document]
live1206: It can be removed from the test since `summary` inside YAML header is not in use anyway, the content part would cause the circular reference.

src/docfx/build/xref/XRefSpec.cs
```diff
@@ -20,10 +21,24 @@ public class XrefSpec
         [JsonExtensionData]
         public JObject ExtensionData { get; } = new JObject();
```
yufeih: This class feels like a mix of external xrefspec and internal xrefspec, may be we could separate the two.

docs/specs/xref.yml
```diff
@@ -390,3 +390,123 @@ outputs:
...
+    {
+      "content": 
+      {
+        "xref": "c.json"
+      }
+    }
+  build.log: |
+    ["error","circular-reference","Found circular reference","docs/b.json"]
+    ["error","circular-reference","Found circular reference","docs/a.json"]
+    ["error","circular-reference","Found circular reference"]
```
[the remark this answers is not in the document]
live1206: The last one is caused by `OutputXrefMap` since it does not have a reference file, it is treated as a different error. This could be solved once we have the dependency chain. The only way I can think of to get it now is to add a call stack to the building file. Can we do it in another PR since this one is getting big and long-living

src/docfx/build/xref/XRefSpec.cs
live1206: Sure, we can think about splitting them, but maybe in another PR

src/docfx/build/page/AttributeTransformer.cs
```diff
@@ -4,78 +4,141 @@
...
+                extensionData[jsonPath] = new Lazy<Func<string, string, Document, Document, JValue>>(() => Load);
+                return null;
 
-                if (extensionData != null && attributes.Any(attr => attr is XrefPropertyAttribute))
+                JValue Load(string propertyName, string uid, Document referencedFile, Document rootFile)
                 {
-                    extensionData[jsonPath] = new JValue(result);
+                    try
+                    {
+                        if (rootFile != null)
```
yufeih: when is `rootFile` null?

src/docfx/build/page/AttributeTransformer.cs
```diff
@@ -4,78 +4,141 @@
...
+            public UidPropertyReference(string uid, string propertyName, Document referencedFile, Document rootFile)
+            {
+                Uid = uid;
+                PropertyName = propertyName;
+                ReferencedFile = referencedFile;
+                RootFile = rootFile;
             }
         }
+
+        private class UidPropertyReferenceComparer : IEqualityComparer<UidPropertyReference>
```
yufeih: nit: tuples have equality comparer out of the box

src/docfx/build/xref/InternalXrefSpec.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Docs.Build
+{
+    internal class InternalXrefSpec : XrefSpec
+    {
+        public new Dictionary<string, Lazy<Func<string, string, Document, Document, JValue>>> ExtensionData { get; } = new Dictionary<string, Lazy<Func<string, string, Document, Document, JValue>>>();
```
yufeih: avoid inheritance, `new XXX ExtensionData` makes it harder to reason about base class logic and deserialization/serialization behavior.

src/docfx/build/page/AttributeTransformer.cs
live1206: No need to check recursion for "Name" property, maybe a flag would be better to identify whether to check recursion or not.

src/docfx/build/xref/InternalXrefSpec.cs
live1206: `InternalXrefSpec` is not used for de-serialization/serialization, but sure to clarify the logic we can rename it to something else.

src/docfx/build/page/AttributeTransformer.cs
live1206: I was wrong, we can add check recursion to "Name" as well

src/docfx/build/page/AttributeTransformer.cs
```diff
@@ -47,19 +48,16 @@ JValue Load(string propertyName, string uid, Document referencedFile, Document r
...
-                            t_recursionDetector.Value.Push(new UidPropertyReference(uid, propertyName, referencedFile, rootFile));
+                            throw Errors.CircularReference(rootFile, t_recursionDetector.Value.Where(x => x.rootFile == rootFile).Select(x => x.referencedFile).ToList()).ToException();
                         }
+                        t_recursionDetector.Value.Push((propertyName, uid, referencedFile, rootFile));
                         return new JValue(TransformContent(context, errors, attribute, value, file, buildChild));
                     }
                     finally
                     {
-                        if (rootFile != null && t_recursionDetector.Value.Count > 0)
+                        if (t_recursionDetector.Value.Count > 0)
```
yufeih: nit: Assert the count must > 0, so if there're some bugs in code, `Pop` will throw and let us know.

src/docfx/build/xref/InternalXrefSpec.cs
```diff
@@ -0,0 +1,80 @@
...
+        public Dictionary<string, Lazy<JValue>> ExtensionData { get; } = new Dictionary<string, Lazy<JValue>>();
+
+        public string GetXrefPropertyValue(string propertyName, Document rootFile)
+        {
+            if (propertyName is null)
+                return null;
+
+            try
+            {
+                if (t_recursionDetector.Value.Contains((propertyName, Uid, ReferencedFile, rootFile)))
```
OsmondJiang: consider A -> A

src/docfx/build/xref/XrefMap.cs
```diff
@@ -27,39 +27,79 @@ public IEnumerable<XrefSpec> InternalReferences
                 {
                     if (TryGetValidXrefSpecs(uid, specsWithSameUid, out var validInternalSpecs))
                     {
-                        loadedInternalSpecs.Add(GetLatestInternalXrefMap(validInternalSpecs).spec);
+                        var (internalSpec, referencedFile) = GetLatestInternalXrefMap(validInternalSpecs);
```
yufeih: var (uid, (specs, file))

src/docfx/build/xref/XrefMap.cs
```diff
@@ -161,13 +208,14 @@ private bool TryGetValidXrefSpecs(string uid, List<Lazy<(List<Error> errors, Xre
...
             {
-                if (value is null)
+                if (input.value is null)
                     return default;
 
-                var isValueCreated = value.IsValueCreated;
-                var (errors, spec, doc) = value.Value;
+                var isValueCreated = input.value.IsValueCreated;
+
+                var (errors, spec) = input.value.Value;
```
yufeih: catch exception here

src/docfx/lib/CollectionUtility.cs
```diff
@@ -15,5 +15,13 @@ internal static class CollectionUtility
                 list.Add(value);
             }
         }
+
+        public static void AddRange<TKey, TValue>(this IDictionary<TKey, TValue> dictionary, IDictionary<TKey, TValue> range)
+        {
+            foreach (var (key, value) in range)
+            {
+                dictionary[key] = value;
```
yufeih: use `Add`

src/docfx/build/xref/XrefMap.cs
live1206: The second element is another list of tuples, can not just load it like a tuple

PR #3772, opened 2018-12-17
Do not mix source and loc dependencies in dependency map

yufeih:


src/docfx/build/dependency/DependencyMapBuilder.cs
```diff
@@ -18,7 +18,14 @@ public void AddDependencyItem(Document relativeTo, Document dependencyDoc, Depen
...
             Debug.Assert(relativeTo != null);
 
-            if (dependencyDoc == null)
+            if (relativeTo == null || dependencyDoc == null)
+            {
+                return;
+            }
+
+            // Do not mix source document in dependency map for localized build
+            var isLocalizedBuild = relativeTo.Docset.IsLocalizedBuild() || dependencyDoc.Docset.IsLocalizedBuild();
```
OsmondJiang: >relativeTo [](start = 35, length = 10)

only need check relativeTo

src/docfx/build/dependency/DependencyMapBuilder.cs
```diff
@@ -14,16 +14,22 @@ internal class DependencyMapBuilder
...
         private readonly ConcurrentHashSet<DependencyItem> _dependencyItems = new ConcurrentHashSet<DependencyItem>();
 
-        public void AddDependencyItem(Document relativeTo, Document dependencyDoc, DependencyType type)
+        public void AddDependencyItem(Document from, Document to, DependencyType type)
         {
-            Debug.Assert(relativeTo != null);
+            Debug.Assert(from != null);
 
-            if (dependencyDoc == null)
+            if (from == null || to == null)
```
OsmondJiang: from is asserted

PR #3773, opened 2018-12-17
Use HrefUtility.MergeHref over string concatenation.

yufeih:
- Make `HrefUtility.SplitHref` return values without `#`, `?`
- Use `HrefUtility.MergeHref` over `path + query + fragment`


test/docfx.Test/lib/HrefUtilityTest.cs
```diff
@@ -10,13 +10,13 @@ public static class HrefUtilityTest
...
-        [InlineData("a?b#c?d=e", "a", "?b", "#c?d=e")]
-        [InlineData("a?b#c#d", "a", "?b", "#c#d")]
+        [InlineData("a#b", "a", "", "b")]
+        [InlineData("a#b<", "a", "", "b<")]
+        [InlineData("a?c", "a", "c", "")]
+        [InlineData("a?b#c", "a", "b", "c")]
+        [InlineData("a#b?c=d", "a", "", "b?c=d")]
+        [InlineData("a?b#c?d=e", "a", "b", "c?d=e")]
+        [InlineData("a?b#c#d", "a", "b", "c#d")]
         public static void SplitHref(string href, string path, string query, string fragment)
```
OsmondJiang: A fragment is an internal page reference, sometimes called a named anchor. It usually appears at the end of a URL and **begins with a hash (#) character followed by an identifier**. It refers to a section within a web page

src/docfx/lib/HrefUtility.cs
```diff
@@ -24,11 +26,11 @@ public static (string path, string query, string fragment) SplitHref(string href
             var fragmentIndex = href.IndexOf('#');
             if (fragmentIndex >= 0)
             {
-                fragment = href.Substring(fragmentIndex);
+                fragment = href.Substring(fragmentIndex + 1);
```
OsmondJiang: we shouldn't do that, conflicted with fragment definition

live1206: Yes. I didn't fix it in the first place because of the nasty url concatenation. We can use 'MergeHref' to get rid of it now.

PR #3774, opened 2018-12-17
fix moniker output empty for toc build result

928PJY:


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -48,7 +48,7 @@ public DependencyResolver(GitCommitProvider gitCommitProvider, Lazy<XrefMap> xre
             return (error, link, file);
         }
 
-        public (Error error, string href, string display) ResolveXref(string href, Document file)
+        public (Error error, string href, string display, Document referencedFile) ResolveXref(string href, Document file)
```
yufeih: nit: `referencedFile` -> `file` to be consistent with signature of other methods

Issue #3686, opened 2018-11-27
Move redirection to dedicated files


yufeih:
Remove redirection from config, use dedicated `redirection.yml`. Similar to `TOC.yml`, `redirection.yml` can be put in any directory.

For migration, just generate a sxs `redirection.yml`

#2913

IEvangelist:
Are there any plans for this to happen in the near future? I'd selfishly love to see this become available soon.

Issue #2913, opened 2018-06-21
Redirection with better configuration


OsmondJiang:
- [ ] RedirectTo can be absolute link or file path
- [ ] Decouple document id transfer and redirection
- [ ] Report warnings if the RedirectTo doesn't belong to build scope, just file link resolving

cc @herohua  @yufeih

yufeih:
Per discussion, we could have
- a **renames** section that maps from *file path* to *file path*, preserves document id
- a **redirections** section that maps from *file path* to *url*, does not preserve document id

Need to consider the impact on migration if we did do this.

PR #3775, opened 2018-12-17
Remove RedirectionOutOfScope

yufeih:
Because redirections are append only (immutable), build scope may change after a redirection is added. it is theatrically possible that a valid redirection once was in build scope is currently not in build scope, due to changes to build scope. We need to respect that redirection.

#3686 #2913


OsmondJiang:
We can't make sure the change of build scope which causes redirection invalid is expected or not, so report errors to user to let the user make their own choices(delete redirection config or change build scope), not just eat the errors, which can cause unexpected LSI.

OsmondJiang:
let's do something from requirement

yufeih:
Another redirection issue that might relate: If there is both a file and a redirection entry that shares the same source path, do we _report a `publish-url-conflict` error and don't generate an output_ or _still produce a redirection output_ ? https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/8082?_a=files

PR #3781, opened 2018-12-18
Support folder level redirection.yml

yufeih:
Implements #3686 

This PR does not address the open question left in #3775 , so some of the related tests failed right now.

File path `redirection.yml` in sub folders are relative to the root docset folder. This is to preserve source path when renaming directories.

@mingwli , the migrate tool needs to be changed accordingly to generate a sxs redirection.yml file.


docs/specs/config.yml
```diff
@@ -313,8 +317,26 @@ inputs:
...
+    ["error","redirection-conflict","The 'docs/a.md' appears twice or more in the redirection mappings","redirection.yml"]
+---
+# Folder level redirection.yml or redirection.json files
+inputs:
+  docfx.yml:
+  redirection.yml: |
+    redirections:
+      docs/a.md: /a
+  docs/folder-1/redirection.json: |
+      { "redirections" : { "docs/folder-1/a.md": "/b" } }
```
superyyrrzz: Feels rather confusing that  `docs/folder-1/a.md` is relative to docset root instead of the current folder...

PR #3782, opened 2018-12-18
extract monikers from toc resolve delegate

928PJY:
- Remove monikers from resolve callback
- Move moniker union logic to BuildTableOfContents
- MonikersMap, monikersProvider -> monikerMap, monikerProvider
- Extract a MonikerMap class like others

### DIFF
--- a/src/docfx/build/moniker/MonikerMap.cs
+++ b/src/docfx/build/moniker/MonikerMap.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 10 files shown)


src/docfx/build/moniker/MonikerMap.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// The mappings between document and monikers
+    /// </summary>
+    internal class MonikerMap
+    {
+        private readonly HashSet<Document> _docs = new HashSet<Document>();
+
+        private readonly ConcurrentDictionary<Document, List<string>> _documentToMonikers = new ConcurrentDictionary<Document, List<string>>();
+
+        public bool Contains(Document doc) => _docs.Contains(doc);
```
yufeih: use `_documentToMonikers.ContainsKey` instead?

src/docfx/build/moniker/MonikersProvider.cs
```diff
@@ -8,14 +8,14 @@
...
-    internal class MonikersProvider
+    internal class MonikerProvider
     {
         private readonly List<(Func<string, bool> glob, string monikerRange)> _rules = new List<(Func<string, bool>, string)>();
         private readonly MonikerRangeParser _rangeParser;
 
         public MonikerComparer Comparer { get; }
 
-        public MonikersProvider(Docset docset)
+        public MonikerProvider(Docset docset)
```
OsmondJiang: i am a little confused with MonikerProvider and MonikerMap, seems they are all used to provide monikers, why not combine them together, as i know, all monikers are provided by moniker provider, so moniker provider should already know the mapping

OsmondJiang: as we talked offline, it would be great if you can only have one `monikerResolver` to
1. provider moniker
2. tracing moniker dependency

then we don't need add moniker map everywhere and also don't need return monikers everywhere, just like DependencyResolver

Issue #3288, opened 2018-08-27
Support UID in TOC for SDP pipeline


OsmondJiang:
as title

yufeih:
Track xref/uid support in #3050, this include everything from TOC to markdown to SDP to config. Close as dup.

OsmondJiang:
seems SDP pipeline doesn't cover this case, @yufeih  and @live1206 , i am reopening this issue and do it now. 

Find some example in MicrosoftDocs, seems all reference content toc using uid.

OsmondJiang:
example here: https://github.com/MicrosoftDocs/dynamics365-docs-odata-apis/blob/master/WebAPI-reference/dynamics-ce-odata-9/toc.yml

PR #3783, opened 2018-12-18
support sdp(yaml/json) toc uid reference

OsmondJiang:
yaml/json toc can use `uid` to reference topic files.
in docfx v2, we also support `topicUid`, but after searching the MicrosoftDocs org, there is no one using this new syntax, we in v3, we are not going to support that

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1079,3 +1079,85 @@ outputs:
         { "toc_title":"Title","href":"a#bookmark","monikers":[] }
       ]
     }
+---
+# uid reference
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    ---
+    title: Title from yaml header a
+    uid: a
+    ---
+  docs/TOC.yml: |
+    - name: Uid Ref
+      uid: a
+outputs:
+  docs/a.json:
+  docs/toc.json: |
+    {
+      "items": [
+        { "toc_title":"Uid Ref","href":"a","monikers":[] }
+      ]
+    }
+---
+# uid reference first
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    ---
+    title: Title from yaml header a
+    uid: a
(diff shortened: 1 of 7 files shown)


docs/specs/toc.yml
```diff
@@ -1079,3 +1079,85 @@ outputs:
...
+  docs/a.md: |
+    ---
+    title: Title from yaml header a
+    uid: a
+    ---
+  docs/b.md:
+  docs/TOC.yml: |
+    - name: Uid Ref
+      uid: a
+      href: b.md
```
928PJY: should we report a warning to user that the href will not work?

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -215,8 +217,19 @@ string GetTopicHref(TableOfContentsInputItem tocInputModel)
                 return default;
             }
 
-            (string resolvedTopicHref, List<string> monikers) ProcessTopicHref(string topicHref)
+            (string resolvedTopicHref, List<string> monikers) ProcessTopicItem(string uid, string topicHref)
             {
+                // process uid first
+                if (!string.IsNullOrEmpty(uid))
+                {
+                    var (resolvedTopicHref, monikers) = resolveXref.Invoke(filePath, uid);
```
928PJY: use `rootPath`?

docs/specs/toc.yml
superyyrrzz: Just checked v2 code, v2 does not log a warning in this case, but it will use href if uid-not-found.

src/docfx/build/toc/parser/TableOfContentsParser.cs
OsmondJiang: rootPath can make the error message not correct

OsmondJiang:
#3288

Issue #3149, opened 2018-08-03
azure-docs-pr issue: how to deal with empty value in *.metadata.html


928PJY:

herohua:
@herohua to follow up first on desired behavior

PR #3784, opened 2018-12-18
Only remove nulls from the middle of arrays

yufeih:
#3149

### DIFF
--- a/docs/specs/metadata.yml
+++ b/docs/specs/metadata.yml
@@ -9,6 +9,19 @@ outputs:
   docs/a.json: |
     { "metadata": { "robots": "noindex, nofollow" } }
 ---
+# Preserve null custom metadata
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    ---
+    ms.custom-metadata:
+    ---
+outputs:
+  docs/a.json: |
+    { "metadata": { "ms.custom-metadata": null } }
+  build.log: |
+    ["info","null-value","'ms.custom-metadata' contains null value","docs/a.md","['ms.custom-metadata']",1,1]
+---
 # generate normal document id and version id
 # backward compatible to docs.com
 # document id = md5("Win.sccm|mdt/release-notes.md")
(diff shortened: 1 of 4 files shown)


src/docfx/lib/JsonUtility.cs
```diff
@@ -346,7 +357,7 @@ private static bool IsNullOrUndefined(this JToken token)
                 (token.Type == JTokenType.Undefined);
         }
 
-        private static void FindNulls(this JToken token, List<Error> errors, List<JToken> nullNodes)
+        private static void RemoveNullsCore(JToken token, List<Error> errors, List<JToken> nullNodes, List<JToken> removeNodes)
```
mingwli: >private static void RemoveNullsCore(JToken token, List<Error> errors, Li [](start = 8, length = 72)

can this be a local function?

superyyrrzz: >List<JToken> nullNodes, List<JToken> removeNodes [](start = 78, length = 48)

seems `nullNodes` is always identical to `removeNodes`, merge into one?

yufeih: `nullNodes` contains nulls for object and array, `removeNodes` only contains nulls in arrays.

yufeih: If could be, I was just trying to keep the existing convention here

docs/specs/metadata.yml
```diff
@@ -9,6 +9,19 @@ outputs:
   docs/a.json: |
     { "metadata": { "robots": "noindex, nofollow" } }
 ---
+# Preserve null custom metadata
```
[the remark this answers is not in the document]
yufeih: @live1206 , is there an existing test that covers this?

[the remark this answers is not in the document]
live1206: https://github.com/dotnet/docfx/blob/bc85753222af5147b53e265b1ff77908bdf3f87e/test/docfx.Test/lib/JsonUtilityTest.cs#L225

PR #3788, opened 2018-12-19
Markdown TOC support UID referencing

OsmondJiang:
1. support uid reference in markdown toc
2. fill toc item name using uid display name if it's not set
3. fix bug of root file relative path

### DIFF
--- a/src/docfx/build/toc/parser/MarkdownTocMarkup.cs
+++ b/src/docfx/build/toc/parser/MarkdownTocMarkup.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using Markdig.Extensions.Yaml;
 using Markdig.Syntax;
 using Markdig.Syntax.Inlines;
+using Microsoft.DocAsCode.MarkdigEngine.Extensions;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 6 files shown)


OsmondJiang:
#3288

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -94,43 +95,36 @@ private static (List<TableOfContentsInputItem> children, int count) ConvertTo(st
...
+                {
+                    currentItem.Uid = xrefInline.Href;
+                    return currentItem;
                 }
 
-                string name = null;
-                string displayName = null;
-                string href = null;
                 var link = block.Inline.FirstOrDefault(l => l is LinkInline);
+                var literalLink = block.Inline.FirstOrDefault(l => l is LiteralInline);
```
yufeih: Probably not related to this PR but `FirstOrDefault` might be too permissive, when there are more inline elements, it abandoned the remaining ones without a warning. Same is true for `LinkInline` and `LiteralInline` handling

```
# @b xyz
```

OsmondJiang: yes, i remember we talked about that, will test it later and create an new pr if we need change it.

PR #3794, opened 2018-12-20
add test for multiple inlines in one heading block

OsmondJiang:
also move should build file in one place

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1236,6 +1236,7 @@ outputs:
     {"items":[{"toc_title":"TOC Ref","children":[{"toc_title":"Uid Ref","href":"a","monikers":[]}],"monikers":[]}]}
 ---
 # markdown toc using bad syntax(multiple inlines for one heading block)
+inputs:
   docfx.yml:
   docs/TOC.md: |
     # @b abc
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1238,12 +1238,15 @@ outputs:
 inputs:
   docfx.yml:
   docs/TOC.md: |
+    # [Test <X> Y](a.md)
+    # [Test <X> Y]
     # @b abc
     # @b @a
     # [Title X](a.md) bcd
     # [Title X](a.md) [Title Y](b.md)
     # @b [Title X](a.md)
     # [Title X](a.md) @b
+    # [Title X](xref: a) @b
   docs/a.md: |
     ---
     title: Title A
@@ -1258,12 +1261,15 @@ outputs:
   docs/toc.json: |
     {
       "items": [
+        { "toc_title":"Test <X> Y","href":"a","monikers":[] },
+        { "toc_title":"[Test <X> Y]","monikers":[] },
         { "toc_title":"Title A","href":"a","monikers":[] },
         { "toc_title":"Title A","href":"a","monikers":[] },
         { "toc_title":"Title X","href":"a","monikers":[] },
         { "toc_title":"Title X","href":"a","monikers":[] },
         { "toc_title":"Title A","href":"a","monikers":[] },
-        { "toc_title":"Title A","href":"a","monikers":[] }
+        { "toc_title":"Title A","href":"a","monikers":[] },
+        { "toc_title":"Title A","href":"a","monikers":[] },
       ]
     }
   docs/a.json:
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1001,17 +1001,19 @@ outputs:
   build.log: |
     ["error","missing-toc-head","The toc head name is missing","docs/TOC.md","",1]
 ---
-# Multiple Toc head names, following the below picking up order
-# 1. link inlie display name
-# 1. link inline name
-# 2. literal inline name
+# Multiple leaf inlines in toc title
 inputs:
   docfx.yml:
   docs/TOC.md: |
-    # Title **Title2** [Title3](https://github.com/docfx)
+    # [Test <X> Y](a.md)
+    # [Test <X> Y]
+    # [Test **Title2** Title3](a.md)
+    # [Test **Title2** Title3]
+  docs/a.md:
 outputs:
+  docs/a.json:
   docs/toc.json: |
-    {"items":[{"toc_title":"Title3", "href": "https://github.com/docfx"}]}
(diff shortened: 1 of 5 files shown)


docs/specs/toc.yml
```diff
@@ -1232,3 +1232,39 @@ outputs:
...
   docs/toc.json: |
     {"items":[{"toc_title":"TOC Ref","children":[{"toc_title":"Uid Ref","href":"a","monikers":[]}],"monikers":[]}]}
+---
+# uid reference in markdown toc using bad syntax(multiple inlines for one heading block)
+# the xref inline has highest priority, then we only retrieve the first one
+inputs:
+  docfx.yml:
+  docs/TOC.md: |
+    # @b abc
+    # @b @a
```
mingwli: what about mixed two syntax? [xref](xref:uid) & @uid
``` text
# [xxx](xref:uidA) @uidB
```

OsmondJiang: good question....

docs/specs/toc.yml
```diff
@@ -1232,3 +1232,39 @@ outputs:
   docs/a.json:
   docs/toc.json: |
     {"items":[{"toc_title":"TOC Ref","children":[{"toc_title":"Uid Ref","href":"a","monikers":[]}],"monikers":[]}]}
+---
+# uid reference in markdown toc using bad syntax(multiple inlines for one heading block)
+# the xref inline has highest priority, then we only retrieve the first one
```
yufeih: @superyyrrzz what is the behavior for v3 in this case? I feel like its some kind of error or warning

superyyrrzz: v2 will error if toc.md has wrong syntax.

OsmondJiang: ok, i will treat them as erros

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -110,23 +116,36 @@ TableOfContentsInputItem GetItem(HeadingBlock block)
...
+                string GetName(IEnumerable<LeafInline> literalInlines)
+                {
+                    literalInlines = literalInlines?.Where(l => l != null && !l.Span.IsEmpty);
+                    if (literalInlines == null || !literalInlines.Any())
+                    {
+                        return null;
+                    }
+
+                    var start = literalInlines.First().Span.Start;
+                    var lenght = literalInlines.Last().Span.End - start + 1;
```
mingwli: >lenght [](start = 24, length = 6)

length

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -102,6 +102,12 @@ TableOfContentsInputItem GetItem(HeadingBlock block)
                     return currentItem;
                 }
 
+                if (block.Inline.Count() > 1 && block.Inline.Any(l => l is XrefInline || l is LinkInline))
+                {
+                    errors.Add(Errors.InvalidTocSyntax(new Range(block.Line, block.Column), filePath, tocContent.Substring(block.Span.Start, block.Span.Length), "multiple inlines in one heading block is not allowed"));
```
superyyrrzz: It may be hard for writers to understand terms like `inline` and `heading block`.

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -110,23 +116,36 @@ TableOfContentsInputItem GetItem(HeadingBlock block)
...
-                var literalLink = block.Inline.FirstOrDefault(l => l is LiteralInline);
                 if (link != null && link is LinkInline linkInline)
                 {
                     if (!string.IsNullOrEmpty(linkInline.Url))
                         currentItem.Href = linkInline.Url;
                     if (!string.IsNullOrEmpty(linkInline.Title))
                         currentItem.DisplayName = linkInline.Title;
 
-                    literalLink = linkInline.FirstOrDefault(l => l is LiteralInline);
+                    currentItem.Name = GetName(linkInline.Select(l => l as LeafInline));
```
yufeih: nit: When `GetName` returns `null` here, the `LinkInline` code branch will mix with `LiteralInline` code branch

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -110,23 +116,36 @@ TableOfContentsInputItem GetItem(HeadingBlock block)
...
 
-                if (literalLink != null && literalLink is LiteralInline literalInline)
-                {
-                    currentItem.Name = literalInline.Content.ToString();
-                }
+                currentItem.Name = currentItem.Name ?? GetName(block.Inline.Select(l => l as LeafInline));
 
                 return currentItem;
+
+                string GetName(IEnumerable<LeafInline> literalInlines)
```
yufeih: Is `IEnumerable<Inline>` sufficient. I see `ContainerInline` has several derived types like `DelimiterInline` and `EmphasisInline`, will it affect the behavior here?

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -110,23 +116,36 @@ TableOfContentsInputItem GetItem(HeadingBlock block)
...
+                    literalInlines = literalInlines?.Where(l => l != null && !l.Span.IsEmpty);
+                    if (literalInlines == null || !literalInlines.Any())
+                    {
+                        return null;
+                    }
+
+                    var start = literalInlines.First().Span.Start;
+                    var lenght = literalInlines.Last().Span.End - start + 1;
+
+                    if (start >= 0 && lenght > 0)
```
yufeih: when `start < 0` or `length < 0`, it means there is a bug somewhere in markdig or our code, shall we let docfx crash?

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
OsmondJiang: it's a hint, but the last message is important, referencing to the docs.

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
OsmondJiang: yes, check my first changed test case

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
OsmondJiang: crash is not really user friendly, let's add Debug.Assert here?

PR #3796, opened 2018-12-20
fix perf issue when report toc dependency

superyyrrzz:
In local test, it can reduce time to calculate dependency for not in toc files from 162s to 13s


herohua:
👍

PR #3801, opened 2018-12-20
Transform code language extracted from triple slash comments to stye class

akihikodaki:
This change is just adding `lang-` prefix and rather minimal. Note that this does not normalize the given language name, take `lang` attribute into account, nor add a proper documentation. These will remain undone.


superyyrrzz:
@akihikodaki Thank you for contribution!

PR #3803, opened 2018-12-20
Fix null leak by upgrading Json.Net

yufeih:
Pickup the [fix](https://github.com/JamesNK/Newtonsoft.Json/issues/1719) for `ExtensionData`. 

Fixes https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/7153?_a=overview

#3149


OsmondJiang:
i don't understand the fix here, do we want preserve null for custom input or not? does that fix have impact to metadata scenarios?

PR #3804, opened 2018-12-20
Update _master.tmpl

ChrisChinchilla:
I think the logic in this template is round the wrong way, i.e. `"_disableToc": false` hides the toc, reversing the logic fixes it.

### DIFF
--- a/src/docfx.website.themes/default/layout/_master.tmpl
+++ b/src/docfx.website.themes/default/layout/_master.tmpl
@@ -24,11 +24,11 @@
       </div>
       {{/_enableSearch}}
       <div role="main" class="container body-content hide-when-search">
-      {{^_disableToc}}
+      {{#_disableToc}}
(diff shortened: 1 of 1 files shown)


superyyrrzz:
Hi @ChrisChinchilla , thanks for your PR.

I think the original code is correct. `{{^_disableToc}}` means when `_disableToc` is false, it will render this TOC. TOC show by default as `false` is the default value.

ChrisChinchilla:
@superyyrrzz Yes, I know, but that's not what happens with this template file right now, so maybe the problem is something different. Changing my local copy made it work as expected.

superyyrrzz:
Maybe a detailed reproducing step can help us figure out the root cause. You can create an issue for it. 

However, What this PR changed seems not to make sense to me. The PR check also failed as it breaks the page layout. I will first close this PR.

PR #3805, opened 2018-12-21
update System.Net.Http to 4.3.2

superyyrrzz:
Microsoft Security Advisory 4021279: https://github.com/aspnet/Announcements/issues/239


herohua:
Please also update all dependencies in service side. Thanks!

Issue #3795, opened 2018-12-20
Support feature branches


yufeih:
To implement our new release process #3790:
- [x] Support feature branches in PR validation 
- [x] Support feature branches in CI
- [x] Create a sandbox MyGet feed
- [x] Implement feature branch versioning
- [x] Consume the latest package in OPS sandbox environment
- [x] ~~Deploy PR to sandbox MyGet feed? (cc @OsmondJiang)~~

@mingwli @fenxu

yufeih:
done

PR #3807, opened 2018-12-21
pack version

mingwli:
update pack version logic

only deploy when it's a deployable branch?
set package version based on branch

#3795

### DIFF
--- a/.deploy.yml
+++ b/.deploy.yml
@@ -15,6 +15,10 @@ steps:
   displayName: Build
 
 - task: NuGetCommand@2
+  condition: >-
+    and(succeeded(), or(
+      eq(variables['Build.SourceBranch'], 'refs/heads/v3'),
+      startsWith(variables['Build.SourceBranch'], 'feature/')))
   displayName: Push to Sandbox MyGet Feed
   inputs:
     command: push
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -6,14 +6,26 @@ function exec([string] $cmd) {
...
-$version = "3.0.0-alpha-$revision-$commitSha"
 
-Remove-Item ./drop -Force -Recurse -ErrorAction Ignore
+if ($branch -eq "v3") {
+    $version = "3.0.0-beta-$revision-$commitSha"
+} elseif ($branch.StartsWith($featureBranchPrefix)) {
+    $feature = $branch.SubString($featureBranchPrefix.length)
+    $version = "3.0.0-alpha-$feature-$revision-$commitSha"
+} else {
+    exit 0
```
yufeih: use `3.0.0-alpha-$revision-$sha` for other branches to we could test pack locally on other branches?

.deploy.yml
```diff
@@ -15,6 +15,10 @@ steps:
   displayName: Build
 
 - task: NuGetCommand@2
+  condition: >-
```
yufeih: The `trigger` section already restrict the branch to `v3` or `feature/*`, this is not needed

PR #3811, opened 2018-12-24
always return default config for Config.TryLoad 

OsmondJiang:
and also add TryFindSourceRepo convention
No need LoadIfExists anymore
Some minor restore changes to make it more simple

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -150,7 +150,7 @@ namespace Microsoft.Docs.Build
         /// Gets the config file name.
         /// </summary>
         [JsonIgnore]
-        public string ConfigFileName { get; private set; }
+        public string ConfigFileName { get; private set; } = "docfx.yml";
 
         public IEnumerable<string> GetFileReferences()
         {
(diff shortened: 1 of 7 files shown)


test/docfx.Test/build/ConfigTest.cs
```diff
@@ -85,5 +85,26 @@ public static void LocConfigConventionTheme(string theme, string locale, string
...
+        [InlineData("", "master", null, null, null)]
+        [InlineData("", null, null, null, null)]
+        [InlineData("https://github.com/docs.zh-cn", "master", "https://github.com/docs", "master", "zh-cn")]
+        [InlineData("https://github.com/docs.zh-CN", "master", "https://github.com/docs", "master", "zh-cn")]
+        [InlineData("https://github.com/docs.bs-Cyrl-BA", "master", "https://github.com/docs", "master", "bs-cyrl-ba")]
+        [InlineData("https://test.visualstudio.com/_git/abc", "master", null, null, null)]
+        [InlineData("https://test.visualstudio.com/_git/abc.zh-cn", "master", "https://test.visualstudio.com/_git/abc", "master", "zh-cn")]
+        [InlineData("https://test.visualstudio.com/_git/abc.bs-Cyrl-BA", "master", "https://test.visualstudio.com/_git/abc", "master", "bs-cyrl-ba")]
+        [InlineData("https://github.com/docs.zh-cn", "master-sxs", "https://github.com/docs", "master", "zh-cn")]
+        [InlineData("https://github.com/docs.loc", "master-sxs", "https://github.com/docs", "master", "loc")]
```
mingwli: >[InlineData("https://github.com/docs.loc", "master-sxs", "https://github.com/docs", "master", "loc")] [](start = 8, length = 101)

is this branch mapping ?
{source-repo-name} -> {source-repo-name}.loc + {source-branch} -> {source-branch}.{locale}
e.g. 
```text
loc-branch       ---->      source-branch     locale
master-sxs.zh-cn           master                   zh-cn
```

OsmondJiang: 赞！ changed

src/docfx/config/Config.cs
```diff
@@ -168,81 +168,62 @@ public IEnumerable<string> GetFileReferences()
...
             var deserializeErrors = new List<Error>();
             (deserializeErrors, config) = JsonUtility.ToObjectWithSchemaValidation<Config>(finalConfigObject);
             errors.AddRange(deserializeErrors);
 
+            // validate metadata
             errors.AddRange(MetadataValidator.Validate(config.GlobalMetadata, nameof(GlobalMetadata)));
             errors.AddRange(MetadataValidator.Validate(config.FileMetadata, nameof(FileMetadata)));
 
+            config.ConfigFileName = configPath == null
+                ? config.ConfigFileName
```
superyyrrzz: >ConfigFileName [](start = 25, length = 14)

when configPath is null, config.ConfigFileName should be null or empty, instead of using default value `docfx.yml` that does not exist

src/docfx/restore/Restore.cs
```diff
@@ -23,52 +23,51 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
...
                 }
             }
-        }
 
-        private static async Task RestoreOneDocset(
-            Report report,
-            string docsetPath,
-            CommandLineOptions options,
+            async Task RestoreOneDocset(
+            string docset,
```
superyyrrzz: >string docset, [](start = 12, length = 14)

minor: indent it. Personally prefer to keep it a private method as the logic is not quite simple.

src/docfx/config/Config.cs
OsmondJiang: it's just used to fill the error messages, when the config is empty, there should be no errors for configurations

OsmondJiang: add @live1206  for double confirm

OsmondJiang: double checked, yes

PR #3813, opened 2018-12-24
consolidate existing CIs

mingwli:
seems yaml syntax is different when it's not in `condition` expression

🔁 🚗 💥

### DIFF
--- a/.deploy.yml
+++ b/.deploy.yml
@@ -24,6 +24,7 @@ jobs:
   condition: eq(variables['Build.Reason'], 'PullRequest')
   steps:
     - powershell: ./build.ps1
+    # todo: conditional PR deploy, check myget limits
 # deployment job(impact, build, pack, push package)
 - job: Deploy(private agent DocFX)
   pool:
--- a/.deploy.yml
+++ b/.deploy.yml
@@ -12,8 +12,6 @@ steps:
   displayName: Impact Test
 
 - powershell: ./build.ps1
-  env:
-    SOURCE_BRANCH: $(variables['Build.SourceBranch'])
   displayName: Build
 
 - task: NuGetCommand@2
--- a/.deploy.yml
+++ b/.deploy.yml
@@ -1,32 +1,58 @@
 trigger:
 - v3
 - feature/*
-queue:
-  name: DocFX
-steps:
-- powershell: ./impact.ps1
-  env:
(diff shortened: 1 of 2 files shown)


build.ps1
```diff
@@ -12,21 +12,26 @@ exec "dotnet test test\docfx.Test"
...
+if ($branch -eq "refs/heads/v3") {
+    # CI triggered by v3
     $version = "3.0.0-beta-$revision-$commitSha"
 } elseif ($branch.StartsWith($featureBranchPrefix)) {
+    # CI triggered by feature/*
     $feature = $branch.SubString($featureBranchPrefix.length)
     $version = "3.0.0-alpha-$feature-$revision-$commitSha"
 } else {
+    # local run
+    $branch = $branch.replace('/', '-')
```
928PJY: never used?

.deploy.yml
```diff
@@ -1,34 +1,58 @@
...
+- feature/*
+jobs:
+# unit & e2e test jobs
+- job: Linux Test(ubuntu-16.04)
+  pool:
+    vmImage: 'ubuntu-16.04'
+  condition: eq(variables['Build.Reason'], 'PullRequest')
+  steps:
+    - bash: ./build.sh
+- job: macOS Test(macOS-10.13)
```
OsmondJiang: any of the build job fails the whole build fails?

build.ps1
```diff
@@ -12,22 +12,27 @@ exec "dotnet test test\docfx.Test"
...
-    $feature = $branch.SubString($featureBranchPrefix.length)
-    $version = "3.0.0-alpha-$feature-$revision-$commitSha"
+} elseif ($branch.StartsWith($branchPrefix)) {
+    # CI triggered by any other branch
+    $branch = $branch.SubString($branchPrefix.length).Replace('/', '-')
+    $version = "3.0.0-alpha-$branch-$revision-$commitSha"
 } else {
-    $version = "3.0.0-alpha-$revision-$commitSha"
+    # local run
+    $branch = $branch.Replace('/', '-')
```
OsmondJiang: will there be too many packages versions? is there any limitation of MyGet?

.deploy.yml
mingwli: I haven't found the doc, but during my testing it happened once.

PR #3815, opened 2018-12-25
restore from a localized repo directly

OsmondJiang:
1. only implement the restore logic and a little build logic
2. add test cases against `restore`
3. need to discuss whether need support this behavior for `folder` mapping
4. add a lot of todos
5. will add more build logic like `bilingual` and `folder` mapping in another PR


docs/specs/localization.yml
```diff
@@ -52,6 +52,74 @@ repos:
 outputs:
   docs/dep.json:
 ---
+# restore source repo when build from loc repo directly 1
```
live1206: Add description to avoid 1,2,3?

src/docfx/build/Build.cs
```diff
@@ -16,7 +16,7 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
         {
             var errors = new List<Error>();
 
-            var (configErrors, config) = Config.Load(docsetPath, options);
+            var (configErrors, config) = LocalizationConvention.GetBuildConfig(docsetPath, options);
```
live1206: While building `en-us` repo, we also need to get the build config from `LocalizationConvention`?

docs/specs/localization.yml
OsmondJiang: description is in second line, the first line is helping me to locate the failed test case quickly

live1206: ok

src/docfx/localization/LocalizationMapping.cs
```diff
@@ -21,6 +21,7 @@ public enum LocalizationMapping
         /// /readme.md          -->         /localization/zh-cn/readme.md
         /// /files/a.md         -->         /localization/de-de/files/a.md
         /// </summary>
+        /// TODO: support build from localization folder directly?
```
live1206: can be removed?

src/docfx/localization/LocalizationMapping.cs
```diff
@@ -35,6 +36,9 @@ public enum LocalizationMapping
         /// #live                 -->           #live.zh-cn
         /// #live       -> bilingual ->         #live-sxs.zh-cn
         /// </summary>
+        /// TODO:
```
live1206: can be removed?

src/docfx/localization/LocalizationMapping.cs
OsmondJiang: need to be discussed

src/docfx/localization/LocalizationMapping.cs
OsmondJiang: yes, will removed

PR #3818, opened 2018-12-25
Valid names may only contain alphanumeric characters

mingwli:


.deploy.yml
```diff
@@ -6,19 +6,19 @@ pr:
...
 # unit & e2e test jobs
-- job: Linux Test(ubuntu-16.04)
+- job: LinuxTest_ubuntu_16_04
   pool:
     vmImage: 'ubuntu-16.04'
   condition: eq(variables['Build.Reason'], 'PullRequest')
   steps:
     - bash: ./build.sh
-- job: macOS Test(macOS-10.13)
+- job: macOSTest_macOS_10_13
```
yufeih: what does `Test` mean here?

mingwli: means it's running test(not deploy)  xxTest_platform

PR #3820, opened 2018-12-26
remove monikerRange from output

928PJY:


src/Microsoft.Docs.Template/Models/FileMetadata.cs
```diff
@@ -27,5 +27,7 @@ public class FileMetadata
 
         [JsonExtensionData]
         public JObject ExtensionData { get; set; }
+
+        public bool ShouldSerializeMonikerRange() => false;
```
yufeih: modify a test case to assert `monikerRange` is not in output

928PJY: sure

PR #3832, opened 2018-12-27
Upload code coverage and test report

yufeih:
Right now we hit **>80%** test coverage, this is a very impressive starting point. 

This PR integrates test coverage reports into our CI pipeline, to continuously monitor test coverage data. We use **80%** as our baseline and newer commits should not break this.

```
+-------------------------+--------+--------+--------+
| Module                  | Line   | Branch | Method |
+-------------------------+--------+--------+--------+
| docfx                   | 80.9%  | 81.7%  | 84.6%  |
+-------------------------+--------+--------+--------+
| Microsoft.Docs.Template | 66.1%  | 100%   | 66.1%  |
+-------------------------+--------+--------+--------+

+---------+--------+--------+--------+
|         | Line   | Branch | Method |
+---------+--------+--------+--------+
| Total   | 80.7%  | 81.7%  | 82.4%  |
+---------+--------+--------+--------+
| Average | 40.35% | 40.85% | 41.2%  |
+---------+--------+--------+--------+
```

Reports in VSTS:
image

### DIFF
--- a/.deploy.yml
+++ b/.deploy.yml
@@ -1,8 +1,9 @@
 trigger:
   batch: true
   branches:
-  - v3
-  - feature/*
+    include:
+      - v3
+      - feature/*
 pr:
 - v3
 - feature/*
--- a/.deploy.yml
+++ b/.deploy.yml
@@ -5,7 +5,7 @@ pr:
 - v3
 - feature/*
 jobs:
-# unit & e2e test jobs
+# Test pull requests on windows, linux and mac
 - job: LinuxTest_ubuntu_16_04
   pool:
     vmImage: 'ubuntu-16.04'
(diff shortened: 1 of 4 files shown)


.deploy.yml
```diff
@@ -1,11 +1,14 @@
 trigger:
-- v3
-- feature/*
+  batch: true
```
mingwli: > batch: true [](start = 1, length = 12)

> If you set batch to true, when a build is running, the system waits until the build is completed, then queues another build of all changes that have not yet been built

is this the behavior we want when running PR validation?

yufeih: yep, to reduce # of builds

PR #3834, opened 2018-12-27
Fix some simple code coverage problems

yufeih:
- Remove debug only ToString() implementations
- Added some missing tests
- Some simple cleanup

### DIFF
--- a/test/docfx.Test/lib/ParallelUtilityTest.cs
+++ b/test/docfx.Test/lib/ParallelUtilityTest.cs
@@ -20,7 +20,12 @@ namespace Microsoft.Docs.Build
             var exception = await Assert.ThrowsAnyAsync<Exception>(() => ParallelUtility.ForEach(Enumerable.Range(0, 1000), Run));
             Assert.Equal(exceptionType, exception.GetType());
 
+            exception = await Assert.ThrowsAnyAsync<Exception>(() => ParallelUtility.ForEach(Enumerable.Range(0, 1000), RunExpandChild));
+            Assert.Equal(exceptionType, exception.GetType());
+
             Task Run(int n) => n % 500 == 0 ? throw (Exception)Activator.CreateInstance(exceptionType) : Task.CompletedTask;
+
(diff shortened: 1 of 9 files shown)


test/docfx.Test/lib/ParallelUtilityTest.cs
```diff
@@ -20,7 +20,12 @@ public static async Task ThrowsTheSameException(Type exceptionType)
             var exception = await Assert.ThrowsAnyAsync<Exception>(() => ParallelUtility.ForEach(Enumerable.Range(0, 1000), Run));
             Assert.Equal(exceptionType, exception.GetType());
 
+            exception = await Assert.ThrowsAnyAsync<Exception>(() => ParallelUtility.ForEach(Enumerable.Range(0, 1000), RunExpandChild));
+            Assert.Equal(exceptionType, exception.GetType());
+
             Task Run(int n) => n % 500 == 0 ? throw (Exception)Activator.CreateInstance(exceptionType) : Task.CompletedTask;
+
+            Task RunExpandChild(int n, Action<int> queue) => n % 500 == 0 ? throw (Exception)Activator.CreateInstance(exceptionType) : Task.CompletedTask;
```
live1206: What is the difference between `RunExpandChild` and `Run`? `Action<int> queue` is not in use

test/docfx.Test/lib/ParallelUtilityTest.cs
```diff
@@ -20,7 +20,12 @@ public static async Task ThrowsTheSameException(Type exceptionType)
             var exception = await Assert.ThrowsAnyAsync<Exception>(() => ParallelUtility.ForEach(Enumerable.Range(0, 1000), Run));
             Assert.Equal(exceptionType, exception.GetType());
 
+            exception = await Assert.ThrowsAnyAsync<Exception>(() => ParallelUtility.ForEach(Enumerable.Range(0, 1000), RunExpandChild));
```
live1206: Shall we put this in a separate test? If the test fails It would be hard to see which case is failing

yufeih: then all the `[TheoryData]` would be duplicated

live1206: yes, the duplication of `[TheoryData]` should be fine

PR #3840, opened 2018-12-28
fix cache corruption when shrink multiple times

superyyrrzz:
If multiple outputs' content are same in cache, shrink can remove duplicate files and make these items point to the same file.

**Before**: First shrink per version, then shrink per PostProcessor in all versions. The second shrink can find duplicates cross versions, but will make the existing cache reference invalid, as version manifest has already been written to disk.

**this fix**: shrink only once before write version manifest on disk.

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildInfo.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildInfo.cs
@@ -108,6 +108,16 @@ namespace Microsoft.DocAsCode.Build.Engine.Incrementals
             return buildInfo;
         }
 
+        public void SaveVersionsManifet(string baseDir)
+        {
+            var expanded = Path.GetFullPath(Environment.ExpandEnvironmentVariables(baseDir));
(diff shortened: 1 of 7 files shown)


fenxu:
@vwxyzh @vicancy Please help to take a look if convenient.

src/Microsoft.DocAsCode.Build.Engine/Incrementals/BuildInfo.cs
```diff
@@ -108,6 +108,16 @@ public static BuildInfo Load(string baseDir, bool onlyValid)
             return buildInfo;
         }
 
+        public void SaveVersionsManifet(string baseDir)
+        {
+            var expanded = Path.GetFullPath(Environment.ExpandEnvironmentVariables(baseDir));
+            var targetDirectory = Path.Combine(expanded, DirectoryName);
```
partychen: Seems there is no used for targetDirectory

herohua:
do we have test case for non-versioned case?

superyyrrzz:
@herohua The added case in this PR also have 2 identical outputs within a version.

PR #3842, opened 2018-12-29
Put parameters to Context

yufeih:
This is a refactor that puts behavioral parameters into `Context`, it simplifies parameter passing of the build pipeline. There is no behavior change.

### DIFF
--- a/stylecop.ruleset
+++ b/stylecop.ruleset
@@ -57,8 +57,7 @@
     <!-- CA1060: Move pinvokes to native methods class -->
     <Rule Id="CA1060" Action="None" />
     <Rule Id="CA1061" Action="Warning" />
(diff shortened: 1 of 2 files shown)


stylecop.ruleset
```diff
@@ -57,7 +57,8 @@
     <!-- CA1060: Move pinvokes to native methods class -->
     <Rule Id="CA1060" Action="None" />
     <Rule Id="CA1061" Action="Warning" />
-    <Rule Id="CA1063" Action="Warning" />
+    <!-- CA1063: Implement IDisposable correctly -->
```
OsmondJiang: >IDisposable [](start = 27, length = 11)

make Context as sealed class, no need change that

Issue #3816, opened 2018-12-25
Support incremental build for FileMetadata change


superyyrrzz:
Now if `FileMetadata` in `docfx.json` changed, DocFX will force building all files. DocFX implements this by comparing current `configHash` with last `configHash` to determine whether current build is full.

A smarter solution is to check which files this `FileMetadata` change impacts, and only rebuild these files.

The implementation includes:

- [x] 1. Split the current `configHash` into `configHash` (without `FileMetadata` info) and `FileMetadataHash`. #3843
- [x] 2. Store `FileMetadata` config into cache. #3843
- [x] 3. Compare current `FileMetadata` and last `FileMetadata` to find the impacted files. The output is a list of globs. This is a pre-step before loading ~~models~~ changes. @partychen #3886 
- [x] 4. ~~Before loading model~~ **Along with `LoadChanges`**, check whether a file's `FileMetadata` is changed by checking whether it's in `FileMetadatChangedGlobs`, to determine incremental build or not. We should enshure this check is before `ExpandDependency`. #3843

@herohua @fenxu

PR #3843, opened 2018-12-29
create file metadata cache

superyyrrzz:
#3816

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
@@ -88,6 +88,8 @@ namespace Microsoft.DocAsCode.Build.Engine.Incrementals
                 BaseDir = Path.GetFullPath(Environment.ExpandEnvironmentVariables(baseDir)),
                 VersionName = parameters.VersionName,
                 ConfigHash = ComputeConfigHash(parameters, markdownServiceContextHash),
+                FileMetadataHash = ComputeFileMetadataHash(parameters.FileMetadata),
+                FileMetadata = parameters.FileMetadata,
                 AttributesFile = IncrementalUtility.CreateRandomFileName(baseDir),
                 DependencyFile = IncrementalUtility.CreateRandomFileName(baseDir),
                 ManifestFile = IncrementalUtility.CreateRandomFileName(baseDir),
@@ -278,6 +280,7 @@ namespace Microsoft.DocAsCode.Build.Engine.Incrementals
                     _changeDict[file] = ChangeKindWithDependency.Created;
                 }
             }
+            LoadFileMetadataChanges();
         }
 
         public List<string> ExpandDependency(DependencyGraph dg, Func<DependencyItem, bool> isValid)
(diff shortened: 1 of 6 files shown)


src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -363,6 +366,21 @@ public void InitFileAttributes()
...
+        public void LoadFileMetadataChanges()
+        {
+            var changedGlobs = CurrentBuildVersionInfo.FileMetadata.GetChangedGlobs(LastBuildVersionInfo.FileMetadata).ToArray();
+            foreach (var file in LastBuildVersionInfo.Attributes.Keys)
+            {
+                if (changedGlobs.Any(g => g.Match(file)))
+                {
+                    if (_changeDict[file] == ChangeKindWithDependency.None)
+                    {
+                        _changeDict[file] = ChangeKindWithDependency.Updated;
```
herohua: >Updated [](start = 69, length = 7)

would it break things if the file is actually deleted in the change list?

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -577,6 +595,19 @@ private static string ComputeConfigHash(DocumentBuildParameters parameter, strin
...
 
+        private static string ComputeFileMetadataHash(FileMetadata fileMetadata)
+        {
+            using (new LoggerPhaseScope("ComputeFileMetadataHash", LogLevel.Diagnostic))
+            {
+                if (fileMetadata == null)
+                {
+                    return string.Empty;
+                }
+                var json = JsonUtility.Serialize(fileMetadata);
```
herohua: >fileMetadata [](start = 49, length = 12)

do we need to consider order?

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
```diff
@@ -645,7 +676,8 @@ private bool GetCanVersionIncremental(IncrementalStatus buildInfoIncrementalStat
                 Logger.LogVerbose(message);
                 return false;
             }
-            if (CurrentBuildVersionInfo.ConfigHash != LastBuildVersionInfo.ConfigHash)
+            if (CurrentBuildVersionInfo.ConfigHash != LastBuildVersionInfo.ConfigHash
+                || CurrentBuildVersionInfo.FileMetadataHash != LastBuildVersionInfo.FileMetadataHash) // TODO: move to file level check
             {
                 string message = "Cannot build incrementally because config changed.";
```
herohua: >Cannot build incrementally because config changed. [](start = 34, length = 50)

it is not related to this change, but is it possible to include what part of config is changed in the information?

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
superyyrrzz: If a file is deleted, the value is `Deleted`, and I will keep it `Deleted` here.

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
superyyrrzz: I want to make FileMetadataHash a quick filter to determine if there is nothing changed. If only order is changed in a build, `changedGlobs` of FileMetadata will be empty, and all files can still build incrementally. So sort may not be necessary here.

src/Microsoft.DocAsCode.Build.Engine/Incrementals/IncrementalBuildContext.cs
superyyrrzz: Yes, we can add the whole config into cache. When config is different, we can log a message based on property wise comparison.

herohua:
Is it possible to add a test case on this?

PR #3845, opened 2018-12-29
Split config data model and config loader

yufeih:
This makes `Config.cs` contains pure data definition to make it easier to see all the build configs in docfx


src/docfx/restore/RestoreFile.cs
```diff
@@ -84,6 +84,18 @@ public static EntityTagHeaderValue GetEtag(string restorePath)
             return parts.Length == 2 ? new EntityTagHeaderValue(HrefUtility.UnescapeUrl(parts[1])) : null;
         }
 
+        public static IEnumerable<string> GetFileReferences(this Config config)
```
OsmondJiang: finally :)

src/docfx/config/Config.Loader.cs
```diff
@@ -0,0 +1,203 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Docs.Build
+{
+    internal sealed partial class Config
```
OsmondJiang: how about ConfigProvider?

superyyrrzz: or ConfigLoader as title?

src/docfx/config/Config.Loader.cs
```diff
@@ -0,0 +1,203 @@
...
+    {
+        private const string KFiles = "files";
```
superyyrrzz: What does `K` in `KFiles` mean?

yufeih: https://google.github.io/styleguide/cppguide.html#Constant_Names, maybe `kFiles` 😆

superyyrrzz: according to NO.12 in https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md use `Files`?

PR #3853, opened 2019-01-02
Resource metadata only needs locale and monikers

yufeih:
Reduces size throughout the pipeline. #3856


yufeih:
@928PJY , good question, it will be put inside manifest and I'll covered in later. This is just to make sure we checkin the impact change.

PR #3857, opened 2019-01-02
build from loc folder is not supported

OsmondJiang:
localization folder mapping doesn't support to be built directly
change localization convention to localization utility
add cwd to spec

### DIFF
--- a/docs/specs/localization.yml
+++ b/docs/specs/localization.yml
@@ -1060,7 +1060,7 @@ outputs:
     {"items":[{"toc_title":"reference toc a","children":[{"toc_title":"reference markdown a","href":"../a/a"}]}]}
 ---
 # builf from loc folder for folder mapping is not supported
-root: localization/zh-cn/
+cwd: localization/zh-cn/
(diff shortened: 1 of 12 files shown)


test/docfx.Test/e2e/E2ESpec.cs
```diff
@@ -13,6 +13,8 @@ public class E2ESpec
 
         public readonly string Repo;
 
+        public readonly string Root;
```
yufeih: `cwd` for command execution working directory?

docs/specs/localization.yml
```diff
@@ -1058,3 +1058,17 @@ outputs:
   docs/a/b.json:
   docs/b/toc.json: |
     {"items":[{"toc_title":"reference toc a","children":[{"toc_title":"reference markdown a","href":"../a/a"}]}]}
+---
+# builf from loc folder for folder mapping is not supported
```
mingwli: >builf from loc [](start = 2, length = 14)

build

Issue #3687, opened 2018-11-28
Include all files by default and use exclude to remove files from build scope


yufeih:
Per todays discussion, we want to clean config by not worrying about file extensions:

- `files` by default include all markdown, yaml and json files: `**/*.{md,yml,json}`
- Use `exclude` to remove files without having to specify extension, e.g. `markdown-template/**`
- `exclude` by default exclude `*` (files in root folder) (How to keep REDIRECTION.yml??)

yufeih:
The default value for `files` is `md,yml,json`, images like `svg,png` are not published by default unless they are referenced by an article.

yufeih:
Default `*` exclude glob has some [bugs](https://github.com/kthompson/glob/issues/42)

PR #3860, opened 2019-01-03
Include all markdown, yml, json files by default

yufeih:
Implements #3687


yufeih:
Failing only on linux https://github.com/kthompson/glob/issues/42

docs/specs/resolve.yml
```diff
@@ -111,7 +111,8 @@ outputs:
 ---
 # Show warning if the link is empty, missing or points to absolute file path
```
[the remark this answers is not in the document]
yufeih: covered in the next test case

src/docfx/config/Config.cs
```diff
@@ -10,8 +10,13 @@ namespace Microsoft.Docs.Build
 {
     internal sealed class Config
     {
-        private static readonly string[] s_defaultInclude = new[] { "docs/**/*.{md,yml,json}" };
-        private static readonly string[] s_defaultExclude = new[] { "_site/**/*", "localization/**/*" };
+        private static readonly string[] s_defaultInclude = new[] { "**/*.{md,yml,json}" };
+        private static readonly string[] s_defaultExclude = new[]
+        {
+            "_site/**",             // Default output location
```
live1206: Why is `_site/**` working the same as `_site/**/*`?

yufeih: `**` matches zero or more directories, we are do string comparison, so it literally means everything in that sub-directory. This behavior is the same as [node-glob](https://github.com/isaacs/node-glob)

docs/specs/resolve.yml
```diff
@@ -109,9 +109,10 @@ outputs:
     { "content": "<p><a target=\"_blank\" href=\"c\">c</a></p>" }
   docs/dir/c.json:
 ---
-# Show warning if the link is empty, missing or points to absolute file path
+# Show warning if the link is empty, missing
 inputs:
-  docfx.yml:
+  docfx.yml: |
+    exclude: '*'
```
OsmondJiang: >exclude: '*' [](start = 4, length = 12)

this actually change the user behavior, they just need specific which folder need to be build and publish at first, but right now i believe most of the repos need exclude "*"(root level files) in their config

OsmondJiang: and also if you made this change,  configuration migrator tool need to be changed also.

---
In reply to: [245192658](https://github.com/dotnet/docfx/pull/3860#discussion_r245192658) [](ancestors = 245192658)

src/docfx/config/Config.cs
```diff
@@ -10,8 +10,13 @@ namespace Microsoft.Docs.Build
...
+            "localization/**",      // Localization file when using folder convention
+            "**/includes/**",       // Default convention for reusable markdown tokens
```
OsmondJiang: >includes [](start = 16, length = 8)

do we also want to add `include`, `token`, `tokens` and like `README.md`, `CONTRIBUTING.md`, `LICENSE`

yufeih: solved with `defaultExclude`

src/docfx/config/Config.cs
```diff
@@ -31,7 +35,14 @@ internal sealed class Config
...
         /// </summary>
-        public readonly string[] Exclude = s_defaultExclude;
+        public readonly string[] Exclude = Array.Empty<string>();
+
+        /// <summary>
+        /// Gets the default file glob patterns excluded from this docset.
+        /// This typically contains system reserved exclude patterns,
+        /// users are recommended to use <see cref="Exclude"/> to exclude files.
+        /// </summary>
+        public readonly string[] DefaultExclude = s_defaultExclude;
```
OsmondJiang: do you want to add a spec test for:
1. default exclude root folder/includes files event user specific the other `exclude`
2. user can overwrite the default exclude if they want to publish/build the root folder files

superyyrrzz: `defaultExclude` has been limited to docfx reserved folders like `_site`, `localization`, which should not be overwrittern by user. It may confuse user whether to use `exclude` or `defaultExclude`, so we can only expose `exclude`

superyyrrzz: No, `defaultExclude` should not contain `*`.

Issue #3852, opened 2019-01-02
configuration generation for loc build directly from loc repo


OsmondJiang:
- [ ] configuration generated for restored source repo
- [ ] build history file generated for restored source repo

OsmondJiang:
#3789

OsmondJiang:
we are going to support full localization config:
1. reduce the dependencies  between docfx v3 and configuration design
2. reduce the migration steps for loc repos

cc @mingwli  @herohua  @yufeih

PR #3864, opened 2019-01-03
support build loc repo with docfx config

OsmondJiang:
if loc repo has docfx config file, we will use it, otherwise, we use source docset's config + overwrite config


OsmondJiang:
#3852

PR #3865, opened 2019-01-03
Add publish manifest file spec

yufeih:
This is a draft design spec of publish manifest file, some notable changes:

- Group system generated outputs to `_build` folder as we have more and more generated output now.
- Creates a dedicated `publish.json` specifically for publish purposes.
- Puts article metadata to `publish.json`
- Remove dependency map from ideal map for now. We can design and add it to a standalone file later.

@fenxu

### DIFF
--- a/docs/designs/output.md
+++ b/docs/designs/output.md
@@ -52,7 +52,7 @@ https://docs.microsoft.com/en-us/netstandard-2.0/dotnet/api/system.string.html#I
 > When a docset is not localized, we will not create the `{locale}` folder.
 > This allows static websites with minimal url when localized and verioning is not needed.
 
-## File Outputs
+## Content Outputs
 
 Each input file transforms to zero or more output files depending on its content type. `docfx` places output files in `output-path` relative to output directory. **Regardless of static rendering or dynamic rendering, output path shares the same schema**:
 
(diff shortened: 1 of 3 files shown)


fenxu:
👍

yufeih:
> 👍 question about publish.json, I think it looks like publish service configuration. The data contract may be decided by publish component.

Right, this is kind of the draft outline, let's finalize it together.

docs/designs/publish.md
```diff
@@ -0,0 +1,32 @@
...
+
+`docfx` generates a manifest file for publishing at `{output-path}/_build/publish.json` after build. It is the entry point for publishing. An example `publish.json` looks like this:
+
+```javascript
+{
+    "publish": [
+        {
+            "url": "/dotnet/api/system.string",
+            "path": "dotnet/api/system.string.json",
+            "hash": "d41d8cd98f00b204e9800998ecf8427e",
```
OsmondJiang: consider to add example for resource？

docs/designs/output.md
```diff
@@ -114,80 +114,14 @@ Different files can share the same `{site-url}` or `{site-path}` due to versioni
...
-            {  
-               "source":"dotnet/azure/service-less-app-token.md",
-               "type":"inclusion"
-            }
-         ]
-      }
-   ]
-}
-```
+Besides content files, `docfx` generates miscellaneous files during the build. The name of these system generated fiels starts with `.`. If the format is JSON, then the extension MUST be `.json` for faster DHS download speed.
```
OsmondJiang: it's better to not mention our low lever consumer in our design document, we don't want to be limited by them

docs/designs/publish.md
```diff
@@ -0,0 +1,32 @@
...
+The format is a json file that contains a list of files to be published. Each item under `publish` section contains the following properties:
+
+Name        | Description
+------------|-----------------
+url         | site url without locale.
+path        | location of the content file relative to `{output-path}`. Different files may share the same `{path}` if their contents are the same.
+hash        | MD5 hash of the content file specified by `{path}`.
+locale      | locale of the file.
+monikers    | monikers of the file.
+*additional properties | Additional properties can be added here if they are required by publish. For backward compatibility reasons, article metadata is required by publish, thus they are also put here.
```
OsmondJiang: it has not to, metadata is just like `blob`, need to be published, not 'required' by published, right? otherwise why not also put the `blob` here?

docs/designs/publish.md
```diff
@@ -0,0 +1,43 @@
...
+        },
+        {
+            "url": "/dotnet/media/logo.png",
+            "path": "dotnet/media/logo.png",
+            "hash": "d41d8cd98f00b204e9800998ecf8427e",
+            "locale": "en-us",
+            "monikers": ["netstandard-2.0", "netstandard-2.1"],
+
+            // additional properties needed for publish
+        }
```
fenxu: {
            "url": "/{base_path}/media/logo.png", (propose to be: url -> asset_id, but current, we can keep url and publish step can do normalize)
   "path": "{relative_path_to_publish_json_file}"
            "hash": "d41d8cd98f00b204e9800998ecf8427e",
            "monikers": ["netstandard-2.0", "netstandard-2.1"],
   extention_data:{
   
   }
}

yufeih: To clarify `extension_data` means using `[JsonExtensionData]`, the actual JSON is a flattened list.

928PJY: should `.xremap.json`, `.filemap.json` be added here too?

docs/designs/output.md
```diff
@@ -114,80 +114,14 @@ Different files can share the same `{site-url}` or `{site-path}` due to versioni
...
-            {  
-               "source":"dotnet/azure/service-less-app-token.md",
-               "type":"inclusion"
-            }
-         ]
-      }
-   ]
-}
-```
+Besides content files, `docfx` generates miscellaneous files during the build. The name of these system generated fiels starts with `.`. The extension MUST be `.json` for files in json format.
```
928PJY: typo

928PJY: >fiels [](start = 114, length = 5)

typo

docs/designs/publish.md
yufeih: no, system generated files have well known names, they are not added here.

PR #3876, opened 2019-01-07
Xref recursive reference check should consider `displayPropertyName`

live1206:
Add another test case

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -540,3 +540,26 @@ inputs:
 outputs:
   build.log: |
     ["error","circular-reference","Found circular reference: 'docs/a.json' --> 'docs/a.json'","docs/a.json"]
+---
+# SDP non-cyclic xref reference, whose property references can be resolved
+inputs:
+  docfx.yml:
+  docs/a.json: |
+    {
+      "$schema": "https://raw.githubusercontent.com/dotnet/docfx/v3/schemas/TestData.json",
+      "uid": "a",
+      "name": "name for a",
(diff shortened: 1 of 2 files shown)


src/docfx/build/xref/XrefMap.cs
```diff
@@ -19,21 +19,21 @@ internal class XrefMap
         private readonly IReadOnlyDictionary<string, XrefSpec> _externalXrefMap;
         private readonly Context _context;
 
-        private static ThreadLocal<Stack<Document>> t_recursionDetector = new ThreadLocal<Stack<Document>>(() => new Stack<Document>());
+        private static ThreadLocal<Stack<(string propertyName, Document parent)>> t_recursionDetector = new ThreadLocal<Stack<(string, Document)>>(() => new Stack<(string, Document)>());
```
yufeih: does it need `uid` for completeness? what if a document has multiple xrefspecs (uids)

live1206: We only support uid definition on root level for SDP now, will it be multiple uids later?

live1206: Add `uid` to the key, no test case for it since we can couldn't define multiple `uid` for now

docs/specs/xref.yml
```diff
@@ -540,3 +540,26 @@ inputs:
 outputs:
   build.log: |
     ["error","circular-reference","Found circular reference: 'docs/a.json' --> 'docs/a.json'","docs/a.json"]
+---
+# SDP non-cyclic xref reference, whose property references can be resolved
```
OsmondJiang: ^_^

PR #3875, opened 2019-01-07
Fix InclusionContext stack

yufeih:
### Before

The current `InclusionContext` implementation assumes every call to `InclusionContext.PushFile` creates a new markdown render stack frame and push it to the recursion detection stack within the call graph:

`a` [!include] `b` [!include] `c`:
```csharp
PushFile(a)
{
    PushFile(b)
    {
          PushFile(c) { }
    }
}
```

But this is incorrect when calling markup using xref reference (@) inside the same call graph:

`a` @ `b` @ `c`:

```csharp
// Current call graph is wrong, `b` is treated as inclusion.
PushFile(a)
{
    PushFile(b)
    {
          PushFile(c) { }
    }
}
```

### After

This PR solve the issue by separating `Markdown.ToHtml`  from `Markdown.ToHtml` used to render inclusion:

`a` [!include] `b` [!include] `c`:
```csharp
PushFile(a)
{
    PushInclusion(b)
    {
          PushInclusion(c) { }
    }
}
```

`a` @ `b` @ `c`:

```csharp
// Current call graph is wrong, `b` is treated as inclusion.
PushFile(a) { }
PushFile(b) { }
PushFile(c) { }
```

Related #3771

PR #3880, opened 2019-01-08
Output xref map href as external accessible URL

live1206:
Trying to fix #3875 

In v2, docfx uses xref service to resolve uid:
https://xref.docs.microsoft.com/query?uid=Microsoft.AspNetCore.Routing.IRouter
```json
[
    {
        "uid": "Microsoft.AspNetCore.Routing.IRouter",
        "name": "IRouter",
        "fullName": "Microsoft.AspNetCore.Routing.IRouter",
        "href": "https://docs.microsoft.com/dotnet/api/microsoft.aspnetcore.routing.irouter",
        "commentId": "T:Microsoft.AspNetCore.Routing.IRouter",
        "nameWithType": "Microsoft.AspNetCore.Routing.IRouter",
        "tags": [
            "/dotnet",
            "public"
        ]
    }
]
```

In v3, docfx uses xrefmap.json to resolve the uid, and the outputted `xrefmap.json` contains:
```json
[
    {
        "uid": "Microsoft.AspNetCore.Routing.IRouter",
        "name": "IRouter",
        "fullName": "Microsoft.AspNetCore.Routing.IRouter",
        "href": "/dotnet/api/microsoft.aspnetcore.routing.irouter",
        "commentId": "T:Microsoft.AspNetCore.Routing.IRouter",
        "nameWithType": "Microsoft.AspNetCore.Routing.IRouter",
        "tags": [
            "/dotnet",
            "public"
        ]
    }
]
```
Seems like the OPS outputted `xrefmap.json` contains the absolute URL as  href, we need to fix it in OPS for v3 to use. 
Since docfx v3 has the information of `hostname`, adding it to `xrefmap.json` output.

### DIFF
--- a/src/docfx/docset/Document.cs
+++ b/src/docfx/docset/Document.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Docs.Build
         /// _site/en-us/603b739b/dotnet/api/system.string/index.json
         ///
         ///  - Normalized using <see cref="PathUtility.NormalizeFile(string)"/>
-        ///  - Docs not start with '/'
+        ///  - Does not start with '/'
         ///  - Does not end with '/'
         /// </summary>
         public string SitePath { get; }
@@ -66,6 +66,11 @@ namespace Microsoft.Docs.Build
         ///  - Does not escape with <see cref="HrefUtility.EscapeUrl(string)"/>
         /// </summary>
         public string SiteUrl { get; }
+
+        /// <summary>
+        /// Gets the external accessible URL
+        /// </summary>
+        public string ExternalUrl { get; }
 
         /// <summary>
         /// Gets the document id and version independent id
@@ -108,6 +113,7 @@ namespace Microsoft.Docs.Build
             string filePath,
             string sitePath,
             string siteUrl,
+            string externalUrl,
             ContentType contentType,
             string mime,
             Schema schema,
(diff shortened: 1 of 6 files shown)


src/docfx/docset/Document.cs
```diff
@@ -67,6 +67,11 @@ internal class Document
         /// </summary>
         public string SiteUrl { get; }
 
+        /// <summary>
+        /// Gets the external accessible URL
+        /// </summary>
+        public string ExternalUrl { get; }
```
yufeih: This feels like `CanonicalUrlWithoutLocale`, see `BuildPage.GetCanonicalUrl`. Maybe also put `CanonicalUrl` here as well?

live1206: sure

OsmondJiang: Seems duplicated for me, what about just use CanonicalUrl for xref

---
In reply to: [245891513](https://github.com/dotnet/docfx/pull/3880#discussion_r245891513) [](ancestors = 245891513)

test/docfx.Test/docfx.test.yml
```diff
@@ -1,4 +1,5 @@
 # default config for specs in docfx
+baseUrl: https://docs.com
```
superyyrrzz: Feels confused as `docs.com` is use both as baseUrl and in `repos` in spec tests. I think it is more suitable here, so we can use another fake site for `repos`.

superyyrrzz:
Usually, the current docset and the referenced UID shared the same baseUrl `docs.microsoft.com`, so the resolved href could be a relative path instead of absolute path. v2 has feature `removeHost` for this, which v3 can also include when resolving xref.

Issue #3730, opened 2018-12-06
Flaky test: RestoreGitWorkTrees


yufeih:
RestoreTest.RestoreGitWorkTrees occasionally fail:

https://ceapex.visualstudio.com/Engineering/_build/results?buildId=28587

PR #3881, opened 2019-01-08
fix bug of get restored repo with branch staring with xxx-

OsmondJiang:
consider that case, two branches, one is master, another is mater-xxx, then when you want to get the restored master branch, you may get master-xxx.

#3730


OsmondJiang:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/59255

PR #3889, opened 2019-01-09
refactor xref map

live1206:
- remove on-demand load of internal xref spec since we have `Lazy` on property level
- fix error log for xref property in SDP

### DIFF
--- a/src/docfx/build/page/AttributeTransformer.cs
+++ b/src/docfx/build/page/AttributeTransformer.cs
@@ -13,7 +13,6 @@ namespace Microsoft.Docs.Build
     {
         public static Func<IEnumerable<DataTypeAttribute>, object, string, object> TransformSDP(
             Context context,
-            List<Error> errors,
             Document file,
             Action<Document> buildChild)
         {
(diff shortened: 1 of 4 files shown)


src/docfx/build/page/AttributeTransformer.cs
```diff
@@ -38,31 +36,31 @@ object Transform(IEnumerable<DataTypeAttribute> attributes, object value, string
...
         }
 
-        private static object TransformContent(Context context, List<Error> errors, DataTypeAttribute attribute, object value, Document file, Action<Document> buildChild)
+        private static object TransformContent(Context context, DataTypeAttribute attribute, object value, Document file, Action<Document> buildChild)
         {
             if (attribute is HrefAttribute)
             {
                 var (error, link, _) = context.DependencyResolver.ResolveLink((string)value, file, file, buildChild);
-                errors.AddIfNotNull(error);
+                context.Report.Write(file.ToString(), error);
```
yufeih: I feel like errors should be return values, the fact that `TransformContent` takes `errors` as input also feels wrong. Maybe our dev guide needed a section around error handling.

live1206: It would be good to update the dev guide. The method of taking `errors` as input is usually not thread-safe and the errors would not be logged if the method is executed on-demand. 
In this case, I just don't want to change the signature of the method, which is coupled with `JsonUtility`.

Issue #3791, opened 2018-12-19
Edit link for PR build


OsmondJiang:
We need to keep the `branch` info in `edit link` which points to pr branch for PR build, but this can cause  full publishing issue, there are several options:
1. Points all links to source branch using a template metadata **require changes across the pipeline*
2. Points only changed links to source branch **require taking changeset as input*
3. ~~hide the `edit link` in pr build~~
4. ~~points to target branch always~~

yufeih:
Prefer letting `git_content_url` in metadata pointing to source branch, and setting `git_content_url` in content pointing to target branch. This way, content will have consistent hash.

yufeih:
Finalized the spec [here](https://github.com/MicrosoftDocs/engineering/pull/270)

PR #3892, opened 2019-01-10
Generate original_content_git_url_template

yufeih:
#3791

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -288,6 +288,7 @@ outputs:
       "open_to_public_contributors": true,
       "content_git_url": "https://github.com/testowner/testrepo/blob/master-content-git-url/docs/index.md",
       "original_content_git_url": "https://github.com/testowner/testrepo/blob/master-content-git-url/docs/index.md",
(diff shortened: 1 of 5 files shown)


docs/specs/git.yml
```diff
@@ -288,6 +288,7 @@ outputs:
       "open_to_public_contributors": true,
       "content_git_url": "https://github.com/testowner/testrepo/blob/master-content-git-url/docs/index.md",
       "original_content_git_url": "https://github.com/testowner/testrepo/blob/master-content-git-url/docs/index.md",
+      "original_content_git_url_template": "{repo}/blob/{branch}/docs/index.md",
```
superyyrrzz: From the naming `original_content_git_url_template`, I would think it is used to generate `original_content_git_url` in template later, but `original_content_git_url` is also there. Consider better naming?

yufeih: this is the final name we came up with, do you have any suggestions?

superyyrrzz: Offline synced and this seems reasonable under the current template system.

PR #3894, opened 2019-01-10
add resolve alias config to support ~

928PJY:
in v2 `~` means the docset root path, which may be different from the repository root path.
so we add a `tildePath` config to support `~` in v3.

### DIFF
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -286,7 +286,8 @@ outputs:
 # Link or Image with `~`
 inputs:
   docfx.yml: |
-    tildePath: docs
+    resolveAlias:
+      ~: docs
   docs/a.md: |
     Link to [B1](~/b.md)
     Link to [B2](~/docs/b.md)
@@ -304,7 +305,8 @@ outputs:
 # <a> or <img> tag with `~`
 inputs:
   docfx.yml: |
-    tildePath: docs
+    resolveAlias:
+      ~: docs
   docs/a.md: |
     Link to <a href="~/b.md">b</a>
     <div>Link to <img src="~/c.png" /></div>
@@ -320,7 +322,8 @@ outputs:
 inputs:
   docfx.yml: |
     exclude: '*'
-    tildePath: docs/
+    resolveAlias:
+      ~: docs
   docs/a.md: |
     [!include[B1](~/b.md)]
     [!include[B2](~/../b.md)]
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -57,19 +57,6 @@ outputs:
   docs/dir-b/b.json: |
     { "content": "<p>Link to <a href=\"../dir-a/a\">a</a></p>" }
 ---
-# Path starting with ~ resolves from docset root directory
-inputs:
-  docfx.yml:
-  docs/dir-a/a.md: Link to [b](~/docs/dir-b/b.md)
-  docs/dir-b/b.md: Link to [a](~docs/dir-a/a.md)
-outputs:
-  docs/dir-a/a.json: |
-    { "content": "<p>Link to <a href=\"../dir-b/b\">b</a></p>" }
-  docs/dir-b/b.json: |
-    { "content": "<p>Link to <a href=\"%7Edocs/dir-a/a.md\">a</a></p>" }
-  build.log: |
-    ["warning","file-not-found","Cannot find file '~docs/dir-a/a.md' relative to 'docs/dir-b/b.md'","docs/dir-b/b.md"]
----
 # Links in include files are resolved relative to the include file
 inputs:
   docfx.yml:
@@ -87,7 +74,7 @@ outputs:
 inputs:
   docfx.yml:
   docs/a.md: Link to <a href="b.md" target="_blank"><em>b</em></a>
-  docs/b.md: <div>Link to <img src="~/docs/c.png" /></div>
+  docs/b.md: <div>Link to <img src="c.png" /></div>
   docs/c.png:
 outputs:
   docs/a.json: |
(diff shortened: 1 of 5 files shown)


OsmondJiang:
much like dependency repo alias

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -245,5 +238,17 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
 
             return (file != null ? null : Errors.FileNotFound(relativeTo.ToString(), path), file, null, query, fragment, false, pathToDocset);
         }
+
+        private string ResolveToRelativePath(string path, Document relativeTo)
+        {
+            foreach (var (alias, aliasPath) in relativeTo.Docset.ResolveAlias)
+            {
```
yufeih: assert that `alias` MUST ends with '/' otherwise `StartsWith` check is not sufficient.

docs/specs/resolve.yml
```diff
@@ -320,7 +322,8 @@ outputs:
 inputs:
   docfx.yml: |
     exclude: '*'
-    tildePath: docs/
+    resolveAlias:
```
yufeih: test that if an alias conflict with a physical path, peek the physical file.

docs/specs/resolve.yml
```diff
@@ -304,7 +305,8 @@ outputs:
 # <a> or <img> tag with `~`
 inputs:
   docfx.yml: |
-    tildePath: docs
+    resolveAlias:
```
yufeih: test that you can add multiple alias entries

src/docfx/config/Config.cs
```diff
@@ -83,6 +83,12 @@ internal sealed class Config
         /// </summary>
         public readonly Dictionary<string, string> Dependencies = new Dictionary<string, string>(PathUtility.PathComparer);
 
+        /// <summary>
+        /// Gets the map from resolve alias to relative path relatived to `docfx.yml` file
+        /// Default path of tilde symbol `~` is '.'
+        /// </summary>
+        public readonly Dictionary<string, string> ResolveAlias = new Dictionary<string, string>(PathUtility.PathComparer);
```
yufeih: nit: can just initialize `ResolveAlias` to `{ '~': '.' }` here so we can see the default value simply by looking at the definition here

src/docfx/docset/Docset.cs
```diff
@@ -163,6 +169,22 @@ private CultureInfo CreateCultureInfo(string locale)
...
+
+            foreach (var (alias, aliasPath) in config.ResolveAlias)
+            {
+                result[PathUtility.NormalizeFolder(alias)] = PathUtility.NormalizeFolder(aliasPath);
+            }
+
+            if (!result.ContainsKey("~/"))
+            {
+                result.Add("~/", "./");
+            }
```
OsmondJiang: add this to default value?

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -245,5 +238,17 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
 
             return (file != null ? null : Errors.FileNotFound(relativeTo.ToString(), path), file, null, query, fragment, false, pathToDocset);
         }
+
+        private string ResolveToRelativePath(string path, Document relativeTo)
```
OsmondJiang: ResolveToDocset

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -245,5 +238,17 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
...
+        {
+            foreach (var (alias, aliasPath) in relativeTo.Docset.ResolveAlias)
+            {
+                if (path.StartsWith(alias, PathUtility.PathComparison))
```
OsmondJiang: are there any case that we need consider the match order?

OsmondJiang: like ~/ and ~/xxx/

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -241,14 +239,18 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
 
         private string ResolveToDocsetRelativePath(string path, Document relativeTo)
         {
-            foreach (var (alias, aliasPath) in relativeTo.Docset.ResolveAlias.Reverse())
+            var docsetRelativePath = PathUtility.NormalizeFile(Path.Combine(Path.GetDirectoryName(relativeTo.FilePath), path));
+            if (!File.Exists(Path.Combine(relativeTo.Docset.DocsetPath, docsetRelativePath)))
             {
-                if (path.StartsWith(alias, PathUtility.PathComparison))
+                foreach (var (alias, aliasPath) in relativeTo.Docset.ResolveAlias.Reverse())
```
yufeih: can we reverse it in `Docset.ResolveAlias` to make resolve faster?

src/docfx/docset/Docset.cs
```diff
@@ -163,6 +169,18 @@ private CultureInfo CreateCultureInfo(string locale)
             }
         }
 
+        private Dictionary<string, string> LoadResolveAlias(Config config)
+        {
+            var result = new Dictionary<string, string>(PathUtility.PathComparer);
+
+            foreach (var (alias, aliasPath) in config.ResolveAlias)
```
live1206: `ResolveAlias` is used as a collection of tuples, not fetching value using a key. Does it have to be `Dictionary`?

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -245,5 +236,21 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
 
             return (file != null ? null : Errors.FileNotFound(relativeTo.ToString(), path), file, null, query, fragment, false, pathToDocset);
         }
+
+        private string ResolveToDocsetRelativePath(string path, Document relativeTo)
+        {
+            var docsetRelativePath = PathUtility.NormalizeFile(Path.Combine(Path.GetDirectoryName(relativeTo.FilePath), path));
+            if (!File.Exists(Path.Combine(relativeTo.Docset.DocsetPath, docsetRelativePath)))
+            {
+                foreach (var (alias, aliasPath) in relativeTo.Docset.ResolveAlias)
```
live1206: `ResolveAlias` is used as a collection of tuples, not fetching value using the key. Does it need to be `Dictionary`?

live1206: ok

PR #3896, opened 2019-01-10
Revert "Revert "set range as attribute of monikerRangeBlock token (#3621)""

928PJY:
This reverts commit 747bdb17d9f72d6a6dee404d4f2d05735267d146.


928PJY:
Since the op will decode the range attribute, so we can revert this revert.

Issue #3891, opened 2019-01-09
implict and etag may affect the LastWriteTime of file/worktree


OsmondJiang:
these two flags may have affects to the LstWriteTime of downloaded files and restored worktree and GC may delete the wrong files/worktrees.

@yufeih , @mingwli

yufeih:
Will this problem disappear say when we use `.lock.json` to track dependency version?

OsmondJiang:
I don't think so, as we talked offline yesterday, GC should not use lock files as input, still need check the last last write time, to avoid the dead lock.

OsmondJiang:
#3888

OsmondJiang:
let's consider more:
- In real production environment, the restored file version count or git repo's size/version count would be very large.
- We also need consider the restore worktree checkout speed limited by the worktree size.

Issue #3893, opened 2019-01-10
apply localization convention for the theme in output CRR info


OsmondJiang:
The theme put in CRR info need apply localization convention, otherwise the theme for loc CRR is source theme.

@928PJY

PR #3897, opened 2019-01-10
always need refresh the last write time for restored items

OsmondJiang:
no matter it's implicit or matching etag or not.

Try to fix the small problem of #3891, let's consider the lock file and big size later.

### DIFF
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -27,21 +27,40 @@ namespace Microsoft.Docs.Build
                 group (git.branch, git.flags)
                 by git.remote;
 
-            await ParallelUtility.ForEach(gitDependencies, RestoreGitRepo, Progress.Update);
+            var workTrees = new List<string>();
+
+            await ParallelUtility.ForEach(gitDependencies, async group => { workTrees.AddRange(await RestoreGitRepo(group)); }, Progress.Update);
+
+            foreach (var workTree in workTrees)
+            {
+                // update the last write time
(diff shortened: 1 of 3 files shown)


src/docfx/restore/RestoreGit.cs
```diff
@@ -27,21 +27,40 @@ public static async Task Restore(string docsetPath, Config config, Func<string,
                 group (git.branch, git.flags)
                 by git.remote;
 
-            await ParallelUtility.ForEach(gitDependencies, RestoreGitRepo, Progress.Update);
+            var workTrees = new List<string>();
```
live1206: use `ConcurrentBag` to be thread-safe?

src/docfx/restore/RestoreGit.cs
```diff
@@ -27,21 +27,40 @@ public static async Task Restore(string docsetPath, Config config, Func<string,
...
+                Directory.SetLastWriteTimeUtc(workTree, DateTime.UtcNow);
+            }
 
             if (!isDependencyRepo && LocalizationUtility.TryGetContributionBranch(docsetPath, out var contributionBranch, out var repo))
             {
                 await GitUtility.Fetch(repo.Path, repo.Remote, contributionBranch, config);
             }
 
-            async Task RestoreGitRepo(IGrouping<string, (string branch, GitFlags flags)> group)
+            async Task<List<string>> RestoreGitRepo(IGrouping<string, (string branch, GitFlags flags)> group)
```
yufeih: nit: how about set modified time inside, so you don't have the change the return value and structure that much.

OsmondJiang:
#3893

Issue #3888, opened 2019-01-09
Use the locked version of dependencies for PR build


OsmondJiang:
Here is the discussion summary:

1. docfx `restore` will generate `.lock.json` file to contains CRR(dependeny repository)s' repo+branch+commit
2. build service will upload the `.lock.json` to somewhere
3. When build want to use the locked version of CRR for PR build, pass the latest `.lock.json` file URL to docfx `restore` and `build`, docfx will use it.

cc @fenxu , @yufeih , @herohua

OsmondJiang:
just like [npm package-locks](https://docs.npmjs.com/files/package-locks), it describes the exact what we want to do.

OsmondJiang:
there are more things need to do for restore:  Restore source repo when building loc repo may cause too much disk usage/perf issue when checkout new worktree for every new version of source repo.

OsmondJiang:
We can only provide lock dependency repositories only, not works for files. @fenxu , @yufeih

PR #3900, opened 2019-01-10
restore + dependency lock + gc design spec

OsmondJiang:
#3888

### DIFF
--- a/docs/designs/restore-and-gc.md
+++ b/docs/designs/restore-and-gc.md
@@ -149,3 +149,12 @@ With GC command docfx `gc`, you can clean up the `%DOCFX_APPDATA_PATH%`.
 Whenever the `restore` command is executed, the `last write time` of downloaded files and git repository folder will be refreshed always, so the `gc` will clean up these files/repositories based on the `last write time`.
 
 The default `retention` rules are 15 days, which means `gc` will only keep these files/repository work trees which are accessed within 15 days, you can also overwrite this default rule by using `--retention-days {days}` command options.
+
+# Future consideration of Restore and GC
+
+The `restore` is to `add new and forget` and `gc` is to `delete`, considering the work tree size and check-out speed, we may want to `re-use` instead of `delete`.
+
+`re-use` existing work-tree of git can help us reduce the disk size and improve the checkout speed, this actually also fits the dependency files.
+
+But this solution need track usage status of each work tree, so let's put it in the future improvements.
+
--- a/docs/designs/restore-and-gc.md
+++ b/docs/designs/restore-and-gc.md
@@ -1,0 +1,115 @@
+# Restore
+
+Docfx `restore` helps you to restore all you dependency repositories/files(defined in [config](ocs/designs/config.md), fallback source repository and theme repository) into your local `%DOCFX_APPDATA_PATH%` for build preparation.
+
+The default `%DOCFX_APPDATA_PATH%` is `%USERPROFILE%/.docfx`, but you can reset it to any other place.
+
+The restore need make sure:
+
+  - Once one version of file/repository is restored, this version of the file/repository should NOT be updated until be cleaned up by `GC`. To make sure the `build`s which using this version are not broken causing by `restore`.
+  - Provide a mechanism to lock down the dependency file/repository version for different `build`s, just like [npm package locks](https://docs.npmjs.com/files/package-locks)
+
+## Restore files
+
+There are some cases that the docfx `build` depends on external files for help. 
+
+For example, if you want the docfx to resolve `github contributors` from building file git commit history, the `gitHub.userCache` file can be set up to reduce the API calls to [GitHub](https://github.com). The file can be just a http(s) url, which `restore` will help to download for `build` usage.
+
+```yml
+github:
+  userCache: https:///some-blob-service/github-user-cache.json
+```
+
+The restored files will be stored at `%DOCFX_APPDATA_PATH%/downloads/{url-short-path}/{version}+{etage}` like:
+
+```text
+`%DOCFX_APPDATA_PATH%`
+    | - downloads
+        | - raw.gith..tent.com+docascode+docfx-te..ndencies+62b0448+extend1.yml+dc363b0e
+            | - 18b265bc4e9a6b4fdff8b5210caf9e3541f4cb2d+%22f02fde5b4f3817266458170ba048c7fc2de46287%22
+            | - 3111eb522b092c9cef5cd0ccd86892cf847592d6+%2299a5dbb8d73da1ec452d74c0ab45b11ebe8615a0%22
+
+```
+
+- `{url-short-path}` is calculated from `file` url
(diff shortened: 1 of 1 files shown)


OsmondJiang:
for build service using dependency lock:

- for non-pr build, setup a no-existing locale `dependencyLock` file path in config and docfx `restore` will generate one in that place.

- for pr build, setup the `dependencyLock`  to a remote url or a locale path, use the `--no-save` to skip `updating` step.

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,151 @@
...
+            "downloads":{
+                "https:///some-blob-service/build-history.json":{
+                    "version": "66458170ba048c7fc2de4+%18b265bc4e9a6b4fdff8b5210caf9e357%22"
+                }
+            }
+        }
+    },
+    "downloads":{
+        "https:///some-blob-service/github-user-cache.json":{
+            "version": "18b265bc4e9a6bf4cb2d+%22f02fde5b4f3817266458170ba048c7fc2de46287%22"
```
yufeih: the example here is not a valid hash string,

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,151 @@
...
+        }
+    }
+}
+```
+
+- `git` is the version mappings of dependency repositories
+- `downloads` is the version mappings of dependency files
+- `version` is `commit id` of the dependency repository  or `content sha1` of the dependency file
```
yufeih: shall we name them `commit` and `sha1`/`hash` respectively then?

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,160 @@
...
+
+This file describes an `exact`, and more importantly `reproducible` `dependencies` trees. Once it's present, any future `restore` will base its work off this file, instead of recalculating`.
+
+The presence of the `dependency lock` changes the `restore` behavior:
+
+- The dependency files/repository described by this `dependency lock` is reproduced, which means reproducing the structure/version described in the file.
```
yufeih: add a section that lock file generation process `docfx.yml + .lock.json --> .lock.json` should be _idempotent_, and that content inside `.lock.json` is always be sorted.

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,160 @@
...
+You can setup your `.lock.json` file(local file or a URL) in the configuration like
+
+```yml
+dependencyLock: https://some-blob-service/repo/branch/.lock.json
+``` 
+
+or 
+
+```yml
+dependencyLock: .lock.json
```
yufeih: add that the default value for `dependencyLock` is _null_, and by default, `docfx restore` will not generate a lock file.

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+Whenever you run docfx `restore`, docfx will:
+
+- [GET] **Try** to get the `.lock.json` file from configuration, if it's present, any future dependencies restore will base its work off this file.
+- [UPDATE] Once finished the restore, docfx will **try** to create/update the `.lock.json` file defined in configuration, no matter it's a local or remote location.
```
yufeih: `githubUserCache` has a flag `updateRemoteUserCache=false`  to determine whether to update remote cache. Maybe similar behavior here as well?

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,160 @@
...
+
+For the last step [UPDATE], you can use command `--no-save` to skip.
```
yufeih: `npm` generates lock files by default, so it has a `--no-save` shortcut to disable it. We are quite the opposite: we don't generate lock files by default, unless you set `dependencyLock` somewhere, so maybe we don't need this option.

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,160 @@
+# Restore
+
+Docfx `restore` helps you to restore all you dependency repositories/files(defined in [config](ocs/designs/config.md), fallback source repository and theme repository) into your local `%DOCFX_APPDATA_PATH%` for build preparation.
```
superyyrrzz: `[config](ocs/designs/config.md)` invalid link

OsmondJiang: changed. 赞

---
In reply to: [247026326](https://github.com/dotnet/docfx/pull/3900#discussion_r247026326) [](ancestors = 247026326)

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,160 @@
...
+            | - master-fd6968dec9a9a39aec1845232466fe35fca520da
+            | - live-a82cb2a51f6925b2fa87275218b01678c5aadfe1
+```
+
+- `{url-short-path}` is calculated from `git remote` url
+- `{worktree-name}` is combined by `branch name` and `HEAD commit id`
+
+## Dependency lock
+
+There are some cases you want to keep using the **same version** of dependency files/repository between different `build`s.
```
superyyrrzz: This works for git, but not for file. We can go to a specific commit for a git repo, but cannot ask server to return a file with a specific e-tag existed before.

So, lock file can only consider git. DocFX always fetches latest for file dependency if cache miss.

live1206: For files, could `Etag` work similar to `version`?

docs/designs/restore-and-gc.md
OsmondJiang: good idea

---
In reply to: [247021515](https://github.com/dotnet/docfx/pull/3900#discussion_r247021515) [](ancestors = 247021515)

docs/designs/restore-and-gc.md
OsmondJiang: already have NOTE about that

---
In reply to: [247030971](https://github.com/dotnet/docfx/pull/3900#discussion_r247030971) [](ancestors = 247030971)

docs/designs/restore-and-gc.md
```diff
@@ -0,0 +1,217 @@
...
+```text
+`%DOCFX_APPDATA_PATH%`
+    | - downloads
+        | - raw.gith..tent.com+docascode+docfx-te..ndencies+62b0448+extend1.yml+dc363b0e
+            | - 18b265bc4e9a6b4fdff8b5210caf9e3541f4cb2d+%22f02fde5b4f3817266458170ba048c7fc2de46287%22
+            | - 3111eb522b092c9cef5cd0ccd86892cf847592d6+%2299a5dbb8d73da1ec452d74c0ab45b11ebe8615a0%22
+
+```
+
+- `{url-short-path}` is calculated from `file` url
```
928PJY: >is calculated from `file` url [](start = 21, length = 29)

what is the calculate logic? should we add it here?

OsmondJiang: the above `restore file` and `restore git`

Issue #3838, opened 2018-12-28
refine error code/message/level before first release


mingwli:
#### level change: ** need work, change level **
  - [x] circular-reference: add 'error' in config migration-tool(for the ones thrown by markdig)
  - [ ] merge-conflict: suggestion
  - [x] include-is-redirection: error

#### split into more specific errors
  - [x] null-value: keep v2 behavior; scalar null value: info & null array value: warning
  - [x] *bookmark-not-found*: internal(suggestion) & external(warning)

#### minor modification 
  - [x] *git-log-error*: shadow => shallow
  - [x] invalid-yaml-header: remove (never used)

#### others
  - [x] absolute-file-path: don't resolve as `link<a>`
  - [x] invalid-redirect-to: #3943
  - [x] reserved-metadata: keep current behavior(throw warning and publish as redirection file)
  - [x] redirection-out-of-scope: throw warning and don't auto add
  - [x] heading-not-found: info, and resolve from h1-h3(follow h2 behavior to be prudent)

#### for backlog(don't need change)
  - [ ] invalid include type: currently we only throw exception when it's redirection, and it could be anything
  - [ ] **internal**: download-failed: more details using http response status code
  - [ ] **internal**: git-clone-failed: addWorkTree / clone failed? need check
  - [ ] **?**: moniker-config-missing: how to handle monikerDefinition & monikerRange conflict
  - [ ] **internal**: upload-failed: alert and monitor
  - [ ] **?**: link-is-dependency: better error code, may need to merge 
  - [ ] **?**: empty-monikers: file / zone level
  - [ ] **?**: *invalid-toc-level/invalid-toc-syntax/missing-toc-head*: only happens in markdown toc
  - [ ] *file-not-found*: inclusion => error(already thrown by markdig); ?link => warning; user_profile (url) => system error
  - [x] unknown-field: update landingData schema?

yufeih:
`invalid-redirection`, `include-is-redirection` are all taking about the same **link type mismatch** problem, that is, you can only reference a set of allowed types. Otherwise we could have `include-is-resource`, `include-is-config` and the list goes on and on.

yufeih:
With #3784, we only have `null-inside-array` now.

OsmondJiang:
bookmark-not-found suggest to be spited into `self-bookmark-not-found`(info) and `bookmark-not-found`(warning)?

mingwli:
[...]

PR #3903, opened 2019-01-11
Split null-array-value from null-value

live1206:
#3838


src/docfx/lib/JsonUtility.cs
```diff
@@ -267,22 +267,31 @@ public static (List<Error>, JToken) RemoveNulls(this JToken token)
...
 
-            RemoveNullsCore(token, errors, nullNodes, removeNodes);
+            RemoveNullsCore(token, errors, nullNodes, nullArrayNodes);
 
             foreach (var node in nullNodes)
             {
-                var name = node is JProperty prop ? prop.Name : (node.Parent?.Parent is JProperty p ? p.Name : node.Path);
-                errors.Add(Errors.NullValue(ToRange(node), name, node.Path));
+                var (lineInfo, name) = Parse(node);
+                errors.Add(Errors.NullValue(new Range(lineInfo.LineNumber, lineInfo.LinePosition), name, node.Path));
```
yufeih: I added a `ToRange` helper for this purpose

PR #3909, opened 2019-01-11
Add console diagnostic logs with --verbose

yufeih:
Add a `--verbose` command line option to show diagnostic logs written using `Log.XXX`

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -192,7 +192,7 @@ namespace Microsoft.Docs.Build
             }
             catch (Exception ex) when (DocfxException.IsDocfxException(ex, out var dex))
             {
-                context.Report.Write(file.ToString(), dex);
+                context.Report.Write(file.ToString(), dex.Error);
                 manifestBuilder.MarkError(file);
                 return new List<string>();
             }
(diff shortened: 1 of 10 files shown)


src/docfx/build/Build.cs
```diff
@@ -192,6 +192,7 @@ void ValidateBookmarks()
             }
             catch (Exception ex) when (DocfxException.IsDocfxException(ex, out var dex))
             {
+                Log.Error(ex);
```
OsmondJiang: how about add it Report.Write? otherwise we need add Log.Error everywhere?

src/docfx/lib/log/Log.cs
```diff
@@ -0,0 +1,63 @@
...
+    {
+        private static AsyncLocal<bool> t_verbose = new AsyncLocal<bool>();
+
+        public static IDisposable BeginScope(bool verbose)
+        {
+            t_verbose.Value = verbose;
+            return new LogScope(() => t_verbose.Value = false);
+        }
+
+        public static void Info(string message)
```
live1206: `Info` is not in use for now? Maybe it can be consolidated with `Error(string message)`, sharing most logic

Issue #3856, opened 2019-01-02
Publish without --legacy


yufeih:
To publish directly from `docfx build` result:

- [x] Finalize publish spec
- [x] Move jint outside `--legacy`
- [x] Add file metadata to publish manifest
- [x] Add output file hash (SHA1).
- [x] ~~Add content type: (This is done in publish step)~~
    - Json files: `application/json`
    - Html files: `text/html`
    - Fallback to aspnet core `FileExtensionContentTypeProvider`
    - Fallback to `application/octet-stream`
- [x] Run jint for TOC
- [x] ~~Remove `--legacy` ??~~

OsmondJiang:
what's the definition of manifest, seems we put too much things into manifest file....

PR #3914, opened 2019-01-15
Update publish output model

yufeih:
This PR touched a lot of files, but most of them are renames according to [publish.md](https://github.com/dotnet/docfx/blob/v3/docs/designs/publish.md).

- build.manifest --> .publish.json
- Split dependency map to `.dependencymap.json`

#3856


src/docfx/build/dependency/DependencyMap.cs
```diff
@@ -14,5 +15,19 @@ public DependencyMap(Dictionary<Document, List<DependencyItem>> map)
             : base(map)
         {
         }
+
+        public object ToDependencyMapModel()
```
live1206: why is the return type `objecct`?

PR #3919, opened 2019-01-15
Move models needed by build into docfx

yufeih:
Move template out of razor template.

#3856


src/docfx/build/page/Contributor.cs
```diff
@@ -0,0 +1,22 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
```
OsmondJiang: put them into one model folder?

yufeih: lets do that during the data modeling casing change.

PR #3846, opened 2018-12-29
Update data modeling developer guide

yufeih:
Added section about how to model data in docfx

PR #3928, opened 2019-01-16
Add metadata to .publish.json

yufeih:
This PR adds metadata needed for publishing to `.publish.json`:

- `locale`
- `monikers`
- `redirect_url`: For redirections. Redirection files now does not have output json files in ideal output, because the content has been redirected anyway.
- `ExtensionData`: Contains custom metadata specified in markdown yaml header. It will be converted to metadata generated from jint in the future.


docs/specs/moniker.yml
```diff
@@ -179,25 +179,24 @@ inputs:
...
       "content": "<p>Moniker: netcore-1.2, netcore-1.3</p>",
       "monikers": ["netcore-1.2", "netcore-1.3"]
     }
-  f961cd88/docs/a-old.json: |
+  .publish.json: |
     {
-      "redirect_url": "/docs/a",
-      "monikers": ["netcore-1.2", "netcore-1.3"]
+      "publish": [
+        { "url": "/docs/a-old", "redirect_url": "/docs/a1", "monikers": [ "netcore-1.0", "netcore-1.1" ] },
```
OsmondJiang: do we have output path here? why need "publish" in ".publish.json"?

src/docfx/build/resource/BuildResource.cs
```diff
@@ -3,21 +3,42 @@
...
+
+            if (file.Docset.Config.Output.CopyResources)
+            {
+                outputPath = file.GetOutputPath(monikers);
+                context.Output.Copy(file, outputPath);
+            }
+            else
             {
+                // Output path is source file path relative to output folder when copy resource is disabled
+                outputPath = PathUtility.NormalizeFile(
```
OsmondJiang: they can be in different disk, what about still use the absolute path?

src/docfx/publish/PublishItem.cs
```diff
@@ -2,15 +2,26 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Docs.Build
 {
+    [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
```
OsmondJiang: can we add it to default setting?

OsmondJiang:
#3856

src/docfx/publish/PublishItem.cs
yufeih: I hope so, that i'll be a separate PR once #3846 is in

src/docfx/build/resource/BuildResource.cs
yufeih: When it is in a different disk, `GetRelativePath` should still produce the correct path. It is defined as path relative to `.publish.json`, so the publish routine can just _combine_ output folder and `path` to locate the file.

docs/specs/moniker.yml
yufeih: yes, `path` is checked in other PRs, this test focuses on `redirect_url`.

OsmondJiang: what about the second question?

superyyrrzz: Maybe to avoid making the root level become array type

PR #3930, opened 2019-01-17
Update TripleColonParser.cs for zone correct closed

v-caxian:
**Related Work:** [User Story 44737: Markdig parser DCR to enable zone end validation](https://ceapex.visualstudio.com/Engineering/_workitems/edit/44737)

**Root cause:** The `SelfClosing` has only the default value of false, which causes the value of the `Closed` to always be true.


v-caxian:
@jeahyoun @yishengjin1413

v-caxian:
@yufeih, done

v-caxian:
Hi @yufeih, 

Actually, I don't have write access to this repository, do you know who can help me merge the pull request?
After the pull request merged, when is the next deployment?

yufeih:
I'll merge it once @yishengjin1413 is ok with this. It'll then be deployed to PPE the next day.

Issue #3457, opened 2018-09-27
Support multiple uid within one SDP


live1206:
https://docs.microsoft.com/en-us/dotnet/api/system.string.endswith?view=netframework-4.7.2 https://docs.microsoft.com/en-us/javascript/api/@azure/cosmos/?view=azure-node-latest

yufeih:
Right now SDP only has one root level uid. In the above example, method overload shares the same `uid`, node sdk is an aggregation page, each type has its own sub page.

superyyrrzz:
For overload, they have different uid like `System.String.EndsWith(System.String)`, `System.String.EndsWith(System.String,System.StringComparison)` in the same page.

For the node sdk, you can scroll down to the end of page, and see "Function Details" who has its own definition there.

PR #3934, opened 2019-01-18
Support multiple uid

live1206:
#3457 
- Users should be able to define multiple uids within one SDP
- For uid defined in non-root level, the related href should append bookmark converted form uid


src/docfx/build/xref/XrefMap.cs
```diff
@@ -354,21 +358,105 @@ private static (List<Error> errors, InternalXrefSpec spec) LoadSchemaDocument(Co
...
                 obj,
                 schema.Type,
                 transform: AttributeTransformer.TransformXref(context, file, null, extensionData));
 
             errors.AddRange(schemaErrors);
-            var xref = new InternalXrefSpec
+
+            var extensionDataByUid = new Dictionary<string, (bool isRoot, Dictionary<string, Lazy<JValue>> properties)>();
+
+            if (extensionData.Count == 0)
```
yufeih: nit: no need to special case `extensionData.Count == 0` here, can just loop over it.

src/docfx/build/xref/XrefMap.cs
```diff
@@ -354,21 +358,105 @@ private static (List<Error> errors, InternalXrefSpec spec) LoadSchemaDocument(Co
...
+                    Href = isRoot ? file.CanonicalUrlWithoutLocale : $"{file.CanonicalUrlWithoutLocale}#{GetBookmarkFromUid(item.Key)}",
+                    ReferencedFile = file,
+                };
+                xref.ExtensionData.AddRange(properties);
+                return xref;
+            }).ToList();
+
+            return (errors, specs);
+
+            string GetBookmarkFromUid(string uid)
```
yufeih: different uids may map to the same bookmark using this algorithm, how about using `Uri.EscapeDataString`?

src/docfx/build/xref/XrefMap.cs
```diff
@@ -354,21 +358,105 @@ private static (List<Error> errors, InternalXrefSpec spec) LoadSchemaDocument(Co
...
+                => Regex.Replace(uid, @"\W", "_");
+
+            (string uid, string jsonPath) MatchExtensionDataToUid(string jsonPath)
+            {
+                var reverse = uids.ToDictionary(x => x.Value, x => x.Key);
```
yufeih: build the `reverse` lookup upfront to save duplicated work here. Name them something like `uidToJsonPath` and `jsonPathToUid` so we know what each `Dict<string, string>` represents

src/docfx/build/xref/XrefMap.cs
```diff
@@ -354,21 +358,105 @@ private static (List<Error> errors, InternalXrefSpec spec) LoadSchemaDocument(Co
...
+                    context.Report.Write(filePath, Errors.UidConflict(str));
+                }
+            }
+
+            foreach (var item in token.Children())
+            {
+                var property = item as JProperty;
+                if (property.Value is JObject obj)
+                {
+                    GetUids(context, filePath, obj, string.IsNullOrEmpty(parentName) ? property.Name : $"{parentName}.{property.Name}", uids);
```
yufeih: use `JToken.Path` over manually building json path

src/docfx/build/xref/XrefMap.cs
live1206: This logic is from the template of v2, just want to follow the same behavior

src/docfx/build/xref/XrefMap.cs
live1206: sounds good

src/docfx/build/xref/XrefMap.cs
```diff
@@ -354,21 +359,102 @@ private static (List<Error> errors, InternalXrefSpec spec) LoadSchemaDocument(Co
...
+        }
+
+        private static void GetUids(Context context, string filePath, JObject token, string parentName, Dictionary<string, string> uidToJsonPath, Dictionary<string, string> jsonPathToUid)
+        {
+            if (token is null)
+                return;
+
+            if (token.TryGetValue("uid", out var value) && value is JValue v && v.Value is string str)
+            {
+                if (!uidToJsonPath.TryAdd(str, parentName))
```
yufeih: is this `token.Path`, if so then `parentName` is not needed.

live1206: u r right 👍

docs/specs/xref.yml
```diff
@@ -222,7 +222,56 @@ outputs:
...
 ---
-# Define uid in non-root object level, it will be ignored
+# support multiple uids in the same array in one SDP
+inputs:
+  docfx.yml:
+  docs/a.yml: |
+    #YamlMime:TestData
+    uid: a
+    description: description for a
+    array:
```
OsmondJiang: add test for single `data`

src/docfx/build/xref/XrefMap.cs
```diff
@@ -354,21 +359,102 @@ private static (List<Error> errors, InternalXrefSpec spec) LoadSchemaDocument(Co
...
+        private static void GetUids(Context context, string filePath, JObject token, Dictionary<string, string> uidToJsonPath, Dictionary<string, string> jsonPathToUid)
+        {
+            if (token is null)
+                return;
+
+            if (token.TryGetValue("uid", out var value) && value is JValue v && v.Value is string str)
+            {
+                if (!uidToJsonPath.TryAdd(str, token.Path))
+                {
+                    context.Report.Write(filePath, Errors.UidConflict(str));
```
OsmondJiang: >context [](start = 20, length = 7)

does it only woking for inside conflicts?

live1206: Here we only want to detect duplicated uids defined in one file.
The case missing is, detect duplicated uid defined already externally, defined again in the build repo

OsmondJiang: no need comparing with external uids, but make sure the uid is unique inside this docset. add test for that

---
In reply to: [250914849](https://github.com/dotnet/docfx/pull/3934#discussion_r250914849) [](ancestors = 250914849)

live1206: test added

PR #3940, opened 2019-01-21
Remove `outputRootRelativePath` and `fileRelativePath`

yufeih:
These two metadata should be safe to remove from our output


OsmondJiang:
we should already apply some change in template side?

yufeih:
> we should already apply some change in template side?

This 2 metadatas are not used by templates

PR #3944, opened 2019-01-21
use dependency lock in restore and build

OsmondJiang:
use dependency lock in restore
use dependency lock in build
sub dependecy git repository can has their own dependency lock
root dependency lock has higher priority to be used


docs/specs/restore.yml
```diff
@@ -166,3 +166,94 @@ repos:
...
+  https://docs.com/lock1/c:
+    - files:
+        docs/dep2.md: "2333"
+    - files:
+        docs/dep2.md: "233333"
+outputs:
+  docs/a.json: |
+    {"content":"<p>&quot;def&quot;</p>\n<p>233333</p>\n<p>2333</p>\n"}
+---
+# restore with sub dependency lock only
```
live1206: add a test case which contains  both root lock and sub-lock, root lock should dominate

OsmondJiang: done

src/docfx/restore/RestoreMap.cs
```diff
@@ -50,8 +54,9 @@ string FindLastModifiedGitRepository()
                 return (
                     from path in Directory.GetDirectories(repoPath, "*", SearchOption.TopDirectoryOnly)
                     let name = Path.GetFileName(path)
-                    where name.StartsWith(HrefUtility.EscapeUrlSegment(branch) + "-" + branch.GetMd5HashShort() + "-") &&
-                          GitUtility.IsWorkTreeCheckoutComplete(repoPath, name)
+                    where GitUtility.IsWorkTreeCheckoutComplete(repoPath, name) &&
+                        ((locked && name == $"{RestoreGit.GetWorkTreeHeadPrefix(branch, locked)}-{commit}") ||
+                        (!locked && name.StartsWith(RestoreGit.GetWorkTreeHeadPrefix(branch))))
                     orderby new DirectoryInfo(path).LastWriteTimeUtc
```
yufeih: This orderby may not be correct, we may need to ask git for the latest version, it can be done when doing the slot/rent/return model.

src/docfx/restore/lock/DependencyLock.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.Docs.Build
+{
+    internal class DependencyLock : DependencyVersion
```
yufeih: consider split models from behaviors

src/docfx/restore/lock/RestoreLock.cs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.IO;
+
+namespace Microsoft.Docs.Build
+{
+    internal static class RestoreLock
```
yufeih: RestoreLock/DependencyLock/RestoreMap feels duplicated

OsmondJiang: u r right, remove RestoreLock

src/docfx/restore/RestoreMap.cs
OsmondJiang: I've spited them into {branch}-xxxx and locked-{branch}-xxx models, make sure the original behavior is correct(last write time order)

PR #3946, opened 2019-01-21
Remove legacy resource and redirection conversion

yufeih:
This PR moves legacy resource and redirection logic, to `BuildResource` and `BuildRedirection`. Legacy page and TOC conversion will be moved later.


OsmondJiang:
#3950

src/docfx/build/legacy/files/LegacyPage.cs
```diff
@@ -14,17 +14,9 @@ internal static class LegacyPage
             Document doc,
             LegacyManifestItem legacyManifestItem)
         {
-            // OPS build use TOC ouput as data page
```
live1206: keep legacy TOC in this PR? legacy logic not added to BuildTableOfContents

src/docfx/build/legacy/files/LegacyPage.cs
```diff
@@ -44,17 +36,9 @@ internal static class LegacyPage
 
                 var themesRelativePathToOutputRoot = "_themes/";
 
-                if (!string.IsNullOrEmpty(doc.RedirectionUrl))
-                {
-                    rawMetadata = LegacyMetadata.GenerateLegacyRedirectionRawMetadata(docset, pageModel);
```
live1206: remove `LegacyMetadata.GenerateLegacyRedirectionRawMetadata()`, not in use any more

yufeih: These lines are covered in LegacyTOC, I think they are kept here accidentally during refactor.

src/docfx/build/legacy/files/LegacyPage.cs
yufeih: These lines are covered in LegacyTOC, I think they are kept here accidentally during refactor.

OsmondJiang: impact test will tell us

OsmondJiang:
redirection and resource are not using JINT to run template, can we add spec test here?

yufeih:
> redirection and resource are not using JINT to run template, can we add spec test here?

The legacy test I added has covered both image and redirections

Issue #3943, opened 2019-01-21
Urls in Redirections must start with '/'


yufeih:
To remove the logic of host name, docfx needs to check that urls in `Redirections` **MUST** start with '/',

OsmondJiang:
The invalid redirection reports `info/warning`, but no output generated.

We can just set the document_id to false in configuration migration if the `redirect_to` is not start with `/` or 'docs.microsoft.com'.

@mingwli

mingwli:
https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.MigrationTool/pullrequest/9416?_a=overview

PR #3955, opened 2019-01-22
Check redirect url

mingwli:
#3943 
config.redirections dest must start with '/'

### DIFF
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -168,17 +168,6 @@ outputs:
   docs/a.json: |
     {"content":"<p><a href=\"/absolute/path?a#b\">redirect</a></p>"}
 ---
-# Files redirected to a relative URL
-inputs:
-  docfx.yml: |
-    redirections:
-      docs/redirect.md: relative/path
-  docs/a.md: |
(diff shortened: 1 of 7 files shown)


docs/specs/resolve.yml
```diff
@@ -168,17 +168,6 @@ outputs:
   docs/a.json: |
     {"content":"<p><a href=\"/absolute/path?a#b\">redirect</a></p>"}
 ---
-# Files redirected to a relative URL
```
OsmondJiang: this case should be added by reason?

OsmondJiang: impact test will test us it's used or not :)

mingwli: there does exist urls other than `docs.microsoft.com`.
```text
articles/machine-learning/service/aml-pricing.md: 'https://azure.microsoft.com/pricing/details/machine-learning-services'
articles/machine-learning/studio/live-chat.md: 'https://social.msdn.microsoft.com/Forums/en-US/home?forum=MachineLearning'
```
how should we handle this?

yufeih: these are simple redirections (without affecting document id)

Issue #3950, opened 2019-01-22
Add test for legacy outputs


OsmondJiang:
Since we are going to move most of the legacy logic inside of docfx, we need treat legacy output as one kind of docfx outputs, and add tests for this outputs(no more legacy only)

@yufeih  @herohua

yufeih:
Legacy output depend on template, but template is right now private.

PR #3956, opened 2019-01-22
Add a basic legacy test

yufeih:
Implements #3950

### DIFF
--- a/src/docfx/template/TemplateEngine.cs
+++ b/src/docfx/template/TemplateEngine.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Docs.Build
 
         public JObject TransformMetadata(string schemaName, JObject metadata)
         {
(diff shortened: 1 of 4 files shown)


src/docfx/template/TemplateEngine.cs
```diff
@@ -51,7 +51,7 @@ public string Render(PageModel model, Document file)
 
         public JObject TransformMetadata(string schemaName, JObject metadata)
         {
-            return JObject.Parse(((JObject)_js.Run($"{schemaName}.mta.json.js", "transform", metadata)).Value<string>("content"));
+            return JObject.Parse(((JObject)_js.Run($"{schemaName}.mta.json.js".ToLowerInvariant(), "transform", metadata)).Value<string>("content"));
```
928PJY: This requires that the filename of files in the template folder are also in lowercase?

yufeih: right, let's just remove ToLower for now

PR #3959, opened 2019-01-23
Refine a config test case

yufeih:
since we've renamed `content --> files`

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -5,10 +5,7 @@ inputs:
     files: docs/**/*.md
   docfx.json: |
     {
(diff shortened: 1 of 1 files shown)


docs/specs/config.yml
```diff
@@ -5,10 +5,7 @@ inputs:
     files: docs/**/*.md
   docfx.json: |
     {
-      "content":
-        {
-          "include": "ignore.md"
-        }
+      "files": { "include": "ignore.md" }
```
superyyrrzz: `"files": "ignore.md",`?

PR #3960, opened 2019-01-23
[WIP] generate dependency lock

OsmondJiang:
still in progress...

### DIFF
--- a/src/docfx/restore/lock/DependencyLock.cs
+++ b/src/docfx/restore/lock/DependencyLock.cs
@@ -10,9 +10,28 @@ namespace Microsoft.Docs.Build
 {
     internal class DependencyLock : DependencyVersion
     {
-        public Dictionary<string, DependencyLock> Git { get; set; } = new Dictionary<string, DependencyLock>();
+        public IReadOnlyDictionary<string, DependencyLock> Git { get; set; } = new Dictionary<string, DependencyLock>();
 
-        public Dictionary<string, DependencyVersion> Downloads { get; set; } = new Dictionary<string, DependencyVersion>();
+        public IReadOnlyDictionary<string, DependencyVersion> Downloads { get; set; } = new Dictionary<string, DependencyVersion>();
+
+        public DependencyLock()
+        {
+        }
+
+        public DependencyLock(IReadOnlyDictionary<string, DependencyLock> gitVersions, IReadOnlyDictionary<string, DependencyVersion> downloads, DependencyVersion version = null)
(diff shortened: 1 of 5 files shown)


test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -267,6 +272,12 @@ private static TheoryData<string> FindTestSpecs()
             }
 
             var docset = Path.Combine(inputFolder, spec.Cwd ?? string.Empty);
+
+            if (GitUtility.IsRepo(docset) && !(spec.Inputs?.Any() ?? false))
+            {
+                Process.Start(new ProcessStartInfo("git", "clean -fdx") { WorkingDirectory = docset }).WaitForExit();
```
yufeih: This is likely going to make test slower, how about removing output files starting with `~/` instead

src/docfx/restore/lock/DependencyVersion.cs
```diff
@@ -1,12 +1,20 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
```
yufeih: nit: not used

src/docfx/restore/lock/DependencyLock.cs
```diff
@@ -3,16 +3,36 @@
 
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 
 namespace Microsoft.Docs.Build
 {
     internal class DependencyLock : DependencyVersion
```
yufeih: can we have a pure data model that represents the content of .lock file, just like how we model other data files like xrefmap, publish manifest, git commits time etc.

test/docfx.Test/e2e/E2ETest.cs
OsmondJiang: not just remove, but make the change revert

yufeih: then only run this command if the output has any files with `~/`. Plus, `spec.Inputs` should never be `null`.

OsmondJiang: 👍

src/docfx/restore/lock/DependencyLock.cs
OsmondJiang: i splitted the model, will clean the constructor in another PR

Issue #3885, opened 2019-01-09
Remove hostname while resolving uid


live1206:
In v2, OPS will inform docfx while calling xref service need to remove some specific hostname.

We need to figure out when we should remove the hostname, for instance:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-2.2

Check the page source to see the href of `IRouter`, even they are sharing the same hostname, the hostname has not been removed. 

So, when should v3 remove hostname while resolving uid?

yufeih:
I'd say never.

OsmondJiang:
it's depends on the difference of host name of different environment?

yufeih:
Close this as docfx should treat urls as opaque data.

OsmondJiang:
just a reminder, we need import different xref files for different branch in prod environment, because different branch using the different host name.

live1206:
https://review.docs.microsoft.com/en-us/engineering/projects/ops/build-service/host-url-usage?branch=master#open-issue

PR #3961, opened 2019-01-23
remove host name if sharing the same one during uid resolving

live1206:
#3885

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -20,8 +20,10 @@ outputs:
     ["info","at-uid-not-found","Cannot find uid 'a' using xref '@a'","docs/a.md"]
 ---
 # Restore xref map from configured url and use the map for resolving
+# hostname should be removed if docset sharing the same one
 inputs:
   docfx.yml: |
+    baseUrl: docs.microsoft.com
     xref: 
       - xrefmap.json
   docs/a.md: Link to @System.String
@@ -37,7 +39,7 @@ inputs:
     }
 outputs:
   docs/a.json: |
-    { "content": "<p>Link to <a href=\"https://docs.microsoft.com/en-us/dotnet/api/system.string\">String</a></p>\n"}
+    { "content": "<p>Link to <a href=\"/en-us/dotnet/api/system.string\">String</a></p>\n"}
 ---
 # Restore xref map from configured url and use the YAML file for resolving
 inputs:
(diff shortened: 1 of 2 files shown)


docs/specs/xref.yml
```diff
@@ -20,8 +20,10 @@ outputs:
     ["info","at-uid-not-found","Cannot find uid 'a' using xref '@a'","docs/a.md"]
 ---
 # Restore xref map from configured url and use the map for resolving
+# hostname should be removed if docset sharing the same one
 inputs:
   docfx.yml: |
+    baseUrl: docs.microsoft.com
```
yufeih: `baseUrl` contains `https://`, so this should be 'https://docs.microsoft.com'

live1206: ok

src/docfx/build/xref/XrefMap.cs
```diff
@@ -70,6 +71,16 @@ internal class XrefMap
             // xrefSpec.displayPropertyName -> xrefSpec.name -> uid
             string display = !string.IsNullOrEmpty(displayPropertyValue) ? displayPropertyValue : (!string.IsNullOrEmpty(name) ? name : uid);
             return (null, resolvedHref, display, referencedFile);
+
+            string RemoveHostnameIfSharingTheSameOne(string input)
+            {
+                var hostname = rootFile.Docset.Config.BaseUrl;
+                if (input.StartsWith(hostname, StringComparison.OrdinalIgnoreCase))
+                {
+                    return input.Substring(input.IndexOf(hostname) + hostname.Length);
```
yufeih:
```suggestion
                    return input.Substring(hostname.Length);
```

src/docfx/build/xref/XrefMap.cs
```diff
@@ -70,6 +71,16 @@ internal class XrefMap
             // xrefSpec.displayPropertyName -> xrefSpec.name -> uid
             string display = !string.IsNullOrEmpty(displayPropertyValue) ? displayPropertyValue : (!string.IsNullOrEmpty(name) ? name : uid);
             return (null, resolvedHref, display, referencedFile);
+
+            string RemoveHostnameIfSharingTheSameOne(string input)
```
OsmondJiang: do we need consider the http or https case? same host name but different schema

live1206: since the hostname contains `http(s)`, it should match?

yufeih: That'll make things a lot complicated, url comparison is itself very complex if you also consider port, www, etc. So we just right now use simple string comparison.

PR #3963, opened 2019-01-23
Ignore empty list during JSON serialization

live1206:


docs/specs/moniker.yml
```diff
@@ -117,7 +115,7 @@ outputs:
       "files": [
           { "url": "/docs/a", "path": "218c13d0/docs/a.json", "monikers": [ "netcore-1.1" ] },
           { "url": "/docs/a", "path": "4667fedf/docs/a.json", "monikers": [ "netcore-2.0" ] },
-          { "url": "/docs/a", "path": "docs/a.json", "monikers": [] },
+          { "url": "/docs/a", "path": "docs/a.json" },
```
yufeih: `!monikers` to test moniker list is emty

docs/specs/git.yml
```diff
@@ -194,8 +194,7 @@ outputs:
   docs/index.json: |
     {
       "author": 
-        { "name": "charlie", "profile_url": "https://github.com/charlie", "display_name": "Charlie", "id": "3" },
-      "contributors": []
+        { "name": "charlie", "profile_url": "https://github.com/charlie", "display_name": "Charlie", "id": "3" }
     }
```
yufeih: `!contributors` to assert contributors are empty

src/docfx/lib/JsonUtility.cs
```diff
@@ -482,6 +482,31 @@ private static void SetFieldWritable(MemberInfo member, JsonProperty prop)
...
+            protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
+            {
+                var prop = base.CreateProperty(member, memberSerialization);
+                var contract = DefaultSerializer.ContractResolver.ResolveContract(prop.PropertyType);
+                if (contract is JsonArrayContract)
+                {
+                    prop.ShouldSerialize =
+                          value =>
+                          {
+                              var obj = JToken.FromObject(value);
```
yufeih: `JToken.FromObject` is to heavy to retrieve a single property.

src/docfx/lib/JsonUtility.cs
```diff
@@ -482,6 +482,31 @@ private static void SetFieldWritable(MemberInfo member, JsonProperty prop)
...
+                                  return false;
+                              }
+                              return true;
+                          };
+                }
+                return prop;
+            }
+        }
+
         private sealed class JsonContractResolver : DefaultContractResolver
```
yufeih: This this also derive from `SerializerContractResolver`?

live1206: intended to split these two, should deserializer be camel case?

src/docfx/lib/JsonUtility.cs
```diff
@@ -482,7 +482,32 @@ private static void SetFieldWritable(MemberInfo member, JsonProperty prop)
             }
         }
 
-        private sealed class JsonContractResolver : DefaultContractResolver
+        private class JsonContractResolver : CamelCasePropertyNamesContractResolver
+        {
+            protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
+            {
+                var prop = base.CreateProperty(member, memberSerialization);
+                var contract = DefaultSerializer.ContractResolver.ResolveContract(prop.PropertyType);
```
yufeih: nit: consider extract this to a method like `don't serialize empty array`

docs/specs/moniker.yml
```diff
@@ -100,7 +100,7 @@ outputs:
   docs/a.json: |
     {
       "content": "<p>No version</p>",
-      "monikers": []
+      "!monikers": null
```
OsmondJiang: would you please help to add test for empty Dictionary?

live1206: for json.net, Dictionary is JObject

live1206:
Calling `DefaultSerializer.ContractResolver.ResolveContract` in `GetProperty` during de-serialization, will cause infinite recusion if the data contract is self referenced like the original `TestData`.

Since this logic is only needed for serialization, palling to move it to a dedicated contract resolver for serialization

yufeih:
> Calling `DefaultSerializer.ContractResolver.ResolveContract` in `GetProperty` during de-serialization, will cause infinite recusion if the data contract is self referenced like the original `TestData`.
> 
> Since this logic is only needed for serialization, palling to move it to a dedicated contract resolver for serialization

Use a dedicated `DefaultContractResolver` to check if a type is array, that can avoid these recursion.

src/docfx/lib/JsonUtility.cs
yufeih: only do this if `ShouldSerialize` is null

live1206: https://github.com/JamesNK/Newtonsoft.Json/blob/509643a8952ce731e0207710c429ad6e67dc43db/Src/Newtonsoft.Json/Serialization/CamelCasePropertyNamesContractResolver.cs#L69
There is a cache from type to JsonContract, the effort is minor

PR #3964, opened 2019-01-23
add noloc extension

daxianji007:


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Noloc/NolocParser.cs
```diff
@@ -0,0 +1,47 @@
...
+        private const string EndString = "\" :::";
+
+        public NolocParser()
+        {
+            OpeningCharacters = new[] { ':' };
+        }
+
+        public override bool Match(InlineProcessor processor, ref StringSlice slice)
+        {
+            if (!ExtensionsHelper.MatchStart(ref slice, StartString, false))
```
yufeih: does the spec said it needs to be exactly match `::: noloc text="`? should `:::no-lock     text  = "` work?

daxianji007: This is not specified in spec, I just prefer to do it strictly. I have called it out and let`s wait for PM`s confirmation.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Noloc/NolocParser.cs
```diff
@@ -0,0 +1,47 @@
...
+            {
+                return false;
+            }
+
+            var text = ExtensionsHelper.TryGetStringBeforeChars(new char[] { '\"', '\n' }, ref slice);
```
yishengjin1413: Please think about this case: 

::: noloc text="dsadsa\
dsadsa" :::

test/Microsoft.DocAsCode.MarkdigEngine.Tests/NolocTest.cs
```diff
@@ -0,0 +1,53 @@
...
+            TestUtility.AssertEqual(expected1, source1, TestUtility.MarkupWithoutSourceInfo);
+
+            // Escape syntax
+            var source2 = "使用 :::noloc text=\"Find a \\\"Quotation\\\"\"::: 方法.";
+            var expected2 = "<p>使用 Find a \"Quotation\" 方法.</p>\n";
+            TestUtility.AssertEqual(expected2, source2, TestUtility.MarkupWithoutSourceInfo);
+
+            // Markdown in noloc
+            var source3 = @":::noloc text=""*Hello*"":::";
+            var expected3 = @"<p>*Hello*</p>
```
superyyrrzz: As the expected3 is not `<p><em>hello</em></p>`, the syntax seems more like `nomarkup` instead of `noloc` to me.

daxianji007: Yes, this is the current design reviewed by @cillroy. Full details:
* Do not render markdown syntax inside this no-loc syntax.
* Do not write multiple lines inside this no-loc syntax.
* Only lowercase allowed.
* No optional spaces allowed.
* Escaped characters are allowed ONLY in text part, :::noloc text="\"Quotation\""::: will be rendered to "Quotation". But escaped line-break character is still not allowed.
* Escaped characters in other part are not allowed.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Noloc/NolocParser.cs
daxianji007: added to tests

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Noloc/NolocParser.cs
daxianji007: The current design is no optional spaces allowed.

PR #3966, opened 2019-01-24
Change TOC pipeline to run jint post processor

yufeih:
- Revert logics in jint: `toc_title` --> `name`, `children` --> `items`
- Remove `homepage`, template has logic to fallback to href
- TOC metadata does not merge metadata from config
- Merge `TocInputModel`, `TocModel`, `LegacyTocModel` into a single `TocModel`
- Remove legacy toc conversion


src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -37,9 +37,9 @@ public static (List<Error> errors, TableOfContentsInputModel model) LoadMdTocMod
 
             var (metaErrors, metadata) = ExtractYamlHeader.Extract(file, context);
             errors.AddRange(metaErrors);
-            var tocModel = new TableOfContentsInputModel
+            var tocModel = new TableOfContentsModel
             {
-                Metadata = metadata,
+                Metadata = metadata.ToObject<TableOfContentsMetadata>(),
```
live1206: validate TOC YAML header here?

yufeih: Second thought: BuildPage needs to warn reserved metadata because we generate these additional metadata in output, but for TOC, we didn't generate these metadata. From that perspective, we may not warn here.

OsmondJiang:
toc has metadata, for A/B testing

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -34,9 +34,11 @@ internal static class BuildTableOfContents
             {
                 output = file.Docset.Template.TransformMetadata("toc.json.js", JsonUtility.ToJObject(model));
 
-                context.Output.WriteText("{}", Path.ChangeExtension(outputPath, ".mta.json"));
+                var metadata = new { experimental = model.Metadata.Experimental, experiment_id = model.Metadata.ExperimentId };
```
OsmondJiang: what about monikers?

src/docfx/build/toc/TableOfContentsItem.cs
```diff
@@ -2,37 +2,38 @@
...
 using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
     public class TableOfContentsItem
     {
-        [JsonProperty(PropertyName = "toc_title")]
```
live1206: does this need to be `SnakeCaseNamingStrategy`? since it was `toc_title`

yufeih: the other properties like `tocHref` are not using `SnakeCase`.

PR #3967, opened 2019-01-24
restore always generate and save dependency lock for build

OsmondJiang:
1. `restore` always generate and save dependency lock for build, to default/user's dependency file path
2. `build` always use dependency lock to retrieve git/file, otherwise errors will be thrown
3.  support version for downloads


OsmondJiang:
#3888

src/docfx/restore/RestoreMap.cs
```diff
@@ -51,18 +72,40 @@ string FindGitRepository()
                     return null;
                 }
 
+                // return specificed version
```
live1206: typo: specified

PR #3971, opened 2019-01-25
fix null value for template

live1206:


OsmondJiang:
https://ceapex.visualstudio.com/Engineering/_build/results?buildId=36727

PR #3980, opened 2019-01-28
add locale to dependency lock path

OsmondJiang:
fix but when build loc content from source repo


OsmondJiang:
https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/8068?_a=overview

PR #3985, opened 2019-01-29
fix build error....

OsmondJiang:
why we can check in code with errors? @yufeih ?


yufeih:
> why we can check in code with errors? @yufeih ?

To reduce PR build time, the current flow does not require a branch to be up to date before merge

PR #3988, opened 2019-01-29
omit empty enumerable value during serializing 

OsmondJiang:
as title, we want to clean up out output json files

### DIFF
--- a/src/docfx/lib/JsonUtility.cs
+++ b/src/docfx/lib/JsonUtility.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.IO;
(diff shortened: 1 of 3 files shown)


test/docfx.Test/lib/JsonUtilityTest.cs
```diff
@@ -544,6 +544,23 @@ public void SyntaxErrorShouldBeThrownWithoutSchemaValidation(string json)
...
+        public void OmitEmptyEnumerableValue()
+        {
+            var content = JsonUtility.Serialize(new EmptyEnumerable());
+            Assert.Equal("{\"a\":\"\"}", content);
+        }
+
+        public class EmptyEnumerable
+        {
+            public string A { get; set; } = "";
+            public List<string> B { get; set; } = new List<string>();
```
live1206: `List` is enough? We are not testing all types implementing `IEnumerable` anyway.
Possible to add a property with nested List? which should be omitted as well

src/docfx/lib/JsonUtility.cs
```diff
@@ -468,18 +469,25 @@ protected override JsonProperty CreateProperty(MemberInfo member, MemberSerializ
 
                 void ShouldNotSerializeEmptyArray()
                 {
-                    if (s_rawContractResolver.ResolveContract(prop.PropertyType) is JsonArrayContract)
```
live1206: minor: `s_rawContractResolver` was involved to avoid infinite recursion. Since we do not use `ResolveContract` here anymore, we can remove it and use `s_serializer.ContractResolver` instead?

src/docfx/lib/JsonUtility.cs
```diff
@@ -468,18 +468,25 @@ protected override JsonProperty CreateProperty(MemberInfo member, MemberSerializ
...
+                        target =>
+                        {
+                            var value = prop.ValueProvider.GetValue(target);
+
+                            if (value is string)
+                            {
+                                return true;
+                            }
+
+                            if (value is IEnumerable enumer && enumer != null && !enumer.GetEnumerator().MoveNext())
```
yufeih: `enum != null` is not needed, `is` operator ensures `enumer` is not null.

src/docfx/lib/JsonUtility.cs
```diff
@@ -468,18 +468,25 @@ protected override JsonProperty CreateProperty(MemberInfo member, MemberSerializ
...
-                                  {
-                                      return false;
-                                  }
-                                  return true;
-                              };
+                        target =>
+                        {
+                            var value = prop.ValueProvider.GetValue(target);
+
+                            if (value is string)
```
yufeih: move this up

PR #3992, opened 2019-01-30
add suggestion to log level

v-pegao:
Add Suggestion severity log method in Microsoft.DocAsCode.Common library as required [here](https://review.docs.microsoft.com/en-us/engineering/projects/ops/suggestion-severity?branch=master#dev-team-tasks)


src/Microsoft.DocAsCode.Common/Loggers/LogLevel.cs
```diff
@@ -8,6 +8,7 @@ public enum LogLevel
         Diagnostic = -1,
         Verbose,
         Info,
+        Suggestion,
```
herohua: shall we keep the original order, i.e. putting `Suggestion` at the end of the list?

superyyrrzz: Order here matters to make [`LogLevelThreshold`](https://github.com/dotnet/docfx/blob/a932308152152250273243968f1f5d161c23c0bd/src/Microsoft.DocAsCode.Common/Loggers/Logger.cs#L90) work.

yufeih:
@herohua , does v3 also need a Suggestion level? If so, @v-pegao , can you help add it [here](https://github.com/dotnet/docfx/blob/v3/src/docfx/lib/log/ErrorLevel.cs) as well?

PR #3994, opened 2019-01-30
get repo url/branch from environments

OsmondJiang:


src/docfx/config/AppData.cs
```diff
@@ -73,5 +77,17 @@ private static string GetAppDataRoot()
                 ? Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.UserProfile), ".docfx")
                 : Path.GetFullPath(docfxAppData);
         }
+
+        private static string GetRepositoryUrlFromEnvironment()
+        {
+            var value = Environment.GetEnvironmentVariable("DOCFX_REPOSITORY_URL");
+            if (!string.IsNullOrEmpty(value))
+            {
+                var (remote, branch) = HrefUtility.SplitGitHref(value);
```
yufeih: don't split href here, the url should be the original url as is. `branch` is coming from another environment variable.

src/docfx/config/AppData.cs
```diff
@@ -25,6 +25,10 @@ internal static class AppData
 
         public static string DefaultGitHubUserCachePath => Path.Combine(CacheRoot, "github-users.json");
 
+        public static string RepositoryUrl => GetRepositoryUrlFromEnvironment();
```
yufeih: `AppData` is where we manage the content inside `.docfx` folder, you can read environment variables directly inside `Repository`.

PR #3996, opened 2019-01-31
Log theme url and resolved theme version

yufeih:


OsmondJiang:
any reason to do that?

yufeih:
> any reason to do that?

To help troubleshoot template related problems. Recently there is a template change that escapes _description_ field, this PR helps diagnose such problems.

PR #3998, opened 2019-01-31
add items to publish to legacy manifest file

OsmondJiang:


src/docfx/build/legacy/manifest/LegacyItemToPublish.cs
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
```
live1206: [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]

PR #4006, opened 2019-02-01
make the max-exceeded-error message more clear

OsmondJiang:


src/docfx/Errors.cs
```diff
@@ -357,8 +357,8 @@ public static Error SchemaNotFound(string schema)
         /// <summary>
         /// Build errors is larger than <see cref="OutputConfig.MaxErrors"/>.
         /// </summary>
-        public static Error ExceedMaxErrors(int maxErrors)
-            => new Error(ErrorLevel.Error, "exceed-max-errors", $"Error or warning count exceed '{maxErrors}'. Build will continue but newer logs will be ignored.");
+        public static Error ExceedMaxErrors(int maxErrors, string type)
+            => new Error(ErrorLevel.Error, "exceed-max-errors", $"{type} count exceed '{maxErrors}'. Build will continue but newer {type} logs will be ignored.");
```
mingwli: This level doesn't take effect in current flow.
I'm not asking to change, just feel a little bit weird.

OsmondJiang: u r right, need check build code to see why publish with `error`

yufeih: Shall we take an `ErrorLevel` here and use it in both message and the first parameter?

src/docfx/Errors.cs
```diff
@@ -357,8 +357,8 @@ public static Error SchemaNotFound(string schema)
         /// <summary>
         /// Build errors is larger than <see cref="OutputConfig.MaxErrors"/>.
         /// </summary>
-        public static Error ExceedMaxErrors(int maxErrors)
-            => new Error(ErrorLevel.Error, "exceed-max-errors", $"Error or warning count exceed '{maxErrors}'. Build will continue but newer logs will be ignored.");
+        public static Error ExceedMaxErrors(int maxErrors, ErrorLevel level)
+            => new Error(level, "exceed-max-errors", $"{level} count exceed '{maxErrors}'. Build will continue but newer {level} logs will be ignored.");
```
yufeih: nit: the last one `level.ToLowerInvarient()`

Issue #3665, opened 2018-11-22
Cannot filter out the APIs 


Kavignon:
**DocFX Version Used**: 
2.40.2

**Template used**: (`default` or `statictoc` or contain custom template)
statistic

**Steps to Reproduce**:

1. Created a filter document (filterConfig.yml)
2. Included APIs
apiRules:

-include: 
    uidRegex: ^Kevin\.Avignon\.People\.Namespace\.Person
    type: Class
-include:
    uidRegex: ^Kevin\.Avignon\.People\.Namespace\.PeopleController\.GetPerson
    type: Method 

-exclude: 
     uidRegex: ^Kevin\.Avignon\.People
     type: Type

3. 

**Expected Behavior**:

I expect to see to the public documentation of Person class, the public documentation for GetPerson method and everything under the People namespace to be excluded.
**Actual Behavior**:
I only see the public documentation of Person.

This is a business case where I have changed the documentation content so I don't divulge any sensible information, but the gist is here.

Kavignon:
apiRules:
- include: 
    uidRegex: ^Kevin\.Avignon\.People\.Controllers\.PeopleController\.GetPersonName.*$
    type: Method
- include: 
    uidRegex: ^Kevin\.Avignon\.People\.Controllers\.PeopleController\.HairColor.*$
    type: Property
- exclude: 
    uidRegex: ^Kevin\.Avignon\.People
    type: Type

Sorry there were some mistakes in the previous message I sent I'm sorry for this @superyyrrzz 

I did consult this page and I've been consulting it while creating my filter file. 


My use case is to only document what is found in the filter file. 

I've debug the code of docFX and it would seem that if my apiRule doesn't include a class, but points to an member within that said class (i.e. Method, Property for instance), it doesn't get documented by docFX.

Most of my solution does not have to be document which is why by the end I follow what's mentioned in your documentation: 

> you need to put the most detailed rule in the top. 

Which is why I make my specific inclusion/exclusion on top of my filterConfig.yml and at the end, we can find that anything else that isn't worth mentioning should **not** be documented. 

That's the behaviour I expected from reading the documentation about filtering unwanted APIs.

superyyrrzz:
@Kavignon Thank you for the feedback. Looks like we need to improve this document.

Kavignon:
@superyyrrzz 

Something also that I'm still uneasy about it's concerning documentation of the internals of a container such as a 

- Type
- Class 
- Interface

[...]

PR #4012, opened 2019-02-01
Document specific member from filter

Kavignon:
This pull request was born out of the following issue: https://github.com/dotnet/docfx/issues/3665

**For those who don't want to go through reading the issue:** 

In a few words, I expressed my feelings and expectations from using the tool. The documentation shows that using user rules in the filtering document will supersede what's define in the default file. Albeit that may be true for types, it isn't true when a user tries to document only specific members of the type (i.e. a property for instance). What this brings on the table? More flexibility to the user. The tool already documents the entire set of members found in a type. Depending on a rule, it may ignore private content. With this, yes we need to go through the source document until we retrieve the correct member to document, but it doesn't stray from the initial implementation.

Also, there's also an issue with the way the filtering document is being used. With any technology, such as SQL or LINQ, or even with mathematical logic, a predicate has the following role: 

> A logical expression which evaluates to TRUE or FALSE, normally to direct the execution path in code
More details here: https://stackoverflow.com/questions/1344015/what-is-a-predicate

But, with DocFX, whatever isn't matched by the filtering document is also included in the resulting documentation. This shouldn't happen.

To that end, I made an update that keeps the same idea whenever a filtering document isn't provided. When a user does provide a document for filtering specific APIs or API members, then the tool will make sure to only include what is found in the filtering document and the rest that can't be matched won't be found in the resulting documentation.


superyyrrzz:
See https://github.com/dotnet/docfx/issues/3665#issuecomment-465018613

Since we have not reached agreement on the design, close this PR for now.

PR #4022, opened 2019-02-12
Update href type logic

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_workitems/edit/64298

### DIFF
--- a/src/docfx/lib/HrefUtility.cs
+++ b/src/docfx/lib/HrefUtility.cs
@@ -3,13 +3,17 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Web;
 
 namespace Microsoft.Docs.Build
 {
     internal static class HrefUtility
     {
+        private static readonly Regex s_uriWithProtocol = new Regex(@"^\w{2,}\:", RegexOptions.Compiled);
+
         /// <summary>
         /// Split href to path, fragement and query
         /// </summary>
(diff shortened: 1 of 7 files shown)


src/docfx/lib/HrefUtility.cs
```diff
@@ -98,11 +100,42 @@ public static DependencyType FragmentToDependencyType(string fragment)
...
+
+            if (Uri.TryCreate(href, UriKind.Absolute, out _))
+            {
+                return HrefType.External;
+            }
+
+            if (char.IsLetter(ch) && href.Contains(':'))
+            {
+                return HrefType.External;
+            }
```
OsmondJiang: what's is this for if we already have Uri.TryCreate above?

live1206: To cover the test case
```csharp
[InlineData("feedback-url:?query=a", HrefType.External)]
```

OsmondJiang: is this a url with port?

live1206: this will launch a windows application

src/docfx/lib/HrefUtility.cs
```diff
@@ -98,11 +100,43 @@ public static DependencyType FragmentToDependencyType(string fragment)
...
+            {
+                return HrefType.AbsolutePath;
+            }
+
+            if (Uri.TryCreate(href, UriKind.Absolute, out _))
+            {
+                return HrefType.External;
+            }
+
+            // Uri.TryCreate does not handle some common errors like http:docs.com, so specialize them here
```
OsmondJiang: humm, do we really want to eat these common errors ?

PR #4023, opened 2019-02-12
Html decode before extracting title

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_workitems/edit/66469

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -221,15 +221,15 @@ inputs:
     ---
     metaKey: metaValue
     ---
-    # This is *a* title
+    # This is *a* title &
     
     body
 outputs:
   docs/a.json: |
     {
       "content": "<p>body</p>",
-      "title": "This is a title",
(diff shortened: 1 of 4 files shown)


docs/specs/markdown.yml
```diff
@@ -221,15 +221,15 @@ inputs:
...
+    # This is *a* title &
     
     body
 outputs:
   docs/a.json: |
     {
       "content": "<p>body</p>",
-      "title": "This is a title",
-      "raw_title": "<h1 id=\"this-is-a-title\">This is <em>a</em> title</h1>",
+      "title": "This is a title &",
```
OsmondJiang: where is the *?

yufeih: * is treated as markdown `<em>`, so it is not part of the text

OsmondJiang: so what will be encoded/decoded, what will be markup? a little confused

PR #4029, opened 2019-02-13
cache toc model, avoid to load twice for building toc and building toc map

OsmondJiang:
as title, we want to avoid toc model loading twice.

split toc model loading and moniker calculation to pure toc model loading for caching 
simplify the moniker calculation for toc

### DIFF
--- a/src/docfx/build/context/Cache.cs
+++ b/src/docfx/build/context/Cache.cs
@@ -41,12 +41,10 @@ namespace Microsoft.Docs.Build
             })).Value;
 
         public (List<Error>, TableOfContentsModel, List<(Document doc, string herf)>, List<Document>) LoadTocModel(Context context, Document file)
-        {
-            return _tocModelCache.GetOrAdd(
+            => _tocModelCache.GetOrAdd(
                 file.FilePath,
                 new Lazy<(List<Error>, TableOfContentsModel, List<(Document doc, string herf)>, List<Document>)>(
                     () => BuildTableOfContents.Load(context, file))).Value;
-        }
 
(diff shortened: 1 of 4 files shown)


src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -17,7 +18,25 @@ internal static class BuildTableOfContents
             Debug.Assert(file.ContentType == ContentType.TableOfContents);
             Debug.Assert(monikerMap != null);
 
-            var (errors, model, refArticles, refTocs) = Load(context, file, monikerMap);
+            // load toc model
+            var hrefMap = new Dictionary<string, List<string>>();
+            var (errors, model, refArticles, refTocs) = context.Cache.LoadTocModel(context, file);
+            foreach (var (doc, href) in refArticles)
+            {
+                if (!hrefMap.TryGetValue(href, out _) && monikerMap.TryGetValue(doc, out var monikers))
```
live1206: minor
```csharp
if (!hrefMap.ContainsKey(href) && monikerMap.TryGetValue(doc, out var monikers))
```

src/docfx/build/context/Cache.cs
```diff
@@ -39,6 +40,14 @@ internal class Cache
                 }
             })).Value;
 
+        public (List<Error>, TableOfContentsModel, List<(Document doc, string herf)>, List<Document>) LoadTocModel(Context context, Document file)
+        {
+            return _tocModelCache.GetOrAdd(
```
mingwli: => _tocModelCache.GetOrAdd(...,...).Value;
more consistent with the rest?

Issue #3995, opened 2019-01-31
restore and gc improvement 


OsmondJiang:
The `restore` is to **add and forget** and `gc` is to **delete**, considering the work **tree size** and check-out speed, we may want to **re-use** instead of **delete**.

re-use existing work-tree of git can help us reduce the disk size and improve the checkout speed, this actually also fits the dependency files.

But this solution need track usage status of each work tree, not a status-less design.

OsmondJiang:
work items:
- [ ] read/write lock cross process
- [ ] pooling model of git work tree
- [ ] pooling model of downloaded file

yufeih:
Do we still need pooling for downloading files?

fenxu:
If the architecture is totally same, reuse that is not bad. However, if it needs more effort, it may not be necessary.

OsmondJiang:
totally same

OsmondJiang:
and in-process reader writer lock doesn't async/await, which doesn't meet our requirement, i am going to first implement an simple shared/exclusive lock using file system.

@yufeih  @fenxu  @herohua

PR #4030, opened 2019-02-13
Restore design spec update

OsmondJiang:
update restore spec to:
1. remove gc logic
2. create reusable download file and work tree
3. using file to tracking each file/work tree detail info


OsmondJiang:
#3995

Issue #3887, opened 2019-01-09
URL decode before link to or include a file


superyyrrzz:
For a file `a b.md`, it should use `[sample](a%20b.md)` to link to it, or use `[!include[](a%20b.md)` to include it. The decode logic is in Markdig's `ReadFile` or `GetLink` delegate or by post processing `MarkupResult`. 

Currently URL decode support:

| syntax | v2 | v3 |
|-|-|-|
| link |  ✅ | ❎   |
| include | ❎ | ❎  |

TODO:

- [ ] URL Decode in v2's `ReadFile` @daxianji007 
- [ ] URL Decode in v3's `ReadFile`
- [x] URL Decode in v3's `GetLink` (needs Q1)

/cc @928PJY

herohua:
This seems like a question we could get PM's input.

yufeih:
Related https://ceapex.visualstudio.com/Engineering/_workitems/edit/66496

yufeih:
What is canonical URL and url in publish manifest for file `a b.md` ?  `a b` or `a%20b`

herohua:
@yufeih to double confirm v3 behavior first

PR #4046, opened 2019-02-18
Url decode before resolving links

yufeih:
Implements #3887 
Fixes https://ceapex.visualstudio.com/Engineering/_workitems/edit/66496/

I'll look at bookmark and query string escape handling in another PR


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -123,7 +123,8 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
                 return (uidError, uidHref, null, referencedFile);
             }
 
-            var (error, file, redirectTo, query, fragment, isSelfBookmark, _) = TryResolveFile(relativeTo, href);
+            var decodedHref = HttpUtility.UrlDecode(href);
```
superyyrrzz: Why not HtmlDecode here? In my understanding, it needs urldecode+htmldecode to get file from href.

yufeih: html decode/encode is handled by markdig, our custom `HtmlUtility.TransformLinks` handles html encode/decode itself.

You can verify the behavior from tests, `a&b` and `a&amp;b` all links to `a%26b`

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -367,7 +367,8 @@ private static void VerifyFile(string file, string content)
                     if (!string.IsNullOrEmpty(content))
                     {
                         TestUtility.VerifyJsonContainEquals(
-                            JToken.Parse(content),
+                            // Test expectation can use YAML for readability
```
live1206: 👍

Issue #4041, opened 2019-02-14
Build VBA-docs TOC take too long


mingwli:
[build report](https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C2%5C14%5Ce2344c78-2e2d-592e-9f3d-af54656882a1%5CCommit%5C201902140546262313-live%5Cworkflow_report.html?sv=2015-02-21&sr=b&sig=6cF4PnWWETpQf%2FMRarKRyGSh684nKI1rZiqK2w3XPTw%3D&st=2019-02-14T06%3A59%3A50Z&se=2019-03-17T07%3A04%3A50Z&sp=r)  

info:

| Total Files | 39063 |
| --- | --- |
| Total Content Files | 38583 |
| Total Resource Files | 478 |
| Total Toc Files | 2 |

log: 
```text
Restore dependencies done in 619.15ms
Globbing files done in 428.27ms
Building Xref map done in 4.38s
** Loading TOC done in 00:10:33.7840000 **
Building files done in 00:14:42.4260000
Convert Legacy Manifest done in 595.19ms
Convert Legacy Dependency Map done in 1.41s
Convert Legacy File Map done in 1.95s
Converting to legacy done in 3.97s
Done in 00:25:27.9280000
```

yufeih:
It is likely because the TOC is huge, and we didn't customize the TOC markdown parser. We could try shrink the TOC markdown pipeline to only include those essential parsers.

Issue #4037, opened 2019-02-14
Build performance hot spot improvements


yufeih:
A quick performance analysis shows the following hot spots areas that affects build performance:

- [ ] Low CPU usage:

![image](https://user-images.githubusercontent.com/511355/52760910-42ed3380-304c-11e9-9c7b-35ae7d042994.png)

- [ ] Jint is taking more than half of page build time:

![image](https://user-images.githubusercontent.com/511355/52760845-0588a600-304c-11e9-86d8-d17c1f54b4f1.png)

- [ ] TabGroupExtension markdown post processor:

![image](https://user-images.githubusercontent.com/511355/52759828-50a0ba00-3048-11e9-8566-9006cdc21285.png)

- [ ] Resource hash calculation

![image](https://user-images.githubusercontent.com/511355/52759941-b2612400-3048-11e9-85a3-07248fc94901.png)

yufeih:
Closed in favor of https://dev.azure.com/ceapex/Engineering/_workitems/edit/113226/

PR #4048, opened 2019-02-20
Optimize ParseHeading by removing linq

yufeih:
`ParseHeading` method is executed for each heading in each article, the LINQ expression causes too much allocation and slows markdown significantly.

#4037 #4041  cc @qinezh


superyyrrzz:
FYI @yishengjin1413 I think you are investigating Markdig performance issue.

PR #4056, opened 2019-02-20
use source repo's branch

mingwli:
when contribution config doesn't specify repository, use current's source repo's

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -390,7 +390,7 @@ outputs:
       "gitcommit": "https://github.com/testowner/testrepo/blob/e662dbc7c5e582a14b863d6a6a82e495d6272f91/docs/index.md",
     }
 ---
-# Use default `master` branch when edit branch isn't specified in contribution config,
+# Use current repo's branch when edit branch isn't specified in contribution config,
 # even contribution config's repository doesn't equal to current repo's
 repos:
   https://github.com/testowner/testrepo#default-building-branch:
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -371,3 +371,39 @@ repos:
 outputs:
   build.log: |
     ["error","committish-not-found","Cannot find branch, tag or commit 'live' for repo '*github.com+git+repo.zh-cn*","a.md"]
+---
+# Use current repo's branch when edit branch isn't specified in contribution config,
+# and contribution config's repository equals to current repo's
+repos:
+  https://github.com/testowner/testrepo#building-branch:
+    - files:
+        docs/index.md: Alice creates this.
+inputs:
+  docfx.yml: |
+    contribution:
+      repository: https://github.com/testowner/testrepo
+outputs:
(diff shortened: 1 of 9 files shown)


mingwli:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/67390

docs/specs/git.yml
```diff
@@ -371,3 +371,39 @@ repos:
...
+      repository: https://github.com/testowner/testrepo
+outputs:
+  docs/index.json: |
+    {
+      "content_git_url": "https://github.com/testowner/testrepo/blob/building-branch/docs/index.md",
+      "original_content_git_url": "https://github.com/testowner/testrepo/blob/building-branch/docs/index.md",
+      "gitcommit": "https://github.com/testowner/testrepo/blob/e662dbc7c5e582a14b863d6a6a82e495d6272f91/docs/index.md",
+    }
+---
+# Use default `master` branch when edit branch isn't specified in contribution config,
```
yufeih:
```suggestion
# Use current repo's branch when edit branch isn't specified in contribution config,
```

src/docfx/lib/HrefUtility.cs
```diff
@@ -80,15 +80,16 @@ public static string MergeHref(string targetHref, string sourceQuery, string sou
         /// Get the git remote information from remote href
         /// </summary>
         /// <param name="remoteHref">The git remote href like https://github.com/dotnet/docfx#master</param>
-        public static (string remote, string refspec) SplitGitHref(string remoteHref)
+        public static (string remote, string refspec, bool hasRefSpec) SplitGitHref(string remoteHref)
```
yufeih: Can check from caller
```suggestion
        public static (string remote, string refspec) SplitGitHref(string remoteHref)
```

docs/specs/git.yml
```diff
@@ -371,3 +371,39 @@ repos:
 outputs:
   build.log: |
     ["error","committish-not-found","Cannot find branch, tag or commit 'live' for repo '*github.com+git+repo.zh-cn*","a.md"]
+---
+# Use current repo's branch when edit branch isn't specified in contribution config,
```
OsmondJiang: would you please add spec test in localization spec or add todo there?

src/docfx/localization/LocalizationUtility.cs
```diff
@@ -248,6 +248,12 @@ private static string GetLocalizationBranch(LocalizationMapping mapping, string
                 return sourceBranch;
             }
 
+            if (TryRemoveLocale(sourceBranch, out string sourceBranchWithoutLocale, out string branchLocale) &&
```
OsmondJiang: why need change that? For contribution url, we already always get contribution branch, check [here](https://github.com/dotnet/docfx/blob/55900c0b8ffd345adbb7271d17800c137c2f7b34/src/docfx/build/metadata/ContributionProvider.cs#L210)

PR #4061, opened 2019-02-21
🔒support cross process read write lock

OsmondJiang:
1. implement init cross process shared/exclusive lock.
2. add unit tests
3. add default expired time, in case lock holder crashed

> NOTE: All acquiring to shared/exclusive lock returns immediately, without waiting, the return value indicate whether lock is acquired or not. 

>NOTE: Since we using asynchronous programming model, the thread info can't be used for tracking acquired lock, we are using `acquirer` info returned from AcquireXXXLock, these info must be provided to ReleaseXXXLock.

generate way to implement share/exclusive lock
```c#
//process lock

//ref count

//process lock release
```

### DIFF
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -246,27 +246,17 @@ namespace Microsoft.Docs.Build
         {
             await RunInsideMutex(path, async () =>
             {
-                var content = File.Exists(path) ? File.ReadAllText(path) : string.Empty;
-                var result = JsonUtility.Deserialize<T>(content);
+                using (var file = File.Open(path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.None))
+                {
+                    var streamReader = new StreamReader(file);
+                    var result = JsonUtility.Deserialize<T>(streamReader.ReadToEnd());
 
-                var updatedResult = await update(result);
-
-                File.WriteAllText(path, JsonUtility.Serialize(updatedResult));
-            });
-        }
-
-        /// <summary>
-        /// Reads the content of a file, update content and write back to file in one atomic operation
-        /// </summary>
-        public static async Task ReadAndWriteFile(string path, Func<string, Task<string>> update)
-        {
-            await RunInsideMutex(path, async () =>
-            {
-                var content = File.Exists(path) ? File.ReadAllText(path) : null;
-
-                var updatedContent = await update(content);
-
-                File.WriteAllText(path, updatedContent);
+                    file.Position = 0;
+                    var updatedResult = await update(result);
+                    var steamWriter = new StreamWriter(file);
+                    steamWriter.Write(JsonUtility.Serialize(updatedResult));
+                    steamWriter.Close();
+                }
             });
         }
 
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -18,7 +18,7 @@ namespace Microsoft.Docs.Build
     /// </summary>
     internal static class ProcessUtility
     {
-        private const int _defaultLockExpireTimeInSecond = 60 * 20;
+        private const int _defaultLockExpireTimeInSecond = 60 * 60 * 6; /*six hours*/
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/docfx/lib/ProcessUtility.cs
```diff
@@ -16,6 +18,147 @@ namespace Microsoft.Docs.Build
     /// </summary>
     internal static class ProcessUtility
     {
+        private const int _defaultLockExpireTimeInSecond = 60 * 20;
```
yufeih: 20min may be to small

src/docfx/lib/ProcessUtility.cs
```diff
@@ -16,6 +18,147 @@ namespace Microsoft.Docs.Build
...
+
+        /// <summary>
+        /// Acquire a shared lock for input lock name
+        /// The returned `acquirer` are used for tracking the acquired lock, instead of thread info, since the thread info may change in asynchronous programming model
+        /// </summary>
+        public static async Task<(bool acquired, string acquirer)> AcquireSharedLock(string lockName)
```
yufeih: just return a bool state makes it harder for others to use this lock, caller have to wait and retry. There are 2 patterns in .NET: [wait with a timeout](https://docs.microsoft.com/en-us/dotnet/api/system.threading.waithandle.waitone?view=netframework-4.7.2#System_Threading_WaitHandle_WaitOne_System_Int32_) or [TryEnter](https://docs.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim.tryenterreadlock?view=netframework-4.7.2)

src/docfx/lib/ProcessUtility.cs
```diff
@@ -251,5 +415,47 @@ private static async Task<T> RetryUntilSucceed<T>(string name, Func<Exception, b
...
+            None,
+            Shared,
+            Exclusive,
+        }
+
+        private class LockInfo
+        {
+            public LockType Type { get; set; }
+
+            public List<Acquirer> AcquiredBy { get; set; } = new List<Acquirer>();
```
yufeih: how about `readers` because acquired could both mean acquired for read and acquired for write.

src/docfx/lib/ProcessUtility.cs
```diff
@@ -251,5 +415,47 @@ private static async Task<T> RetryUntilSucceed<T>(string name, Func<Exception, b
...
+        private enum LockType
+        {
+            None,
+            Shared,
+            Exclusive,
+        }
+
+        private class LockInfo
+        {
+            public LockType Type { get; set; }
```
yufeih: If the process crashed while holding a write lock, this state is leaked and docfx will hang forever.

test/docfx.Test/common/TestFramework.cs
```diff
@@ -32,6 +32,8 @@ protected override ITestFrameworkExecutor CreateExecutor(AssemblyName assemblyNa
             Log.ForceVerbose = true;
 
             MakeDebugAssertThrowException();
+            if (Directory.Exists(AppData.MutexRoot))
```
yufeih: Is this intentional?

src/docfx/lib/ProcessUtility.cs
OsmondJiang: current requirement doesn't need wait, we can do that later if you have the requirments,

src/docfx/lib/ProcessUtility.cs
OsmondJiang: i would like to keep the acquirer  info even it's write lock acquiring, debugging would be easy

test/docfx.Test/common/TestFramework.cs
OsmondJiang: yes, to cleanup mutex folder for my unit tests

src/docfx/lib/ProcessUtility.cs
OsmondJiang: no, it will be cleaned by expiring

test/docfx.Test/common/TestFramework.cs
OsmondJiang: will remove this line

Issue #3981, opened 2019-01-28
Add application insight telemetry


yufeih:
Implement telemetry based on application insight to track usage and statistics:

- Write telemetry to AI specified by [APPINSIGHTS_INSTRUMENTATIONKEY](https://github.com/Microsoft/ApplicationInsights-dotnet/blob/eda4a8ef821298613d67096db88779ba600d1c23/src/Microsoft.ApplicationInsights/Extensibility/Implementation/TelemetryConfigurationFactory.cs#L23) (supported natively by AI).
- Telemetry is opt-in only. To opt-in, set `APPINSIGHTS_INSTRUMENTATIONKEY`.
- Use [pre-aggregated metrics](https://docs.microsoft.com/en-us/azure/azure-monitor/app/pre-aggregated-metrics-log-metrics) to capture full telemetry without sampling.
- Dimensions: _version_, _repo_, _branch_, _os_

Work items:

- [x] Track aggregated operation duration
- [x] Track crash exception
- [x] Track errors and warnings
- [x] Track cache hit rate: _git commit cache_, _repo cache_, _github user cache_
- [x] Provisioning and service integration (https://ceapex.visualstudio.com/Engineering/_git/OpenPublishing.Build/pullrequest/8888?_a=files)

yufeih:
> What about adding a default embedded AI key to track community users information? Of course, we can disable it by default.

👍  I plain to add that after:
- going through all the compliant process
- ready for community release

PR #4062, opened 2019-02-21
📈Add basic application insight telemetry

yufeih:
#3981


src/docfx/lib/log/Telemetry.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        public static void TrackErrorCount(string code, ErrorLevel level)
+        {
+            s_telemetryClient
+                .GetMetric(new MetricIdentifier(null, $"error", "code", "level", "os", "version", "repo", "branch"))
+                .TrackValue(1, code, level.ToString(), s_os, s_version, s_repo, s_branch);
+        }
+
+        public static void TrackCacheTotal(CacheName name)
```
OsmondJiang: what's that means?

OsmondJiang: or TrackCacheHit?

src/docfx/restore/Restore.cs
```diff
@@ -22,6 +22,8 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
             using (Progress.Start("Restore dependencies"))
             {
                 var repository = Repository.Create(docsetPath);
+                Telemetry.SetRepository(repository?.Remote, repository?.Branch);
```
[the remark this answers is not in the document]
yufeih: this is to help answer questions like what is the average build time for azure? so only root repo is needed.

src/docfx/lib/log/Progress.cs
```diff
@@ -67,9 +67,11 @@ public void Dispose()
             {
                 t_scope.Value = t_scope.Value.Pop(out var scope);
 
-                var elapsedMs = Stopwatch.ElapsedMilliseconds;
+                var elapsed = Stopwatch.Elapsed;
+                var elapsedMs = elapsed.TotalMilliseconds;
                 if (Log.Verbose || elapsedMs > ProgressDelayMs)
                 {
+                    Telemetry.TrackOperationDuration(Name, elapsed);
```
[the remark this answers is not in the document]
yufeih: Right. I think we can first disable this telemetry until we need more details in the future.

src/docfx/lib/log/Telemetry.cs
yufeih: Tracking Hit/Miss count is hard then tracking Miss/Total count, see GitHubUser cache as an example, there is no clear if/else switch, but it is easier to find all total accesses and missed accesses.

Issue #4039, opened 2019-02-14
Dependency missing for inclusion/link chain


live1206:
`a.md` includes `b.md`, and then `b.md` includes `c.md`
```yml
inputs:
  docfx.yml:
  docs/a.md: a [!INCLUDE[](b.md)]
  docs/b.md: b [!INCLUDE[](c.md)]
  docs/c.md: c
```

In the output `.dependencymap.json`, dependency from `a.md` to `c.md` is missing
```json
{
  "dependencies": {
    "docs/b.md": [
      {
        "source": "docs/c.md",
        "type": "inclusion"
      }
    ],
    "docs/a.md": [
      {
        "source": "docs/b.md",
        "type": "inclusion"
      }
    ]
  }
}
```
The same problem with link chain also exists.

live1206:
we need to think about how to flatten the dependency map, for instance
```
     include            link
a      --->    b, b    ---> c
```
what would be the dependency type between a and c?
And also, do we need link dependency type since the change of c would not cause the content change of b

PR #4063, opened 2019-02-21
🌏Flatten dependency map

live1206:
#4039 
According to [v2](https://github.com/dotnet/docfx/blob/6375541889bc9559b7991961260f0d10938edfd1/src/Microsoft.DocAsCode.Build.Engine/Incrementals/DependencyGraph.cs#L21), only dependency type `inclusion`, `overwrite` and `overWriteFragment` would transit, v3 only has `inclusion` for now.


docs/specs/markdown.yml
```diff
@@ -400,4 +400,137 @@ outputs:
...
\ No newline at end of file
+---
+# inclusion chain
+inputs:
+  docfx.yml:
+  docs/a.md: a [!INCLUDE[](b.md)]
+  docs/b.md: b [!INCLUDE[](c.md)]
+  docs/c.md: c
+outputs:
+  docs/a.json: |
```
OsmondJiang: no need to check for your purpose

docs/specs/markdown.yml
```diff
@@ -400,4 +400,137 @@ outputs:
...
+          "type": "inclusion"
+        },
+        {
+          "source": "docs/a.md",
+          "type": "link"
+        }
+      ]
+    }
+   }
+---
```
OsmondJiang: add test:

a -> link b -> include c

src/docfx/build/dependency/DependencyMapBuilder.cs
```diff
@@ -34,13 +35,52 @@ public void AddDependencyItem(Document from, Document to, DependencyType type)
...
+            var graph = _dependencyItems
+                .GroupBy(k => k.From)
                 .ToDictionary(
                     k => k.Key,
-                    v => (from r in v.Distinct()
-                         orderby r.Dest.FilePath, r.Type
-                         select r).ToList()));
+                    v => (from r in v
+                          orderby r.To.FilePath, r.Type
+                          select r).ToHashSet());
```
OsmondJiang: you need cache the calculated nodes, otherwise a lot of nodes will be re-calculated multiple times.

like that ?
```C# 
flattenCache

GetFlattenItems(key, graph, items)
{
    return flattenCache.GetOrAdd(key, => 
    {
        var flattenItems = items.ToList();
        foreach (var item in items)
        {
            if (graph.ContainsKey(item) && item != key && CanTransit)
            {
                flattenItems.AddRange(GetFlattenItems(item, graph, graph[item]))
            }
        }
        return flattenItems;
    });
}

var flattenGrap = graph.ToDictionary();
foreach(var (from, items) in graph)
{
    flattenGrap[from] = GetFlattenItems(graph, items);
}
```

live1206: good point, we can reuse the calculated result

src/docfx/build/dependency/DependencyMapBuilder.cs
```diff
@@ -34,13 +35,63 @@ public void AddDependencyItem(Document from, Document to, DependencyType type)
...
+            var result = new Dictionary<Document, HashSet<DependencyItem>>();
+            var graph = _dependencyItems
+                .GroupBy(k => k.From)
                 .ToDictionary(
                     k => k.Key,
-                    v => (from r in v.Distinct()
-                         orderby r.Dest.FilePath, r.Type
-                         select r).ToList()));
+                    v => (from r in v
+                          orderby r.To.FilePath, r.Type
```
yufeih: Just like `Dictionary`, `HashSet` also preserves insertion order, so the result is sorted and deterministic.

src/docfx/build/dependency/DependencyMapBuilder.cs
```diff
@@ -34,13 +35,63 @@ public void AddDependencyItem(Document from, Document to, DependencyType type)
...
+
+        private void Transit(Document from, Dictionary<Document, HashSet<DependencyItem>> graph, DependencyItem current, DependencyItem next, Dictionary<Document, HashSet<DependencyItem>> dependencies)
+        {
+            if (!CanTransit(current))
+            {
+                return;
+            }
+
+            // if the dependency destination is already in the result set, we can reuse it
+            if (next.To != from && dependencies.ContainsKey(next.To))
```
yufeih: Use `TryGetValue` to save another lookup
```suggestion
            if (next.To != from && dependencies.TryGetValue(next.To, out var xxx))
```

src/docfx/build/dependency/DependencyMapBuilder.cs
live1206: yes, found that, so `order by` can be removed

src/docfx/build/dependency/DependencyMapBuilder.cs
```diff
@@ -34,13 +35,63 @@ public void AddDependencyItem(Document from, Document to, DependencyType type)
...
+            if (next.To != from && dependencies.TryGetValue(next.To, out var nextDependencies))
+            {
+                foreach (var item in nextDependencies)
+                {
+                    dependencies[from].Add(new DependencyItem(from, item.To, item.Type));
+                }
+                return;
+            }
+
+            if (!graph.ContainsKey(next.To) || next.To.Equals(from))
```
yufeih: need a hash set to detect recursion.

OsmondJiang: recursion should  be avoid by `CanTransit` and detected during building?

Issue #4047, opened 2019-02-19
Use named mutex for cross process locking


yufeih:
Right now we are using file locks to synchronize cross process access to a shared resource. The current implementation is not robust on linux/mac, a dangling lock file is left on file system whenever the process is crashed or terminated by `Ctrl+C`.

We could instead use global named mutex, it works across platform and survives process crash.

OsmondJiang:
```
new FileStream(
                    lockPath,
                    FileMode.CreateNew,
                    FileAccess.ReadWrite,
                    FileShare.None,
                    1,
                    FileOptions.DeleteOnClose)
```

we are using DeleteOnClose, so the file should be deleted even the docfx crashed, which platform are you testing?

OsmondJiang:
i see, you are testing on linux and mac

OsmondJiang:
windows can always clean the file: https://stackoverflow.com/questions/324045/windows-temporary-files-behaviour-are-they-deleted-by-the-system

OsmondJiang:
https://stackoverflow.com/questions/46302756/in-dotnet-core-how-can-i-ensure-only-one-copy-of-my-application-is-running?answertab=votes#tab-top

yufeih:
The platforms we are targeting are windows, mac (authoring environment) and linux (potential production environment). Unix is currently not a target.

OsmondJiang:
if we change it to that way, we may also want to have Shared Lock and Exclusive Lock.

PR #4068, opened 2019-02-22
🔒Use globally named mutex over file locking

yufeih:
Implements #4047 

`AbandonedMutexException` is thrown as is because when it occurs, it typically indicates some data corruption.


src/docfx/lib/ProcessUtility.cs
```diff
@@ -285,40 +285,39 @@ public static void WriteFile(string path, Action<Stream> write)
...
+                // avoid nested calls to RunInsideMutex with same mutex name
+                t_mutexRecursionStack.Value = t_mutexRecursionStack.Value ?? ImmutableStack<string>.Empty;
+                if (t_mutexRecursionStack.Value.Contains(mutexName))
+                {
+                    throw new ApplicationException($"Nested call to RunInsideMutex is detected, mutex name: {mutexName}");
+                }
+                t_mutexRecursionStack.Value = t_mutexRecursionStack.Value.Push(mutexName);
 
-                using (RetryUntilSucceed(mutexName, IsFileAlreadyExistsException, CreateFile))
+                try
```
OsmondJiang: that looks better😁

PR #4071, opened 2019-02-25
💬 TOC.md supports xml comment

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/8891?_a=files&path=%2Fsource%2Fazure-docs-pr.ja-jp.output%2Fbuild.log

### DIFF
--- a/src/docfx/build/toc/parser/MarkdownTocMarkup.cs
+++ b/src/docfx/build/toc/parser/MarkdownTocMarkup.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Text;
 using Markdig.Extensions.Yaml;
 using Markdig.Syntax;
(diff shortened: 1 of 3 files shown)


src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -15,24 +14,26 @@ namespace Microsoft.Docs.Build
 {
     internal static class MarkdownTocMarkup
     {
-        private static readonly HashSet<Type> s_blockWhiteList = new HashSet<Type> { typeof(HeadingBlock) /*header*/, typeof(YamlFrontMatterBlock) /*yaml header*/, typeof(HtmlBlock) /*comment*/ };
```
OsmondJiang: why need change that? we should only need add html block parser back?

yufeih: There are other types of html block types, we only support comments.

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
```
live1206: seems like this is still needed?

yufeih: yes, introduced by accident

Issue #4020, opened 2019-02-11
Support docs conceptual template


yufeih:
We've settled down to use the existing template system, this issue tracks the missing work needed to support conceptual documents, it does not include schema driven documents and conceptual yaml files going through yaml document processor pipeline.

- [x] Run `toc.json.js` for TOC pipeline.
- [x] Confirm and porting remaining logic in `conceptual.html.primary.js`
- [x] Read localized strings from `/LocalizedTokens/docs({locale}).html/tokens.json`, it is right now reading from razor resources.
- [x] ~~Model `PageModel` to be the input of `{schema}.template.js`. The goal is to minimize template transform as much as possible.~~

PR #4073, opened 2019-02-25
🌐Reading markdown token from template

yufeih:
#4020


src/docfx/docset/Docset.cs
```diff
@@ -299,6 +290,20 @@ private HashSet<Document> CreateBuildScope(IEnumerable<Document> redirections)
             }
         }
 
+        private TemplateEngine CreateTemplate()
+        {
+            if (string.IsNullOrEmpty(Config.Theme))
+            {
+                return new TemplateEngine(Path.Combine(AppContext.BaseDirectory, "theme"), Locale);
```
OsmondJiang: what's that?

OsmondJiang: what's that?

yufeih: Just a fake template

OsmondJiang: this will throw undefined exception for docs repos, so just return null before we have real default theme?

src/docfx/lib/JsonUtility.cs
```diff
@@ -286,20 +286,15 @@ public static (List<Error>, JToken) RemoveNulls(this JToken token)
             return (errors, token);
         }
 
-        public static bool TryGetValue<T>(this JObject obj, string key, out T value) where T : JToken
+        public static bool TryGetValue<T>(this JObject obj, string key, out T value)
         {
-            value = null;
-            if (obj == null || string.IsNullOrEmpty(key))
+            if (obj.TryGetValue(key, out var valueToken) && valueToken is JValue jvalue && jvalue.Value is T result)
```
OsmondJiang: `valueToken is JValue` is too restrict? the method name can't tell me that i can't use this method is not a JValue

src/docfx/docset/Docset.cs
OsmondJiang: this will throw undefined exception for docs repos, so just return null before we have real default theme?

PR #4076, opened 2019-02-25
🔒remove sub dep lock support and change test

OsmondJiang:
simplify restore with lock

### DIFF
--- a/src/docfx/docset/Docset.cs
+++ b/src/docfx/docset/Docset.cs
@@ -43,11 +43,6 @@ namespace Microsoft.Docs.Build
         public bool Legacy => _options.Legacy;
 
         /// <summary>
-        /// Gets the dependent docsets
-        /// </summary>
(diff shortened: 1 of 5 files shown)


src/docfx/docset/Docset.cs
```diff
@@ -312,7 +312,6 @@ private static async Task<(List<Error>, Dictionary<string, Docset>)> LoadDepende
                 var (loadErrors, subConfig) = ConfigLoader.TryLoad(dir, options, locale);
                 errors.AddRange(loadErrors);
 
-                subLock = subLock ?? await Docs.Build.DependencyLock.Load(dir, subConfig.DependencyLock);
                 result.TryAdd(PathUtility.NormalizeFolder(name), await Create(report, dir, locale, subConfig, options, subLock, isDependency: true));
```
yufeih: This method could then be sync

live1206: there is still `await Create` in the next line

PR #4079, opened 2019-02-26
🥽using pooling model to restore git dependencies

OsmondJiang:
#3995 

Restore:
- Acquire a exclusive slot(id)
- Restoring(create or checkout work tree)
- Release exclusive slot(id) with restored flag

Build:
- Acquire a shared slot(id) with flag(restored == true)
- Building
- Release shared slot(id)

There are somethings we need keep in mind:

- Exclusive/Shared lock has a default expire time, which means the lock can't be held in one owner for too long(default is 6 hours) in case it crashed.

- Once slot is restored, there is a lockdown time for this slot to be used for another restore, in case the slot is just restored but not used for build yet.


src/docfx/restore/index/DependencySlotPool.cs
```diff
@@ -0,0 +1,211 @@
...
+                                    {
+                                        slot = i;
+                                        slot.Acquirer = acquirer;
+                                        break;
+                                    }
+                                }
+                            }
+                            break;
+                        default:
+                            throw new NotSupportedException($"{type} is not supported");
```
yufeih: just remove `LockType.None`?

src/docfx/restore/index/DependencySlotPool.cs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        public static List<T> GetSlots<T>(string restoreDir) where T : DependencySlot
+        {
+            Debug.Assert(!string.IsNullOrEmpty(restoreDir));
+
+            var slotsFile = Path.Combine(restoreDir, "index.json");
+            var content = File.Exists(slotsFile) ? File.ReadAllText(slotsFile) : string.Empty;
+
+            return JsonUtility.Deserialize<List<T>>(content) ?? new List<T>();
```
yufeih: consider use a wrapper object like other data models, it is hard to extend an array.

src/docfx/restore/index/DependencyGitPool.cs
```diff
@@ -0,0 +1,175 @@
...
+                remote,
+                type,
+                slot =>
+                {
+                    // update branch and commit info to new rented slot
+                    slot.Commit = commit;
+                    slot.Branch = branch;
+                    return slot;
+                },
+                existingSlot => existingSlot.Branch == branch && existingSlot.Commit == commit);
```
yufeih: `commit` alone is enough to identify a slot, since we are not looking up `branch` from repo.

src/docfx/restore/index/DependencySlot.cs
```diff
@@ -0,0 +1,22 @@
...
+        public string Id { get; set; }
+
+        public string Url { get; set; }
+
+        public DateTime LastAccessDate { get; set; }
+
+        public bool Restored { get; set; }
+
+        [JsonIgnore]
+        public string Acquirer { get; set; }
```
yufeih: Is there a way to transfer acquirer without polluting this data model?

src/docfx/restore/index/DependencySlot.cs
```diff
@@ -0,0 +1,22 @@
...
+{
+    internal abstract class DependencySlot
+    {
+        public string Id { get; set; }
+
+        public string Url { get; set; }
+
+        public DateTime LastAccessDate { get; set; }
+
+        public bool Restored { get; set; }
```
yufeih: `Commit` is enough to identify if a slot is good

src/docfx/restore/index/DependencySlot.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    internal abstract class DependencySlot
+    {
+        public string Id { get; set; }
+
+        public string Url { get; set; }
+
+        public DateTime LastAccessDate { get; set; }
```
yufeih: why is there a need to store time?

src/docfx/lib/ProcessUtility.cs
```diff
@@ -275,12 +282,15 @@ public static async Task<string> ReadFile(string path)
...
+        public static Task<T> ReadFile<T>(string path)
+            => ReadFile(path, stream => JsonUtility.Deserialize<T>(new StreamReader(stream).ReadToEnd()));
+
         public static async Task<T> ReadFile<T>(string path, Func<Stream, T> read)
         {
             T result = default;
             await RunInsideMutex(path, () =>
             {
-                using (var fs = new FileStream(path, FileMode.Open, FileAccess.Read, FileShare.Read, 1024, FileOptions.SequentialScan))
+                using (var fs = new FileStream(path, FileMode.OpenOrCreate, FileAccess.Read, FileShare.Read, 1024, FileOptions.SequentialScan))
```
yufeih: This change the semantic of `ReadFile`, it should not attempt to create a file.

src/docfx/lib/ProcessUtility.cs
```diff
@@ -326,23 +336,20 @@ public static Task WriteFile(string path, Action<Stream> write)
...
         {
             Debug.Assert(!string.IsNullOrEmpty(mutexName));
+            var lockPath = Path.Combine(AppData.MutexRoot, HashUtility.GetMd5Hash(mutexName));
 
-            // avoid the RunInsideMutex to be nested used
-            // doesn't support to require a lock before releasing a lock
-            // which may cause deadlock
-            if (t_innerCall.Value != null)
+            // avoid the RunInsideMutex to be nested used which cause deadlock
+            if (t_innerCall.Value != null && t_innerCall.Value == lockPath)
```
yufeih: can this be done in another PR? to allow fine grained nest detection? This should be a stack/list to detect deadlocks.

src/docfx/restore/lock/DependencyLock.cs
```diff
@@ -10,21 +10,21 @@ namespace Microsoft.Docs.Build
 {
     internal static class DependencyLock
     {
-        public static DependencyLockModel GetGitLock(this DependencyLockModel dependencyLock, string href, string branch)
+        public static (string key, DependencyLockModel value) GetGitLock(this DependencyLockModel dependencyLock, string href, string branch)
```
yufeih: consider not include this change in this PR.

src/docfx/restore/index/DependencyGitPool.cs
OsmondJiang: not enough, we have `@implicit` feature which based on branch.

src/docfx/restore/index/DependencySlot.cs
OsmondJiang: it's last access data, used for ordering and apply lockdown time

src/docfx/restore/index/DependencySlot.cs
OsmondJiang: not enough, i will change the `commit` at the start of restoring.

src/docfx/restore/index/DependencySlot.cs
OsmondJiang: i would like to keep it with slot. becase slot need to be released anyway.

src/docfx/lib/ProcessUtility.cs
OsmondJiang: ok

src/docfx/lib/ProcessUtility.cs
OsmondJiang: will revert that

src/docfx/restore/index/DependencySlotPool.cs
OsmondJiang: ok

src/docfx/build/Build.cs
```diff
@@ -220,15 +234,16 @@ private static async Task<DependencyLockModel> LoadBuildDependencyLock(string do
             return dependencyLock ?? new DependencyLockModel();
         }
 
-        private static (List<Error> errors, Config config) GetBuildConfig(string docset, Repository repository, CommandLineOptions options, DependencyLockModel dependencyLock)
+        private static (List<Error> errors, Config config) GetBuildConfig(string docset, Repository repository, CommandLineOptions options, DependencyLockModel dependencyLock, DependencyGitPool dependencyGitPool)
         {
             if (ConfigLoader.TryGetConfigPath(docset, out _) || !LocalizationUtility.TryGetSourceRepository(repository, out var sourceRemote, out var sourceBranch, out var locale))
             {
-                return ConfigLoader.Load(docset, options);
+                var (errors, config) = ConfigLoader.Load(docset, options);
```
OsmondJiang: revert this changes

src/docfx/docset/Docset.cs
```diff
@@ -177,15 +182,15 @@ internal class Docset
                 Debug.Assert(!string.IsNullOrEmpty(Config.Theme));
 
                 var (themeRemote, themeBranch) = LocalizationUtility.GetLocalizedTheme(Config.Theme, Locale, Config.Localization.DefaultLocale);
-                var (themePath, themeLock) = RestoreMap.GetGitRestorePath($"{themeRemote}#{themeBranch}", DependencyLock);
+                var (themePath, themeLock) = _dependencyGitPool.AcquireSharedGit($"{themeRemote}#{themeBranch}", DependencyLock);
```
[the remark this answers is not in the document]
OsmondJiang: will change back to 'GetGitRestorePath'

src/docfx/restore/index/DependencySlotPool.cs
OsmondJiang: ok

src/docfx/restore/lock/DependencyLock.cs
OsmondJiang: i will try to not change the structure

src/docfx/restore/RestoreGit.cs
```diff
@@ -144,20 +148,40 @@ async Task AddWorkTrees()
...
+                                    throw Errors.GitCloneFailed(remote, branches).ToException(ex);
+                                }
                             }
-                            catch (Exception ex)
+                            else
                             {
-                                throw Errors.GitCloneFailed(remote, branches).ToException(ex);
+                                // worktree already exists
+                                // checkout to {headCommit}, no need to fetch
+                                await GitUtility.Checkout(workTreePath, headCommit);
```
yufeih: Is add worktree enough to update `workTreePath` to `headCommit`? since we are calling `worktree add --force`

src/docfx/restore/index/DependencySlot.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
+{
+    internal abstract class DependencySlot
+    {
+        public string Id { get; set; }
+
+        public string Url { get; set; }
```
yufeih: Is `Url` just like `Acquirer` for internal tracking purpose? It does not need to persist on disk.

src/docfx/restore/index/DependencyGitPool.cs
```diff
@@ -0,0 +1,176 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Docs.Build
+{
+    internal class DependencyGitPool
+    {
+        private Dictionary<(string remote, string branch, string commit, LockType lockType), (string path, DependencyGit git, bool released)> _acquiredGits = new Dictionary<(string remote, string branch, string commit, LockType lockType), (string path, DependencyGit git, bool released)>();
```
yufeih: can this dictionary be `readonly IReadOnlyDictionary`, to separate mutating operations like `AcquireXXX` from non mutating (thread safe) operations like `GetXXX` and `ReleaseXXX`. Like builder patterns used in other places, mutating operations returns an immutable instance.

src/docfx/restore/index/DependencySlotPool.cs
```diff
@@ -0,0 +1,218 @@
...
+
+                    WriteSlots(restoreDir, slots);
+                });
+
+            Debug.Assert(released);
+
+            return released;
+        }
+
+        public static async Task<(string path, T slot)> AcquireSlots<T>(string url, LockType type, Func<T, T> updateExistingSlot, Func<T, bool> matchExistingSlot) where T : DependencySlot, new()
```
mingwli:
```suggestion
        public static async Task<(string path, T slot)> AcquireSlot<T>(string url, LockType type, Func<T, T> updateExistingSlot, Func<T, bool> matchExistingSlot) where T : DependencySlot, new()
```

src/docfx/restore/index/DependencySlot.cs
OsmondJiang: you r right.

src/docfx/restore/index/DependencyGitPool.cs
OsmondJiang: great idea, try try

src/docfx/restore/index/DependencySlotPool.cs
OsmondJiang: 赞！

src/docfx/restore/RestoreGit.cs
OsmondJiang: no, it will throw work tree already exists error, because the work tree name is not changed

src/docfx/restore/RestoreMap.cs
```diff
@@ -12,100 +12,8 @@ namespace Microsoft.Docs.Build
 {
     internal static class RestoreMap
```
mingwli: >RestoreMap [](start = 26, length = 10)

Only restore-file-related logic in this class, would it be better to merge it with RestoreFile?

OsmondJiang: i will change `restore file` part in another pr

src/docfx/build/Build.cs
```diff
@@ -15,22 +15,35 @@ internal static class Build
...
-            XrefMap xrefMap = null;
-
             var repository = Repository.Create(docsetPath);
             Telemetry.SetRepository(repository?.Remote, repository?.Branch);
 
             var locale = LocalizationUtility.GetLocale(repository, options);
             var dependencyLock = await LoadBuildDependencyLock(docsetPath, locale, repository, options);
-            var (configErrors, config) = GetBuildConfig(docsetPath, repository, options, dependencyLock);
+
+            var dependencyGitPool = await DependencyGitPool.AcquireGitPool(dependencyLock);
```
live1206: Does `AcquireGitPool` mean `Create`?

src/docfx/restore/index/DependencyGitPool.cs
```diff
@@ -0,0 +1,175 @@
...
+            catch
+            {
+                successed = false;
+                throw;
+            }
+            finally
+            {
+                if (!successed && root)
+                {
+                    foreach (var (k, v) in acquired)
```
live1206: Can we just call `Release` here?
Since `AcquireGitPool` is supposed to be invoked only once to get the instance?

OsmondJiang: offline discussed

src/docfx/lib/ProcessUtility.cs
```diff
@@ -22,6 +22,25 @@ internal static class ProcessUtility
         private const int _defaultLockExpireTimeInSecond = 60 * 60 * 6; /*six hours*/
         private static AsyncLocal<ImmutableStack<string>> t_innerCall = new AsyncLocal<ImmutableStack<string>>();
 
+        public static async Task<bool> IsExclusiveLockHeld(string lockName)
+        {
```
live1206: update UT to check this?

src/docfx/restore/index/DependencySlotPool.cs
```diff
@@ -0,0 +1,219 @@
...
+            var released = false;
+            await ProcessUtility.RunInsideMutex(
+                url + "/index.json",
+                async () =>
+                {
+                    var slots = GetSlots<T>(restoreDir);
+                    var slotsToRelease = slots.Where(i => i.Id == slot.Id);
+                    Debug.Assert(slotsToRelease.Count() == 1);
+
+                    var slotToRelease = slotsToRelease.First();
```
live1206: use `slotsToRelease.Single()` to get rid of `Debug.Assert` above and below

src/docfx/restore/index/DependencySlotPool.cs
```diff
@@ -0,0 +1,219 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Microsoft.Docs.Build
+{
+    internal static class DependencySlotPool
+    {
+        private static int _defaultLockdownTimeInSecond = 10 * 60;
+
+        public static async Task<(string path, T slot)> TryGetSlot<T>(string remote, Func<IReadOnlyList<T>, IReadOnlyList<T>> matchExistingSlots) where T : DependencySlot
```
live1206: possible to put `<T>` on class level to get rid of `where` constraint on every method?
if not, do we need where `T : new ()` here?

OsmondJiang: we need`T :  new()` here to new a T.

PR #4080, opened 2019-02-26
upgrade markdig package

yishengjin1413:


yishengjin1413:
For the test case change: Please refer to the change log https://github.com/lunet-io/markdig/blob/master/changelog.md .

Output attached attributes to a <tr> from a table row ((PR #300))

PR #4084, opened 2019-02-26
🌀make the nested detection more correct for RunInsideMutex

OsmondJiang:
it's a dependency change for #4079

### DIFF
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -328,13 +328,11 @@ namespace Microsoft.Docs.Build
 
             var lockPath = Path.Combine(AppData.MutexRoot, HashUtility.GetMd5Hash(mutexName));
 
-            // avoid the RunInsideMutex to be nested used
-            // doesn't support to require a lock before releasing a lock
-            // which may cause deadlock
+            // avoid the RunInsideMutex to be nested used with same mutex name
             t_innerCall.Value = t_innerCall.Value ?? new HashSet<string>();
             if (!t_innerCall.Value.Add(lockPath))
             {
-                throw new NotImplementedException($"Nested call to RunInsideMutex is not detected, {mutexName}");
+                throw new ApplicationException($"Nested call to RunInsideMutex is detected, mutex name: {mutexName}");
             }
 
             try
--- a/src/docfx/lib/ProcessUtility.cs
+++ b/src/docfx/lib/ProcessUtility.cs
@@ -325,18 +325,17 @@ namespace Microsoft.Docs.Build
         public static async Task RunInsideMutex(string mutexName, Func<Task> action)
         {
             Debug.Assert(!string.IsNullOrEmpty(mutexName));
-
             var lockPath = Path.Combine(AppData.MutexRoot, HashUtility.GetMd5Hash(mutexName));
-
-            // avoid the RunInsideMutex to be nested used with same mutex name
-            t_innerCall.Value = t_innerCall.Value ?? new HashSet<string>();
-            if (!t_innerCall.Value.Add(lockPath))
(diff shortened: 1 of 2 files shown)


src/docfx/lib/ProcessUtility.cs
```diff
@@ -327,22 +326,19 @@ public static async Task RunInsideMutex(string mutexName, Func<Task> action)
...
 
-            // avoid the RunInsideMutex to be nested used
-            // doesn't support to require a lock before releasing a lock
-            // which may cause deadlock
-            if (t_innerCall.Value != null)
+            var lockPath = Path.Combine(AppData.MutexRoot, HashUtility.GetMd5Hash(mutexName));
+
+            // avoid the RunInsideMutex to be nested used with same mutex name
+            t_innerCall.Value = t_innerCall.Value ?? new HashSet<string>();
+            if (!t_innerCall.Value.Add(lockPath))
```
live1206: put `Add` action into `try` block? since there is a `remove` in `finally` block

test/docfx.Test/lib/ProcessUtilityTest.cs
```diff
@@ -65,22 +65,34 @@ public static async Task FileMutexTest()
...
         {
-            // works for one lock one file
-            await ProcessUtility.RunInsideMutex($"process-test/{Guid.NewGuid()}", async () => { await Task.Delay(100); });
-            await ProcessUtility.RunInsideMutex($"process-test/{Guid.NewGuid()}", async () => { await Task.Delay(100); });
-
-            // doesn't work for requiring a lock before releasing a lock
-            await Assert.ThrowsAnyAsync<Exception>(async () =>
-            {
-                await ProcessUtility.RunInsideMutex($"process-test/{Guid.NewGuid()}",
+            // nested run works for different names
```
live1206: shall we split this into 2 test cases, it would be easier to detect which one is failing

src/docfx/lib/ProcessUtility.cs
OsmondJiang: ok!

test/docfx.Test/lib/ProcessUtilityTest.cs
OsmondJiang: great idea

src/docfx/lib/ProcessUtility.cs
```diff
@@ -326,22 +325,18 @@ public static Task WriteFile(string path, Action<Stream> write)
...
-            }
-
-            t_innerCall.Value = new object();
+            var lockPath = Path.Combine(AppData.MutexRoot, HashUtility.GetMd5Hash(mutexName));
 
             try
             {
-                Directory.CreateDirectory(AppData.MutexRoot);
+                // avoid the RunInsideMutex to be nested used with same mutex name
+                t_innerCall.Value = t_innerCall.Value ?? new HashSet<string>();
```
yufeih: nit: `AsyncLocal` constructor has takes a factory where you can initialize an empty hashset, if you do so, this check is no needed.

src/docfx/lib/ProcessUtility.cs
```diff
@@ -326,22 +325,18 @@ public static Task WriteFile(string path, Action<Stream> write)
...
+                if (!t_innerCall.Value.Add(lockPath))
+                {
+                    throw new ApplicationException($"Nested call to RunInsideMutex is detected, mutex name: {mutexName}");
```
yufeih: nit: moving this line inside `try` will cause `finally` to remove `lockPath` in case of exception.

OsmondJiang: it's ok, remove just return false if the lockPath can't be found.

PR #4097, opened 2019-03-01
only find up-level toc, same behavior like v2 

OsmondJiang:


src/docfx/build/toc/TableOfContentsMap.cs
```diff
@@ -75,7 +75,7 @@ public Document GetNearestToc(Document file)
             var fileNames = Path.GetFileNameWithoutExtension(file.SitePath).Split(wordSplitChars, StringSplitOptions.RemoveEmptyEntries);
             var tocCandidates = from toc in filteredTocs
                                 let dirInfo = GetRelativeDirectoryInfo(file, toc)
-                                where hasReferencedTocs || dirInfo.parentDirectoryCount >= dirInfo.subDirectoryCount
+                                where hasReferencedTocs || (dirInfo.parentDirectoryCount >= dirInfo.subDirectoryCount && dirInfo.subDirectoryCount == 0)
```
mingwli:
```suggestion
                                where hasReferencedTocs || dirInfo.subDirectoryCount == 0
```
parentDirectoryCount is always positive xD

yufeih: maybe note some where that this is due to `breadcrumb`

PR #4098, opened 2019-03-01
🧹clean up version support for file downloading

OsmondJiang:
this is the first step to support dependency file -> clean up versioning support for file downloading

right now our file system doesn't provide any file version supporting, so we are going to:

1. always download file to one fixed place, re-use this place.
2. always add process lock to before reading/writing to this file
3. file should be released after content being loaded to process
4. same url/file should always be loaded once and used forever during this `build`,  in case the file changed in the second loading.

### DIFF
--- a/src/docfx/restore/lock/DependencyLockModel.cs
+++ b/src/docfx/restore/lock/DependencyLockModel.cs
@@ -8,7 +8,5 @@ namespace Microsoft.Docs.Build
     internal class DependencyLockModel : DependencyVersion
     {
         public IReadOnlyDictionary<string, DependencyLockModel> Git { get; set; } = new Dictionary<string, DependencyLockModel>();
-
-        public IReadOnlyDictionary<string, DependencyVersion> Downloads { get; set; } = new Dictionary<string, DependencyVersion>();
     }
 }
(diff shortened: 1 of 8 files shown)


mingwli:
does [restore-and-gc.md](https://github.com/dotnet/docfx/blob/v3/docs/designs/restore-and-gc.md) needs to be updated accordingly?

OsmondJiang:
> does [restore-and-gc.md](https://github.com/dotnet/docfx/blob/v3/docs/designs/restore-and-gc.md) need to be updated accordingly?

There is a note:

image

Anyway i will update this spec later.

src/docfx/restore/lock/DependencyLockModel.cs
```diff
@@ -8,7 +8,5 @@ namespace Microsoft.Docs.Build
     internal class DependencyLockModel : DependencyVersion
```
yufeih: also remove `DependencyVersion.Hash`, then it feels `DependencyVersion` could be consolidated into `DependencyLockModel`

yufeih: Also we don't need to calculate hash in  https://github.com/OsmondJiang/docfx/blob/3cd291a04546a14956fd9fabd19689a80b9ff135/src/docfx/restore/RestoreFile.cs#L53

OsmondJiang: agree with your first point, for the second, we need keep that until we change the file restore logic(always restore to one place for one url)

PR #4099, opened 2019-03-01
🎰make dependency-map and op_aggregated_file_map_info deterministic

live1206:
Then we can remove the sorting of `.dependency-map.json` and `op_aggregated_file_map_info.json` from the impact tool


OsmondJiang:
it's better to find out why the sort in `normalize.json` doesn't work

PR #4102, opened 2019-03-04
🧹continually clean up restore file

OsmondJiang:
prepare work for restoring file


OsmondJiang:
#4098

OsmondJiang:
#4102 #3995

src/docfx/lib/ProcessUtility.cs
```diff
@@ -453,6 +462,23 @@ private static LockInfo FilterExpiredAcquirers(LockInfo lockInfo)
             return lockInfo;
         }
 
+        private static Task<(bool, string)> AcquireWithRetry(string lockName, Func<Task<(bool, string)>> acquireLock)
+        {
+            return RetryUntilSucceed(lockName, ex => ex is ApplicationException, ThrowOnNotAcquired);
```
mingwli: >ThrowOnNotAcquired [](start = 81, length = 18)

This name is a bit confusing, it's like a callback to invoke when acquireLock failed

src/docfx/lib/ProcessUtility.cs
```diff
@@ -45,32 +45,36 @@ public static async Task<bool> IsExclusiveLockHeld(string lockName)
...
-                if (lockInfo.Type == LockType.Exclusive)
+            async Task<(bool, string)> AcquireLock()
+            {
+                var acquired = false;
+                var acquirer = (string)null;
+                await ReadAndWriteFile<LockInfo>(lockPath, lockInfo =>
                 {
-                    return Task.FromResult(lockInfo);
-                }
+                    lockInfo = lockInfo ?? new LockInfo();
```
live1206: put this line into `FilterExpiredAcquirers`? then we can get rid of the null assertion

src/docfx/lib/ProcessUtility.cs
```diff
@@ -45,32 +45,36 @@ public static async Task<bool> IsExclusiveLockHeld(string lockName)
         /// Acquire a shared lock for input lock name
         /// The returned `acquirer` are used for tracking the acquired lock, instead of thread info, since the thread info may change in asynchronous programming model
         /// </summary>
-        public static async Task<(bool acquired, string acquirer)> AcquireSharedLock(string lockName)
+        public static Task<(bool acquired, string acquirer)> AcquireSharedLock(string lockName, bool retryUntilSuccess = false)
```
live1206: any test coverage for the input `retryUntilSuccess` true?

src/docfx/lib/ProcessUtility.cs
```diff
@@ -114,32 +118,37 @@ public static async Task<bool> ReleaseSharedLock(string lockName, string acquire
...
-            });
+                    if (lockInfo.Type != LockType.None)
+                    {
+                        return Task.FromResult(lockInfo);
+                    }
+
+                    acquirer = Guid.NewGuid().ToString();
+                    acquired = true;
+                    lockInfo.Type = LockType.Exclusive;
+                    lockInfo.AcquiredBy = new List<Acquirer> { new Acquirer { Id = acquirer, Date = DateTime.UtcNow } };
```
live1206: Is `lockInfo.AcquiredBy.Add()` enough here instead of allocating a new list?
And possible to consolidate `AcquireLock` with the one in `AcquireSharedLock`?

test/docfx.Test/lib/ProcessUtilityTest.cs
```diff
@@ -227,5 +227,32 @@ public static async Task RunSharedAndExclusiveLock(string[] steps, bool[] result
...
+                Assert.True(acquired);
+
+                await Task.Delay(exclusiveWaitingTime);
+
+                Assert.True(await ProcessUtility.ReleaseExclusiveLock(name, acquirer));
+            }
+
+            async Task AcquireSharedLock(int sharedWaitingTime)
+            {
+                var (acquired, acquirer) = await ProcessUtility.AcquireExclusiveLock(name, true);
```
live1206: should this be `ProcessUtility.AcquireSharedLock`?

src/docfx/lib/ProcessUtility.cs
OsmondJiang: yes, changed

src/docfx/lib/ProcessUtility.cs
OsmondJiang: filter  is to filter existing timeout lock, not to create a new lock info.

src/docfx/lib/ProcessUtility.cs
OsmondJiang: I would like to keep them splitted, the shared logic and different logic are half and half

PR #4106, opened 2019-03-04
⬇🗄always restore one url to one place

OsmondJiang:
1. always return one ulr to one place
2. remove etag from file name
3. add process lock when reading/writing restored file

will rename RestoreMap later

PR #4108, opened 2019-03-05
🧹clean up todo and gc in restore

OsmondJiang:
no need gc and cleanup some todos 

depends on #4106


src/docfx/cli/Program.cs
```diff
@@ -77,10 +77,6 @@ internal static async Task<int> Run(string[] args)
                             await Restore.Run(docset, options, report, @implicit: true);
                             await Watch.Run(docset, options);
                             break;
```
yufeih: also remove from README.md

PR #4120, opened 2019-03-06
fix dependencyType when contains query string

superyyrrzz:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/71858/


src/Microsoft.DocAsCode.Build.Engine/CompilePhaseHandlerWithIncremental.cs
```diff
@@ -303,6 +291,11 @@ private IEnumerable<DependencyItem> GetUidDependencyCore(FileModel model)
             }
         }
 
+        private bool IsBookmark(LinkSourceInfo info)
+        {
+            return info.Anchor?.Contains("#") == true;
```
herohua: do we think `a#` is a bookmark?

superyyrrzz: Seems no... updated

PR #4123, opened 2019-03-06
feat: add throttling to suggestion level logs

herohua:
@superyyrrzz @928PJY @fenxu

### DIFF
--- a/src/Microsoft.DocAsCode.Common/Loggers/Logger.cs
+++ b/src/Microsoft.DocAsCode.Common/Loggers/Logger.cs
@@ -9,13 +9,16 @@ namespace Microsoft.DocAsCode.Common
     public static class Logger
     {
         public const int WarningThrottling = 10000;
+        public const int SuggestionThrottling = 10000;
         public static bool HasError { get; private set; }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Common/Loggers/Logger.cs
```diff
@@ -117,6 +120,26 @@ public static void Log(ILogItem item)
                 }
             }
 
+            if (item.LogLevel == LogLevel.Suggestion)
```
superyyrrzz: The logic seems similar with above. What about extract a method?

superyyrrzz:
To make suggestion replay work properly, it also need to update this:
https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs#L307

However, the side effect is that if suggestion count exceeds throttling currently, it will always disable incremental build.

herohua:
We won't add the throttling for suggestion for now as it would invalidate incremental cache. We will revisit the issue when suggestion starts to be massively showing up for all repos.

Issue #4104, opened 2019-03-04
treat 'null' in yaml header as string


mingwli:
As per [yaml spec](https://yaml.org/spec/1.2/spec.html#id2805712), `null` should be resolved as a null value, while v3 currently treating it as string scalar.

yufeih:
Also consider `.inf, -.Inf, +.INF, .NAN` as mentioned in that spec. Do you want to contribute a fix?

PR #4126, opened 2019-03-07
📄handle YAML special string value

mingwli:
#4104

### DIFF
--- a/src/docfx/lib/YamlUtility.cs
+++ b/src/docfx/lib/YamlUtility.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Reflection;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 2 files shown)


src/docfx/lib/YamlUtility.cs
```diff
@@ -144,23 +145,21 @@ private static JToken ToJson(YamlNode node)
...
                     {
                         return PopulateLineInfoToJToken(new JValue(n), node);
                     }
-                    if (double.TryParse(scalar.Value, out var d))
+                    if (double.TryParse(scalar.Value, NumberStyles.Any, CultureInfo.InvariantCulture, out var d))
                     {
                         return PopulateLineInfoToJToken(new JValue(d), node);
                     }
-                    if (bool.TryParse(scalar.Value, out var b))
+                    if (bool.TryParse(scalar.Value,out var b))
```
yufeih: <space>

mingwli:
The explicit null is resolved as null value in v2.
https://raw.githubusercontent.com/openpublish/docfx-migration-to-v3/null-value/docfx-v3-migration/schema-syntax-check/others/null-value/explicit/a.md
image

test/docfx.Test/lib/YamlUtilityTest.cs
```diff
@@ -158,6 +158,37 @@ public void TestBoolean()
             Assert.Equal(new[] { true, true, true, false, false, false }, value2.Select(j => (bool)j).ToArray());
         }
 
+        [Fact]
+        public void TestNull()
```
live1206: better to use `InlineData` to specify input and output? would be easier to add more test cases

mingwli: updated

PR #4127, opened 2019-03-07
Expose markdown line info through MarkdownContext

yufeih:
To support line info in error messages, test case in #4130 

#2874

PR #4130, opened 2019-03-08
📄Add line/column info to markdown errors

yufeih:
Implements https://ceapex.visualstudio.com/Engineering/_workitems/edit/72242/
Depending on v2 change #4127 
Not all errors has line info, this PR only adds line info errors raised as part of markdown parsing.

#2874

### DIFF
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -108,9 +108,9 @@ outputs:
   docs/a.json: |
     { "content": "<p><a href=\"\">Empty</a><a href=\"a%3C%3Eb.md\">Invalid</a><a href=\"c.md\">Missing</a></p>" }
   build.log: |
-    ["info","link-is-empty","Link is empty","docs/a.md"]
-    ["warning","file-not-found","Cannot find file 'a<>b.md' relative to 'docs/a.md'","docs/a.md"]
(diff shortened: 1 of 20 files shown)


docs/specs/resolve.yml
```diff
@@ -237,7 +237,7 @@ outputs:
   docs/a.json: |
     {"content":"<p>Link to <a href=\"../dep/logo.svg\">dep</a></p>"}
   build.log: |
-    ["warning","link-is-dependency","File 'logo.svg' referenced by link '../dep/logo.svg' will not be built because it is from a dependency docset","docs/a.md"]
+    ["warning","link-is-dependency","File 'logo.svg' referenced by link '../dep/logo.svg' will not be built because it is from a dependency docset","docs/a.md","",1,9]
```
OsmondJiang: why is there empty string for most of the error message?

yufeih: It is `jsonpath`, we could consider turning it into an object

OsmondJiang: let's add a comment or todo :)

yufeih: https://github.com/dotnet/docfx/issues/4141

PR #4131, opened 2019-03-08
📝Add suggestion level

mingwli:
#3838

### DIFF
--- a/src/docfx/lib/log/ErrorLevel.cs
+++ b/src/docfx/lib/log/ErrorLevel.cs
@@ -7,6 +7,7 @@ namespace Microsoft.Docs.Build
     {
         Off,
         Info,
+        Suggestion,
         Warning,
         Error,
     }
--- a/schemas/docfx.json
+++ b/schemas/docfx.json
@@ -216,6 +216,7 @@
         "enum": [
           "off",
           "info",
+          "suggestion",
           "warning",
           "error"
         ]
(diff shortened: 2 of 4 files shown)


src/docfx/lib/log/Error.cs
```diff
@@ -13,7 +13,7 @@ internal class Error
     {
         public static readonly IEqualityComparer<Error> Comparer = new EqualityComparer();
 
-        public ErrorLevel Level { get; private set; }
+        public ErrorLevel Level { get; set; }
```
yufeih:
```suggestion
        public ErrorLevel Level { get; }
```

src/docfx/lib/log/ErrorLevel.cs
```diff
@@ -7,6 +7,7 @@ internal enum ErrorLevel
     {
         Off,
         Info,
+        Suggestion,
```
yufeih: Also needs to update Report.cs to limit suggestion count

src/docfx/lib/log/Error.cs
```diff
@@ -65,7 +65,6 @@ public string ToString(ErrorLevel level)
 
         public DocfxException ToException(Exception innerException = null)
         {
-            Level = ErrorLevel.Error;
```
live1206: without public 'set', this line could not be removed

live1206: should this be kept?

PR #4134, opened 2019-03-08
📈 Limit telemetry flush time to 2s

yufeih:
We've seen problems with flushing telemetry

### DIFF
--- a/src/docfx/lib/log/Telemetry.cs
+++ b/src/docfx/lib/log/Telemetry.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using System.Runtime.InteropServices;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.Metrics;
--- a/src/docfx/lib/log/Telemetry.cs
+++ b/src/docfx/lib/log/Telemetry.cs
@@ -5,6 +5,8 @@ using System;
 using System.Collections.Generic;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.ApplicationInsights;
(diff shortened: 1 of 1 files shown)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -63,8 +64,9 @@ public static void TrackException(Exception ex)
 
         public static void Flush()
         {
-            // Default timeout of 100 sec is used
-            s_telemetryClient.Flush();
+            // Default timeout of TelemetryClient.Flush is 100 seconds,
+            // but we only want to wait for 2 seconds at most.
+            Task.WaitAny(Task.Run(s_telemetryClient.Flush), Task.Delay(2000));
```
OsmondJiang: what will happen if it's time out?

yufeih: nothing, just hand there doing nothing for 100s

OsmondJiang: https://github.com/Microsoft/ApplicationInsights-dotnet/issues/407

OsmondJiang: will we lost any traces here if it's timeout?

herohua: Nice catch, @OsmondJiang ! I just recalled in the legacy `TraceEx` library, there is also a workaround to sleep for 10s.

herohua: do we need to swallow exceptions here?

yufeih: It is swallowed and printed at `Program.cs`

yufeih: @OsmondJiang @herohua Flush is sync when using `InMemoryChannel`, which is the default channel used in docfx, so we don't have the problem mentioned in that issue. This change is just to [give Flush a timeout](https://github.com/Microsoft/ApplicationInsights-dotnet/issues/1088)

PR #4146, opened 2019-03-11
resolve absolute file path as empty string

mingwli:
#3838 

resolve absoute file path as empty href.
rename `HrefType.Bookmark` to `HrefType.SelfBookmark`


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -137,6 +137,8 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
             // Cannot resolve the file, leave href as is
             if (file == null)
             {
+                if (hrefType == HrefType.WindowsAbsolutePath)
```
yufeih: move this outside `file == null` ?

src/docfx/lib/GlobUtility.cs
```diff
@@ -29,17 +29,18 @@ bool IsMatch(string path)
                     return false;
                 }
 
+                foreach (var exclude in excludeGlobs)
```
yufeih: nit: turn this change into another PR?

mingwli: yep, forgot to revert it.

PR #4147, opened 2019-03-11
glob apply exclude first

mingwli:


src/docfx/lib/GlobUtility.cs
```diff
@@ -29,17 +29,18 @@ bool IsMatch(string path)
                     return false;
                 }
 
+                foreach (var exclude in excludeGlobs)
+                {
```
OsmondJiang: add a test for that?

PR #4148, opened 2019-03-12
⏬ Keep using HtmlAgilityPack 1.4.9 during migration

yufeih:
Fixes HTML normalization problems caused by different HtmlAgilityPack version: https://ceapex.visualstudio.com/Engineering/_workitems/edit/73033/

A post migration solution might be to keep the original HTML intact when we do html transformation.


src/docfx/docfx.csproj
```diff
@@ -8,6 +8,7 @@
     <RootNamespace>Microsoft.Docs.Build</RootNamespace>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackAsTool>true</PackAsTool>
+    <NoWarn>NU1701</NoWarn>
```
mingwli: can this only apply to `HtmlAgilityPack`?

yufeih: sadly, no

Issue #4032, opened 2019-02-13
Support resolve of Git info for Azure DevOps repos


herohua:

OsmondJiang:
init investigation about the work items:

-  contribution provider
  - contributors (test pass)
  - author (test pass)
  - updated at (test pass)
  - content git url (new feature)
  - original content git url (new feature)
  - git commit url (new feature)

- Restore Git Dependency(test pass)

- Localization Convention(test pass)

- Migration tool(@928PJY )

- Configuration Migration tool(@mingwli )

@fenxu  @herohua  @yufeih  @superyyrrzz , please help to add more if I miss anything here.

OsmondJiang:
# Restore Git Dependency
-  We still using the same git url convention {repo_url}#{branch} like `https://ceapex.visualstudio.com/Engineering/_git/OpenPublishing.Build#master`
- Spec test(Restore + Lock) need to be added to make sure it's supported

# Localization Convention
- The localization convention will follow the same rules defined [here](https://github.com/dotnet/docfx/blob/v3/docs/designs/conceptual-localization.md#loc-content-as-assets-of-source-repo)
- Spec test(Repo/Branch/Folder/Bilingual) need to be added

# Contribution
- `Contributors/Author` will be resolved from committer's email using GITHUB cache/API.
- `UpdatedAt` should be the same logic
- `OriginalContentGitUrl/GitCommitUrl/OriginalContentGitUrlTemplate` should be new url VSTS format
- `ContentGitUrl` should be null if the contribution configuration is not setup?

# Configuration Migration Tool
Should be working, test should be added

# Migration Tool
Unknown states

# Build Service
-[ ] token(resolveUserProfile/CloneTemplate/Restore)
-[ ] localization overwrite config

herohua:
pending confirmation on priority

herohua:
we may need to consider #3173 when doing design around `Contributors/Author`

OsmondJiang:
Based on the confirmation from PM, we will continue work on this feature as a high priority.

cc @herohua  @yufeih  @fenxu

OsmondJiang:
For #3173, it's related with `OriginalContentGitUrl/GitCommitUrl/OriginalContentGitUrlTemplate/ContentGitUrl/Contributor/Author`, So i will provide another design for that to make it more common.

@herohua  @yufeih

PR #4151, opened 2019-03-12
🌐☁add spec test for localization convention vsts support

OsmondJiang:
only add basic tests for localization convention of vsts repo,  to make sure:
1. folder/branch/repo mapping convention works
2. bilingual works


OsmondJiang:
#4032

PR #4152, opened 2019-03-12
📄Make JSON property deserialization case sensitive

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_workitems/edit/73018/

### DIFF
--- a/docs/specs/casing.yml
+++ b/docs/specs/casing.yml
@@ -55,7 +55,7 @@ outputs:
   output/a.json:
   output/b.json:
 ---
-# Resolve to the same lowercase image on windows and mac
+# Resolve to the same lower case image on windows and mac
 os: windows, osx
 inputs:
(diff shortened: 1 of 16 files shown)


superyyrrzz:
Why change `gitHub` to `github`? This will break integration with OPS

yufeih:
> Why change `gitHub` to `github`? This will break integration with OPS

`github` comes more natural in config then `gitHub`, imagine `github_user_cache` vs `git_hub_user_cache`. I DO need to update OPS powershell before this is in.

yufeih:
> > Why change `gitHub` to `github`? This will break integration with OPS
> 
> `github` comes more natural in config then `gitHub`, imagine `github_user_cache` vs `git_hub_user_cache`. I DO need to update OPS powershell before this is in.

Change that later

PR #4153, opened 2019-03-13
Split bookmark-not-found

mingwli:
split into:
internal-bookmark-not-found => suggestion
external-bookmark-not-found => warning


mingwli:
#3838

PR #4156, opened 2019-03-13
include-is-redirection should be error

mingwli:
#3838

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -270,8 +270,8 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Inclusion is a <see cref="Config.Redirections"/> entry like [!INCLUDE[](redirect.md)]
         /// </summary>
-        public static Error IncludeRedirection(Document relativeTo, string path)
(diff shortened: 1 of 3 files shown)


src/docfx/Errors.cs
```diff
@@ -270,8 +270,8 @@ public static Error PublishUrlConflict(string url, IEnumerable<Document> files,
         /// <summary>
         /// Inclusion is a <see cref="Config.Redirections"/> entry like [!INCLUDE[](redirect.md)]
         /// </summary>
-        public static Error IncludeRedirection(Document relativeTo, string path)
-            => new Error(ErrorLevel.Warning, "include-is-redirection", $"Referenced inclusion {path} relative to '{relativeTo}' shouldn't belong to redirections", relativeTo.ToString());
+        public static Error IncludeIsRedirection(Document relativeTo, string path)
```
yufeih: There are maybe other cases like including a PNG file or including a YAML file or a config file, to me these all belongs to inclusion is not an expected content type

OsmondJiang: we can focus on the level change right now

mingwli: I've added this to issue #3838 for next time discussion with our PMs.

Issue #4139, opened 2019-03-11
Avoid GitHub abuse detection


superyyrrzz:
After enable GitHub user cache expiry, sometimes we call GitHub failed:

> You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.

From [GitHub's guideline](https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits), currently we can try these:

* Make requests for a single user or client ID serially. Do not make requests for a single user or client ID concurrently.
* When you have been limited, use the Retry-After response header to slow down. The value of the `Retry-After` header will always be an integer, representing the number of seconds you should wait before making requests again. For example, `Retry-After: 30` means you should wait 30 seconds before sending more requests. (Related: #3749)

PR #4157, opened 2019-03-13
log how long GitHub expect us to wait for retry

superyyrrzz:
#4139

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -86,15 +86,8 @@ namespace Microsoft.Docs.Build
         ///   - the api call reach github limit
         ///   - using invalid access token(more detailed info in ex.Message)
         /// </summary>
-        public static Error GitHubApiFailed(string api, Exception ex, int? retryAfterSeconds = null)
-        {
-            var message = $"Failed calling GitHub API '{api}': {ex.Message}";
-            if (retryAfterSeconds.HasValue)
-            {
-                message += $" It is recommended to retry after {retryAfterSeconds} seconds.";
-            }
-            return new Error(ErrorLevel.Warning, "github-api-failed", message);
-        }
+        public static Error GitHubApiFailed(string api, Exception ex)
+            => new Error(ErrorLevel.Warning, "github-api-failed", $"Failed calling GitHub API '{api}': {ex.Message}");
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -60,6 +60,10 @@ public async Task<(Error, GitHubUser)> GetUserByLogin(string login)
                 _rateLimitError = Errors.GitHubApiFailed(apiDetail, ex);
                 return (_rateLimitError, null);
             }
+            catch (AbuseException ex)
+            {
+                return (Errors.GitHubApiFailed(apiDetail, ex, ex.RetryAfterSeconds), null);
```
OsmondJiang: can we combine these two api exception handler together? seems they are almost same

src/docfx/Errors.cs
```diff
@@ -86,8 +86,15 @@ public static Error GitHubUserNotFound(string login)
         ///   - the api call reach github limit
         ///   - using invalid access token(more detailed info in ex.Message)
         /// </summary>
-        public static Error GitHubApiFailed(string api, Exception ex)
-            => new Error(ErrorLevel.Warning, "github-api-failed", $"Failed calling GitHub API '{api}': {ex.Message}");
+        public static Error GitHubApiFailed(string api, Exception ex, int? retryAfterSeconds = null)
```
yufeih: Don't use error messages to log diagnostics, if you want to log just `Log.Write()`

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -102,6 +102,10 @@ public async Task<(Error, string login)> GetLoginByCommit(string repoOwner, stri
             }
             catch (Exception ex)
             {
+                if (ex is AbuseException aex)
+                {
+                    Log.Write($"Failed calling GitHub API '{apiDetail}', message: '{ex.Message}', retryAfterSeconds: '{aex.RetryAfterSeconds}'");
```
OsmondJiang: only applied to GetLoginByCommit? What about GetUserByLogin?

superyyrrzz: Forgot about this... added

PR #4158, opened 2019-03-13
Compile a list of error code without line info

yufeih:
#4130

### DIFF
--- a/test/docfx.Test/e2e/E2ETest.cs
+++ b/test/docfx.Test/e2e/E2ETest.cs
@@ -26,11 +26,13 @@ namespace Microsoft.Docs.Build
             "need-restore", "publish-url-conflict", "output-path-conflict", "download-failed", "heading-not-found", "config-not-found",
 
             // These error codes are the ones we could have line info but haven't implement them yet:
-            "bookmark-not-found", "file-not-found", "uid-not-found", "committish-not-found", "invalid-toc-syntax", "yaml-header-not-object",
+            "external-bookmark-not-found", "internal-bookmark-not-found", "file-not-found", "uid-not-found", "committish-not-found",
+            "invalid-toc-syntax", "yaml-header-not-object",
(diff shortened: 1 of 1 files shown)


test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -21,6 +21,18 @@ namespace Microsoft.Docs.Build
 {
     public static class E2ETest
     {
+        private static readonly string[] s_errorCodesWithoutLineInfo =
+        {
+            "need-restore", "publish-url-conflict", "output-path-conflict", "download-failed", "heading-not-found", "config-not-found",
+
+            // These error codes are the ones we could have line info but haven't implement them yet:
+            "bookmark-not-found", "file-not-found", "uid-not-found", "committish-not-found", "invalid-toc-syntax", "yaml-header-not-object",
```
mingwli:
```suggestion
            "internal-bookmark-not-found", ""external-bookmark-not-found"", "file-not-found", "uid-not-found", "committish-not-found", "invalid-toc-syntax", "yaml-header-not-object",
```

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -405,5 +420,16 @@ private static void VerifyFile(string file, string content)
...
             }
         }
+
+        private static void VerifyLogsHasLineInfo(string[] logs)
+        {
+            foreach (var log in Array.ConvertAll(logs, JArray.Parse))
+            {
+                if (!s_errorCodesWithoutLineInfo.Contains(log[1].ToString()) && log.Count < 6)
+                {
+                    Assert.True(false, $"Error code {log[1].ToString()} must have line info");
```
OsmondJiang: that's means by default newly added error should have line info, unless we add it to the exclusion list?

yufeih: yep!

Issue #4005, opened 2019-02-01
landingData throws a lot of `file not found` warnings


OsmondJiang:
https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C2%5C1%5C0c07d4eb-0562-6512-9227-12dbf89d9a0f%5CCommit%5C201902010600590723-markdig-migration-with-docfx-v3%5Crawlog.txt?sv=2015-02-21&sr=b&sig=ZnOOX4jsHR5bdk8qo8E6hws%2F15Ei7ldojYFDSIQuUDE%3D&st=2019-02-01T06%3A05%3A03Z&se=2019-03-04T06%3A10%3A03Z&sp=r

![image](https://user-images.githubusercontent.com/19990166/52108468-193c1180-2635-11e9-9435-c7e2ae925670.png)

live1206:
If we take a look at [speech-service](https://github.com/MicrosoftDocs/azure-docs-pr/tree/master/articles/cognitive-services/Speech-Service), should these files be found?

OsmondJiang:
There are still any warning on PROD azure-docs-pr, check report: https://opbuildstorageprod.blob.core.windows.net/report/2019%5C3%5C12%5Cdbb1031b-495e-6f6d-ad5f-4e1ef332e502%5CCommit%5C201903120309590274-markdig-migration-with-docfx-v3%5Cworkflow_report.html?sv=2015-02-21&sr=b&sig=XWk1a%2BngrW44fO6wAshhX5e5MmThK1EkTPuA20yK1TA%3D&st=2019-03-12T03%3A24%3A37Z&se=2019-04-12T03%3A29%3A37Z&sp=r

live1206:
There are some content issues, the href link within html should be absolute path, such as this one: https://github.com/MicrosoftDocs/azure-docs-pr/blob/95c1c2b248d131b60a967868fd8dbe84a408ee7d/articles/cosmos-db/index.yml#L149

But there are another error messages as below
```
Cannot find file 'en-us/azure/app-service/containers/tutorial-dotnetcore-sqldb-app' relative to 'articles/app-service/containers/index.yml'
```
Check [the original content](https://github.com/OPS-E2E-PPE/azure-docs-pr/blob/9b7cb3a417599510758d7d2fec7c88b3bd371ac1/articles/app-service/containers/index.yml#L95), seems like the href got locale added somehow. Created a test case which works as expected. Any idea where might this locale come from?
@OsmondJiang  @yufeih

PR #4161, opened 2019-03-14
👀Remove [HTML] since there is no need to resolve href within HTML

live1206:
#4005

### DIFF
--- a/src/Microsoft.Docs.Template/Models/LandingData.cs
+++ b/src/Microsoft.Docs.Template/Models/LandingData.cs
@@ -109,7 +109,6 @@ namespace Microsoft.Docs.Build
 
         public LandingDataRow[] Rows { get; set; }
 
-        [Html]
         public string Html { get; set; }
     }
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Docs.Template/Models/LandingData.cs
```diff
@@ -109,7 +109,6 @@ public sealed class LandingDataItem
 
         public LandingDataRow[] Rows { get; set; }
 
-        [Html]
```
OsmondJiang: For html attribute, it also do something else, can we just remove the dependency resolving from these logics:

```
            if (attribute is HtmlAttribute)
            {
                var html = HtmlUtility.TransformLinks((string)value, href =>
                {
                    var (error, link, _) = context.DependencyResolver.ResolveLink(href, file, file, buildChild);
                    context.Report.Write(file.ToString(), error);
                    return link;
                });
                return HtmlUtility.StripTags(HtmlUtility.LoadHtml(html)).OuterHtml;
            }
```

PR #4166, opened 2019-03-15
✍contributor, author, updated_at for vsts repo

OsmondJiang:


OsmondJiang:
#4032

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -54,7 +54,10 @@ public static async Task<ContributionProvider> Create(Docset docset, GitHubUserC
             var userIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             var updatedDateTime = GetUpdatedAt(document, commits);
 
-            var resolveGitHubUsers = GitHubUtility.TryParse(repo?.Remote, out var gitHubOwner, out var gitHubRepoName) && document.Docset.Config.GitHub.ResolveUsers;
+            _ = GitHubUtility.TryParse(repo?.Remote, out var gitHubOwner, out var gitHubRepoName) ||
+                GitHubUtility.TryParse(document.Docset.Config.Contribution.Repository, out gitHubOwner, out gitHubRepoName);
+
+            var resolveGitHubUsers = document.Docset.Config.GitHub.ResolveUsers;
```
yufeih: we only resolve github users if both (resolve users is true) and (the repository or contribution repository) is github like, this change omits the second condition and will cause unneeded github api calls

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -142,15 +142,15 @@ public async Task<(Error error, GitHubUser user)> GetByLogin(string login)
...
         {
             if (string.IsNullOrEmpty(authorEmail))
                 return default;
 
             Telemetry.TrackCacheTotalCount(TelemetryName.GitHubUserCache);
             var user = TryGetByEmail(authorEmail);
-            if (user != null)
-                return (null, user.IsValid() ? user : null);
+            if (user != null || string.IsNullOrEmpty(repoOwner) || string.IsNullOrEmpty(repoName))
+                return (null, (user?.IsValid() ?? false) ? user : null);
```
yufeih: consider adding a new method `GetByEmail` that only calls `TryGetByEmail` and leave `GetByCommit` intact.

src/docfx/build/metadata/ContributionProvider.cs
OsmondJiang: make senses

src/docfx/lib/github/GitHubUserCache.cs
OsmondJiang: this may change the first condition trigger, need confirmed with @superyyrrzz.

yufeih:
conflict due to my force push (my bad), would you create a new PR or update existing?

docs/specs/git.yml
```diff
@@ -407,3 +407,56 @@ outputs:
...
+      ]
+    }
+outputs:
+  docs/index.json: |
+    {
+      "author": 
+        { "name": "charlie", "profile_url": "https://github.com/charlie", "display_name": "Charlie", "id": "3" },
+      "contributors": [
+        { "name": "bob", "profile_url": "https://github.com/bob", "display_name": "Bob", "id": "2" },
+        { "name": "OsmondJiang", "profile_url": "https://github.com/OsmondJiang", "display_name": "osmondjiang", "id": "19990166" }
```
superyyrrzz: As GitHub limits API calls, this test may fail unexpectedly.

OsmondJiang: when we met this error, we can add github token for this test case

PR #4167, opened 2019-03-15
Update copyright text from 2018 to 2019.

scovetta:


yufeih:
@superyyrrzz @herohua , should we remove the years from copyright headers? Copyright headers in our  source code looks like `Copyright (c) Microsoft. All rights reserved.`

yufeih:
@scovetta , Thank you for taking the time to come up with this PR, could you tweak it a little bit to exclude the years all together from the copyright headers, so we don't need to deal with it each year?

herohua:
@superyyrrzz this PR was merged into `master` branch. do we need to bring it back to `dev`?

PR #4168, opened 2019-03-15
🔗Do not escape all URLs

yufeih:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/73036/


superyyrrzz:
Will you add a guard for filename including `#`?

yufeih:
> Will you add a guard for filename including `#`?

Not in this PR. That part should be covered by validation library in v2, as part of porting validation library to v3, we can get that feature in v3.

PR #4170, opened 2019-03-15
Add back resolving href for landing page, mute file-not-found for landing page

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/74434

### DIFF
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -348,20 +348,3 @@ outputs:
     { "content": "[!include[B1](~/b.md)]<p>Token</p>\n" }
   build.log: |
     ["warning","include-not-found","Cannot resolve '~/b.md' relative to 'docs/a.md'.","docs/a.md","",1,1]
----
-# Resolve link for landing page
-inputs:
-  docfx.yml:
-  docs/index.yml: |
-    ### YamlMime:YamlDocument
-    documentType: LandingData
-    title: REST API Documentation
-    abstract:
-      menu:
-        items:
-        - href: ~/b
-  b.md: b
-outputs:
-  docs/index.json: |
-    { "content": {"title":"REST API Documentation","abstract":{"menu":{"items":[{"href":"../b"}]}},"documentType":"LandingData"} }
-  b.json:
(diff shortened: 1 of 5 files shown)


src/docfx/build/page/AttributeTransformer.cs
```diff
@@ -45,7 +45,8 @@ private static object TransformContent(Context context, DataTypeAttribute attrib
         {
             if (attribute is HrefAttribute)
             {
-                var (error, link, _) = context.DependencyResolver.ResolveLink((string)value, file, file, buildChild);
+                var (error, link, _) = context.DependencyResolver.ResolveLink((string)value, file, file, buildChild, file.FilePath.EndsWith("index.yml"));
```
yufeih: file.Schema.Name == "LandingData"

live1206: right

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -224,7 +224,14 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
 
                     var file = Document.TryCreateFromFile(relativeTo.Docset, pathToDocset);
 
-                    return (file != null ? null : Errors.FileNotFound(relativeTo.ToString(), path), file, null, query, fragment, null, pathToDocset);
+                    // try to resolve with .md for landing page
+                    if (file is null && forLandingPage)
```
OsmondJiang: just a suggestion, move this out of resolveLink method, hack it in caller.

live1206: Added a comment, this flag should not be used. And the hack needs to be removed once the user input is correct.

src/docfx/build/page/AttributeTransformer.cs
live1206: Actually, we could not use `file.Schema.Name == "LandingData"` here, because `file.Schame` could be null and we are getting `documentType` instead as the test shows.

docs/specs/resolve.yml
```diff
@@ -348,3 +348,20 @@ outputs:
     { "content": "[!include[B1](~/b.md)]<p>Token</p>\n" }
   build.log: |
     ["warning","include-not-found","Cannot resolve '~/b.md' relative to 'docs/a.md'.","docs/a.md","",1,1]
+---
```
yufeih: nit: move this spec to `legacy.yml` to make it easier to find hacks like this?

live1206: sure, we should get rid of it asap

Issue #4174, opened 2019-03-15
azure-docs-pr build performance downgrade


OsmondJiang:
~300s -> ~450s

https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C3%5C15%5C0c07d4eb-0562-6512-9227-12dbf89d9a0f%5CCommit%5C201903150600420894-markdig-migration-with-docfx-v3%5Cworkflow_report.html?sv=2015-02-21&sr=b&sig=IdH97D0pOLQPKUshi9tRBYTcCLUZqimTJmgpytAEq6k%3D&st=2019-03-15T08%3A53%3A57Z&se=2019-04-15T08%3A58%3A57Z&sp=r

OsmondJiang:
@yufeih  @superyyrrzz  @herohua

PR #4175, opened 2019-03-15
Log github access calls

yufeih:
#4174 #4139


superyyrrzz:
Move into `Retry`'s action? then we can know if we call GitHub multiple times for the same API.

PR #4185, opened 2019-03-18
perf: don't apply global/file metadata for resource

superyyrrzz:


src/Microsoft.DocAsCode.Build.Engine/HostServiceCreator.cs
```diff
@@ -72,10 +72,12 @@ public virtual bool ShouldProcessorTraceInfo(IDocumentProcessor processor)
             {
                 Logger.LogDiagnostic($"Processor {processor.Name}, File {file.FullPath}: Loading...");
 
-                metadata = ApplyFileMetadata(file.FullPath, metadata, fileMetadata);
+                var fileMeta = NeedApplyMetadata()
```
herohua: `fileMeta` -> `fileMetadata`?

superyyrrzz: already used in method's parameter list...

PR #4193, opened 2019-03-18
Do not log github-user-not-found error in GetByCommit

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/9745?_a=files

### DIFF
--- a/test/docfx.Test/lib/GitHubUserCacheTest.cs
+++ b/test/docfx.Test/lib/GitHubUserCacheTest.cs
@@ -127,6 +127,28 @@ namespace Microsoft.Docs.Build
             };
             yield return new object[]
             {
+                "Get user by commit does not return user not found error",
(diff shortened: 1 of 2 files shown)


test/docfx.Test/lib/GitHubUserCacheTest.cs
```diff
@@ -126,6 +126,28 @@ private static IEnumerable<object[]> PrepareDataCore()
                 1
             };
             yield return new object[]
+            {
+                "Get user by commit does not return user not found error",
+                (Func<GitHubUserCache, Task>) ( async (cache) =>
+                {
+                    Assert.Null((await cache.GetByCommit("me@contoso.com", "owner", "name", "3")).error);
```
live1206: seems a bit odd to assert here, possible to return it and assert in `TestCache`?

PR #4194, opened 2019-03-19
🛠Config validation

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/73054
Add validation for config file metadata and global metadata, to avoid duplicated errors on page building.

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -83,7 +83,7 @@ inputs:
   docs/b.md:
 outputs:
   build.log: |
-    ["error","violate-schema","Error reading string. Unexpected token: StartArray. Path 'title'.","docfx.yml"]
+    ["error","violate-schema","Expected type string, please input string or type compatible with string.","docfx.yml","title"]
 ---
 # File metadata of object form
 inputs:
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -72,6 +72,19 @@ outputs:
   docs/b.json: |
     { "metadata": { "key1": "value", "key2": 2 } }
 ---
+# Global metadata schema violation
+inputs:
+  docfx.yml: |
+    globalMetadata:
+      title: 
+        - 1
+        - 2
+  docs/a.md:
+  docs/b.md:
+outputs:
+  build.log: |
+    ["error","violate-schema","Error reading string. Unexpected token: StartArray. Path 'title'.","docfx.yml"]
+---
 # File metadata of object form
 inputs:
   docfx.yml: |
@@ -413,3 +426,15 @@ inputs:
       a.md: https://docs.microsoft.com/docfx/a
       b.md: a
 outputs:
+---
+# file metadata schema violation
+inputs:
+  docfx.yml: |  
+    fileMetadata:
+      title:
+        docs/**: 
+          - 1
+          - 2
+outputs:
+  build.log: |
+    ["error","violate-schema","Expected 'String' for 'title', but got 'Array'","docfx.yml","file metadata"]
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -83,7 +83,7 @@ inputs:
   docs/b.md:
 outputs:
   build.log: |
-    ["error","violate-schema","Expected type string, please input string or type compatible with string.","docfx.yml","title"]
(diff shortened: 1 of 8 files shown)


docs/specs/config.yml
```diff
@@ -72,6 +72,19 @@ outputs:
...
+  docfx.yml: |
+    globalMetadata:
+      title: 
+        - 1
+        - 2
+  docs/a.md:
+  docs/b.md:
+outputs:
+  build.log: |
+    ["error","violate-schema","Error reading string. Unexpected token: StartArray. Path 'title'.","docfx.yml"]
```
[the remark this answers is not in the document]
live1206: For file metadata, since it has a nesting structure of `{metadata-name} -> {glob} -> {metadata-value}`, we can not just use Json.Net for validation, so came up with some manual validation.
For global metadata and other schema validations, the error message has been rewritten.

src/docfx/build/metadata/MetadataValidator.cs
```diff
@@ -2,34 +2,73 @@
...
                     errors.Add(Errors.ReservedMetadata(new Range(lineInfo?.LineNumber ?? 0, lineInfo?.LinePosition ?? 0), key, from));
                 }
+                else
+                {
+                    var type = s_fileMetadataTypes.GetOrAdd(
+                       key,
+                       new Lazy<Type>(() => typeof(FileMetadata).GetProperty(key, BindingFlags.IgnoreCase | BindingFlags.Public | BindingFlags.Instance)?.PropertyType));
+                    if (type.Value is null)
+                        continue;
+                    var values = token as IEnumerable<KeyValuePair<string, JToken>>;
```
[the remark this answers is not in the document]
live1206: For global metadata, we can just use Json.Net, so no need to do this manual validation. And the error message has been rewritten, we can tweak on it further.

docs/specs/config.yml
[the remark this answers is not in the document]
live1206: Per offline discussion, line info would be enough to guide the user.

docs/specs/config.yml
```diff
@@ -413,3 +426,15 @@ inputs:
...
+inputs:
+  docfx.yml: |  
+    fileMetadata:
+      title:
+        docs/**: 
+          - 1
+          - 2
+outputs:
+  build.log: |
+    ["error","violate-schema","Expected type String, please input string or type compatible with String.","docfx.yml","fileMetadata.title['docs/**']",4,7]
```
OsmondJiang: 👍

docs/specs/metadata.yml
```diff
@@ -341,7 +341,7 @@ inputs:
     ---
 outputs:
   build.log: |
-    ["warning","reserved-metadata","Metadata 'redirect_url' is reserved by docfx, remove this metadata from yaml header","docs/a.md","",1,15]
+    ["warning","reserved-metadata","Metadata 'redirect_url' is reserved by docfx, remove this metadata from redirect_url","docs/a.md","",1,15]
```
OsmondJiang: a little strange here?

src/docfx/lib/JsonUtility.cs
```diff
@@ -266,7 +268,7 @@ public static (List<Error>, JToken) Deserialize(string json)
         public static void Merge(JObject container, JObject overwrite)
         {
             container.Merge(overwrite, s_mergeSettings);
-        }
+                    }
```
OsmondJiang: space

src/docfx/lib/JsonUtility.cs
```diff
@@ -345,9 +347,24 @@ private static (Range, string message, string path) ParseException(Exception ex)
...
+                return (default, RewriteErrorMessage(match.Groups[1].Value), match.Groups[2].Value);
+            }
+            return (default, RewriteErrorMessage(ex.Message), null);
+        }
+
+        private static string RewriteErrorMessage(string message)
+        {
+            if (message.StartsWith("Error reading string. Unexpected token"))
+            {
+                return "Expected type String, please input String or type compatible with String.";
```
OsmondJiang: only works for expecting string? what about other expecting case or add a other type mismatch test case?

live1206: https://github.com/dotnet/docfx/blob/8569891a307da86ae7704810d59b475a6fcff4cc/test/docfx.Test/lib/JsonUtilityTest.cs#L331
Covered here

Issue #4078, opened 2019-02-26
Use absolute URL for intermediate representation


yufeih:
We are currently using relative URL for intermediate representation, this complicates URL handling logic and makes html caching difficult for markdown inclusion and SDP. We could instead use absolute URL for intermediate representation and convert absolute URLs to relative URLs in the final HTML post processing stage. We are going to have an HTML post processing stage regardless so there is unlikely performance penalty.

> SCOPE: This issue does not track how we cache inclusion or SDP, it just defines the intermediate HTML representation that is cache friendly.

### Design

Intermediate HTML contains `data-linktype` and absolute URL (starting with `/`) for links that points to files:

Type | Markdown | HTML
---|---|---
External | `[](https://docs.com)` | `<a data-linktype="external" href="https://docs.com"></a>`
AbsolutePath | `[](/azure)` | `<a data-linktype="absolute-path" href="/azure"></a>`
RelativePath | `[](a.md)` | `<a data-linktype="relative-path" href="/basepath/a"></a>`
RelativePath with error | `[](a.md)` | `<a data-linktype="relative-path" href="a.md"></a>`
Bookmark | `[](#anchor)` | `<a data-linktype="self-bookmark" href="#anchor"></a>`

In HTML post process, all links are turned into relative URL if it satisfy both:
1. `data-linktype` is `relative-path`
2. `href` starts with `/`

> The above design also applies to images (`<img src="xxx">`)

### Work Items

- [ ] Move `data-linktype` generation from HTML post process to markup
- [ ] Generate absolute URLs in markup
- [ ] Convert absolute URL to relative URL in HTML post process

PR #4197, opened 2019-03-19
Move resolve link outside markdown extension core

yufeih:
To do #4078 , we need to handle link post process differently in v3, this PR moves link resolve logic outside markdig extensions to markdig service (v2 only), so v3 can have it's own link resolve logic.


herohua:
@daxianji007 @yishengjin1413 @qinezh

src/Microsoft.DocAsCode.MarkdigEngine/ResolveLink/ResolveLinkExtension.cs
```diff
@@ -0,0 +1,67 @@
...
+                }
+
+                if (markdownObject is LinkInline linkInline && !linkInline.IsAutoLink)
+                {
+                    linkInline.GetDynamicUrl = () => GetLink(linkInline.Url, InclusionContext.File, InclusionContext.RootFile, linkInline);
+                }
+            }
+        }
+
+        private static string GetLink(string path, object relativeTo, object resultRelativeTo, MarkdownObject origin)
```
yishengjin1413: The same with https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine/MarkdigMarkdownService.cs#L165? Can we put them in one place?

yishengjin1413:
Looks good to me~

PR #4199, opened 2019-03-19
🐛add auth token and log for spec test 

OsmondJiang:


.deploy.yml
```diff
@@ -28,7 +28,12 @@ jobs:
   condition: eq(variables['Build.Reason'], 'PullRequest')
   steps:
     - powershell: ./build.ps1
-
+    env:
+      OP_BUILD_USER_TOKEN: $(OpBuildUserToken)
```
yufeih: we don't need `OP_BUILD_USER_TOKEN` for build, it is to support running config migration script in impact testing

docs/specs/git.yml
```diff
@@ -434,6 +434,7 @@ inputs:
     gitHub:
       userCache: user_profile.json
       resolveUsers: true
+      auth: {GITHUB_ACCESS_TOKEN}
```
superyyrrzz: authToken

.deploy.yml
```diff
@@ -56,14 +56,18 @@ jobs:
   - powershell: ./impact.ps1
     condition: eq(variables['Build.SourceBranch'], 'refs/heads/v3')
     env:
-      OP_BUILD_USER_TOKEN: $(OpBuildUserToken)
```
mingwli: this one is used for running config migration in edge-developer

OsmondJiang: @yufeih  is saying that you don't need it anymore?

PR #4200, opened 2019-03-19
🐞 Make legacy output deterministic in case of conflict

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/9784?_a=files

### DIFF
--- a/src/docfx/build/context/Output.cs
+++ b/src/docfx/build/context/Output.cs
@@ -85,7 +85,7 @@ namespace Microsoft.Docs.Build
             File.Copy(sourcePath, GetDestinationPath(destRelativePath), overwrite: true);
         }
 
-        public void Delete(string destRelativePath)
+        public void Delete(string destRelativePath, bool legacy = false)
         {
             Debug.Assert(!Path.IsPathRooted(destRelativePath));
 
@@ -94,6 +94,15 @@ namespace Microsoft.Docs.Build
             if (File.Exists(destinationPath))
             {
                 File.Delete(destinationPath);
+            }
+
+            if (legacy)
+            {
+                var mtaJsonPath = Path.ChangeExtension(destinationPath.Replace(".raw.page.json", ".json"), "mta.json");
+                if (File.Exists(mtaJsonPath))
+                {
+                    File.Delete(mtaJsonPath);
+                }
             }
         }
 
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -225,7 +225,7 @@ namespace Microsoft.Docs.Build
 
             if (file.Docset.Legacy)
             {
-                if (isPage)
+                if (isPage && file.Docset.Template != null)
                 {
                     return TemplateTransform.Transform(model, file);
                 }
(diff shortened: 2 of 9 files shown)


src/docfx/build/context/Output.cs
```diff
@@ -95,6 +95,15 @@ public void Delete(string destRelativePath)
             {
                 File.Delete(destinationPath);
             }
+
+            if (legacy)
+            {
+                var mtaJsonPath = Path.ChangeExtension(destinationPath.Replace(".raw.page.json", ".json"), "mta.json");
```
OsmondJiang: why not directly repleace the extension?

yufeih: There could be other formats like `.png` for resource and `.json` for TOC.

src/docfx/build/context/Output.cs
```diff
@@ -95,6 +95,15 @@ public void Delete(string destRelativePath)
             {
                 File.Delete(destinationPath);
             }
+
+            if (legacy)
```
OsmondJiang: it's only works for legacy?  what about ideal output?

yufeih: For ideal, we don't have `.mta.json` so there is no problem. There is also an existing tests that covers conflicts for ideal output

src/docfx/build/page/BuildPage.cs
```diff
@@ -225,7 +225,7 @@ private static (object output, JObject extensionData) ApplyTemplate(Document fil
 
             if (file.Docset.Legacy)
             {
-                if (isPage)
+                if (isPage && file.Docset.Template != null)
```
OsmondJiang: the condition is not really matched with below using item, not really straightforward.

src/docfx/build/context/Output.cs
OsmondJiang: it's the Delete is deleting multiple files instead of one file, why not passing in the publish item, using it to decide which files need to be delete from file system, in case we have more in the future

src/docfx/build/page/BuildPage.cs
yufeih: If put inside, the method signature does not match. When `Template == null`, it is currently best to return a `PageModel`, but `Template.Transform` currently returns a `TemplateModel`, so I'd rather keep the current change.

OsmondJiang: `Docset.Template` is used inside of `TemplateTransform.Transform`, so it's better checked inside of `TemplateTransform.Transform`. For the signature, we can change the it right?

OsmondJiang: just an example;
```C#
            if (file.Docset.Legacy)
            {
                if (isPage && TemplateTransform.TryTransform(model, file, out var templateModel, out var metadata))
                {
                    return (templateModel, metadata);
                }

                return (model, null);
            }
```
```C#
        public static bool TryTransform(PageModel pageModel, Document file, out TemplateModel model, out JObject metadata)
        {
            model = default;
            metadata = default;
            if (file.Docset.Template == null)
            {
                return false;
            }

            (model, metadata) = Transform(pageModel, file);
            return true;
        }
```

OsmondJiang: offline discussed, add TODO here

PR #4203, opened 2019-03-20
perf: use StringComparison.Ordinal(IgnoreCase)

superyyrrzz:
[Use comparisons with `StringComparison.Ordinal` or `StringComparison.OrdinalIgnoreCase` for better performance](https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings?view=netframework-4.7.2#recommendations-for-string-usage)


herohua:
Cool 👍

herohua:
@jsntcy, @TianqiZhang shall we consider adding this into best practice?

yufeih:
@superyyrrzz , how much performance gain to we get from this and shall we do the same for v3?

superyyrrzz:
@yufeih it saves about 30%~40% time when executing the method.

In azure-docs-pr, here is the data from profiling a [method](https://github.com/dotnet/docfx/blob/da48dd4c820f4b15dbcf4cc350ab68fabc5292b4/src/docfx/SubCommands/DocumentBuilderWrapper.cs#L218) mainly to evaluate glob patters:

| | method total time | `StartsWith` CPU | `EndsWith` CPU |
|---|---|---|---|
| before | ~12s | 186 (1.44%) | 302 (2.34%) |
| after | ~10s | 106 (1.05%) | 162 (1.60%) |

cc @herohua

jsntcy:
@herohua, I think we can add it to best practices, but will think how to organize it (may be just a tip)

PR #4208, opened 2019-03-20
🌐Bulk get GitHub commits

yufeih:
This PR fixes GitHub API abuse/rate limit and performance problem by:

- Using bulk API calls to get commits and cache users more aggressively (bulk API consumes the same rate as non-bulk API)
- Make GitHubUserCache serial, to avoid API abuse
- Make default cache expiration in 15 ~ 30 days

Test Result:

Repo | User profile cache | # of GitHub API calls | Build Time | Abused / rate limited?
------|--------------------------|------------------------|------------|-------
azure-docs-pr | no | 2091 |  00:26:50 | no
azure-docs-pr | yes (today's sandbox cache, may have expired records) | 210 | 00:03:36 | no
azure-docs-pr | yes (impact test cache, full) | 0 | 00:01:43 | no
edge-developer | no | 13 | 10s | no
edge-developer | yes (today's sandbox cache, may have expired records) | 1 | 3s | no

#4139

### DIFF
--- a/src/docfx/lib/github/GitHubAccessor.cs
+++ b/src/docfx/lib/github/GitHubAccessor.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Octokit;
 
--- a/src/docfx/lib/github/GitHubAccessor.cs
+++ b/src/docfx/lib/github/GitHubAccessor.cs
@@ -2,7 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Octokit;
@@ -12,10 +14,6 @@ namespace Microsoft.Docs.Build
     internal class GitHubAccessor
     {
         private readonly GitHubClient _client;
-
-        // Bypass GitHub abuse detection:
-        // https://developer.github.com/v3/guides/best-practices-for-integrators/#dealing-with-abuse-rate-limits
-        private static readonly SemaphoreSlim _semaphore = new SemaphoreSlim(10, 10);
 
         private static volatile Error _rateLimitError;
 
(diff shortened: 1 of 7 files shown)


src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -81,17 +78,15 @@ public async Task<(Error, GitHubUser)> GetUserByCommit(string repoOwner, string
             var apiDetail = $"GET /repos/{repoOwner}/{repoName}/commits/{commitSha}";
```
OsmondJiang: is `apiDetail` need to be changed to new API?

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -154,55 +149,59 @@ async Task<(Error, GitHubUser)> GetByCommitCore()
...
                 // only mark the email as invalid when the user is not found
-                if (user != null)
+                if (users != null)
                 {
-                    UpdateUser(user);
+                    UpdateUsers(users);
                 }
-                return (error, user);
+
+                return (error, _usersByEmail.TryGetValue(authorEmail, out var user) && user.IsValid() ? user : null);
```
OsmondJiang: will that cause any differences  between `get by commit` and `get by email`?

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -138,13 +133,13 @@ public Task<(Error error, GitHubUser user)> GetByCommit(string authorEmail, stri
             if (string.IsNullOrEmpty(authorEmail))
```
OsmondJiang: does that mean if the author email is empty, we can't get any user from github even using commit?

yufeih: yep

OsmondJiang: we should be able to get user by login(got from commit) even if the author email is empty?

src/docfx/lib/github/GitHubUserCache.cs
yufeih: The `string.IsNullOrEmpty(repoOwner)` will make `get by email` return early, so there is no different. But it is bad to mix these into one method, I need to split them in another PR later.

PR #4215, opened 2019-03-21
🧚‍♀️Support inclusion in titles

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_workitems/edit/75662/


docs/specs/manifest.yml
```diff
@@ -55,7 +55,7 @@ outputs:
     {
       "files": [
         { "url": "/c", "!hash": null },
-        { "url": "/a", "hash": "cc98fe5ff90ff67265c7d6f3f7ed70298d2f19a7" },
+        { "url": "/a", "hash": "d9f0ca10e7db6c5a0a5e0c3279d38576aea69869" },
```
yufeih: this is due to `raw_title=""` --> `!raw_title`. docs-ui does not depend on raw_title having a value, so the change is safe.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -159,6 +163,32 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+                {
+                    continue;
+                }
+
+                if (child.NodeType == HtmlNodeType.Text || string.IsNullOrWhiteSpace(child.OuterHtml))
+                {
+                    continue;
+                }
+
+                if (child.NodeType == HtmlNodeType.Element && child.Name == "h1")
```
mingwli: >  if (child.NodeType == HtmlNodeType.Element && child.Name == "h1") [](start = 14, length = 67)

we follow the v2 behavior here?

yufeih: yep, let me do it in another PR, want to keep one PR for one purpose

Issue #4206, opened 2019-03-20
Get latest version for dependency without lock


OsmondJiang:
Current implementation is trying to find latest `last access commit` version, not real latest version.

PR #4216, opened 2019-03-21
🐛get true latest version of dependency git without lock

OsmondJiang:
#4206

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -124,9 +124,9 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Get a list of commits using git log
         /// </summary>
-        public static Task<string[]> GetCommits(string path, int top = 1)
+        public static Task<string[]> GetCommits(string path, string committish, int top = 1)
         {
-            return Execute(path, $"--no-pager log --pretty=format:\"%H\" -{top}", (stdout, stderr) =>
+            return Execute(path, $"--no-pager log {committish} --pretty=format:\"%H\" -{top}", (stdout, stderr) =>
             {
                 Debug.Assert(stdout != null);
 
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -122,6 +122,19 @@ namespace Microsoft.Docs.Build
         }
 
         /// <summary>
+        /// Get a list of commits using git log
+        /// </summary>
+        public static Task<string[]> GetCommits(string path, int top = 1)
+        {
+            return Execute(path, $"--no-pager log --pretty=format:\"%H\" -{top}", (stdout, stderr) =>
+            {
+                Debug.Assert(stdout != null);
+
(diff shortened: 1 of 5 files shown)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -121,6 +121,19 @@ public static Task Fetch(string path, string url, string committish, Config conf
             return CloneOrUpdate(path, url, new[] { committish }, bare: false, depthOne: false, prune: false, config);
         }
 
+        /// <summary>
+        /// Get a list of commits using git log
+        /// </summary>
+        public static Task<string[]> GetCommits(string path, int top = 1)
```
yufeih: this may take too long (consider azure-docs-pr with xxxxxxxxx of commits). consider use commit time for comparison. time could be stored in `DependencyGit` so you only get the time once on checkout.

OsmondJiang: i only take top 100 for comparing

yufeih: > i only take top 100 for comparing

then it may give you wrong result

OsmondJiang: commit time is not pushing time, not really correct.

yufeih: tree-depth?

OsmondJiang: > > i only take top 100 for comparing
> 
> then it may give you wrong result

I believe it's enough, i then sort by last access time. considering our build frequency.

OsmondJiang: considering perf, even i changed it to top 1000, it finished very fast, and it only happens on restoring per repo, not building per file.

src/docfx/restore/RestoreMap.cs
```diff
@@ -193,17 +194,28 @@ string GetFileContentIfExists(string file)
...
-            });
+                    if (!string.IsNullOrEmpty(commit))
+                    {
+                        // found commit matched slot
+                        filteredGits = filteredGits.Where(i => i.Commit == commit);
+                    }
+
+                    var commits = Array.Empty<string>();
+                    if (filteredGits.Count() > 1)
+                    {
```
mingwli: when would there be multiple slots with same commit and branch?

src/docfx/restore/RestoreMap.cs
```diff
@@ -193,17 +194,28 @@ string GetFileContentIfExists(string file)
...
+                async gits =>
                 {
-                    // found commit matched slot
-                    filteredGits = filteredGits.Where(i => i.Commit == commit);
-                }
+                    var filteredGits = gits.Where(i => i.Branch == branch);
 
-                return filteredGits.ToList();
-            });
+                    if (!string.IsNullOrEmpty(commit))
```
mingwli: suggestion:
 check commit first to return earlier?

OsmondJiang: here we are assuming that we can get multiple slots with same commit.

PR #4220, opened 2019-03-21
Log whether repo cache or git commit cache is hit

yufeih:


src/docfx/lib/git/GitUtility.cs
```diff
@@ -281,6 +281,7 @@ private static async Task CloneOrUpdate(string path, string url, IEnumerable<str
             Telemetry.TrackCacheTotalCount(TelemetryName.GitRepositoryCache);
             if (git_remote_create(out var remote, repo, "origin", url) == 0)
             {
+                Log.Write($"Using existing repository '{path}' for '{url}'");
```
OsmondJiang: Seems we always trace a `log` and `telemetry cache missing` at the place where the cache is missing(no matter what kinds of cache), do we consider to combine them together?

yufeih: Good point. I thought about it but decided against it. We don't want too many console logs (slows us down and makes it harder to diagnose), currently we log cache hit info, for github user cache, this will generate enormous amount of logs.

PR #4221, opened 2019-03-21
🧹Remove MarkupResult and ConceptualMarkdown

yufeih:
Clean things up after #4215

### DIFF
--- a/src/docfx/lib/markdown/MarkdownPipelineType.cs
+++ b/src/docfx/lib/markdown/MarkdownPipelineType.cs
@@ -1,13 +1,12 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Collections.Generic;
-
 namespace Microsoft.Docs.Build
 {
-    internal class MarkupResult
+    public enum MarkdownPipelineType
     {
-        // TODO: remove this class
-        public List<Error> Errors = new List<Error>();
+        Markdown,
+        InlineMarkdown,
+        TocMarkdown,
     }
 }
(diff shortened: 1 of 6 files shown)


src/docfx/lib/markdown/MarkdownPipelineType.cs
```diff
@@ -1,13 +1,12 @@
...
-using System.Collections.Generic;
-
 namespace Microsoft.Docs.Build
 {
-    internal class MarkupResult
+    public enum MarkdownPipelineType
     {
-        // TODO: remove this class
-        public List<Error> Errors = new List<Error>();
+        Markdown,
```
OsmondJiang: here the markdown is conceptual markdown right? why rename it, we can have more markdown like overwrite markdown?

src/docfx/lib/markdown/MarkdownUtility.cs
```diff
@@ -16,47 +12,36 @@
 
 namespace Microsoft.Docs.Build
 {
-    public enum MarkdownPipelineType
-    {
-        ConceptualMarkdown,
```
OsmondJiang: seems previous `markdown/inlinemarkdown` attribute doesn't resolve moniker zone,  will this brings perf impact or result diff?

yufeih: That seems to me like a bug @928PJY markdown/inline markdown is also used in SDP, we should also handle moniker zones in SDP right?

928PJY: we don't have the plan to support moniker zone syntax in SDP markdown before, so it hasn't been added to the inlineMarkdown pipeline, but seems it will just bring perf impact to the repos which contain moniker zone syntax in SDP files, to keep the consistency of markdown pipelines and since it matches our future goal, we can add it to inlineMarkdown pipeline now.

PR #4223, opened 2019-03-21
remove ending '/' in git remote url

mingwli:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/75963/


src/docfx/lib/HrefUtility.cs
```diff
@@ -86,6 +86,8 @@ public static (string remote, string refspec, bool hasRefSpec) SplitGitHref(stri
 
             var (path, _, fragment) = SplitHref(remoteHref);
 
+            if (!string.IsNullOrEmpty(path) && path.EndsWith("/"))
```
yufeih:
```suggestion
            path = path.TrimEnd('/', '\\');
```

PR #4224, opened 2019-03-21
Fix h1 extraction skipping non empty text

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/9867?_a=overview

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -89,7 +89,8 @@ inputs:
     [!include[]()]
     # Heading
 outputs:
-  docs/a.json:
+  docs/a.json: |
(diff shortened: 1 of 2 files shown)


docs/specs/markdown.yml
```diff
@@ -80,6 +80,20 @@ outputs:
...
+# Include block before h1
+inputs:
+  docfx.yml: |
+    rules:
+      heading-not-found: warning
+  docs/a.md: |
+    [!include[]()]
+    # Heading
+outputs:
+  docs/a.json:
```
live1206: verify h1 remained in the output?

yufeih: because there is a visible block before h1. We only extract the h1 as title if it is the first visible block

PR #4230, opened 2019-03-22
resolve title from h1 - h3

mingwli:
#3838

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -46,6 +46,44 @@ outputs:
       "metadata": { "title": "title from yaml header" }
     }
 ---
+# resolve raw_title from h1 to h3
+inputs:
+  docfx.yml: |
+    rules:
+      heading-not-found: warning
+  h1.md: |
+    # h1 title
+  h2.md: |
+    ## h2 title
+  h3.md: |
+    ### h3 title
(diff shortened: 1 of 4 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -177,7 +179,7 @@ public static HtmlNode ExtractTitle(HtmlNode node)
                     continue;
                 }
 
-                if (child.NodeType == HtmlNodeType.Element && child.Name == "h1")
+                if (child.NodeType == HtmlNodeType.Element && s_titleElementNames.Contains(child.Name))
```
yufeih: nit: `(name == "h1" || name == "h2" || name == "h3")` is usually faster then `Contains` for simple small checks

docs/specs/markdown.yml
```diff
@@ -60,7 +98,7 @@ inputs:
 outputs:
   docs/a.json:
   build.log: |
-    ["warning","heading-not-found","The first visible block is not a heading block with `#`","docs/a.md"]
+    ["warning","heading-not-found","The first visible block is not a heading block with `#`, `##` or `###`","docs/a.md"]
```
yufeih: why is it still "warning" ?

Issue #3987, opened 2019-01-29
pr builds' metadata is different target builds'


OsmondJiang:
![image](https://user-images.githubusercontent.com/19990166/51898935-ed2e4f80-23ec-11e9-9349-94803324af7e.png)

We may want to use configuration to get the repo + branch instead of git info itself.

@yufeih @fenxu

OsmondJiang:
The main reason is that the branch and repo info maybe incorrect for PR build. 

I would like to use environment variables like [VSTS build variables](https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azdevops) to setup the current repo url and repo branch, what do you think? @yufeih  @fenxu 

![image](https://user-images.githubusercontent.com/19990166/51953796-b140cc00-2478-11e9-8b48-784cd7937d1c.png)

OsmondJiang:
# Background

During PR build, we found the metadata generated is different with target branch build, the only different is the `original_content_git_url`.

The difference has impacts to our incremental publish, which cause all PR publish is **FULL publish**. We only want to publish modified files.

The main reason is that docfx get the repo info(url + branch) from local git directly, which is incorrect during PR build, because the PR build using branch info fetched from GITHUB which is actually PR name.

# Solution

We want make this metadata `original_content_git_url` generated in PR build same with target branch builld's, which means the repo info(url + branch) provided by local git is not correct. 

Docfx need a way to get the git info(url + branch) from the caller: using environment variables `DOCFX_REPOSITORY_URL` and `DOCFX_REPOSITORY_BRANCH`.

Docfx will try to get repo url and repo branch info from environment variables firstly, if they are not set, using local git as fallback.

# Work Items

- [ ] Retrieve repo url and repo branch info from environment variable for docfx build
- [ ] Build service setup these two environment variables before run docfx.

# Propose Change

```C#
Repository Repository.Create(string docsetPath)
{
    return Repository.Create(docsetPath, AppData.GetRepoUrl(), AppData.GetBranchName());
}
string AppData.GetRepoUrl()
{
    return Environment.GetEnvironmentVariable("DOCFX_REPOSITORY_URL");
}
string AppData.GetBranchName()
{
    return Environment.GetEnvironmentVariable("DOCFX_REPOSITORY_BRANCH");
}
```

# Open Question

How about the sub modules in docset?

cc @yufeih  @fenxu  @superyyrrzz

fenxu:
[...]

PR #4234, opened 2019-03-25
Prefer existing user name when updating user cache from commit

yufeih:
This makes github user cache stable before an record is expired. Fixes PR unexpected change caused by display name.

image

### DIFF
--- a/src/docfx/lib/github/GitHubUserCache.cs
+++ b/src/docfx/lib/github/GitHubUserCache.cs
@@ -299,7 +299,9 @@ namespace Microsoft.Docs.Build
                     user.Id = existingUser.Id;
                 if (user.Login is null)
                     user.Login = existingUser.Login;
-                if (user.Name is null || preferExistingName)
+                if (user.Name is null)
+                    user.Name = existingUser.Name;
+                if (preferExistingName && existingUser.Name != null)
                     user.Name = existingUser.Name;
                 user.Emails = user.Emails.Concat(existingUser.Emails).Distinct().ToArray();
             }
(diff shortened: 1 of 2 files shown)


OsmondJiang:
#3987

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -248,7 +248,7 @@ async Task<(Error error, bool collide)> SaveRemote(string content)
         /// 2. User missing with the specified login: { "login": "..." }
         /// 3. User missing with the specified email: { "emails": [ "..." ] }
         /// </summary>
-        private void UpdateUser(GitHubUser user)
+        private void UpdateUser(GitHubUser user, bool preferExistingName = false)
```
OsmondJiang: when will it be false?

yufeih: All other places where users are updated:
- Users returned from _get by login` API
- Users read from cache file itself

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -291,25 +291,25 @@ private void UpdateUser(GitHubUser user)
...
+        private static void MergeUser(GitHubUser user, GitHubUser existingUser, bool preferExistingName)
         {
             if (existingUser.IsValid())
             {
                 if (user.Id is null)
                     user.Id = existingUser.Id;
                 if (user.Login is null)
                     user.Login = existingUser.Login;
-                if (user.Name is null)
+                if (user.Name is null || preferExistingName)
```
superyyrrzz: `if (user.Name is null || preferExistingName && existingUser.Name != null)` ?

PR #4237, opened 2019-03-25
Add line info for bookmark validation

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/75652/
- Add line info of markdown files
- Add TODO for TOC
- No usage of bookmark for breadcrumb path

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1070,26 +1070,6 @@ outputs:
   build.log: |
     ["error","output-path-conflict","Two or more files output to the same path 'docs/toc.json': 'docs/TOC.md', 'docs/TOC.yml'"]
 ---
-# validate bookmarks in toc
-inputs:
-  docfx.yml:
-  docs/a.md: |
-    # title 1
-  docs/TOC.md: |
-    # [link to title 1](a.md#title-1)
-    # [link to title 2](a.md#title-2)
-outputs:
-  docs/a.json:
-  docs/toc.json: |
-    {
-      "items": [
-        { "name": "link to title 1", "href": "a#title-1", "!monikers":null },
-        { "name": "link to title 2", "href": "a#title-2", "!monikers":null }
-      ]
-    }
-  build.log: |
-    ["warning","external-bookmark-not-found","Cannot find bookmark '#title-2' in 'docs/a.md', did you mean '#title-1'?","docs/TOC.md"]
----
 # reference uid within link
 inputs:
   docfx.yml:
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1070,6 +1070,22 @@ outputs:
   build.log: |
     ["error","output-path-conflict","Two or more files output to the same path 'docs/toc.json': 'docs/TOC.md', 'docs/TOC.yml'"]
 ---
+# validate bookmarks in toc	
+inputs:	
+  docfx.yml:	
+  docs/a.md: |	
+    # title 1	
+  docs/TOC.md: |	
+    # [link to title 1](a.md#title-1)	
+outputs:	
+  docs/a.json:	
+  docs/toc.json: |	
+    {	
+      "items": [	
+        { "name": "link to title 1", "href": "a#title-1", "!monikers":null },	
+      ]	
+    }
+---
 # reference uid within link
 inputs:
   docfx.yml:
--- a/src/docfx/build/dependency/DependencyResolver.cs
+++ b/src/docfx/build/dependency/DependencyResolver.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Docs.Build
 
         // forLandingPage should not be used, it is a hack to handle some specific logic for landing page based on the user input for now
         // which needs to be removed once the user input is correct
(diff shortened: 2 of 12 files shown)


src/docfx/build/page/BookmarkValidator.cs
```diff
@@ -38,14 +38,14 @@ public List<(Error error, Document file)> Validate()
...
-            foreach (var (file, reference, bookmark) in _references)
+            foreach (var (file, reference, bookmark, range) in _references)
             {
                 if (_bookmarksByFile.TryGetValue(reference, out var bookmarks) && bookmarks.Contains(bookmark))
                 {
                     continue;
                 }
-                result.Add(file == reference ? (Errors.InternalBookmarkNotFound(file, bookmark, bookmarks), file) :
-                                               (Errors.ExternalBookmarkNotFound(file, reference, bookmark, bookmarks), file));
+                result.Add(file == reference ? (Errors.InternalBookmarkNotFound(file, bookmark, bookmarks).WithRange(range), file) :
```
yufeih: nit: if the InternalBookmarkNotFound method takes an `in Range` then you don't need `WithRange` to create a copy of `Error`

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -36,7 +36,7 @@ public DependencyResolver(GitCommitProvider gitCommitProvider, BookmarkValidator
 
         // forLandingPage should not be used, it is a hack to handle some specific logic for landing page based on the user input for now
         // which needs to be removed once the user input is correct
-        public (Error error, string link, Document file) ResolveLink(string path, Document relativeTo, Document resultRelativeTo, Action<Document> buildChild, bool forLandingPage = false)
+        public (Error error, string link, Document file) ResolveLink(string path, Document relativeTo, Document resultRelativeTo, Action<Document> buildChild, Range range, bool forLandingPage = false)
```
yufeih:
```suggestion
        public (Error error, string link, Document file) ResolveLink(string path, Document relativeTo, Document resultRelativeTo, Action<Document> buildChild, in Range range, bool forLandingPage = false)
```

src/docfx/build/page/BuildPage.cs
```diff
@@ -26,7 +26,8 @@ internal static class BuildPage
 
             if (!string.IsNullOrEmpty(metadata.BreadcrumbPath))
             {
-                var (breadcrumbError, breadcrumbPath, _) = context.DependencyResolver.ResolveLink(metadata.BreadcrumbPath, file, file, buildChild);
+                // there is no bookmark validation for breadcrumb path, so no need of range here
+                var (breadcrumbError, breadcrumbPath, _) = context.DependencyResolver.ResolveLink(metadata.BreadcrumbPath, file, file, buildChild, new Range(0, 0));
```
yufeih:
```suggestion
                var (breadcrumbError, breadcrumbPath, _) = context.DependencyResolver.ResolveLink(metadata.BreadcrumbPath, file, file, buildChild, default);
```

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -26,7 +26,7 @@ public static class E2ETest
             "need-restore", "publish-url-conflict", "output-path-conflict", "download-failed", "heading-not-found", "config-not-found",
 
             // These error codes are the ones we could have line info but haven't implement them yet:
-            "external-bookmark-not-found", "internal-bookmark-not-found", "file-not-found", "uid-not-found", "committish-not-found",
+            "external-bookmark-not-found", "file-not-found", "uid-not-found", "committish-not-found",
```
yufeih:
```suggestion
           "file-not-found", "uid-not-found", "committish-not-found",
```

live1206: `external-bookmark-not-found` could not be removed for now, because there is a TOC bookmark validation which involves it

src/docfx/build/dependency/DependencyResolver.cs
live1206: why do we need reference input here? `range` is not in use for the caller

yufeih: `Range` is currently a `struct` and `in` makes it pass by ref instead of copying the whole value.

OsmondJiang: what's the benefit to make it as a struct? it's used a lot in Errors.cs as parameters

live1206: > `Range` is currently a `struct` and `in` makes it pass by ref instead of copying the whole value.

Ah, I remember now :)

test/docfx.Test/e2e/E2ETest.cs
live1206: remove this test to make sure `external-bookmark-not-found` warning with line info

docs/specs/toc.yml
```diff
@@ -1070,26 +1070,6 @@ outputs:
   build.log: |
     ["error","output-path-conflict","Two or more files output to the same path 'docs/toc.json': 'docs/TOC.md', 'docs/TOC.yml'"]
 ---
-# validate bookmarks in toc
```
OsmondJiang: don't remove the whole test, just remove the invalid bookmark reference `title-2`

live1206: ok

PR #4238, opened 2019-03-25
fix abstract for landing page

live1206:


src/Microsoft.Docs.Template/Views/Template/LandingData.cshtml
```diff
@@ -30,7 +30,7 @@ private void UpdateAbstract()
         @if (!string.IsNullOrEmpty(Model.Abstract?.Menu?.Title) || (Model.Abstract?.Menu?.Items != null && Model.Abstract.Menu.Items.Any()))
         {
             <div class="menu">
-                if (!string.IsNullOrEmpty(Model.Abstract?.Menu?.Title))
+                @if (!string.IsNullOrEmpty(Model.Abstract?.Menu?.Title))
```
OsmondJiang: could we add a legacy spec test for that?

live1206: I think it would be hard to add the template tests at this moment

PR #4240, opened 2019-03-25
🐛fix h1 remove and extract logic

OsmondJiang:
fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/76616

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -170,19 +170,19 @@ namespace Microsoft.Docs.Build
 
             if (titleNode != null)
             {
-                var previousSibling = titleNode.PreviousSibling;
-                while (previousSibling != null)
+                var previousVisibleNode = titleNode.PreviousSibling;
+                while (previousVisibleNode != null)
                 {
-                    if (previousSibling.NodeType == HtmlNodeType.Comment ||
-                        (previousSibling.NodeType == HtmlNodeType.Text && string.IsNullOrWhiteSpace(previousSibling.OuterHtml)))
+                    if (previousVisibleNode.NodeType == HtmlNodeType.Comment ||
+                        (previousVisibleNode.NodeType == HtmlNodeType.Text && string.IsNullOrWhiteSpace(previousVisibleNode.OuterHtml)))
                     {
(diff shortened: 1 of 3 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
+using System.Linq;
```
mingwli:
```suggestion
```

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,27 +164,24 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+            var previousVisableNode = false;
             foreach (var child in node.ChildNodes)
             {
-                if (child.NodeType == HtmlNodeType.Comment)
+                if (child.NodeType != HtmlNodeType.Comment &&
+                    !(child.NodeType == HtmlNodeType.Text && string.IsNullOrWhiteSpace(child.OuterHtml)))
                 {
-                    continue;
-                }
+                    if (child.NodeType == HtmlNodeType.Element && (child.Name == "h1" || child.Name == "h2" || child.Name == "h3"))
```
yufeih: Should check if all the nodes before h1 is visible instead of just the previous one, e.g.
```
asdf

<!---comment-->
h1

```

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,27 +164,24 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
         }
 
-        public static HtmlNode ExtractTitle(HtmlNode node)
+        public static HtmlNode RemoveTitle(HtmlNode node)
         {
+            var previousVisableNode = false;
             foreach (var child in node.ChildNodes)
             {
-                if (child.NodeType == HtmlNodeType.Comment)
+                if (child.NodeType != HtmlNodeType.Comment &&
```
yufeih: consider extracting a method IsHtmlNodeVisible

src/docfx/lib/HtmlUtility.cs
OsmondJiang: just check any one is visible is enough, no need to check all are invisible

PR #4247, opened 2019-03-26
fix unexpected warning on self bookmark in included file

928PJY:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/76318

### DIFF
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -58,13 +58,13 @@ inputs:
     [!INCLUDE[](includes/b.md)]
   docs/includes/b.md: |
     ## test
-    Link to [self](#test)
-    Link to [self](../a.md/#heading)
-    Link to [self](#heading1)
-    Link to [self](../a.md/#heading2)
+    Link to [valiad self](#test)
+    Link to [valiad parent](../a.md/#heading)
+    Link to [invalid self](#heading1)
+    Link to [invalid parent](../a.md/#heading2)
 outputs:
   docs/a.json: |
-    { "content": "<h2 id=\"test\">test</h2>\n<p>Link to <a href=\"#test\">self</a>\nLink to <a href=\"a#heading\">self</a>\nLink to <a href=\"#heading1\">self</a>\nLink to <a href=\"a#heading2\">self</a></p>" }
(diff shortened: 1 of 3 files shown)


docs/specs/resolve.yml
```diff
@@ -49,6 +49,20 @@ outputs:
...
+# Links to self bookmarks in the included file
+inputs:
+  docfx.yml: |
+    files: docs/a.md
+  docs/a.md: |
+    # heading
+    [!INCLUDE[](includes/b.md)]
+  docs/includes/b.md: |
+    ## test
+    Link to [self](#test)
```
OsmondJiang: add one new line 
`Link to [parent](#heading)`?

live1206:
As we offline discussed, do we support bookmark in inclusion referencing to the parent file?

928PJY:
yes, we support it, they can reference the parent file with explicit path instead self bookmark,

---
In reply to: [476936095](https://github.com/dotnet/docfx/pull/4247#issuecomment-476936095) [](ancestors = 476936095)

docs/specs/resolve.yml
```diff
@@ -49,6 +49,26 @@ outputs:
...
+    ## test
+    Link to [valiad self](#test)
+    Link to [valiad parent](../a.md/#heading)
+    Link to [invalid self](#heading1)
+    Link to [invalid parent](../a.md/#heading2)
+outputs:
+  docs/a.json: |
+    { "content": "<h2 id=\"test\">test</h2>\n<p>Link to <a href=\"#test\">valiad self</a>\nLink to <a href=\"a#heading\">valiad parent</a>\nLink to <a href=\"#heading1\">invalid self</a>\nLink to <a href=\"a#heading2\">invalid parent</a></p>" }
+  build.log: |
+    ["suggestion","internal-bookmark-not-found","Cannot find bookmark '#heading2' in 'docs/includes/b.md', did you mean '#heading'?","docs/includes/b.md","",5,9]
```
OsmondJiang: it's the #heading2 is treated as internal bookmark, then could we add a case `Link to [valiad parent](#heading)`

928PJY: we don't recommend the user to use selfbookmark to reference the  parent

---
In reply to: [269406727](https://github.com/dotnet/docfx/pull/4247#discussion_r269406727) [](ancestors = 269406727)

PR #4251, opened 2019-03-27
remove include-is-redirection

mingwli:
only resolve content from toc/md


mingwli:
resolving https://ceapex.visualstudio.com/Engineering/_workitems/edit/77046

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -92,11 +92,6 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
         {
             var (error, file, redirect, _, _, _, pathToDocset) = TryResolveFile(relativeTo, href);
 
-            if (redirect != null)
-            {
-                return (Errors.IncludeIsRedirection(relativeTo, href), null, null);
```
yufeih: does `return default` works without changing other parts of the code?

PR #4252, opened 2019-03-27
only assign value to raw title when title is removed from html

OsmondJiang:
fix: https://ceapex.visualstudio.com/web/wi.aspx?pcguid=7d644393-99ad-41c8-ac53-7fa79294c720&id=77409

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -78,7 +78,7 @@ outputs:
   h4.json: |
     {
       "!title": null,
-      "!raw_title": null,
+      "raw_title": "",
       "content": "<h4 id=\"h4-title\">h4 title</h4>",
     }
   build.log: |
@@ -97,11 +97,13 @@ inputs:
 outputs:
   h1toremove.json: |
     {
-      "content":"<div></div>"
+      "content":"<div></div>",
+      "raw_title": "<h1 id=\"h1-title\">h1 title</h1>"
     }
   h1tokeep.json: |
     {
-      "content":"<p>welcome to docfx</p>\n<h1 id=\"h1-title\">h1 title</h1>\n"
+      "content":"<p>welcome to docfx</p>\n<h1 id=\"h1-title\">h1 title</h1>\n",
+      "raw_title": ""
     }
 ---
 # No H1
@@ -182,7 +184,7 @@ inputs:
     hello
 outputs:
   docs/a.json: |
-    { "raw_title": "<h1 id=\"title-from-h1\">Title from H1</h1>" }
+    { "raw_title": "", "title": "Title from yaml header" }
 ---
 # include another file
 inputs:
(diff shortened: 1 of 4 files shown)


docs/specs/markdown.yml
```diff
@@ -78,7 +78,7 @@ outputs:
   h4.json: |
     {
       "!title": null,
-      "!raw_title": null,
+      "raw_title": "",
```
yufeih: Does this conflict with https://github.com/dotnet/docfx/commit/e55823dd9777c157ca5e1282a20e3d6ccfc69cda, since `""` and `undefined` is treated differently for raw_title?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,7 +163,10 @@ public static string TransformLinks(string html, Func<string, string> transform)
             return result.ToString();
         }
 
-        public static HtmlNode RemoveTitle(HtmlNode node)
+        /// <summary>
+        /// Get title node and remove title node if all previous nodes are invisible
+        /// </summary>
+        public static (HtmlNode headerNode, bool removed) ExtractTitle(HtmlNode node)
```
yufeih: consider 
 - Having 2 methods `string GetTitle` and `string ExtractRawTitle` to indicate `rawTitle` and `title` are distinct properties
--or--
 - Having 1 method that returns `(string title, string rawTitle, HtmlDom content) SplitHtml`

That might better express the intend then leaking the `removed` intermediate state to the caller

OsmondJiang: ExtractTitle method has clear described this method will return header node and try remove header node. 

And RawTitle is just to reuse the result returned from this method, for me it's really clear.

docs/specs/markdown.yml
OsmondJiang: no, rawTitle is different with title, based on current template implementation and v2 logic https://github.com/dotnet/docfx/blob/da48dd4c820f4b15dbcf4cc350ab68fabc5292b4/src/Microsoft.DocAsCode.Build.ConceptualDocuments/HtmlDocumentUtility.cs#L32

src/docfx/lib/HtmlUtility.cs
```diff
@@ -164,28 +164,36 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+
+            title = null;
+            rawTitle = string.Empty;
             foreach (var child in node.ChildNodes)
             {
                 if (!IsInvisibleNode(child))
                 {
                     if (child.NodeType == HtmlNodeType.Element && (child.Name == "h1" || child.Name == "h2" || child.Name == "h3"))
                     {
+                        title = child.InnerText == null ? null : HttpUtility.HtmlDecode(child.InnerText);
```
live1206: suggest using `is null`

PR #4253, opened 2019-03-27
revert to resolve redirection to unblock impact

mingwli:


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -221,7 +221,7 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
                         //
                         // TODO: In case of file rename, we should warn if the content is not inside build scope.
                         //       But we should not warn or do anything with absolute URLs.
-                        var (error, redirectFile) = Document.TryCreate(relativeTo.Docset, pathToDocset, redirectTo);
```
OsmondJiang: fix at here: https://github.com/dotnet/docfx/blob/v3/src/docfx/docset/Document.cs#L368

PR #4256, opened 2019-03-27
fix internal bookmark validation error:

928PJY:
https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/10072?path=%2Fsource%2Fazure-docs-pr.output%2Fbuild.log&_a=files

### DIFF
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -56,6 +56,7 @@ inputs:
   docs/a.md: |
     # heading
     [!INCLUDE[](includes/b.md)]
+    Link to [invaliad self](./a.md#heading3)
   docs/includes/b.md: |
     ## test
     Link to [valiad self](#test)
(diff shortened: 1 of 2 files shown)


docs/specs/resolve.yml
```diff
@@ -56,6 +56,7 @@ inputs:
   docs/a.md: |
     # heading
     [!INCLUDE[](includes/b.md)]
+    Link to [invaliad self](./a.md#heading3)
```
yufeih:
```suggestion
    Link to [invalid self](./a.md#heading3)
```

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -159,13 +159,13 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
                     }
                     var selfUrl = Document.PathToRelativeUrl(
                         Path.GetFileName(file.SitePath), file.ContentType, file.Schema, file.Docset.Config.Output.Json);
-                    return (error, selfUrl + query + fragment, fragment, hrefType, null);
```
live1206: should `hrefType` be `HrefType.SelfBookmark` in the first place? I mean, the output of https://github.com/dotnet/docfx/blob/c55b52837b77150f9c37020323231e9ca3129321/src/docfx/build/dependency/DependencyResolver.cs#L196

928PJY: it should be, but to keep the behavior same with v2([Self](a.md#test) => <a href=“a#test”>), we cannot generate the right hrefType  in the first place

Issue #4259, opened 2019-03-28
Code `TODO` clean up and improvement


OsmondJiang:
- [x] cleanup docset model
  - [x] split data and behavioral objects
  - [x] define what configs should passed to dependency docsets and localization docsets (future)
- [x] cleanup document model
  - [x] split data and behavioral objects
- [ ] remove landing page link resolver hacking (future)
- [x] get line info of TOC href for bookmark validation&file-not-found (progressing)
- [x] get ride of `.GetAwaiter().GetResult()` if possible
- [x] reduce unnecessary Async/Await caused by RunInsideMutex()
- [ ] word count is not really correct (future)
- [x] clear model (interface) between docfx and template (progressing)
- [x] add missing line info (progressing)
- [x] Change `Range` from `Struct` to `Class` to avoid write `in` everywhere (progressing)
- [x] combine restore map and dependency lock model in build pipeline

yufeih:
For loc:
- Loc docset is right now a doubly linked list (docset <--> localization docset <--> fallback docset), a singly linked list is easier to reason if we just have (docset --> fallback docset)
- If it is too complicated to support building loc from english, we can drop folder convention and only support build from loc.

OsmondJiang:
For Config:

- [ ] Simplify `Global Config`. `Extended Config` and `Overwrite Config`
- [ ] Define `options`, `environment variables` and `config`

PR #4262, opened 2019-03-28
⚒split docset and template

OsmondJiang:
#4259

### DIFF
--- a/src/docfx/template/TemplateEngine.cs
+++ b/src/docfx/template/TemplateEngine.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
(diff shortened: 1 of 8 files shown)


src/docfx/template/TemplateEngine.cs
```diff
@@ -29,11 +30,27 @@ public TemplateEngine(string templateDir, string locale)
...
+
+            return new TemplateEngine(themePath, docset.Locale);
+        }
+
         public string Render(PageModel model, Document file)
         {
             // TODO: only works for conceptual
             var content = model.Content.ToString();
-            var (templateModel, metadata) = TemplateTransform.Transform(model, file);
+            var (templateModel, metadata) = TemplateTransform.Transform(this, model, file);
```
live1206: seems a bit odd to pass `this` here, should `TemplateTransform` be part of `TemplateEngine`?

OsmondJiang: i am going to leave this to another PR or discussion, @live1206  @yufeih

PR #4265, opened 2019-03-29
🛠Consolidate JsonUtility and YamlUtility to share the same interface

yufeih:
There are some in-consistency between the public interface of `JsonUtility` and `YamlUtility`, this PR:

- Makes all `Parse` and `ToObject` method return an additional `error`, which means the input needs schema validation and any errors or warnings are reported through that return value.
- Makes all `DeserializeData` method turn the result directly, which means the input is a data file and we don't do validation for them, we'll throw if the data is corrupted.
- Removes `nullValidation` since they are always true (except for tests)

- Fix a TOC metadata validation bug caused by not calling the correct `ToObject` method
- Fix yaml header error does not have line info

@live1206 , do you know why this is changed? https://github.com/dotnet/docfx/pull/4194/files#diff-531ae800ae2cae4aa437350ad3816b00R102 It makes Yaml using `Deserialize` but leaves Json as `DeserializeWithSchemaValidation<JToken>`.

#4259


live1206:
Both `docfx.yml` and `docfx.json` should use `Deserialize` instead of `DeserializeWithSchemaValidation`, I am changing it now while adding back line info

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -207,7 +207,7 @@ private async Task LoadCommitsTime(Docset docset)
             {
                 var (_, content, _) = await RestoreMap.GetRestoredFileContent(docset, docset.Config.Contribution.GitCommitsTime);
 
-                foreach (var commit in JsonUtility.Deserialize<GitCommitsTime>(content).Commits)
+                foreach (var commit in JsonUtility.DeserializeData<GitCommitsTime>(content).Commits)
```
OsmondJiang: `DeserializeData` and `Deserialize` is not clear than previous `Deserialize` and `DeserializeWithSchemaValidation` from the method name.

yufeih: Both are fine, as long as they are consistent, right now JToken parse with schema validation is also called `(Error, JToken) Deseriaize(string)`. I tend to want to make  `DeserializeWithSchemaValidation` the default behavior for docfx to detect more warnings, so called it `Deserialize`

live1206: https://github.com/dotnet/docfx/blob/7e9e7db60909ad1ea80fe22d2a6b7a4e90029c0d/src/docfx/lib/YamlUtility.cs#L82
https://github.com/dotnet/docfx/blob/7e9e7db60909ad1ea80fe22d2a6b7a4e90029c0d/src/docfx/lib/YamlUtility.cs#L87
These two were missing during the last renaming

OsmondJiang: From the usage of these two methods, seems `DeserializeData` is used more than `Deserialize`, apart from the tests.

test/docfx.Test/lib/JsonUtilityTest.cs
```diff
@@ -47,7 +47,7 @@ public void TestObjectWithStringProperty(string input)
             var sw = new StringWriter();
             JsonUtility.Serialize(sw, new BasicClass { C = input });
             var json = sw.ToString();
-            var (errors, value) = JsonUtility.DeserializeWithSchemaValidation<BasicClass>(json);
+            var (errors, value) = Deserialize<BasicClass>(json);
```
OsmondJiang: this Deserialize will cause confusion with JsonUtility/YamlUtility,Deserialize, rename it to DeserializeWithSchemaValidation?

PR #4267, opened 2019-03-29
👨‍🏭combine template engine and transform to reduce confusion

OsmondJiang:
Combine the template engine and transform together, to provide `render`, `transform(model+metadata)` and `transformMetadata`.

#4259


OsmondJiang:
> can you keep the file `TemplateTransform`? It is related to `TemplateEngine` but could be treated as a layer below TemplateEngine. I'd like to keep two files where `TemplateTransform` is right now a place mostly for backward compatibility. You could rename it to `TemplateEngineTransform` indicate the relationship.

From my view, template engine is to `render` and `transform`, so it's not reasonable to split `TransformMetadata` and `TransformMetadataAndModel` to two classes, which causes confusions.

`TemplateEngine` actually also depends on `TemplateTransform`, so from legacy view point, they are all run in legacy model.

mingwli:
> > can you keep the file `TemplateTransform`? It is related to `TemplateEngine` but could be treated as a layer below TemplateEngine. I'd like to keep two files where `TemplateTransform` is right now a place mostly for backward compatibility. You could rename it to `TemplateEngineTransform` indicate the relationship.
> 
> From my view, template engine is to `render` and `transform`, so it's not reasonable to split `TransformMetadata` and `TransformMetadataAndModel` to two classes, which causes confusions.
> 
> `TemplateEngine` actually also depends on `TemplateTransform`, so from legacy view point, they are all run in legacy model.

yet the `TemplateEngine.TransformMetadata` and `TemplateTransform.Transform` do have loop reference

PR #4270, opened 2019-03-29
🛠Keep the line info of config schema validation

live1206:


src/docfx/config/ConfigLoader.cs
```diff
@@ -154,8 +157,11 @@ private static (List<Error>, JObject) ExtendConfigs(JObject config, string docse
                 }
             }
 
-            JsonUtility.Merge(result, config);
-            return (errors, result);
+            // to keep the line info of config
```
OsmondJiang: may we extract this as a method?

live1206: Thought it was not a good idea to put the workaround into lib, but maybe it's ok

src/docfx/lib/JsonUtility.cs
```diff
@@ -268,6 +268,17 @@ public static void Merge(JObject container, JObject overwrite)
             container.Merge(overwrite, s_mergeSettings);
         }
 
+        /// <summary>
+        /// Merge overwrite into container
+        /// And keep the line info of properties in overwrite which are not in container
+        /// </summary>
+        public static void MergeWithLineInfo(JObject container, JObject overwrite)
+        {
+            var original = overwrite.DeepClone() as JObject;
```
OsmondJiang: how about make it as default behavior?

live1206: sounds good

live1206: Actually, it depends if we need this workaround

OsmondJiang: move this method to configLoader as a private method

PR #4271, opened 2019-03-29
👨‍🏭remove un-necessary await/async method and prepare for mutex changes

OsmondJiang:
a lot of async/await methods are caused by RunInsideMutex, but this method can be synchronous, and the actions executed inside mutex are mostly synchronous.

So remove un-necessary asynchronous methods, also a prepare work for mutex implementation changing work

### DIFF
--- a/test/docfx.Test/lib/ProcessUtilityTest.cs
+++ b/test/docfx.Test/lib/ProcessUtilityTest.cs
@@ -38,21 +38,21 @@ namespace Microsoft.Docs.Build
         }
 
         [Fact]
-        public static async Task FileMutexTest()
+        public static void FileMutexTest()
         {
             var concurrencyLevel = 0;
             var fileName = $"process-test/{Guid.NewGuid()}";
 
             try
             {
-                await ParallelUtility.ForEach(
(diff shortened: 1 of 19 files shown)


OsmondJiang:
#4259

test/docfx.Test/lib/ProcessUtilityTest.cs
```diff
@@ -63,56 +63,56 @@ public static async Task FileMutexTest()
...
-        public static async Task ParallelNestedRunInMutexWithSameNameTest()
+        public static void ParallelNestedRunInMutexWithSameNameTest()
         {
             var name = Guid.NewGuid().ToString();
 
-            await ProcessUtility.RunInsideMutex($"process-test/123", async () =>
+            ProcessUtility.RunInsideMutex($"process-test/123", () =>
             {
-                await ParallelUtility.ForEach(new[] { 1, 2, 3, 4, 5 }, async i =>
+                Parallel.ForEach(new[] { 1, 2, 3, 4, 5 }, i =>
```
live1206: monir: `ParallelUtility` also contains a synchronized `Foreach`

PR #4272, opened 2019-03-29
🔗Add line info for file-not-found

live1206:
- update config model to include line info where needed


src/docfx/restore/RestoreMap.cs
```diff
@@ -72,12 +72,12 @@ public bool Release()
...
 
-        public static (string localPath, string content, string etag) GetRestoredFileContent(Docset docset, string url)
+        public static (string localPath, string content, string etag) GetRestoredFileContent(Docset docset, string url, in Range range)
         {
-            return GetRestoredFileContent(docset.DocsetPath, url, docset.FallbackDocset?.DocsetPath);
+            return GetRestoredFileContent(docset.DocsetPath, url, range, docset.FallbackDocset?.DocsetPath);
         }
 
-        public static (string localPath, string content, string etag) GetRestoredFileContent(string docsetPath, string url, string fallbackDocset = null)
+        public static (string localPath, string content, string etag) GetRestoredFileContent(string docsetPath, string url, in Range range, string fallbackDocset = null)
```
OsmondJiang: how about change the `url` to `SourceInfo<string>`

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -200,7 +200,7 @@ private void LoadCommitsTime(Docset docset)
         {
             if (!string.IsNullOrEmpty(docset.Config.Contribution.GitCommitsTime))
             {
-                var (_, content, _) = RestoreMap.GetRestoredFileContent(docset, docset.Config.Contribution.GitCommitsTime);
+                var (_, content, _) = RestoreMap.GetRestoredFileContent(docset, docset.Config.Contribution.GitCommitsTime, docset.Config.Contribution.GitCommitsTime.Range);
```
superyyrrzz: Add an overload `GetRestoredFileContent(Docset docset, SourceInfo<string> url)`? Seems it needn't 3 params. Same for `DependencyLock.Load`

live1206: updated

PR #4288, opened 2019-04-02
🐛fix bug of key already added

OsmondJiang:
fix: https://ceapex.visualstudio.com/Engineering/_build/results?buildId=47926


OsmondJiang:
The root cause is causing by parallel getting one file commits without any cache, the init cache adding(multiple times) for this file will throw that exception, @yufeih

PR #4290, opened 2019-04-02
Decouple get link logic from v2

yufeih:
This PR picks up the latest markdown extension changes.

#4078


docs/specs/xref.yml
```diff
@@ -436,7 +436,7 @@ inputs:
 outputs:
   docs/a.json:
   docs/b.json: |
-    { "content": "<p>Link to <a href=\"xref:c#bookmark\">title</a></p>" }
+    { "content": "<p>Link to <a href=\"\">title</a></p>" }
```
live1206: is this a behavior change?

yufeih: it is, does it currently impact our comparison with v2?

live1206: I think it should be as it is written while uid not resolved

live1206: will this affect v2 as well?

yufeih: nop, it'll not affect v2.

live1206: then we are having a behavior change compared to v2

PR #4291, opened 2019-04-02
log file which causes build collapse

OsmondJiang:


src/docfx/build/Build.cs
```diff
@@ -182,11 +182,17 @@ void ValidateBookmarks()
...
+                if (DocfxException.IsDocfxException(ex, out var dex))
+                {
+                    context.Report.Write(file.ToString(), dex.Error);
+                    context.PublishModelBuilder.MarkError(file);
+                    return new List<string>();
+                }
+
+                Console.WriteLine($"Build {file.FilePath} failed");
+                throw;
             }
```
yufeih: nit: you can nest catch like this:
```
catch (Exception ex) when (DocfxException.IsDocfxException(ex, out var dex))
{
    xxx
}
catch
{
    Console.WriteLine(xxx);
    throw;
}
```

PR #4292, opened 2019-04-02
Add namespace to CommitsTimeItem

yufeih:


src/docfx/build/metadata/CommitsTimeItem.cs
```diff
@@ -4,11 +4,14 @@
 using System;
 using Newtonsoft.Json;
 
-internal class CommitsTimeItem
+namespace Microsoft.Docs.Build
```
OsmondJiang: is there a run to valid?

yufeih: I cannot find an existing stylecop rule that enforce that

928PJY:
what is this class used for?

yufeih:
> what is this class used for?

A data model to resolve updated at based on git commit time

src/docfx/build/metadata/CommitsTimeItem.cs
```diff
@@ -4,11 +4,14 @@
...
-    [JsonProperty("sha")]
-    public string Sha { get; set; }
+    internal class CommitsTimeItem
+    {
+        [JsonProperty("sha")]
+        public string Sha { get; set; }
 
-    [JsonProperty("built_at")]
-    public DateTime BuiltAt { get; set; }
+        [JsonProperty("built_at")]
```
live1206: why not just use `[JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]`?

PR #4293, opened 2019-04-02
Make ProcessUtility.Execute sync

yufeih:
#4047 #4259


src/docfx/lib/git/GitUtility.cs
```diff
@@ -314,23 +311,23 @@ private static async Task CloneOrUpdate(string path, string url, IEnumerable<str
 
             try
             {
-                await ExecuteNonQuery(path, $"{httpConfig} fetch --tags --progress --update-head-ok {pruneSwitch} {depth} \"{url}\" {refspecs}", stderr: true);
+                ExecuteNonQuery(path, $"{httpConfig} fetch --tags --progress --update-head-ok {pruneSwitch} {depth} \"{url}\" {refspecs}");
             }
-            catch (InvalidOperationException ex) when (committishes.Any(rev => ex.Message.Contains(rev)))
+            catch (InvalidOperationException)
```
yufeih: removed `stderr` from `ProcessUtility.Execute`, we'll retry here in most cases

src/docfx/lib/log/DocfxException.cs
```diff
@@ -21,8 +21,9 @@ public DocfxException(Error error, Exception innerException = null)
 
         public static bool IsDocfxException(Exception ex, out DocfxException docfxException)
         {
-            docfxException = ex as DocfxException ?? ex.InnerException as DocfxException;
-            return docfxException != null;
+            docfxException = ex as DocfxException;
+
+            return docfxException != null ? true : IsDocfxException(ex.InnerException, out docfxException);
         }
```
yufeih: This is because of nested `AggregatedException`

src/docfx/lib/ProcessUtility.cs
```diff
@@ -181,81 +179,30 @@ public static bool ReleaseExclusiveLock(string lockName, string acquirer)
...
-                }
-                else
-                {
-                    var message = $"'\"{fileName}\" {commandLineArgs}' failed in directory '{cwd}' with exit code {process.ExitCode}: \nSTDOUT:'{output}'\nSTDERR:\n'{error}'";
-
-                    tcs.TrySetException(new InvalidOperationException(message));
-                }
-            };
+            var process = Process.Start(psi);
+            var result = stdout ? process.StandardOutput.ReadToEnd() : null;
```
OsmondJiang: will this meet out of buffer issue?

yufeih: I think it shouldn't hang as long as `ReadToEnd` is called before `WaitForExit` and we only read `stdout`. Test it with `git log` on azure-docs, it completed successfully.

src/docfx/lib/git/GitUtility.cs
OsmondJiang: what about the log info for debugging when the execution failed?

yufeih: `stderr` will be piped to output, so you can see that from the log

PR #4294, opened 2019-04-02
🔗Add line info for Toc href validation

live1206:
- build a line info map key as `name+value`
- use the previous map to get line info while resolving the link
- TODO: line info for TOC.md href validation

PR #4298, opened 2019-04-03
Add line info to github-user-not-found using SourceInfo<T>

yufeih:
- Add line info to `github-user-not-found` message
- Added a `SourceInfo<T>` to pass line info down to strongly typed models
- Make `JsonUtility.Merge` respect line info

#4294


src/docfx/lib/json/JsonUtility.cs
```diff
@@ -238,7 +245,18 @@ public static (List<Error>, JToken) Parse(string json)
...
         public static void Merge(JObject container, JObject overwrite)
         {
-            container.Merge(overwrite, s_mergeSettings);
+            foreach (var (key, value) in overwrite)
+            {
+                if (container[key] is JObject containerObj && value is JObject overwriteObj)
+                {
+                    Merge(containerObj, overwriteObj);
+                }
+                else if (IsNullOrUndefined(container[key]) || !IsNullOrUndefined(value))
```
live1206: how to handle JArray?

src/docfx/lib/json/SourceInfoJsonConverter.cs
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
+{
+    internal class SourceInfoJsonConverter : JsonConverter
```
live1206: Separate other converter and contract resolver of `JsonUtility` into an isolated file as well?

src/docfx/lib/json/SourceInfoJsonConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+            return objectType.IsGenericType && objectType.GetGenericTypeDefinition() == typeof(SourceInfo<>);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var range = JsonUtility.ToRange((IJsonLineInfo)reader);
+            var valueType = objectType.GenericTypeArguments[0];
+            var value = serializer.Deserialize(reader, valueType);
+
+            // TODO: populate file info
```
live1206: just out of curiosity, is it possible to get the file info here?

src/docfx/lib/json/JsonUtility.cs
live1206: It seems not trivial to rewrite `Merge`, there is a workaround to keep line info inside overwrite
https://github.com/dotnet/docfx/blob/611e336cb75493dad8a5979306148d8ccabdcdb9/src/docfx/config/ConfigLoader.cs#L167

src/docfx/lib/json/SourceInfoJsonConverter.cs
yufeih: do it in another PR

src/docfx/lib/json/JsonUtility.cs
live1206: Add more tests to make sure the implementation could fulfill our needs

src/docfx/lib/json/SourceInfoJsonConverter.cs
yufeih: through some thread static foo

src/docfx/lib/log/SourceInfo.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace Microsoft.Docs.Build
+{
+    [DebuggerDisplay("{Value}")]
+    public class SourceInfo<T> : ISourceInfo
+    {
+        public readonly T Value;
+
+        public readonly string File;
+
+        public readonly Range Range;
```
OsmondJiang: can we add a `implicit` operator here, so we don't need to change most of the interface？

src/docfx/lib/json/SourceInfoJsonConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
+{
+    internal class SourceInfoJsonConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType)
+        {
+            return objectType.IsGenericType && objectType.GetGenericTypeDefinition() == typeof(SourceInfo<>);
```
OsmondJiang: may have multiple level inherits

src/docfx/lib/json/SourceInfoJsonConverter.cs
```diff
@@ -0,0 +1,31 @@
...
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var range = JsonUtility.ToRange((IJsonLineInfo)reader);
+            var valueType = objectType.GenericTypeArguments[0];
```
OsmondJiang: same here, unless you make SourceInfo sealed

src/docfx/lib/log/SourceInfo.cs
yufeih: I am thinking about turning `Range` into `SourceInfo` eventually, because a line number alone cannot identify the true source. (need both line info + file info)

src/docfx/lib/json/SourceInfoJsonConverter.cs
live1206: make `SourceInfo` sealed?

src/docfx/lib/json/SourceInfoJsonConverter.cs
yufeih: yeah, it could be sealed, I do not want inheritance at all on this.

src/docfx/lib/log/SourceInfo.cs
OsmondJiang: i mean the `value`

yufeih: There was once a weird bug due to implicit conversion of `JValue` https://github.com/dotnet/docfx/blob/v3/src/docfx/cli/CommandLineOptions.cs#L23 when it is used with `? :` and other operators, type inference may produce unexpected result. It used to be `a ? a : null` but was changed to `a ? (JValue)a : JValue.CreateNull()` 

Let's see how it goes for this one

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -99,7 +99,7 @@ async Task<Contributor> GetAuthor()
                     {
                         // Remove author from contributors if author name is specified
                         var (error, result) = await _gitHubUserCache.GetByLogin(authorName);
-                        errors.AddIfNotNull(error);
+                        errors.AddIfNotNull(error?.WithRange(authorName?.Range ?? default));
```
live1206: can just be `authorName`

yufeih: this one takes a `Range` that are not implicitly convertible.

live1206: I see

PR #4300, opened 2019-04-03
perf: deserialize only needed in dependency cmd

superyyrrzz:
`docfx dependency` for azure-docs-pr:
* before: ~12s
* after: ~8.5s


src/docfx/SubCommands/DependencyCommand.cs
```diff
@@ -31,34 +31,46 @@ public void Exec(SubCommandRunningContext context)
...
+        private DependencyGraph Load(string intermediateFolder, string versionName)
+        {
+            var expandedBaseFolder = Path.GetFullPath(Environment.ExpandEnvironmentVariables(intermediateFolder));
+            var buildInfoFile = Path.Combine(expandedBaseFolder, BuildInfo.FileName);
+            var buildInfo = JsonUtility.Deserialize<BuildInfo>(buildInfoFile);
+            if (buildInfo == null)
+            {
+                LogErrorAndThrow($"Cache files in the folder {intermediateFolder} are corrupted!", null);
+            }
+            var versionInfo = buildInfo.Versions.FirstOrDefault(v => v.VersionName == versionName);
```
herohua: is there an opportunity to pass in all versions and run the export just once?

superyyrrzz: Agree there is no need to run multiple times. Will sync with @fenxu confirming the interface.

PR #4303, opened 2019-04-03
Split contract resolvers from json utility

yufeih:
Refactor only.

### DIFF
--- a/src/docfx/lib/json/JsonUtility.cs
+++ b/src/docfx/lib/json/JsonUtility.cs
@@ -2,12 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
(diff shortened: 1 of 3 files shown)


src/docfx/lib/json/JsonUtility.cs
```diff
@@ -519,129 +510,7 @@ private static JsonPropertyCollection GetPropertiesFromJsonArrayContract(JsonArr
...
-                    }
-                }
-
-                var transform = t_status.Value.Peek().Transform;
-                return transform != null ? transform(_attributes, value, reader.Path) : value;
-            }
-        }
-
-        private sealed class Status
+        internal class Status
```
live1206: does styleCop allow two internal class in the same file?

PR #4306, opened 2019-04-03
Improve backslash replace in getOverrideFolder.

scovetta:
The current `getOverrideFolder` function looks like it has a minor bug, where `.replace('\\', '/')` only replaces the first occurrence of a backslash with a forward slash. From the context, I'm guessing that the intent is to replace them all, so a regex-based match seems like a reasonable approach.


superyyrrzz:
@scovetta Definitely a correct fix. Seems it is challenging to write C# code and JS code simultaneously... Thank you for contribution!

superyyrrzz:
Just find your pull request targets `master` branch, while it should be `dev` branch. I have picked up in `dev`.

PR #4311, opened 2019-04-04
Range --> SourceInfo

yufeih:
This is a refactor only PR to merge `Range` into `SourceInfo`.


src/docfx/lib/log/SourceInfo{T}.cs
```diff
@@ -0,0 +1,37 @@
...
+
+        public SourceInfo<T> WithValue(T value)
+        {
+            if (value == default)
+                return null;
+
+            return new SourceInfo<T>(value, this);
+        }
+
+        public override string ToString()
```
yufeih: who is relying on this `ToString()` overload?

src/docfx/lib/log/SourceInfo{T}.cs
```diff
@@ -0,0 +1,37 @@
...
+        public SourceInfo(T value, SourceInfo source = null)
+            : base(source?.File, source?.Line ?? 0, source?.Column ?? 0, source?.EndLine ?? 0, source?.EndColumn ?? 0)
+        {
+            Value = value;
+        }
+
+        public SourceInfo<T> WithValue(T value)
+        {
+            if (value == default)
+                return null;
```
yufeih: line info will be lost if you return `null` here. If `Value` needs permutation, we can make `Value` settable.

src/docfx/lib/log/SourceInfo{T}.cs
live1206: https://github.com/dotnet/docfx/blob/575c0ad2b242ac79947270f68cedd3e057ae33b0/src/docfx/Errors.cs#L198

src/docfx/lib/log/SourceInfo{T}.cs
live1206: yes, it was not easy to make it settable in the previous PR, will update it here

src/docfx/lib/log/SourceInfo{T}.cs
yufeih: ok, then `DebuggerDisplay` is not needed given the `ToString()` overload. And the `ToString` implementation needs to check `null`

live1206: ok

src/docfx/lib/log/SourceInfo{T}.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        internal override object GetValue() => Value;
+
+        public SourceInfo(T value, SourceInfo source = null)
+            : base(source?.File, source?.Line ?? 0, source?.Column ?? 0, source?.EndLine ?? 0, source?.EndColumn ?? 0)
+        {
+            Value = value;
+        }
+
+        public SourceInfo<T> WithValue(T value)
```
OsmondJiang: we don't need this anymore?

src/docfx/lib/log/SourceInfo.cs
```diff
@@ -1,42 +1,53 @@
...
         public readonly string File;
 
-        public readonly Range Range;
+        /// <summary>
+        /// A one based start line value.
+        /// </summary>
+        public readonly int Line;
 
-        object ISourceInfo.Value => Value;
+        /// <summary>
```
OsmondJiang: since "endLine" and "endColumn" are not used anyway, just delete them?

src/docfx/lib/log/SourceInfo{T}.cs
live1206: forgot to remove

src/docfx/lib/log/SourceInfo.cs
live1206: https://github.com/Microsoft/vscode/blob/master/src/vs/editor/common/core/range.ts
Even vscode define `range` with `endLine` and `endColumn`
We only have them from YamlNode https://github.com/dotnet/docfx/blob/575c0ad2b242ac79947270f68cedd3e057ae33b0/src/docfx/lib/json/YamlUtility.cs#L178
And we are not exposing them anyway, so they can be removed? @yufeih

src/docfx/Errors.cs
```diff
@@ -99,14 +99,14 @@ public static Error InvalidTopicHref(Document relativeTo, string topicHref)
         /// <summary>
         /// In markdown-format toc, link(treated as inclusion) CAN ONLY be toc file, folder or absolute path.
         /// </summary>
-        public static Error InvalidTocHref(Document relativeTo, string tocHref, in Range range)
-            => new Error(ErrorLevel.Error, "invalid-toc-href", $"The toc href '{tocHref}' can only reference to a local TOC file, folder or absolute path", relativeTo.ToString(), range);
+        public static Error InvalidTocHref(Document relativeTo, SourceInfo<string> source)
+            => new Error(ErrorLevel.Error, "invalid-toc-href", $"The toc href '{source}' can only reference to a local TOC file, folder or absolute path", relativeTo.ToString(), source);
```
OsmondJiang: this make me confused wht "file" and "source.file"

src/docfx/lib/log/SourceInfo.cs
yufeih: We have end line|column info for markdown nodes as well. Can add `EndLine, EndColumn` to `Error` to expose it.

src/docfx/Errors.cs
yufeih: @live1206 , I feel like we don't want to merge the two into once parameter. This also solves the need for `ToString` overload

live1206: We will eventually use `File` from `SourceInfo` after we pass file into JsonUtility.Parse and fill `File` correctly into `SourceInfo`

src/docfx/lib/log/SourceInfo.cs
OsmondJiang: that;s what i mean, seems current `EndLine and EndColumn` are removed from `Error`

src/docfx/Errors.cs
live1206: Added TODO

live1206: > @live1206 , I also feel like we don't want to merge the two parameters into 1. This also solves the need for `ToString` overload

I think these two are just fit for `SourceInfo`, it should be fine to have an override `ToString`

src/docfx/lib/log/SourceInfo.cs
live1206: added end line|column back to `Error.cs`

PR #4314, opened 2019-04-04
feat: resolve xref recursively

superyyrrzz:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/78412/


superyyrrzz:
@live1206 Does v3 already support this?

herohua:
so it is not possible to have `summary: something could resolve to another <xref:uid?displayProperties=summary>`?

superyyrrzz:
@herohua docfx will not abandon this usage, but this will not be a valid usage, as the referenced summary will be HTMLEncoded, which will make the `<xref>` in it will not be resolved again.

Issue #4172, opened 2019-03-15
onboard localization repo to docfx v3


OsmondJiang:
- [x] integration localization pipeline with build service 
- [x] add e2e test for localization repo build for v3
- [x] no overwrite config, using existing config
- [x] build from localization repo directly
- [x] bilingual configuration migration @mingwli 
- [x] localization build perf improvement
- [x] resolve deleted referenced resource link
- [x] remove "_repo.en-us" dependency repo from migrated `docfx.yml` @mingwli

OsmondJiang:
cc @herohua  @fenxu  @yufeih

OsmondJiang:
meet some TOC issue during test, the toc file in loc repo has some external field like `ms.contentlocale`, which is unknown for docfx.

![image](https://user-images.githubusercontent.com/19990166/55719915-00374e80-5a32-11e9-813c-ff0cb054d940.png)

![image](https://user-images.githubusercontent.com/19990166/55719923-062d2f80-5a32-11e9-9a3c-0cbe1df13a33.png)

OsmondJiang:
# Build from localized reposotory directly

## Configuration Migration

Configuration migration will migrate configuration in loc repo to `docfx.yml`, the only specified feature for localized repo is to add `bilingual` config to migrated `docfx.yml`.

The current strategy is that configuration migrator will add `bilingual: true` to migrated `docfx.yml` is current building branch end with `-sxs`. @mingwli 

## DocFX Store

```
docfx restore {localized-repo-path}
```
No need add `--locale {locale}` to above execution command, docfx will automatically detect the `locale` from current building repository remote, the localized repository name must follow bellow pattern:
```
{source-repo-name}.{locale}
```

Restore will restore:
  - Dependency repositories defined in `docfx.yml` (dep == 1)
  - Dependency files defined in `docfx.yml`
  - Localized theme repository (dep == 1)
  - Source repository (full history)
  - `contribution branch` if current branch is bilingual branch

## DocFX Build

```
docfx build {localized-repo-path}
```
No need add `--locale {locale}` to above execution command, docfx will automatically detect the `locale` from current building repository remote, the localized repository name must follow bellow pattern:
```
{source-repo-name}.{locale}
```

Docfx will load `docfx.yml` from current `{localized-repo-path}`,  and build:
  - localized files defined build scope
  - localized TOC files and their source parent TOC files if has
  - source xref file map(TODO, @live1206 )
  - localized + source token/code-snippet files if referenced
[...]

PR #4327, opened 2019-04-09
support toc with extension data

OsmondJiang:
only deserialize but serialize, which is the same behavior like v2.


OsmondJiang:
#4172

PR #4336, opened 2019-04-09
📝Xref line info

live1206:
- Add line info for xref related errors
- Add test for TOC bookmark validation

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1192,7 +1192,7 @@ outputs:
       ]
     }
   build.log: |
-    ["warning","uid-not-found","Cannot find uid 'c' using xref 'c'","docs/TOC.yml"]
+    ["warning","uid-not-found","Cannot find uid 'c' using xref 'c'","docs/TOC.yml",5,8]
 ---
 # uid reference with monikers in yaml toc
 inputs:
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1106,21 +1106,25 @@ outputs:
   build.log: |
     ["error","output-path-conflict","Two or more files output to the same path 'docs/toc.json': 'docs/TOC.md', 'docs/TOC.yml'"]
 ---
-# validate bookmarks in toc	
-inputs:	
-  docfx.yml:	
-  docs/a.md: |	
-    # title 1	
-  docs/TOC.md: |	
-    # [link to title 1](a.md#title-1)	
-outputs:	
-  docs/a.json:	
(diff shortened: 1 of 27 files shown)


docs/specs/toc.yml
```diff
@@ -1106,21 +1106,25 @@ outputs:
...
-  docs/TOC.md: |	
-    # [link to title 1](a.md#title-1)	
-outputs:	
-  docs/a.json:	
-  docs/toc.json: |	
-    {	
-      "items": [	
-        { "name": "link to title 1", "href": "a#title-1", "!monikers":null },	
-      ]	
+# validate bookmarks in toc
```
OsmondJiang: would you split this to another PR?

live1206: Just added this test case back, the implementation was done in the previous PR

src/docfx/build/page/AttributeTransformer.cs
```diff
@@ -12,44 +12,43 @@ namespace Microsoft.Docs.Build
...
+            object Transform(IEnumerable<DataTypeAttribute> attributes, object value, string jsonPath, Range range)
             {
                 var attribute = attributes.SingleOrDefault(attr => !(attr is XrefPropertyAttribute));
-                return TransformContent(context, attribute, value, file, buildChild);
+                return TransformContent(context, attribute, value, range, file, buildChild);
             }
         }
 
-        public static Func<IEnumerable<DataTypeAttribute>, object, string, object> TransformXref(
+        public static Func<IEnumerable<DataTypeAttribute>, object, string, Range, object> TransformXref(
```
OsmondJiang: Suggest to:
```
 Func<IEnumerable<DataTypeAttribute>, SourceInfo<object>, string, object>
```

OsmondJiang: and finally we can get ride of range, haha

src/docfx/lib/log/SourceInfo.cs
```diff
@@ -16,7 +16,7 @@ public sealed class SourceInfo<T> : ISourceInfo
 
         object ISourceInfo.Value => Value;
 
-        public SourceInfo(T value, string file, Range range)
+        public SourceInfo(T value, Range range = default, string file = null)
```
superyyrrzz: As `SourceInfo` has `file` already, seems we don't need `Document file` as a parameter in most methods of `errors.cs`.

live1206: will clean this up after we merge #4311

PR #4338, opened 2019-04-09
👨‍🏭support link to deleted resource to get ride of tons of noisy warnings

OsmondJiang:
actually that's a missing feature in v2, with this feature, we can get ride of lots of noisy warning about "file-not-found":

https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C4%5C9%5C4709ceea-246d-76fd-3615-65fb72516268%5CCommit%5C201904090844459536-docfx-v3-e2e-test-sxs%5Cworkflow_report.html?sv=2015-02-21&sr=b&sig=ML%2BFm4dTXklAc%2FNO%2FiORKhJ8qoVXTnPB964iiP6bAV4%3D&st=2019-04-09T09%3A08%3A27Z&se=2019-05-10T09%3A13%3A27Z&sp=r

### DIFF
--- a/src/docfx/build/dependency/DependencyResolver.cs
+++ b/src/docfx/build/dependency/DependencyResolver.cs
@@ -24,6 +24,9 @@ namespace Microsoft.Docs.Build
             _dependencyMapBuilder = dependencyMapBuilder;
             _gitCommitProvider = gitCommitProvider;
             _xrefMap = xrefMap;
+
+            // forLandingPage should not be used, it is a hack to handle some specific logic for landing page based on the user input for now
+            // which needs to be removed once the user input is correct
             _forLandingPage = forLandingPage;
         }
 
@@ -36,8 +39,6 @@ namespace Microsoft.Docs.Build
             return (error, content, child);
         }
 
-        // forLandingPage should not be used, it is a hack to handle some specific logic for landing page based on the user input for now
-        // which needs to be removed once the user input is correct
         public (Error error, string link, Document file) ResolveLink(string path, Document relativeTo, Document resultRelativeTo, Action<Document> buildChild, in Range range)
         {
             var (error, link, fragment, hrefType, file) = TryResolveHref(relativeTo, path, resultRelativeTo, range);
@@ -127,7 +128,7 @@ namespace Microsoft.Docs.Build
             }
 
             var decodedHref = Uri.UnescapeDataString(href);
-            var (error, file, redirectTo, query, fragment, hrefType, _) = TryResolveFile(relativeTo, decodedHref, range);
+            var (error, file, redirectTo, query, fragment, hrefType, pathToDocset) = TryResolveFile(relativeTo, decodedHref, range);
 
             // Redirection
             // follow redirections
@@ -145,7 +146,20 @@ namespace Microsoft.Docs.Build
             // Cannot resolve the file, leave href as is
             if (file is null)
             {
-                return (error, href, fragment, hrefType, null);
+                if (string.IsNullOrEmpty(pathToDocset))
+                {
+                    return (error, href, fragment, hrefType, null);
+                }
+
+                var (errorFromHistory, resourceFromHistory) = TryResolveResourceFromHistory(_gitCommitProvider, relativeTo.Docset, pathToDocset);
+                if (errorFromHistory != null || resourceFromHistory == null)
+                {
+                    return (errorFromHistory ?? error, href, fragment, hrefType, null);
+                }
+
+                // set file to resource got from histroy, reset the error
+                file = resourceFromHistory;
+                error = null;
             }
 
             // Self reference, don't build the file, leave href as is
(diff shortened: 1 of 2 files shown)


herohua:
Would this bring in unexpected performance cost?

OsmondJiang:
> Would this bring in unexpected performance cost?

That only happens in loc repo building and happened on deleted resource links, with git commit history cache, I believe it's OK.

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -145,7 +146,20 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
             // Cannot resolve the file, leave href as is
             if (file is null)
             {
-                return (error, href, fragment, hrefType, null);
+                if (string.IsNullOrEmpty(pathToDocset))
```
live1206: how about putting this check into `TryResolveResourceFromHistory`, so that it can return null `resourceFromHistory`

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -145,7 +146,20 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
...
+
+                var (errorFromHistory, resourceFromHistory) = TryResolveResourceFromHistory(_gitCommitProvider, relativeTo.Docset, pathToDocset);
+                if (errorFromHistory != null || resourceFromHistory == null)
+                {
+                    return (errorFromHistory ?? error, href, fragment, hrefType, null);
+                }
+
+                // set file to resource got from histroy, reset the error
+                file = resourceFromHistory;
+                error = null;
```
live1206: why do we need to reset error here?

src/docfx/build/dependency/DependencyResolver.cs
OsmondJiang: I am trying to keep it consistent with TryResolveContentFromHistory

src/docfx/build/dependency/DependencyResolver.cs
OsmondJiang: otherwise the previous "file-not-found" error will be returned and reported

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -145,7 +146,20 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
...
-                return (error, href, fragment, hrefType, null);
+                if (string.IsNullOrEmpty(pathToDocset))
+                {
+                    return (error, href, fragment, hrefType, null);
+                }
+
+                var (errorFromHistory, resourceFromHistory) = TryResolveResourceFromHistory(_gitCommitProvider, relativeTo.Docset, pathToDocset);
+                if (errorFromHistory != null || resourceFromHistory == null)
+                {
+                    return (errorFromHistory ?? error, href, fragment, hrefType, null);
```
mingwli: >errorFromHistory ?? error [](start = 28, length = 25)

if `errorFromHistory != null`, then the previous 'file-not-found' error will be overwritten?

OsmondJiang: yes, then we want to throw the error why we can't get it from history, it a root cause why file is not found

src/docfx/build/dependency/DependencyResolver.cs
superyyrrzz: what about reuse them? `(error, file) = TryResolveResourceFromHistory(_gitCommitProvider, relativeTo.Docset, pathToDocset);`

src/docfx/build/dependency/DependencyResolver.cs
live1206: do not quite get the consistency here

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -145,7 +146,16 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
             // Cannot resolve the file, leave href as is
             if (file is null)
             {
-                return (error, href, fragment, hrefType, null);
+                if (string.IsNullOrEmpty(pathToDocset))
+                {
+                    return (error, href, fragment, hrefType, null);
+                }
+
+                (error, file) = TryResolveResourceFromHistory(_gitCommitProvider, relativeTo.Docset, pathToDocset);
```
yufeih: consider conslidating `TryResolveResourceFromHistory` and `TryResolveContentFromHistory` logic inside `TryResolveFile`. Logically, `ResolveContent` and `ResolveLink` boths build on top of `ResolveFile`, `ResolveContent` just takes the result of `ResolveFile` and read its content, `ResolveLink` just takes the result of `ResolveFile` and calculate url.

OsmondJiang: the result `content` is the key point, the file resolved from history can't read `content` directly from `document`, so that's the reason why we put it outside of resolveFile. And for perf consideration, we don't want to read content for deleted resource.

src/docfx/build/dependency/DependencyResolver.cs
OsmondJiang: error can't be reused, when file is not resolved from history, error maybe null.

OsmondJiang:
we reduced 600+ warnings:

https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C4%5C10%5C4709ceea-246d-76fd-3615-65fb72516268%5CCommit%5C201904100729008254-v3-loc-sxs%5Cworkflow_report.html?sv=2015-02-21&sr=b&sig=goYZtFuErQFj6grE%2FdqEmSpbNAsL6bPbjmMwawI8Uwk%3D&st=2019-04-10T07%3A52%3A46Z&se=2019-05-11T07%3A57%3A46Z&sp=r

cc @yufeih  @herohua  @fenxu

PR #4344, opened 2019-04-10
🎫Clean up legacy tests

live1206:
- make empty output files optional for legacy tests
- add null check for `JsonUtility.Merge`


docs/specs/legacy.yml
```diff
@@ -64,10 +64,6 @@ outputs:
     { "locale": "en-us", "ms.author": "yufeih", "is_dynamic_rendering": true }
   toc.json:
   toc.mta.json:
-  .manifest.json:
```
OsmondJiang: keep at least one here

PR #4351, opened 2019-04-11
🍂support multiple hashtags for fragment

OsmondJiang:
fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/79102?src=WorkItemMention&src-action=artifact_link

### DIFF
--- a/src/docfx/lib/HrefUtility.cs
+++ b/src/docfx/lib/HrefUtility.cs
@@ -50,7 +50,25 @@ namespace Microsoft.Docs.Build
                 }
             }
 
-            return (path, query, fragment);
+            return (path, query, TrimStart(fragment));
+
+            // TODO: this is to support a legacy behavior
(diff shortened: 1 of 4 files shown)


src/docfx/lib/HrefUtility.cs
```diff
@@ -50,7 +50,25 @@ public static (string path, string query, string fragment) SplitHref(string href
...
+                var lastFragmentIndex = -1;
+                foreach (var ch in fragmentToTrim)
+                {
+                    if (ch == '#')
+                    {
+                        lastFragmentIndex++;
+                        continue;
+                    }
+
+                    break;
```
superyyrrzz: minor: seems a bit clearer if reverse `if`:
```cs
if (ch != '#') break;
lastFragmentIndex++
```

OsmondJiang: I think both are OK

PR #4358, opened 2019-04-15
Add line info to invalid-redirect-to and invalid-redirection

yufeih:


docs/specs/casing.yml
```diff
@@ -29,7 +29,7 @@ inputs:
     [redirect](REDIRECT.md)
 outputs:
   docs/a.json: |
-    {"content":"<p><a href=\"/path\">redirect</a></p>"}
+    {"content":"<p><a href=\"redirect\">redirect</a></p>"}
```
yufeih: I removed a non-legacy behavior to simplify our test metric.

OsmondJiang: is it worth to change our prod behavior just for simplifing our test metric?

yufeih: > is it worth to change our prod behavior just for simplifing our test metric?

Our prod is using `legacy` behavior, the one on the right side.

src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -79,31 +73,25 @@ void AddRedirections(Dictionary<string, string> items, bool checkRedirectTo = fa
                         continue;
                     }
 
-                    if (checkRedirectTo && !redirectTo.StartsWith('/'))
+                    if (checkRedirectTo && !redirectTo.Value.StartsWith('/'))
```
live1206: could `redirectTo` be null?

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -188,9 +175,9 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
             // Make result relative to `resultRelativeTo`
             var relativeUrl = GetRelativeUrl(resultRelativeTo, file);
 
-            if (redirectTo != null)
+            if (file?.RedirectionUrl != null)
             {
-                return (error, relativeUrl + query + fragment, fragment, hrefType, null);
+                return (error, relativeUrl + query + fragment, null, hrefType, null);
```
live1206: why is `fragment` returning null here?

yufeih: to suppress bookmark validation

src/docfx/build/redirection/RedirectionMap.cs
yufeih: I think it shouldn't be `null`, otherwise that'll complicate everything (see existing references to `SourceInfo<T>.Value`), I added a TODO to eliminate null problems associated with `SourceInfo`

PR #4359, opened 2019-04-15
Add line info to merge conflict

yufeih:


src/docfx/lib/git/GitUtility.cs
```diff
@@ -222,7 +222,16 @@ public static void CheckMergeConflictMarker(string content, string file)
                 content.Contains("\n>>>>>>>") &&
                 content.Contains("\n======="))
             {
-                throw Errors.MergeConflict(file).ToException();
+                var start = content.IndexOf("<<<<<<<");
```
OsmondJiang: can we do some optimization here? seems one content will be looped checked multiple times.

Issue #4357, opened 2019-04-15
docfx restore failed in a localized repo test branch(no corresponding source branch)


OsmondJiang:
https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C4%5C11%5C4709ceea-246d-76fd-3615-65fb72516268%5CCommit%5C201904110821274992-OsmondJiang-patch-1%5Crawlog.txt?sv=2015-02-21&sr=b&sig=wdjasXhkCWUr605weDuGk5S4z38vWLKzJWXjQh2R9%2Bo%3D&st=2019-04-11T08%3A24%3A16Z&se=2019-05-12T08%3A29%3A16Z&sp=r

OsmondJiang:
The root cause is that, if there is no corresponding source branch for fallback, we should not failed, there are two options:
1. fallback to master
2. no fallback repo to be restored

Considering the [design spec](https://github.com/dotnet/docfx/blob/v3/docs/designs/conceptual-localization.md#fallback-to-corresponding-branch-only), we are going to option 2.

cc @yufeih  @herohua  @fenxu

PR #4360, opened 2019-04-15
🌐always restore source master for loc non-live branch building

OsmondJiang:
#4357

### DIFF
--- a/src/docfx/localization/LocalizationUtility.cs
+++ b/src/docfx/localization/LocalizationUtility.cs
@@ -130,6 +130,12 @@ namespace Microsoft.Docs.Build
                 {
                     sourceBranch = contributionBranch;
                 }
+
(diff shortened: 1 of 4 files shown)


src/docfx/localization/LocalizationUtility.cs
```diff
@@ -130,6 +130,12 @@ public static bool TryGetSourceRepository(string remote, string branch, out stri
                 {
                     sourceBranch = contributionBranch;
                 }
+
+                if (sourceBranch != "live")
```
yufeih: if it is designed as always fallback to `master`, then we could remove this `live` hack because live will first attempt to fallback to `live` first then `master`?

OsmondJiang: non-live branch fallback to "master", live always fallback to "live"

PR #4362, opened 2019-04-15
Refine raw model for template part-1

OsmondJiang:
1. reuse page model to generate raw model as many as possible
2. add test to check generated raw model, in case we will break in the future changing, will remove it later once our page model become raw model
3. fix some bugs

### DIFF
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -71,6 +71,67 @@ outputs:
   .publish.json:
   xrefmap.json:
 ---
+# Raw model generated for template
+repos:
+  https://docs.com/rawModel:
+  - files:
+      docfx.yml: |
+        theme: https://docs.com/template
+        redirections:
+          a.md: /b
+      c.md: |
+        ---
+        api_name:
+          - "API Name"
+        ms.author: yufeih
+        ---
+        # Title
+        [d](d.png)
+      d.png:
+      TOC.md: |
+        # [c](c.md)
+  https://docs.com/template:
+  - files:
+      readme.md:
+outputs:
+  c.json:
+  d.png:
+  toc.json:
+  .publish.json: |
+    {"files": [
+      {"url": "/a"}, 
+      {
+        "url": "/c",
+        "locale": "en-us",
+        "api_name": ["API Name"],
+        "ms.author": "yufeih",
+        "depot_name": ".",
+        "site_name": "Docs",
+        "__global": {},
+        "conceptual": "<p><a href=\"d.png\">d</a></p>\n",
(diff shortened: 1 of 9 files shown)


docs/specs/legacy.yml
```diff
@@ -71,6 +71,67 @@ outputs:
...
+        "api_name": ["API Name"],
+        "ms.author": "yufeih",
+        "depot_name": ".",
+        "site_name": "Docs",
+        "__global": {},
+        "conceptual": "<p><a href=\"d.png\">d</a></p>\n",
+        "_path": "c.json",
+        "toc_rel": "toc.json",
+        "word_count": 2,
+        "wordCount": 2,
```
OsmondJiang: two word_count and two locale here

docs/specs/legacy.yml
```diff
@@ -71,6 +71,67 @@ outputs:
...
+        "title": "Title",
+        "rawTitle": "<h1 id=\"title\">Title</h1>",
+        "_op_canonicalUrlPrefix": "https://docs.com/en-us/./",
```
OsmondJiang: this is somehow strange

src/docfx/template/TemplateEngine.cs
```diff
@@ -301,7 +295,7 @@ private static JObject CreateMetadata(JObject rawMetadata)
 
         private static string CreateHtmlMetaTags(JObject metadata)
         {
-            var result = new StringBuilder();
+            var meta = new List<string>();
 
             foreach (var (key, value) in metadata)
```
yufeih: order here

src/docfx/docset/Document.cs
```diff
@@ -471,7 +471,8 @@ private static bool IsValidRelativePath(string path)
             // get set path from site path
             // site path doesn't contain version info according to the output spec
             var sitePathWithoutExtension = Path.Combine(Path.GetDirectoryName(SitePath), Path.GetFileNameWithoutExtension(SitePath));
-            var sitePath = PathUtility.NormalizeFile(Path.GetRelativePath(Docset.Config.DocumentId.SiteBasePath, sitePathWithoutExtension));
+            var siteBasePath = Docset.Config.Output.LowerCaseUrl ? Docset.Config.DocumentId.SiteBasePath.ToLowerInvariant() : Docset.Config.DocumentId.SiteBasePath;
```
yufeih: Is there a bug for this one? suggest not change this in this PR.

OsmondJiang: this bug is found in my test, so i need fix it to make my test pass in linux.

PR #4363, opened 2019-04-15
should not check conflict id if the redirection file is out of scope

928PJY:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/84112/

### DIFF
--- a/src/docfx/docset/Docset.cs
+++ b/src/docfx/docset/Docset.cs
@@ -264,9 +264,7 @@ namespace Microsoft.Docs.Build
                         }
                     });
 
-                var result = new HashSet<Document>(files);
-                result.UnionWith(redirections);
-                return result;
+                return new HashSet<Document>(files.Concat(redirections));
             }
         }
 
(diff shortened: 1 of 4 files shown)


928PJY:
@yufeih offline discussed with @herohua, we agree to fix it in docfx v3, any different idea?

src/docfx/docset/Docset.cs
```diff
@@ -168,7 +168,12 @@ internal class Docset
...
-            _buildScope = new Lazy<HashSet<Document>>(() => CreateBuildScope(Redirections.Files));
+            _buildScope = new Lazy<HashSet<Document>>(() =>
+            {
+                var (errors, scope) = CreateBuildScope(Redirections.Files);
+                report.Write(Config.ConfigFileName, errors);
+                return scope;
+            });
             _redirections = new Lazy<RedirectionMap>(() =>
             {
                 var (errors, map) = RedirectionMap.Create(this);
```
yufeih: consider passing `glob` to `RedirectionMap.Create` and handle redirection related errors in `RedirectionMap.Create`

src/docfx/docset/Docset.cs
```diff
@@ -265,19 +265,7 @@ private HashSet<Document> CreateBuildScope(IEnumerable<Document> redirections)
                     });
 
                 var result = new HashSet<Document>(files);
```
yufeih:
```suggestion
return new HashSet<Document>(files.Concat(redirections));
```

PR #4364, opened 2019-04-15
Revert "🍂support multiple hashtags for fragment (#4351)"

superyyrrzz:
This reverts commit 88e49072ab6c4b31f785049e7aef09e594a89514.

If user want to link to a bookmark like `<h1 id="#test"></h1>`, it should use `<a href="##test"></a>`. Merging mutiple `#` into one is wrong in this case. This is a valid case as [HTML5 allows any character except any type of space character](http://xahlee.info/js/html_allowed_chars_in_attribute.html)

So this diff is actually a [v2 bug](https://github.com/dotnet/docfx/blob/f1ff9c21dd5e7c9cd21185bca3693d9e460f1263/src/Microsoft.DocAsCode.Build.Engine/PostProcessors/ValidateBookmark.cs#L85), we can fisrt fix it in diff tool.
https://ceapex.visualstudio.com/Engineering/_workitems/edit/79102?src=WorkItemMention&src-action=artifact_link

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -339,14 +339,12 @@ inputs:
     [link to title 3](#title-3)
     [link to html title](#html-title)
     [link to yaml](c.yml#no-such-bookmark) does not warn, only validate against markdown files
-    [link to title 2 again](####title-2)
     ## title 2
     <h2 id="html-title">title</h2>
     [!include[](d.md)]
   docs/b.md: |
     [link to title 2](a.md#title-2)
     [link to title 3](a.md#title-3)
-    [link to title 2 again](a.md##title-2)
   docs/c.yml: |
     #YamlMime:LandingData
     title: test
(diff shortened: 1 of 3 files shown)


OsmondJiang:
let's firstly check if this change is valid: https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/commit/3a4ab681a2c8047cea16dd5b7bc1bc0d0dcaa16d?refName=refs%2Fheads%2Fmaster

OsmondJiang:
offline talked with @herohua  and @superyyrrzz , we should fix the bug in v2 eventually

OsmondJiang:
or add a test for test the case you mentioned.

superyyrrzz:
As the fix in v2 will introduce extra warnings, will contact with PM to decide whether to fix it in v2 or when migrating to v3.

PR #4365, opened 2019-04-15
Add support for <xref href="uid"/> in HTML inside markdown files

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/77048

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -661,7 +661,6 @@ inputs:
   docs/a.md: |
     str <xref href="c"></xref>
     str <xref href="d" data-raw-source="@mention"></xref>
-    str <xref href="d" data-throw-if-not-resolved="False" data-raw-html="@higher" data-raw-source="@mention"></xref>
     some text
     str <xref href="d" data-raw-html="@higher" data-raw-source="@mention"></xref>
     multiple xref <xref href="e" data-raw-source="@mentione"></xref> some text <xref href="f" data-raw-source="@mentionf"></xref>
@@ -677,12 +676,12 @@ inputs:
     ---
 outputs:
   docs/a.json: |
-    {"content" : "<p>str <a href='c'>Title from yaml header c</a>\nstr @mention\nstr @higher\nsome text\nstr @higher\nmultiple xref @mentione some text @mentionf</p>\n<p>\n  @mentione\n  some text\n  mentionf\n</p>\n"}
+    {"content" : "<p>str <a href='c'>Title from yaml header c</a>\nstr @mention\nsome text\nstr @higher\nmultiple xref @mentione some text @mentionf</p>\n<p>\n  @mentione\n  some text\n  mentionf\n</p>\n"}
   docs/c.json:
   build.log: |
     ["info","at-uid-not-found","Cannot find uid 'd' using xref '<xref href=\"d\" data-raw-source=\"@mention\">'","docs/a.md",2,12]
-    ["info","at-uid-not-found","Cannot find uid 'd' using xref '<xref href=\"d\" data-raw-html=\"@higher\" data-raw-source=\"@mention\">'","docs/a.md",5,12]
-    ["info","at-uid-not-found","Cannot find uid 'e' using xref '<xref href=\"e\" data-raw-source=\"@mentione\">'","docs/a.md",6,12]
-    ["info","at-uid-not-found","Cannot find uid 'f' using xref '<xref href=\"f\" data-raw-source=\"@mentionf\">'","docs/a.md",6,12]
-    ["info","at-uid-not-found","Cannot find uid 'e' using xref '<p>\n  <xref href=\"e\" data-raw-source=\"@mentione\" />\n  some text\n  <xref href=\"f\" data-raw-source=\"mentionf\" />\n</p>'","docs/a.md",7,18]
-    ["warning","uid-not-found","Cannot find uid 'f' using xref '<p>\n  <xref href=\"e\" data-raw-source=\"@mentione\" />\n  some text\n  <xref href=\"f\" data-raw-source=\"mentionf\" />\n</p>'","docs/a.md",7,78]
+    ["info","at-uid-not-found","Cannot find uid 'd' using xref '<xref href=\"d\" data-raw-html=\"@higher\" data-raw-source=\"@mention\">'","docs/a.md",4,12]
+    ["info","at-uid-not-found","Cannot find uid 'e' using xref '<xref href=\"e\" data-raw-source=\"@mentione\">'","docs/a.md",5,12]
+    ["info","at-uid-not-found","Cannot find uid 'f' using xref '<xref href=\"f\" data-raw-source=\"@mentionf\">'","docs/a.md",5,12]
+    ["info","at-uid-not-found","Cannot find uid 'e' using xref '<p>\n  <xref href=\"e\" data-raw-source=\"@mentione\" />\n  some text\n  <xref href=\"f\" data-raw-source=\"mentionf\" />\n</p>'","docs/a.md",6,18]
+    ["warning","uid-not-found","Cannot find uid 'f' using xref '<p>\n  <xref href=\"e\" data-raw-source=\"@mentione\" />\n  some text\n  <xref href=\"f\" data-raw-source=\"mentionf\" />\n</p>'","docs/a.md",6,78]
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -654,3 +654,21 @@ inputs:
 outputs:
   build.log: |
     ["error","schema-not-found","Unknown schema ''","docs/a.json"]
+---
+# support <xref href="uid"></xref>
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    str <xref href="c"></xref>
+    str <xref href="d" data-raw-source="mention"></xref>
+  docs/c.md: |
+    ---
+    title: Title from yaml header c
+    uid: c
+    ---
+outputs:
+  docs/a.json: |
+    {"content":"<p>str <a href='c'>Title from yaml header c</a>\nstr @mention</p>\n"}
+  docs/c.json:
+  build.log: |
+    ["warning","uid-not-found","Cannot find uid 'c' using xref '<xref href=\"c\">'","docs/a.md",2,12]
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -662,7 +662,6 @@ inputs:
     str <xref href="c"></xref>
     str <xref href="d" data-raw-source="@mention"></xref>
     str <xref href="d" data-raw-html="@higher" data-raw-source="@mention"></xref>
-    str <xref href="d" />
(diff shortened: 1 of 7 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,52 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+                }
+
+                var rawSource = node.GetAttributeValue("data-raw-source", null);
+                var rawHtml = node.GetAttributeValue("data-raw-html", null);
+                var raw = HttpUtility.HtmlDecode(!string.IsNullOrEmpty(rawHtml) ? rawHtml : rawSource) ?? xref.Value;
+                var (_, resolvedHref, display, _) = transform(xref.Value);
+                if (string.IsNullOrEmpty(resolvedHref))
+                {
+                    errors.Add((xref.Value, currentLine, s_getValueStartIndex(xref)));
+                    result.Append($"@{raw}");
```
superyyrrzz: Remove `@` here. `@` should be within `raw`.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,52 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+                    continue;
+                }
+
+                var rawSource = node.GetAttributeValue("data-raw-source", null);
+                var rawHtml = node.GetAttributeValue("data-raw-html", null);
+                var raw = HttpUtility.HtmlDecode(!string.IsNullOrEmpty(rawHtml) ? rawHtml : rawSource) ?? xref.Value;
+                var (_, resolvedHref, display, _) = transform(xref.Value);
+                if (string.IsNullOrEmpty(resolvedHref))
+                {
+                    errors.Add((xref.Value, currentLine, s_getValueStartIndex(xref)));
```
superyyrrzz: No need to log warning if `data-throw-if-not-resolved="False" `

src/docfx/lib/HtmlUtility.cs
live1206: According to the test, seems not?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -195,15 +195,15 @@ public static (List<(string uid, int line, int column)>, string) TransformXrefs(
                 var logWarning = node.GetAttributeValue("data-throw-if-not-resolved", null);
                 var rawSource = node.GetAttributeValue("data-raw-source", null);
                 var rawHtml = node.GetAttributeValue("data-raw-html", null);
-                var raw = HttpUtility.HtmlDecode(!string.IsNullOrEmpty(rawHtml) ? rawHtml : rawSource) ?? xref.Value;
+                var raw = HttpUtility.HtmlDecode(!string.IsNullOrEmpty(rawHtml) ? rawHtml : rawSource) ?? $"@{xref.Value}";
```
superyyrrzz: We can use innerHtmas last fallback as in v2 `$"@{xref.Value}"` => `node.InnerHtml`

live1206: simply the fallback for now only takes `raw`. fix it later if we need to

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,56 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+                    {
+                        errors.Add((xref.Value, currentLine, s_getValueStartIndex(xref)));
+                    }
+                    result.Append($"{raw}");
+                }
+                else
+                {
+                    result.Append($"<a href='{resolvedHref}'>{display}</a>");
+                }
+                currentLine += 1;
```
superyyrrzz: Is there a more accurate way to calculate line? This may be wrong if `continue` is hit, or multiple `<xref>` in the same like.

live1206: For the `continue` case, you are right, updated the test accordingly.
Processing node by node here, I assume each node should be 1 line.
For multiple `<xref>` in the same line, they should have the same line number

live1206: good point, updated test

docs/specs/xref.yml
```diff
@@ -654,3 +654,31 @@ inputs:
...
+  docs/a.json: |
+    {"content" : "<p>str <a href='c'>Title from yaml header c</a>\nstr @mention\nstr @higher\nsome text\nstr @higher\nmultiple xref @mentione some text @mentionf\nmultiple xref @mentione@mentionf</p>\n"}
+  docs/c.json:
+  build.log: |
+    ["warning","uid-not-found","Cannot find uid 'd' using xref '<xref href=\"d\" data-raw-source=\"@mention\">'","docs/a.md",2,12]
+    ["warning","uid-not-found","Cannot find uid 'd' using xref '<xref href=\"d\" data-raw-html=\"@higher\" data-raw-source=\"@mention\">'","docs/a.md",5,12]
+    ["warning","uid-not-found","Cannot find uid 'e' using xref '<xref href=\"e\" data-raw-source=\"@mentione\">'","docs/a.md",6,12]
+    ["warning","uid-not-found","Cannot find uid 'f' using xref '<xref href=\"f\" data-raw-source=\"@mentionf\">'","docs/a.md",6,12]
+    ["warning","uid-not-found","Cannot find uid 'e' using xref '<xref href=\"e\" data-raw-source=\"@mentione\">'","docs/a.md",7,12]
+    ["warning","uid-not-found","Cannot find uid 'f' using xref '<xref href=\"f\" data-raw-source=\"@mentionf\">'","docs/a.md",7,12]
```
superyyrrzz: column is 12 for both `e` and `f`...

live1206: Yes, do not see a good way to get the proper column number. And also while the HTML block has multiple lines, just use the starting line since there is not an accurate way to get the line number within the block.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,53 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+            var rawHtml = node.GetAttributeValue("data-raw-html", null);
+            var raw = HttpUtility.HtmlDecode(!string.IsNullOrEmpty(rawHtml) ? rawHtml : rawSource);
+            var (_, resolvedHref, display, _) = transform(xref.Value);
+            if (string.IsNullOrEmpty(resolvedHref))
+            {
+                if (string.Compare(logWarning, "False", StringComparison.OrdinalIgnoreCase) != 0)
+                {
+                    errors.Add((xref.Value, lineNumber, s_getValueStartIndex(xref)));
+                }
+                result.Append($"{raw}");
```
yufeih:
```suggestion
                result.Append(raw);
```

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,53 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+            }
+            else
+            {
+                result.Append($"<a href='{resolvedHref}'>{display}</a>");
```
yufeih: `HtmlEncode` and `HtmlDecode`

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,53 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+                return (errors, html);
+            }
+
+            var xref = node.Attributes["href"];
+            if (xref is null)
+            {
+                return (errors, html);
+            }
+
+            var logWarning = node.GetAttributeValue("data-throw-if-not-resolved", null);
```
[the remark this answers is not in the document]
live1206: As I understand, `data-throw-if-not-resolved` is default as `true`. Just follow the same logic with `data-raw-source` and `data-raw-html`

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,63 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+                }
+
+                var xref = node.Attributes["href"];
+                if (xref is null)
+                {
+                    continue;
+                }
+
+                var logWarning = node.GetAttributeValue("data-throw-if-not-resolved", null);
+                var rawSource = node.GetAttributeValue("data-raw-source", null);
```
yufeih: handle `at-uid-not-found` vs `uid-not-found`

src/docfx/lib/HtmlUtility.cs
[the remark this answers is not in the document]
live1206: remove `data-throw-if-not-resolved`

src/docfx/lib/markdown/ResolveXrefExtension.cs
```diff
@@ -21,24 +23,55 @@ internal static class ResolveXrefExtension
...
+                         LogErrors(errors, file, inline.Tag);
+                         inline.Tag = result;
+                     }
                      return node;
                  });
              });
+
+            void LogErrors(List<(string uid, int line, int column, string errorCode)> errors, Document file, string html)
+            {
+                foreach (var (uid, line, column, errorCode) in errors)
```
yufeih: why not make `TransformXref` return a `List<Error>` to avoid doing the transformation here?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -163,6 +164,61 @@ public static string TransformLinks(string html, Func<string, string> transform)
...
+
+                var resolvedNode = new HtmlDocument();
+                if (string.IsNullOrEmpty(resolvedHref))
+                {
+                    errors.Add((xref.Value, lineNumber, s_getValueStartIndex(xref), raw.StartsWith("@") ? nameof(Errors.AtUidNotFound) : nameof(Errors.UidNotFound)));
+                    resolvedNode.LoadHtml(raw);
+                }
+                else
+                {
+                    resolvedNode.LoadHtml($"<a href='{HttpUtility.HtmlEncode(resolvedHref)}'>{display}</a>");
```
yufeih: do you also need to encode `{display}` in case it contains `<`? Also add a test case about encoding handling in `encoding.yml`?

src/docfx/lib/markdown/ResolveXrefExtension.cs
live1206: Because then we need to send in File as input for `TransformXref`, which would cause problems for the tests sending in null file

src/docfx/lib/HtmlUtility.cs
live1206: Actually, I am planning to do this within `Xrefmap.Resolve()`, decode `href` and encode `resolveXref` and `display`, feel it should be in a later PR

src/docfx/lib/markdown/ResolveXrefExtension.cs
yufeih: How about modifying `Errors.UidNotFound` to take `SourceInfo` over `Document` then `TransformXref` could depend on a `string file`?

live1206: sure, sounds like a plan

PR #4368, opened 2019-04-16
Fix NRE in IsDocfxException

yufeih:
Otherwise `IsDocfxException` throws `NRE`

### DIFF
--- a/src/docfx/lib/log/DocfxException.cs
+++ b/src/docfx/lib/log/DocfxException.cs
@@ -21,9 +21,18 @@ namespace Microsoft.Docs.Build
 
         public static bool IsDocfxException(Exception ex, out DocfxException docfxException)
         {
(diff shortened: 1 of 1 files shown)


src/docfx/lib/log/DocfxException.cs
```diff
@@ -21,9 +21,18 @@ public DocfxException(Error error, Exception innerException = null)
...
         {
-            docfxException = ex as DocfxException;
+            while (ex != null)
+            {
+                if (ex is DocfxException de)
+                {
+                    docfxException = de;
+                    return true;
+                }
+                ex = ex.InnerException;
```
OsmondJiang: would we avoid change the input exception?

yufeih: it is a recursive algorithm so I need to change it, do you suggest having a new variable?

PR #4381, opened 2019-04-17
👨‍🏭 move more logic out of `create raw model`

OsmondJiang:
to make out output model more like raw model prepared for template

PR #4382, opened 2019-04-18
🛠Not Serialize empty Sourceinfo<>

live1206:
- Should not serialize `SourceInfo` with `null` value or empty array #4381 
- Consolidate `JsonMerge` tests

### DIFF
--- a/src/docfx/lib/json/JsonContractResolver.cs
+++ b/src/docfx/lib/json/JsonContractResolver.cs
@@ -30,13 +30,21 @@ namespace Microsoft.Docs.Build
         protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
         {
             var prop = base.CreateProperty(member, memberSerialization);
-            ShouldNotSerializeEmptyArray();
+            ShouldNotSerialize();
             SetFieldWritable();
             return prop;
 
-            void ShouldNotSerializeEmptyArray()
+            void ShouldNotSerialize()
             {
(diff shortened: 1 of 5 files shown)


src/docfx/lib/json/JsonContractResolver.cs
```diff
@@ -30,13 +30,21 @@ protected override JsonObjectContract CreateObjectContract(Type objectType)
...
             {
-                if (typeof(IEnumerable).IsAssignableFrom(prop.PropertyType) && !(prop.PropertyType == typeof(string)))
+                // SourceInfo with null value
+                if (prop.PropertyType.IsGenericType && prop.PropertyType.GetGenericTypeDefinition() == typeof(SourceInfo<>))
+                {
+                    prop.ShouldSerialize =
+                        target => ((SourceInfo)prop.ValueProvider.GetValue(target))?.GetValue() != null;
+                }
+
+                // Empty array
```
OsmondJiang: below rule(empty array) also apply to SourceInfo.Value

src/docfx/lib/json/JsonContractResolver.cs
```diff
@@ -30,13 +30,21 @@ protected override JsonObjectContract CreateObjectContract(Type objectType)
         protected override JsonProperty CreateProperty(MemberInfo member, MemberSerialization memberSerialization)
         {
             var prop = base.CreateProperty(member, memberSerialization);
-            ShouldNotSerializeEmptyArray();
+            ShouldNotSerialize();
```
yufeih: Consider having two methods: `DoNotSerializeEmptyArray` and `DoNotSerializeNullSourceInfo`

PR #4383, opened 2019-04-18
🌐 restore corresponding source branch firstly

OsmondJiang:
build loc xxx branch
  - restore source xxx branch
  - if source xxx branch doesn't exist, restore `master` branch

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -143,7 +143,7 @@ namespace Microsoft.Docs.Build
 
                 return Execute(".", $"ls-remote --heads \"{remote}\" {branch}").Split('\n', StringSplitOptions.RemoveEmptyEntries).Any();
             }
-            catch
+            catch (InvalidOperationException)
             {
                 return false;
             }
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -130,6 +130,26 @@ namespace Microsoft.Docs.Build
         }
 
         /// <summary>
+        /// Check if remote branch exists
+        /// </summary>
+        public static bool RemoteBranchExists(string remote, string branch)
+        {
+            try
+            {
+                if (GitRemoteProxy != null)
+                {
+                    remote = GitRemoteProxy(remote);
+                }
+
+                return Execute(".", $"ls-remote --heads \"{remote}\" {branch}").Split('\n', StringSplitOptions.RemoveEmptyEntries).Any();
+            }
+            catch
+            {
+                return false;
+            }
+       }
+
+        /// <summary>
         /// List work trees for a given repo
         /// </summary>
         public static List<string> ListWorkTree(string repoPath)
--- a/src/docfx/localization/LocalizationUtility.cs
+++ b/src/docfx/localization/LocalizationUtility.cs
@@ -207,12 +207,6 @@ namespace Microsoft.Docs.Build
                     sourceBranch = contributionBranch;
                 }
 
-                if (sourceBranch != "master" &&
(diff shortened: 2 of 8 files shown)


OsmondJiang:
#4357

src/docfx/localization/LocalizationUtility.cs
```diff
@@ -131,11 +131,12 @@ public static bool TryGetSourceRepository(string remote, string branch, out stri
                     sourceBranch = contributionBranch;
                 }
 
-                if (sourceBranch != "live")
+                if (sourceBranch != "master" &&
+                    !GitUtility.RemoteBranchExists(sourceRemote, sourceBranch))
```
yufeih: `ls-remote` is a network call, better not to access network in a utility function, this function is used in build as well.

OsmondJiang: docfx make sure the source repo are only get once during restore and build. so it's should be OK

OsmondJiang:
this PR current is waiting for #4386

OsmondJiang:
> this PR current is waiting for #4386

Done

src/docfx/lib/git/GitUtility.cs
```diff
@@ -129,6 +129,26 @@ public static string[] GetCommits(string path, string committish, int top = 1)
...
+            try
+            {
+                if (GitRemoteProxy != null)
+                {
+                    remote = GitRemoteProxy(remote);
+                }
+
+                return Execute(".", $"ls-remote --heads \"{remote}\" {branch}").Split('\n', StringSplitOptions.RemoveEmptyEntries).Any();
+            }
+            catch
```
yufeih: when is there an exception? can it be implemented without using exception handling

test/docfx.Test/build/ConfigTest.cs
```diff
@@ -90,10 +90,6 @@ public static void LocConfigConventionTheme(string theme, string locale, string
         [InlineData("https://test.visualstudio.com/_git/abc.bs-Cyrl-BA", "master", "https://test.visualstudio.com/_git/abc", "master", "bs-cyrl-ba")]
         [InlineData("https://github.com/docs.zh-cn", "master-sxs", "https://github.com/docs", "master", "zh-cn")]
         [InlineData("https://github.com/docs.loc", "master-sxs.zh-cn", "https://github.com/docs", "master", "zh-cn")]
-        [InlineData("https://github.com/docs.zh-cn", "test", "https://github.com/docs", "master", "zh-cn")]
-        [InlineData("https://github.com/docs.loc", "test.zh-cn", "https://github.com/docs", "master", "zh-cn")]
-        [InlineData("https://github.com/docs.zh-cn", "test-sxs", "https://github.com/docs", "master", "zh-cn")]
-        [InlineData("https://github.com/docs.loc", "test-sxs.zh-cn", "https://github.com/docs", "master", "zh-cn")]
         public static void LocConfigConventionSourceRepo(string remote, string branch, string expectedSourceRemote, string expectedSourceBranch, string expectedLocale)
```
yufeih: why remove these tests?

OsmondJiang: fallback logic is not covered by TryGetSourceRepo method, it's controlled by restore/build pipeline using different methods

src/docfx/lib/git/GitUtility.cs
OsmondJiang: when the remote doesn't exists, just return false, I think `the remote doesn't exist` error should be thrown by `restore`

test/docfx.Test/build/ConfigTest.cs
OsmondJiang: we have spec test to cover that case

PR #4386, opened 2019-04-19
🌐make sure the source repo are get only once during restore and build

OsmondJiang:
that's a pre-work for #4383

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -18,12 +18,16 @@ namespace Microsoft.Docs.Build
             var repository = Repository.Create(docsetPath);
             Telemetry.SetRepository(repository?.Remote, repository?.Branch);
 
-            var sourceRepo = LocalizationUtility.TryGetSourceRepository(repository, out var remote, out string branch, out var locale) ? new SourceRepo(remote, branch) : default;
+            var sourceRepoInfo = LocalizationUtility.TryGetSourceRepository(repository, out var remote, out string branch, out var locale) ? (remote, branch) : default;
 
             locale = locale ?? options.Locale;
-            var dependencyLock = LoadBuildDependencyLock(docsetPath, locale, options, sourceRepo);
-
+            var dependencyLock = LoadBuildDependencyLock(docsetPath, locale, options, sourceRepoInfo);
             var restoreMap = RestoreMap.Create(dependencyLock);
+
+            var sourceRepo = sourceRepoInfo != default
+                ? Repository.Create(restoreMap.GetGitRestorePath(sourceRepoInfo.remote, sourceRepoInfo.branch, dependencyLock).path, sourceRepoInfo.branch, sourceRepoInfo.remote)
+                : default;
+
             try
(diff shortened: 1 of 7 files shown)


src/docfx/build/Build.cs
```diff
@@ -266,5 +286,29 @@ private static IReadOnlyList<Document> GetTableOfContentsScope(Docset docset, Ta
 
             return result;
         }
+
+        private class SourceRepo
```
yufeih: we already have a `Repository` class, can that be reused?

OsmondJiang: `Repository` need have "path"

src/docfx/build/Build.cs
```diff
@@ -18,32 +18,44 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Repo
...
-        private static async Task Run(string docsetPath, Repository repository, string locale, CommandLineOptions options, Report report, DependencyLockModel dependencyLock, RestoreMap restoreMap)
+        private static async Task Run(
+            string docsetPath,
+            Repository repository,
+            string locale,
+            CommandLineOptions options,
+            Report report,
+            DependencyLockModel dependencyLock,
+            RestoreMap restoreMap,
+            Repository sourceRepo = null)
```
yufeih: consider name it `fallbackRepo` or `locFallbackRepo`. `sourceRepo` could also mean source/target repo for a pull request.

src/docfx/build/Build.cs
```diff
@@ -197,41 +209,50 @@ void ValidateBookmarks()
             }
         }
 
-        private static DependencyLockModel LoadBuildDependencyLock(string docset, string locale, Repository repository, CommandLineOptions commandLineOptions)
+        private static (DependencyLockModel dependencyLock, (string remote, string branch) sourceRepoInfo) LoadBuildDependencyLock(
```
yufeih:
```suggestion
        private static (DependencyLockModel dependencyLock, string remote, string branch) LoadBuildDependencyLock(
```

src/docfx/localization/LocalizationUtility.cs
```diff
@@ -237,6 +182,42 @@ public static bool TryRemoveLocale(string name, out string nameWithoutLocale, ou
             return false;
         }
 
+        internal static bool TryGetSourceRepository(string remote, string branch, out string sourceRemote, out string sourceBranch, out string locale)
```
yufeih: nit: Still make it public to make diff smaller?

```suggestion
        public static bool TryGetSourceRepository(string remote, string branch, out string sourceRemote, out string sourceBranch, out string locale)
```

src/docfx/build/Build.cs
OsmondJiang: good idea

src/docfx/localization/LocalizationUtility.cs
OsmondJiang: or let me try to make it private

PR #4390, opened 2019-04-19
Comply with v2 behavior, allow additional property for landing page

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/85613

### DIFF
--- a/src/Microsoft.Docs.Template/Models/LandingData.cs
+++ b/src/Microsoft.Docs.Template/Models/LandingData.cs
@@ -7,7 +7,7 @@ using Newtonsoft.Json.Linq;
 namespace Microsoft.Docs.Build
 {
     [PageSchema]
-    public class LandingData
+    public class LandingData : ClassWithExtensionData
     {
         public string Title { get; set; }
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Docs.Template/Models/LandingData.cs
```diff
@@ -20,12 +20,9 @@ public class LandingData
         public LandingDataSection[] Sections { get; set; }
 
         public string DocumentType { get; set; }
-
-        [JsonExtensionData(WriteData = false)]
-        public JObject ExtensionData { get; set; }
     }
 
-    public sealed class LandingDataAbstract
+    public sealed class LandingDataAbstract : ClassWithExtensionData
```
OsmondJiang: what if we just remove sealed?

live1206: We need `ExtensionData` to store additional properties. `sealed` only affects `AdditaionalProperties` of the JSON schema while using `Json.net schema`, which we do not use any more.

live1206:
Fix the content for now since we only have 1 occurrence, will get back to the behavior once PM has reviewed `unknown-field` or we are getting more

PR #4395, opened 2019-04-22
👨‍🏭 combine restore map and dependency lock in build pipeline

OsmondJiang:
these two model are logically one thing in build pipeline

### DIFF
--- a/src/docfx/restore/RestoreMap.cs
+++ b/src/docfx/restore/RestoreMap.cs
@@ -6,7 +6,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Threading.Tasks;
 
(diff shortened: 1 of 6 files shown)


src/docfx/restore/RestoreMap.cs
```diff
@@ -6,15 +6,16 @@
...
-using System.Threading.Tasks;
 
 namespace Microsoft.Docs.Build
 {
     internal class RestoreMap
     {
         private IReadOnlyDictionary<(string remote, string branch, string commit), (string path, DependencyGit git)> _acquiredGits;
 
-        public RestoreMap(IReadOnlyDictionary<(string remote, string branch, string commit), (string path, DependencyGit git)> acquiredGits)
+        public DependencyLockModel DependencyLock { get; private set; }
```
yufeih: can this be private?

PR #4397, opened 2019-04-22
Add line info for more error codes

live1206:


src/docfx/Errors.cs
```diff
@@ -93,8 +93,8 @@ public static Error GitHubApiFailed(string api, Exception ex)
         /// In yaml-format toc, topicHref SHOULD reference an article,
         /// rather than relative path or another toc file.
         /// </summary>
-        public static Error InvalidTopicHref(Document relativeTo, string topicHref)
-            => new Error(ErrorLevel.Error, "invalid-topic-href", $"The topic href '{topicHref}' can only reference to a local file or absolute path", relativeTo.ToString());
+        public static Error InvalidTopicHref(Document relativeTo, SourceInfo<string> source)
```
yufeih: consider make `SourceInfo` the first parameter to make it easier to see which one does not have source info

live1206: refer to the TODOs in `E2ETest.cs`

PR #4398, opened 2019-04-23
use null to overwrite when merge jObjects

mingwli:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/81106/

### DIFF
--- a/src/docfx/build/metadata/MetadataProvider.cs
+++ b/src/docfx/build/metadata/MetadataProvider.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Docs.Build
             if (yamlHeader != null)
             {
                 errors.AddRange(MetadataValidator.ValidateGlobalMetadata(yamlHeader));
-                JsonUtility.Merge(result, yamlHeader);
+                JsonUtility.Merge(result, yamlHeader, overwriteWithNull: true);
             }
 
             // We are validating against the merged JObject so discard the validation result here.
--- a/src/docfx/build/metadata/MetadataProvider.cs
+++ b/src/docfx/build/metadata/MetadataProvider.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Docs.Build
             if (yamlHeader != null)
             {
                 errors.AddRange(MetadataValidator.ValidateGlobalMetadata(yamlHeader));
-                JsonUtility.Merge(result, yamlHeader, overwriteWithNull: true);
+                JsonUtility.Merge(result, yamlHeader);
             }
 
             // We are validating against the merged JObject so discard the validation result here.
--- a/src/docfx/cli/CommandLineOptions.cs
+++ b/src/docfx/cli/CommandLineOptions.cs
@@ -16,10 +16,9 @@ namespace Microsoft.Docs.Build
 
         public JObject ToJObject()
         {
-            var output = new JObject
-            {
-                ["path"] = Output != null ? (JValue)Output : JValue.CreateNull(),
-            };
+            var output = new JObject();
+            if (Output != null)
+                output["path"] = Output;
 
             if (Legacy)
             {
--- a/src/docfx/cli/CommandLineOptions.cs
+++ b/src/docfx/cli/CommandLineOptions.cs
@@ -16,14 +16,20 @@ namespace Microsoft.Docs.Build
 
         public JObject ToJObject()
         {
+            var output = new JObject
+            {
+                ["path"] = Output != null ? (JValue)Output : JValue.CreateNull(),
+            };
+
+            if (Legacy)
+            {
+                output["json"] = true;
+                output["copyResources"] = false;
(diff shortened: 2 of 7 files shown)


src/docfx/lib/json/JsonUtility.cs
```diff
@@ -233,7 +233,7 @@ public static (List<Error>, JToken) Parse(string json, string file)
             }
         }
 
-        public static void Merge(JObject container, JObject overwrite)
+        public static void Merge(JObject container, JObject overwrite, bool overwriteWithNull = false)
```
OsmondJiang: that's should also apply to OverwriteConfig, would you please help to change and add a small test for that?

mingwli: plan to make the `overwrite-with-null` default behavior instead of just a flag

yufeih:
```suggestion
        public static void Merge(JObject container, JObject overwrite, bool ignoreNulls = false)
```

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -245,9 +245,9 @@ public static void Merge(JObject container, JObject overwrite)
 
                 if (container[key] is JObject containerObj && value is JObject overwriteObj)
                 {
-                    Merge(containerObj, overwriteObj);
+                    Merge(containerObj, overwriteObj, overwriteWithNull);
                 }
-                else if (IsNullOrUndefined(container[key]) || !IsNullOrUndefined(value))
+                else if (overwriteWithNull || IsNullOrUndefined(container[key]) || !IsNullOrUndefined(value))
```
yufeih: Test `ignoreNulls` flag in `TestJsonMerge`

src/docfx/lib/json/JsonUtility.cs
live1206: `overwriteWithNull` means `NotIgnoreNulls`

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -233,7 +233,7 @@ public static (List<Error>, JToken) Parse(string json, string file)
             }
         }
 
-        public static void Merge(JObject container, JObject overwrite)
+        public static void Merge(JObject container, JObject overwrite, bool overwriteWithNull = true)
```
yufeih:
```suggestion
        public static void Merge(JObject container, JObject overwrite, NullValueHandling nullValueHandling = NullValueHandling.Include)
```

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -49,7 +49,7 @@ public MetadataProvider(Config config)
             if (yamlHeader != null)
             {
                 errors.AddRange(MetadataValidator.ValidateGlobalMetadata(yamlHeader));
-                JsonUtility.Merge(result, yamlHeader);
+                JsonUtility.Merge(result, yamlHeader, overwriteWithNull: true);
```
live1206: isn't `overwriteWithNull: true` default value already?

src/docfx/config/ConfigLoader.cs
```diff
@@ -3,18 +3,18 @@
...
 namespace Microsoft.Docs.Build
 {
     internal static class ConfigLoader
     {
-        private const string Files = "files";
-        private const string Exclude = "exclude";
         private const string Extend = "extend";
         private const string DefaultLocale = "defaultLocale";
         private const string Localization = "localization";
+        private static readonly string[] KeysToExpand = new string[] { "files", "exclude", "extend" };
```
yufeih:
```suggestion
        private static readonly string[] s_keysToExpand = new string[] { "files", "exclude", "extend" };
```

OsmondJiang: reuse the const string

src/docfx/cli/CommandLineOptions.cs
```diff
@@ -16,14 +16,20 @@ internal class CommandLineOptions
 
         public JObject ToJObject()
         {
+            var output = new JObject
+            {
+                ["path"] = Output != null ? (JValue)Output : JValue.CreateNull(),
```
yufeih:
```csharp
if (Output != null)
    output["path"] = Output,
```

src/docfx/config/ConfigLoader.cs
```diff
@@ -201,16 +201,18 @@ private static void OverwriteConfig(JObject config, string locale, string branch
...
+                    config[key] = ExpandStringArray(config[key]);
+            }
             return config;
         }
 
         private static JArray ExpandStringArray(JToken e)
         {
-            if (e is null)
-                return null;
+            Debug.Assert(!(e is null));
```
OsmondJiang: we can use `e != null`....

PR #4401, opened 2019-04-23
Support use commit build time for updated at

yufeih:
With this we can remove `GitCommitUpdateTimeResolver`.

- Add `updateTimeAsCommitBuildTime` config, when enabled, it'll update and save current build time and use it to populate updated at
- Add `DOCFX_CACHE_PATH`  to simplify service integration (save file copies)
- Add `cache` to e2e spec to test cache content

### DIFF
--- a/src/docfx/build/metadata/ContributionProvider.cs
+++ b/src/docfx/build/metadata/ContributionProvider.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Docs.Build
     {
         private readonly GitHubUserCache _gitHubUserCache;
 
-        private readonly Dictionary<string, DateTime> _updateTimeByCommit = new Dictionary<string, DateTime>();
+        private readonly CommitBuildTimeProvider _commitBuildTimeProvider;
 
         private readonly GitCommitProvider _gitCommitProvider;
 
@@ -24,8 +24,8 @@ namespace Microsoft.Docs.Build
 
             _gitHubUserCache = gitHubUserCache;
             _gitCommitProvider = gitCommitProvider;
-
-            LoadCommitsTime(docset);
+            _commitBuildTimeProvider = docset.Repository != null && docset.Config.UpdateTimeAsCommitBuildTime
+                ? new CommitBuildTimeProvider(docset.Repository) : null;
         }
 
         public async Task<(List<Error> error, Contributor author, List<Contributor> contributors, DateTime updatedAt)> GetAuthorAndContributors(
@@ -136,7 +136,7 @@ namespace Microsoft.Docs.Build
         {
             if (fileCommits?.Count > 0)
             {
-                return _updateTimeByCommit.TryGetValue(fileCommits[0].Sha, out var timeFromHistory)
+                return _commitBuildTimeProvider != null && _commitBuildTimeProvider.TryGetCommitBuildTime(fileCommits[0].Sha, out var timeFromHistory)
                     ? timeFromHistory
                     : fileCommits[0].Time.UtcDateTime;
             }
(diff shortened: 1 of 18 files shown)


src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -24,8 +24,8 @@ public ContributionProvider(Docset docset, GitHubUserCache gitHubUserCache, GitC
 
             _gitHubUserCache = gitHubUserCache;
             _gitCommitProvider = gitCommitProvider;
-
-            LoadCommitsTime(docset);
+            _commitBuildTimeProvider = docset.Repository != null && docset.Config.UpdateTimeAsCommitBuildTime
+                ? new CommitBuildTimeProvider(docset.Repository) : null;
```
OsmondJiang: what about to avoid "null" here?

src/docfx/lib/git/CommitBuildTimeProvider.cs
```diff
@@ -0,0 +1,59 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    internal class CommitBuildTimeProvider
+    {
+        private readonly Repository _repo;
+        private readonly string _commitBuildTimePath;
+        private readonly IReadOnlyDictionary<string, DateTime> _buildTimeByCommit;
+
+        public CommitBuildTimeProvider(Repository repo)
```
OsmondJiang: considering the submodule case?

src/docfx/config/AppData.cs
```diff
@@ -46,6 +49,11 @@ public static string GetCommitCachePath(string remote)
             return Path.Combine(CacheRoot, "commits", HashUtility.GetMd5Hash(remote));
         }
 
+        public static string GetCommitBuildTimePath(string remote, string branch)
+        {
+            return Path.Combine(CacheRoot, "history", $"build_history_{HashUtility.GetMd5Guid(remote)}_{HashUtility.GetMd5Guid(branch)}.json");
```
fenxu: >remote [](start = 94, length = 6)

Does this remote normalize?

fenxu: >GetMd5Guid [](start = 116, length = 10)

The method is different with original one. Make sure the value will be same

yufeih: It is the value of `DOCS_REPOSITORY_URL` or the raw value got from the current `git remote show`. docfx does not normalize it. If build service ensures `DOCS_REPOSITORY_URL` and git remote URL is normalized then this is correct.

yufeih: we do remove `.git` extension from `git remote show`

src/docfx/lib/git/CommitBuildTimeProvider.cs
yufeih: added a TODO

src/docfx/config/AppData.cs
yufeih: It is the same, I've updated test case to use `edge-developer` and the cache file path matches production path.

src/docfx/lib/git/CommitBuildTimeProvider.cs
```diff
@@ -0,0 +1,62 @@
...
+
+            _buildTimeByCommit = commitBuildTime.Commits.ToDictionary(item => item.Sha, item => item.BuiltAt);
+        }
+
+        public bool TryGetCommitBuildTime(string commitId, out DateTime time)
+        {
+            return _buildTimeByCommit.TryGetValue(commitId, out time);
+        }
+
+        public void UpdateAndSaveCache()
```
superyyrrzz: update test case to cover UpdateAndSave?

yufeih: that part is a bit tricky to do with spec test, since the result is unstable (built_at = now)

src/docfx/lib/git/CommitBuildTimeProvider.cs
OsmondJiang: what's the meaning of current Docs.Build implementation?

src/docfx/lib/git/CommitBuildTimeProvider.cs
OsmondJiang: use `commit` time for this test?

PR #4403, opened 2019-04-23
parse xref config to jArray

mingwli:


docs/specs/config.yml
```diff
@@ -491,3 +491,19 @@ inputs:
 outputs:
   build.log: |
     ["error","violate-schema","Expected type String, please input string or type compatible with String.","docfx.yml",4,7]
+---
```
yufeih: can just change an existing xref test case

PR #4413, opened 2019-04-24
remove fake user/contributors when github.resolveusers = false

OsmondJiang:
fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/86802/

### DIFF
--- a/src/docfx/build/metadata/ContributionProvider.cs
+++ b/src/docfx/build/metadata/ContributionProvider.cs
@@ -51,10 +51,12 @@ namespace Microsoft.Docs.Build
             var userIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
             var updatedDateTime = GetUpdatedAt(document, commits);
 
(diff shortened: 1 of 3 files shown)


src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -97,13 +94,10 @@ async Task<Contributor> GetAuthor()
...
                 if (!string.IsNullOrEmpty(authorName))
                 {
-                    if (resolveGitHubUsers)
-                    {
-                        // Remove author from contributors if author name is specified
-                        var (error, result) = await _gitHubUserCache.GetByLogin(authorName);
-                        errors.AddIfNotNull(error?.WithSourceInfo(authorName));
-                        return result?.ToContributor();
-                    }
+                    // Remove author from contributors if author name is specified
```
mingwli: >// Remove author from contributors if author name is specified [](start = 20, length = 62)

this line of comment seems misplaced before?

PR #4414, opened 2019-04-24
📃 Unify metadata validation source info

yufeih:
Unify line info handling for global metadata, file metadata and yaml header, so we have one entry for metadata validation with the correct source info


src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -41,20 +39,21 @@ public MetadataProvider(Config config)
...
             if (yamlHeader != null)
             {
-                errors.AddRange(MetadataValidator.ValidateGlobalMetadata(yamlHeader));
                 JsonUtility.Merge(result, yamlHeader);
             }
 
-            // We are validating against the merged JObject so discard the validation result here.
-            var (_, obj) = JsonUtility.ToObject<T>(result);
-            return (errors, obj);
+            var (errors, metadata) = JsonUtility.ToObject<T>(result);
```
yufeih: with this change, the merged `JObject` here has the correct line info for each value, so we have a single entry for metadata validation.

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -445,35 +441,31 @@ private static void HandleError(object sender, Newtonsoft.Json.Serialization.Err
...
-                    t_status.Value.Peek().Errors.Add(Errors.ViolateSchema(source, message));
+                    var state = t_status.Value.Peek();
+                    state.Errors.Add(Errors.ViolateSchema(GetSourceInfo(state.Reader.CurrentToken), RewriteErrorMessage(args.ErrorContext.Error.Message)));
                     args.ErrorContext.Handled = true;
                 }
             }
         }
 
-        private static (SourceInfo, string message) ParseException(Exception ex)
+        private static Error ToError(JsonReaderException ex, string file = null)
```
live1206: possible to get rid of null file here, like adding `file` to current `State`?

yufeih: `JsonUtility.Deserialize` currently does not have file path, so that one is null, we could at file path to `JsonUtility.Deserialize` in another PR.

docs/specs/legacy.yml
```diff
@@ -227,8 +227,8 @@ inputs:
 outputs:
   docs/a.json:
```
OsmondJiang: would you please add one test case like below:
```
inputs:
  docfx.yml: |
    fileMetadata:
      author:
        '**/*': XXX
  docs/a.md:
outputs:

yufeih: sure

PR #4426, opened 2019-04-25
 💬Usage -> Getting Started

yufeih:
Removed usage, as they are always out of date. Instead, add a simple getting started guide. `docfx build` runs `docfx restore` implicitly so there is no need to document that in getting started.


README.md
```diff
@@ -5,43 +5,17 @@
...
-```
-- docset_path: The docset directory which contains all files need to build and the `docfx.yml` config file.
-- options:
-  - -o/--output: The output directory in which to place built artifacts.
-  - --log: The output build log path.
-  - --github-token: The GitHub token used to get contribution information from GitHub API
-
--options:
-  - --retention-days: Only keep files/folder which are accessed/written within <retention-days> days.
+- Create a directory with a `docfx.yml` config file, markdown files and other contents.
```
mingwli: a help link on `docfx.yml` schema?

Issue #3882, opened 2019-01-09
resolve XRef for conceptuals


OsmondJiang:
Docfx has done the generate xref map file and uploding xref map file part, missing the import/restore xref map feature.

live1206:
https://github.com/dotnet/docfx/blob/70b9bfd98f9f2c1325655184d17345d8aada8486/docs/specs/xref.yml#L42
Is this scenario same as this test?

OsmondJiang:
Maybe we need config migration tool support?  @mingwli

yufeih:
Should compare external xref url with current host name, remove host name if they are the same.

OsmondJiang:
@yufeih, may we know why? seems it's totally different with what we talked in this issue #3885

live1206:
> Should compare external xref url with current host name, remove host name if they are the same.

will handle it in #3885

herohua:
we need to consider #3487 together with this issue

live1206:
## Restore related `xrefmap.json` to resolve external `uid`
With this, it would save all the network traffic to xref service, which would be much faster for build and authoring.
The main concern here is how to config where to get `xrefmap.json`, which depot/product name/share base path? For new repos, we assume the user should know where to get the uids. 

 - [ ] Needs to confirm with PM if it is OK for users to do so
 - [ ] For existing repos, we need to set up the configuration during migration, some thoughts:
      - [ ] v3 adds `depots` to the config file, which points where `xrefmap.json` locates
      - [ ] v2 helps to output the external uids referenced in the current repo during the build
      - [ ] migration tool calls xref service to somehow get the `depots` of the previous external uids, then add them into v3 config file
 - [ ] Given a `depot`, there could be multiple `xrefmap.json`, they need to be merged somewhere #3487 
      - [ ] short term: merge them during OP build, send the urls to docFX v3 config, the merging is already implemented [OP.Build](https://ceapex.visualstudio.com/Engineering/All/_git/OpenPublishing.Build?path=%2FLibraries%2FWorkflowTasks%2FXRefMap%2FXrefMapRestorer.cs&version=GBdevelop&line=36&lineStyle=plain&lineEnd=37&lineStartColumn=1&lineEndColumn=1), which is based on the shared base path, adjust it to use the config `depots`
      - [ ] long term: merge them in xref service, we had a discussion before but needs more confirmation

live1206:
https://github.com/MicrosoftDocs/engineering/pull/430

PR #4428, opened 2019-04-26
📝Add design of resolving external UIDs

live1206:
Add design of resolving external UIDs

#3882

### DIFF
--- a/docs/designs/xref.md
+++ b/docs/designs/xref.md
@@ -111,7 +111,6 @@ With this setting in place, during the OPS build we will loop throug the list an
 After getting all the urls, we will extend the DocFX config with them into `xref` settings, the one in the beginning of this section. Then, DocFX may restore the xref files as usual.
 
 #### Migration
-
 For the existing repositories using DocFX v2, we need to add the xref settings into configuration during the migration to v3.
 - Get the external UIDs used in the repository:
     - v2 help output the list of external UIDs during the build to prepare for v3
@@ -120,6 +119,9 @@ For the existing repositories using DocFX v2, we need to add the xref settings i
       For instance, given `System.String`, by calling `https://xref.docs.microsoft.com/query?uid=System.String`, we can get `"href":"https://docs.microsoft.com/dotnet/api/system.string"`. The first segment after the host will be the base path, which is `dotnet`. 
       > shall we always call xref service in production here?
     - Put the list of unique base paths from the previous step into `xrefBathPath` of `docfx.yml`
+
+#### New users or new uids
+Although the users of the existing respositories do not need to take action on the `xref` configuration, but if they need to use new external UIDs not covered by `xrefBasePath` after the migration to v3, they still need to add the related base path by following the steps under [migration](#migration). And the new users of DocFX v3 should always set the `xrefBasePath` by following the steps under [migration](#migration)
 
 ## Resolve xref
 ### Using `xref` to reference a uid in markdown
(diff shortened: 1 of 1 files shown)


docs/designs/xref.md
```diff
@@ -91,8 +91,42 @@ xref:
   - http://url1/.xrefmap.json
   - http://url2/.xrefmap.json
 ```
+User should be able to set a string instead of array, docfx will expand it to an array during the build.\
 During `docfx restore`, all the JSON files will be restored and merged.
 
+#### Xref base path
+Since we do not have a clear picture how we are going to support the local build, we will focus on OPS build firstly.\
+Unlike DocFX v2, the user of v3 needs to set where to get the `.xrefmap.json` files in `docfx.yml`.
```
yufeih: Should this be in `.openpublish.config.json`? `docfx.yml` currently does not exist in repo.

docs/designs/xref.md
```diff
@@ -91,8 +91,42 @@ xref:
...
+```yml
+xrefBasePath:
```
yufeih: Is this first URL segment or docset base path? because base path can contain multiple URL segments like `uwp/api`

live1206: we found that the user of docfx v2 always needs to config this in `xref_query_tags` under `.openpublishing.publish.config.json`, updating the doc

docs/designs/xref.md
live1206: Actually, it is in `.openpublishing.publish.config.json` already, maybe there is no need to do the migration

docs/designs/xref.md
```diff
@@ -91,8 +91,35 @@ xref:
...
 ```
+User should be able to set a string instead of array, docfx will expand it to an array during the build.\
 During `docfx restore`, all the JSON files will be restored and merged.
 
+#### Xref base path
+Since we do not have a clear picture how we are going to support the local build, we will focus on OPS build firstly.\
+The user needs to set where to get the `.xrefmap.json` files in `.openpublishing.publish.config.json`.
+```json
+{
+    "docsets_to_publish": [
```
OsmondJiang: why are we using ops config? is this a workaround or final design?

docs/designs/xref.md
```diff
@@ -91,8 +91,35 @@ xref:
...
+        }]
+}
+```
+
+#### OPS build
+With this setting in place, during the OPS build we will loop through `xref_query_tags` and get the urls of all `.xrefmap.json` files under each base path via calling DHS APIs. The reason to do so is, the public URL for each base path only contains 1 file of 1 depot with the highest priority, and there are many depots under 1 base path. These urls should only be valid for a short time.\
+\
+After getting all the urls, we will extend the DocFX config with them into `xref` settings, the one in the beginning of this section. Then, DocFX may restore the xref files as usual.
+> - The base path of current building repository should be added as default.
+> - The urls should exclude the current depot of the current buiding repository. Bucause this will be handled as internal UIDs.
```
OsmondJiang: will this be done in build service or docfx internal?

928PJY: typo

docs/designs/xref.md
live1206: This config in v2 also belongs to OPS config.
And as it describes, we are focusing on docs build for now, not local build.

docs/designs/xref.md
live1206: This logic is handled in OPS build

live1206:
moving it to OPS build

PR #4429, opened 2019-04-26
🛰Load html metadata name, reserved metadata from schema

yufeih:
This PR extends JSON schema to implement the ability of `SchemaValidationConsole`:

- `reserved`: an array of metadata names that are reserved by build, same as `blaclist.json`.
- `htmlMetaHidden`: a list of metadata names that should not show in HTML <meta> tags, same as `is_visible` in `op_schema.json`
- `htmlMetaName`: overwrite name for HTML <meta> tag, same as `display_name` in `op_schema.json`

Metadata type validation is also turned on in this PR, so after this PR is in, we can completely remove `SchemaValidationConsole` in docs build.

Corresponding changes in Docs.Build: https://ceapex.visualstudio.com/Engineering/_git/Docs.Build/pullrequest/11089?_a=files&path=%2FLibraries%2FMsBuildDebuggerProject%2FmetadataSchema.json

Corresponding changes in impact test: https://ceapex.visualstudio.com/Engineering/_git/Docs.DocFX.Impact/pullrequest/11090?_a=files


yufeih:
There is currently a null related problem, if we run this against any real docs repo, we'll get a lot of `unexpected-type` errors:

```
unexpected-type microsoft-edge/dev-guide/understanding-leaks.md(8,8): Expect type 'String' but got 'Null'
```

The source looks like `author:`, and YAML deserialize turn it into `{"author": null}`, per JSON schema validation, `null` is not a `string` type, thus the `unexpected-type` error.

yufeih:
To properly handle string or null, it requires the schema to be written as `{ "type": [ "string", "null" ] }`

src/docfx/lib/schema/JsonSchema.cs
```diff
@@ -10,15 +10,51 @@ namespace Microsoft.Docs.Build
...
+        /// </summary>
         public string[] Required { get; set; } = Array.Empty<string>();
+
+        // JSON schema custom extensions
+        //-------------------------------------------
+
+        /// <summary>
+        /// Alternative name used in output HTML <meta> tag. If not set, the original metadata name is used.
+        /// </summary>
+        public string HtmlMetaName { get; set; }
```
OsmondJiang: `HtmlMetaName` and `HtmlMetaHidden` are not clear for me since the schema is a nested object, will they be always describing root level metadata names or current level names?

yufeih: always root level because metadata is a flat list.

yufeih: This follows the same syntax as json schema `required`

OsmondJiang: `required` is not always in root level. So the `HtmlMetadataName` could contains `xxxx.yyy.zzz` or `xxx/yyy/zzz`?

yufeih: metadata is a flat list, it cannot be nested, so there is no need to access sub component.

PR #4432, opened 2019-04-28
Fix for index.png asset_id

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/87204

### DIFF
--- a/src/docfx/build/legacy/LegacyUtility.cs
+++ b/src/docfx/build/legacy/LegacyUtility.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Docs.Build
             }
 
             return PathUtility.NormalizeFile(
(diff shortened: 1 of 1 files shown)


src/docfx/build/legacy/LegacyUtility.cs
```diff
@@ -47,7 +47,7 @@ public static string ToLegacySiteUrlRelativeToBaseSitePath(this Document doc, Do
             }
 
             return PathUtility.NormalizeFile(
-                Path.GetFileNameWithoutExtension(doc.FilePath).Equals("index", PathUtility.PathComparison)
+                Path.GetFileNameWithoutExtension(doc.FilePath).Equals("index", PathUtility.PathComparison) && doc.ContentType == ContentType.Page
```
928PJY: `|| redirection`

live1206: Then it means `!= ContentType.Resource`

PR #4435, opened 2019-04-28
👨‍🏭 reuse existing work-tree

OsmondJiang:
to reduce the work tree count and improve performance

### DIFF
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -141,41 +141,46 @@ namespace Microsoft.Docs.Build
                         var gitDependencyLock = dependencyLock?.GetGitLock(remote, branch);
                         headCommit = gitDependencyLock?.Commit ?? headCommit;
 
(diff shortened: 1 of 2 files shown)


src/docfx/restore/RestoreGit.cs
```diff
@@ -141,41 +141,46 @@ void AddWorkTrees()
                         var gitDependencyLock = dependencyLock?.GetGitLock(remote, branch);
                         headCommit = gitDependencyLock?.Commit ?? headCommit;
 
-                        var (workTreePath, gitSlot) = RestoreMap.AcquireExclusiveGit(remote, branch, headCommit);
-                        workTreePath = Path.GetFullPath(workTreePath).Replace('\\', '/');
-                        var restored = true;
-
-                        try
+                        var (workTreePath, gitSlot) = RestoreMap.TryGetGitRestorePath(remote, branch, headCommit);
+                        if (workTreePath is null)
```
OsmondJiang: only change is here

PR #4438, opened 2019-04-29
remove checking for toc items

OsmondJiang:
there is no reason to check the length of toc items need to be bigger than 0

fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/88155

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -986,6 +986,7 @@ inputs:
       tocHref: 
       href: f1/b.md
       additionalProperty: a
+      items: []
   docs/f1/a.md:
   docs/f1/b.md:
 outputs:
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -986,7 +986,6 @@ inputs:
       tocHref: 
       href: f1/b.md
       additionalProperty: a
-      items: []
   docs/f1/a.md:
   docs/f1/b.md:
 outputs:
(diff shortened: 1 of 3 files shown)


yufeih:
@live1206 , where is the original TOC validation spec and does it mentioned anything about this validation?

live1206:
> @live1206 , where is the original TOC validation spec and does it mentioned anything about this validation?

https://review.docs.microsoft.com/en-us/new-hope/specs/ops/toc_schema_validation?branch=master
Seems like there is no requirement of minimum length 1, we can remove it

docs/specs/toc.yml
```diff
@@ -986,7 +986,6 @@ inputs:
       tocHref: 
       href: f1/b.md
       additionalProperty: a
-      items: []
```
live1206: keep this line to verify no error log

src/docfx/build/toc/TableOfContentsItem.cs
```diff
@@ -31,7 +30,6 @@ public class TableOfContentsItem
 
         public List<string> Monikers { get; set; }
 
-        [MinLength(1)]
         public List<TableOfContentsItem> Items { get; set; } = new List<TableOfContentsItem>();
```
yufeih: also remove from `TableOfContentsModel`

docs/specs/toc.yml
OsmondJiang: good idea

src/docfx/build/toc/TableOfContentsItem.cs
OsmondJiang: nice catch!

PR #4451, opened 2019-04-30
Fix for document id of landing page

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/88686

### DIFF
--- a/src/docfx/docset/Document.cs
+++ b/src/docfx/docset/Document.cs
@@ -456,8 +456,8 @@ namespace Microsoft.Docs.Build
                 ? sourcePath
                 : mappedSourcePath;
 
-            // if source ends with index.yml, change it to index.md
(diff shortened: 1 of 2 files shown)


src/docfx/docset/Document.cs
```diff
@@ -456,8 +456,8 @@ private static bool IsValidRelativePath(string path)
                 ? sourcePath
                 : mappedSourcePath;
 
-            // if source ends with index.yml, change it to index.md
-            if ("index.yml".Equals(Path.GetFileName(sourcePath).ToLowerInvariant()))
+            // if source is landing page and not resource, change it to *.md
+            if (Schema.Type == typeof(LandingData) && ContentType != ContentType.Resource)
```
OsmondJiang: do we need extra `ContentType != ContentType.Resource` ?

live1206: you are right, we don't need it

PR #4452, opened 2019-04-30
🐞 Fixes weak etag parsing error

yufeih:
When the service returns a weak etag that looks like `W/"99a5dbb8d73da1ec452d74c0ab45b11ebe8615a0"`, restore throws `Message: System.FormatException : The specified value is not a valid quoted string.`

This PR fix this bug.


OsmondJiang:
maybe add a test would be better if it's doable

yufeih:
> maybe add a test would be better if it's doable

RestoreUrl test current fail due to this bug

Issue #4434, opened 2019-04-28
remove implicit option for restore


OsmondJiang:
“implicit” option of `docfx restore` brings too many unpredictable inputs for `docfx build`.

cc @yufeih

yufeih:
go do it!

yufeih:
Most of our dependencies are floating range version, downloads dont have versions at all, and we cannot make implicit restore both fast and correct, removing it makes build fast and correct. The only downside is an extra madantory command, so lets drop it.

PR #4453, opened 2019-04-30
remove implicit restore incorrect optimization

OsmondJiang:
#4434

### DIFF
--- a/src/docfx/cli/Program.cs
+++ b/src/docfx/cli/Program.cs
@@ -71,18 +71,10 @@ namespace Microsoft.Docs.Build
                             Done(command, stopwatch.Elapsed, report);
                             break;
                         case "build":
-                            if (options.Restore)
-                            {
-                                await Restore.Run(docset, options, report);
-                            }
                             await Build.Run(docset, options, report);
                             Done(command, stopwatch.Elapsed, report);
                             break;
                         case "watch":
-                            if (options.Restore)
-                            {
-                                await Restore.Run(docset, options, report);
-                            }
                             await Watch.Run(docset, options);
                             break;
                     }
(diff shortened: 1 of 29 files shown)


OsmondJiang:
We will hold on this PR until we find a easy way to:
1. Remove implicit wrong logic
2. Keep the perf as same as right now.

I am going to add "--no-restore" to all our docfx callers

OsmondJiang:
The main perf concern is here:

image


Every test now will fetch remote, no matter the test dependencies repo exists or not.

cc @live1206  @yufeih

src/docfx/cli/Program.cs
```diff
@@ -71,12 +71,14 @@ internal static async Task<int> Run(string[] args)
                             Done(command, stopwatch.Elapsed, report);
                             break;
                         case "build":
-                            await Restore.Run(docset, options, report, @implicit: true);
+                            if (!options.NoRestore)
```
yufeih: remove `--no-restore`

OsmondJiang: i will do it in another PR

herohua:
what the error message will be like if someone runs `build` without explicitly `restore` first?

OsmondJiang:
> what the error message will be like if someone runs `build` without explicitly `restore` first?

In this PR, no error will thrown because docfx will automatically run `restore` before build. this PR just removed some incorrectly optimization for restore.

In my next PR, we will totally remove `implicit restore` feature, use will get `need-restore` error if some dependency repositories are not restored before build like below:

image

src/docfx/restore/RestoreFile.cs
```diff
@@ -13,21 +13,18 @@ namespace Microsoft.Docs.Build
...
+                restoreUrl => Restore(restoreUrl, config));
         }
 
-        public static async Task Restore(string url, Config config, bool @implicit = false)
+        public static async Task Restore(string url, Config config)
         {
             var filePath = GetRestoreContentPath(url);
 
-            var (existingContent, existingEtagContent) = RestoreMap.TryGetRestoredFileContent(url);
-            if (!string.IsNullOrEmpty(existingContent) && @implicit)
```
live1206: @OsmondJiang Just realized, we always need to restore from remote even the content already exists?

OsmondJiang: no, we don't

Issue #4455, opened 2019-04-30
Error handling for command line parser


OsmondJiang:
Currently wrong command line options in our spec test will cause test crashed without any useful information.

The root cause is that we didn't handle the command line parser errors.

PR #4457, opened 2019-04-30
avoid test crashed when options are wrong

OsmondJiang:
handle the command line parser error by our self, instead of exit the whole process


src/docfx/cli/Program.cs
```diff
@@ -103,36 +107,47 @@ private static (string command, string docset, CommandLineOptions options) Parse
...
+                    syntax.DefineOption("v|verbose", ref options.Verbose, "Enable diagnostics console output.");
+                    syntax.DefineParameter("docset", ref docset, "Docset directory that contains docfx.yml/docfx.json.");
+                });
+
+                options.Locale = options.Locale?.ToLowerInvariant();
+                return (command, docset, options);
+            }
+            catch (ArgumentSyntaxException ex)
+            {
+                Log.Write(ex);
```
yufeih: this changed the existing command line experience

OsmondJiang: image

image

image

OsmondJiang:
#4455

PR #4459, opened 2019-04-30
🚀Replace Concurrent with manual lock for add only operation

live1206:
https://github.com/dotnet/corefx/issues/12067
For add only operation
- `ConcurrentDictionary` to `Dictionary` with manual lock
- `ConcurrentBag` to `List` with manual lock


src/docfx/build/legacy/LegacyDependencyMap.cs
```diff
@@ -17,7 +15,7 @@ internal static class LegacyDependencyMap
         {
             using (Progress.Start("Convert Legacy Dependency Map"))
             {
-                var legacyDependencyMap = new ConcurrentBag<LegacyDependencyMapItem>();
+                var legacyDependencyMap = new List<LegacyDependencyMapItem>();
```
yufeih: better yet, we can have an `ArrayBuilder` and a `DictionaryBuilder` that encapsulate the lock logic?

live1206: sounds good

src/docfx/lib/ArrayBuilder.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) Microsoft. All rights reserved.
```
yufeih: consider move `ArrayBuilder`, `DictionaryBuilder`, `ConcurrentHashSet`, `CollectionUtility` into `lib/collections`

src/docfx/lib/ArrayBuilder.cs
```diff
@@ -0,0 +1,26 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+
+namespace System.Collections.Concurrent
+{
+    internal class ArrayBuilder<T> : IEnumerable<T>
```
yufeih:
```suggestion
    internal readonly struct ArrayBuilder<T>
```

live1206: use `class` because `struct` doesnt support parameterless constructor and field initialization

src/docfx/lib/collections/DictionaryBuilder.cs
```diff
@@ -0,0 +1,23 @@
...
+
+        public bool TryAdd(TKey key, TValue value)
+        {
+            lock (_dictionary)
+            {
+                return _dictionary.TryAdd(key, value);
+            }
+        }
+
+        public IReadOnlyDictionary<TKey, TValue> ToDictionary()
```
OsmondJiang: how about return ReadOnlyDictionary and ReadOnlyList? then the user can't modify the inside _dictionary and _list

live1206: you are right

Issue #4461, opened 2019-04-30
Build log are missing for docfx not found case


OsmondJiang:
We should always be able to generate build.log as long as docfx doesn't crashed.

PR #4462, opened 2019-04-30
report to default output if config is not found

OsmondJiang:
#4461


src/docfx/lib/log/Report.cs
```diff
@@ -33,28 +33,27 @@ internal sealed class Report : IDisposable
...
         }
 
-        public void Configure(string docsetPath, Config config)
+        public void Configure(Config config)
         {
-            var outputPath = Path.Combine(docsetPath, config.Output.Path, "build.log");
-            Debug.Assert(_outputPath is null || _outputPath == outputPath, "Cannot change report output path");
+            Debug.Assert(!_output.IsValueCreated || _config.Output.Path == config.Output.Path, "Cannot change report output path");
 
             _config = config;
```
yufeih: With this change, If I set `output.path` in config, that config will not be respected

Issue #4209, opened 2019-03-20
h1 and metadata validation in v3


herohua:

herohua:
@kimizhu

yufeih:
track this work in #4605

PR #4465, opened 2019-04-30
Allow multiple metadata schema definitions

yufeih:
This allows `op_schema.json` and metadata validation schema to be defined and managed at different places. #4209


src/docfx/docset/Docset.cs
```diff
@@ -251,12 +252,13 @@ private CultureInfo CreateCultureInfo(string locale)
...
         private JsonSchema LoadMetadataSchema(Config config)
         {
-            if (!string.IsNullOrEmpty(config.MetadataSchema))
+            var token = new JObject();
+            foreach (var metadataSchema in config.MetadataSchema)
             {
-                var (_, content, _) = RestoreMap.GetRestoredFileContent(this, config.MetadataSchema);
-                return JsonUtility.Deserialize<JsonSchema>(content);
+                var (_, content, _) = RestoreMap.GetRestoredFileContent(this, metadataSchema);
+                JsonUtility.Merge(token, (JObject)JsonUtility.Parse(content, metadataSchema).value);
```
live1206: what if `content` is not `JObject`?

yufeih: If will throw and crash docfx.

PR #4468, opened 2019-04-30
Fixed truncated strings in "In This Article" nav

### LINKED ISSUE #3166: String truncation in the 'IN THIS ARTICLE' section when window width exceed the max. content width.
**DocFX Version Used**: 
2.37.2.0
2.37.1.0

Unknown

**Steps to Reproduce**:

1. Go to http://dotnet.github.io/docfx/tutorial/docfx_getting_started.html
2. Maximize the browser window so that there is no text wrapping in the 'IN THIS ARTICLE' section
3. Observe the truncation on the #3 topic (3. Use DocFX integrated with Visual Studio)
4. Resize the window width so that the text wrapping happens in the 'IN THIS ARTICLE' section
5. Above mentioned string is wrapped properly, and there is no truncation any more

This issue repros both in Edge and Chrome

**Expected Behavior**:
There is no string truncation regardless of the window width.

**Actual Behavior**:
There is a string truncation when window width exceed the max. content width.

kborchers:
- Remove overflow:hidden styles
 - Set .affix > ul.level1 width to 100%

Fixed #3166

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.css
+++ b/src/docfx.website.themes/default/styles/docfx.css
@@ -633,17 +633,11 @@ body .toc{
   height: calc(100% - 100px);
   margin-right: -20px;
 }
-.affix ul > li > a:before {
-  color: #cccccc;
-  position: absolute;
-}
 .affix ul > li > a:hover {
   background: none;
   color: #666666;
-}
-.affix ul > li.active > a,
-.affix ul > li.active > a:before {
-  color: #337ab7;
+  border-left: 3px solid #cccccc;
+  padding-left: 9px;
 }
 .affix ul > li > a {
   padding: 5px 12px;
@@ -652,16 +646,12 @@ body .toc{
 .affix > ul > li.active:last-child {
   margin-bottom: 50px;
 }
-.affix > ul > li > a:before {
-  content: "|";
-  font-size: 16px;
-  top: 1px;
-  left: 0;
-}
-.affix > ul > li.active > a,
-.affix > ul > li.active > a:before {
+.affix > ul > li.active > a {
   color: #337ab7;
   font-weight: bold;
+}
+.affix > ul > li.active > a:hover {
+  border-color: #337ab7;
 }
 .affix ul ul > li > a {
   padding: 2px 15px;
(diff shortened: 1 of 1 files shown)


superyyrrzz:
Thank you for contribution! I just tested it. It works well on H2 in side affix, but has some issue in H3. Note that the `>` mark is separated from title to different line:

image

Could you please help fix this issue?

kborchers:
Ah, yes I missed that. I will push a fix to the PR ASAP.

src/docfx.website.themes/default/styles/docfx.css
```diff
@@ -629,22 +628,16 @@ body .toc{
   font-size: 12px;
 }
 .affix > ul.level1 {
-  overflow: hidden;
+  width: 100%;
   padding-bottom: 10px;
   height: calc(100% - 100px);
   margin-right: -20px;
```
KalleOlaviNiemitalo: Before I saw this pull request, I tried to hack around the problem by editing the CSS in a browser, and just removing this `margin-right: -20px;` setting seemed to do the trick. Is it needed for anything?

superyyrrzz: I cannot think of why this is needed... It's ok to remove this line.

kborchers: I will try to work on this PR again today to address the earlier comments but specifically on this one, that should get you another 20px but will not eliminate the issue of truncating long lines if they go beyond that extra 20px.

superyyrrzz:
ping @kborchers Any update on this PR?

superyyrrzz:
@kborchers I will first close this as #4771 fixed it.

kborchers:
My apologies as I have just been buried with work and haven't been able to get back to this. That said, I'm pretty sure that other PR only partially fixes this issue. The entire navigation is truncated at the bottom in certain screen/view port sizes.

superyyrrzz:
Thank you for the updates! I have the same feeling that it does not address the root cause, while it can make the situation slightly better to address the specific issue on that page. BTW, you can go back to fix it completely when you have time.

PR #4474, opened 2019-05-05
update output path

928PJY:
output path schema:
`{output-dir}/{monikerListHash}}?/{site-path}`
->
`{output-dir}/{siteBasePath}/{monikerListHash}?/{path-to-site-base-path}`
> `path-to-site-base-path` means the relative path of sitePath related to siteBasePath

### DIFF
--- a/docs/specs/moniker.yml
+++ b/docs/specs/moniker.yml
@@ -2,6 +2,8 @@
 # Apply moniker range setting in the config to markdown file's output path and output content
 inputs:
   docfx.yml: |
+    documentId:
+      siteBasePath: docs
     monikerRange:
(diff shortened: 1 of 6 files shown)


docs/specs/moniker.yml
```diff
@@ -2,6 +2,8 @@
 # Apply moniker range setting in the config to markdown file's output path and output content
 inputs:
   docfx.yml: |
+    documentId:
+      siteBasePath: docs
```
yufeih: consider promote `documentId/siteBasePath` to `/baseUrl` in another PR, where `BaseUrl` is `https://{hostName}/{basePath}`

PR #4475, opened 2019-05-05
🐞fix for parsing uid spec string

live1206:


src/docfx/build/xref/XrefMap.cs
```diff
@@ -248,6 +252,13 @@ private static string GetSubstringFromContent(string content, int startLine, int
             var result = new StringBuilder();
             var currentLine = 1;
             var currentColumn = 1;
+
+            // for better performance by accessing index when content is 1 line
+            if (currentLine == startLine && currentLine == endLine)
+            {
+                return content.Substring(startColumn - 1, endColumn - startColumn + 1);
```
superyyrrzz: Move this to L275? Then this optimization is not limited to currentLine == startLine == endLine **== 1**

live1206: This only works for **1** line content. If `currentLine == startLine == endLine > 1` we do not know the index directly

yufeih:
@live1206 , what is the performance data for loading `/dotnet` xrefmap before and after JSON reader optimization?

live1206:
> @live1206 , what is the performance data for loading `/dotnet` xrefmap before and after JSON reader optimization?

Original, it was around 48s and now it is around 13s

PR #4476, opened 2019-05-06
Make yaml-duplicated-key a non fatal error

yufeih:
Now `yaml-duplicated-key` is a fatal error that cannot be suppressed to warning or info because it throws, this PR turns `yaml-duplicated-key` into a non-fatal error by using YamlDotNet event API.


src/docfx/lib/json/YamlUtility.cs
```diff
@@ -62,7 +59,7 @@ public static string ReadHeader(string yaml)
         /// </summary>
         public static T Deserialize<T>(string input)
         {
-            var token = ParseAsJToken(input, file: null);
+            var (_, token) = ParseAsJToken(input, file: null);
```
live1206: why `RemoveNulls` is not applied here?

yufeih: Same as `JsonUtility.Deserialize` not removing nulls

herohua:
This is great! It would make migration process more smooth. :)

PR #4480, opened 2019-05-06
docfx changes need for `Docs.Build` regression test

yufeih:
- Add `noTest` to `build.ps1` for faster build
- Stablize legacy manifest
- Rename legacy manifest item to match output

### DIFF
--- a/src/docfx/build/legacy/manifest/LegacyManifest.cs
+++ b/src/docfx/build/legacy/manifest/LegacyManifest.cs
@@ -95,9 +95,9 @@ namespace Microsoft.Docs.Build
                         }
                         var file = new LegacyManifestItem
                         {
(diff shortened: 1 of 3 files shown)


src/docfx/build/legacy/manifest/LegacyManifest.cs
```diff
@@ -127,7 +127,7 @@ public static void Convert(Docset docset, Context context, Dictionary<Document,
                         version_folder = string.Empty,
                         xref_map = "xrefmap.yml",
                     },
-                    files = convertedItems.Select(f => f.manifestItem),
+                    files = convertedItems.OrderBy(f => f.manifestItem.AssetId).Select(f => f.manifestItem),
```
live1206: what's the `OrderBy` for?

Issue #4222, opened 2019-03-21
Remove base path related config


yufeih:
To change base path, it is best to only have 1 config that contains base path info like `base_url`. Right now `routes` also has base path info, we should remove them.

- [ ] Remove base path from `routes`
- [ ] Remove base path from `publish.json`

OsmondJiang:
the site base path config is also related with "output.lowerCaseUrl", @yufeih

mingwli:
https://ceapex.visualstudio.com/Engineering/_git/Docs.Build/pullrequest/11307?_a=overview

mingwli:
https://ceapex.visualstudio.com/Engineering/_git/Docs.Build/pullrequest/11311?_a=overview

yufeih:
The output `url` should also be an URL without `base_url`

yufeih:
For routes, the conclusion is only remove `site_base_path`, leave `source_base_path` for document id only.

herohua:
@928PJY @mingwli to check whether this has been finished or not

PR #4481, opened 2019-05-06
remove siteBasePath from routes and always prepend to sitePath

928PJY:
#4222 

require changes of config migration tool @mingwli :
1. remove `siteBasePath` from routes


docs/specs/metadata.yml
```diff
@@ -95,6 +97,8 @@ inputs:
     product: Win
     files: "**/*"
     baseUrl: https://docs.com/sccm
+    routes:
```
OsmondJiang: there are two concerns:
- User need combine two config "baseUrl" and "routes" to get the finally site url
- Routes expose something like `internal mapping` out to end user, it doesn't map to output path or site path

OsmondJiang: actually the `routes` in this change is more like `sourceBasePath`.

OsmondJiang: we can provide `sourceBasePath` + `baseUrl` experiences to docs user and `routes` to community users

src/docfx/docset/Docset.cs
```diff
@@ -170,11 +170,11 @@ internal class Docset
             Config = config;
             DocsetPath = PathUtility.NormalizeFolder(Path.GetFullPath(docsetPath));
             Locale = locale.ToLowerInvariant();
-            Routes = NormalizeRoutes(config.Routes);
+            (HostName, SiteBasePath) = SplitBaseUrl(config.BaseUrl);
+            Routes = NormalizeRoutes(config.Routes,  config.SourceBasePath, SiteBasePath);
```
yufeih:
```suggestion
            Routes = NormalizeRoutes(config.Routes, config.SourceBasePath, SiteBasePath);
```

src/docfx/docset/Docset.cs
```diff
@@ -224,7 +224,7 @@ Repository GetRepositoryInternal(string fullPath)
             }
         }
 
-        private static IReadOnlyDictionary<string, string> NormalizeRoutes(Dictionary<string, string> routes)
+        private static IReadOnlyDictionary<string, string> NormalizeRoutes(Dictionary<string, string> routes, string sourceBasePath, string siteBasePath)
```
yufeih: where is `sourceBasePath` used?

928PJY: line 239

---
In reply to: [281133469](https://github.com/dotnet/docfx/pull/4481#discussion_r281133469) [](ancestors = 281133469)

docs/specs/metadata.yml
```diff
@@ -199,16 +193,15 @@ inputs:
...
+    sourceBasePath: articles
     baseUrl: https://docs.com/azure
-    routes:
-      articles/: "azure/"
   articles/iot-central/concepts-architecture.md:
   articles/billing/billing-add-change-azure-subscription-administrator.md:
 outputs:
   azure/iot-central/concepts-architecture.json: |
-    {"document_id":"7e13132e-7847-dc4f-90d5-32a233d1c81a","document_version_independent_id":"d9cc5ab8-926a-8a5a-312c-10ca36619bad"}
+    {"document_id":"57901a4c-fd14-9bd5-0dc2-fbd965db5287","document_version_independent_id":"7d7a36da-9cce-c2dd-3e04-ce9ae5e197e5"}
```
OsmondJiang: if the document id changed, means we did something wrong here

src/docfx/docset/Docset.cs
```diff
@@ -225,15 +225,24 @@ Repository GetRepositoryInternal(string fullPath)
             }
         }
 
-        private static IReadOnlyDictionary<string, string> NormalizeRoutes(Dictionary<string, string> routes)
+        private static IReadOnlyDictionary<string, string> NormalizeRoutes(Dictionary<string, string> routes, string sourceBasePath)
```
yufeih: consider changing the logic inside `ApplyRoutes` over here, that is where all the logic is applied, this method just does path normalization.

928PJY: if I move this logic into ApplyRoutes, this logic will run everytime the apply routes is called
this method is to normalize the routes, which is exactly what I am doing, right? 

---
In reply to: [284527991](https://github.com/dotnet/docfx/pull/4481#discussion_r284527991) [](ancestors = 284527991)

PR #4484, opened 2019-05-07
🚁 change output model to template consuming raw model

OsmondJiang:
content -> conceptual
word_count ❌
schema_type ❌
DepotName/_path/CanonicalUrlPrefix/PdfUrlPrefixTemplate -> output model


OsmondJiang:
no need keep both `word_count` and `wordCount`, blow is the template logic:

image

cc @superyyrrzz  @yufeih

src/docfx/build/page/OutputModel.cs
```diff
@@ -49,7 +53,16 @@ public class OutputModel : InputMetadata
 
         public string SiteName { get; set; }
 
-        public string Conceptual => Content as string;
+        public string DepotName { get; set; }
+
+        [JsonProperty("_path")]
```
yufeih: `_path` can can be removed

OsmondJiang: _path is used to calculate canonical url in template:

image

yufeih: ok, we then need to update template code after next weeks meeting, because `_versionPath` is also missing.

PR #4487, opened 2019-05-07
Schema error level

mingwli:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/90328?src=WorkItemMention&src-action=artifact_link

### DIFF
--- a/docs/specs/validation.yml
+++ b/docs/specs/validation.yml
@@ -57,7 +57,12 @@ inputs:
     key3: net462
     ---
 outputs:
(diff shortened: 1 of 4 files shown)


docs/specs/validation.yml
```diff
@@ -57,10 +57,11 @@ inputs:
     key3: net462
     ---
 outputs:
+  a.json:
```
OsmondJiang: give a example of a.json here?

PR #4490, opened 2019-05-08
📃Json/yaml deserialize must have file info

yufeih:


test/docfx.Test/build/TocTest.cs
```diff
@@ -13,7 +13,7 @@ public static class TocTest
             new Report(),
             Directory.GetCurrentDirectory(),
             "en-us",
-            JsonUtility.Deserialize<Config>("{'output': { 'json': true } }".Replace('\'', '\"')),
+            JsonUtility.Deserialize<Config>("{'output': { 'json': true } }".Replace('\'', '\"'), null),
```
live1206: consider input `string.Empty` instead of `null` here? we can have a non-null check for file

PR #4491, opened 2019-05-08
Remove sas token for blob url while create short folder name

live1206:
For azure blob URL, remove the sas token from the URL to identify if the file has been restored locally

### DIFF
--- a/src/docfx/lib/PathUtility.cs
+++ b/src/docfx/lib/PathUtility.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Docs.Build
 
         private static readonly HashSet<char> s_invalidPathChars = Path.GetInvalidPathChars().Concat(Path.GetInvalidFileNameChars()).Distinct().ToHashSet();
 
-        private static readonly Regex s_blobUrl = new Regex(@"^https:\/\/\w+.blob.core.windows.net(/\w+)*\?(.*)$", RegexOptions.IgnoreCase);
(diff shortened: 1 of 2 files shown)


src/docfx/lib/PathUtility.cs
```diff
@@ -25,6 +26,8 @@ internal static class PathUtility
 
         private static readonly HashSet<char> s_invalidPathChars = Path.GetInvalidPathChars().Concat(Path.GetInvalidFileNameChars()).Distinct().ToHashSet();
 
+        private static readonly Regex s_blobUrl = new Regex(@"^https:\/\/\w+.blob.core.windows.net(/\w+)*\?(.*)$", RegexOptions.IgnoreCase);
```
yufeih:
```suggestion
        private static readonly Regex s_blobUrl = new Regex(@"^https:\/\/.+?.blob.core.windows.net\/.*?\?(.*)$", RegexOptions.IgnoreCase);
```

yufeih: `\w` only match against alphabet or underscore, url could contain other characters like `-`

src/docfx/lib/PathUtility.cs
```diff
@@ -202,6 +207,18 @@ public static string UrlToShortName(string url)
             return result.ToString();
         }
 
+        // For azure blob url, url without sas token should identify if the content has changed
+        // https://docs.microsoft.com/en-us/azure/storage/common/storage-dotnet-shared-access-signature-part-1#how-a-shared-access-signature-works
+        private static string RemoveQueryForBlobUrl(string url)
```
yufeih: consider inside this and use `Regex.Replace`

PR #4492, opened 2019-05-08
Use OneOrManyJsonConverter to expand config

yufeih:
makes things cleaner


src/docfx/config/Config.cs
```diff
@@ -62,6 +64,7 @@ internal sealed class Config
         /// The extend file addresses
         /// The addresses can be absolute url or relative path
         /// </summary>
+        [JsonConverter(typeof(OneOrManyConverter))]
```
OsmondJiang: Is this OneOrManyConverter only working for string type?

yufeih: works for arrays of any type

PR #4498, opened 2019-05-08
🧹report --> errorlog

yufeih:
- Rename `Report` to `ErrorLog` because `Report` could also mean the generated HTML build report.
- `build.log` --> `.errors.log` (restore also writes to build.log)


OsmondJiang:
Will HTML build report be part of our docfx output? If not, i don't think this change is necessary

yufeih:
> Will HTML build report be part of our docfx output? If not, i don't think this change is necessary

It is likely to be in `Docs.Build` output

OsmondJiang:
> > Will HTML build report be part of our docfx output? If not, i don't think this change is necessary
> 
> It is likely to be in `Docs.Build` output

So why we need change code in docfx?

OsmondJiang:
> > > Will HTML build report be part of our docfx output? If not, i don't think this change is necessary
> > 
> > 
> > It is likely to be in `Docs.Build` output
> 
> So why we need change code in docfx?

Let's consider from another point, what's the report will be using docfx directly?

yufeih:
> > > Will HTML build report be part of our docfx output? If not, i don't think this change is necessary
> > 
> > 
> > It is likely to be in `Docs.Build` output
> 
> So why we need change code in docfx?

To make integration easier. Right now `Docs.Build` copies and append each log file after each command in a finally block, if every command append report to the same location, this post step could be removed.

Issue #4489, opened 2019-05-08
Prepare monikerDefinition file for versioning repository build


928PJY:
Three options:
1. Call OP Build rest API `GET /v2/monikertrees/allfamiliesproductsmonikers`:
    - There will be two new configs in the global config, the definition file will be downloaded in the restore step:
    ```
    monikerDefinition: https://op-build-sandbox2.azurewebsites.net/v2/monikertrees/allfamiliesproductsmonikers
    http:
        https://op-build-sandbox2.azurewebsites.net/v2/monikertrees/allfamiliesproductsmonikers:
            headers:
                X-OP-BuildUserToken: xxxx-xxxxx
    ```
    1. By running original `docsv3.ps1`, the token is already ready in the global metadata now, we can use it directly.
    2. By running the docsBuild.exe, The token can be saved in a key vault, and restore the token with the key vault connect string passed from environment variables.

2. Call DHS API
    To call DHS API, we need the `BuildServiceAADClientId` and `BuildServiceAADClientSecret` to use the `BuildServiceAADClient`, which is stored in a key vault, so we still need a key vault connecting string to restore those tokens.
    1. By running original `docsv3.ps1`, the connecting string can be passed by the global metadata.
    2. By running the docsBuild.exe,  the connecting string can be passed from environment variables.

3. Call docs API
    There is no auth required for those APIs, but currently, there are just API to get moniker by moniker name.

928PJY:
Use the first approach

PR #4500, opened 2019-05-08
rename moniker Model properties

928PJY:
#4489

### DIFF
--- a/src/docfx/lib/moniker/Moniker.cs
+++ b/src/docfx/lib/moniker/Moniker.cs
@@ -3,16 +3,16 @@
 
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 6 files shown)


src/docfx/lib/moniker/Moniker.cs
```diff
@@ -8,8 +8,10 @@ namespace Microsoft.Docs.Build
 {
     internal class Moniker
     {
+        [JsonProperty(PropertyName = "moniker_name")]
         public string Name { get; set; } = string.Empty;
```
yufeih: To be consistent in with data model:
```suggestion
        public string MonikerName { get; set; } = string.Empty;
```

Also use `[JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]` over `[JsonProperty]`

PR #4502, opened 2019-05-08
Update log code 'at-uid-not-found'

v-pegao:
| Log Code | Changes |
|----------|---------------|
| at-uid-not-found | log code, message  |
| uid-not-found | log code, message |


src/docfx/Errors.cs
```diff
@@ -256,8 +256,8 @@ public static Error MergeConflict(SourceInfo source)
         /// <summary>
         /// Failed to resolve uid defined by @ syntax.
         /// </summary>
-        public static Error AtUidNotFound(SourceInfo<string> source, string uid)
-            => new Error(ErrorLevel.Info, "at-uid-not-found", $"Cannot find uid '{uid}' using xref '{source}'", source);
+        public static Error XrefNotFound(SourceInfo<string> source)
```
yufeih: Keep `AtXrefNotFound` and `XrefNotFound` and there error level, only change code and message, also can you help move these error codes adjacent to each other.

v-pegao: Updated. But will set the PR to WIP status and add the log codes to confirmation list.

- `at-uid-not-found` is expected to be renamed as `xref-not-found` and its log level is `warning` in PM's request
- When checking PM's notes, they said `Why have a special case for @ UIDs? Just show them as <xref:...> or @... as appropriate.`. Maybe they want to merge `at-uid-not-found` and `uid-not-found`.

yufeih: @mingwli @v-pegao ,

> Why have a special case for @ UIDs? Just show them as <xref:...> or @... as appropriate

Here are a huge number of cases where ordinary `@` is parsed as xref by mistake, so this is also an info in v2.

PR #4504, opened 2019-05-08
Update 'config-not-found', 'download-failed'

v-pegao:
| Log Code | Changes |
|----------|---------------|
| config-not-found | message |
| download-failed | message |

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -104,7 +104,7 @@ inputs:
     extend: https://this-url-does-not-exit.com/this-is-a-bad-url
 outputs:
   build.log: |
-    ["error","download-failed","Download 'https://this-url-does-not-exit.com/this-is-a-bad-url' failed: *"]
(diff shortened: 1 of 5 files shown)


docs/specs/restore.yml
```diff
@@ -104,7 +104,7 @@ inputs:
     extend: https://this-url-does-not-exit.com/this-is-a-bad-url
 outputs:
   build.log: |
-    ["error","download-failed","Download 'https://this-url-does-not-exit.com/this-is-a-bad-url' failed: *"]
+    ["error","download-failed","System: Download failed for file 'https://this-url-does-not-exit.com/this-is-a-bad-url'. Try closing and reopening the PR. If you get this Error again, file an issue."]
```
yufeih: We should have a separate category indicating whether this is a system error, with that, we may remove `System: ` prefix.

v-pegao: Removed. The category seems not working for now, build report always shows `Unspecific` type for docfx log codes.

PR #4511, opened 2019-05-09
Update log code `github-api-failed`, `git-log-error`, `git-not-found`

v-pegao:
| Log Code | Changes |
|----------|---------------|
| github-api-failed | message |
| ~~git-log-error~~ | deleted|
| git-not-found | message |

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -306,7 +306,7 @@ namespace Microsoft.Docs.Build
         /// Git.exe isn't installed.
         /// </summary>
         public static Error GitNotFound()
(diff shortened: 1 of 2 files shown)


src/docfx/Errors.cs
```diff
@@ -300,13 +300,13 @@ public static Error InvalidMetadataType(SourceInfo source, string name)
         /// Failed to compute specific info of a commit.
         /// </summary>
         public static Error GitLogError(string repoPath, int errorCode)
```
yufeih: @mingwli, I think we can remove this error code and make it an `NotSupportedException`, if we reach here, better let docfx crash

Issue #4510, opened 2019-05-09
Allow custom error message, code and level per metadata per error


yufeih:
Related #4209

yufeih:
track this work in #4605

PR #4514, opened 2019-05-09
Update log code `invalid-redirection`, `attribute-reserved`, `upload-failed`

v-pegao:
| Log Code | Changes |
|----------|---------------|
| invalid-redirection | log code, message |
| attribute-reserved | log code, message |
| upload-failed | message |


docs/specs/legacy.yml
```diff
@@ -267,8 +267,8 @@ inputs:
 outputs:
   docs/a.json:
   build.log: |
-    ["warning","reserved-metadata","Metadata 'content_type' is reserved by docfx, remove this metadata","docfx.yml",3,16]
-    ["warning","reserved-metadata","Metadata 'ms.contentlang' is reserved by docfx, remove this metadata","docs/a.md",2,1]
+    ["warning","attribute-reserved","Attribute content_type is reserved for use by Docs. Remove it from your file metadata.","docfx.yml",3,16]
+    ["warning","attribute-reserved","Attribute ms.contentlang is reserved for use by Docs. Remove it from your file metadata.","docs/a.md",2,1]
```
OsmondJiang: that's more like a error message/code for Docs, not for docfx? @yufeih  @herohua

yufeih: yes, #4510 is tracking this, right now my thinking is we can first checkin all the docs message, then gradually move then outside to config files once #4510 is done

PR #4515, opened 2019-05-09
Update log code `schema-not-found`, `uid-conflict`

v-pegao:
| Log Code | Changes |
|----------|---------------|
| schema-not-found | message |
| uid-conflict | message |


src/docfx/Errors.cs
```diff
@@ -404,7 +404,7 @@ public static Error UidConflict(string uid, IEnumerable<string> conflicts = null
             }
 
             var hint = conflicts.Count() > 5 ? "(Only 5 duplicates displayed)" : "";
-            return new Error(ErrorLevel.Error, "uid-conflict", $"Two or more documents have defined the same Uid '{uid}': {string.Join(',', conflicts.Take(5))}{hint}");
+            return new Error(ErrorLevel.Error, "uid-conflict", $"UID '{uid}' is defined in more than one file: {string.Join(',', conflicts.Take(5))}{hint}");
```
yufeih: consider replace all the `string.Join` in `Errors.cs` with `Join` method to ensure all values a quoted.

src/docfx/Errors.cs
```diff
@@ -404,7 +404,7 @@ public static Error UidConflict(string uid, IEnumerable<string> conflicts = null
             }
```
yufeih: there is another case in line 403 where multiple UID is defined in the same file

src/docfx/Errors.cs
v-pegao: Updated. Only `circular-reference` is using `string.Join()` in `Errors.cs` now

src/docfx/Errors.cs
v-pegao: I didn't find the new message for the case in line 403 and the new message is not fit for the case in line 403 as it mentioned `more than one file`. Let me know if It needs update.

yufeih: It needs update, but probably not in this PR, @mingwli this is another item to confirm

OsmondJiang:
you need merge v3 firstly

PR #4516, opened 2019-05-09
fix remove duplicate output files in manifestitems

partychen:


src/Microsoft.DocAsCode.Build.Engine/ManifestUtility.cs
```diff
@@ -19,8 +19,17 @@ public static void RemoveDuplicateOutputFiles(ManifestItemCollection manifestIte
             {
                 throw new ArgumentNullException(nameof(manifestItems));
             }
+            var items = new List<ManifestItem>();
+            foreach (var uniques in (from m in manifestItems
```
herohua: combine two iteration together

superyyrrzz: avoid duplicate code

PR #4518, opened 2019-05-09
don't remove resource output when the content type is resource and copyResources is false

928PJY:
when there is some error happen in the resource file, and the `coprResource` config is false, we will delete the original resource file in the current build folder.

### DIFF
--- a/src/docfx/publish/PublishModelBuilder.cs
+++ b/src/docfx/publish/PublishModelBuilder.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Docs.Build
 
                     foreach (var conflictingFile in files.Keys)
                     {
-                        if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null)
+                        if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && (conflictingFile.ContentType != ContentType.Resource || conflictingFile.Docset.Config.Output.CopyResources))
                         {
                             context.Output.Delete(item.Path, legacy);
                         }
@@ -93,7 +93,7 @@ namespace Microsoft.Docs.Build
 
                 foreach (var conflictingFile in conflictingFiles)
                 {
-                    if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null)
+                    if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && (conflictingFile.ContentType != ContentType.Resource || conflictingFile.Docset.Config.Output.CopyResources))
                     {
                         context.Output.Delete(item.Path, legacy);
                     }
@@ -105,7 +105,7 @@ namespace Microsoft.Docs.Build
             {
                 if (_filesBySiteUrl.TryRemove(file.SiteUrl, out _))
                 {
-                    if (_publishItems.TryRemove(file, out var item) && item.Path != null)
+                    if (_publishItems.TryRemove(file, out var item) && item.Path != null && (file.ContentType != ContentType.Resource || file.Docset.Config.Output.CopyResources))
                     {
                         context.Output.Delete(item.Path, legacy);
                     }
--- a/src/docfx/publish/PublishModelBuilder.cs
+++ b/src/docfx/publish/PublishModelBuilder.cs
@@ -66,7 +66,7 @@ namespace Microsoft.Docs.Build
 
                     foreach (var conflictingFile in files.Keys)
                     {
-                        if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && (conflictingFile.ContentType != ContentType.Resource || conflictingFile.Docset.Config.Output.CopyResources))
+                        if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && IsDeletableFile(item))
                         {
                             context.Output.Delete(item.Path, legacy);
                         }
@@ -93,7 +93,7 @@ namespace Microsoft.Docs.Build
 
                 foreach (var conflictingFile in conflictingFiles)
                 {
-                    if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && (conflictingFile.ContentType != ContentType.Resource || conflictingFile.Docset.Config.Output.CopyResources))
+                    if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && IsDeletableFile(item))
                     {
                         context.Output.Delete(item.Path, legacy);
                     }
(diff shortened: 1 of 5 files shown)


src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -66,7 +66,7 @@ public bool TryAdd(Document file, PublishItem item)
 
                     foreach (var conflictingFile in files.Keys)
                     {
-                        if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null)
+                        if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && (conflictingFile.ContentType != ContentType.Resource || conflictingFile.Docset.Config.Output.CopyResources))
```
yufeih: check whether `item.Path` is inside output folder

OsmondJiang:
what issue we are going to resolve here? would you please add it to description？

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -93,7 +93,7 @@ public bool TryAdd(Document file, PublishItem item)
 
                 foreach (var conflictingFile in conflictingFiles)
                 {
-                    if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null)
+                    if (_publishItems.TryRemove(conflictingFile, out var item) && item.Path != null && (conflictingFile.ContentType != ContentType.Resource || conflictingFile.Docset.Config.Output.CopyResources))
```
yufeih: @928PJY , can you change the test framework to assert that tests shouldn't change any files in source folder? You can simplify compare the last modified time of each file in the source repo.

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -453,5 +458,18 @@ private static void VerifyLogsHasLineInfo(string[] logs)
...
+        private static void VerifyNoChangesOnInputFiles(FileInfo[] inputFiles, string[] skippableInputs)
+        {
+            foreach(var file in inputFiles)
+            {
+                if (skippableInputs.Contains(file.Name))
+                {
+                    continue;
+                }
+                Assert.True(File.Exists(file.FullName), $"Input file {file.Name} has been deleted");
+                Assert.True(file.LastWriteTime == new FileInfo(file.FullName).LastWriteTime, $"Input file {file.Name} has been changed");
```
live1206: Seems like the `LastWriteTime` assertion is always true?

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -126,5 +126,10 @@ public bool TryAdd(Document file, PublishItem item)
 
             return (model, fileManifests);
         }
+
+        private bool IsDeletableFile(PublishItem item)
+        {
+            return !item.Path.StartsWith("..");
```
live1206: Is `item.Path` always a relative path?

live1206:
add a test case

test/docfx.Test/e2e/E2ETest.cs
928PJY: No, the `FileInfo` will not change once it has been created

---
In reply to: [284090364](https://github.com/dotnet/docfx/pull/4518#discussion_r284090364) [](ancestors = 284090364)

src/docfx/publish/PublishModelBuilder.cs
928PJY: yes, related to manifest file

---
In reply to: [284090685](https://github.com/dotnet/docfx/pull/4518#discussion_r284090685) [](ancestors = 284090685)

yufeih: If the output path is in another windows directory, `item.Path` will look like `D:/a/b/c`, thus this check will fail. There is a `PathUtility.Match` method that tries to do the same thing, use that method and update the implementation?

test/docfx.Test/e2e/E2ETest.cs
yufeih: @928PJY , better not relying on this implementation details, suggest using a `Dictionary<string, DateTime>`

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -453,5 +458,18 @@ private static void VerifyLogsHasLineInfo(string[] logs)
...
+
+        private static void VerifyNoChangesOnInputFiles(FileInfo[] inputFiles, string[] skippableInputs)
+        {
+            foreach(var file in inputFiles)
+            {
+                if (skippableInputs.Contains(file.Name))
+                {
+                    continue;
+                }
+                Assert.True(File.Exists(file.FullName), $"Input file {file.Name} has been deleted");
```
yufeih: Also need to assert that we don't generate new files in inputs, so

```csharp
Assert.True(inputs.OrderBy(path).SequentialEquals(outputs.OrderBy(path)))
```

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -126,5 +126,10 @@ public bool TryAdd(Document file, PublishItem item)
 
             return (model, fileManifests);
         }
+
+        private bool IsDeletableFile(PublishItem item)
```
yufeih:
```suggestion
        private bool IsInsideOutputPath(PublishItem item)
```

test/docfx.Test/e2e/E2ETest.cs
928PJY: sure

---
In reply to: [284527224](https://github.com/dotnet/docfx/pull/4518#discussion_r284527224) [](ancestors = 284527224)

test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -453,5 +458,26 @@ private static void VerifyLogsHasLineInfo(string[] logs)
...
+                                .Select(file => file.FullName)
+                                .Except(dir.GetFiles($"_site/*", SearchOption.AllDirectories).Select(file => file.FullName)).ToList();
+            foreach (var (file, lastModifyTime) in inputFiles)
+            {
+                var currentFile = new FileInfo(file);
+                if (skippableInputs.Contains(currentFile.Name))
+                {
+                    continue;
+                }
+                Assert.True(File.Exists(file), $"Input file {currentFile.Name} has been deleted");
```
yufeih: file.Exists
```suggestion
                Assert.True(currentFile.Exists, $"Input file {currentFile.Name} has been deleted");
```

Issue #4493, opened 2019-05-08
azure devops repos are using new format url


OsmondJiang:
support identification of Azure DevOps repo by new format:

https://dev.azure.com/{account}/{project}/_git/{repo} like https://dev.azure.com/ceapex/Engineering/_git/Docs.DocFX.Impact

OsmondJiang:
Now new issue is coming, the auth token current is not working for xxx.visualstudio.com but dev.azure.com

cc @yufeih  @mingwli  @fenxu @herohua 

Check this failed test:

https://ceapex.visualstudio.com/Engineering/_build/results?buildId=54247

OsmondJiang:
Work Items:
  - [ ] Docfx support new URL format
  - [ ] Build service support new URL format(including Configuration Migration @mingwli  @OsmondJiang )
  - [ ] Migration tool(@928PJY )

superyyrrzz:
Also need to support this in v2.

chrisrichards:
Just taking a look at this, does the code in [GitUtility.cs](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Common/Git/GitUtility.cs) need updating as well? e.g:

https://github.com/dotnet/docfx/blob/973fe1ec959aea1805e6d49f96c4ada84fecfe14/src/Microsoft.DocAsCode.Common/Git/GitUtility.cs#L29-L30

and:

https://github.com/dotnet/docfx/blob/973fe1ec959aea1805e6d49f96c4ada84fecfe14/src/Microsoft.DocAsCode.Common/Git/GitUtility.cs#L33

superyyrrzz:
@chrisrichards Actually they have no reference except tests. It can be `Obsolete`.

PR #4519, opened 2019-05-09
🐛remove user token and name from git remote

OsmondJiang:
depends one this PR to add test #4513


src/docfx/docset/Repository.cs
```diff
@@ -46,11 +47,10 @@ public static Repository Create(string path, string branch, string repoUrl = nul
...
 
             var (remote, repoBranch, commit) = GitUtility.GetRepoInfo(repoPath);
-            var gitIndex = remote.IndexOf(".git");
-            if (gitIndex >= 0)
-            {
-                remote = remote.Remove(gitIndex);
-            }
+
+            // remove user name, token and .git from url like https://xxxxx@dev.azure.com/xxxx.git
+            remote = Regex.Replace(repoUrl ?? remote, @"^((http|https):\/\/)?([^\/\s]+@)?([\S]+?)(\.git)?$", "$1$4");
```
yufeih: I think the conclusion is to not do any url normalization in docfx, no?

OsmondJiang: no, we need do it because docfx is not always used by service. And now the account info will always be part of remote when user clone repo from azure devops:

image

yufeih: @live1206 c# `Regex.Replace` also supports `$1`

OsmondJiang:
#4493

OsmondJiang:
> Also test for `https://username:password@a.com
> 👍

Already added

PR #4521, opened 2019-05-10
Implement yaml tags according to spec

yufeih:
Fixes https://dev.azure.com/ceapex/Engineering/_workitems/edit/90583/

### DIFF
--- a/src/docfx/lib/json/YamlUtility.cs
+++ b/src/docfx/lib/json/YamlUtility.cs
@@ -188,10 +188,6 @@ namespace Microsoft.Docs.Build
             {
                 return new JValue(b);
             }
-            if (int.TryParse(value, out var n))
-            {
-                return new JValue(n);
-            }
             if (long.TryParse(value, out var l))
             {
                 return new JValue(l);
--- a/src/docfx/lib/json/YamlUtility.cs
+++ b/src/docfx/lib/json/YamlUtility.cs
@@ -127,24 +127,7 @@ namespace Microsoft.Docs.Build
                 case Scalar scalar:
                     if (scalar.Style == ScalarStyle.Plain)
                     {
-                        if (string.IsNullOrWhiteSpace(scalar.Value) ||
-                            scalar.Value == "~" ||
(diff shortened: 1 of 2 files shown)


src/docfx/lib/json/YamlUtility.cs
```diff
@@ -181,6 +164,58 @@ private static JToken ParseAsJToken(IParser parser, string file, List<Error> err
...
+            {
+                return JValue.CreateNull();
+            }
+            if (bool.TryParse(value, out var b))
+            {
+                return new JValue(b);
+            }
+            if (int.TryParse(value, out var n))
+            {
+                return new JValue(n);
```
OsmondJiang: long is enough?

src/docfx/lib/json/YamlUtility.cs
```diff
@@ -181,6 +164,58 @@ private static JToken ParseAsJToken(IParser parser, string file, List<Error> err
...
+            //    0x[0 - 9a - fA - F] +                           tag:yaml.org,2002:int(Base 16)
+            //    [-+] ? ( \. [0-9]+ | [0-9]+ ( \. [0-9]* )? ) ( [eE][-+]?[0 - 9]+ )?   tag:yaml.org,2002:float (Number)
+            //    [-+]? ( \.inf | \.Inf | \.INF )                 tag:yaml.org,2002:float (Infinity)
+            //    \.nan | \.NaN | \.NAN                           tag:yaml.org,2002:float (Not a number)
+            //    *                                               tag:yaml.org,2002:str(Default)
+            if (string.IsNullOrEmpty(value) || value == "~" || value.Equals("null", StringComparison.OrdinalIgnoreCase))
+            {
+                return JValue.CreateNull();
+            }
+            if (bool.TryParse(value, out var b))
```
OsmondJiang: what about 1/0 for bool? seems you changed the order?

yufeih: 1, 0 is not a valid bool, `bool.TryParse("0")` returns `false`

Issue #4458, opened 2019-04-30
NRE in dependency slot pool


yufeih:
# docfx crash report: System.AggregateException

docfx: `1.0.0`
x64: `True`
cmd: `D:\docfx\src\docfx\bin\Release\netcoreapp2.2\docfx.dll restore D:\docfx-impact\source\azure-docs-pr --verbose`
cwd: `D:\docfx\src\docfx`
git: `git version 2.21.0.windows.1`

## callstack

```
System.AggregateException: One or more errors occurred. (One or more errors occurred. (Object reference not set to an instance of an object.)) ---> System.AggregateException: One or more errors occurred. (Object reference not set to an instance of an object.) ---> System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.Docs.Build.DependencySlotPool`1.<>c__DisplayClass3_0.<AcquireSlot>b__0() in D:\docfx\src\docfx\restore\index\DependencySlotPool.cs:line 138
   at Microsoft.Docs.Build.ProcessUtility.RunInsideMutex(String mutexName, Action action) in D:\docfx\src\docfx\lib\ProcessUtility.cs:line 314
   at Microsoft.Docs.Build.DependencySlotPool`1.AcquireSlot(String url, LockType type, Func`2 updateExistingSlot, Func`2 matchExistingSlot) in D:\docfx\src\docfx\restore\index\DependencySlotPool.cs:line 102
   at Microsoft.Docs.Build.RestoreMap.AcquireGit(String remote, String branch, String commit, LockType type) in D:\docfx\src\docfx\restore\RestoreMap.cs:line 247
   at Microsoft.Docs.Build.RestoreMap.AcquireExclusiveGit(String remote, String branch, String commit) in D:\docfx\src\docfx\restore\RestoreMap.cs:line 231
   at Microsoft.Docs.Build.RestoreGit.<>c__DisplayClass1_2.<Restore>b__8(String branch) in D:\docfx\src\docfx\restore\RestoreGit.cs:line 145
   at Microsoft.Docs.Build.ParallelUtility.<>c__DisplayClass1_0`1.<ForEach>b__0(T item) in D:\docfx\src\docfx\lib\ParallelUtility.cs:line 31
   at System.Threading.Tasks.Parallel.<>c__DisplayClass44_0`2.<PartitionerForEachWorker>b__1(IEnumerator& partitionState, Int32 timeout, Boolean& replicationDelegateYieldedBeforeCompletion)
[...]

PR #4523, opened 2019-05-10
🐛force acquire new dependency slots

OsmondJiang:
force acquire new dependency slot to avoid out-of-sync problem.
#4458


OsmondJiang:
> just to prevent accidental merge before I review this

It would be great if you can take 5 minutes to take a look...🤷

Issue #4443, opened 2019-04-29
restore crashed due to git worktree and index.json out of sync


yufeih:
Exception (first time):

```
Using docfx 3.0.0-beta-00833-20832f903
Using existing repository 'C:\Users\yufeih\AppData\Local\Temp\appdata\git2\github.com+Microsoft+template..ocs.msft+95403dd1\.git' for 'https://github.com/Microsoft/templates.docs.msft'
remote: Enumerating objects: 90, done.
remote: Counting objects: 100% (83/83), done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 44 (delta 26), reused 25 (delta 14), pack-reused 0
Unpacking objects: 100% (44/44), done.
From https://github.com/Microsoft/templates.docs.msft
 + 1825ec8...f615a60 master     -> master  (forced update)
fatal: 'C:/Users/yufeih/AppData/Local/Temp/appdata/git2/github.com+Microsoft+template..ocs.msft+95403dd1/1' already exists
Preparing worktree (detached HEAD f615a60)
Restore dependencies done in 5.57s
Microsoft.Docs.Build.DocfxException: Cloning git repository 'https://github.com/Microsoft/templates.docs.msft' ('master') failed. ---> System.InvalidOperationException: '"git" -c core.longpaths=true worktree add C:/Users/yufeih/AppData/Local/Temp/appdata/git2/github.com+Microsoft+template..ocs.msft+95403dd1/1 f615a60861b2393d84323892167657235cb1f5d1 --force' failed in directory 'C:\Users\yufeih\AppData\Local\Temp\appdata\git2\github.com+Microsoft+template..ocs.msft+95403dd1\.git' with exit code 128:
STDOUT:''
   at Microsoft.Docs.Build.ProcessUtility.Execute(String fileName, String commandLineArgs, String cwd, Boolean stdout) in D:\work\1\s\src\docfx\lib\ProcessUtility.cs:line 202
   at Microsoft.Docs.Build.GitUtility.Execute(String cwd, String commandLineArgs, Boolean stdout) in D:\work\1\s\src\docfx\lib\git\GitUtility.cs:line 346
   at Microsoft.Docs.Build.GitUtility.AddWorkTree(String cwd, String committish, String path) in D:\work\1\s\src\docfx\lib\git\GitUtility.cs:line 181
[...]

PR #4526, opened 2019-05-10
👨‍🏭 only trust slot status for restored work tree

OsmondJiang:
#4443 

Test has already been added


src/docfx/restore/RestoreGit.cs
```diff
@@ -130,24 +129,29 @@ void AddWorkTrees()
 
                             try
                             {
-                                if (existingWorkTreeFolders.TryAdd(workTreePath))
+                                if (gitSlot.Restored && Directory.Exists(workTreePath))
+                                {
+                                    // re-use existing work tree
+                                    // checkout to {headCommit}, no need to fetch
+                                    Debug.Assert(!GitUtility.IsDirty(workTreePath));
```
yufeih: remove this `Debug.Assert` since it only runs in debug mode

OsmondJiang: The `Debug.Assert` purpose is to let us find problem in Debug model, that's what I want to do.

Issue #4444, opened 2019-04-29
restore does not gracefully handle force push


yufeih:
If a dependent git repo is force pushed, `docfx restore` currently does not handle unknown commit gracefully.

OsmondJiang:
@fenxu , what you think the behavior like for PR building with dependency lock?

fenxu:
It's a corner case IMO. The error will happen only when
1. User do a force push
2. Before branch build finishing, user create a PR.

There's workaround, that's close and reopen the PR after target branch build finished.

I think we can fix that but not so high priority because usually partners doesn't allow user to do force push.

cc @yufeih @OsmondJiang

fenxu:
About fix, I don't think we have ability to detect whether a force push is happened. So that we don't know whether we should respect the cache or not.

Maybe we can improve error message to let uesr close/reopen the PR after target branch finishing build.

PR #4534, opened 2019-05-13
👨‍🏭improve error message for wrong version in dependency lock

OsmondJiang:
#4444

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -268,6 +268,31 @@ outputs:
        "!downloads": null
     }
 ---
+# restore with wrong dependency version
+repos:
+  https://docs.com/wrong/a#master:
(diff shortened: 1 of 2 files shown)


docs/specs/restore.yml
```diff
@@ -268,6 +268,31 @@ outputs:
...
+                   }
+                }
+             }
+          }
+  https://docs.com/wrong/b:
+    - files:
+        docs/dep1.md:
+outputs:
+  .errors.log: |
+    ["error","committish-not-found","Can't find branch, tag, or commit 'missing-commit-id' for repo https://docs.com/wrong/b."]
```
OsmondJiang: this error code and error message should be reviewed, @mingwli  @herohua

PR #4536, opened 2019-05-14
Ensures updated_at fallbacks to now when updateTimeAsCommitBuildTime is true

yufeih:
Currently when `updateTimeAsCommitBuildTime` is true, the first time a repo is build, `update_at` fallbacks to commit time, it should fallback to now.

### DIFF
--- a/src/docfx/lib/git/CommitBuildTimeProvider.cs
+++ b/src/docfx/lib/git/CommitBuildTimeProvider.cs
@@ -35,7 +35,6 @@ namespace Microsoft.Docs.Build
             if (!_buildTimeByCommit.TryGetValue(commitId, out time))
             {
                 time = _buildTime;
-                return false;
             }
             return true;
         }
--- a/src/docfx/lib/git/CommitBuildTimeProvider.cs
+++ b/src/docfx/lib/git/CommitBuildTimeProvider.cs
@@ -10,6 +10,7 @@ namespace Microsoft.Docs.Build
 {
     internal class CommitBuildTimeProvider
     {
+        private readonly DateTime _buildTime = DateTime.UtcNow;
         private readonly Repository _repo;
         private readonly string _commitBuildTimePath;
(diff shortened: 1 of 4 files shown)


src/docfx/lib/git/CommitBuildTimeProvider.cs
```diff
@@ -31,7 +32,12 @@ public CommitBuildTimeProvider(Repository repo)
 
         public bool TryGetCommitBuildTime(string commitId, out DateTime time)
         {
-            return _buildTimeByCommit.TryGetValue(commitId, out time);
+            if (!_buildTimeByCommit.TryGetValue(commitId, out time))
+            {
+                time = _buildTime;
```
live1206: why setting `time` here, seems it will only be used when `commitId` found in `_buildTimeByCommit`:
https://github.com/dotnet/docfx/blob/56c5fa22bc9122e039c78f05d7fa58d302cfeec5/src/docfx/build/metadata/ContributionProvider.cs#L147

yufeih: it should always return false

src/docfx/lib/git/CommitBuildTimeProvider.cs
```diff
@@ -31,7 +32,11 @@ public CommitBuildTimeProvider(Repository repo)
 
         public bool TryGetCommitBuildTime(string commitId, out DateTime time)
         {
-            return _buildTimeByCommit.TryGetValue(commitId, out time);
+            if (!_buildTimeByCommit.TryGetValue(commitId, out time))
```
OsmondJiang: can we use GetOrAdd here, then `UpdateAndSaveCache` don't need retrieve commits again?

yufeih: `UpdateAndSaveCache` have to set the time for commits that are not retrieved using this method, so GetOrAdd here does not solve the problem.

PR #4537, opened 2019-05-14
 Only call github once if we don't have access to the repo

yufeih:
The current implementation will call github for each commit when we know for sure that the repo does not exists or we don't have access to the repo. This is stupid, we should only call it once.

### DIFF
--- a/src/docfx/lib/github/GitHubAccessor.cs
+++ b/src/docfx/lib/github/GitHubAccessor.cs
@@ -2,8 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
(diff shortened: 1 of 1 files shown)


src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -13,8 +13,9 @@ namespace Microsoft.Docs.Build
     internal class GitHubAccessor
     {
         private readonly GitHubClient _client;
+        private readonly ConcurrentHashSet<(string owner, string name)> _unknownRepos = new ConcurrentHashSet<(string owner, string name)>();
 
-        private static volatile Error _rateLimitError;
```
superyyrrzz: why remove static?

Issue #4542, opened 2019-05-14
RestoreGitWorkTrees failed occasionally.


yufeih:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=55061

Building files: 100% (1/1), 00:00:05 
Failed   Microsoft.Docs.Build.RestoreTest.RestoreGitWorkTrees
Error Message:
 Assert.Equal() Failure
Expected: 5
Actual:   4
Stack Trace:
   at Microsoft.Docs.Build.RestoreTest.RestoreGitWorkTrees() in D:\a\1\s\test\docfx.Test\restore\RestoreTest.cs:line 117

OsmondJiang:
Here is the detail log:

```
Microsoft.Docs.Build.DocfxException: Cloning git repository 'https://github.com/docascode/docfx-test-dependencies-clean' ('test-4-clean') failed. ---> System.InvalidOperationException: '"git" -c core.longpaths=true worktree add D:/a/1/s/test/docfx.Test/bin/Debug/netcoreapp2.2/appdata/git2/github.com+docascode+docfx-te..es-clean+0336104e/5 01fa8106eb0b1de66fa75fefe49f1ce35cb239de --force' failed in directory 'D:\a\1\s\test\docfx.Test\bin\Debug\netcoreapp2.2\appdata\git2\github.com+docascode+docfx-te..es-clean+0336104e\.git' with exit code 128: 
STDOUT:''
   at Microsoft.Docs.Build.ProcessUtility.Execute(String fileName, String commandLineArgs, String cwd, Boolean stdout)
   at Microsoft.Docs.Build.GitUtility.Execute(String cwd, String commandLineArgs, Boolean stdout)
   at Microsoft.Docs.Build.GitUtility.ExecuteNonQuery(String cwd, String commandLineArgs)
   at Microsoft.Docs.Build.GitUtility.AddWorkTree(String cwd, String committish, String path)
   at Microsoft.Docs.Build.RestoreGit.<>c__DisplayClass1_1.<Restore>b__8(String branch)
   --- End of inner exception stack trace ---
   at Microsoft.Docs.Build.RestoreGit.<>c__DisplayClass1_1.<Restore>b__8(String branch)
   at Microsoft.Docs.Build.ParallelUtility.<>c__DisplayClass1_0`1.<ForEach>b__0(T item)
[...]

PR #4546, opened 2019-05-14
clean work tree folder before using

OsmondJiang:
#4542 

This PR is add more checking to the test, make sure we know the root cause if it happened again.
And also found some business logic bug. The newly added checks will cover these part changes

### DIFF
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -143,6 +143,9 @@ namespace Microsoft.Docs.Build
                                         // it may be dirty caused by last failed restore action
                                         if (Directory.Exists(workTreePath))
                                             Directory.Delete(workTreePath, true);
+
+                                        // https://stackoverflow.com/questions/24265481/after-directory-delete-the-directory-exists-returning-true-sometimes
+                                        Debug.Assert(!Directory.Exists(workTreePath));
                                         GitUtility.AddWorkTree(repoPath, headCommit, workTreePath);
                                     }
                                     catch (Exception ex)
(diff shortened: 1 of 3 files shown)


src/docfx/restore/index/DependencySlotPool.cs
```diff
@@ -111,7 +111,7 @@ public static (string path, T slot) AcquireSlot(string url, LockType type, Func<
                             var existed = false;
                             foreach (var i in slots)
                             {
-                                if (DateTime.UtcNow - i.LastAccessDate > TimeSpan.FromSeconds(_defaultLockdownTimeInSecond))
+                                if (!i.Restored || DateTime.UtcNow - i.LastAccessDate > TimeSpan.FromSeconds(_defaultLockdownTimeInSecond))
```
mingwli: >!i.Restored [](start = 36, length = 11)

the slot maybe right being used by another docfx-restore?

src/docfx/restore/RestoreGit.cs
```diff
@@ -143,6 +143,9 @@ void AddWorkTrees()
                                         // it may be dirty caused by last failed restore action
                                         if (Directory.Exists(workTreePath))
                                             Directory.Delete(workTreePath, true);
+
+                                        // https://stackoverflow.com/questions/24265481/after-directory-delete-the-directory-exists-returning-true-sometimes
+                                        Debug.Assert(!Directory.Exists(workTreePath));
```
live1206: This only fails on debug, shall we log an exception for the normal run as well?

src/docfx/restore/index/DependencySlotPool.cs
OsmondJiang: no, one index file can be only access by one consumer at the same time

src/docfx/restore/RestoreGit.cs
OsmondJiang: this is only to identify the root cause, once we know it's the root cause, we will something here.

Issue #4552, opened 2019-05-15
Error/warning/info/suggrestion count telemetry should be calculated correctly


OsmondJiang:
Even they reach the max count(1000), we should calculate the count telemetry correctly per error code.

cc @herohua  @yufeih

yufeih:
where does the requirement came from?

OsmondJiang:
@yufeih, As we talked offline, either we break the build when reach the max error count or we should the correct error count to user.

OsmondJiang:
The conclusion is that once the error code reach the max count, the telemetry actually is not really useful, so let's keep it untouched

PR #4554, opened 2019-05-15
break build once docfx reach the max count of error/warning/suggestion count

OsmondJiang:
#4552


OsmondJiang:
Then we will set a large number to `max error count` for docs building. @yufeih  @herohua  @live1206

yufeih:
Let's discuss it with the team before merge

PR #4560, opened 2019-05-16
fix output path in legacy filemap

928PJY:
https://dev.azure.com/ceapex/Engineering/_git/Docs.DocFX.Impact/pullrequest/11660?_a=files&path=%2Fsource%2Fedge-developer.output%2F.publish.json

### DIFF
--- a/src/docfx/lib/PathUtility.cs
+++ b/src/docfx/lib/PathUtility.cs
@@ -205,6 +205,24 @@ namespace Microsoft.Docs.Build
             return result.ToString();
         }
 
+        public static string ChangeExtension(string filePath, string extension, string[] acceptableExtension = null)
+        {
+            acceptableExtension = acceptableExtension ?? new string[] { "raw.page.json", "mta.json" };
+            if (!acceptableExtension.Any(ext =>
+             {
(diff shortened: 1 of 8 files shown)


live1206:
Could you add some description of what problem has been fixed in this PR and add a test for it?

src/docfx/lib/PathUtility.cs
```diff
@@ -205,6 +205,24 @@ public static string UrlToShortName(string url)
             return result.ToString();
         }
 
+        public static string ChangeExtension(string filePath, string extension, string[] acceptableExtension = null)
+        {
+            acceptableExtension = acceptableExtension ?? new string[] { "raw.page.json", "mta.json" };
```
yufeih: `raw.page.json` and `mta.json` this is for legacy only (and will be removed shortly once docs-publish) is done, so I suggest to not extract this shared method and don't change code for non-legacy files for this fix

928PJY: so where should we put the shared method used in legacy code

---
In reply to: [284538603](https://github.com/dotnet/docfx/pull/4560#discussion_r284538603) [](ancestors = 284538603)

PR #4566, opened 2019-05-16
Wait until dir is really deleted

OsmondJiang:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=55603

### DIFF
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -145,9 +145,9 @@ namespace Microsoft.Docs.Build
                                         if (Directory.Exists(workTreePath))
                                         {
                                             // https://stackoverflow.com/questions/24265481/after-directory-delete-the-directory-exists-returning-true-sometimes
-                                            var dir = new DirectoryInfo(workTreePath);
-                                            dir.Delete(true);
-                                            dir.Refresh();
+                                            Directory.Delete(workTreePath, true);
+                                            while (Directory.Exists(workTreePath))
+                                                Thread.Sleep(1);
                                         }
 
                                         Debug.Assert(!Directory.Exists(workTreePath));
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Microsoft.Docs.Build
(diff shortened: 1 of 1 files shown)


src/docfx/restore/RestoreGit.cs
```diff
@@ -142,9 +143,13 @@ void AddWorkTrees()
                                         // clean existing work tree folder
                                         // it may be dirty caused by last failed restore action
                                         if (Directory.Exists(workTreePath))
+                                        {
+                                            // https://stackoverflow.com/questions/24265481/after-directory-delete-the-directory-exists-returning-true-sometimes
                                             Directory.Delete(workTreePath, true);
+                                            while (Directory.Exists(workTreePath))
+                                                Thread.Sleep(1);
```
yufeih: this can hang forever if some other process holds a lock.

OsmondJiang: you are right, but you can see that work tree path is hidden from end user, controlled by docfx, so I don't think it's a really big problem, but I can add a timeout here or find another way to make sure it's deleted.

OsmondJiang: @yufeih  and @herohua , no actually, because it the work tree path is using by other process, the Directory.Delete will throw IO exceptions

image

OsmondJiang: that code actually is just waiting the directory handler to be close after being deleted.

yufeih: If we do need to wait, can there be a timeout, or a console log (like `RunInsideMutex`) to indicate that we are waiting for directory delete to complete.

OsmondJiang: good suggestion for console log point, i will add one

src/docfx/restore/RestoreGit.cs
```diff
@@ -147,7 +147,10 @@ void AddWorkTrees()
                                             // https://stackoverflow.com/questions/24265481/after-directory-delete-the-directory-exists-returning-true-sometimes
                                             Directory.Delete(workTreePath, true);
                                             while (Directory.Exists(workTreePath))
+                                            {
+                                                Log.Write($"Wait for the {workTreePath} to be deleted");
                                                 Thread.Sleep(1);
```
yufeih: this is going to generate a zillion logs, consider writing a log every 30 seconds

yufeih: also could just use `Console.WriteLine` to show this message when `--verbose` is not specified

OsmondJiang: it's not going to generate a zillion logs because the directly deleting delay is really short. And i want it to be showed when --verbose for debugging only.

Issue #4567, opened 2019-05-16
[SDP] schema validation rules


mingwli:
- [x] Json Schema Validation For SDP
  - [x] type
  - [x] required
  - [x] ~additionProperties~
  - [x] additionalProperties
  - [x] ~~calls [e.g.](https://ceapex.visualstudio.com/Engineering/_git/docs-ui?path=%2Fpackages%2Fschemas%2FPowerConnector.schema.json&version=GBdevelop&line=161&lineStyle=plain&lineEnd=162&lineStartColumn=1&lineEndColumn=1) only one. should be a mistake.~~
  - [x] definitions
  - [x] '$ref'
  - [x] enum
  - [x] ~footer~
  - [x] format [e.g.](https://ceapex.visualstudio.com/Engineering/_git/docs-ui?path=%2Fpackages%2Fschemas%2FExamination.schema.json&version=GBdevelop&line=162&lineStyle=plain&lineEnd=163&lineStartColumn=1&lineEndColumn=1) only one
  - [x] items
  - [x] maxItems (item & length both are used on array)
  - [x] maxLength
  - [x] ~~metadata/pointer~~
  - [x] minItems
  - [x] minLength
  - [x] ~sections~
  - [x] ~~tags~~

#### Scanned result here:
`forTypes` stands for what type of properties the rule is applied
`valueType` stands for the type of the value of the rule in JavaScript
[rules.json.txt](https://github.com/dotnet/docfx/files/3186430/rules.json.txt)

OsmondJiang:
`ref` and `definitions` need to be supported.

cc @yufeih  @herohua

OsmondJiang:
Add some example link to these items, @mingwli

OsmondJiang:
- [ ] Json Schema Validation For SDP
  - ~~[ ] additionProperties~~
  - [ ] additionalProperties
  - [ ] calls
  - [ ] definitions
  - [ ] '$ref'
  - [x] enum
  - [ ] footer
  - [ ] format
  - [x] items
  - [ ] maxItems
  - [ ] maxLength
  - [x] ~~metadata/pointer~~
  - [ ] minItems
  - [ ] minLength
  - [ ] sections
  - [x] ~~tags~~
- [ ] Json Schema Transform
  - [ ] ContentType
  - [ ] XrefProperties
- [ ] Remove Razor rendering, using template rendering
- [ ] Remove Strong Model
  - [ ] Remove Strong Model Kinds of Attributes
[...]

PR #4585, opened 2019-05-20
👨‍🏭 re-implement SDP pipeline(validation and transform) using JSON schema

OsmondJiang:
1. add some json schema file back
2. implement schema transform for `SDP`
3. integration schema validation and transform with `BuildPage`
4. fix some tests

todo: 
Remove validation on `JsonUtility.ToObject`
Remove strong model class for `landingData` and `testData/Page`
implement `TransformXref` part
support `$ref` and `definitions` in json schema


live1206:
we should remove strong type schema then since JSON schema is in use

OsmondJiang:
> we should remove strong type schema then since JSON schema is in use

Check to description, added it in TODO list

src/docfx/data/ContextObject.json
```diff
@@ -0,0 +1,39 @@
+{
```
superyyrrzz: These schemas shall go to Microsoft.Docs.Template.csproj

src/docfx/build/page/BuildPage.cs
```diff
@@ -178,14 +178,25 @@ private static (List<Error> errors, Schema schema, OutputModel model)
             LoadSchemaDocument(Context context, List<Error> errors, JToken token, Document file, Action<Document> buildChild)
         {
             var obj = token as JObject;
-            if (file.Schema is null)
+            if (file.Schema is null || !File.Exists($"data/{file.Schema.Type.Name}.json"))
```
superyyrrzz: This looks like a hack that relying on that you have not removed strong model schemas. Will your TODO "remove strong model schemas" include refining this?

src/docfx/data/ContextObject.json
OsmondJiang: Yes, will integration with Template later, TODO already added

src/docfx/build/page/BuildPage.cs
```diff
@@ -178,14 +178,25 @@ private static (List<Error> errors, Schema schema, OutputModel model)
...
             {
                 throw Errors.SchemaNotFound(file.Mime).ToException();
             }
 
-            // todo: why not directly use strong model here?
-            var (schemaViolationErrors, content) = JsonUtility.ToObject(token, file.Schema.Type, transform: AttributeTransformer.TransformSDP(context, file, buildChild));
-            errors.AddRange(schemaViolationErrors);
+            // TODO: load json schema from template
+            var schemaFilePath = $"data/{file.Schema.Type.Name}.json";
+            var jsonSchema = JsonUtility.Deserialize<JsonSchema>(File.ReadAllText(schemaFilePath), schemaFilePath);
```
superyyrrzz: Load schema only once, instead of per file.

src/docfx/build/page/BuildPage.cs
OsmondJiang: Currently validation and transform are not relying on strong model.  The only reason to keep strong model is the Razor part.

src/docfx/build/page/BuildPage.cs
OsmondJiang: good suggestion

src/docfx/lib/schema/JsonSchemaTransform.cs
```diff
@@ -0,0 +1,119 @@
...
+            {
+                case JsonSchemaContentType.Href:
+                    var (error, link, _) = dependencyResolver.ResolveLink(content, file, file, buildChild);
+                    errors.AddIfNotNull(error);
+                    content = new SourceInfo<string>(link, content);
+                    break;
+
+                case JsonSchemaContentType.Markdown:
+                    var (markupErrors, html) = MarkdownUtility.ToHtml(
+                    content,
```
superyyrrzz: style: indent parameters

docs/specs/schema.yml
```diff
@@ -154,6 +155,7 @@ outputs:
     { "content": {} }
   .errors.log: |
     ["info","null-value","'description' contains null value","docs/a.yml",2,1]
+    ["warning","unexpected-type","Expect type 'String' but got 'Null'","docs/a.yml",2,13]
```
superyyrrzz: duplicate with above `null-value`. Should log only one message. Looks like a bug in schema validation.

OsmondJiang: yes, already opened a issue for that: #4591

src/docfx/docfx.csproj
```diff
@@ -41,4 +41,19 @@
     <ProjectReference Include="../Microsoft.Docs.Template/Microsoft.Docs.Template.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
yufeih: remove this, it is already copied

src/docfx/lib/schema/JsonSchemaContentType.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Docs.Build
+{
+    public enum JsonSchemaContentType
+    {
+        None,
+        Href, // resolve link
```
yufeih: prefer ///<summary> comment

src/docfx/lib/schema/JsonSchemaProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.IO;
+
+namespace Microsoft.Docs.Build
+{
+    internal class JsonSchemaProvider
+    {
+        private ConcurrentDictionary<string, JsonSchema> _jsonSchemas = new ConcurrentDictionary<string, JsonSchema>();
+
+        // TODO: get schema from template
+        public JsonSchema GetJsonSchema(Schema schema)
```
yufeih: move this method to `TemplateEngine`, there is no need to create a new type and pass it down

src/docfx/lib/schema/JsonSchemaTransform.cs
```diff
@@ -0,0 +1,119 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    internal static class JsonSchemaTransform
+    {
+        public static (List<Error> errors, JToken token) Transform(Document file, Context context, JsonSchema schema, JToken token, Action<Document> buildChild)
+        {
+            var errors = new List<Error>();
+
+            var transformedToken = JsonUtility.DeepClone(token);
```
yufeih: we could do an in place transform for perf reasons without this deep clone, if `JToken` is removed from output, it is clear from method signature that it is an in place transform.

src/docfx/lib/schema/JsonSchemaTransform.cs
yufeih: still not the right indent

src/docfx/lib/schema/JsonSchemaTransform.cs
```diff
@@ -0,0 +1,119 @@
...
+            var dependencyResolver = file.Schema.Type == typeof(LandingData) ? context.LandingPageDependencyResolver : context.DependencyResolver;
+            var sourceInfo = JsonUtility.GetSourceInfo(value);
+            var content = new SourceInfo<string>(value.Value<string>(), sourceInfo);
+
+            switch (schema.ContentType)
+            {
+                case JsonSchemaContentType.Href:
+                    var (error, link, _) = dependencyResolver.ResolveLink(content, file, file, buildChild);
+                    errors.AddIfNotNull(error);
+                    content = new SourceInfo<string>(link, content);
```
yufeih: Can directly return here, source info will not be useful afterwards (some transforms like xref does not return a simple string)

```suggestion
                    return link;
```

OsmondJiang:
#4567

src/docfx/lib/schema/JsonSchemaTransform.cs
OsmondJiang: sounds good

src/docfx/lib/schema/JsonSchemaTransform.cs
OsmondJiang: i would like to keep the original source info in case we use it somewhere someday.

src/docfx/docfx.csproj
OsmondJiang: what do you mean it's already copied?

yufeih: My mistake, I thought we copied everything under `data` link we did for test project, so use:
```suggestion
   <None Include="data/**/*" CopyToOutputDirectory="PreserveNewest" />
```

src/docfx/lib/schema/JsonSchemaProvider.cs
OsmondJiang: context.Template could be null in our test pipeline, so do you mean put a static method in Template?

src/docfx/lib/schema/JsonSchemaTransform.cs
```diff
@@ -0,0 +1,118 @@
...
+                        MarkdownPipelineType.InlineMarkdown);
+
+                    errors.AddRange(inlineMarkupErrors);
+                    content = new SourceInfo<string>(inlineHtml, content);
+                    break;
+
+                case JsonSchemaContentType.Html:
+                    var htmlWithLinks = HtmlUtility.TransformLinks(content, (href, _) =>
+                    {
+                        var (e, l, _) = dependencyResolver.ResolveLink(new SourceInfo<string>(href, content), file, file, buildChild);
```
live1206: For naming consistency, it would be better to have `error` and `link` here

src/docfx/lib/schema/JsonSchemaTransform.cs
```diff
@@ -0,0 +1,118 @@
...
+            Transform(file, context, schema, token, errors, buildChild);
+            return (errors, token);
+        }
+
+        private static void Transform(Document file, Context context, JsonSchema schema, JToken token, List<Error> errors, Action<Document> buildChild)
+        {
+            switch (token)
+            {
+                case JValue scalar:
+                    token.Replace(TransformScalar(file, context, schema, scalar, errors, buildChild));
```
live1206: this will just replace the value and keep the line info? can we do this in other places as well?

src/docfx/template/TemplateEngine.cs
```diff
@@ -40,6 +42,20 @@ private TemplateEngine(string templateDir, JsonSchema metadataSchema)
...
         }
 
+        public static JsonSchema GetJsonSchema(Schema schema)
+        {
+            if (schema == null)
+            {
+                return null;
+            }
+
+            // TODO: get schema from template
```
live1206: I guess test schemas should not go to Template?

src/docfx/lib/schema/JsonSchemaTransform.cs
OsmondJiang: TransformScalar return an new JToken with line info, so line info will be there

src/docfx/template/TemplateEngine.cs
OsmondJiang: At that time, we should always have a test template for test purpose, just like legacy spec test do.

Issue #4450, opened 2019-04-30
Legacy cleanup tasks after v3 migration


yufeih:
Some of our content output has different behaviors depending on `--legacy` flag, but we are only testing against non legacy behavior. This issue tracks the works we've done to remove the behavior differences and test against production behavior. We can ship our ideal output as improvements after content migration is done.

OsmondJiang:
dynamic resource scope -> static resource scope

PR #4587, opened 2019-05-20
Consolidate HtmlPostProcess

yufeih:
- Let SDP use HtmlPostProcess for `[html]`, make details of `HtmlPostProcess` private
- Always add locale for absolute path (removed a legacy switch)
- Adjust tests

#4078 #4450

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using System.Globalization;
-using System.Linq;
 using System.Text;
 using System.Web;
 using HtmlAgilityPack;
@@ -16,25 +15,27 @@ namespace Microsoft.Docs.Build
         private static readonly Func<HtmlAgilityPack.HtmlAttribute, int> s_getValueStartIndex =
             ReflectionUtility.CreateInstanceFieldGetter<HtmlAgilityPack.HtmlAttribute, int>("_valuestartindex");
 
+        public static string HtmlPostProcess(string html, CultureInfo culture) => HtmlPostProcess(LoadHtml(html), culture);
+
+        public static string HtmlPostProcess(HtmlNode html, CultureInfo culture)
+        {
+            html = html.StripTags();
+            html = html.AddLinkType(culture.Name.ToLowerInvariant())
+                       .RemoveRerunCodepenIframes();
+
+            if (string.IsNullOrWhiteSpace(html.OuterHtml))
+            {
+                return "<div></div>";
+            }
+
+            return LocalizationUtility.AddLeftToRightMarker(culture, html.OuterHtml);
+        }
+
         public static HtmlNode LoadHtml(string html)
         {
             var doc = new HtmlDocument();
             doc.LoadHtml(html);
             return doc.DocumentNode;
-        }
-
-        public static string TransformHtml(string html, Func<HtmlNode, HtmlNode> transform)
-        {
-            HtmlDocument document = new HtmlDocument();
-            document.LoadHtml(html);
-            return transform(document.DocumentNode).OuterHtml;
-        }
-
-        public static HtmlNode AddLinkType(this HtmlNode html, string locale)
-        {
-            AddLinkType(html, "a", "href", locale);
-            AddLinkType(html, "img", "src", locale);
-            return html;
         }
 
         public static long CountWord(this HtmlNode node)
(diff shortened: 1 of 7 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -16,25 +15,27 @@ internal static class HtmlUtility
...
-        public static HtmlNode LoadHtml(string html)
-        {
-            var doc = new HtmlDocument();
-            doc.LoadHtml(html);
-            return doc.DocumentNode;
-        }
+        public static string HtmlPostProcess(string html, CultureInfo culture) => HtmlPostProcess(LoadHtml(html), culture);
 
-        public static string TransformHtml(string html, Func<HtmlNode, HtmlNode> transform)
+        public static string HtmlPostProcess(HtmlNode html, CultureInfo culture)
```
OsmondJiang: What's the meaning of PostProcess for an utility?  Seems it's belongs to BuildPage html post process, not HtmlUtility.
We should name it using functional description for utility method, right？

OsmondJiang: Suggestion to move this method to BuildPage

test/docfx.Test/lib/HtmlUtilityTest.cs
```diff
@@ -20,33 +23,21 @@ public class HtmlUtilityTest
...
         [InlineData("<a href='https://[abc]' />", "<a href='https://[abc]' data-linktype='external' />")]
-        public void AddLinkType(string input, string output)
-        {
-            var actual = HtmlUtility.TransformHtml(input, node => node.AddLinkType("zh-cn"));
-
-            Assert.Equal(TestUtility.NormalizeHtml(output), TestUtility.NormalizeHtml(actual));
-        }
 
-        [Theory]
+        // Rerun codepen iframes
```
OsmondJiang: suggest to keep them spited.

src/docfx/lib/HtmlUtility.cs
yufeih: It will be shared with conceptual and html/markdown format processing (attribute/schematransform) in SDP, thus extracted to HtmlUtil. The named was picked and agreed with the template team to replace logics in conceptual jint, in case they want to apply some post processing inside build pipeline after we generate HTML.

OsmondJiang: this name has no difference with `DoSomeThings` in HtmlUtility, for me, it's more like a business logic combination.

test/docfx.Test/common/TestUtility.cs
```diff
@@ -104,7 +104,10 @@ void Walk(HtmlNode node, int level)
                         sb.Append(node.Name);
                         foreach (var attr in node.Attributes.OrderBy(a => a.Name))
                         {
-                            sb.Append($" {attr.Name}=\"{TrimWhiteSpace(attr.Value)}\"");
+                            if (attr.Name != "data-linktype")
```
live1206: why not apply for `data-linktype`?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -16,25 +15,27 @@ internal static class HtmlUtility
...
         {
-            HtmlDocument document = new HtmlDocument();
-            document.LoadHtml(html);
-            return transform(document.DocumentNode).OuterHtml;
+            html = html.StripTags();
+            html = html.AddLinkType(culture.Name.ToLowerInvariant())
```
live1206: so the culture name would always be appended to the resolved URL. We have some logic of appending the locale in the compare tool, we can remove them then.

test/docfx.Test/common/TestUtility.cs
yufeih: I not exclude here, that means modifying a lot of test cases, because this is not behind a legacy flag now, and every link will have data-linktype attribute

src/docfx/lib/HtmlUtility.cs
yufeih: It will only append if it is an absolute path (like before), it does not affect the compare tool.

live1206: https://apidrop.visualstudio.com/Content%20CI/_git/docfx-v3-migration?path=%2Fscripts%2Fmigration%2Fcompare%2Fnormalize-html.js&version=GBmaster&line=65&lineStyle=plain&lineEnd=66&lineStartColumn=1&lineEndColumn=1

OsmondJiang:
I still don't think it's a good name in HtmlUtility, it's something like "DoSomeThings" in HtmlUtility, as utility helper, it's can't tell user when and where to use it. Suggest to move these business combination logic into your business code, not utility helper code,

PR #4588, opened 2019-05-20
Use absolute URL for intermediate representation

yufeih:
#4078


yufeih:
The `azure-docs-pr` diff is expected, we've trimmed leading `./` in compare tool so it'll not cause migration failure.

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -41,26 +42,52 @@ internal class DependencyResolver
             return (error, content, child);
         }
 
-        public (Error error, string link, Document file) ResolveLink(SourceInfo<string> path, Document relativeTo, Document resultRelativeTo)
+        public (Error error, string link, Document file) ResolveRelativeLink(SourceInfo<string> path, Document relativeTo, Document resultRelativeTo)
```
```suggestion
        public (Error error, string link, Document file) ResolveRelativeLink(SourceInfo<string> path, Document declaringFile, Document referencingFile)
```

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -41,26 +42,52 @@ internal class DependencyResolver
...
                 _dependencyMapBuilder.AddDependencyItem(relativeTo, file, UrlUtility.FragmentToDependencyType(fragment));
                 _bookmarkValidator.AddBookmarkReference(relativeTo, isSelfBookmark ? resultRelativeTo : file, fragment, isSelfBookmark, path);
             }
 
-            return (error, link, file);
+            return (error, link, linkType, file);
         }
 
-        public (Error error, string href, string display, IXrefSpec spec) ResolveXref(SourceInfo<string> href, Document relativeTo, Document rootFile)
+        public (Error error, string href, string display, IXrefSpec spec) ResolveRelativeXref(SourceInfo<string> href, Document relativeTo, Document resultRelativeTo)
```
```suggestion
        public (Error error, string href, string display, IXrefSpec spec) ResolveRelativeXref(SourceInfo<string> href, Document declaringFile, Document referencingFile)
```

src/docfx/build/dependency/DependencyResolver.cs
yufeih: I've used `declaringFile` and `relativeToFile` to mimic `Path.GetRelativePath(string relativeTo, string path)`

src/docfx/lib/markdown/MarkdownUtility.cs
```diff
@@ -61,7 +65,16 @@ public static (List<Error> errors, MarkdownDocument ast) Parse(string content, M
 
                     var html = Markdown.ToHtml(markdown, s_markdownPipelines[(int)pipelineType]);
 
-                    return (status.Errors, html);
+                    var htmlWithRelativeLink = HtmlUtility.TransformLinks(html, (href, _) =>
+                    {
+                        if (href.StartsWith(RelativeUrlMarker))
+                        {
+                            return UrlUtility.GetRelativeUrl(file.SiteUrl, href.Substring(RelativeUrlMarker.Length));
+                        }
```
OsmondJiang: will this bring perf concern?

src/docfx/lib/UrlUtility.cs
```diff
@@ -110,6 +111,69 @@ public static DependencyType FragmentToDependencyType(string fragment)
             return fragment != null && fragment.Length > 1 ? DependencyType.Bookmark : DependencyType.Link;
         }
 
+        public static string GetRelativeUrl(string relativeToUrl, string url)
```
OsmondJiang: Why not just directly use Path.GetRelativeTo after removing the beginning `/`?

src/docfx/lib/markdown/MarkdownUtility.cs
```diff
@@ -75,20 +88,31 @@ internal static void LogError(Error error)
             t_status.Value.Peek().Errors.Add(error);
         }
 
-        internal static string GetLink(string path, object relativeTo, object resultRelativeTo, MarkdownObject origin, int columnOffset = 0)
+        internal static string GetLink(string path, object relativeTo, MarkdownObject origin, int columnOffset = 0)
         {
             var status = t_status.Value.Peek();
-            var (error, link, _) = status.Context.DependencyResolver.ResolveLink(new SourceInfo<string>(path, origin.ToSourceInfo(columnOffset: columnOffset)), (Document)relativeTo, (Document)resultRelativeTo);
+            var (error, link, linkType, _) = status.Context.DependencyResolver.ResolveLink(new SourceInfo<string>(path, origin.ToSourceInfo(columnOffset: columnOffset)), (Document)relativeTo);
```
OsmondJiang: Can we use the same way as ResolveXref to determine whether this is a relative path?
```
if (doc != null)
{
    link = RelativeUrlMarker + link
}
``` 

So then we don't need change the return value

src/docfx/lib/UrlUtility.cs
yufeih: `Path.GetRelativePath` produces different result, I initially used it, see the diff in [this report](https://github.com/dotnet/docfx/pull/4588#issuecomment-501539552)

src/docfx/lib/markdown/MarkdownUtility.cs
yufeih: Just tried, there are some subtle differences: when the link target does not exist, `doc` is null, but the links should still be treated as relative link.

src/docfx/lib/markdown/MarkdownUtility.cs
yufeih: not likely, `azure-docs-pr (00:02:28.9939835, 0 more diff)`

src/docfx/lib/markdown/MarkdownUtility.cs
OsmondJiang: When the line is not resolved, we leave it there, should be fine right?

src/docfx/lib/markdown/MarkdownUtility.cs
```diff
@@ -75,20 +88,31 @@ internal static void LogError(Error error)
...
+        internal static string GetLink(string path, object relativeTo, MarkdownObject origin, int columnOffset = 0)
         {
             var status = t_status.Value.Peek();
-            var (error, link, _) = status.Context.DependencyResolver.ResolveLink(new SourceInfo<string>(path, origin.ToSourceInfo(columnOffset: columnOffset)), (Document)relativeTo, (Document)resultRelativeTo);
+            var (error, link, file) = status.Context.DependencyResolver.ResolveLink(new SourceInfo<string>(path, origin.ToSourceInfo(columnOffset: columnOffset)), (Document)relativeTo);
             status.Errors.AddIfNotNull(error?.WithSourceInfo(origin.ToSourceInfo()));
+
+            if (file != null)
+            {
+                link = RelativeUrlMarker + link;
```
OsmondJiang: 👍

PR #4589, opened 2019-05-20
extract monikers of publishItems to decrease the manifest file size

928PJY:
and match the new publish manifest design

### DIFF
--- a/docs/specs/moniker.yml
+++ b/docs/specs/moniker.yml
@@ -43,7 +43,17 @@ outputs:
       "!monikers": null
     }
   .publish.json: |
-    {"files":[{"url":"/docs/a","group":"23d205a5"},{"url":"/docs/a","group":"9d4e15fd"},{"url":"/docs/a","!group":null}],"groups":{"23d205a5":["netcore-2.0","netcore-2.1"],"9d4e15fd":["netcore-1.0","netcore-1.1"]}}
+    {
+      "files":[
+        {"url":"/docs/a","moniker_group":"23d205a5"},
+        {"url":"/docs/a","moniker_group":"9d4e15fd"},
+        {"url":"/docs/a","!moniker_group":null}
+      ],
+      "monikerGroups":{
+        "23d205a5":["netcore-2.0","netcore-2.1"],
+        "9d4e15fd":["netcore-1.0","netcore-1.1"]
+      }
+    }
 ---
 # Multiple matched moniker range setting in the config, should use the last one
 inputs:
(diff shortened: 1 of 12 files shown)


src/docfx/publish/PublishItem.cs
```diff
@@ -15,12 +15,15 @@ internal class PublishItem
 
         public string Path { get; set; }
 
+        public string Group { get; set; }
```
yufeih:
```suggestion
        public string MonikerGroup { get; set; }
```

src/docfx/publish/PublishModel.cs
```diff
@@ -1,10 +1,14 @@
...
 
+using System.Collections.Generic;
+
 namespace Microsoft.Docs.Build
 {
     internal class PublishModel
     {
         public PublishItem[] Files { get; set; }
+
+        public Dictionary<string, List<string>> Groups { get; set; } = new Dictionary<string, List<string>>();
```
yufeih: `MonikerGroups`

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -53,6 +53,7 @@ internal static class BuildTableOfContents
                 Path = outputPath,
                 Locale = file.Docset.Locale,
                 Monikers = model.Metadata.Monikers,
+                Group = HashUtility.GetMd5HashShort(model.Metadata.Monikers),
```
yufeih: calculating hash is slow, `MonikerGroup` can simply be an increasing number starting from 1 and assigned during `PublishModelBuilder.Build`, this can also produce deterministic result.

928PJY: that will cause different monikers in the different repo to produce the same group name, which may be not acceptable for DHS. @fenxu any suggestion?
For perf concern, we can add a cache layer to avoid duplicate calculation.

OsmondJiang:
what's the reason to reduce the size of manifest file? since the new publish pipeline is still using a really large publish file, will this finally resolve our problem?

src/docfx/build/toc/BuildTableOfContents.cs
fenxu: 1. Keep result stable by using Hash.
2. Improve perf by cache

---
In reply to: [285826404](https://github.com/dotnet/docfx/pull/4589#discussion_r285826404) [](ancestors = 285826404)

src/docfx/publish/PublishItem.cs
```diff
@@ -15,12 +15,15 @@ internal class PublishItem
...
+
         public string Hash { get; set; }
 
         public string Locale { get; set; }
 
         public string RedirectUrl { get; set; }
 
+        [JsonIgnore]
         public List<string> Monikers { get; set; }
```
fenxu: >public List<string> Monikers { get; set; } [](start = 8, length = 42)

Add default constructor to avoid null case

928PJY:
why not? it can avoid duplicate data

---
In reply to: [494211364](https://github.com/dotnet/docfx/pull/4589#issuecomment-494211364) [](ancestors = 494211364)

yufeih:
Are there any data of how much the file is reduced on a real repo with monikers?

fenxu:
For .net, the file size will reduce to half based on Tianqi previous input.
cc @jipe@microsoft.com 

---
In reply to: [494222638](https://github.com/dotnet/docfx/pull/4589#issuecomment-494222638) [](ancestors = 494222638)

docs/specs/moniker.yml
```diff
@@ -193,11 +209,15 @@ outputs:
...
-        { "url": "/docs/a-old", "redirect_url": "/docs/a2", "monikers": [ "netcore-1.2", "netcore-1.3" ] },
-        { "url": "/docs/a", "monikers": [ "netcore-1.0", "netcore-1.1" ] },
-        { "url": "/docs/a", "monikers": [ "netcore-1.2", "netcore-1.3" ] },
-      ]
+        { "url": "/docs/a-old", "redirect_url": "/docs/a1", "moniker_group": "9d4e15fd" },
+        { "url": "/docs/a-old", "redirect_url": "/docs/a2", "moniker_group": "f961cd88" },
+        { "url": "/docs/a", "moniker_group": "9d4e15fd" },
+        { "url": "/docs/a", "moniker_group": "f961cd88" },
+      ],
+      "monikerGroups": {
```
yufeih: should be moniker_groups

PR #4590, opened 2019-05-20
Implement additionalProperties

mingwli:
#4567

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -371,12 +371,6 @@ namespace Microsoft.Docs.Build
             => new Error(ErrorLevel.Warning, "field-required", $"Missing required field '{name}'", source);
 
         /// <summary>
-        /// Additional keys are defined in given object
-        /// </summary>
-        public static Error AdditionalProperty(SourceInfo source, string propertyName)
-            => new Error(ErrorLevel.Warning, "additional-property", $"Additional property name '{propertyName}' is not allowed", source);
-
-        /// <summary>
         /// Used unknown YamlMime.
         /// Examples:
         ///   - forgot to define schema in schema document(yml)
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -371,6 +371,12 @@ namespace Microsoft.Docs.Build
             => new Error(ErrorLevel.Warning, "field-required", $"Missing required field '{name}'", source);
 
         /// <summary>
+        /// Additional keys are defined in given object
+        /// </summary>
+        public static Error AdditionalProperty(SourceInfo source, string propertyName)
+            => new Error(ErrorLevel.Warning, "additional-property", $"Additional property name '{propertyName}' is not allowed", source);
+
+        /// <summary>
         /// Used unknown YamlMime.
         /// Examples:
         ///   - forgot to define schema in schema document(yml)
--- a/src/docfx/lib/json/ValueOrObjectConverter.cs
+++ b/src/docfx/lib/json/ValueOrObjectConverter.cs
@@ -1,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Newtonsoft.Json;
+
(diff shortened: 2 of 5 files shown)


src/docfx/lib/schema/JsonSchema.cs
```diff
@@ -34,6 +34,14 @@ internal class JsonSchema
         /// </summary>
         public JValue[] Enum { get; set; }
 
+        /// <summary>
+        /// An object can have extra keys not defined in properties.
+        /// This can be:
+        ///     - boolean: allow/disallow additional properties
+        ///     - object: the schema for the additional properties
+        /// </summary>
+        public JToken AdditionalProperties { get; set; }
```
yufeih: We can only support `bool` for now

```suggestion
        public bool AdditionalProperties { get; set; }
```

or write an `ObjectOrValueConverter` similar to `OneOrManyConverter`

```suggestion
        [JsonConverter(typeof(ObjectOrValueConverter))]
        public (JsonSchema obj, bool value) AdditionalProperties { get; set; }
```

test/docfx.Test/lib/JsonSchemaTest.cs
```diff
@@ -132,6 +132,19 @@ public void TestJsonSchemaConfirmance(string description, string schemaText, str
         [InlineData("{'properties': {'key': {'type': 'string'}}}", "{'key': 1}",
             "['warning','unexpected-type','Expect type 'String' but got 'Integer'','file',1,9]")]
 
+        // additional properties validation
+        // AdditionalProperty is enabled with explicit false(can be named additionalProperties or additionProperties)
+        [InlineData("{'properties': {'key': {'type': 'string'}}, 'additionalProperties': {}}", "{'key': 'value', 'key1': 'value1'}", "")]
```
yufeih: remove line 18 `"additionalProperties"` to test against the spec

src/docfx/lib/schema/JsonSchema.cs
```diff
@@ -56,5 +64,18 @@ internal class JsonSchema
         /// Properties that are reserved by the system.
         /// </summary>
         public string[] Reserved { get; set; } = Array.Empty<string>();
+
+        /// <summary>
+        /// Alias for <see cref="AdditionalProperties"/>.
+        /// </summary>
+        [JsonProperty("additionProperties")]
+        private JToken AdditionProperties
```
yufeih: remove this, it is not part of the spec

test/docfx.Test/lib/JsonSchemaTest.cs
mingwli: @yufeih  there's an extra `patternProperties` which use regex to match keys, that we don't support yet. and it's not used in docs-ui.

src/docfx/Errors.cs
```diff
@@ -370,6 +370,12 @@ public static Error UndefinedValue(SourceInfo source, object value, IEnumerable<
         public static Error FieldRequired(SourceInfo source, string name)
             => new Error(ErrorLevel.Warning, "field-required", $"Missing required field '{name}'", source);
 
+        /// <summary>
+        /// Additional keys are defined in given object
+        /// </summary>
+        public static Error AdditionalProperty(SourceInfo source, string propertyName)
```
yufeih: this could be merged with `unknown field` error, they serve the same purpose.

src/docfx/lib/json/ValueOrObjectConverter.cs
```diff
@@ -0,0 +1,33 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
+{
+    internal class ValueOrObjectConverter<T1, T2> : JsonConverter where T2 : class
```
yufeih: You can dynamically get these types from `objectType` property in `ReadJson` method

```suggestion
    internal class ValueOrObjectConverter : JsonConverter
```

OsmondJiang:
#4567

PR #4594, opened 2019-05-21
build should not faile when sdp reference an empty UID

superyyrrzz:


herohua:
So, what will be the resolve behavior if `uid` == `string.Empty`

superyyrrzz:
@herohua Will not special handle empty string. It will be resolved as normal UID, and finally get a `UidNotFound`.

PR #4596, opened 2019-05-21
support parts of $ref and definitions for json schema

OsmondJiang:
1. root level ref pointer
2. definition ref pointer


test/docfx.Test/data/jschema/draft7/definitions.json
```diff
@@ -1,7 +1,7 @@
 [
     {
         "description": "valid definition",
-        "schema": {"$ref": "http://json-schema.org/draft-07/schema#"},
```
yufeih: don't change this, this means we don't support the `definitions` test suite, `ref` test suite already covers inline definitions

src/docfx/lib/schema/JsonSchemaValidation.cs
```diff
@@ -12,12 +12,14 @@ internal static class JsonSchemaValidation
         public static List<Error> Validate(JsonSchema schema, JToken token)
         {
             var errors = new List<Error>();
-            Validate(schema, token, errors);
+            Validate(schema, token, errors, schema);
             return errors;
         }
 
-        private static void Validate(JsonSchema schema, JToken token, List<Error> errors)
+        private static void Validate(JsonSchema schema, JToken token, List<Error> errors, JsonSchema root)
```
yufeih: consider having a `JsonSchemaValidationContext` class that holds the logic of JsonSchemaUtility, we may need instance fields.

```suggestion
        private static void Validate(JsonSchemaValidationContext context, JsonSchema schema, JToken token, List<Error> errors)
```

OsmondJiang:
#4567

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -24,6 +24,14 @@ internal static class JsonUtility
             new JTokenJsonConverter { },
         };
 
+        private static readonly JsonSerializer s_jsonSchemaSerializer = JsonSerializer.Create(new JsonSerializerSettings
+        {
+            NullValueHandling = NullValueHandling.Ignore,
+            MetadataPropertyHandling = MetadataPropertyHandling.Ignore,
```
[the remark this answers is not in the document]
OsmondJiang: sounds good

test/docfx.Test/lib/JsonSchemaTest.cs
```diff
@@ -52,6 +51,18 @@ public class JsonSchemaTest
...
+             // ref
+            "relative pointer ref to object",
+            "relative pointer ref to array",
+            "escaped pointer ref",
+            "remote ref, containing refs itself",
+            "$ref to boolean schema true",
+            "$ref to boolean schema false",
+            "Recursive references between schemas",
+            "refs with quote",
+            "root pointer ref/mismatch", // additional properties
```
yufeih: now `additional properties` has been implemented

src/docfx/lib/schema/JsonSchemaValidationContext.cs
```diff
@@ -0,0 +1,77 @@
...
+            {
+                // not supported
+                throw new NotSupportedException("Schema id is not supported");
+            }
+
+            var parts = jsonPointer.Split(new char[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
+            if (parts.Length != 2)
+            {
+                // not supported
+                throw new NotSupportedException("Non-root level schema definition is not supported");
```
yufeih: If we only need to support the root level definitions and references, can build a `Dictionary<string, JsonSchema>` field upfront and `GetDefinition` becomes a simple dictionary lookup.

test/docfx.Test/lib/JsonSchemaTest.cs
OsmondJiang: it's already included

src/docfx/lib/schema/JsonSchemaValidationContext.cs
OsmondJiang: it's actually a simple dictionary lookup, just add some validation before that

yufeih: It does more work like parsing and split #ref, the dictionary key should look like `#/definitions/xxx`, so `GetDefinition` becomes allocation free.

src/docfx/lib/schema/JsonSchemaValidationContext.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            Debug.Assert(!string.IsNullOrEmpty(fragment));
+
+            if (!string.IsNullOrEmpty(path))
+            {
+                // not supported
+                throw new NotSupportedException("Schema definition file is not supported");
+            }
+
+            string jsonPointer = fragment.TrimStart(new char[] { '#', '/' });
```
yufeih: this trim makes `#/definitions/a` , `####/definitions/a` , `/#definitions/a` behaves the same

OsmondJiang: this fragment is get from previous UrlUtility.Split, so can NOT the last one. For the first one and second ones, I believe we can treat them same. Do you want really strict rules on that?

src/docfx/lib/schema/JsonSchemaValidationContext.cs
```diff
@@ -0,0 +1,47 @@
...
+            return jsonSchema;
+        }
+
+        private (string key, JsonSchema schema) GetDefinitionCore(SourceInfo<string> @ref)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(@ref));
+
+            return _definitions.TryGetValue(@ref, out var schema)
+                ? (@ref, schema)
+                : throw Errors.SchemaDefinitionNotFound(@ref).ToException();
```
yufeih: this is not an error that writers could make, so an ordinary exception is enough, don't throw DocfxException

src/docfx/lib/schema/JsonSchemaValidationContext.cs
```diff
@@ -0,0 +1,47 @@
...
+        public JsonSchemaValidationContext(JsonSchema root)
+        {
+            Debug.Assert(root != null);
+
+            _root = root;
+            _definitions = _root.Definitions.ToDictionary(k => $"#/definitions/{k.Key}", v => v.Value);
+            _definitions.Add("#", root);
+        }
+
+        public JsonSchema GetDefinition(JsonSchema jsonSchema, HashSet<string> recursions = null, string key = "#")
```
yufeih: consider simplify public interface make recursion parameters private. It might be easier to pass `ref` to convey that we are getting a schema from a $ref

```suggestion
        public JsonSchema GetDefinition(string ref)
```

OsmondJiang: We need the whole schema

src/docfx/lib/schema/JsonSchemaValidationContext.cs
OsmondJiang: what ordinary exception would you prefer?  ApplicationException?

yufeih: ApplicationException looks good to me

src/docfx/lib/schema/JsonSchemaValidationContext.cs
```diff
@@ -0,0 +1,45 @@
...
+            if (jsonSchema != null &&
+                !string.IsNullOrEmpty(jsonSchema.Ref) &&
+                recursions.Add(jsonSchema.Ref))
+            {
+                if (_definitions.TryGetValue(jsonSchema.Ref, out var schema))
+                {
+                    return GetDefinitionCore(schema, recursions);
+                }
+
+                throw new ApplicationException($"Could not found `{jsonSchema.Ref}` schema definition");
```
live1206: found -> find

PR #4602, opened 2019-05-22
min/max items validation

mingwli:
#4567

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaValidation.cs
+++ b/src/docfx/lib/schema/JsonSchemaValidation.cs
@@ -12,11 +12,11 @@ namespace Microsoft.Docs.Build
         public static List<Error> Validate(JsonSchema schema, JToken token)
         {
             var errors = new List<Error>();
-            Validate(schema, token, errors, null);
+            Validate(schema, token, errors);
             return errors;
         }
 
-        private static void Validate(JsonSchema schema, JToken token, List<Error> errors, string name)
+        private static void Validate(JsonSchema schema, JToken token, List<Error> errors)
         {
             if (!ValidateType(schema, token, errors))
             {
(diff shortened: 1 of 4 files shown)


src/docfx/lib/schema/JsonSchemaValidation.cs
```diff
@@ -60,15 +60,21 @@ private static void ValidateScalar(JsonSchema schema, List<Error> errors, JValue
...
             {
                 foreach (var item in array)
                 {
-                    Validate(schema.Items, item, errors);
+                    Validate(schema.Items, item, errors, name);
                 }
             }
+
+            if (schema.MaxItems.HasValue && array.Count > schema.MaxItems.Value)
+                errors.Add(Errors.ArrayLengthInvalid(JsonUtility.GetSourceInfo(array), name ?? array.Path, maxItems: schema.MaxItems));
```
yufeih: `array.Path` is enough, there is no need to pass `name` all the way down

test/docfx.Test/lib/JsonSchemaTest.cs
```diff
@@ -149,6 +147,15 @@ public void TestJsonSchemaConfirmance(string description, string schemaText, str
             @"['warning','unexpected-type','Expect type 'Boolean' but got 'String'','file',1,4]
               ['warning','unexpected-type','Expect type 'Boolean' but got 'String'','file',1,8]")]
 
+        [InlineData("{'maxItems': 3, 'minItems': 1}", "['a','b']", "")]
+        [InlineData("{'properties': {'arr': {'maxItems': 3, 'minItems': 1}}}", "{'arr': ['a','b','c','d']}",
+            "['warning','array-length-invalid','Array 'arr' should be <= 3','file',1,9]")]
```
yufeih:
```suggestion
            "['warning','array-length-invalid','Array length should be <= 3 for 'xxx'','file',1,9]")]
```

PR #4604, opened 2019-05-22
consume order property in monikerDefinition file

928PJY:
The moniker definition list return by API `/v2/monikertrees/allfamiliesproductsmonikers` are not sorted

### DIFF
--- a/src/docfx/lib/moniker/EvaluatorWithMonikersVisitor.cs
+++ b/src/docfx/lib/moniker/EvaluatorWithMonikersVisitor.cs
@@ -19,6 +19,11 @@ namespace Microsoft.Docs.Build
             _monikerProductInfoDictionary = InitializeMonikers(monikerDefinition.Monikers);
         }
 
+        public List<string> GetSortedMonikerNameList()
+        {
+            return _monikerProductInfoDictionary.Keys.ToList();
+        }
+
         public IEnumerable<string> Visit(ComparatorExpression expression)
         {
             Debug.Assert(expression.Operand != null);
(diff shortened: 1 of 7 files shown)


src/docfx/lib/moniker/Moniker.cs
```diff
@@ -14,6 +14,8 @@ internal class Moniker
 
         public string ProductName { get; set; } = string.Empty;
 
+        public int Order { get; set; } = 0;
```
yufeih:
```suggestion
        public int Order { get; set; }
```

src/docfx/lib/moniker/EvaluatorWithMonikersVisitor.cs
```diff
@@ -66,7 +71,7 @@ public IEnumerable<string> Visit(LogicExpression expression)
         private static Dictionary<string, MonikerProductInfo> InitializeMonikers(IEnumerable<Moniker> monikers)
         {
             var monikerNameList = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-            var productNameDictionary = new Dictionary<string, List<string>>(StringComparer.OrdinalIgnoreCase);
+            var productNameDictionary = new SortedDictionary<string, List<Moniker>>(StringComparer.OrdinalIgnoreCase);
```
yufeih: why SortedDictionary

928PJY: I want the stable evaluation output of monikerRange which contains multiple products like `>netcore-1.0 || < netframework-2.0`

PR #4606, opened 2019-05-22
Environment.UserInteractive does not work as we expected

yufeih:
So drop it for now


OsmondJiang:
unless you have the other way to get the stderr out in ci pipeline

yufeih:
> unless you have the other way to get the stderr out in ci pipeline

it is not piped correctly

PR #4612, opened 2019-05-23
implement min/maxLength for string

mingwli:
#4567

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -371,7 +371,7 @@ namespace Microsoft.Docs.Build
             => new Error(ErrorLevel.Warning, "array-length-invalid", $"Array {(string.IsNullOrEmpty(propName) ? "" : $"'{propName}' ")}length should be {(minItems.HasValue ? $">= {minItems.Value}" : $"<= {maxItems.Value}")}", source);
 
         /// <summary>
-        /// String length not with min and max.
+        /// String length not within min and max.
         /// </summary>
         public static Error StringLengthInvalid(SourceInfo source, string propName, int? minLength = null, int? maxLength = null)
             => new Error(ErrorLevel.Warning, "string-length-invalid", $"String {(string.IsNullOrEmpty(propName) ? "" : $"'{propName}' ")}length should be {(minLength.HasValue ? $">= {minLength.Value}" : $"<= {maxLength.Value}")}", source);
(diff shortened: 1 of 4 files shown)


src/docfx/lib/schema/JsonSchemaValidation.cs
```diff
@@ -58,6 +58,24 @@ private static void ValidateScalar(JsonSchema schema, List<Error> errors, JValue
             {
                 errors.Add(Errors.UndefinedValue(JsonUtility.GetSourceInfo(scalar), scalar, schema.Enum));
             }
+
+            if ((schema.MaxLength.HasValue || schema.MinLength.HasValue) && CanTreatAsString())
```
yufeih: Is checking the value good enough to detect whether it is a string?

```suggestion
            if ((schema.MaxLength.HasValue || schema.MinLength.HasValue) && scalar.Value is string str)
```

src/docfx/lib/schema/JsonSchemaValidation.cs
```diff
@@ -58,6 +58,16 @@ private static void ValidateScalar(JsonSchema schema, List<Error> errors, JValue
             {
                 errors.Add(Errors.UndefinedValue(JsonUtility.GetSourceInfo(scalar), scalar, schema.Enum));
             }
+
+            if ((schema.MaxLength.HasValue || schema.MinLength.HasValue) && scalar.Value is string str)
```
OsmondJiang: do we need throw exception here if value is not string?

mingwli: tried that at "https://www.jsonschemavalidator.net/", no validation error thrown.

src/docfx/Errors.cs
```diff
@@ -364,9 +364,18 @@ public static Error UnexpectedType(SourceInfo source, string expectedType, strin
...
             => new Error(ErrorLevel.Warning, "undefined-value", $"Value '{value}' is not accepted. Valid values: {Join(validValues)}", source);
 
+        /// <summary>
+        /// Array length not within min and max.
+        /// </summary>
         public static Error ArrayLengthInvalid(SourceInfo source, string propName, int? minItems = null, int? maxItems = null)
             => new Error(ErrorLevel.Warning, "array-length-invalid", $"Array {(string.IsNullOrEmpty(propName) ? "" : $"'{propName}' ")}length should be {(minItems.HasValue ? $">= {minItems.Value}" : $"<= {maxItems.Value}")}", source);
 
+        /// <summary>
+        /// String length not with min and max.
```
live1206:
```suggestion
        /// String length not within min and max.
```

Issue #4067, opened 2019-02-22
Support schema driven documents from JSON schema


yufeih:
### Phases
- 1️⃣ Support Conceptual SDP like `ContextObject`, `LandingData`
  - Page Model to generate `.raw.page.json` + `.meta.json`
  - Data Model to generate `.json`
- 2️⃣ Support Reference SDP without fragments like `OData`
- 3️⃣ Support Reference SDP with fragments.

### Feature Scope
- 1️⃣  Integration with Template json schema
  - Load json schema from Template.
  - Apply template to docfx output model.
  - Pre-process template to adapt the new output model and pipeline.
  - Local debugging and testing
- 1️⃣  Json Schema validation(#4567)
  - Some basic rules implementation.
  - Testing.
- 1️⃣  Json Schema Transform
  - Support Kinds of `content-type`
    - Markdown
    - InlineMarkdown
    - Href
    - Html
    - Xref
      - Transform xref from string to xrefSpec.
  - Testing
- 2️⃣  Build Internal Xref Map
  - Retrieve UIDs
    - Validate duplicated uid in same file.
    - Validate duplicate uid cross whole repo.
  - Get all uid and corresponding xref properties based on `XrefProperties`
    - Support leaf nodes only? @yufeih  @live1206 
    - Monikers(dotnet files)
  - Output internal xref map files.
- 2️⃣  Resolve and export XREF map by versioning
  - Respect versioning defined in config
  - Respect versioning defined per file (metadata)
  - Respect versioning defined in node (dotnet files)
- 1️⃣  Setup Integration Test and Migration Test
  - Setup integration test
  - Setup migration test
- 3️⃣  Support fragments
  - UNKNOWN
- Legacy
  - 1️⃣ Keep landing data working for a while
  - Future clean up

### Json schema validation

- [x] Json schema validation (#4567)

### Json schema transform

- [x] Basic content type transforms Html/Markdown/InlineMarkdown/Link
- [x] Basic content type testing

### Template integration

- [x] Allow template from local folder (#4559)
- [x] Read schema from template
- [x] Run JINT and mustache (both optional)
- [ ] Handle `data-linktype` for links in mustache template if template is pushed futher down the stack

### Xref handling

- [ ] Build Xref Map from XrefProperties
  - [x] Respect `XrefProperties` defined in JSON schema
  - [x] Confirm whether we need support non-root level xref properties
  - [ ] output xref map of sdp pipeline
[...]

PR #4613, opened 2019-05-23
Temporarily move HtmlPostProcess after applying template

yufeih:
First to unblock https://dev.azure.com/ceapex/Engineering/_git/Docs.DocFX.Impact/pullrequest/11885?_a=files, it needs to be tackled later

#4067


OsmondJiang:
do we need create a issue for tracking that?

yufeih:
> do we need create a issue for tracking that?

All will be covered in #4067

Issue #4623, opened 2019-05-24
Build a private repo without a github token causes lots of duplicated github-api-failure


yufeih:
Build a private repo without a github token triggers a lot of duplicated calls to GitHub, all failed with:

```json
{
"message":"This endpoint requires you to be authenticated.",
"documentation_url":"https://developer.github.com/v3/#authentication"
}
```

Command line output, notice the duplicated names:

```
Calling GitHub commit API to resolve wibjorn@microsoft.com
heading-not-found devsandbox/redirect.md(0,0): The first visible block is not a heading block with `#`, `##` or `###`
github-api-failed devsandbox/redirect.md(0,0): Call to GitHub API 'https://api.github.com/graphql' failed: {"message":"This endpoint requires you to be authenticated.","documentation_url":"https://developer.github.com/v3/#authentication"} Try closing and reopening the PR. If you get this Error again, file an issue.
Calling GitHub commit API to resolve jeremy.danyow@microsoft.com
heading-not-found devsandbox/notFound.md(0,0): The first visible block is not a heading block with `#`, `##` or `###`
github-api-failed devsandbox/notFound.md(0,0): Call to GitHub API 'https://api.github.com/graphql' failed: {"message":"This endpoint requires you to be authenticated.","documentation_url":"https://developer.github.com/v3/#authentication"} Try closing and reopening the PR. If you get this Error again, file an issue.
Calling GitHub commit API to resolve adkinn@microsoft.com
Calling GitHub commit API to resolve wibjorn@microsoft.com
Calling GitHub commit API to resolve jeremy.danyow@microsoft.com
heading-not-found devsandbox/archive.md(0,0): The first visible block is not a heading block with `#`, `##` or `###`
github-api-failed devsandbox/archive.md(0,0): Call to GitHub API 'https://api.github.com/graphql' failed: {"message":"This endpoint requires you to be authenticated.","documentation_url":"https://developer.github.com/v3/#authentication"} Try closing and reopening the PR. If you get this Error again, file an issue.
Calling GitHub commit API to resolve wibjorn@microsoft.com
Calling GitHub commit API to resolve duncanma@microsoft.com
Calling GitHub commit API to resolve jedanyow@microsoft.com
[...]

PR #4624, opened 2019-05-24
dont call github graph again when auth failed

OsmondJiang:
#4623


OsmondJiang:
> check that test case to assert it always return an error?

That's a weird but good idea

PR #4627, opened 2019-05-24
use github graph api tro get github user by login

OsmondJiang:


OsmondJiang:
image

The reason to change the test is that the "docascode" is not a user, it's a org.

OsmondJiang:
#4627

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -7,32 +7,33 @@
...
 using System.Threading.Tasks;
 using Newtonsoft.Json;
-using Octokit;
 
 namespace Microsoft.Docs.Build
 {
-    internal class GitHubAccessor
+    internal sealed class GitHubAccessor
     {
         private static readonly HttpClient s_httpClient = new HttpClient();
```
superyyrrzz: As the constructor will add token to this client, suggest not to use static.

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -42,39 +43,85 @@ public async Task<(Error, GitHubUser)> GetUserByLogin(string login)
...
+                    login,
+                });
+
+                if (error != null)
+                {
+                    return (error, null);
+                }
+
+                var grahpApiResponse = JsonConvert.DeserializeAnonymousType(await response.Content.ReadAsStringAsync(), new
+                {
```
superyyrrzz: Concrete the type definition into a strong type model? Feels a bit too long here and not very straightforward.

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -42,39 +43,85 @@ public async Task<(Error, GitHubUser)> GetUserByLogin(string login)
...
+                        {
+                            name = "",
+                            email = "",
+                            login = "",
+                            databaseId = 0,
+                        },
+                    },
+                });
+
+                if (grahpApiResponse.errors != null && grahpApiResponse.errors.Any())
```
superyyrrzz: `Any()` => `.Length`

src/docfx/lib/github/GitHubAccessor.cs
OsmondJiang: good suggestion, but need change a lot of class to IDispoase...

src/docfx/lib/github/GitHubAccessor.cs
OsmondJiang: there are too many depths of Grahp API Result, and most of them are not re-usable,  so I would like use anonymous type,

live1206: how about just loading the string into JToken? what is the benefit of another de-serialization into anonymous type?

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -42,39 +43,85 @@ public async Task<(Error, GitHubUser)> GetUserByLogin(string login)
...
+                    if (notFoundError != null)
+                        return default;
+
+                    var rateLimitError = grahpApiResponse.errors.FirstOrDefault(e => e.type == "MAX_NODE_LIMIT_EXCEEDED" || e.type == "RATE_LIMITED");
+                    if (rateLimitError != null)
+                    {
+                        _rateLimitError = Errors.GitHubApiFailed(_url, rateLimitError.message);
+                    }
+
+                    return (Errors.GitHubApiFailed(_url, rateLimitError?.message ?? grahpApiResponse.errors.First().message), null);
```
yufeih: The github api failed error message is already very long, consider remove the detailed diagnostic message from error to hide it from error report, write it using `Log.Write` instead.

OsmondJiang: I would like expose this error message to end user to let them know the reason.

OsmondJiang: the message is really helpful.

image

yufeih: I once see a full json dumped into the message, it it because line 286 simply dumps the whole JSON response?

yufeih: this part is fine, only including the message

OsmondJiang: unless we give the wrong query, we should get clear and clean error message

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -226,27 +257,38 @@ public async Task<(Error, IEnumerable<GitHubUser>)> GetUsersByCommit(string repo
...
+            if (!response.IsSuccessStatusCode)
             {
-                Log.Write($"Failed calling GitHub API '{api}', message: '{ex.Message}', retryAfterSeconds: '{aex.RetryAfterSeconds}'");
+                var message = await response.Content.ReadAsStringAsync();
+
+                if (response.StatusCode == (HttpStatusCode)401)
+                {
+                    _unauthorizedError = Errors.GitHubApiFailed(_url, message);
+                }
+                return (Errors.GitHubApiFailed(_url, message), null);
```
yufeih: I mean the message here

PR #4629, opened 2019-05-24
Fixes CI failure

yufeih:
Fixes CI failure

### DIFF
--- a/src/docfx/lib/github/GitHubAccessor.cs
+++ b/src/docfx/lib/github/GitHubAccessor.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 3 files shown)


src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -231,7 +232,7 @@ HttpRequestMessage CreateHttpRequest(HttpContent content)
                 var message = new HttpRequestMessage();
                 message.Headers.Add("User-Agent", "DocFX");
                 if (!string.IsNullOrEmpty(_token))
-                    message.Headers.Add("Authorization", $"bearer {_token}");
```
OsmondJiang: from here: https://developer.github.com/v4/guides/forming-calls/#communicating-with-graphql

It should be a bearer token

PR #4630, opened 2019-05-24
output monikers soted by OrdinalIgnoreCase stringComparer:

928PJY:


src/docfx/lib/moniker/EvaluatorWithMonikersVisitor.cs
```diff
@@ -71,7 +71,7 @@ public IEnumerable<string> Visit(LogicExpression expression)
         private static Dictionary<string, MonikerProductInfo> InitializeMonikers(IEnumerable<Moniker> monikers)
         {
             var monikerNameList = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-            var productNameDictionary = new SortedDictionary<string, List<Moniker>>(StringComparer.OrdinalIgnoreCase);
+            var productNameDictionary = new Dictionary<string, List<Moniker>>(StringComparer.OrdinalIgnoreCase);
```
live1206: no need for the ordering of product name any more?

928PJY: since we will always sort by stringComparer in the end, so it is necessary here.

PR #4635, opened 2019-05-27
Enable integration testing for forked repos

yufeih:
- Replace System.AccessToken with a custom devops token
- Removes a private repo spec test because it is covered by integration test

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -163,25 +163,6 @@ outputs:
       "conceptual": "<p>for long file path test.</p>"
     }
 ---
-# [Skip] Restore private azure repos with dependencies
-# VSTS recently are not really stable, so skip this test for now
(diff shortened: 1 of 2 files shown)


docs/specs/restore.yml
```diff
@@ -163,25 +163,6 @@ outputs:
       "conceptual": "<p>for long file path test.</p>"
     }
 ---
-# [Skip] Restore private azure repos with dependencies
-# VSTS recently are not really stable, so skip this test for now
```
OsmondJiang: Can we enable it after this PR is merged?

yufeih: enabling that requires modifying test framework to support base64 encoding (expressions?), since this scenarios is main for integration, it is already covered by integration tests, so I don't plan to enable it again.

PR #4636, opened 2019-05-27
Resolving same uid with multiple versioning

live1206:


docs/designs/xref.md
```diff
@@ -177,6 +177,33 @@ There are some other UID reference formats supported in docfx V2, which are not
...
+resolving uid `a` should take the one with the highest version and also respect the version of file `b`
+```
+folder-1(1.0, 2.0):
+   - a.md(uid: a, title: A1)
+   - b.md(@a)
+folder-2(3.0,4.0):
+   - a.md(uid: a, title: A2)
+   - b.md(@a)
+```
+- multiple versioning with different product names, take the one with highest product name alphabetically
```
OsmondJiang: what's about the custom order defined for monikers?

OsmondJiang: can we add output example here？

docs/designs/xref.md
```diff
@@ -177,6 +177,42 @@ There are some other UID reference formats supported in docfx V2, which are not
...
+outputs:
+    folder-3(c-1.0, c-2.0):
+        - b.md(<a href="url">A2</a>)
+```
+
+### External uid with multiple versioning
+The resolving logic for external uid is the same as internal uid, but how to consume `xrefmap` is different.
+v3 is consuming `xrefmap.json` of v2 output for now. And for multiple versioning, v2 would output multiple `xrefmap-versionxx.json` files. There are two options here:
+- switch to consume `xrefmap.json` from v3 output, we need to publish `xrefmap.json` with multiple versioning in v3
+- v2 needs to modify `xrefspec` model to include versioning information, and combine multiple `xrefmap-versionxx.json` into one `xrefmap.json`
```
yufeih: if 2 is the preferred approach, we can only say 2 in the spec. I also don't like 1.

live1206: does this mean v3  would always consume the `xrefmap.json` output from v2?

yufeih: I see, it is not choose one of the two options, it is that there are two scenarios to support.

> There are two options here

Issue #4591, opened 2019-05-20
schema validation checks are in different places and duplicated


OsmondJiang:
`is-null`  is checked at JsonUtility.Parse (only object, not in array) and also JsonSchemaValidation.Validate

PR #4639, opened 2019-05-27
👨‍🏭remove null value check because it will be covered by json schema

OsmondJiang:
#4591

### DIFF
--- a/src/docfx/lib/json/JsonUtility.cs
+++ b/src/docfx/lib/json/JsonUtility.cs
@@ -284,11 +284,6 @@ namespace Microsoft.Docs.Build
 
             RemoveNullsCore(root, null);
 
-            foreach (var (node, name) in nullNodes)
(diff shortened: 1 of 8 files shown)


OsmondJiang:
The integration test failure is expected:

image

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -284,11 +284,6 @@ public static (List<Error>, JToken) RemoveNulls(this JToken root)
 
             RemoveNullsCore(root, null);
 
-            foreach (var (node, name) in nullNodes)
```
yufeih: also remove `nullNodes` traversal in `RemoveNullsCore`

OsmondJiang: sorry, cleaned up more

PR #4641, opened 2019-05-27
Remove ConfigFileName in favor of source info

yufeih:
Now that we have source info, `ConfigFileName` can be removed. Most of the changes are related to redirection because it is the only error that still uses that property.

### DIFF
--- a/docs/specs/casing.yml
+++ b/docs/specs/casing.yml
@@ -24,7 +24,7 @@ inputs:
   docfx.yml: |
     redirections:
       docs/redirect.md: /path
-      docs/redirect1.md: /PATH
+      docs/redirect1.md: /PATH1
   docs/a.md: |
     [redirect](REDIRECT.md)
 outputs:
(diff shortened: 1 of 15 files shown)


yufeih:
Integration test expected:
```diff
 {
   "message_severity": "warning",
   "code": "link-out-of-scope",
-  "message": "File 'devsandbox/azure-selector/includes/single-selector.md' referenced by link './azure-selector/includes/single-selector.md' will not be built because it is not included in docfx.yml",
+  "message": "File 'devsandbox/azure-selector/includes/single-selector.md' referenced by link './azure-selector/includes/single-selector.md' will not be built because it is not included in build scope",
   "file": "devsandbox/toc.yml",
   "line": 18
 }

```

OsmondJiang:
from your change, the error message actually lost the ability to tell user which config file has problem

yufeih:
> from your change, the error message actually lost the ability to tell user which config file has problem

The message has't been decided yet, there are other ways to communicate the problem without explicitly saying the name of the config file, I just want to remove this parameter so it doesn't complicate the pipeline.

Issue #4559, opened 2019-05-16
Allow git dependencies (theme) to point to a local folder


yufeih:
Currently git dependencies must be a remote git repository. To develop template locally, it needs to use template from a local folder. For file references in config, we allow file to be both an external URL or a local file, we could to the same to allow git dependencies to be either a git repository (starting with http:// or https://) or a local folder.

KalleOlaviNiemitalo:
> Currently git dependencies must be a remote git repository.

Having not used DocFX v3 yet, I was wondering how the remote git access is authenticated, but I see it is well documented in [docs/designs/credential.md](https://github.com/dotnet/docfx/blob/1408757302b2f66c9fb1515e97462104533bf9d2/docs/designs/credential.md) on the v3 branch.

yufeih:
@KalleOlaviNiemitalo , docfx calls git.exe, so if you have all credentials hooked up in git credential managers, it will automatically authenticate. 

In other scenarios where you might want to pass the credential using a config, we support setting  [http.extraheader](https://git-scm.com/docs/git-config#Documentation/git-config.txt-httpextraHeader) in docfx config file as described in [credentials.md](https://github.com/dotnet/docfx/blob/1408757302b2f66c9fb1515e97462104533bf9d2/docs/designs/credential.md#request-header)

yufeih:
With git dependencies supporting local folders, we might be able to deprecate `resolveAlias` and instead use:

```yaml
dependencies:
    '~': 'source_base_path'
```

PR #4642, opened 2019-05-27
Allow git dependencies to be local files

yufeih:
#4559

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -166,10 +166,9 @@ namespace Microsoft.Docs.Build
         public readonly SourceInfo<string>[] MetadataSchema = Array.Empty<SourceInfo<string>>();
 
         /// <summary>
-        /// Get the theme repo url like https://github.com/docs/theme#master
-        /// It's used for legacy doc(docs.com) sites build
+        /// Get the template folder or git repository url (like https://github.com/docs/theme#master)
         /// </summary>
-        public readonly string Theme = string.Empty;
+        public readonly string Template = string.Empty;
 
         /// <summary>
         /// Gets the dependency lock file path
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -168,7 +168,7 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Get the template folder or git repository url (like https://github.com/docs/theme#master)
         /// </summary>
-        public readonly string Template = string.Empty;
+        public readonly string Theme = string.Empty;
 
         /// <summary>
         /// Gets the dependency lock file path
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -192,9 +192,9 @@ namespace Microsoft.Docs.Build
                 }
             }
 
-            if (UrlUtility.IsHttp(config.Template))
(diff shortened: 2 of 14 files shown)


src/docfx/config/Config.cs
```diff
@@ -166,10 +166,9 @@ internal sealed class Config
         public readonly SourceInfo<string>[] MetadataSchema = Array.Empty<SourceInfo<string>>();
 
         /// <summary>
-        /// Get the theme repo url like https://github.com/docs/theme#master
-        /// It's used for legacy doc(docs.com) sites build
+        /// Get the template folder or git repository url (like https://github.com/docs/theme#master)
         /// </summary>
-        public readonly string Theme = string.Empty;
+        public readonly string Template = string.Empty;
```
OsmondJiang: that's a break change, also need change it Configuration Migration Tool, @mingwli

src/docfx/restore/RestoreGit.cs
```diff
@@ -183,32 +183,26 @@ void AddWorkTrees()
 
         private static IEnumerable<(string remote, string branch, GitFlags flags)> GetGitDependencies(Config config, string locale, Repository rootRepository)
         {
-            var dependencies = config.Dependencies.Values.Select(url =>
+            foreach (var (_, url) in config.Dependencies)
```
OsmondJiang: will this change also applies to Dependencies? How we know the branch info for local dependency folder? Can we add a spec test for that?

src/docfx/restore/RestoreGit.cs
```diff
@@ -183,32 +183,26 @@ void AddWorkTrees()
...
-            dependencies = dependencies.Concat(GetThemeGitDependencies(config, locale));
+                if (UrlUtility.IsHttp(url))
+                {
+                    var (remote, branch, _) = UrlUtility.SplitGitUrl(url);
+                    yield return (remote, branch, GitFlags.DepthOne);
+                }
+            }
 
-            if (rootRepository != null)
+            if (UrlUtility.IsHttp(config.Template))
```
OsmondJiang: only root repo need get template

src/docfx/restore/RestoreMap.cs
```diff
@@ -177,10 +186,10 @@ public static RestoreMap
             {
                 if (!successed && root)
                 {
-                    foreach (var (k, v) in acquired)
-                    {
-                        ReleaseGit(v.git, LockType.Shared, false);
-                    }
+                foreach (var (k, v) in acquired)
```
OsmondJiang: spaces

src/docfx/config/Config.cs
yufeih: yep, intentional

src/docfx/restore/RestoreMap.cs
```diff
@@ -21,20 +20,26 @@ public RestoreMap(IReadOnlyDictionary<(string remote, string branch, string comm
...
-        }
+            if (!UrlUtility.IsHttp(remote))
+            {
+                var fullPath = Path.Combine(docsetPath, remote);
+                if (Directory.Exists(fullPath))
+                {
+                    return (fullPath, new RestoreMap(_acquiredGits));
+                }
+
+                // TODO: Intentionally don't fallback to fallbackDocset for git restore path,
```
OsmondJiang: it;s overwrite config scenario, not for test happy

PR #4643, opened 2019-05-27
date-time format validation

mingwli:
#4567


src/docfx/Errors.cs
```diff
@@ -364,6 +364,12 @@ public static Error UnexpectedType(SourceInfo source, string expectedType, strin
         public static Error UndefinedValue(SourceInfo source, object value, IEnumerable<object> validValues)
             => new Error(ErrorLevel.Warning, "undefined-value", $"Value '{value}' is not accepted. Valid values: {Join(validValues)}", source);
 
+        /// <summary>
+        /// The string type's value doesn't match given format.
+        /// </summary>
+        public static Error StringFormatInvalid(SourceInfo source, string value, JsonSchemaStringFormat type)
+            => new Error(ErrorLevel.Warning, "string-format-invalid", $"String '{value}' is not a valid '{type}'", source);
```
```suggestion
            => new Error(ErrorLevel.Warning, "format-invalid", $"'{value}' is not a valid '{type}'", source);
```

src/docfx/lib/json/StringEnumDashConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+using System;
+using System.Linq;
+using System.Text.RegularExpressions;
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
+{
+    internal class StringEnumDashConverter : JsonConverter
+    {
+        public override bool CanConvert(Type objectType) => objectType == typeof(string);
```
yufeih: drop this in favor of attribute

src/docfx/lib/schema/JsonSchemaStringFormat.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace Microsoft.Docs.Build
+{
+    internal enum JsonSchemaStringFormat
+    {
+        DateTime,
+        Uri,
```
yufeih: If we don't want to support Uri/UriReference, can remove them from here

src/docfx/lib/schema/JsonSchemaValidation.cs
```diff
@@ -70,6 +71,14 @@ private static void ValidateScalar(JsonSchema schema, List<Error> errors, JValue
                 if (schema.MinLength.HasValue && unicodeLength < schema.MinLength.Value)
                     errors.Add(Errors.StringLengthInvalid(JsonUtility.GetSourceInfo(scalar), scalar.Path, minLength: schema.MinLength));
             }
+
+            switch (schema.Format)
+            {
+                case JsonSchemaStringFormat.DateTime:
+                    if (!DateTime.TryParse((string)scalar, out var _))
```
yufeih: scalar may not be a string, cast here will throw an exception

test/docfx.Test/lib/JsonSchemaTest.cs
```diff
@@ -145,6 +144,11 @@ public void TestJsonSchemaConfirmance(string description, string schemaText, str
             @"['warning','string-length-invalid','String 'str' length should be <= 2','file',1,13]
               ['warning','string-length-invalid','String 'str' length should be >= 4','file',1,13]")]
 
+        // string format validation
+        [InlineData("{'type': ['string'], 'format': 'date-time'}", "'1963-06-19T08:30:06Z'", "")]
+        [InlineData("{'type': ['string'], 'format': 'date-time'}", "'invalid'",
+            "['warning','format-invalid','String 'invalid' is not a valid 'DateTime'','file',1,9]")]
+
```
yufeih: can you add this test? `value<string>` might also throw

```csharp
[InlineData("{'type': ['string'], 'format': 'date-time'}", "1"]
```

mingwli: I also tried with
```
var v = new JValue(); // with all 13 overloads except JValue(JValue) & JValue(object)
var vStr = (string)v;
```
and they are all parsed with no exception.

src/docfx/lib/schema/JsonSchemaStringFormat.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.Serialization;
+
+namespace Microsoft.Docs.Build
+{
+    internal enum JsonSchemaStringFormat
+    {
+        None,
+        [EnumMember(Value = "date-time")]
```
yufeih: nit: add an empty line after `None,`

test/docfx.Test/lib/JsonSchemaTest.cs
```diff
@@ -145,6 +144,13 @@ public void TestJsonSchemaConfirmance(string description, string schemaText, str
             @"['warning','string-length-invalid','String 'str' length should be <= 2','file',1,13]
               ['warning','string-length-invalid','String 'str' length should be >= 4','file',1,13]")]
 
+        // string format validation
+        [InlineData("{'type': ['string'], 'format': 'date-time'}", "'1963-06-19T08:30:06Z'", "")]
+        [InlineData("{'type': ['string', 'number'], 'format': 'date-time'}", "1",
```
yufeih: How about
```suggestion
        [InlineData("{'format': 'date-time'}", "[]",
```

PR #4645, opened 2019-05-27
🐱‍👤Consolidate resolve internal and external xref map

live1206:


src/docfx/build/xref/XrefMap.cs
```diff
@@ -18,8 +18,8 @@ namespace Microsoft.Docs.Build
     internal class XrefMap
     {
         // TODO: key could be uid+moniker+locale
-        private readonly IReadOnlyDictionary<string, List<(InternalXrefSpec specs, Document file)>> _internalXrefMap;
-        private readonly IReadOnlyDictionary<string, Lazy<XrefSpec>> _externalXrefMap;
+        private readonly Dictionary<string, List<Lazy<IXrefSpec>>> _map = new Dictionary<string, List<Lazy<IXrefSpec>>>();
```
yufeih: try split xrefmap build from xrefmap resolve

src/docfx/build/xref/IXrefSpec.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using System.Collections.Generic;
+
+namespace Microsoft.Docs.Build
+{
+    internal interface IXrefSpec
+    {
+        string Href { get; }
+
+        Document ReferencedFile { get; }
```
yufeih: nit: consider rename it to `DeclaringFile` to indicate that it is the file that declares this xref.

src/docfx/build/xref/ExternalXRefSpec.cs
```diff
@@ -7,16 +7,18 @@
...
 namespace Microsoft.Docs.Build
 {
-    public class XrefSpec
+    internal class ExternalXrefSpec : IXrefSpec
     {
         public string Uid { get; set; }
 
+        public string Href { get; set; }
+
+        public Document ReferencedFile { get; set; }
```
yufeih: Since this class is still _THE_ data model, use `Document IXrefSpec.ReferencedFile => null` to hide this member both from serialization and from callers.

I would also like to prefix every data model with `Model` to clearly indicate that the class is a data contract and will be saved to disk, thus any changes to it must consider serialization and be backward compatible.

src/docfx/build/xref/XrefMap.cs
```diff
@@ -104,222 +106,79 @@ private bool TryResolveFromInternal(string uid, SourceInfo<string> href, string
             return false;
         }
 
-        private bool TryResolveFromExternal(string uid, out XrefSpec spec)
+        private IXrefSpec GetSpec(string uid, SourceInfo<string> href, string moniker, List<IXrefSpec> specs)
```
yufeih:
```suggestion
        private IXrefSpec GetXrefSpec(string uid, SourceInfo<string> href, string moniker, List<IXrefSpec> xrefSpecs)
```

src/docfx/build/xref/ExternalXRefSpec.cs
live1206: Will hold the `Model` naming change here, we can update the data contracts once we agree on the naming

yufeih: yep

src/docfx/build/xref/XrefMap.cs
```diff
@@ -2,28 +2,29 @@
...
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal class XrefMap
     {
         // TODO: key could be uid+moniker+locale
-        private readonly IReadOnlyDictionary<string, List<(InternalXrefSpec specs, Document file)>> _internalXrefMap;
-        private readonly IReadOnlyDictionary<string, Lazy<XrefSpec>> _externalXrefMap;
+        private readonly IReadOnlyDictionary<string, List<Lazy<IXrefSpec>>> _map;
```
OsmondJiang: add a test case for same uid exist in both external xref map and internal xref map

PR #4648, opened 2019-05-27
validate toc name during build

OsmondJiang:
Remove [JsonRequired] attribute and validate name during TOC build


src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -138,6 +139,13 @@ private static (List<Error>, TableOfContentsModel) LoadTocModel(JToken tocToken)
                 tocModelItem.TocHref = resolvedTocHref;
                 tocModelItem.Name = tocModelItem.Name ?? resolvedTopicName;
                 tocModelItem.Items = subChildren?.Items ?? tocModelItem.Items;
+
+                // validate
+                // todo: how to do required validation in strong model
```
OsmondJiang: Any idea here? @yufeih  @live1206 ?

OsmondJiang: If the name is not defined in input model, then we don't know the line info when reporting error.

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -138,6 +139,13 @@ private static (List<Error>, TableOfContentsModel) LoadTocModel(JToken tocToken)
...
+                if (tocModelItem.Name != null && string.IsNullOrEmpty(tocModelItem.Name))
```
yufeih: This check looks wierd, did you mean

```suggestion
                if (string.IsNullOrEmpty(tocModelItem.Name))
```

OsmondJiang: we need make sure tocModelItem.Name is not null to get source info...

yufeih: I see, if I turn `SourceInfo<T>` into a struct, does it makes null handling easier?
```
struct SourceInfo<T>
{
    public T Value;
    public SourceInfo SourceInfo;
}
```

OsmondJiang: I doubt about that, it will make the sourceInf usage complex and don't resolve our problem since the source info is still not correct(should be parent's)

Issue #4115, opened 2019-03-06
Docfx throws unexpected exception and crash (segfault)


yufeih:
Replacing all `docs/` with empty string in all .yml files and run tests, there are two issues:
1. docfx crash (segfault)
2. Lots of unexpected exception and assertions throws from source code, these should all be errors in reports or test exceptions.

PR #4649, opened 2019-05-28
Allow source info propagation for null values

### LINKED ISSUE #4647: source info doesn't work when key is undefined
as title, even with `key` defined.

yufeih:
Fixes #4647 , related #4115


src/docfx/lib/json/JsonContractResolver.cs
```diff
@@ -29,61 +29,69 @@ protected override JsonObjectContract CreateObjectContract(Type objectType)
...
+            return !(value is string) && value is IEnumerable enumerable && !enumerable.GetEnumerator().MoveNext();
+        }
+
+        private static void HandleSourceInfo(JsonProperty property)
+        {
+            if (property.PropertyType.IsGenericType && property.PropertyType.GetGenericTypeDefinition() == typeof(SourceInfo<>))
             {
-                if (!prop.Writable)
+                // Allow source info propagation for null values
+                property.NullValueHandling = NullValueHandling.Include;
```
OsmondJiang: keep the behavior consistent with normal type null handling: null value means using docfx default value.

src/docfx/lib/json/JsonContractResolver.cs
```diff
@@ -29,61 +29,69 @@ protected override JsonObjectContract CreateObjectContract(Type objectType)
...
+
+                // Do not serialize empty array of null
```
live1206:
```suggestion
                // Do not serialize empty array or null
```

test/docfx.Test/lib/JsonUtilityTest.cs
```diff
@@ -554,6 +554,37 @@ public void TestJsonMerge(string a, string b, string result)
...
+        }
+
+        [Fact]
+        public void JsonWithTypeErrorsHasCorrectSourceInfo()
+        {
+            var (_, json) = JsonUtility.Parse(
+                "{'next': {'a': ['a1','a2'],'b': 'b'}, 'c': 'c'}".Replace('\'', '"'), "file");
+            var (errors, obj) = JsonUtility.ToObject<ClassWithSourceInfo>(json);
+
+            Assert.Null(obj.Next.A.Value);
```
yufeih: this test showcased that if we simply `return null` without skipping the token in a complex structure, the return object may corrupt. That is why _Report error for invalid yaml header value type as array_ failed when we simply not ignore nulls.

OsmondJiang: No.
With "NullValueHandler = Ignore" This test will pass. I believe IgnoreNullValueHandler did the same thing as you did in this PR.

image

OsmondJiang: @live1206

PR #4656, opened 2019-05-29
Fix monikers of toc node

928PJY:
- If the node is a text node(`## Header`), the *monikers* of this node is the collection of its children's monikers.
- If the node is an absolute link node(`## [Header](/a)` or `## [Header](http://test)`), the *monikers* of this node is the monikers of this toc file, which means this node will always be displayed.

### DIFF
--- a/src/docfx/build/toc/BuildTableOfContents.cs
+++ b/src/docfx/build/toc/BuildTableOfContents.cs
@@ -82,11 +82,19 @@ namespace Microsoft.Docs.Build
                     }
 
                     List<string> monikers = null;
+
+                    var linkType = UrlUtility.GetLinkType(item.Href?.Value);
+                    if (linkType == LinkType.External || linkType == LinkType.AbsolutePath)
+                    {
+                        item.Monikers = fileMonikers;
+                        continue;
+                    }
+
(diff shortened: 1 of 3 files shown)


src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -82,11 +82,19 @@ void ResolveItemMonikers(List<TableOfContentsItem> items)
                     }
 
                     List<string> monikers = null;
+
+                    var linkType = UrlUtility.GetLinkType(item.Href?.Value);
+                    if (linkType == LinkType.External || linkType == LinkType.AbsolutePath)
```
yufeih: Instead of checking for link type, is it easier to detect pure text node?

928PJY: there are three cases need to check

mingwli: can we reuse [TocHrefType](https://github.com/dotnet/docfx/blob/b67e013094b9a7b868f60d86133a7e27943cd3c7/src/docfx/build/toc/parser/TableOfContentsParser.cs#L350) here?

928PJY: that require saving that info in the tocModel

PR #4659, opened 2019-05-29
👨‍🏭 respect xrefProperties to generate xref spec

OsmondJiang:
#4067


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -35,15 +36,21 @@ public JsonSchemaTransformer(JsonSchema schema)
 
             void TransformXrefScalar(JsonSchema schema, JValue value)
             {
-                extensions[value.Path] = new Lazy<JValue>(
-                    () => TransformScalar(schema, file, context, value, errors, buildChild: null),
-                    LazyThreadSafetyMode.PublicationOnly);
+                var name = value.Path.Split('.').Last();
```
live1206: property name could contain `.`, like `ms.date`

OsmondJiang: u r right!

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -54,6 +67,7 @@ private void Traverse(JsonSchema schema, JToken token, Action<JsonSchema, JValue
                 case JArray array:
                     if (schema.Items != null)
                     {
+                        schema.Items.Parent = schema.Parent;
```
mingwli: still confused about this line

OsmondJiang: go through the json schema, array is different with object

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -35,15 +37,26 @@ public JsonSchemaTransformer(JsonSchema schema)
...
+                if (schema.Parent?.XrefProperties.Contains(name) ?? false)
+                {
+                    extensions[value.Path] = new Lazy<JValue>(
+                        () => TransformScalar(schema, file, context, value, errors, buildChild: null),
+                        LazyThreadSafetyMode.PublicationOnly);
+                }
             }
         }
 
         private void Traverse(JsonSchema schema, JToken token, Action<JsonSchema, JValue> transformScalar)
```
yufeih: Use a property in a data model with `[JsonIgnore]` to pass parameters is __NOT__ thread safe. The same JsonSchema instance is shared by multiple concurrent calls to Traverse.

```suggestion
        private void Traverse(JsonSchema schema, JsonSchema parentSchema, JToken token, Action<JsonSchema, JsonSchema, JValue> transformScalar)
```

OsmondJiang: OK, but can we make it not shared with multiple concurrent calls just like the JToken passed in?

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -66,7 +71,7 @@ private void Traverse(JsonSchema schema, JToken token, Action<JsonSchema, JValue
                     {
                         if (schema.Properties.TryGetValue(key, out var propertySchema))
                         {
-                            Traverse(propertySchema, value, transformScalar);
+                            Traverse(propertySchema, value, transformScalar, schema);
```
yufeih: If you pass `key` down to traverse, than the complex `GetPropertyNameFromJsonPath` is not needed
```suggestion
                            Traverse(propertySchema, value, key, transformScalar, schema);
```

OsmondJiang: don't want to do that  since it's change too much of Traverse and GetPropertyNameFromJsonPath is also used by XrefMapBuilder

OsmondJiang:
It's a expected integration test failure, the uid will not be loaded if we can't found corresponding json schema.

image

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -22,41 +23,66 @@ public JsonSchemaTransformer(JsonSchema schema)
         public (List<Error> errors, JToken token) TransformContent(Document file, Context context, JToken token, Action<Document> buildChild)
         {
             var errors = new List<Error>();
-            Traverse(_schema, token, (schema, value) => value.Replace(TransformScalar(schema, file, context, value, errors, buildChild)));
+            Traverse(_schema, token, (_, schema, name, value) => value.Replace(TransformScalar(schema, file, context, value, errors, buildChild)));
             return (errors, token);
         }
 
-        public (List<Error> errors, Dictionary<string, Lazy<JValue>>) TransformXref(Document file, Context context, JToken token)
+        public (List<Error> errors, Dictionary<string, (bool, Dictionary<string, Lazy<JValue>>)> properties) TransformXref(Document file, Context context, JToken token)
```
live1206:
```suggestion
        public (List<Error> errors, Dictionary<string, (bool isRoot, Dictionary<string, Lazy<JValue> properties>)> xrefProperties) TransformXref(Document file, Context context, JToken token)
```

live1206: shall we follow the naming convention of the return type? it helps me to understand what these mean, especially for this kind of complex types

src/Microsoft.Docs.Template/Models/TestData.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Newtonsoft.Json;
```
yufeih: do we still need to update this test model? This could be removed since we are moving to json schema.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -22,55 +23,135 @@ public JsonSchemaTransformer(JsonSchema schema)
...
+                                xrefPropertiesGroupByUid[uid] = (isRoot, new Dictionary<string, Lazy<JToken>>());
+                            }
+                        }
+
+                        foreach (var (key, value) in obj)
+                        {
+                            if (uid != null && schema.XrefProperties.Contains(key))
+                            {
+                                var propertySchema = TryGetPropertyJsonSchema(schema, key, out var subSchema) ? subSchema : null;
+                                xrefPropertiesGroupByUid[uid].Item2[key] = new Lazy<JToken>(
```
live1206: can we do this object implementation in a separate PR?

src/Microsoft.Docs.Template/Models/TestData.cs
OsmondJiang: I will removed it in #4694  PR

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -22,57 +23,130 @@ public JsonSchemaTransformer(JsonSchema schema)
...
+                node.Replace(TransformScalar(subSchema, file, context, node as JValue, errors));
+
+                return default;
+            });
+        }
+
+        private void Traverse(JsonSchema schema, JToken token, Func<JsonSchema, JToken, string[]> transform)
         {
             schema = _definitions.GetDefinition(schema);
+            if (schema == null)
```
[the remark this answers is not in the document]
OsmondJiang: schema for xref properties may be null

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -22,57 +23,130 @@ public JsonSchemaTransformer(JsonSchema schema)
...
+            {
+                return false;
+            }
+
+            if (jsonSchema.Properties.TryGetValue(key, out propertySchema))
+            {
+                return true;
+            }
+
+            if (jsonSchema.AdditionalProperties.additionalPropertyJsonSchema != null)
```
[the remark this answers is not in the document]
OsmondJiang: AdditionProperties is supported in schema validation, so I'd like to support it in transfrom.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -22,57 +23,130 @@ public JsonSchemaTransformer(JsonSchema schema)
...
+            Traverse(_schema, token, (schema, node) =>
             {
-                extensions[value.Path] = new Lazy<JValue>(
-                    () => (JValue)TransformScalar(schema, file, context, value, errors), // todo: support JToken
-                    LazyThreadSafetyMode.PublicationOnly);
-            }
+                if (node is JObject obj)
+                {
+                    var uid = obj.TryGetValue("uid", out var uidValue) && uidValue is JValue uidJValue && uidJValue.Value is string uidStr ? uidStr : null;
+                    if (uid != null)
```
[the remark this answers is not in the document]
OsmondJiang: good point!

PR #4621, opened 2019-05-23
Resolve multiple same UID definitions with different versions

live1206:
#3607 

When there are multiple same UIDs with different versions, take the latest one respecting the document version

PR #4660, opened 2019-05-29
Refactoring and handle comparison between different products

live1206:
- Preparing for #4621 expose `Moniker` model instead of moniker name to `IXrefSpec` to be able to get additional infomation
- refactoring


src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -70,23 +70,29 @@ public MonikerProvider(Docset docset)
...
-                    error = Errors.EmptyMonikers($"No moniker intersection between docfx.yml/docfx.json and file metadata. Config moniker range '{configMonikerRange}' is '{string.Join(',', configMonikers)}', while file moniker range '{fileMonikerRange}' is '{string.Join(',', fileMonikers)}'");
+                    error = Errors.EmptyMonikers($"No moniker intersection between docfx.yml/docfx.json and file metadata. Config moniker range '{configMonikerRange}' is '{string.Join(',', configMonikers.Select(x => x.MonikerName))}', while file moniker range '{fileMonikerRange}' is '{string.Join(',', fileMonikers.Select(x => x.MonikerName))}'");
                 }
-                intersection.Sort(Comparer);
+                intersection.OrderBy(x => x.MonikerName, Comparer);
                 return (error, intersection);
             }
 
-            configMonikers.Sort(Comparer);
+            configMonikers.OrderBy(x => x.MonikerName, Comparer);
```
928PJY: this will not change the order of items in configMonikers, the return value is  sorted

live1206: You are right. why are the tests green here...

live1206: remove this line

PR #4653, opened 2019-05-28
remove siteBasePath from routes and always prepend to sitePath

928PJY:
#4222 

require changes of config migration tool @mingwli :
1. remove `siteBasePath` from routes

PR #4662, opened 2019-05-29
remove siteBasePath from routes and always prepend to sitePath 

928PJY:
From #4653


yufeih:
@mingwli , need to merge `mingwli/route-remove-base-path` to latest for integration test to pass

PR #4670, opened 2019-05-30
get inner exception message while sending github http request

OsmondJiang:
image

### DIFF
--- a/src/docfx/lib/github/GitHubAccessor.cs
+++ b/src/docfx/lib/github/GitHubAccessor.cs
@@ -117,7 +117,7 @@ query ($login: String!) {
             }
             catch (Exception ex)
             {
-                return (Errors.GitHubApiFailed(_url, ex.Message), null);
+                return (Errors.GitHubApiFailed(_url, ex.InnerException?.Message ?? ex.Message), null);
             }
         }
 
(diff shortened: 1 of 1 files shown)


OsmondJiang:
image

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -117,7 +117,7 @@ public async Task<(Error, GitHubUser)> GetUserByLogin(string login)
             }
             catch (Exception ex)
             {
-                return (Errors.GitHubApiFailed(_url, ex.Message), null);
+                return (Errors.GitHubApiFailed(_url, ex.InnerException?.Message ?? ex.Message), null);
```
yufeih: can you write these detailed diagnostic info using Log.Write?, the Error message is mainly for writers, not for troubleshooting

OsmondJiang: so what the message do you want to return to the end user?

yufeih:
```suggestion
                Log.Write(ex);
                return (Errors.GitHubApiFailed(_url, ex.Message), null);
```

OsmondJiang: if we already have the really message, why not directly return to end user? let's say we are using by some developers, they can find more info from here. And yes, we can also write to disgnostic.

yufeih: sure

Issue #4655, opened 2019-05-29
The E2E Test ‘Build with --legacy’ Failed in Chinese Environment


LowEntropyBody:
![image](https://user-images.githubusercontent.com/26829444/58526256-f85d8480-8200-11e9-9d5e-e1897c6157f1.png)

LowEntropyBody:
@yufeih

yufeih:
@LowEntropyBody would you mind taking a look and come up with a fix? the related code should be in `BuildPage.cs` line 45

PR #4673, opened 2019-05-30
Fix the bug in time format culture-affected

LowEntropyBody:
Related issue: #4655

### DIFF
--- a/src/docfx/cli/Docfx.cs
+++ b/src/docfx/cli/Docfx.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Docs.Build
                 return 1;
             }
 
-            CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = new CultureInfo("en-US", false);
+            CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = new CultureInfo("en-US");
 
             using (Log.BeginScope(options.Verbose))
             using (var errorLog = new ErrorLog(docset, options.Legacy))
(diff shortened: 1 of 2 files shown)


yufeih:
I'm thinking about setting `CultureInfo.CurrentCulture` and `CultureInfo.CurrentUICulture` globally to invariant culture at program startup, because there would be a lot of places where we call `ToString` without specifying invariant culture, and docfx as a tool doesn't really have much culture dependent text.

src/docfx/cli/Docfx.cs
```diff
@@ -62,6 +63,9 @@ internal static async Task<int> Run(string[] args)
                 return 1;
             }
 
+            CultureInfo.CurrentCulture = new CultureInfo("en-US", false);
```
yufeih:
```suggestion
            CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = new CultureInfo("en-US", false);
```

LowEntropyBody: Cool

src/docfx/cli/Docfx.cs
```diff
@@ -62,6 +63,8 @@ internal static async Task<int> Run(string[] args)
                 return 1;
             }
 
+            CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = new CultureInfo("en-US", false);
```
yufeih: what does `false` mean here, is `new CultureInfo("en-US")` enough?

LowEntropyBody: You are right

Issue #4672, opened 2019-05-30
Xrefmap failed to load for complex jsons


yufeih:
The xrefmap load optimization is buggy when JSON is a little bit complex like [this one](https://github.com/docascode/docfx-test-dependencies/blob/xref/xref1.json)

PR #4674, opened 2019-05-30
Fix for populating xref map

live1206:
#4672


test/docfx.Test/build/XrefMapLoaderTest.cs
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using Microsoft.VisualStudio.TestPlatform.ObjectModel;
```
yufeih: what is this using for?

test/docfx.Test/build/XrefMapLoaderTest.cs
```diff
@@ -0,0 +1,67 @@
...
+
+            if (File.Exists(filePath))
+            {
+                File.Delete(filePath);
+            }
+        }
+
+        private string WriteJsonToTempFile(string json)
+        {
+            var tempFilePath = Path.GetTempFileName();
```
yufeih: put these files inside a folder? we generally dont delete test generated files

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,84 @@
...
+            var bytesToRead = end - start + 1;
+            var bytes = new byte[bytesToRead];
+            stream.Position = start;
+
+            do
+            {
+                var read = stream.Read(bytes, bytesRead, (int)bytesToRead);
+                bytesRead += read;
+                bytesToRead -= read;
+            } while (bytesToRead > 0);
```
yufeih: in case of data corruption, read is 0, this loop will loop forever

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,84 @@
...
+            stream.Close();
+
```
yufeih: the stream will be closed by the using statement  in line 29, so no need to close here

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,84 @@
...
+                        break;
+                }
+            }
+            return result;
+        }
+
+        private static string ReadJsonFragment(Stream stream, long start, long end)
+        {
+            var bytesRead = 0;
+            var bytesToRead = end - start + 1;
```
yufeih: generally end means exclusive end position, suggest to +1 in line 57

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,84 @@
...
+            var reader = new Utf8JsonReader(content, isFinalBlock: true, default);
+            string uid = null;
+            int startIndex = 0;
+            int endIndex = 0;
+            while (reader.Read())
+            {
+                switch (reader.TokenType)
+                {
+                    case JsonTokenType.PropertyName:
+                        if (reader.TextEquals(Encoding.UTF8.GetBytes("uid")) && reader.Read() && reader.TokenType == JsonTokenType.String)
```
yufeih: cache utf8 uid as a static var can avoid allocation and expensive getbytes calls

src/docfx/build/xref/XrefMapBuilder.cs
```diff
@@ -66,95 +66,38 @@ public static XrefMap Build(Context context, Docset docset)
             return map;
         }
 
-        private static void DeserializeAndPopulateXrefMap(Action<string, int, int, int, int> populate, string content)
+        private static void DeserializeAndPopulateXrefMap(Action<string, int, int> populate, ReadOnlySpan<byte> content)
```
yufeih: can this method be removed?

live1206: yes

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+                switch (reader.TokenType)
+                {
+                    case JsonTokenType.PropertyName:
+                        if (reader.TextEquals(s_uidBytes) && reader.Read() && reader.TokenType == JsonTokenType.String)
+                        {
+                            uid = Encoding.UTF8.GetString(reader.ValueSpan);
+                        }
+                        break;
+                    case JsonTokenType.StartObject:
+                        startIndex = (int)reader.TokenStartIndex;
```
yufeih: this needs to be pushed to a stack for cases like, add a test case for it.
```json
{
 "uid": "a",
  "prop": { }
}
```

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+            }
+            return result;
+        }
+
+        private static string ReadJsonFragment(Stream stream, long start, long end)
+        {
+            var offset = 0;
+            var bytesRead = 0;
+            var bytesToRead = end - start;
```
yufeih: save the cast below
```suggestion
            var bytesToRead = (int)(end - start);
```

src/docfx/restore/RestoreMap.cs
```diff
@@ -223,6 +257,19 @@ public static (string path, DependencyGit git) AcquireExclusiveGit(string remote
         public static bool ReleaseGit(DependencyGit git, LockType lockType, bool successed = true)
             => DependencySlotPool<DependencyGit>.ReleaseSlot(git, lockType, successed);
 
+        private static string TryGetRestoredFilePath(string url)
```
yufeih: consider layer `GetRestoredFileContent` on top of `GetRestoredFilePath`

live1206: it would not save much logic since there are both `content` and `etag` to be prepared

OsmondJiang:
I have two concerns about this PR:
- Reading xref map files during the whole build may have read/write conflicts with restore operations. We may need make restored file read-only
- This PR is mainly to optimize `Load Xref Map`, but may cause a lot of IO operations when `Resolve UID`, especially parallel reading file content at same time.

My suggestion is to test a repo with lots of UIDs to resolve, check the performance.

src/docfx/build/xref/XrefMapLoader.cs
live1206: I thought this is not a valid case for the current `XrefSpec` model, but if xref property can be an JObject, sure...

live1206:
> I have two concerns about this PR:
> 
> * Reading xref map files during the whole build may have read/write conflicts with restore operations. We may need make restored file read-only
**_While using FileStream to read content for `uid`, it means the file should be read-only. If meanwhile another process is writing the same file, which means we can not reuse the `uid -> byte index` mapping.
It should throw._**

> * This PR is mainly to optimize `Load Xref Map`, but may cause a lot of IO operations when `Resolve UID`, especially parallel reading file content at same time.
_**https://github.com/dotnet/corefx/blob/a78bd7308555926f88aad3d0ecaaebae0f9204cc/src/Common/src/CoreLib/System/IO/FileStream.cs#L185
Let's take a look at the constructor of `FileStream`, it simply creates a [`FileHanlder`](https://github.com/dotnet/corefx/blob/a78bd7308555926f88aad3d0ecaaebae0f9204cc/src/Common/src/CoreLib/System/IO/FileStream.cs#L240), the actual IO operation should be pretty quick and then the `FileHandler` will be disposed. Basically, we want to make sure this cost is not high.**_

> My suggestion is to test a repo with lots of UIDs to resolve, check the performance.
_**Sure, we can try it out with https://github.com/OPS-E2E-PPE/docs, let's fix #4717 firstly to be able to test locally**_

OsmondJiang:
> > I have two concerns about this PR:
> > 
> > * Reading xref map files during the whole build may have read/write conflicts with restore operations. We may need make restored file read-only
> >   **_While using FileStream to read content for `uid`, it means the file should be read-only. If meanwhile another process is writing the same file, which means we can not reuse the `uid -> byte index` mapping.
> >   It should throw._**
> > > Our original design is that there are no conflicts between build and restore, so we should avoid throw here.
> 
> > * This PR is mainly to optimize `Load Xref Map`, but may cause a lot of IO operations when `Resolve UID`, especially parallel reading file content at same time.
> >   _**https://github.com/dotnet/corefx/blob/a78bd7308555926f88aad3d0ecaaebae0f9204cc/src/Common/src/CoreLib/System/IO/FileStream.cs#L185
> >   Let's take a look at the constructor of `FileStream`, it simply creates a [`FileHanlder`](https://github.com/dotnet/corefx/blob/a78bd7308555926f88aad3d0ecaaebae0f9204cc/src/Common/src/CoreLib/System/IO/FileStream.cs#L240), the actual IO operation should be pretty quick and then the `FileHandler` will be disposed. Basically, we want to make sure this cost is not high.**_
> > > The main concerns is about the paralleled reading content from disk out of order, we all known sequence reading from disk is more efficient.
> 
> > My suggestion is to test a repo with lots of UIDs to resolve, check the performance.
> > _**Sure, we can try it out with https://github.com/OPS-E2E-PPE/docs, let's fix #4717 firstly to be able to test locally**_

live1206:
Build done in 00:02:31.9510000
Tried with dotet/docs locally, no performance issue for this PR. Once we have cache in place, we do not need to load the entire xrefmap.json, the memory consumption should be reduced.

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Text;
+using System.Text.Json;
+
+namespace Microsoft.Docs.Build
+{
+    internal class XrefMapLoader
+    {
+        private static byte[] s_uidBytes = Encoding.UTF8.GetBytes("uid");
+
+        public static ListBuilder<(string, Lazy<IXrefSpec>)> Load(string filePath)
```
yufeih: consider return `List<>` over `ListBuilder<>`

src/docfx/build/xref/XrefMapLoader.cs
```diff
@@ -0,0 +1,88 @@
...
+                        if (reader.TextEquals(s_uidBytes) && reader.Read() && reader.TokenType == JsonTokenType.String && stack.TryPop(out var top))
+                        {
+                            stack.Push((Encoding.UTF8.GetString(reader.ValueSpan), top.start));
+                        }
+                        break;
+                    case JsonTokenType.StartObject:
+                        stack.Push((null, (int)reader.TokenStartIndex));
+                        break;
+                    case JsonTokenType.EndObject:
+                        if (stack.TryPop(out var item))
```
yufeih:
```suggestion
                        if (stack.TryPop(out var item) && item.uid != null)
```

src/docfx/build/xref/XrefMapLoader.cs
live1206: You are right, no need to use `ListBuilder` here, no concurrency involved

Issue #4664, opened 2019-05-29
Output is not deterministic when link and the link article has different casing


yufeih:
Found during azure-docs-pr integration testing

```diff

--- "a/D:/Docs.Build.TestData\\output\\azure-docs-pr/.dependencymap.json"
+++ "b/D:/Docs.Build.TestData\\.temp\\azure-docs-pr/.dependencymap.json"
@@ -205444,7 +205444,7 @@
         "type": "bookmark"
       },
       {
-        "source": "articles/cognitive-services/LUIS/luis-how-to-azure-subscription.md",
+        "source": "articles/cognitive-services/luis/luis-how-to-azure-subscription.md",
         "type": "link"
       },
       {
@@ -215772,7 +215772,7 @@
         "type": "link"
       },
       {
-        "source": "articles/cognitive-services/Speech-Service/speech-devices-sdk.md",
+        "source": "articles/cognitive-services/speech-service/speech-devices-sdk.md",
         "type": "link"
       },
       {
diff --git "a/D:/Docs.Build.TestData\\output\\azure-docs-pr/.errors.log" "b/D:/Docs.Build.TestData\\.temp\\azure-docs-pr/.errors.log"
index f55ec0eb..0b05f14f 100644
--- "a/D:/Docs.Build.TestData\\output\\azure-docs-pr/.errors.log"
+++ "b/D:/Docs.Build.TestData\\.temp\\azure-docs-pr/.errors.log"
@@ -1889,6 +1889,13 @@ To fix this error either change the JSON to a JSON primitive value (e.g. string,
   "file": "articles/active-directory/saas-apps/korn-ferry-alp-tutorial.md",
   "line": 150
 }
+{
+  "message_severity": "info",
+  "code": "at-xref-not-found",
+  "message": "Cross reference not found: '@contoso.com'",
+  "file": "articles/active-directory/saas-apps/kronos-tutorial.md",
+  "line": 157
+}
 {
   "message_severity": "info",
   "code": "at-xref-not-found",
@@ -2351,13 +2358,6 @@ To fix this error either change the JSON to a JSON primitive value (e.g. string,
   "file": "articles/active-directory/saas-apps/phraseanet-tutorial.md",
   "line": 147
 }
-{
-  "message_severity": "info",
-  "code": "at-xref-not-found",
-  "message": "Cross reference not found: '@contoso.com'",
-  "file": "articles/active-directory/saas-apps/pingboard-tutorial.md",
-  "line": 176
-}
 {
[...]

PR #4682, opened 2019-05-31
🧹 Extract build scope, scan scope, redirection map into context

yufeih:
We need a way to detect the real file name on file system, the simplest way is to load all files names under the current docset, this also gives us the ability to enable case insensitive file resolve on linux.

This PR is a cleanup to give us a place to manage this lookup table, to address #4664 , which is currently a blocker to integration azure-docs-pr to integration test

### DIFF
--- a/src/docfx/docset/BuildScope.cs
+++ b/src/docfx/docset/BuildScope.cs
@@ -1,0 +1,70 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Globalization;
+using System.IO;
+using System.Linq;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Docs.Build
+{
+    internal class BuildScope
+    {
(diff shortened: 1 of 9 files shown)


src/docfx/docset/BuildScope.cs
```diff
@@ -0,0 +1,75 @@
...
+            var glob = CreateGlob(docset);
+
+            var (errors, redirections) = RedirectionMap.Create(docset, glob);
+            var files = GetFiles(docset, glob).Concat(redirections.Files).ToHashSet();
+            var filesWithFallback = files;
+
+            if (docset.FallbackDocset != null)
+            {
+                filesWithFallback = files.Concat(
+                    GetFiles(docset.FallbackDocset, CreateGlob(docset.FallbackDocset))).ToHashSet();
```
OsmondJiang: the key is not Document, is the file path to docset

OsmondJiang:
I didn't see what is to resolve in this PR block the azure-docs-pr integration?

OsmondJiang:
I think the docset should own its build scope and scan scope, what's the standard here? why dependency repositories still exists in docset?

src/docfx/docset/BuildScope.cs
yufeih: why no test failed?

Issue #4605, opened 2019-05-22
Json Schema Validation For Metadata


LowEntropyBody:
- [x] Json Schema Validation
  - [x] type
  - [x] enum
  - [x] required
  - [x] dependencies
  - [x] either
  - [x] precludes
  - [x] dateFormat
  - [x] maxDate
  - [x] minDate
  - [x] replacedby
  - [x] microsoftAlias
  - [x] dependentEnums
- [x] Convert Original Schema to Json Schema
  - [x] get ruleset
  - [x] get allowlists
  - [x] convertion
- [x] Handle Error Message
  - [x] rules errors
  - [x] additional errors
- [ ] Online
  - [ ] joint debugging and testing
  - [ ] optimization and online of the entire metadata validation

PR #4685, opened 2019-05-31
Metadata validation extension rules: 'dateFormat', 'dateRange', 'deprecated'

LowEntropyBody:
Related issue: #4605

Ps: For 'dateRange' rule, i don't add tests. Because this rule does not give complete information in the original validation service and has not been used. 
for eample:
```json
{
    "relativeMax": "5:00:00:00" - 5 days in the future,
    "relativeMin": "-2:00:00" - 2 hours in the past
}
```
It uses relative time to build the date range, which causes the value to expire regardless of the date value set in metadata in the future. I will discuss this issue with PM later. This pr only implements the function of dateRange rule.


src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -64,13 +66,15 @@ private static bool ValidateType(JsonSchema schema, JToken token, List<Error> er
...
 
-        private static void ValidateScalar(JsonSchema schema, List<Error> errors, JValue scalar)
+        private void ValidateScalar(JsonSchema schema, JValue scalar, List<Error> errors)
         {
             if (schema.Enum != null && !schema.Enum.Contains(scalar))
             {
                 errors.Add(Errors.UndefinedValue(JsonUtility.GetSourceInfo(scalar), scalar, schema.Enum));
             }
 
+            ValidateDateFormat(schema, scalar, errors);
```
yufeih: can put this under the if check in line 74, then you can access `str` without using `ToString()`. Also consider extract `ValidateFormat` and `ValidateLength`

LowEntropyBody: Sure, good suggestion

LowEntropyBody:
Update: Jason gave a good suggestion. Since the dateRange rule can't be used for successful tests, I will add a few tests beyond the range of dateRange.

src/docfx/lib/schema/JsonSchema.cs
```diff
@@ -129,5 +129,25 @@ internal class JsonSchema
...
+
+        /// <summary>
+        /// Properties that are used to set minimum time range
+        /// </summary>
+        public TimeSpan? RelativeMinDate { get; set; }
+
+        /// <summary>
+        /// Properties that are used to indicate the deprecated field
+        /// </summary>
+        public JValue ReplacedBy { get; set; }
```
mingwli: >JValue Repl [](start = 15, length = 11)

why JValue, not string?

LowEntropyBody: Because replaceBy can be set "", this means that field has been deprecated, but there is no other field to replace it. you can take a look at test for more details.

yufeih: A `string` can also be null or empty

LowEntropyBody: OK, I will fix this

LowEntropyBody:
Work item link: [100602](https://dev.azure.com/ceapex/Engineering/_workitems/edit/100602)

PR #4686, opened 2019-05-31
Sdp/resolve schema

mingwli:
#4067

e2e test load schema from default template folder


src/docfx/build/page/BuildPage.cs
```diff
@@ -183,7 +183,7 @@ private static (List<Error> errors, Schema schema, OutputModel model)
                 throw Errors.SchemaNotFound(file.Mime).ToException();
             }
 
-            var (schemaValidator, schemaTransformer) = TemplateEngine.GetJsonSchema(file.Schema);
+            var (schemaValidator, schemaTransformer) = context.Template.GetJsonSchema(file.Schema);
```
OsmondJiang: check if it's null

mingwli:
will continue with #4753

PR #4687, opened 2019-05-31
Add Feature Requirements of XREF

live1206:


docs/designs/xref.md
```diff
@@ -4,6 +4,21 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+- Define an internal UID with versioning and reference to this UID without versioning within the current repository
+- Define multiple UID with same value internally with different versioning and reference to this UID without versioning within the current repository
+    - If all versionings for this UID are within the same product, take the one with highest versioning respecting the referencing file
+    - If all versionings for this UID are from different products, take the highest one base on proudct namme alphabetically
+- Reference to an external UID without versioning
+    - The href of UID is from the same host name as the referencing repository
+    - The href of UID is from a different host name as the referencing repository
+        - If the current branch is live branch, the outout path could be 'review.docs', but the resolved url is 'docs', it will go to another site clicking on the url
+        - If the current branch is from a different site other than 'docs', we need to prioritize the uid from this site firstly if same uid defined in this site and 'docs' 
+- Reference to an external UID with versioning
```
928PJY: does this item mean the referenced Uid has just one version?

live1206: The next item(UID with multiple versionings) is not in scope for versioning during migration from v2 to v3

docs/designs/xref.md
```diff
@@ -4,6 +4,143 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+            ---
+            title: Title from v2
+            uid: a
+            ---
+            docs/b.md: Link to @a
+        outputs:
+            docs/b.json: |
+                {"conceptual":"<p>Link to <a href=\"a\">Title from v2</a></p>\n"}
+        ```
+- Reference to an external UID without versioning
```
yufeih: Its great to list the status qoo, I would be more interested in why things are implemented the way it is, e.g. because of some limitations in the hosting layer. Bottom line is we implement these as is and later rework the whole fallback story with hosting together

docs/designs/xref.md
```diff
@@ -4,6 +4,152 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+    title: Title from yaml header a
+    uid: a
+    description: some description
+    ---
+    docs/b.md: Link to @a
+  outputs:
+    docs/b.json: |
+        {"conceptual":"<p>Link to <a href=\"a\">Title from yaml header a</a></p>\n"}
+  ```
+  > **_Notice_**: Only title will be considered as xref properrty for `uid` definition in `.md` files
```
928PJY: typo

928PJY: >properrty [](start = 56, length = 9)

typo

docs/designs/xref.md
```diff
@@ -4,6 +4,152 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+- Define multiple UID with same value internally with different versions and reference to this UID without versioning within the current repository
+    - If all versions for this UID are within the same product, take the one with highest versioning respecting the referencing file
+      ```yaml
+        # v1: 1.0, 2.0, 3.0
+        # v2: 4.0, 5.0
+        # v3: 3.0
```
928PJY: > v3: 3.0  [](start = 9, length = 9)

what if the monikers of v3 folder is `4.0`

live1206: we can come up with more tests when the PR is out

docs/designs/xref.md
```diff
@@ -4,6 +4,152 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+            ---
+            title: Title from v1
+            uid: a
+            ---
+            docs/v2/a.md: |
+            ---
+            title: Title from v2
+            uid: a
+            ---
+            docs/v3/b.md: Link to @a
```
herohua: >Link to @a [](start = 26, length = 10)

actually, it depends on the metadata moniker range of the file and/or the moniker range Markdig extension. 

below samples are correct, but may be a bit simplified. we may need some detailed write-up of more cases.

herohua: and this part only applies to internal XREF

---
In reply to: [295599235](https://github.com/dotnet/docfx/pull/4687#discussion_r295599235) [](ancestors = 295599235)

docs/designs/xref.md
```diff
@@ -4,6 +4,152 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+            ---
+            title: Title from v2
+            uid: a
+            ---
+            docs/b.md: Link to @a
+        outputs:
+            docs/b.json: |
+                {"conceptual":"<p>Link to <a href=\"a\">Title from v2</a></p>\n"}
+        ```
+    - If the versions for this UID are from different products, take the highest one base on proudct name alphabetically
```
herohua: >take the highest one base on proudct name alphabetically [](start = 64, length = 56)

is this correct? I thought we should follow fallback order (not necessarily the alphabetical order)

docs/designs/xref.md
```diff
@@ -4,6 +4,152 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+            ---
+            title: Title from azure-cn
+            uid: a
+            ---
+            azure-cn/b.md: Link to @a
+          outputs:
+            azure-cn/b.json: |
+                {"conceptual":"<p>Link to <a href=\"a\">Title from azure-cn</a></p>\n"}
+          ```
+- Reference to an external UID with versioning, the resolving logic should be the same as internal UID without versioning. We only take consideration of 1 version for now.
```
herohua: >consideration of 1 version for now. [](start = 136, length = 35)

and do not output `view=?`

docs/designs/xref.md
```diff
@@ -4,6 +4,152 @@ Besides using file path to link to another file, DocFX also allows you to give a
...
+- Reference to an external UID with multiple versionings (not support for now)
+    - The resolve logic should be smilar to the internal resolving
+- [UID defnition](#define-uid)
+- Needs to check if we have any conceptual repo output non-empty xrefmap.json. If yes, and auto-complete in xref service is still in use xref service needs to consume `xrefmap.json` instead of `xrefmap.yml` before v3 xref on-boarding.
```
herohua: >Needs to check if we have any conceptual repo output non-empty xrefmap.json. If yes, and auto-complete in xref service is still in use xref service needs to consume `xrefmap.json` instead of `xrefmap.yml` before v3 xref on-boarding. [](start = 2, length = 232)

[https://github.com/aspnet/AspNetCore.Docs/search?q=uid&unscoped_q=uid](https://github.com/aspnet/AspNetCore.Docs/search?q=uid&unscoped_q=uid)

docs/designs/xref.md
herohua: >If the versions for this UID are from different products [](start = 6, length = 56)

what if we have overlapping of some products, but not all

docs/designs/xref.md
live1206: right, adding it

docs/designs/xref.md
live1206: Trying to add complete requirements to this file, so yes, this part is talking about internal UID, adding h3 for it.
And just want to have a simple example to explain the case, not having all cases. We should have more test cases in our code to cover more scenarios.

docs/designs/xref.md
live1206: Yes, it should only consider the products that overlap

PR #4688, opened 2019-05-31
💣 Fix a critical test validation bug and corresponding cases

yufeih:
💣💣💣💣💣💣💣💣💣💣💣💣💣💣💣💣💣💣💣💣

@928PJY , there is a moniker test case that produces non-deterministic output, I've marked it as [skip]

There is another SDP html `<script>` removal test that I marked as [skip], it will be fixed in #4588

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -523,7 +523,7 @@ outputs:
           { "name": "charlie", "profile_url": "https://github.com/charlie", "display_name": "Charlie", "id": "3" },
         "contributors": [
           { "name": "bob", "profile_url": "https://github.com/bob", "display_name": "Bob", "id": "2" },
-          { "name": "OsmondJiang", "profile_url": "https://github.com/OsmondJiang", "display_name": "osmondjiang", "id": "19990166" }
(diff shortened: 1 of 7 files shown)


docs/specs/git.yml
```diff
@@ -523,7 +523,7 @@ outputs:
           { "name": "charlie", "profile_url": "https://github.com/charlie", "display_name": "Charlie", "id": "3" },
         "contributors": [
           { "name": "bob", "profile_url": "https://github.com/bob", "display_name": "Bob", "id": "2" },
-          { "name": "OsmondJiang", "profile_url": "https://github.com/OsmondJiang", "display_name": "osmondjiang", "id": "19990166" }
+          { "name": "Osmond Jiang", "profile_url": "https://github.com/OsmondJiang", "display_name": "osmondjiang", "id": "19990166" }
```
OsmondJiang: change the display name instead of name

yufeih:
Linux build recently has a lot of GitHub API related failures:

```
Call to GitHub API 'https://api.github.com/graphql' failed: The server returned an invalid or unrecognized response. Try closing and reopening the PR. If you get this Error again, file an issue.
```

PR #4690, opened 2019-06-03
Some user doesn't have name, use login instead

OsmondJiang:
image
fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/95561

### DIFF
--- a/src/docfx/lib/github/GitHubAccessor.cs
+++ b/src/docfx/lib/github/GitHubAccessor.cs
@@ -118,6 +118,7 @@ query ($login: String!) {
             }
             catch (Exception ex)
             {
+                Log.Write(ex);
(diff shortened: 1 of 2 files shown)


yufeih:
The integration test of this PR passed, so we have no users in our integration test having this issue.

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -244,7 +246,7 @@ public async Task<(Error, IEnumerable<GitHubUser>)> GetUsersByCommit(string owne
                             {
                                 Id = node.author.user?.databaseId,
```
yufeih: are there any case where `author.user` is null?
```suggestion
                                Id = node.author.user.databaseId,
```

PR #4692, opened 2019-06-03
update axios

superyyrrzz:


superyyrrzz:
I reverted this change as it will meet below error:
```
Error: Request body larger than maxBodyLength limit
```

This upgrade originally aims to fix a DoS vulnerability. However, this is not the case when we use it as deployment script, instead of handling user inputs.

Issue #4705, opened 2019-06-10
Test case "Path is case insensitive on windows and mac" failed on Linux


OsmondJiang:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=59686 cc @yufeih

PR #4706, opened 2019-06-10
Fix linux test failure

yufeih:
#4705

### DIFF
--- a/docs/specs/casing.yml
+++ b/docs/specs/casing.yml
@@ -87,6 +87,7 @@ outputs:
   docs/B.json:
 ---
 # Use real file casing in manifest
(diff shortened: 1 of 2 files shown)


docs/specs/casing.yml
```diff
@@ -1,5 +1,4 @@
-# Path is case insensitive on windows and mac
```
OsmondJiang: is that a break change in v3?

OsmondJiang: if we want to all are case insensitive, we need re-design it, a lot of comparing logic is also applied too.

yufeih: sure, defer that change

PR #4710, opened 2019-06-11
disable incremental when schema changes

superyyrrzz:
TODO: add test


src/Microsoft.DocAsCode.Build.SchemaDriven/Models/DocumentSchema.cs
```diff
@@ -97,6 +97,8 @@ public static DocumentSchema Load(TextReader reader, string title)
 
                 schema.MetadataReference = pointer;
                 schema.AllowOverwrite = CheckOverwriteAbility(schema);
+                schema.Hash = new JTokenEqualityComparer().GetHashCode(jObject).ToString();
```
herohua: >GetHashCode [](start = 59, length = 11)

this may not be safe enough for incremental hash

PR #4702, opened 2019-06-10
resolve toc node moniker inside LoadToc function since the document info is lost after LoadToc

928PJY:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/96568/
Since we resolve the moniker of toc node in a post processor, we just have the href resolved result and lost the document info, we cannot get the accurate moniker resolve result.

Sample:
# toc.md
```
# [A-v1](v1/a.md)
# [A-v2](v2/a.md)
```
In the post processor, we cannot distinguish those two nodes, the moniker resolved result will be the same for those two nodes which have a different version.

there are two solution:
- add document property to TOCItem Model
- resolve moniker inside the LoadToc function

PR #4711, opened 2019-06-11
Consolidate MetadataProvider and MonikerProvider

yufeih:
#4702, #4605 , https://ceapex.visualstudio.com/Engineering/_workitems/edit/96568/

- Cache metadata in `MetadataProvider`.
- Simplify `MonikerProvider` interface to only expose one `GetFileLevelMonikers` method, cache monikers internally.
- Fix places where `Markup` is called without passing `MonikerProvider`

### DIFF
--- a/src/docfx/build/metadata/MetadataProvider.cs
+++ b/src/docfx/build/metadata/MetadataProvider.cs
@@ -2,8 +2,9 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using Newtonsoft.Json.Linq;
 
@@ -11,13 +12,18 @@ namespace Microsoft.Docs.Build
 {
     internal class MetadataProvider
     {
+        private readonly Cache _cache;
         private readonly JsonSchemaValidator _schemaValidator;
         private readonly JObject _globalMetadata;
         private readonly HashSet<string> _reservedMetadata;
         private readonly List<(Func<string, bool> glob, string key, JToken value)> _rules = new List<(Func<string, bool> glob, string key, JToken value)>();
 
-        public MetadataProvider(Docset docset)
+        private readonly ConcurrentDictionary<Document, (List<Error> errors, OutputModel metadata)> _metadataCache
+                   = new ConcurrentDictionary<Document, (List<Error> errors, OutputModel metadata)>();
+
+        public MetadataProvider(Docset docset, Cache cache)
         {
+            _cache = cache;
             _schemaValidator = new JsonSchemaValidator(docset.MetadataSchema);
             _globalMetadata = docset.Config.GlobalMetadata;
 
@@ -35,15 +41,22 @@ namespace Microsoft.Docs.Build
             }
         }
 
-        public (List<Error> errors, T metadata) GetInputMetadata<T>(Document file, JObject yamlHeader = null) where T : InputMetadata
+        public (List<Error> errors, OutputModel metadata) GetMetadata(Document file)
         {
-            Debug.Assert(file != null);
+            return _metadataCache.GetOrAdd(file, GetMetadataCore);
+        }
+
+        private (List<Error> errors, OutputModel metadata) GetMetadataCore(Document file)
+        {
+            if (file.ContentType != ContentType.Page && file.ContentType != ContentType.TableOfContents)
+            {
+                return (new List<Error>(), new OutputModel());
+            }
 
             var result = new JObject();
-            if (yamlHeader != null)
-            {
-                JsonUtility.SetSourceInfo(result, JsonUtility.GetSourceInfo(yamlHeader));
-            }
+
+            var (errors, yamlHeader) = LoadMetadata(file);
+            JsonUtility.SetSourceInfo(result, JsonUtility.GetSourceInfo(yamlHeader));
 
             JsonUtility.Merge(result, _globalMetadata);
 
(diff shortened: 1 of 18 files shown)


src/docfx/build/xref/ExternalXRefSpec.cs
```diff
@@ -15,10 +15,8 @@ internal class ExternalXrefSpec : IXrefSpec
 
         Document IXrefSpec.DeclairingFile => null;
 
-        // TODO: need a lookup table of moniker definition to get Moniker from moniker name
-        // not into output for now
         [JsonIgnore]
-        public HashSet<Moniker> Monikers { get; set; } = new HashSet<Moniker>();
```
928PJY: conflict with this [PR](https://github.com/dotnet/docfx/pull/4621)

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -2,20 +2,26 @@
...
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal class MonikerProvider
     {
-        private readonly List<(Func<string, bool> glob, (string monikerRange, IEnumerable<Moniker> monikers))> _rules = new List<(Func<string, bool>, (string, IEnumerable<Moniker>))>();
+        private readonly List<(Func<string, bool> glob, (string monikerRange, IEnumerable<string> monikers))> _rules = new List<(Func<string, bool>, (string, IEnumerable<string>))>();
```
live1206: why is this changed from `Moniker` to `string` again? Needed in #4621

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -98,7 +98,7 @@ private JValue TransformScalar(JsonSchema schema, Document file, Context context
                         file,
                         dependencyResolver,
                         buildChild,
-                        null,
+                        context.MonikerProvider,
```
928PJY: >context.MonikerProvider [](start = 24, length = 23)

we haven't design the version for sdp, keep this unchanged?

yufeih: Versioning should be the same for both sdp and conceptual, or the markdown pipeline should be the same for sdp and conceptual.

src/docfx/build/moniker/MonikerProvider.cs
yufeih: to move and hide moniker resolve logic inside `MonikerProvider`

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -98,5 +108,46 @@ private static bool IsValidMetadataType(JToken token)
...
+            {
+                return LoadSchemaDocumentMetadata(_cache.LoadYamlFile(file), file.FilePath);
+            }
+
+            if (file.FilePath.EndsWith(".json", PathUtility.PathComparison))
+            {
+                return LoadSchemaDocumentMetadata(_cache.LoadJsonFile(file), file.FilePath);
+            }
+
+            return (new List<Error>(), new JObject());
```
live1206:
```suggestion
            return default;
```

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -28,28 +34,47 @@ public MonikerProvider(Docset docset)
...
 
-            var (monikerError, monikers) = GetFileLevelMonikers(file, metadata.MonikerRange);
-            errors.AddIfNotNull(monikerError);
+            var zoneLevelMonikers = _rangeParser.Parse(rangeString);
+            var monikers = fileLevelMonikers.Intersect(zoneLevelMonikers, StringComparer.OrdinalIgnoreCase).ToList();
 
-            return (errors, monikers);
+            if (monikers.Count == 0)
+            {
+                var error = Errors.EmptyMonikers($"No intersection between zone and file level monikers. The result of zone level range string '{rangeString}' is '{string.Join(',', zoneLevelMonikers)}', while file level monikers is '{string.Join(',', fileLevelMonikers)}'.");
```
live1206: Do we have criteria when we should pass an error message and when it should be prepared within the `Error` itself?
Because with this way, it would be harder for us to gather the error messages for PM to review

928PJY: this is a special case, because we want to use the same code `empty-moniker` for the config/file and file/zone empty moniker intersection, but with different message, so we pass an error message here.

live1206: The message can still be prepared

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -35,16 +41,23 @@ public MetadataProvider(Docset docset)
...
+        }
 
-            var result = new JObject();
-            if (yamlHeader != null)
+        private (List<Error> errors, OutputModel metadata) GetMetadataCore(Document file)
+        {
+            if (file.ContentType != ContentType.Page && file.ContentType != ContentType.TableOfContents)
             {
-                JsonUtility.SetSourceInfo(result, JsonUtility.GetSourceInfo(yamlHeader));
+                return (new List<Error>(), new OutputModel());
```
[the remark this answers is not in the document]
yufeih: default would return a null error list, and caller would need to check null before calling `AddRange`. I also think the currently error list propagation approach is not ideal, created an [issue](https://dev.azure.com/ceapex/Engineering/_workitems/edit/98621) to track.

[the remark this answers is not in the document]
live1206: right

src/docfx/build/moniker/MonikerProvider.cs
yufeih: @928PJY will you do this refactor later?

928PJY: OK, leave it to me.

PR #4712, opened 2019-06-11
add locale checker before appending locale to absolute url

OsmondJiang:
fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/98130/

### DIFF
--- a/src/docfx/localization/LocalizationUtility.cs
+++ b/src/docfx/localization/LocalizationUtility.cs
@@ -2,17 +2,23 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.Docs.Build
 {
     internal static class LocalizationUtility
     {
+        private static readonly HashSet<string> s_locales = new HashSet<string>(CultureInfo.GetCultures(CultureTypes.AllCultures).Except(CultureInfo.GetCultures(CultureTypes.NeutralCultures)).Select(c => c.Name), StringComparer.OrdinalIgnoreCase);
         private static readonly Regex s_nameWithLocale = new Regex(@"^.+?(\.[a-z]{2,4}-[a-z]{2,4}(-[a-z]{2,4})?|\.loc)?$", RegexOptions.IgnoreCase);
         private static readonly Regex s_lrmAdjustment = new Regex(@"(^|\s|\>)(C#|F#|C\+\+)(\s*|[.!?;:]*)(\<|[\n\r]|$)", RegexOptions.IgnoreCase);
+
+        public static bool IsValidLocale(string locale)
+            => s_locales.Contains(locale);
 
         public static string AddLeftToRightMarker(CultureInfo culture, string text)
         {
--- a/src/docfx/localization/LocalizationUtility.cs
+++ b/src/docfx/localization/LocalizationUtility.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Docs.Build
 {
     internal static class LocalizationUtility
     {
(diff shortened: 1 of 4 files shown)


src/docfx/localization/LocalizationUtility.cs
```diff
@@ -2,18 +2,24 @@
...
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.Docs.Build
 {
     internal static class LocalizationUtility
     {
+        private static readonly HashSet<string> s_locales = new HashSet<string>(CultureInfo.GetCultures(CultureTypes.AllCultures).Except(CultureInfo.GetCultures(CultureTypes.NeutralCultures)).Select(c => c.Name), StringComparer.OrdinalIgnoreCase);
         private static readonly Regex s_nameWithLocale = new Regex(@"^.+?(\.[a-z]{2,4}-[a-z]{2,4}(-[a-z]{2,4})?|\.loc)?$", RegexOptions.IgnoreCase);
```
yufeih: can `s_nameWIthLocale` also share the same local list?

src/docfx/localization/LocalizationUtility.cs
```diff
@@ -2,18 +2,24 @@
...
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.Docs.Build
 {
     internal static class LocalizationUtility
     {
+        private static readonly HashSet<string> s_locales = new HashSet<string>(CultureInfo.GetCultures(CultureTypes.AllCultures).Except(CultureInfo.GetCultures(CultureTypes.NeutralCultures)).Select(c => c.Name), StringComparer.OrdinalIgnoreCase);
```
yufeih: is `Contains('-')` a better condition to ensure we don't include strings without `-` into the locale list?

src/docfx/localization/LocalizationUtility.cs
OsmondJiang: yes we can, I will do it in another PR, it ends with this locale list

src/docfx/localization/LocalizationUtility.cs
herohua: I am wondering how is this list compared to the actual list Docs supports...

OsmondJiang: I think it's the same from the summary:
```
  Cultures that are associated with a language but are not specific to a country/region.
        //     The names of .NET Framework cultures consist of the lowercase two-letter code
        //     derived from ISO 639-1. For example: &quot;en&quot; (English) is a neutral culture.
        NeutralCultures = 1,

Comment (OsmondJiang) [reply src/docfx/localization/LocalizationUtility.cs]: > I am wondering how is this list compared to the actual list Docs supports...

I added docs support locales to the test

**Comment by OsmondJiang** [comment]
expected diff:

image

File: src/docfx/localization/LocalizationUtility.cs (hunk 2 above)
Comment (yufeih) [reply src/docfx/localization/LocalizationUtility.cs]: I would like to suggest we start with the docs locale list

Comment (OsmondJiang) [reply src/docfx/localization/LocalizationUtility.cs]: don't want to add docs' limitation to docfx.

File: src/docfx/localization/LocalizationUtility.cs [hunk 3]
```diff
@@ -2,18 +2,24 @@
...
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.Docs.Build
 {
     internal static class LocalizationUtility
     {
+        private static readonly HashSet<string> s_locales = new HashSet<string>(CultureInfo.GetCultures(CultureTypes.AllCultures).Except(CultureInfo.GetCultures(CultureTypes.NeutralCultures)).Select(c => c.Name).Concat(new[] { "zh-cn", "zh-tw", "zh-hk", "zh-sg", "zh-mo" }), StringComparer.OrdinalIgnoreCase);
```
live1206: It would be helpful to have some comments why we concat some additional locales here

PR #4713, opened 2019-06-11
Remove error level info

yufeih:
Fixes https://dev.azure.com/ceapex/Engineering/_git/Docs.Build.TestData/commit/be67d9a7ac58a8c2f29918ee78422e2514372a8d?refName=refs%2Fheads%2Fmaster

azure-docs-pr test is unstable because there are too much info messages in `.errors.log`, this PR turns all info to off, so they do not appear in `.errors.log`. To keep existing tests intact, they remain info during tests.

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -220,27 +220,11 @@ namespace Microsoft.Docs.Build
             => new Error(ErrorLevel.Error, "json-syntax-error", message, source);
 
         /// <summary>
-        /// Used empty link in article.md.
-        /// Examples:
-        ///   - [link]()
-        /// </summary>
-        /// Behavior: ✔️ Message: ❌
-        public static Error LinkIsEmpty(Document relativeTo)
-            => new Error(ErrorLevel.Off, "link-is-empty", "Link is empty", relativeTo.ToString());
-
-        /// <summary>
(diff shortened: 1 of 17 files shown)


OsmondJiang:
I don't think it's good to change our error level for testing purpose.

src/docfx/Errors.cs
```diff
@@ -259,11 +243,11 @@ public static Error LocalFilePath(Document relativeTo, string path)
             => new Error(ErrorLevel.Warning, "local-file-path", $"Link '{path}' points to a local file. Use a relative path instead", relativeTo.ToString());
 
         /// <summary>
-        /// The fisrt tag in an article.md isn't h1 tag.
+        /// The first tag in an article.md isn't h1 tag.
         /// </summary>
         /// Behavior: ✔️ Message: ❌
         public static Error HeadingNotFound(Document file)
-            => new Error(ErrorLevel.Info, "heading-not-found", $"The first visible block is not a heading block with `#`, `##` or `###`", file.ToString());
+            => new Error(ErrorLevel.Off, "heading-not-found", $"The first visible block is not a heading block with `#`, `##` or `###`", file.ToString());
```
yufeih: this is also kept because there will be H1 validation shortly

PR #4718, opened 2019-06-11
Render original content for short xref inline inside links

yufeih:
Fixes https://dev.azure.com/ceapex/Engineering/_workitems/edit/87621/

@yishengjin1413

### DIFF
--- a/test/Microsoft.DocAsCode.MarkdigEngine.Tests/XrefTest.cs
+++ b/test/Microsoft.DocAsCode.MarkdigEngine.Tests/XrefTest.cs
@@ -17,7 +17,12 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Tests
 @Microsoft.Build.Tasks
 ""@""a[test](link)
 @hehe
+
 ![img @hehe a](src ""title @hehe b"")
+![img @hehe a
+![img @hehe a](src ""title @hehe b""
+![this **img @hehe a** haha](src ""title @hehe b"")
+
 @""Microsoft.Build.Tasks?text=Tasks""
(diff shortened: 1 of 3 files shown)


test/Microsoft.DocAsCode.MarkdigEngine.Tests/XrefTest.cs
```diff
@@ -17,6 +17,7 @@ public void XrefTestGeneral()
 @Microsoft.Build.Tasks
 ""@""a[test](link)
 @hehe
+![img @hehe a](src ""title @hehe b"")
```
yishengjin1413: Would you also add cases like "![img @hehe a" and `![img @hehe a](src ""title @hehe b""`

yishengjin1413: Also `this **img @hehe a** haha`

yishengjin1413: @yufeih From the test case, For "![img @hehe a", the XrefInlineShortParser will not work also. Is there any way to fix this case?

yufeih:
The trick is during render, `LinkInline` will use an `HtmlRenderer` with `EnableHtmlForInline = false`. The cost to fix it does not justify the reason, thus close this.

test/Microsoft.DocAsCode.MarkdigEngine.Tests/XrefTest.cs
yufeih: Fixed in Renderer

Issue #4488, opened 2019-05-07
PDF Build fails in Azure DevOps


ChristopherDonnelly:
**Operation System**: Windows

**DocFX Version Used**: DocFx Build Tasks - [Visual Studio Marketplace Extension](https://marketplace.visualstudio.com/items?itemName=chrismason.vsts-docfxtasks)

**Template used**: pdf.default

**Additional Information**: I am able to create the PDF locally just not in a build pipeline in Azure DevOps.

**Steps to Reproduce**:
1. Get Sources from repository
2. Add Chocolatey task to Install wkhtmltopdf
3. Run Create DocFx task to create PDF
4. Project File "docfx_pdf.json"  <-- Just testing PDF build.  Content builds work correctly.
```
"pdf": {
        "content": [
            {
                "files": [
                    "**/*.md",
                    "**/**.yml"
                ],
                "exclude": [
                    "**/toc.yml",
                    "**/toc.md"
                ]
            },
            {
                "files": "pdf/toc.yml"
            }
        ],
        "resource": [
            {
                "files": [
                    "**/*.png",
                    "**/*.jpg",
                    "**/*.gif",
                    "**/*.pdf",
                    "**/*.pptx",
                    "favicon.ico"
                ],
                "exclude": [
                    "**/obj/**",
                    "**/includes/**",
                    "**/_site_pdf/**"
                ]
            }
        ],
        "overwrite": [],
        "dest": "_site_pdf"
    }
```

**Expected Behavior**:
Create PDF file

**Actual Behavior**:
PDF file is not create with these errors in the Build Pipeline.

```
[19-05-06 06:59:08.197]Error:[PdfCommand.PDF]Error happen when converting pdf/toc.json to Pdf. Details: System.InvalidOperationException: The file is not a valid PDF document.
   at PdfSharp.Pdf.IO.PdfReader.Open(Stream stream, String password, PdfDocumentOpenMode openmode, PdfPasswordProvider passwordProvider)
   at Microsoft.DocAsCode.HtmlToPdf.HtmlToPdfConverter.SaveCore(Stream stream)
   at Microsoft.DocAsCode.HtmlToPdf.HtmlToPdfConverter.Save(String outputFileName)
   at Microsoft.DocAsCode.HtmlToPdf.ConvertWrapper.<>c__DisplayClass7_0.<ConvertCore>b__1(ManifestItem tocFile)
	24 Warning(s)
	1 Error(s)
[...]

PR #4719, opened 2019-06-11
new pdf option noStdin

oleksabor:
In order to deal with #4488 I'd like to suggest to add new `noStdin` pdf option 
It can be set using `docfx.json` file to run `docfx pdf` on the Azure build pipeline.

````Json
"pdf": {
 "noStdin": true,    
"content": [

````

### CODE REVIEW

File: src/Microsoft.DocAsCode.HtmlToPdf/ConvertWrapper.cs [hunk 1]
```diff
@@ -328,15 +328,16 @@ private IList<HtmlModel> BuildHtmlModels(string basePath, IList<TocModel> tocMod
...
-            var converter = new HtmlToPdfConverter(
-                htmlModels,
-                new HtmlToPdfOptions
-                {
-                    BasePath = basePath,
-                    UserStyleSheet = _pdfOptions.CssFilePath,
-                    LoadErrorHandling = _pdfOptions.LoadErrorHandling,
-                    AdditionalArguments = _pdfOptions.AdditionalPdfCommandArgs,
-                    OutlineOption = _pdfOptions.OutlineOption
+			var converter = new HtmlToPdfConverter(
```
superyyrrzz: Replace <kbd>tab</kbd> with 4 <kbd>space</kbd>s

superyyrrzz:
This PR makes me think whether do we need this option. Could we always disable `IsReadArgsFromStdin`?

/cc @partychen

src/Microsoft.DocAsCode.HtmlToPdf/ConvertWrapper.cs
oleksabor: Please find tabs were replaced with spaces in the commit af6b07fe862fd572a7ee76b5607bd3e6bf273816

superyyrrzz:
@oleksabor I just talked with @partychen . Looks like the root cause is stderr in wkhtmltopdf is not handled correctly. I will soon create another PR to fix this. We can first hold on this PR and see if that PR can work for you.

oleksabor:
@superyyrrzz I see, thank you for information

I'm not in a hurry since have patched docfx using v2.42.4 for Azure pipelines. 

However I'd like to add that had tried to read from the `wkhtmltopdf`s `stderr` and failed - there is no output even if it returns 1
I've used `RedirectStandardError = config.RedirectError,` to redirect stderr and then read it as

````CSharp
if (IsOutputToStdout)
{
	Log.Debug("reading stdout");
	using (var standardOutput = process.StandardOutput)
	{
		//standardOutput.BaseStream.CopyTo(stream);
		using (var reader = new StreamReader(standardOutput.BaseStream))
		{
			bufStr = reader.ReadToEnd();
		}
	}
}
if (config.RedirectError)
{
	Log.Debug("reading stderr");
	using (var sout = process.StandardError)
	{
		using (var reader = new StreamReader(sout.BaseStream))
		{
			var error = reader.ReadToEnd();
			if (!string.IsNullOrEmpty(error))
				throw new ApplicationException(error);
		}
	}
}
var exited = process.WaitForExit(TimeoutInMilliseconds);
````

`var error = reader.ReadToEnd();` returns `null` even if `exited` is not zero and `wkhtmltopdf` does not output pdf.

I've tried to run `wkhtmltopdf` using Azure pipeline and it there is nothing in the `stderr` as well.

**Comment by superyyrrzz** [comment]
@oleksabor I created #4733. Could you help review and try whether it can fix your issue? From your latest comment, I feel this is what you have tried. I am not sure whether it can work.

**Comment by superyyrrzz** [comment]
Sorry for the late reply. I see this is a feasible fix for this issue, while there are still some changes required for this PR:
* Replace all tabs with 4 spaces
* "Update branch" to pick up latest changes in dev branch
* Update the table in `Documentation/tutorial/docfx.exe_user_manual.md` for this new option: https://dotnet.github.io/docfx/tutorial/docfx.exe_user_manual.html#33-properties-for-pdf

File: src/docfx/SubCommands/PdfCommand.cs [hunk 2]
```diff
@@ -184,10 +184,10 @@ private static void MergeOptionsToConfig(PdfCommandOptions options, PdfJsonConfi
                 config.GeneratesExternalLink = options.GeneratesExternalLink.Value;
             }
 
-			if (options.NoInputStreamArgs.HasValue)
-			{
-				config.NoInputStreamArgs = options.NoInputStreamArgs.Value;
-			}
+            if (options.NoInputStreamArgs.HasValue)
+            {
+	            config.NoInputStreamArgs = options.NoInputStreamArgs.Value;
```
superyyrrzz: This should be the last tab 😄

PR #4722, opened 2019-06-12
append version and is_moniker_range info in filemap.json

928PJY:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/96584

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4722)

### DIFF
--- a/src/docfx/build/legacy/LegacyFileMap.cs
+++ b/src/docfx/build/legacy/LegacyFileMap.cs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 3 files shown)


src/docfx/build/legacy/LegacyFileMap.cs
```diff
@@ -33,7 +42,8 @@ internal static class LegacyFileMap
                         var legacyOutputFilePathRelativeToSiteBasePath = document.ToLegacyOutputPathRelativeToSiteBasePath(docset, publishItem);
                         var legacySiteUrlRelativeToSiteBasePath = document.ToLegacySiteUrlRelativeToSiteBasePath(docset);
 
-                        var fileItem = LegacyFileMapItem.Instance(legacyOutputFilePathRelativeToSiteBasePath, legacySiteUrlRelativeToSiteBasePath, document.ContentType);
+                        var monikerRange = GetMonikerRangeConfig(document);
```
OsmondJiang: can we remove this? i am assuming that we can also write moniker range in file metadata

OsmondJiang: i mean markdown yaml header

928PJY: The monikerRange used here is just a idenetity of group, it won't be used to be  evaluated to monikers.

---
In reply to: [293207261](https://github.com/dotnet/docfx/pull/4722#discussion_r293207261) [](ancestors = 293207261)

PR #4725, opened 2019-06-12
🧹 Remove GitHub user cache remote update

yufeih:
Since we put all GitHub user cache inside the cache folder, and you can only get GitHub with a PAT token, the explicit GitHub user location is not needed. Most of the changes are updating test cases that uses `github_user_cache` to test restore.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4725)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -58,8 +58,6 @@ namespace Microsoft.Docs.Build
 
                 var (publishManifest, fileManifests, sourceDependencies) = await BuildFiles(context, docset, tocMap);
 
-                var saveGitHubUserCache = context.GitHubUserCache.SaveChanges(config);
-
                 xrefMap.OutputXrefMap(context);
                 context.Output.WriteJson(publishManifest, ".publish.json");
                 context.Output.WriteJson(sourceDependencies.ToDependencyMapModel(), ".dependencymap.json");
@@ -77,9 +75,8 @@ namespace Microsoft.Docs.Build
                     }
                 }
 
-                context.ErrorLog.Write(await saveGitHubUserCache);
-
-                context.ContributionProvider.UpdateCommitBuildTime();
+                context.GitHubUserCache.Save();
+                context.ContributionProvider.Save();
             }
         }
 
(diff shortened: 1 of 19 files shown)


src/docfx/build/Build.cs
```diff
@@ -58,8 +58,6 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Erro
 
                 var (publishManifest, fileManifests, sourceDependencies) = await BuildFiles(context, docset, tocMap);
 
-                var saveGitHubUserCache = context.GitHubUserCache.SaveChanges(config);
```
OsmondJiang: then docfx doesn't have the ablitity to update remote github user cache, this part will be done by Docs.Build including the merge logic?

OsmondJiang: or my question is that in the future, the cahche will be per repo/branch?

---
In reply to: [293177576](https://github.com/dotnet/docfx/pull/4725#discussion_r293177576) [](ancestors = 293177576)

test/docfx.Test/lib/GitHubUserCacheTest.cs
```diff
@@ -196,19 +196,6 @@ private static IEnumerable<object[]> PrepareDataCore()
                 1,
                 1
             };
-            yield return new object[]
-            {
-                "Prefer existing email name",
```
OsmondJiang: >Prefer existing email name", [](start = 17, length = 28)

why this test is removed?

src/docfx/lib/github/GitHubUserCache.cs
```diff
@@ -296,7 +206,7 @@ private void UpdateUser(GitHubUser user, bool preferExistingName = false)
             _updated = true;
         }
 
-        private static void MergeUser(GitHubUser user, GitHubUser existingUser, bool preferExistingName)
```
OsmondJiang: >preferExistingName [](start = 85, length = 18)

is this change covered by test?

Issue #4703, opened 2019-06-10
null value handling doesn't work for non-strong model output


OsmondJiang:
When doing some investigation work for SDP, we found that if we remove strong model for SDP pipeline and directly output JToken to output file, the null value handling doesn't work any more.

That breaks the current behavior "docfx only preserve custom null value" behavior.

cc @yufeih  @mingwli  @herohua

OsmondJiang:
When user write "author" metadata, but it's invalid(not resolved correctly), docfx will remove "author" property from output model.

OsmondJiang:
The conclusion is:
1. for sdp pipeline, for these properties which need to be transformed, we need case by case determine which key need to be removed from output model, like `xref`, when it's can't be resolved, remove it from output.
2. for these properties which are unknown for docfx, keep them untouched, no matter they are null value/empty array.
3. for the metadata, keep the read-only metadata untouched, if the metadata presents in docfx output model, they are treated as docfx output, controlled by docfx.

OsmondJiang:
cc @yufeih  @live1206  @herohua  @zhiliangxu

PR #4726, opened 2019-06-12
👨‍🏭 keep user write metadata as much as possible

OsmondJiang:
remove the limitation of converting input metadata to strong model:
- only read what we want to read
- don't touch any thing we don't want to write back
- make less changes
- author and title belongs to our output model, controlled by docfx

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4726)


OsmondJiang:
#4703

OsmondJiang:
diff are expected

src/docfx/build/page/BuildPage.cs
```diff
@@ -94,9 +76,49 @@ internal static class BuildPage
             }
 
             return (errors, publishItem);
+
+            async Task<OutputModel> GenerateOutputMetadata()
```
yufeih: static method.

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -91,7 +88,10 @@ public MetadataProvider(Docset docset, Cache cache)
 
             errors.AddRange(_schemaValidator.Validate(result));
 
-            return (errors, metadata);
+            var (validationErrors, metadataModel) = JsonUtility.ToObject<InputMetadata>(result);
+            errors.AddRange(validationErrors);
+
+            return (errors, result, metadataModel);
```
live1206: so there is some content overlapping between `result` and `metadataModel`?

src/docfx/build/page/BuildPage.cs
```diff
@@ -18,55 +18,37 @@ internal static class BuildPage
...
+            var (errors, isPage, outputModel, (inputMetadata, metadataObject)) = await Load(context, file);
+            var (generateErrors, outputMetadata) = await GenerateOutputMetadata(context, file, tocMap, inputMetadata, outputModel);
+            errors.AddRange(generateErrors);
+
+            var mergedOutputMetadata = new JObject();
+            JsonUtility.Merge(mergedOutputMetadata, metadataObject);
+            JsonUtility.Merge(mergedOutputMetadata, JsonUtility.ToJObject(outputMetadata));
+            var outputPath = file.GetOutputPath(outputMetadata.Monikers, file.Docset.SiteBasePath, isPage);
+
+            var output = (object)null;
```
live1206:
```suggestion
            object output = null;
```

src/docfx/build/page/BuildPage.cs
```diff
@@ -18,55 +18,37 @@ internal static class BuildPage
...
+            var metadata = (JObject)null;
```
live1206:
```suggestion
            JObject metadata = null;
```

src/docfx/build/page/BuildPage.cs
OsmondJiang: hmmm...

src/docfx/build/metadata/MetadataProvider.cs
OsmondJiang: result will be only for reading

PR #4730, opened 2019-06-13
homepage only shows when topic href is set

OsmondJiang:
fix: https://apidrop.visualstudio.com/Content%20CI/_build%2Fresults?buildId=60910&_a=summary

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4730)

### DIFF
--- a/src/docfx/build/toc/parser/TableOfContentsParser.cs
+++ b/src/docfx/build/toc/parser/TableOfContentsParser.cs
@@ -137,7 +137,7 @@ namespace Microsoft.Docs.Build
                 // set resolved href back
                 tocModelItem.Href = resolvedTocHref ?? resolvedTopicHref ?? resolvedTopicItemFromTocHref?.Href;
                 tocModelItem.TocHref = resolvedTocHref;
-                tocModelItem.Homepage = resolvedTopicHref;
+                tocModelItem.Homepage = tocModelItem.TopicHref != null ? resolvedTopicHref : null;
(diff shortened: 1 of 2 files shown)


src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -137,7 +137,7 @@ private static (List<Error>, TableOfContentsModel) LoadTocModel(JToken tocToken)
                 // set resolved href back
                 tocModelItem.Href = resolvedTocHref ?? resolvedTopicHref ?? resolvedTopicItemFromTocHref?.Href;
                 tocModelItem.TocHref = resolvedTocHref;
-                tocModelItem.Homepage = resolvedTopicHref;
+                tocModelItem.Homepage = !string.IsNullOrEmpty(tocModelItem.TopicHref) ? resolvedTopicHref : null;
```
yufeih:
```suggestion
                tocModelItem.Homepage = !string.IsNullOrEmpty(resolvedTopicHref) ? resolvedTopicHref : null;
```

OsmondJiang: not that case, when resolvedTopicHref is not null, it may be resolved from `topicHref`, but also from `href`

PR #4735, opened 2019-06-13
Turn SourceInfo<T> into a struct

yufeih:
The current `SourceInfo<T>` is a class, with `var a = xxx`, it is unclear if `a` is `T` or `SourceInfo<T>` in expressions like `if (a != null)` or `a ?? b`, this leads to some bugs found in code reviews. Further more, if you turn `a` from `T` to `SourceInfo<T>`, there will be no compiler error that tell the semantic of `if (a != null)` checks has changed.

By changing `Source<T>` to a struct, 

- You simply cannot do `if (a != null)` or `a ?? b` when a is `SourceInfo<T>`. A helper `Or` method is provided to do null coalesce on struct using `a.Or(b)`
- Can avoid null checks on `Source<T>`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4735)


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -123,8 +123,8 @@ public string GetRelativeUrl(Document fileRelativeTo, Document file)
 
             if (href.Value.StartsWith("xref:") != false)
```
live1206: could you help fix this line? :)

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -111,7 +116,7 @@ TableOfContentsItem GetItem(HeadingBlock block)
                 {
                     var invalidTocSyntaxContent = tocContent.Substring(block.Span.Start, block.Span.Length);
                     errors.Add(Errors.InvalidTocSyntax(new SourceInfo<string>(invalidTocSyntaxContent, block.ToSourceInfo(file: filePath)), "multiple inlines in one heading block is not allowed"));
-                    return currentItem;
+                    return null;
```
live1206: what could go wrong if add an empty instance of `TableOfContentsItem` into `items`?

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
```diff
@@ -134,7 +139,10 @@ TableOfContentsItem GetItem(HeadingBlock block)
                     currentItem.Name = GetLiteral(linkInline);
                 }
 
-                currentItem.Name = currentItem.Name ?? GetLiteral(block.Inline);
+                if (currentItem.Name.Value is null)
```
live1206: use `Or` operator here as well?

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
OsmondJiang: >null [](start = 27, length = 4)

Better to avoid to return null here

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -141,16 +141,16 @@ private static (List<Error>, TableOfContentsModel) LoadTocModel(JToken tocToken)
...
-                tocModelItem.Name = tocModelItem.Name ?? resolvedTopicName;
+                tocModelItem.Homepage = !string.IsNullOrEmpty(tocModelItem.TopicHref) ? resolvedTopicHref : default;
+                tocModelItem.Name = tocModelItem.Name.Or(resolvedTopicName);
                 tocModelItem.Items = subChildren?.Items ?? tocModelItem.Items;
                 tocModelItem.Monikers = GetMonikers(resolvedTocHref, resolvedTopicHref, resolvedTopicItemFromTocHref, tocModelItem, document);
 
                 // validate
                 // todo: how to do required validation in strong model
-                if (tocModelItem.Name != null && string.IsNullOrEmpty(tocModelItem.Name))
+                if (string.IsNullOrEmpty(tocModelItem.Name))
```
OsmondJiang: >string.IsNullOrEmpty(tocModelItem.Name) [](start = 20, length = 39)

keep the default value check in case the input doesn't have `name` property?

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
yufeih: It adds an empty entry to the TOC node tree, and triggers [name check](https://github.com/dotnet/docfx/pull/4735/files#diff-d483b592e55ca934337e0e273921ee1cR153), results in syntax error and name missing error on the same node.

It does not have this problem before because there was a `tocModelItem.Name != null` check that happened to by-pass this case

cc @OsmondJiang

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
yufeih: 👍 There is only one subtle difference between `Or` and `??`, that is `Or` will always evaluate right hand side parameter, while `??` will *NOT* evaluate right hand side if left hand side is not null. In this case, `GetLiteral` is a relatively complex operation, so I am skipping the call intentionally to make it faster.

live1206: ok, maybe add a comment for `Or` operation method?

src/docfx/build/toc/parser/MarkdownTocMarkup.cs
live1206: I see

PR #4736, opened 2019-06-13
UID resolving should respect the order of xref list config.

live1206:
fix https://dev.azure.com/ceapex/Engineering/_workitems/edit/98868

This change does not affect the build performance.
Tested with dotnet/docs locally:
Build done in 00:02:13.0870000

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4736)

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -728,14 +728,11 @@ inputs:
     {
       "references":[{
           "uid": "System.String",
-          "name": "String.1",
-          "fullName": "System.String.1",
-          "href": "https://docs.microsoft.com/en-us/dotnet/api/system.string",
-          "nameWithType": "System.String"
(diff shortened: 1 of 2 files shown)


docs/specs/xref.yml
```diff
@@ -715,3 +715,27 @@ outputs:
...
+    xref: 
+      - 1.xrefmap.json
+      - 2.xrefmap.json
+  docs/a.md: Link to @System.String
+  1.xrefmap.json: |
+    {
+      "references":[{
+          "uid": "System.String",
+          "name": "String.1",
+          "fullName": "System.String.1",
```
yufeih: simplify the test case to only include `uid` and `href` ?

live1206: sounds good

PR #4742, opened 2019-06-14
Adjust build pipeline to ensure Markup gets the same context

yufeih:
- Turns `buildChild` to `context.Queue.Enqueue`
- Move LandingData logic inside `DependencyResolver`
- Some cleanup on build pipeline

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4742)


src/docfx/lib/collections/WorkQueue.cs
```diff
@@ -0,0 +1,85 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Concurrent;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Docs.Build
+{
+    internal class WorkQueue<T>
```
OsmondJiang: >WorkQueue [](start = 19, length = 9)

with this, do we still need Parallel.ForEach? Can we only keep one?

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -4,13 +4,12 @@
...
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal static class BuildTableOfContents
     {
-        public static (IEnumerable<Error>, PublishItem publishItem) Build(Context context, Document file)
+        public static void Build(Context context, Document file)
```
OsmondJiang: >public [](start = 8, length = 6)

How can we avoid to build children during building TOC map? We don't want queue every referenced files when build toc map.

yufeih: Unlike `buildChild`, `WorkQueue.Enqueue` does not trigger build immediately, it just adds to a backing queue, the queue is only build when `Drain` is called.

src/docfx/lib/collections/WorkQueue.cs
yufeih: nop, removed

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -10,19 +10,32 @@ namespace Microsoft.Docs.Build
...
+            _errorLog = errorLog;
+        }
+
         public void MarkError(Document file)
         {
             _filesWithErrors.Add(file);
         }
 
-        public bool TryAdd(Document file, PublishItem item)
+        public bool TryAdd(Document file, IEnumerable<Error> errors, PublishItem item)
```
OsmondJiang: >IEnumerable<Error> errors [](start = 42, length = 25)

could you split this change into another PR then in this PR we can focus on the change of `BuildChildren`?

src/docfx/lib/collections/WorkQueue.cs
```diff
@@ -0,0 +1,108 @@
...
+                return;
+            }
+
+            _queue.Enqueue(item);
+
+            Interlocked.Increment(ref _totalCount);
+            Interlocked.Increment(ref _remainingCount);
+        }
+
+        public Task Drain(Func<T, Task> run, Action<int, int> progress = null)
```
OsmondJiang: >Func<T, Task> run [](start = 26, length = 17)

may we make the `run` as a member of WorkQueue? then EnQueue and Drain will both use it.  Currently it's a little confuse for me that the relationship between Enqueue and Drain, seems Enqueue is using Drain parameter

src/docfx/publish/PublishModelBuilder.cs
yufeih: ok

src/docfx/lib/collections/WorkQueue.cs
yufeih: `Enqueue` will not use `run`, it just add it to the queue, this is intentional because we only have enough information to call `run` when `Drain` is called.

OsmondJiang: What i mean is that Enqueue is using Drain parameter which is confusing me, but since it's limited by the existing workflow, i am ok with that.

---
In reply to: [294582612](https://github.com/dotnet/docfx/pull/4742#discussion_r294582612) [](ancestors = 294582612)

src/docfx/build/context/Context.cs
```diff
@@ -13,33 +13,33 @@ internal sealed class Context : IDisposable
         public readonly ErrorLog ErrorLog;
         public readonly Cache Cache;
         public readonly Output Output;
+        public readonly WorkQueue<Document> BuildQueue;
```
OsmondJiang: >BuildQueue [](start = 44, length = 10)

remove it and only keep it inside of DependencyResolver?

src/docfx/build/context/Context.cs
```diff
@@ -13,33 +13,33 @@ internal sealed class Context : IDisposable
...
         public readonly DependencyMapBuilder DependencyMapBuilder;
         public readonly DependencyResolver DependencyResolver;
-        public readonly DependencyResolver LandingPageDependencyResolver;
         public readonly GitHubUserCache GitHubUserCache;
         public readonly ContributionProvider ContributionProvider;
         public readonly PublishModelBuilder PublishModelBuilder;
         public readonly TemplateEngine Template;
 
-        public Context(string outputPath, ErrorLog errorLog, Docset docset, Func<XrefMap> xrefMap)
+        public Context(string outputPath, ErrorLog errorLog,  Docset docset, Func<XrefMap> xrefMap)
```
mingwli: >g,  Docset [](start = 58, length = 10)

an extra space

PR #4744, opened 2019-06-14
register mono msbuild path

### LINKED ISSUE #4728: Does not work under Linux with Mono since 2.43
After support MSBuild16/VS2019 (#4437), DocFX cannot run under Linux/Mono. From the [error message](https://github.com/dotnet/docfx/pull/4708#issuecomment-501339741), I suspect this is because Mono is still using MSBuild15.

Once Mono supports MSBuild16, Maybe we need to add back msbuild DLLs removed in #4708.

The current workaround is to keep using 2.42.4.

tibel:
Register mono msbuild path to MSBuildLocator to enable assembly resolving.

This should fix #4728

@superyyrrzz May you run your tests with this change again please.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4744)


superyyrrzz:
Thank you for the quick fix! I will try this on my office machine next Monday.

PR #4753, opened 2019-06-18
remove schema from document and delete some strong models

OsmondJiang:
1. remove `schema` class
2. remove some strong model output class
3. use `mime` to control the build workflow
4. move `schema` logic into `template`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4753)


src/docfx/template/TemplateEngine.cs
```diff
@@ -43,16 +46,75 @@ private TemplateEngine(string templateDir, JsonSchema metadataSchema)
...
+            // todo: get `isData` from template JINT script name
+            if (mime != null && s_schemas.TryGetValue(mime, out var schema))
+            {
+                return string.Equals(schema, "ContextObject", StringComparison.OrdinalIgnoreCase) || string.Equals(schema, "TestData", StringComparison.OrdinalIgnoreCase);
+            }
+
+            return false;
+        }
+
+        public static bool Is(string mime, Type type)
```
yufeih: consider `IsLandingData(string mime)` since that is the only one we need to check.

src/docfx/template/TemplateEngine.cs
```diff
@@ -43,16 +46,75 @@ private TemplateEngine(string templateDir, JsonSchema metadataSchema)
...
+                    {
+                        mime = new SourceInfo<string>(YamlUtility.ReadMime(reader), new SourceInfo(pathToDocset, 1, 1));
+                    }
+                }
+            }
+
+            return mime;
+        }
+
+        public static string GetSchemaName(string mime)
```
yufeih: `schemaName` is the same as `mime`, so this method might be unnecessary. We can consolidate the term used internally to always be `schemaName`, could also remove OutputModel.SchemaType

src/docfx/template/TemplateEngine.cs
```diff
@@ -43,16 +46,75 @@ private TemplateEngine(string templateDir, JsonSchema metadataSchema)
                 .ToDictionary(prop => prop.Key, prop => prop.Value.HtmlMetaName);
         }
 
-        public static (JsonSchemaValidator, JsonSchemaTransformer) GetJsonSchema(Schema schema)
+        public static SourceInfo<string> ReadMimeFromFile(string pathToDocset, string filePath)
```
yufeih: put this inside `Document`?

src/docfx/docset/Document.cs
```diff
@@ -325,12 +318,12 @@ internal static ContentType GetContentType(string path)
...
         }
 
-        internal static string FilePathToSitePath(string path, ContentType contentType, Schema schema, bool json, bool uglifyUrl)
+        internal static string FilePathToSitePath(string path, ContentType contentType, string mime, bool json, bool uglifyUrl)
         {
             switch (contentType)
             {
                 case ContentType.Page:
-                    if (schema is null || schema.Attribute is PageSchemaAttribute)
+                    if (mime is null || !TemplateEngine.IsData(mime))
```
live1206: how about having another method `TemplateEngine.IsPage(mime)` which is `!IsData(mime)`. Then the reader would not wonder it's not data, then it's page, implicitly assuming that either `Data` or `Page`.

PR #4754, opened 2019-06-18
flatten toc (folder) inclusion dependency map

OsmondJiang:
fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/96108

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4754)


docs/specs/toc.yml
```diff
@@ -486,6 +486,41 @@ outputs:
...
+                {
+                    "source": "docs/f1/b.md",
+                    "type": "link"
+                },
+                {
+                    "source": "docs/f1/TOC.md",
+                    "type": "tocInclusion"
+                },
+                {
+                    "source": "docs/f2/a.md",
```
superyyrrzz: Not all the dependency in referenced TOC is transit. Actually `toc.md` does not depend on `f1/a.md`.

OsmondJiang: good point, seems we need know it's referencing a folder or referencing a toc file

OsmondJiang:
The toc dependencies changes are expected.

image

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -419,6 +426,7 @@ private static TocHrefType GetHrefType(string href)
 
         private enum TocHrefType
         {
+            None,
```
superyyrrzz: no use?

src/docfx/build/dependency/DependencyType.cs
```diff
@@ -10,6 +10,7 @@ public enum DependencyType
         UidInclusion, // uid reference with display property
         Inclusion, // token or codesnippet
         Overwrite, // overwrite markdown reference
-        TocInclusion, // toc reference toc
+        TocInclusion, // toc includes toc
+        TocFolderInclusion, // toc includes toc first item
```
superyyrrzz: naming: what about `TocFolderReference`? Just to make it clear that it is not an inclusion.

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -244,36 +280,35 @@ SourceInfo<string> GetTopicHref(TableOfContentsItem tocInputModel)
                 return default;
             }
 
-            (SourceInfo<string> resolvedTocHref, TableOfContentsItem resolvedTopicItem, TableOfContentsModel subChildren) ProcessTocHref(SourceInfo<string> tocHref)
+            (SourceInfo<string> resolvedTocHref, TableOfContentsModel subChildren, (TableOfContentsItem item, Document doc) firstChild, TocHrefType tocHrefType) ProcessTocHref(SourceInfo<string> tocHref)
```
superyyrrzz: >(TableOfContentsItem item, Document doc)  [](start = 83, length = 41)

From the caller, seems it is not necessary to make return type a nested tuple.

src/docfx/build/toc/parser/TableOfContentsParser.cs
OsmondJiang: used, return `default`

src/docfx/build/dependency/DependencyType.cs
OsmondJiang: it's treated as inclusion in v2

src/docfx/build/toc/parser/TableOfContentsParser.cs
OsmondJiang: it's necessary, firstItem is used separated

src/docfx/build/dependency/DependencyType.cs
OsmondJiang: sorry, toc and toc relationships are missing in v2, so v3 generate it according to our own understanding.

OsmondJiang:
@superyyrrzz v2 doesn't flatten toc inclusion dependencies, so this PR is focusing on flatten folder inclusion dependencies

OsmondJiang:
the diff are expected:

image

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -151,7 +169,25 @@ private static (List<Error>, TableOfContentsModel) LoadTocModel(JToken tocToken)
...
+
+            return (errors, firstChild);
+
+            void ProcessSubChildren((TableOfContentsItem item, Document doc) subChildFromItems, (TableOfContentsItem item, Document doc) subChildFromTocHref)
+            {
+                if (subChildFromTocHref != default)
+                {
+                    subFirstChildren.Add(subChildFromTocHref);
+                }
+                else
```
live1206: else if (...)

PR #4757, opened 2019-06-18
trim trailing 'index' in rediret to when with id

mingwli:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/96087/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4757)

### DIFF
--- a/src/docfx/build/redirection/RedirectionMap.cs
+++ b/src/docfx/build/redirection/RedirectionMap.cs
@@ -80,6 +80,7 @@ namespace Microsoft.Docs.Build
                     }
 
                     var combineRedirectUrl = false;
+                    var mutableRedirectUrl = redirectUrl.Value;
                     if (redirectDocumentId)
                     {
                         switch (UrlUtility.GetLinkType(redirectUrl))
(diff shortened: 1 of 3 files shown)


src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -93,9 +94,14 @@ void AddRedirections(Dictionary<string, SourceInfo<string>> items, bool redirect
                                 errors.Add(Errors.RedirectionUrlInvalid(redirectUrl));
                                 continue;
                         }
+
+                        if (redirectUrl.Value.EndsWith("index"))
```
yufeih: this check also matches `reindex`, do it in `Document` after absolute URL is combined, then check against `/index`

src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -93,9 +94,14 @@ void AddRedirections(Dictionary<string, SourceInfo<string>> items, bool redirect
...
+                        {
+                            mutableRedirectUrl = redirectUrl.Value.Substring(0, redirectUrl.Value.LastIndexOf("index"));
```
yufeih: avoid scan in `LastIndexOf`:
```suggestion
                            mutableRedirectUrl = redirectUrl.Value.Substring(0, redirectUrl.Value.Length - "index".Length);
```

src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -95,7 +96,7 @@ void AddRedirections(Dictionary<string, SourceInfo<string>> items, bool redirect
                         }
                     }
 
-                    Document redirect = Document.Create(docset, pathToDocset, redirectUrl, combineRedirectUrl: combineRedirectUrl);
+                    Document redirect = Document.Create(docset, pathToDocset, mutableRedirectUrl, combineRedirectUrl: combineRedirectUrl);
```
yufeih: revert this change

PR #4760, opened 2019-06-19
Add JsonSchema const validation

yufeih:
#4605 

Implement the `const` standard JSON schema validation.

@mingwli @LowEntropyBody 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4760)


LowEntropyBody:
InlineData("{'const': {}}", "{‘a’ : 2}") What is the behavior of this situation? The current logic seems to be no error.

src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -301,5 +310,47 @@ private static bool TypeMatches(JsonSchemaType schemaType, JTokenType tokenType)
...
+
+                case JObject mapA when b is JObject mapB:
+                    if (mapA.Count != mapB.Count)
+                    {
+                        return false;
+                    }
+
+                    foreach (var (key, valueA) in mapA)
+                    {
+                        if (!mapB.TryGetValue(key, out var valueB) || !JTokenDeepEquals(valueA, valueB))
```
LowEntropyBody: InlineData("{'const': {}}", "{‘a’ : 2}") What is the behavior of this situation? The current logic seems to be no error.

yufeih: It will fail, there is a check for count in line 337, the JSON schema test suite has this case.

PR #4763, opened 2019-06-20
Fix for parsing query string when typo happens.

live1206:
Fix https://dev.azure.com/ceapex/Engineering/_workitems/edit/99306/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4763)

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -153,7 +153,14 @@ inputs:
     Link to <xref:a?displayProperty=notExisting>
 outputs:
   docs/c.json: | 
-    {"conceptual":"<p>Link to <a href=\"a.json\">test.test</a>\nLink to <a href=\"a.json\">test</a>\nLink to <a href=\"a.json\">test.test</a>\nLink to <a href=\"a.json\">test</a>??displayProperty=fullName\nLink to <a href=\"a.json\">test</a></p>\n"}
+    {
+      "conceptual":"<p>
+                        Link to <a href=\"a.json\">test.test</a>
+                        Link to <a href=\"a.json\">test</a>
+                        Link to <a href=\"a.json\">test.test</a>
+                        Link to <a href=\"a.json\">test</a>??displayProperty=fullName
+                        Link to <a href=\"a.json\">test</a></p>"
+    }
   docs/a.json:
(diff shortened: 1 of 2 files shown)


docs/specs/xref.yml
```diff
@@ -147,11 +147,13 @@ inputs:
...
   docs/c.md: |
     Link to <xref:a?displayProperty=fullName>
+    Link to <xref:a??displayProperty=fullName>
     Link to @a?displayProperty=fullName
+    Link to @a??displayProperty=fullName
     Link to <xref:a?displayProperty=notExisting>
 outputs:
   docs/c.json: | 
-    {"conceptual":"<p>Link to <a href=\"a.json\">test.test</a>\nLink to <a href=\"a.json\">test.test</a>\nLink to <a href=\"a.json\">test</a></p>\n"}
+    {"conceptual":"<p>Link to <a href=\"a.json\">test.test</a>\nLink to <a href=\"a.json\">test</a>\nLink to <a href=\"a.json\">test.test</a>\nLink to <a href=\"a.json\">test</a>??displayProperty=fullName\nLink to <a href=\"a.json\">test</a></p>\n"}
```
herohua: why would we want behaviors are different between `xref` and `@`

live1206: Because @uid??displayProperty, the href in xref block is `uid`
and <xref: uid??displayProperty>, the href in xref block is `uid??displayProperty`

live1206: https://github.com/dotnet/docfx/blob/cc1757a8e01658b2de49a8dfa135ec7ccd7c7ea4/src/docfx/lib/markdown/ResolveXrefExtension.cs#L29
I am talking about this one

superyyrrzz: This shares the same behavior with v2.

yufeih: I see, this is because shorthand syntax match does not allow consecutive symbols in `XrefInlineShortParser`

yufeih: replace `\n` with real newline for readability

PR #4767, opened 2019-06-20
change json schema transform from `replace` to `return`

OsmondJiang:
avoid replacing existing model, which is not thread safe.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4767)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -87,7 +87,6 @@ namespace Microsoft.Docs.Build
                 }
 
                 var transformedScalar = TransformScalar(subSchema, file, context, node as JValue, errors);
-                JsonUtility.SetSourceInfo(transformedScalar, JsonUtility.GetSourceInfo(node));
                 return (default, transformedScalar);
             });
         }
@@ -95,7 +94,6 @@ namespace Microsoft.Docs.Build
         private JToken Traverse(JsonSchema schema, JToken token, Func<JsonSchema, JToken, (string[], JToken)> transform)
         {
             schema = _definitions.GetDefinition(schema);
-            var sourceInfo = JsonUtility.GetSourceInfo(token);
             if (schema == null)
             {
                 return token;
@@ -116,7 +114,6 @@ namespace Microsoft.Docs.Build
                         return array;
 
                     var newArray = new JArray();
-                    JsonUtility.SetSourceInfo(newArray, sourceInfo);
                     foreach (var item in array)
                     {
                         var newItem = Traverse(schema.Items, item, transform);
@@ -128,7 +125,6 @@ namespace Microsoft.Docs.Build
                 case JObject obj:
 
                     var newObject = new JObject();
-                    JsonUtility.SetSourceInfo(newObject, sourceInfo);
                     foreach (var (key, value) in obj)
                     {
                         if (!transformedKeys.Contains(key) && TryGetPropertyJsonSchema(schema, key, out var propertySchema))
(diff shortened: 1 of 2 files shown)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -63,70 +63,91 @@ public JsonSchemaTransformer(JsonSchema schema)
...
+                JsonUtility.SetSourceInfo(transformedScalar, JsonUtility.GetSourceInfo(node));
+                return (default, transformedScalar);
             });
         }
 
-        private void Traverse(JsonSchema schema, JToken token, Func<JsonSchema, JToken, string[]> transform)
+        private JToken Traverse(JsonSchema schema, JToken token, Func<JsonSchema, JToken, (string[], JToken)> transform)
         {
             schema = _definitions.GetDefinition(schema);
+            var sourceInfo = JsonUtility.GetSourceInfo(token);
```
yufeih: To make things simpler, the transformed token does not need to preserve source info, there is no downstream consumers that needs source info afterwards. Same thing can be removed in  line 183.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -63,70 +63,91 @@ public JsonSchemaTransformer(JsonSchema schema)
...
             if (schema == null)
             {
-                return;
+                return token;
             }
 
-            var transformedKeys = transform(schema, token) ?? Array.Empty<string>();
+            var (transformedKeys, transformedToken) = transform(schema, token);
+            transformedKeys = transformedKeys ?? Array.Empty<string>();
```
yufeih: can compare `token` and `transformedToken` for reference equality to detect if the token has been processed. If the token is processed, skip the entire switch block in line 107.
With this, the callback does not need to return `transformedKeys`

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -63,70 +63,91 @@ public JsonSchemaTransformer(JsonSchema schema)
...
 
-            switch (token)
+            switch (transformedToken)
             {
+                case JValue scalar:
+
+                    return scalar;
+
                 case JArray array:
```
yufeih: There is no need to copy the array if `schema.Items == null`:
```suggestion
                case JArray array when schema.Items != null:
```

With this, can replace line 109 ~ line 112 and line 150 with a default block that return `token`

src/docfx/lib/schema/JsonSchemaTransformer.cs
OsmondJiang: I am not sure about that, will xref need that?

src/docfx/lib/schema/JsonSchemaTransformer.cs
OsmondJiang: good point!

src/docfx/lib/schema/JsonSchemaTransformer.cs
OsmondJiang: when the token is process, if the token is JObject, maybe only a few properties are process, so reference equality is not working here

src/docfx/lib/schema/JsonSchemaTransformer.cs
yufeih: nop, after schema transform, it is turned into a json object and pass to template

OsmondJiang: @live1206  confirmed it's not used in xref, so delete it

PR #4768, opened 2019-06-20
🧹no need href info anymore since it's calculated in parser

OsmondJiang:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4768)


src/docfx/build/context/Cache.cs
```diff
@@ -30,10 +30,10 @@ internal class Cache
                 return JsonUtility.Parse(content, file.FilePath);
             })).Value;
 
-        public (List<Error>, TableOfContentsModel, List<(Document doc, string herf)>, List<Document>) LoadTocModel(Context context, Document file)
+        public (List<Error>, TableOfContentsModel, List<Document>, List<Document>) LoadTocModel(Context context, Document file)
```
live1206: what are the returns here? especially two `List<Document>`

PR #4773, opened 2019-06-21
split output model and output metadata, consistent conceptual and sdp build pipeline

OsmondJiang:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4773)


src/docfx/build/page/BuildPage.cs
```diff
@@ -18,25 +19,26 @@ internal static class BuildPage
...
-            JsonUtility.Merge(mergedOutputMetadata, JsonUtility.ToJObject(outputMetadata));
             var outputPath = file.GetOutputPath(outputMetadata.Monikers, file.Docset.SiteBasePath, isPage);
 
             object output = null;
             JObject metadata = null;
             if (isPage)
             {
-                (output, metadata) = ApplyPageTemplate(context, file, mergedOutputMetadata, outputMetadata.Conceptual);
+                var mergedOutputModel = new JObject();
+                JsonUtility.Merge(mergedOutputModel, metadataObject);
```
live1206: How about adding a new method for this usage:
```csharp
        public static void Merge(JObject container, params JObject[] overwrites)
        {
            foreach (var item in overwrites)
            {
                Merge(container, item);
            }
        }
```

OsmondJiang: we can add it later once we need apply data page template and need merge there.

OsmondJiang:
the change is expected, now the data page doesn't have any system generated metadata anymore, output only contains transformed jtoken.

PR #4775, opened 2019-06-21
Extract BuildScope outside Docset

yufeih:
To eventually make `Document` a pure data model. cc @mingwli 

- Extract BuildScope and put it to context
- Drop the concept of scan scope, loc build scope contains fallback, but fallback contents are skipped when building.
- Move TOC build scope related logic to `BuildTableOfContents`
- Remove publishItem to BuildXXX return value since we are not using it
- Move resolve alias outside of Docset

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4775)


src/docfx/build/Build.cs
```diff
@@ -102,38 +102,26 @@ void ValidateBookmarks()
...
                 {
-                    case ContentType.Resource:
-                        (errors, publishItem) = BuildResource.Build(context, file);
+                    case ContentType.Resource when !file.Docset.IsFallback():
+                        errors = BuildResource.Build(context, file);
                         break;
-                    case ContentType.Page:
-                        (errors, publishItem) = await BuildPage.Build(context, file, tocMap);
+                    case ContentType.Page when !file.Docset.IsFallback():
+                        errors = await BuildPage.Build(context, file, tocMap);
```
OsmondJiang: can we still keep the controlling of which file to build here?

src/docfx/build/Build.cs
```diff
@@ -53,13 +53,13 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Erro
             using (var context = new Context(outputPath, errorLog, docset, () => xrefMap))
             {
                 xrefMap = XrefMapBuilder.Build(context, docset);
-                var tocMap = TableOfContentsMap.Create(context, docset);
+                var tocMap = TableOfContentsMap.Create(context);
 
-                context.BuildQueue.Enqueue(GetBuildScope(docset, tocMap));
+                context.BuildQueue.Enqueue(context.BuildScope.Files);
```
OsmondJiang: we are moving more and more things into context, what's the different to keep them inside of docset?

yufeih: I'm thinking eventually there will be no `Docset` class 😁, `Document` becomes a pure data model, and everything else goes to their respective behavioral classes (e.g., resolveAlias goes to `DependencyResolver`)

src/docfx/build/Build.cs
```diff
@@ -102,38 +102,26 @@ void ValidateBookmarks()
...
                         // TODO: improve error message for toc monikers overlap
-                        (errors, publishItem) = BuildTableOfContents.Build(context, file);
+                        errors = BuildTableOfContents.Build(context, file, tocMap);
                         break;
-                    case ContentType.Redirection:
-                        (errors, publishItem) = BuildRedirection.Build(context, file);
+                    case ContentType.Redirection when !file.Docset.IsFallback():
+                        errors = BuildRedirection.Build(context, file);
                         break;
                 }
```
OsmondJiang: The telemetry is not correct here: @yufeih 
```
Telemetry.TrackBuildItemCount(file.ContentType);
```

PR #4777, opened 2019-06-21
Append branch information to xref map output

live1206:
Append ?branch=master to xref map output URL when the current build repo branch is master

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4777)


superyyrrzz:
OPS has some logic to patch `xrefmap.yml`. We can move there.

PR #4780, opened 2019-06-21
ResolveLink --> ResolveAbsoluteLink

yufeih:
- DependencyResolver --> LinkResolver
- ResolveLink --> ResolveAbsoluteLink
- ResolveXref --> ResolveAbsoluteXref

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4780)

### DIFF
--- a/src/docfx/build/context/LinkResolver.cs
+++ b/src/docfx/build/context/LinkResolver.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Docs.Build
 
         public (Error error, string href, string display, IXrefSpec spec) ResolveRelativeXref(Document relativeToFile, SourceInfo<string> href, Document declaringFile)
         {
-            var (error, link, display, spec) = ResolveXref(href, declaringFile);
+            var (error, link, display, spec) = ResolveAbsoluteXref(href, declaringFile);
 
             if (spec?.DeclairingFile != null)
             {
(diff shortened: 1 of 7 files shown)


src/docfx/build/context/LinkResolver.cs
```diff
@@ -54,9 +54,9 @@ internal class DependencyResolver
             return (error, link, file);
         }
 
-        public (Error error, string link, Document file) ResolveLink(SourceInfo<string> path, Document declaringFile)
+        public (Error error, string link, Document file) ResolveAbsoluteLink(SourceInfo<string> path, Document declaringFile)
```
live1206: Does `ResolveAbsoluteLink` mean the resolved link would be absolute? same for `ResovleAbsoluteXref`

yufeih: yep, to pair with `ResolveRelativeLink`

src/docfx/build/context/LinkResolver.cs
```diff
@@ -11,15 +11,15 @@
 
 namespace Microsoft.Docs.Build
 {
-    internal class DependencyResolver
+    internal class LinkResolver
```
live1206: This also contains `ResolveContent`, which is inclusion. Is `LinkResolver` accurate?

yufeih: The term `Link` is more about input rather than output, `ResolveLink` uses an input link to resolve an output link, in `ResolveContent` uses an input link to resolve the output content. `ResolveXref` uses an input _uid link_ to resolve and output link.

OsmondJiang:
What's the purpose to change `DependencyResolver` to `LinkResolver`? We have `ResolveContent`, `ResolveLink` and `ResolveXref` in `DependencyResolver` and also manage the dependency relationships, so I'd like to keep the original name

PR #4787, opened 2019-06-24
change output model to _tocRel

OsmondJiang:
fix bug: https://dev.azure.com/ceapex/Engineering/_workitems/edit/94425

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4787)

### DIFF
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -62,7 +62,6 @@ outputs:
         <meta name=\"search.ms_product\" content=\"\" />
         <meta name=\"search.ms_sitename\" content=\"Docs\" />
         <meta name=\"site_name\" content=\"Docs\" />
-        <meta name=\"toc_rel\" content=\"toc.json\" />
         <meta name=\"updated_at\" content=\"2018-10-30 12:00 AM\" />
         <meta name=\"wordCount\" content=\"1\" />"
     }
@@ -145,7 +144,7 @@ outputs:
         },
         "conceptual": "\n<p><a href=\"d.png\">d</a></p>\n",
         "_path": "c.json",
-        "toc_rel": "toc.json",
+        "_tocRel": "toc.json",
         "wordCount": 2,
         "title": "Title",
         "rawTitle": "<h1 id=\"title\">Title</h1>",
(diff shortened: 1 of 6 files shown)


OsmondJiang:
diff are expected

src/docfx/build/page/BuildPage.cs
```diff
@@ -93,7 +93,7 @@ public static async Task<IEnumerable<Error>> Build(Context context, Document fil
             }
 
             outputMetadata.Locale = file.Docset.Locale;
-            outputMetadata.TocRel = tocMap.FindTocRelativePath(file);
+            outputMetadata.TocRel = !string.IsNullOrEmpty(inputMetadata.TocRel) ? inputMetadata.TocRel : tocMap.FindTocRelativePath(file);
```
yufeih: So for this type of archived content, it does not need TOC map. So TOC map build can be lazily evaluated.

OsmondJiang:
> Instead of changing docfx, is it better to change `docs-ui` ?

Change in template need confirmation with template team, and also we already have '_op_contributionInfo' output name which is also ugly as this, so let's do this naming thing together later.

src/docfx/build/page/BuildPage.cs
OsmondJiang: We don't know if all files in archived content has "_tocRel" metadata, but you are right, we can lazy the toc map. I prefer to do it in another PR.

yufeih: yep, just bring it up.

docs/specs/legacy.yml
```diff
@@ -62,7 +62,6 @@ outputs:
         <meta name=\"search.ms_product\" content=\"\" />
         <meta name=\"search.ms_sitename\" content=\"Docs\" />
         <meta name=\"site_name\" content=\"Docs\" />
-        <meta name=\"toc_rel\" content=\"toc.json\" />
```
superyyrrzz: We still need `toc_rel` in output page meta. Does it still work?

OsmondJiang: That's template model output, not covered by our spec test now. It's converted by integration test, check the test result

Issue #4375, opened 2019-04-16
lowercase url config miss match with site base path


OsmondJiang:
Docfx using site base path to calculate "document id" and kinds of  "relative path", but when the "output.lowercaseUrl" is true, and the "siteBasePath" has uppercase characters, the "document id“and kinds of "relative path" will be calculated wrong in linux system.

@yufeih @live1206

PR #4788, opened 2019-06-24
Make BaseUrl respect LowerCaseUrl

yufeih:
Fixes https://github.com/dotnet/docfx/issues/4375

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4788)

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Runtime.Serialization;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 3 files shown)


src/docfx/config/Config.cs
```diff
@@ -181,5 +182,14 @@ internal sealed class Config
         /// for the latest commit that touches that document.
         /// </summary>
         public readonly bool UpdateTimeAsCommitBuildTime = false;
+
+        [OnDeserialized]
+        private void OnDeserialized(StreamingContext context)
```
OsmondJiang: can we add [ToLowercase] attribute for that? then we can also use it for other property like locale.

yufeih: attributes are very restrictive, in this case, `BaseUrl` depend on the value of `Output.LowerCaseUrl` to determine whether it needs to lower.

PR #4790, opened 2019-06-24
lazy toc map and xref map in context, control the init order

OsmondJiang:
- lazy toc map and xref map in context
- init build action in WorkQueue constructor instead of Drain
- One single place to control which file could be build

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4790)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -50,16 +50,13 @@ namespace Microsoft.Docs.Build
             var docset = GetBuildDocset(new Docset(errorLog, docsetPath, locale, config, options, restoreMap, repository, fallbackRepo));
             var outputPath = Path.Combine(docsetPath, config.Output.Path);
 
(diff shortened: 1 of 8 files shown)


src/docfx/build/Build.cs
```diff
@@ -143,6 +144,22 @@ private static async Task BuildFile(Context context, Document file, TableOfConte
                 Console.WriteLine($"Build {file.FilePath} failed");
                 throw;
             }
+
+            bool ShouldBuildFile()
+            {
+                if (file.ContentType == ContentType.TableOfContents)
```
live1206: Does a repo always have a Toc file? If yes, `TocMap` will always be loaded, seems no need to lazy load?

PR #4797, opened 2019-06-25
dont parse datetime when load swagger json

yishengjin1413:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4797)


yishengjin1413:
Yes. I think so, From the code, no one else referenced this method in Docfx now.

PR #4802, opened 2019-06-26
Remove filePath and etag from `GetRestoredFileContent` 

yufeih:
- Don't read the whole file content just to retrieve etag, because some content may be huge like xrefmap.
- Remove filePath and etag from `GetRestoredFileContent` since they are not used outside of restore
- Reworked `RestoreMap.Create` to lift `try...catch` logic outside.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4802)


src/docfx/restore/lock/DependencyLock.cs
```diff
@@ -51,12 +51,7 @@ public static DependencyLockModel Load(string docset, SourceInfo<string> depende
                 }
             }
 
-            var (_, content, _) = RestoreMap.GetRestoredFileContent(docset, dependencyLockPath);
-
-            if (string.IsNullOrEmpty(content))
```
yufeih: this shouldn't use `string.IsNullOrEmpty`

OsmondJiang: so to fix this line, changed so many files..? can we split the other changes to other PRs?

yufeih: sure

src/docfx/restore/RestoreMap.cs
```diff
@@ -88,28 +87,31 @@ public static (string localPath, string content, string etag) GetRestoredFileCon
...
+                        return File.ReadAllText(fullPath);
                     }
                 }
 
                 throw Errors.FileNotFound(url).ToException();
             }
 
-            var (content, etag) = TryGetRestoredFileContent(url);
-            if (string.IsNullOrEmpty(content))
+            var filePath = RestoreFile.GetRestoreContentPath(url);
```
OsmondJiang: put it inside ProcessMutex?

PR #4804, opened 2019-06-26
 Don't throw need restore when the content is empty

yufeih:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4804)

### DIFF
--- a/src/docfx/restore/RestoreMap.cs
+++ b/src/docfx/restore/RestoreMap.cs
@@ -104,7 +104,7 @@ namespace Microsoft.Docs.Build
             }
 
             var (content, etag) = TryGetRestoredFileContent(url);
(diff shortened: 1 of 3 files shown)


src/docfx/restore/RestoreMap.cs
```diff
@@ -104,7 +104,7 @@ public static (string localPath, string content, string etag) GetRestoredFileCon
             }
 
             var (content, etag) = TryGetRestoredFileContent(url);
-            if (string.IsNullOrEmpty(content))
+            if (content == null)
```
live1206:
```suggestion
            if (content is null)
```

PR #4807, opened 2019-06-27
Make error immutable

yufeih:
Set level to error in `ErrorLog.Write` instead of `ToException`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4807)

### DIFF
--- a/src/docfx/lib/log/Error.cs
+++ b/src/docfx/lib/log/Error.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Docs.Build
     {
         public static readonly IEqualityComparer<Error> Comparer = new EqualityComparer();
 
-        public ErrorLevel Level { get; private set; }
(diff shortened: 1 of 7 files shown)


src/docfx/lib/log/Error.cs
```diff
@@ -13,7 +13,7 @@ internal class Error
     {
         public static readonly IEqualityComparer<Error> Comparer = new EqualityComparer();
 
-        public ErrorLevel Level { get; private set; }
+        public ErrorLevel Level { get; set; }
```
mingwli: @msftbot hold this pr for another 30 minutes so that @yufeih has a chance to consider my feedback

PR #4808, opened 2019-06-27
check isdata from template dir

mingwli:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4808)


src/docfx/template/TemplateEngine.cs
```diff
@@ -36,15 +40,43 @@ private TemplateEngine(string templateDir, JsonSchema metadataSchema)
             var contentTemplateDir = Path.Combine(templateDir, "ContentTemplate");
             var schemaDir = Path.Combine(contentTemplateDir, "schemas");
 
+            Global = LoadGlobalTokens(contentTemplateDir);
             _templateDir = templateDir;
             _schemaDir = schemaDir;
             _liquid = new LiquidTemplate(templateDir);
             _js = new JavascriptEngine(contentTemplateDir);
-            Global = LoadGlobalTokens(contentTemplateDir);
             _schemas = Directory.Exists(schemaDir) ? Directory.EnumerateFiles(schemaDir, "*.schema.json", SearchOption.TopDirectoryOnly)
```
yufeih: consider extract a method here:

```suggestion
            _schemas = LoadSchemas(xxx)
```

src/docfx/template/TemplateEngine.cs
```diff
@@ -36,15 +40,43 @@ private TemplateEngine(string templateDir, JsonSchema metadataSchema)
...
+                _schemaTemplates.Add(schema, (true, false));
+            }
             _schemas.Add("LandingData");
+            _schemaTemplates.Add("LandingData", (false, false));
+
+            if (Directory.Exists(contentTemplateDir))
+            {
+                var allJsFiles = Directory.EnumerateFiles(contentTemplateDir, "*.js", SearchOption.TopDirectoryOnly)
+                             .Select(k => Path.GetFileNameWithoutExtension(k))
+                             .GroupBy(k => k.Split('.').First());
```
yufeih: Don't assume schema name cannot contain `.`

src/docfx/template/TemplateEngine.cs
```diff
@@ -191,8 +170,61 @@ public JObject PreprocessMetadata(JObject outputMetadata, Document file)
         public JObject TransformTocMetadata(object model)
             => TransformMetadata("toc.json.js", JsonUtility.ToJObject(model));
 
+        private IReadOnlyDictionary<string, (bool isData, bool transformMetadata, (JsonSchemaValidator, JsonSchemaTransformer) jsonSchema)>
+            LoadSchemas(string schemaDir, string contentTemplateDir)
+        {
+            var schemas = new Dictionary<string, (bool isData, bool transformMetadata, (JsonSchemaValidator, JsonSchemaTransformer) jsonSchema)>();
+            var schemaNames = Directory.Exists(schemaDir) ? Directory.EnumerateFiles(schemaDir, "*.schema.json", SearchOption.TopDirectoryOnly)
+                                                    .Select(k => Path.GetFileNameWithoutExtension(k))
+                                                    .Select(k => k.Substring(0, k.Length - ".schema".Length)).ToHashSet() : new HashSet<string>();
```
live1206: If we only want to remove duplicates here, why not `List.Distict` other than `HashSet`?

src/docfx/template/TemplateEngine.cs
```diff
@@ -191,8 +170,61 @@ public JObject PreprocessMetadata(JObject outputMetadata, Document file)
...
+            {
+                if (schemaName is null)
+                {
+                    return default;
+                }
+
+                var schemaFilePath = Path.Combine(schemaDir, $"{schemaName}.schema.json");
+                if (string.Equals(schemaName, "LandingData", StringComparison.OrdinalIgnoreCase))
+                {
+                    schemaFilePath = Path.Combine(AppContext.BaseDirectory, "data", "schemas", "LandingData.json");
```
live1206: why is `LandingData` special here not like the others?

mingwli: https://github.com/search?utf8=%E2%9C%93&q=org%3Amicrosoftdocs+landingdata+in%3Afile+extension%3Ayml&ref=simplesearch

LandingData used to belong to Yaml-Pipeline(which I'm not really familiar with), and it doesn't have a corresponding `LandingData.schema.json` defined in template repo

src/docfx/template/TemplateEngine.cs
mingwli: We only need to enumerate the collection with no further operation on the collection itself, and the order doesn't matter. So I think both HashSet and List are acceptable here.

live1206: Since we already use Linq here, I think it would be more readable to use `Distinct` in the end instead of doing it implicitly by introducing `HashSet`

src/docfx/template/TemplateEngine.cs
```diff
@@ -18,16 +17,13 @@ internal class TemplateEngine
...
-                          = new ConcurrentDictionary<string, Lazy<(JsonSchemaValidator, JsonSchemaTransformer)>>();
 
         private readonly string _templateDir;
-        private readonly string _schemaDir;
         private readonly LiquidTemplate _liquid;
         private readonly JavascriptEngine _js;
         private readonly HashSet<string> _htmlMetaHidden;
         private readonly Dictionary<string, string> _htmlMetaNames;
-        private readonly HashSet<string> _schemas;
+        private readonly IReadOnlyDictionary<string, TemplateSchema> _schemas;
```
yufeih: Move Lazies inside `TemplateSchema` over here:
```suggestion
        private readonly IReadOnlyDictionary<string, Lazy<TemplateSchema>> _schemas;
```

src/docfx/template/TemplateEngine.cs
```diff
@@ -191,8 +167,22 @@ public JObject PreprocessMetadata(JObject outputMetadata, Document file)
...
 
+        private IReadOnlyDictionary<string, TemplateSchema>
+            LoadSchemas(string schemaDir, string contentTemplateDir)
+        {
+            var schemas = new Dictionary<string, TemplateSchema>();
+            var schemaNames = Directory.Exists(schemaDir) ? (from k in Directory.EnumerateFiles(schemaDir, "*.schema.json", SearchOption.TopDirectoryOnly)
+                                                             let fileName = Path.GetFileName(k)
+                                                             select fileName.Substring(0, fileName.Length - ".schema.json".Length)).ToList() : new List<string>();
+
+            schemaNames.ForEach(schemaName => schemas.Add(schemaName, new TemplateSchema(schemaName, schemaDir, contentTemplateDir)));
```
yufeih: can just use `ToDictionary`

src/docfx/template/TemplateEngine.cs
```diff
@@ -191,8 +167,22 @@ public JObject PreprocessMetadata(JObject outputMetadata, Document file)
         public JObject TransformTocMetadata(object model)
             => TransformMetadata("toc.json.js", JsonUtility.ToJObject(model));
 
+        private IReadOnlyDictionary<string, TemplateSchema>
```
yufeih:
```suggestion
        private static IReadOnlyDictionary<string, TemplateSchema>
```

mingwli: what's the benefit of this?

yufeih: To ensure this method is pure and does not access member variables

PR #4809, opened 2019-06-28
Remove TemplateEngine.PreprocessMetadata

yufeih:
- Move global to `JavascriptEngine` so we don't convert it for each jint run (__global is huge)
- Remove `TemplateEngine.PreprocessMetadata`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4809)

### DIFF
--- a/test/docfx.Test/template/JavascriptEngineTest.cs
+++ b/test/docfx.Test/template/JavascriptEngineTest.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Docs.Build
         {
             var inputJson = JObject.Parse(input.Replace('\'', '"'));
             var outputJson = _js.Run("index.js", "main", inputJson);
-            (outputJson as JObject).Property("__global")?.Remove();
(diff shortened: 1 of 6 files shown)


yufeih:
The only diff is ordering

test/docfx.Test/template/JavascriptEngineTest.cs
```diff
@@ -19,6 +19,7 @@ public void RunJavascript(string input, string output)
         {
             var inputJson = JObject.Parse(input.Replace('\'', '"'));
             var outputJson = _js.Run("index.js", "main", inputJson);
+            (outputJson as JObject).Property("__global")?.Remove();
```
live1206: we do not need `__global` in the output model anymore?

yufeih: yep, this is a temp variable used inside JINT

PR #4812, opened 2019-06-28
Only produce XrefNotFound error from XrefMap

yufeih:
Currently XrefNotFound error is produced from both XrefMap and xref resolve, this PR aligns xref resolve and link resolve, so errors are handled and reported consistently.

With this, it is easier to consolidate link resolve and xref resolve inside HTML and produce the correct line info (https://dev.azure.com/ceapex/Engineering/_workitems/edit/78751)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4812)


src/docfx/lib/markdown/MarkdownUtility.cs
```diff
@@ -185,7 +153,40 @@ private static (string content, object file) ReadFile(string path, object relati
             return (content, file);
         }
 
-        private static List<string> ParseMonikerRange(SourceInfo<string> monikerRange)
+        private static string GetLink(SourceInfo<string> href)
```
OsmondJiang: still keep it inside of xxxExtension? why move it out?

yufeih: To make it so only `MarkdownUtility` references `MarkdigExtensions`, currently they references each other

PR #4820, opened 2019-07-02
consist conceptual and landing data build

OsmondJiang:
1. define conceptual model, used by building conceptual and landing data
2. null/empty array value's key should be removed from output

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4820)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -5,7 +5,9 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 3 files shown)


src/docfx/build/page/BuildPage.cs
```diff
@@ -234,15 +234,14 @@ private static (List<Error> errors, JObject model, InputMetadata inputMetadata)
...
+                    Conceptual = HtmlUtility.LoadHtml(await RazorTemplate.Render(file.Mime, content)).HtmlPostProcess(file.Docset.Culture),
+                    Title = inputMetadata.Title ?? obj?.Value<string>("title"),
+                    RawTitle = $"<h1>{obj?.Value<string>("title")}</h1>",
+                });
             }
 
-            pageModel["title"] = inputMetadata.Title ?? obj?.Value<string>("title");
-            pageModel["rawTitle"] = file.Docset.Legacy ? $"<h1>{obj?.Value<string>("title")}</h1>" : null;
-
             return (errors, pageModel as JObject, inputMetadata);
```
live1206:
```suggestion
            return (errors, pageModel, inputMetadata);
```

src/docfx/build/page/BuildPage.cs
```diff
@@ -234,15 +234,14 @@ private static (List<Error> errors, JObject model, InputMetadata inputMetadata)
                 var landingData = (LandingData)content;
                 JsonUtility.Merge(inputMetadata.RawJObject, landingData.ExtensionData);
 
-                pageModel = new JObject()
+                pageModel = JsonUtility.ToJObject(new ConceptualModel
```
live1206: do it  the same way as line 176 and 186

src/docfx/build/page/BuildPage.cs
```diff
@@ -269,5 +268,19 @@ private static (object model, JObject metadata) ApplyPageTemplate(Context contex
 
             return (pageModel, processedMetadata);
         }
+
+        [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+        internal class ConceptualModel
```
yufeih: put it in a separate file

src/docfx/build/page/BuildPage.cs
live1206:
```suggestion
            return (errors, pageModel, inputMetadata);
```

OsmondJiang: page model is not a JObject always, it can be a transformed token

yufeih: Need to check and report like I did [here](https://github.com/dotnet/docfx/pull/4810/files#diff-54e86e4d213ae7f283c6a9bf56b3f8f8R48) and [here](https://github.com/dotnet/docfx/pull/4810/files#diff-c329d0f4c7d5076b957537c2818b4a03R238)

PR #4824, opened 2019-07-02
🎈Add validation that same uid should have same name

live1206:
- Add `xref-property-conflict` as a warning and take the first UID order by the declaring file
- Change `moniker-overlapping` to warning and take the first UID order by the declaring file
- remove ordering based on monikers

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4824)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -523,6 +523,17 @@ namespace Microsoft.Docs.Build
         }
 
         /// <summary>
+        /// Same uid defined within different versions with the different name.
+        /// Examples:
+        ///   - Same uid defined in multiple .md files with different versions have different titles.
+        /// </summary>
+        /// Behavior: ❌ Message: ❌
+        public static Error UidNameConflict(string uid, IEnumerable<string> nameConflicts)
+        {
+            return new Error(ErrorLevel.Error, "uid-name-conflict", $"UID '{uid}' is defined with different names: {Join(nameConflicts)}");
+        }
+
+        /// <summary>
         /// Multiple articles with same uid contain overlapped monikers,
         /// and can't decide which article to use when referencing that uid with this overlapped version
         /// </summary>
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -537,7 +537,7 @@ namespace Microsoft.Docs.Build
         /// Behavior: ✔️ Message: ❌
         public static Error UidPropertyConflict(string uid, string propertyName, IEnumerable<string> conflicts)
         {
-            return new Error(ErrorLevel.Warning, "uid-property-conflict", $"UID '{uid}' is defined with different {propertyName}s: {Join(conflicts)}");
+            return new Error(ErrorLevel.Warning, "xref-property-conflict", $"UID '{uid}' is defined with different {propertyName}s: {Join(conflicts)}");
         }
 
         /// <summary>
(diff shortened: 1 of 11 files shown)


test/docfx.Test/e2e/E2ETest.cs
```diff
@@ -32,7 +32,7 @@ public static class E2ETest
             "download-failed", "locale-invalid",
 
             // show multiple errors with line info
-            "publish-url-conflict", "output-path-conflict", "uid-conflict", "redirection-conflict",
+            "publish-url-conflict", "output-path-conflict", "uid-conflict", "uid-name-conflict", "redirection-conflict",
```
yufeih: we should be able to report where the `uid` is defined, but it could be done in another PR.

live1206: We should fix the errors of conflicting at some point.

src/docfx/Errors.cs
```diff
@@ -522,6 +522,17 @@ public static Error UidConflict(string uid, IEnumerable<string> conflicts = null
...
 
+        /// <summary>
+        /// Same uid defined within different versions with the different name.
+        /// Examples:
+        ///   - Same uid defined in multiple .md files with different versions have different titles.
+        /// </summary>
+        /// Behavior: ❌ Message: ❌
+        public static Error UidNameConflict(string uid, IEnumerable<string> nameConflicts)
+        {
+            return new Error(ErrorLevel.Error, "uid-name-conflict", $"UID '{uid}' is defined with different names: {Join(nameConflicts)}");
```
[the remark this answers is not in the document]
live1206: sounds good

src/docfx/Errors.cs
```diff
@@ -529,13 +529,24 @@ public static Error UidConflict(string uid, IEnumerable<string> conflicts = null
...
+        }
+
         /// <summary>
         /// Multiple articles with same uid contain overlapped monikers,
         /// and can't decide which article to use when referencing that uid with this overlapped version
         /// </summary>
         /// Behavior: ✔️ Message: ❌
         public static Error MonikerOverlapping(IEnumerable<string> overlappingmonikers)
-            => new Error(ErrorLevel.Error, "moniker-overlapping", $"Two or more documents have defined overlapping moniker: {Join(overlappingmonikers)}");
+            => new Error(ErrorLevel.Warning, "moniker-overlapping", $"Two or more documents have defined overlapping moniker: {Join(overlappingmonikers)}");
```
928PJY: >moniker-overlapping [](start = 46, length = 19)

since we changed the level and add new error code, do we need to confirm with PM?

live1206: This follows the behavior of `uid-conflict`(warning) from v2, I would think this is fine

live1206: @herohua

src/docfx/build/xref/InternalXrefMapBuilder.cs
```diff
@@ -147,18 +147,25 @@ private static InternalXrefSpec[] AggregateXrefSpecs(Context context, string uid
...
 
-            // uid conflicts with overlapping monikers, drop the uid and log an error
+            // uid conflicts with overlapping monikers
+            // log an warning and take the first one order by the declaring file
             var conflictsWithMoniker = specsWithSameUid.Where(x => x.Monikers.Count > 0).ToArray();
             if (CheckOverlappingMonikers(specsWithSameUid, out var overlappingMonikers))
             {
                 context.ErrorLog.Write(Errors.MonikerOverlapping(overlappingMonikers));
-                return Array.Empty<InternalXrefSpec>();
+                return new InternalXrefSpec[] { conflictsWithMoniker.OrderBy(item => item.DeclairingFile).First() };
```
OsmondJiang: We can simply return the first one of order list which contains all xrefSpec with this UID no matter it has moniker or not by Declaring File here?

live1206: sounds good, this could simplify the resolving logic as well

PR #4828, opened 2019-07-02
JsonSchema: pattern

yufeih:
#4605 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4828)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaValidator.cs
+++ b/src/docfx/lib/schema/JsonSchemaValidator.cs
@@ -171,11 +171,16 @@ namespace Microsoft.Docs.Build
                     errors.Add(Errors.StringLengthInvalid(JsonUtility.GetSourceInfo(scalar), scalar.Path, $">= {schema.MinLength}"));
             }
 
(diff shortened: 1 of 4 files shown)


src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -171,11 +171,16 @@ private void ValidateString(JsonSchema schema, string name, JValue scalar, strin
                     errors.Add(Errors.StringLengthInvalid(JsonUtility.GetSourceInfo(scalar), scalar.Path, $">= {schema.MinLength}"));
             }
 
+            if (schema.Pattern != null && !Regex.IsMatch(str, schema.Pattern))
```
mingwli: >schema.Pattern [](start = 62, length = 14)

what if schema.Pattern is invalid e.g. '[' ?

yufeih: We'll just throw whatever `Regex.IsMatch` throws. Currently we consider template related errors system errors.

PR #4830, opened 2019-07-02
consist conceptual and landing data build

OsmondJiang:
1. define conceptual model, used by building conceptual and landing data
2. null/empty array value's key should be removed from output

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4830)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -384,7 +384,7 @@ namespace Microsoft.Docs.Build
         /// The input value type does not match expected value type.
         /// </summary>
         /// Behavior: ✔️ Message: ❌
-        public static Error UnexpectedType(SourceInfo source, string expectedType, string actualType)
+        public static Error UnexpectedType(SourceInfo source, object expectedType, object actualType)
             => new Error(ErrorLevel.Warning, "unexpected-type", $"Expect type '{expectedType}' but got '{actualType}'", source);
 
         /// <summary>
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -5,9 +5,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Threading.Tasks;
-using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
-using Newtonsoft.Json.Serialization;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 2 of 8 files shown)


src/docfx/build/page/BuildPage.cs
```diff
@@ -212,19 +217,17 @@ private static (List<Error> errors, JObject model, InputMetadata inputMetadata)
...
+            var (schemaTransformError, transformedToken) = schemaTemplate.JsonSchemaTransformer.TransformContent(file, context, obj);
             errors.AddRange(schemaTransformError);
 
             var (metaErrors, inputMetadata) = context.MetadataProvider.GetMetadata(file);
             errors.AddRange(metaErrors);
 
-            var pageMetadata = new OutputMetadata();
-            var pageModel = transformedToken;
-
+            var pageModel = transformedToken as JObject;
```
live1206: where shall we add the type validation and throw `UnexpectedType`?

src/docfx/Errors.cs
```diff
@@ -384,7 +384,7 @@ public static Error ViolateSchema(SourceInfo source, string message)
         /// The input value type does not match expected value type.
         /// </summary>
         /// Behavior: ✔️ Message: ❌
-        public static Error UnexpectedType(SourceInfo source, string expectedType, string actualType)
+        public static Error UnexpectedType(SourceInfo source, object expectedType, object actualType)
```
live1206: why not do a `ToString()` from the caller to avoid the implicit cast?

src/docfx/build/page/BuildPage.cs
OsmondJiang: I add it at the beginning of LoadSchemaDocument

src/docfx/build/xref/InternalXrefMapBuilder.cs
```diff
@@ -98,8 +98,13 @@ private static (Error error, InternalXrefSpec spec, Document doc) LoadMarkdown(C
             return (error, xref, file);
         }
 
-        private static (List<Error> errors, IReadOnlyList<InternalXrefSpec> specs) LoadSchemaDocument(Context context, JObject obj, Document file)
+        private static (List<Error> errors, IReadOnlyList<InternalXrefSpec> specs) LoadSchemaDocument(Context context, JToken token, Document file)
         {
+            if (!(token is JObject obj))
```
yufeih: this check here is not needed

OsmondJiang:
The diff are expected.

PR #4831, opened 2019-07-02
Fix bug of restoring latest file content

OsmondJiang:
since we changed the logic of restoring file to one places calculated by url, we should always overwrite existing file using latest version of restored file. v

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4831)

### DIFF
--- a/src/docfx/restore/RestoreFile.cs
+++ b/src/docfx/restore/RestoreFile.cs
@@ -46,8 +46,10 @@ namespace Microsoft.Docs.Build
             {
                 PathUtility.CreateDirectoryFromFilePath(filePath);
 
-                File.Copy(tempFile, filePath, true);
-                File.Delete(tempFile);
+                if (File.Exists(filePath))
+                    File.Delete(filePath);
+
+                File.Move(tempFile, filePath);
 
                 if (etag != null)
                 {
--- a/src/docfx/restore/RestoreFile.cs
+++ b/src/docfx/restore/RestoreFile.cs
@@ -44,15 +44,10 @@ namespace Microsoft.Docs.Build
 
             using (InterProcessMutex.Create(filePath))
             {
-                if (!File.Exists(filePath))
-                {
-                    PathUtility.CreateDirectoryFromFilePath(filePath);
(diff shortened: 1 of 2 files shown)


src/docfx/restore/RestoreFile.cs
```diff
@@ -44,15 +44,10 @@ public static async Task Restore(string url, Config config)
...
-                    PathUtility.CreateDirectoryFromFilePath(filePath);
-                    File.Move(tempFile, filePath);
-                }
-                else
-                {
-                    File.Delete(tempFile);
-                }
+                PathUtility.CreateDirectoryFromFilePath(filePath);
+
+                File.Copy(tempFile, filePath, true);
```
yufeih: Prefer `File.Move` over `File.Copy` here to prevent data corruption. `File.Move` is mostly an atomic operation while `File.Copy` isn't. When docfx crashed in the middle of `File.Copy` for a large files, the target file exist but only contains partial data.

src/docfx/restore/RestoreFile.cs
```diff
@@ -44,15 +44,12 @@ public static async Task Restore(string url, Config config)
...
-                    PathUtility.CreateDirectoryFromFilePath(filePath);
-                    File.Move(tempFile, filePath);
-                }
-                else
-                {
-                    File.Delete(tempFile);
-                }
+                PathUtility.CreateDirectoryFromFilePath(filePath);
+
+                if (File.Exists(filePath))
```
yufeih: line 49 - 52 can become a single atomic operation using `File.Replace`, it internally forwards the call to [ReplaceFileW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-replacefilew)
```suggestion
                File.Replace(tempFile, filePath, null);
```

OsmondJiang: File.Replace will throw errors when dest file doesn't exists

PR #4840, opened 2019-07-04
share transformed model between transform xref and transform content

OsmondJiang:
refine the transform code, share models between transform xref and transform content, avoid duplicated markup for same file same properties.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4840)


live1206:
Failed   Microsoft.Docs.Build.E2ETest.Run(name: "config/17. latter extend config takes precedence o"...)
Error Message:
 Assert.Equal() Failure
Expected: OrderedEnumerable<String, String> ["a.json"]
Actual:   OrderedEnumerable<String, String> [".errors.log"]

OsmondJiang:
> Failed Microsoft.Docs.Build.E2ETest.Run(name: "config/17. latter extend config takes precedence o"...)
> Error Message:
> Assert.Equal() Failure
> Expected: OrderedEnumerable<String, String> ["a.json"]
> Actual: OrderedEnumerable<String, String> [".errors.log"]

It's not related to this PR, so be related some other issue, opened a bug for tracking.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -13,136 +14,149 @@ internal class JsonSchemaTransformer
...
                 {
-                    var uid = obj.TryGetValue("uid", out var uidValue) && uidValue is JValue uidJValue && uidJValue.Value is string uidStr ? uidStr : null;
+                    case JObject obj:
+                        var uid = obj.TryGetValue("uid", out var uidValue) && uidValue is JValue uidJValue && uidJValue.Value is string uidStr ? uidStr : null;
 
-                    if (uid == null)
-                    {
-                        return (default, node);
-                    }
+                        if (uid == null)
```
live1206:
```suggestion
                        if (uid is null)
```

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -13,136 +14,149 @@ internal class JsonSchemaTransformer
...
+                                }
+                                else
+                                {
+                                    newObject[key] = value;
+                                }
+                            }
+                            return newObject;
+
+                        case JValue value:
+                            return TransformScalar(subSchema, file, context, value, errors);
```
live1206: nit: remove line 157 and replace it with
```csharp
default:
  throw new NotSupportedException();
```

OsmondJiang:
The integration test failed is caused by other issue, #4848 was created to fix this issue by @yufeih

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -13,136 +14,149 @@ internal class JsonSchemaTransformer
...
         public (List<Error> errors, JToken token) TransformContent(Document file, Context context, JToken token)
         {
             var errors = new List<Error>();
-            var transformedToken = Transform(file, context, _schema, token, errors);
+            var transformedToken = TransformToken(file, context, _schema, token, errors);
             return (errors, transformedToken);
         }
 
-        public (List<Error> errors, Dictionary<string, List<(bool isRoot, SourceInfo source, Dictionary<string, Lazy<JToken>> propertiesByUid)>> xrefPropertiesGroupByUid) TransformXref(Document file, Context context, JToken token)
+        public (List<Error> errors, Dictionary<string, List<(bool isRoot, SourceInfo source, Dictionary<string, Lazy<JToken>> propertiesByUid)>> xrefPropertiesGroupByUid) TraverseXref(Document file, Context context, JToken token)
```
yufeih: can simplify the interface to return a list of `InternalXrefSpec`

```suggestion
        public (List<Error> errors, List<InternalXrefSpec> xrefs)>) LoadXrefSpecs(Document file, Context context, JToken token)
```

OsmondJiang: I think we can do this in another PR, since it need change XrefMapBuild also, which is not this PR focus.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -13,136 +14,146 @@ internal class JsonSchemaTransformer
...
+                {
+                    return node;
+                }
 
-                case JObject obj:
+                return _transformedProperties.GetOrAdd(node, _ => Transform());
 
-                    var newObject = new JObject();
-                    foreach (var (key, value) in obj)
+                JToken Transform()
```
yufeih: consider _not_ use nested function here because it is hard to tell what variables it captures. the second parameter to `dictionary.GetOrAdd` needs to be as pure as possible.

In this case, `errors` is captured and used inside `Transform` function, when `TransformScalar` generates an error, the error will be added to the list of the first calls `TransformToken`, other calls to `TransformToken` will not have correct errors.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -13,136 +14,146 @@ internal class JsonSchemaTransformer
...
-                        xrefPropertiesGroupByUid[uid].Add((obj.Parent is null, JsonUtility.GetSourceInfo(obj), BuildXrefPropertiesForUid(schema, obj)));
-                    }
+                        if (!xrefPropertiesGroupByUid.TryGetValue(uid, out var xrefPropertyList))
+                        {
+                            xrefPropertiesGroupByUid[uid] = xrefPropertyList = new List<(bool, SourceInfo, Dictionary<string, Lazy<JToken>>)>();
+                        }
 
-                    return (schema.XrefProperties, node);
+                        var xrefProperties = new Dictionary<string, Lazy<JToken>>();
+                        TraverseObjectXref(obj, (propertySchema, key, value) =>
```
yufeih: prefer calls over callbacks

OsmondJiang: TraverseObjectXref need to be used in two places, just avoid duplicated code

src/docfx/lib/schema/JsonSchemaTransformer.cs
OsmondJiang: good point, will change that

PR #4842, opened 2019-07-04
ignore update_at_date_time

mingwli:
fix https://apidrop.visualstudio.com/Content%20CI/_build/results?buildId=64525

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4842)

### DIFF
--- a/src/docfx/build/page/ContributionInfo.cs
+++ b/src/docfx/build/page/ContributionInfo.cs
@@ -14,6 +14,7 @@ namespace Microsoft.Docs.Build
     {
         public string UpdateAt { get; set; }
 
+        [JsonIgnore]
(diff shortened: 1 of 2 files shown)


src/docfx/build/page/ContributionInfo.cs
```diff
@@ -14,6 +14,7 @@ public class ContributionInfo
     {
         public string UpdateAt { get; set; }
 
+        [JsonIgnore]
```
live1206: `JsonIgnore` will affect during de-serialization as well, is that what we want here?

PR #4850, opened 2019-07-08
✨sort properties during serialization

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4850)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -69,10 +69,6 @@ namespace Microsoft.Docs.Build
                 {
                     context.Output.WriteText(str, publishItem.Path);
                 }
-                else if (output is JObject obj)
-                {
-                    context.Output.WriteJson(obj is null ? null : new JObject(obj.Properties().OrderBy(p => p.Name)), publishItem.Path);
-                }
                 else
                 {
                     context.Output.WriteJson(output, publishItem.Path);
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -1,12 +1,11 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
-using Microsoft.AspNetCore.Mvc.RazorPages;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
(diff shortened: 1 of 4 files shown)


src/docfx/lib/json/JsonContractResolver.cs
```diff
@@ -27,6 +28,11 @@ protected override JsonObjectContract CreateObjectContract(Type objectType)
             return contract;
         }
 
+        protected override IList<JsonProperty> CreateProperties(Type type, MemberSerialization memberSerialization)
```
OsmondJiang: Only works for strong model

src/docfx/build/page/BuildPage.cs
```diff
@@ -37,6 +36,7 @@ public static async Task<List<Error>> Build(Context context, Document file)
 
                 var isConceptual = string.IsNullOrEmpty(file.Mime) || TemplateEngine.IsLandingData(file.Mime);
                 (output, metadata) = ApplyPageTemplate(context, file, mergedMetadata, mergeModel, isConceptual);
+                metadata = metadata is null ? null : new JObject(metadata.Properties().OrderBy(p => p.Name));
```
yufeih: if sort is done here, only `.publish.json` are sorted, `.raw.page.json` and `.mta.json` are not sorted, should sort `mergedMetadata`

live1206: https://github.com/dotnet/docfx/blob/75c0979bc96d94ab170f6b8a2fccda0fff7ff121/src/docfx/build/page/BuildPage.cs#L78
This could cover `.mta.json` as well, but where do we write `.raw.page.json`

yufeih: This is where content in `.raw.page.json` is created: https://github.com/dotnet/docfx/blob/ba2c1d924c6f1ccb7e1efda2258506b03837e777/src/docfx/build/page/BuildPage.cs#L286

live1206: ok, then `output` in line 38 would be the content written into `.raw.page.json`. Shall we sort `output` as well? If we only sort `mergedMetadata`, there are other parts merged afterward, they will not be in order, right?

live1206: But since `pageModel` is actually a strong model, do we care that much about it, do we only focus on the `JObject` part here?

yufeih: there is no merge afterward today, we can sort output when there is need

live1206: https://github.com/dotnet/docfx/blob/ba2c1d924c6f1ccb7e1efda2258506b03837e777/src/docfx/build/page/BuildPage.cs#L35
`pageModel` is merged into `mergeModel` after `mergedMetadata`, and then `mergedMetadata` and `mergeModel` input into `ApplyPageTemplate` to get `metadata` as content of `.mta.json` and `ExtensionData` in `PublishItem`. I still believe it is more straightforward to just sort `metadata` before written.

yufeih: sure, that make sense in the current code flow. Just ensure both `.raw.page.json` and `.mta.json` are sorted.

yufeih:
With this, the `orderby` in `HtmlUtility.CreateHtmlMetaTags` can be removed

PR #4851, opened 2019-07-08
change toc transform from replace to return new model

OsmondJiang:
1. reduce the complexity of code to calculate first child
2. return new model instead of change inputting model
3. reduce the complexity of code to calculate monikers 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4851)

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1526,3 +1526,61 @@ outputs:
             }
         ]
     }
+---
+# don't respect monikers of input toc model
+inputs:
+  docfx.yml: |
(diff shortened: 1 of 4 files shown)


OsmondJiang:
The diff changes seems right, @928PJY  would you please help double confirm?

toc content:

image

OsmondJiang:
@928PJY Here is the v2 output:

image

PR #4856, opened 2019-07-09
 WorkQueue.Drain wait for pending jobs in case of exception

yufeih:
This should be a proper fix over a mitigation

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4856)


OsmondJiang:
I have a little concerns here that we switch ActionBlock to TaskPool here. We did a lot of work for ActionBlock before, now we switch to TaskPool, we may meet same problems before and still need a lot of efforts to fix these problem, just like this exception handling.

PR #4858, opened 2019-07-10
Do not validate bookmark if the target file doesn't report bookmarks

yufeih:
Fixes https://dev.azure.com/ceapex/Engineering/_workitems/edit/104033?src=WorkItemMention&src-action=artifact_link

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4858)

### DIFF
--- a/src/docfx/build/page/BookmarkValidator.cs
+++ b/src/docfx/build/page/BookmarkValidator.cs
@@ -54,7 +54,8 @@ namespace Microsoft.Docs.Build
                 if (bookmark == "top")
                     continue;
 
-                // Do not validate bookmark if the target file doesn't report bookmarks
+                // Do not validate bookmark if the target file doesn't report bookmarks.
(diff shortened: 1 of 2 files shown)


src/docfx/build/page/BookmarkValidator.cs
```diff
@@ -52,13 +52,14 @@ public void Validate()
             {
                 // #top is HTMl predefined URL, which points to the top of the page
                 if (bookmark == "top")
-                {
                     continue;
-                }
-                if (bookmarksByFile.TryGetValue(reference, out var bookmarks) && bookmarks.Contains(bookmark))
-                {
+
+                // Do not validate bookmark if the target file doesn't report bookmarks
```
OsmondJiang: what's the meaning of "doesn't report bookmakrs"? is it built without any bookmarks or not built at all?

PR #4861, opened 2019-07-10
Author can be excluded if it is not specified in YAML header

928PJY:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/104362

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4861)

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -161,7 +161,7 @@ outputs:
 ---
 # Author can be excluded if it is not specified in YAML header
 repos:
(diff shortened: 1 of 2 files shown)


docs/specs/git.yml
```diff
@@ -159,6 +159,44 @@ outputs:
       }
     }
 ---
+# Author can be excluded if it is not specified in YAML header
```
yufeih: can this test case be simplified, minimized?

PR #4862, opened 2019-07-10
fix bug of page date after applying template

OsmondJiang:
page data is calculated from two sources:
1. ms.date which is wrote by writer
2. _gitContributorInformation.updated_at_date_time

the second source will be removed after being used by template to calculate the page date

fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/104093

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4862)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -281,6 +281,13 @@ namespace Microsoft.Docs.Build
                 pageModel.Remove("_op_gitContributorInformation");
                 pageModel.Remove("_op_allContributorsStr");
             }
+
+            // remove updated_at_date_time after applying template - legacy behavior
+            if (pageModel.TryGetValue("_op_gitContributorInformation", out var gitInfoToken) && gitInfoToken is JObject gitInfo && gitInfo.ContainsKey("updated_at_date_time"))
+            {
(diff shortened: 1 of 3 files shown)


src/docfx/build/page/BuildPage.cs
```diff
@@ -281,6 +281,13 @@ private static (TemplateModel model, JObject metadata) TransformToTemplateModel(
                 pageModel.Remove("_op_gitContributorInformation");
                 pageModel.Remove("_op_allContributorsStr");
             }
+
+            // remove updated_at_date_time after applying template - legacy behavior
+            if (pageModel.TryGetValue("_op_gitContributorInformation", out var gitInfoToken) && gitInfoToken is JObject gitInfo && gitInfo.ContainsKey("updated_at_date_time"))
```
yufeih: can this be done in migration tool? It's weird to remove it here for no reason.

OsmondJiang: yes we can, but we need know the reason why it's removed from v2 page model, @fenxu  would you please help to confirm.

OsmondJiang:
diff are expected.

PR #4864, opened 2019-07-11
combine the xref spec builder and json schema xref loader

OsmondJiang:
1. no need sort xref specs built from sdp
2. reduce un-necessary inner function 
3. combine internal xref spec build logic with json schema xref loader

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4864)

### DIFF
--- a/src/docfx/build/xref/IXrefSpec.cs
+++ b/src/docfx/build/xref/IXrefSpec.cs
@@ -7,6 +7,8 @@ namespace Microsoft.Docs.Build
 {
     internal interface IXrefSpec
     {
+        SourceInfo Source { get; }
+
         string Href { get; }
 
         Document DeclaringFile { get; }
--- a/src/docfx/build/xref/IXrefSpec.cs
+++ b/src/docfx/build/xref/IXrefSpec.cs
@@ -7,8 +7,6 @@ namespace Microsoft.Docs.Build
 {
     internal interface IXrefSpec
     {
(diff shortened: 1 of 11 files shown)


src/docfx/build/xref/IXrefSpec.cs
```diff
@@ -7,6 +7,8 @@ namespace Microsoft.Docs.Build
 {
     internal interface IXrefSpec
     {
+        SourceInfo Source { get; }
```
yufeih: is adding source info to `InternalXrefSpec` is enough?

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -5,7 +5,9 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
+using Microsoft.Graph;
```
yufeih: remove this?

OsmondJiang: my IDE will automatically add some strange namespace... :(

yufeih: me too, it is a new annoying IDE feature

PR #4866, opened 2019-07-12
support has error flag in publish item file

OsmondJiang:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/104945

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4866)

### DIFF
--- a/docs/specs/manifest.yml
+++ b/docs/specs/manifest.yml
@@ -238,7 +238,7 @@ outputs:
        }
     }
 ---
-# Manifest does not include files with errors
+# Manifest includes files with errors
 inputs:
   docfx.yml: |
     rules:
@@ -251,8 +251,6 @@ outputs:
       [
         {
           "url": "/docs/a",
-          "path": "docs/a.json",
-          "source_path": "docs/a.md",
           "locale": "en-us",
           "has_error": true
         }
--- a/docs/specs/manifest.yml
+++ b/docs/specs/manifest.yml
@@ -253,7 +253,8 @@ outputs:
           "url": "/docs/a",
           "path": "docs/a.json",
           "source_path": "docs/a.md",
-          "locale": "en-us"
+          "locale": "en-us",
+          "has_error": true
         }
       ]
     }
--- a/docs/specs/manifest.yml
+++ b/docs/specs/manifest.yml
@@ -247,7 +247,16 @@ inputs:
     a [](b)
 outputs:
   .publish.json: |
-    { "!files": null }
+    { "files": 
+      [
+        {
+          "url": "/docs/a",
+          "path": "docs/a.json",
+          "source_path": "docs/a.md",
+          "locale": "en-us"
+        }
+      ]
+    }
   .errors.log: |
     ["error","file-not-found","Invalid file link: 'b'.","docs/a.md",1,3]
 ---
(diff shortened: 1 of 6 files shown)


docs/specs/manifest.yml
```diff
@@ -247,7 +247,17 @@ inputs:
     a [](b)
 outputs:
   .publish.json: |
-    { "!files": null }
+    { "files":
```
yufeih: instead of modifying an exist spec, can you add a new spec that describes what `hasError` does

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -133,6 +124,17 @@ public bool TryAdd(Document file, PublishItem item)
             var fileManifests = _publishItems.ToDictionary(item => item.Key, item => item.Value);
 
             return (model, fileManifests);
+
+            void HandleFileWithError(Document file)
```
yufeih: prefer member function over inner function here

docs/specs/manifest.yml
OsmondJiang: all these modified cased are related to my changes, so I just reuse the case instead of write duplicated multiple cases

src/docfx/publish/PublishModelBuilder.cs
OsmondJiang: what's the reason?

yufeih: It just makes it harder to reason about correctness. Scanning local functions for captured variables is no fun.
https://github.com/dotnet/csharplang/issues/1277 https://github.com/dotnet/csharplang/issues/1565

OsmondJiang: hmm, if we want to avoid the inner function, can we add it to our develop guideline?  seems we are use to use inner function, right?

live1206: I think we should not forbid inner function here. For instance, in the case of recursion, we use inner function to avoid passing along too many inputs.

OsmondJiang:
> # Manifest does not include files with errors

Change the title

---
Refers to: docs/specs/manifest.yml:241 in 821213d. [](commit_id = 821213d98074fedd3760dfd28aa02622ef551f12, deletion_comment = False)

yufeih:
@OsmondJiang , don't forget to ignore this in migration tool

OsmondJiang:
> @OsmondJiang , don't forget to ignore this in migration tool

Sure, i will first fix the migration tool, then merge this PR.

OsmondJiang:
> @OsmondJiang , don't forget to ignore this in migration tool

@yufeih , no need, the .publish.json file is ignored in comparing tool.

image

src/docfx/publish/PublishItem.cs
```diff
@@ -23,6 +23,8 @@ internal class PublishItem
 
         public string RedirectUrl { get; set; }
 
+        public bool HasError { get; set; }
```
yufeih: nit: this adds a `has_error` to every output, maybe we only include it for files that actually has errors:
```suggestion
        [JsonProperty(DefaultValueHandling = DefaultValueHandling.Ignore)]
        public bool HasError { get; set; }
```

yufeih: Or ignore default value for all booleans in JsonUtiltiy.

PR #4867, opened 2019-07-12
log template hash

superyyrrzz:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4867)

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/TemplateProcessors/TemplateManager.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/TemplateProcessors/TemplateManager.cs
@@ -43,6 +43,9 @@ namespace Microsoft.DocAsCode.Build.Engine
             {
                 return null;
             }
+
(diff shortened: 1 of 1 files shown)


src/Microsoft.DocAsCode.Build.Engine/TemplateProcessors/TemplateManager.cs
```diff
@@ -43,6 +43,9 @@ public string GetTemplatesHash()
             {
                 return null;
             }
+
+            Logger.LogVerbose("Calculating template hash...");
```
herohua: this line seems useless

superyyrrzz: Just to be consistent with plugin hash logs. We can also remove them later.

PR #4871, opened 2019-07-15
Fix thread pool starve by adding parallel limit to workqueue

yufeih:
WorkQueue is currently too aggressive and causes starvation.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4871)

### DIFF
--- a/src/docfx/cli/Docfx.cs
+++ b/src/docfx/cli/Docfx.cs
@@ -10,6 +10,7 @@ using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Web;
 
@@ -69,6 +70,8 @@ namespace Microsoft.Docs.Build
             using (var errorLog = new ErrorLog(docset, options.Legacy))
             {
                 Log.Write($"Using docfx {GetDocfxVersion()}");
+
+                ThreadPool.SetMinThreads(32, 32);
 
                 try
                 {
--- a/src/docfx/cli/Docfx.cs
+++ b/src/docfx/cli/Docfx.cs
@@ -71,7 +71,8 @@ namespace Microsoft.Docs.Build
             {
                 Log.Write($"Using docfx {GetDocfxVersion()}");
 
-                ThreadPool.SetMinThreads(32, 32);
+                var minThreads = Math.Max(32, Environment.ProcessorCount * 4);
+                ThreadPool.SetMinThreads(minThreads, minThreads);
 
                 try
                 {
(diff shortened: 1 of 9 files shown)


yufeih:
This has overall positive impact to build performance, can see `azure-docs-pr` build time dropped from ~2:40 to 2:00

yufeih:
`azure-docs-pr` docfx build time on my machine decreased from 3:24 to 2:14 due to this change

src/docfx/cli/Docfx.cs
```diff
@@ -70,6 +71,8 @@ internal static async Task<int> Run(string[] args)
             {
                 Log.Write($"Using docfx {GetDocfxVersion()}");
 
+                ThreadPool.SetMinThreads(32, 32);
```
live1206: should this magic number be related to `Environment.ProcessorCount` as well?

yufeih:
`azure-docs-pr` build time on my machine after removing lazy evaluation:

max parallel unlimited: 2:38
max parallel 16: 2:20

src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -321,12 +321,12 @@ private void ValidateMicrosoftAlias(JsonSchema schema, string name, JValue scala
                 {
                     if (_microsoftGraphCache != null)
                     {
-                        var (error, msAlias) = _microsoftGraphCache.GetMicrosoftAliasAsync(alias).GetAwaiter().GetResult();
+                        var (error, msAlias) = _microsoftGraphCache.GetMicrosoftAlias(alias).GetAwaiter().GetResult();
```
OsmondJiang: could we get ride of this GetResult?

yufeih: I'd like to keep it here for now because:
- It is currently the most straightforward structure
- Most of the time (upon cache hit) the returned Task is a completed task, which doesn't involves wait
- There aren't much performance difference now after the change. Retrieving all msalias using @LowEntropyBody 's benchmark without cache is under 7 seconds (similar to bulk call).

OsmondJiang: Will we plan to change it in the future if you say for now? can we add a TODO here?

yufeih: no plan to change it in the future, added a note there

OsmondJiang: I just want to say that we depends on cache too much, @yufeih

PR #4877, opened 2019-07-15
Metadata validation modify error message

LowEntropyBody:
Related issues: #4605

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4877)


yufeih:
@LowEntropyBody , with this, do we still need to map error code based on metadata name? say `attribute-missing` to `author-missing`, do we still need to support dynamic level adjustment based on the value of metadata validation service ?

LowEntropyBody:
@yufeih For now, it is not supported. Then we will discuss this issue with Megan again. These do not affect the logic of the program.

docs/specs/schema.yml
```diff
@@ -136,7 +136,7 @@ outputs:
   toc.json:
   docs/toc.json:
   .errors.log: |
-    ["warning","unexpected-type","Expect type 'String' but got 'Object'","docs/a.yml",3,18]
+    ["suggestion","unexpected-type","Expected type 'String' but got 'Object'","docs/a.yml",3,18]
```
yufeih: This error code is shared with other scenarios, changing it to suggestion breaks existing build error level.

LowEntropyBody: sure, i will change back

src/docfx/Errors.cs
```diff
@@ -373,16 +373,16 @@ public static Error ViolateSchema(SourceInfo source, string message)
...
 
         /// <summary>
         /// The input value is not defined in a valid value list.
         /// </summary>
-        /// Behavior: ✔️ Message: ❌
-        public static Error UndefinedValue(SourceInfo source, object value, IEnumerable<object> validValues)
-            => new Error(ErrorLevel.Warning, "undefined-value", $"Value '{value}' is not accepted. Valid values: {Join(validValues)}", source);
+        /// Behavior: ✔️ Message: ✔️
+        public static Error InvalidValue(SourceInfo source, string name, object value)
+            => new Error(ErrorLevel.Suggestion, "invalid-value", $"Invalid value for '{name}': '{value}'", source);
```
kimizhu: The ErrorLevel should be defined in schema instead of hardcoding as Suggestion.

src/docfx/Errors.cs
```diff
@@ -373,16 +373,16 @@ public static Error ViolateSchema(SourceInfo source, string message)
         /// <summary>
         /// The input value type does not match expected value type.
         /// </summary>
-        /// Behavior: ✔️ Message: ❌
+        /// Behavior: ✔️ Message: ✔️
         public static Error UnexpectedType(SourceInfo source, object expectedType, object actualType)
-            => new Error(ErrorLevel.Warning, "unexpected-type", $"Expect type '{expectedType}' but got '{actualType}'", source);
+            => new Error(ErrorLevel.Warning, "unexpected-type", $"Expected type '{expectedType}' but got '{actualType}'", source);
```
kimizhu: The by default ErrorLevel for UnexpectedType in docfx is warning? Shall we override it using ErrorLevel in schema?

src/docfx/Errors.cs
```diff
@@ -427,64 +427,67 @@ public static Error NumberInvalid(SourceInfo source, string propName, string cri
...
         /// </summary>
-        /// Behavior: ✔️ Message: ❌
-        public static Error OverDateRange(SourceInfo source, string name, TimeSpan? relativeMinDate, TimeSpan? relativeMaxDate)
-            => new Error(ErrorLevel.Warning, "over-date-range", $"Based on the current time, '{name}' needs to be in this range: {(relativeMinDate.HasValue ? $"{relativeMinDate} <= " : "")}'{name}'{(relativeMaxDate.HasValue ? $" <= {relativeMaxDate}" : "")}", source);
+        /// Behavior: ✔️ Message: ✔️
+        public static Error DateOutOfRange(SourceInfo source, string name, string value)
+            => new Error(ErrorLevel.Suggestion, "date-out-of-range", $"Value out of range for '{name}': '{value}'", source);
 
         /// <summary>
         /// A field is deprecated.
```
kimizhu: Attribute

src/docfx/Errors.cs
LowEntropyBody: This will be done in docs build

src/docfx/Errors.cs
LowEntropyBody: good idea

PR #4878, opened 2019-07-15
redirection files with conflicted document id should be built

OsmondJiang:
as title

fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/105471

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4878)

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -493,3 +493,38 @@ outputs:
   .publish.json: |
   .errors.log: |
       ["error","publish-url-conflict","Two or more files publish to the same url '/a.png': 'a.png', 'docs/a.png'"]
+---
+# Redirection files have conflicted ids
+# all redirection files will be built
+# redirection document id is retrieved from the first redirection file (a.md for below case)
+inputs:
+  docfx.yml: |
+    redirections:
+      a.md: /docs/path1
+      b.md: /docs/path1
+  docs/path1.md:
+outputs:
+  .errors.log: |
+    ["warning","redirection-url-conflict","The '/docs/path1' appears twice or more in the redirection mappings","docfx.yml",3,9]
+  .publish.json: |
+    {
+        "files": [
+            {
+                "url": "/a",
+                "source_path": "a.md",
+                "locale": "en-us",
+                "redirect_url": "/docs/path1"
+            },
+            {
+                "url": "/b",
+                "source_path": "b.md",
+                "locale": "en-us",
+                "redirect_url": "/docs/path1"
+            },
+            {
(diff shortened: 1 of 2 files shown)


src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -47,7 +47,7 @@ public static RedirectionMap Create(ErrorLog errorLog, Docset docset, Func<strin
             AddRedirections(docset.Config.Redirections, redirectDocumentId: true);
             var redirectionsByRedirectionUrl = redirections
                 .GroupBy(file => file.RedirectionUrl, PathUtility.PathComparer)
-                .ToDictionary(group => group.Key, group => group.First(), PathUtility.PathComparer);
+                .ToDictionary(group => group.Key, group => group.OrderBy(r => r).First(), PathUtility.PathComparer);
```
yufeih: If we don't order here, we pick the first one that appears in redirection.json, the implication is that writers can safely add new redirection rules to the end and does not affect document_id of existing articles.

If we sort here, the above assumption is broken, new redirection rules may change document_id of an existing article. What is the expected behavior here?

docs/specs/config.yml
```diff
@@ -493,3 +493,38 @@ outputs:
...
+    redirections:
+      a.md: /docs/path1
+      b.md: /docs/path1
+  docs/path1.md:
+outputs:
+  .errors.log: |
+    ["warning","redirection-url-conflict","The '/docs/path1' appears twice or more in the redirection mappings","docfx.yml",3,9]
+  .publish.json: |
+    {
+        "files": [
```
yufeih: consider condense output expectation to only include what must be verified:

```suggestion
        "files": [
          { "url": "/a", "redirect_url": "/docs/path1" },
          { "url": "/b", "redirect_url": "/docs/path1" },
          { "url": "/docs/path1" },
```

src/docfx/build/redirection/RedirectionMap.cs
OsmondJiang: U r right, we should pick up the first on wrote by user

docs/specs/config.yml
OsmondJiang: considered, the reason why we fix it is we missed one of them in the publish items

OsmondJiang:
@yufeih  have you updated the azure-docs-pr content which may cause this diff?

OsmondJiang:
The diff are expected:

image

PR #4882, opened 2019-07-16
JsonSchema: contains

yufeih:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4882)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaValidator.cs
+++ b/src/docfx/lib/schema/JsonSchemaValidator.cs
@@ -115,7 +115,6 @@ namespace Microsoft.Docs.Build
 
             if (schema.Contains != null && !array.Any(item => Validate(schema.Contains, item).Count == 0))
             {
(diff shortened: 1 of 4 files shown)


src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -337,7 +347,7 @@ private void ValidateMicrosoftAlias(JsonSchema schema, string name, JValue scala
 
         private void ValidateDateRange(JsonSchema schema, string name, JValue scalar, DateTime date, List<Error> errors)
         {
-            var diff = date - DateTime.Now;
+            var diff = date - DateTime.UtcNow;
```
yufeih: @LowEntropyBody should this be `UtcNow` ?

LowEntropyBody: Right, UTC is better

PR #4883, opened 2019-07-16
avoid duplicated call to get user by commits

OsmondJiang:
1. reduce the duplicated call to get user by commits
2. reused the contributors got from commits
3. only filter user by user id, not email.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4883)


src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -120,14 +122,7 @@ async Task<Contributor> GetAuthor()
...
                     // When author name is not specified, last contributor is author
-                    for (var i = contributionCommits.Count - 1; i >= 0; i--)
-                    {
-                        var user = await GetContributor(contributionCommits[i]);
-                        if (user != null && !excludes.Contains(user.Name))
-                        {
-                            return user;
-                        }
-                    }
+                    return contributors[contributors.Count - 1];
```
yufeih: @superyyrrzz to confirm the expected behavior, I remember he added this loop for a purpose

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -74,7 +73,7 @@ public async Task<(List<Error> errors, ContributionInfo contributionInfo)> GetCo
                         continue;
 
                     var contributor = await GetContributor(commit);
-                    if (contributor != null && !excludes.Contains(contributor.Name) && userIds.Add(contributor.Id))
+                    if (contributor != null && !excludes.Contains(contributor.Name))
```
yufeih: Add a debug assertion in line 79 `Log(login)` and line 114 `Log(email)` to ensure a single `login` or `email` is called at most once to prevent bugs like this:

image

src/docfx/build/metadata/ContributionProvider.cs
superyyrrzz: After #4861, these 2 loops can be unified.

PR #4887, opened 2019-07-17
🐛fix xref transform bug

OsmondJiang:
xref transfrom fallback logic:

resolvedHref -> data-raw-html->data-raw-source -> <span class="xref">xref-value</span>

fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/105007

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4887)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Docs.Build
                 var xref = HttpUtility.HtmlDecode(node.GetAttributeValue("href", ""));
 
                 var raw = HttpUtility.HtmlDecode(
-                    node.GetAttributeValue("data-raw-html", null) ?? node.GetAttributeValue("data-raw-source", null) ?? "");
+                    node.GetAttributeValue("data-raw-html", null) ?? node.GetAttributeValue("data-raw-source", null) ?? $"<span class=\"xref\">{UrlUtility.SplitUrl(xref).path}</span>");
 
                 var isShorthand = raw.StartsWith("@");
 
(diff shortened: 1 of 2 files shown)


OsmondJiang:
@yufeih  Integration test sometimes failed with below exception:

image

OsmondJiang:
That's what I got from test PR:

image

yufeih:
@fenxu , log in Docs.Build does not flush on dispose?

OsmondJiang:
> @fenxu , log in Docs.Build does not flush on dispose?

It's related to concurrency append lines to same file, has created a PR to fix it.

https://dev.azure.com/ceapex/Engineering/_git/Docs.Build/pullrequest/13676?_a=files&path=%2Fsrc%2Fshared%2Flogger%2FLogger.cs

src/docfx/lib/HtmlUtility.cs
```diff
@@ -152,7 +152,7 @@ public static string TransformXref(string html, Func<string, bool, int, (string
                 var xref = HttpUtility.HtmlDecode(node.GetAttributeValue("href", ""));
 
                 var raw = HttpUtility.HtmlDecode(
-                    node.GetAttributeValue("data-raw-html", null) ?? node.GetAttributeValue("data-raw-source", null) ?? "");
+                    node.GetAttributeValue("data-raw-html", null) ?? node.GetAttributeValue("data-raw-source", null) ?? $"<span class=\"xref\">{UrlUtility.SplitUrl(xref).path}</span>");
```
yufeih: Also share this logic with `XrefExtension.cs`, line 27, 37 ?

OsmondJiang: @live1206 , do we need share? seems the logic is not totally same

OsmondJiang: We don't need to, the "data-raw-source" generated from markdig here is always itself, so we don't need add fallback logic there.

fenxu:
The log issue is fixed, not flush but concurrent. Some code is removed. Anyway, it works now.

src/docfx/lib/HtmlUtility.cs
yufeih: I see, should HTML encode the content

PR #4889, opened 2019-07-17
output stable order publish manifest

928PJY:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4889)

### DIFF
--- a/src/docfx/publish/PublishModel.cs
+++ b/src/docfx/publish/PublishModel.cs
@@ -13,6 +13,6 @@ namespace Microsoft.Docs.Build
     {
         public PublishItem[] Files { get; set; }
 
-        public SortedDictionary<string, List<string>> MonikerGroups { get; set; } = new SortedDictionary<string, List<string>>();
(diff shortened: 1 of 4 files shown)


src/docfx/publish/PublishModel.cs
```diff
@@ -13,6 +13,6 @@ internal class PublishModel
     {
         public PublishItem[] Files { get; set; }
 
-        public Dictionary<string, List<string>> MonikerGroups { get; set; } = new Dictionary<string, List<string>>();
+        public SortedDictionary<string, List<string>> MonikerGroups { get; set; } = new SortedDictionary<string, List<string>>();
```
live1206:
```suggestion
        public IReadOnlyDictionary<string, List<string>> MonikerGroups { get; set; } = new Dictionary<string, List<string>>();
```

928PJY:
bnjnhgjh

PR #4891, opened 2019-07-18
fix page title and metadata merge order

OsmondJiang:
fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/105452

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4891)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Docs.Build
             var (inputMetaErrors, inputMetadata) = context.MetadataProvider.GetMetadata(file);
             errors.AddRange(inputMetaErrors);
 
-            var (outputMetadataErrors, outputMetadata) = await CreateOutputMetadata(context, file, inputMetadata, sourceModel);
+            var (outputMetadataErrors, outputMetadata) = await CreateOutputMetadata(context, file, inputMetadata);
             errors.AddRange(outputMetadataErrors);
 
             var (output, metadata) = file.IsPage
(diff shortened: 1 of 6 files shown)


src/docfx/lib/json/JsonUtility.cs
```diff
@@ -191,6 +191,17 @@ public static (List<Error> errors, JToken value) Parse(string json, string file)
...
+            if (overwrites == null)
+                return;
+
+            foreach (var overwrite in overwrites)
+            {
+                Merge(container, overwrite);
+            }
+        }
+
         public static void Merge(JObject container, JObject overwrite)
```
live1206: should this method be `private` and only expose `Merge(JObject container, params JObject[] overwrites)`?

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -191,25 +190,14 @@ public static (List<Error> errors, JToken value) Parse(string json, string file)
             }
         }
 
-        public static void Merge(JObject container, JObject overwrite)
+        public static void Merge(JObject container, params JObject[] overwrites)
```
yufeih: keep the original non-allocating Merge method

src/docfx/build/page/BuildPage.cs
```diff
@@ -198,12 +196,13 @@ private static (List<Error> errors, JObject model)
                 errors.Add(Errors.HeadingNotFound(file));
             }
 
+            var (_, inputMetadata) = context.MetadataProvider.GetMetadata(file);
```
yufeih: accumulate errors is

OsmondJiang: it's calculated in build page, here we don't need do it.

OsmondJiang:
image
Integration test failed caused by publishing, @fenxu would you please help to take a look?

src/docfx/build/page/BuildPage.cs
yufeih: I know, but it is a good practice to not relying on behaviors elsewhere.

OsmondJiang:
The diffs are expected, same with PROD page:

image

PR #4905, opened 2019-07-22
Update .editorconfig to include .NET style rules

yufeih:
Adopts `.editorconfig` from https://github.com/dotnet/roslyn/blob/master/.editorconfig

Shows `IDExxx` suggestions in error list for files currently been edited.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4905)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,15 +1,216 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
 root = true
 
+# Don't use tabs for indentation.
 [*]
 charset = utf-8
 indent_style = space
-indent_size = 4
-end_of_line = lf
-insert_final_newline = true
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -1,15 +1,216 @@
...
 charset = utf-8
 indent_style = space
+# (Please don't specify an indent_size here; that has too many unintended consequences.)
+
+# Code files
+[*.{cs,csx,vb,vbx}]
 indent_size = 4
-end_of_line = lf
 insert_final_newline = true
-trim_trailing_whitespace = true
```
herohua: do we want to keep this rule?

yufeih: yep

PR #4907, opened 2019-07-22
Metadata validation: Enumdependenies rule supports self dependency

LowEntropyBody:
Related issues: #4605
Docs Build PR: https://dev.azure.com/ceapex/Engineering/_git/Docs.Build/pullrequest/13861?_a=overview

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4907)


src/docfx/lib/schema/EnumDependenciesSchema.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Docs.Build
+{
+    internal class EnumDependenciesSchema : Dictionary<string, Dictionary<JToken, EnumDependenciesSchema>>
```
yufeih: this is a recursive data structure, did you intent it to support more than 2 levels of dependencies?

LowEntropyBody: Yes, because this rule in v2 also supports recursive model. i think we need to be consistent with it

src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -381,31 +381,88 @@ private void ValidateDeprecated(JsonSchema schema, string name, JToken token, Li
...
+                return;
+            }
+
+            foreach (var (fieldNameWithIndex, allowList) in enumDependencies)
+            {
+                var (fieldName, fieldIndex) = GetFieldNameAndIndex(fieldNameWithIndex);
+                if (map.TryGetValue(fieldName, out var fieldRawValue))
                 {
-                    foreach (var (dependentFieldName, allowLists) in enumDependencyRules)
+                    var fieldValue = GetFieldValue(fieldRawValue, fieldIndex);
```
yufeih:
```suggestion
                    var fieldValue = fieldRawValue is JArray array && fieldIndex < array.Count ? array[fieldIndex] : null;
```

src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -381,31 +381,88 @@ private void ValidateDeprecated(JsonSchema schema, string name, JToken token, Li
...
                     }
                 }
             }
         }
 
+        // For string type: name = 'topic', output = ('topic', 0) or name = 'topic[0]', output = ('topic', 0)
+        // For array type: name = 'topic[0]', output = ('topic', 0) or name = 'topic[1]', output = ('topic', 1) or ...
+        private (string, int) GetFieldNameAndIndex(string name)
+        {
+            var match = Regex.Match(name, @"\[\d+\]$");
```
yufeih: check if name contains `[` and `]` before using `Regex` because `Regex is potentially slow

src/docfx/lib/schema/EnumDependenciesSchema.cs
yufeih: add a little description about the data structure, what the key and value means

src/docfx/lib/schema/JsonSchemaValidator.cs
LowEntropyBody: cool~~

src/docfx/lib/schema/JsonSchemaValidator.cs
LowEntropyBody: make sense

src/docfx/lib/schema/EnumDependenciesSchema.cs
LowEntropyBody: sure

PR #4919, opened 2019-07-25
Fix stackoverflow exception caused by WorkQueue

yufeih:
Fixes https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C7%5C25%5C0c07d4eb-0562-6512-9227-12dbf89d9a0f%5CCommit%5C201907250556214514-docs-build-v3%5Cworkflow_report.html?sv=2016-05-31&sr=b&sig=o7zCEV3x%2F%2B6X%2Fiff%2FRJgM0e5Nk%2FBbPgxqOiOMTlWQ4w%3D&st=2019-07-25T06%3A00%3A03Z&se=2019-08-25T06%3A05%3A03Z&sp=r

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4919)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Generic;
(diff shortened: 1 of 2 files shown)


yufeih:
> can we add a unit test or spec test for this case? seems we have a few tests for our WorkQueue

There is a UT to cover this case (a random exception occurred in the middle of the build): https://github.com/dotnet/docfx/blob/v3/test/docfx.Test/lib/WorkQueueTest.cs#L26

But this only occurs under heavy load, I might need to figure out a different way to test it.

PR #4920, opened 2019-07-25
Metadata Validation: Not validate TOC

LowEntropyBody:
Related issues: #4605
Work item: https://ceapex.visualstudio.com/Engineering/_workitems/edit/108255

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4920)


src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -87,7 +87,10 @@ public MetadataProvider(Docset docset, Cache cache, MicrosoftGraphCache microsof
                 }
             }
 
-            errors.AddRange(_schemaValidator.Validate(result));
+            if (file.ContentType == ContentType.Page && string.IsNullOrEmpty(file.Mime))
```
kimizhu: Is it correct we hardcode the ContentType here?
I see validation rule includes the type such as conceptual, include, is it the same thing?
https://docsmetadatatool.azurewebsites.net/rules

mingwli: this makes it validating against ONLY 'Conceptual'(not other yaml schemas), is this what we want?

OsmondJiang: >file [](start = 16, length = 4)

If (file.ContentType != ContentType.TableOfContents)

yufeih: @kimizhu , I think we will support different rules per types eventually, but we can start with just conceptual content. docfx today have some small problems identifying if a markdown is a include file (token)

yufeih: There is now a `Document.IsConceptual` property

OsmondJiang: Seems he only wants to validate markdown files.

LowEntropyBody: In order to be consistent with v2, currently we only verify markdown file

LowEntropyBody: Sorry for confuse, I have added a commend.

kimizhu: @yufeih what is the dev effort to implement the ability to identify the include files?
The problem here is there are some rules only applies to the conceptual content - for example, title is a required metadata for conceptual file but not required for include file. If docfx v3 cannot tell if the file is include file or not, we cannot apply this rule only on conceptual files.

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -87,7 +87,11 @@ public MetadataProvider(Docset docset, Cache cache, MicrosoftGraphCache microsof
                 }
             }
 
-            errors.AddRange(_schemaValidator.Validate(result));
+            // Only validate markdown files
+            if (file.IsConceptual)
```
OsmondJiang: it contains LandingData?

PR #4922, opened 2019-07-26
Exit code is 0 when unless we have unexpected exceptions

yufeih:
Now that we've moved to `Docs.Build`, we can return 0 in case of user error.

**IMPORTANT**: Merge this **after** we've enabled the new publish pipeline

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4922)


src/docfx/cli/Docfx.cs
```diff
@@ -96,7 +99,7 @@ internal static async Task<int> Run(string[] args)
                 {
                     Log.Write(dex);
                     errorLog.Write(dex.Error, isException: true);
-                    return 1;
+                    return 0;
```
fenxu: The publish.json will still mark error item as has_error, then the change should be safe.

PR #4923, opened 2019-07-26
clear restore map interface for file and git

OsmondJiang:
1. split file and git into two map, because they are using totally different context
2. clear the fallback interface 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4923)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -18,15 +18,10 @@ namespace Microsoft.Docs.Build
             Telemetry.SetRepository(repository?.Remote, repository?.Branch);
 
             var locale = LocalizationUtility.GetLocale(repository?.Remote, repository?.Branch, options);
-            var (restoreGitMap, fallbackRepo) = LoadRestoreMap(docsetPath, locale, repository, options);
-
-            try
+            using (var restoreGitMap = GetRestoreGitMap(docsetPath, locale, options))
             {
-                await Run(docsetPath, repository, locale, options, errorLog, restoreGitMap, fallbackRepo);
-            }
-            finally
-            {
-                restoreGitMap.Release();
+                var (fallbackRepo, fallbackRestoreGitMap) = GetFallbackRepository(docsetPath, repository, restoreGitMap);
+                await Run(docsetPath, repository, locale, options, errorLog, fallbackRestoreGitMap ?? restoreGitMap, fallbackRepo);
             }
         }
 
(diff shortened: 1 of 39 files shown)


src/docfx/build/Build.cs
```diff
@@ -36,7 +36,7 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Erro
             string locale,
             CommandLineOptions options,
             ErrorLog errorLog,
-            RestoreMap restoreMap,
+            RestoreGitMap restoreMap,
```
live1206: change the parameter to `restoreGitMap` as well?

src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -30,8 +30,7 @@ internal class ExternalXrefMapLoader
                 }
                 else
                 {
-                    // TODO: It is very easy to forget passing fallbackDocsetPath, the RestoreMap interface needs improvement
-                    var filePath = RestoreMap.GetRestoredFilePath(docset.DocsetPath, url, docset.FallbackDocset?.DocsetPath);
+                    var filePath = RestoreFileMap.GetRestoredFilePath(docset, url);
```
yufeih: It is possible to abstract away all the unrelated concept, so the caller only needs to pass in an `url`?

```suggestion
                    var filePath = xxx.GetFileRestorePath(url);
```

src/docfx/build/Build.cs
OsmondJiang: sounds good.

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -18,8 +18,6 @@ internal class PublishModelBuilder
 
         public void MarkError(Document file)
         {
-            // TODO: If Error has a Document identifier, we can retrieve files with errors from
-            //       error log without explicitly call PublishModelBuilder.MarkError
             _filesWithErrors.Add(file);
```
yufeih: Leave this TODO: this is about decouple `PublishModelBuilder` from `Document` and use `FilePath` as the key to here

cc @928PJY

OsmondJiang: > Change to hide fallback path as we discussed

I have hidden the fallback path

PR #4925, opened 2019-07-29
support titleOverwriteH1

superyyrrzz:
"title" in YAML header > "titleOverwriteH1" in globalMetadata/fileMetadata > auto extract from H1 > "title" globalMetadata/fileMetadata

https://ceapex.visualstudio.com/Engineering/_workitems/edit/104689?src=WorkItemMention&src-action=artifact_link

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4925)


herohua:
does v3 need to support `titleOverwriteH1`?

yufeih:
Yes, and I think it is good opportunity to consolidate title logic, not just override order, but also null value handling. cc @OsmondJiang

PR #4930, opened 2019-07-29
Added Image Long Description Extension to MarkdigEngine.

bharney0:
Handles accessibility issue with Images that must contain long description. Azure DevOps Work Item can be [found here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/101079)
- transforms image to include visibly-hidden long description in aria-describedby tag that can be read by screen readers.
- Added :::image::: tag to MarkdigEngine
   - Includes source, alt-text, and optional id parameters.
   - unique id is generated for aria-describeby tag if id is not provided by user.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4930)


bharney0:
@jeahyoun @jdanyow please let me know if any changes are required for approval.

yufeih:
@bharney0 , is there an azure devops link to the original work item? On the code side, this seems to be a variant of triple colon extension, can [TripleColonParser](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonParser.cs) be reused?

herohua:
@daxianji007 @yishengjin1413 @wh-alice

bharney0:
@yufeih the Work Item can be [found here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/101079). I tried to use the triple colon extension as my starting point, but I ran into issues trying to implement the first time. I will take another look to see if I can just add another implementation of the ITripleColonBlock similar to Zone and ChromelessForm

wh-alice:
@yishengjin1413 , would you please take a look for this markdig engine change?

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Image/Image.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft. All rights reserved.
```
yishengjin1413: Since you add a new extension contain path need to resolve, you may also need to change this file https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine/ResolveLink/ResolveLinkExtension.cs#L52 . Or when the extension is written in a token, the path will not be resolved correctly.

yishengjin1413: In the same way, you may also need to add a test case about new syntax in token, to check whether the image path is resolved successfully.

jdanyow:
If possible this should reuse the triple colon extension, primarily to take advantage of the existing parsing logic but also there’s a perf hit for each new extension that’s added.

I don’t think we can permit authors to choose their own ids because they have to be unique. Let’s generate them automatically using Guid.

yufeih:
using a guid produces different result for each build, which is bad for incemential publish, can use the current source position as id.

jdanyow:
@yufeih that makes sense. What’s the best example we can refer to for generating an incremental build friendly html id, compatible with inclusion? I imagine this will come up frequently. Could be a good candidate for a shared method if you wanted to add a utility we could all use.

yufeih:
@jdanyow something like below? 

You can get line and column from `processor.Line` and `processor.Column` while inside a `Parser`, `{InclusionContext.File}` gives you a string representation of the current file path relative to docset folder.

```csharp
        public static string GetHtmlId(int line, int column)
        {
            using (var md5 = MD5.Create())
            {
                var id = $"{InclusionContext.File}-{line}-{column}";
                var fileBytes = md5.ComputeHash(Encoding.UTF8.GetBytes(id));

                return new Guid(fileBytes).ToString("N").Substring(0, 5);
            }
        }
```

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Image/ImageParser.cs
```diff
@@ -0,0 +1,289 @@
...
+                var value = attribute.Value;
+                switch (name)
+                {
+                    case "alt-text":
+                        alt = value;
+                        break;
+                    case "source":
+                        src = value;
+                        break;
+                    case "id":
```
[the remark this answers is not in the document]
bharney0: Only reason was because It was in the spec that writers should be able to set "id" by themselves, but based on feedback that "id" must be unique, so we are now only allowing it to be generated.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Image/ImageParser.cs
```diff
@@ -0,0 +1,289 @@
...
+                {
+                    Id = Id.Split('.').First();
+                }
+                //If the user does not have the id attribute set,
+                //then use src filename with randomly generated id to be unique.
+                Random random = new Random();
+                const string chars = "abcdefghijklmnopqrstuvwxyz0123456789";
+                var generated = new string(Enumerable.Repeat(chars, 8)
+                    .Select(s => s[random.Next(s.Length)]).ToArray());
+                Id = $"{Id}-{generated}";
```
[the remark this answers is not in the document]
bharney0: I am now computing Id based on MD5 hash based on feedback from @yufeih

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Image/Image.cs
bharney0: Can you clarify? I'm not sure I understand what change needs to be done to https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine/ResolveLink/ResolveLinkExtension.cs#L52

bharney0:
@jdanyow unfortunately the `TripleColonExtension` has a different ending syntax from `ImageExtension`
```
::: zone target=""chromeless""
::: form action=""create-resource"" submitText=""Create"" :::
::: zone-end
```
vs
```
:::image source=""example.jpg"" alt-text=""example""::: 
Lorem Ipsum
:::image-end:::
```
The only difference being the ending triple colon after the end tag. :::zone-end vs. :::image-end::: And because the `TryContinue` method does not allow both ending syntax I needed to create a new extension or differ from the spec of the extension work item. I was not able to find a way to include the ending triple colon without breaking the existing extensions for zone and chromeless.

yufeih:
@bharney0 @jdanyow It is a spec problem? The triple colon extension was invented to provide a consistent markdown extension experience, seems to me like `:::` as an line ending is optional.

bharney0:
@yufeih @jdanyow I refactored the work to use the `TripleColonExtension` and added a `EndingTripleColons` flag to the Extension Interface to indicate that the tag ends with TripleColon. Let me know if this works better with the existing design.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Image/Image.cs
yishengjin1413: Sure. You can take a look about this PR https://github.com/bharney0/docfx/pull/1/files. It include one unit test and the related code changes. The code should work but you may need to do some refactor~

jdanyow:
@bharney I think it will confuse authors to have multiple triple colon syntaxes. Moniker range, chromeless forms, and zones don't require the trailing `:::`. Would folks be ok with adjusting the image long description spec to align with the other extensions?

bharney0:
@jdanyow I agree that this will cause confusion with the two types of syntaxes. However, based on communications with PM/Megan, the inconsistencies will be addressed in future Pull Requests. For now merging these changes and closing this one out.

jdanyow:
@bharney0 sounds good 👌 
Thanks for checking into it 👏

bharney0:
@yishengjin1413 @yufeih I guess I don't have permission to merge even with approval. Can you merge? or let me know how I can get these changes in?

PR #4932, opened 2019-07-30
skip recursive lookup value for mustache template

OsmondJiang:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/108774

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4932)

### DIFF
--- a/src/docfx/template/MustacheTemplate.cs
+++ b/src/docfx/template/MustacheTemplate.cs
@@ -16,14 +16,11 @@ namespace Microsoft.Docs.Build
     {
         private readonly string _templateDir;
         private readonly IStubbleRenderer _renderer;
-        private readonly RenderSettings _renderSettings;
         private readonly ConcurrentDictionary<string, Lazy<string>> _templates = new ConcurrentDictionary<string, Lazy<string>>();
 
         public MustacheTemplate(string templateDir)
         {
             _templateDir = templateDir;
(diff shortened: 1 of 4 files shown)


OsmondJiang:
Seems we can't simply skip recursive look up for all, otherwise it can't support below case which is used in many places:

image

@yufeih  @mingwli

src/docfx/template/MustacheTemplate.cs
```diff
@@ -52,14 +51,15 @@ private static RendererSettingsBuilder UseJson(RendererSettingsBuilder settings)
 
             return settings.AddValueGetter(typeof(JObject), GetJObjectValue)
                            .AddTruthyCheck<JObject>(value => value != null)
+                           .AddTruthyCheck<JValue>(value => value.Type == JTokenType.Null)
```
yufeih: add a test case:
```suggestion
                           .AddTruthyCheck<JValue>(value => value.Type != JTokenType.Null)
```

OsmondJiang: oooopsss

OsmondJiang:
the diffs are expected

PR #4935, opened 2019-07-30
🧪Add Microsoft.DocAsTest

yufeih:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/109125/

There are the noticeable changes to the test framework :

- `repo` is replaced with `repos`, we always write the content of repos inside test.
- Replace `commands` with `locale`, `legacy`, `build` and `restore` switches to simplify test logic
- [from loc] test is not a separate test in test explorer, because there is no capability to expand a test with custom rules in Microsoft.DocAsTest
- We always run the first repo, for [from loc] test, always run the first loc repo.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4935)


test/docfx.Test/DocfxTest.cs
```diff
@@ -0,0 +1,313 @@
...
+
+        private async static Task RunCore(string docsetPath, string outputPath, DocfxTestSpec spec)
+        {
+            if (spec.Watch)
+            {
+                await RunWatch(docsetPath, spec);
+                return;
+            }
+
+            await RunBuild(docsetPath, outputPath, spec, spec.Locale);
```
OsmondJiang: >await RunBuild(docsetPath, outputPath, spec, spec.Locale); [](start = 12, length = 58)

add a way to debug [From Loc]

PR #4941, opened 2019-07-31
pass along unknown queries while resolving uid

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/109169

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4941)

### DIFF
--- a/src/docfx/build/dependency/DependencyResolver.cs
+++ b/src/docfx/build/dependency/DependencyResolver.cs
@@ -125,7 +125,10 @@ namespace Microsoft.Docs.Build
                 {
                     queries["view"] = moniker;
                 }
-                resolvedHref = UrlUtility.MergeUrl(resolvedHref, "?" + string.Join('&', queries), fragment.Length == 0 ? "" : fragment.Substring(1));
+                resolvedHref = UrlUtility.MergeUrl(
+                    resolvedHref,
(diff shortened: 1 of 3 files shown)


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -119,8 +121,11 @@ internal class DependencyResolver
 
             if (!string.IsNullOrEmpty(resolvedHref))
             {
-                var monikerQuery = !string.IsNullOrEmpty(moniker) ? $"view={moniker}" : "";
-                resolvedHref = UrlUtility.MergeUrl(resolvedHref, monikerQuery, fragment.Length == 0 ? "" : fragment.Substring(1));
+                if (!string.IsNullOrEmpty(moniker))
+                {
+                    queries["view"] = moniker;
+                }
+                resolvedHref = UrlUtility.MergeUrl(resolvedHref, "?" + string.Join('&', queries), fragment.Length == 0 ? "" : fragment.Substring(1));
```
yufeih: `queries` might be null here, can you make it such that `queries is always not null?

live1206: sounds good

PR #4944, opened 2019-08-01
Update XrefMap model for JSON serialization

live1206:
xrefmap.yml:
```yml
### YamlMime:XRefMap
sorted: true
tags: []
properties:
  vendor: aspnet-core-conceptual-ppe-master
  tags:
  - /aspnet-core-conceptual-ppe
  - internal
```
xrefmap.json:
```json
{
    "Sorted": true,
    "Tags": [],
    "Others": {
        "properties": {
            "vendor": "aspnet-core-conceptual-ppe-master",
            "tags": [
                "/aspnet-core-conceptual-ppe",
                "internal"
            ]
        }
    }
}
```
Fix to let JSON serialization have the same output as YAML serialization

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4944)

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Engine/XRefMaps/XRefMap.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/XRefMaps/XRefMap.cs
@@ -12,28 +12,37 @@ namespace Microsoft.DocAsCode.Build.Engine
     using Microsoft.DocAsCode.Common;
     using Microsoft.DocAsCode.Plugins;
     using Microsoft.DocAsCode.YamlSerialization;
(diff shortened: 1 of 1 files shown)


herohua:
do we need any unit test to cover this change?

live1206:
> do we need any unit test to cover this change?

This is consumed in OP.Build. And the JSON output will be consumed by xref service.
We found this while switching xref service to consume JSON instead of YAML.
So this should be covered by the integration.

PR #4946, opened 2019-08-01
refine fallback and localized docset(repository)

OsmondJiang:
currently fallback/localized docset and fallback/localized repository repository is not really clear. sometime we are using repository, sometimes we are using docset.

And also the fallback/localized docset make the code hard to read and maintain, restore file map and restore git map are used anywhere.

this pr is try to use 
1. localized/fallback docset as much as possible
2. reduce the complexity of current docset and fallback/localized docset.
3. hide fallback docset in the pipeline
4. hide restore file map and restore git map, use it in the init of context

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4946)

### DIFF
--- a/src/docfx/build/context/BuildScope.cs
+++ b/src/docfx/build/context/BuildScope.cs
@@ -56,14 +56,33 @@ namespace Microsoft.Docs.Build
             return _fileNames.TryGetValue(fileName, out actualFileName);
         }
 
-        public Docset GetDocset(Docset docset)
-            => docset == _fallbackDocset ? _docset : null;
-
         public Docset GetFallbackDocset(Docset docset)
-            => docset == _docset || IsFallbackDocset(docset) ? _fallbackDocset : null;
+            => docset == _docset || docset == _fallbackDocset ? _fallbackDocset : null;
 
-        public bool IsFallbackDocset(Docset docset)
-            => docset == _fallbackDocset;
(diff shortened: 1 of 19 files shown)


src/docfx/build/context/BuildScope.cs
```diff
@@ -25,17 +26,22 @@ internal class BuildScope
 
         public RedirectionMap Redirections { get; }
 
-        public BuildScope(ErrorLog errorLog, Docset docset, TemplateEngine templateEngine)
+        public Docset Docset { get; }
```
yufeih: Can this be private?

```suggestion
        private readonly Docset _docset;
```

src/docfx/build/context/BuildScope.cs
```diff
@@ -52,6 +58,12 @@ public bool GetActualFileName(string fileName, out string actualFileName)
             return _fileNames.TryGetValue(fileName, out actualFileName);
         }
 
+        public Docset GetFallbackDocset(Docset docset)
+            => docset == Docset || IsFallbackDocset(docset) ? _fallbackDocset : null;
+
+        public bool IsFallbackDocset(Docset docset)
```
yufeih: is it possible to detect if a `file` is a fallback by checking `file.FilePath.Origin == Fallback` ?

OsmondJiang: They are actually same, sometimes we don't have file, only docset

PR #4959, opened 2019-08-05
add log when template changes

superyyrrzz:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4959)


src/Microsoft.DocAsCode.Common/Loggers/InfoCodes.cs
```diff
@@ -20,6 +20,7 @@ public static class FullBuildReason
             public const string NoAvailableGroupCache = "NoAvailableGroupCache";
             public const string ConfigChanged = "ConfigChanged";
             public const string ForceRebuild = "ForceRebuild";
+            public const string TemplateChanged = "TemplateChanged";
```
herohua: >TemplateChanged [](start = 51, length = 15)

would this impact Real-Time Insight service?

superyyrrzz: I think no. This just adds a code that they do not recognize.

PR #4961, opened 2019-08-05
Adds Azure DevOps url support

chrisrichards:
Adds support for the new Azure DevOps url format.

Fixes issue #4493 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4961)

### DIFF
--- a/RELEASENOTE.md
+++ b/RELEASENOTE.md
@@ -1,5 +1,8 @@
-Version Notes (Current Version: v2.44)
+Version Notes (Current Version: v2.45)
 =======================================
+
+v2.45(Pre-Release)
+-----------
 
 v2.44
 -----------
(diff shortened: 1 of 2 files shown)


superyyrrzz:
Hi @chrisrichards, thank you for your PR. I am thinking could you update `testRegex` in `vso` to achieve this. Thus we can avoid duplicate `generateUrl` code. Actually this makes sense as Azure DevOps is renamed from VSO.

superyyrrzz:
It would be better to support both visualstudio.com and azure.com, as both are supported by Azure DevOps now. You can use a regex to support both, like `/^(https?:\/\/)?(ssh:\/\/\S+\@)?(\S+@)?(\S+\.)?(dev\.azure|visualstudio)\.com(\/|:).*/i`. Keep the comment for old and new form.

superyyrrzz:
@ChrisChinchilla Thank you for contribution!

PR #4962, opened 2019-08-06
respects uid content type to generate xref properties

OsmondJiang:
fix: https://ceapex.visualstudio.com/_workitems/edit/110999?fullScreen=True and https://ceapex.visualstudio.com/_workitems/edit/111001?fullScreen=True

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4962)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -144,6 +144,7 @@ namespace Microsoft.Docs.Build
                 {
                     // transform array and object is not supported yet
                     case JArray array:
+
                         var newArray = new JArray();
                         foreach (var item in array)
                         {
@@ -155,6 +156,7 @@ namespace Microsoft.Docs.Build
                         return (errors, newArray);
 
                     case JObject obj:
+
                         var newObject = new JObject();
                         foreach (var (key, value) in obj)
                         {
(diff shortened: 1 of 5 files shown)


OsmondJiang:
check this: 
image

OsmondJiang:
Diff are expected, take DevSandbox as example, the v2 also throw uid not found errors:

image

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -62,6 +67,7 @@ List<InternalXrefSpec> LoadXrefSpecs(JsonSchema schema, JToken node)
 
                         break;
                     case JArray array:
+
```
mingwli: an empty line

live1206: remove blank line

OsmondJiang: we leave a blank line under `case` habitually

live1206: do we? I barely see that

OsmondJiang: removed, follow the existing style.

PR #4969, opened 2019-08-07
toc_rel calculation sort by site path then file path

mingwli:
fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/110166

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4969)

### DIFF
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -402,6 +402,7 @@ inputs:
   docfx.yml: |
     template: {APP_BASE_PATH}data/template
     monikerRange:
+      'docs/**': 'netcore-2.0'
       'docs/v1/**': 'netcore-1.0'
       'docs/v2/**': 'netcore-2.0'
     baseUrl: https://docs.com/docs
@@ -410,10 +411,13 @@ inputs:
       docs/v1/: .
       docs/v2/: .
     monikerDefinition: monikerDefinition.json
-  docs/v1/TOC.md:
+  docs/v1/TOC.md: |
+    # [link](../c.md)
   docs/v1/f1/a.md:
-  docs/v2/TOC.md:
+  docs/v2/TOC.md: |
+    # [link](../c.md)
   docs/v2/f1/a.md:
+  docs/c.md:
   monikerDefinition.json: |
     {
       "monikers": [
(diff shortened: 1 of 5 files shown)


src/docfx/build/toc/TableOfContentsMap.cs
```diff
@@ -164,25 +168,25 @@ public TocCandidate(int subDirectoryCount, int parentDirectoryCount, Document to
...
             }
-
-            var parentDirCompareResult = candidateX.ParentDirectoryCount - candidateY.ParentDirectoryCount;
-            if (parentDirCompareResult != 0)
-            {
-                return parentDirCompareResult;
-            }
-
-            return candidateX.Toc.CompareTo(candidateY.Toc);
+            if (result == 0)
```
OsmondJiang: first compare sitePath, then compare monikers?

mingwli: the relative path calculation happens here: https://github.com/dotnet/docfx/pull/4969/files#diff-b15f33cf52e14d45fdb52ca50e899984R127

docs/specs/legacy.yml
```diff
@@ -410,10 +411,13 @@ inputs:
       docs/v1/: .
       docs/v2/: .
     monikerDefinition: monikerDefinition.json
-  docs/v1/TOC.md:
+  docs/v1/TOC.md: |
+    # [link](../c.md)
   docs/v1/f1/a.md:
-  docs/v2/TOC.md:
+  docs/v2/TOC.md: |
+    # [link](../c.md)
```
OsmondJiang: Add a new case instead of changing the existing one, seems the existing one already covered this scenario

src/docfx/build/toc/TableOfContentsMap.cs
```diff
@@ -164,25 +168,28 @@ public TocCandidate(int subDirectoryCount, int parentDirectoryCount, Document to
...
+                result = candidateX.Toc.FilePath.CompareTo(candidateY.Toc.FilePath);
```
OsmondJiang: finally we still need compare `FilePath`?

docs/specs/legacy.yml
mingwli: reverted, and this case already cover file path ordering

src/docfx/build/toc/TableOfContentsMap.cs
mingwli: yes, when site/file relative path are both the same, we need to have a stable result

src/docfx/build/toc/TableOfContentsMap.cs
```diff
@@ -147,42 +147,45 @@ private static (int subDirectoryCount, int parentDirectoryCount)
 
         private sealed class TocCandidate
         {
-            public int SubDirectoryCount { get; }
+            private Document _referencedFile;
```
OsmondJiang: no needed

PR #4972, opened 2019-08-08
Add JsonDiff to Microsoft.DocsAsTest

yufeih:
### JsonDiff

Add `JsonDiff` to `Microsoft.DocsAsTest` and integrate it with docfx tests. 

image

### JsonDiff pipeline

`JsonDiff` turns expected and actual object into JSON and produce a diff if they don't match. It comes with an extensible normalization pipeline to transform the objects before the are compared.

`Microsoft.DocAsTest` comes with several reusable normalization building blocks:

name |  description | example
-------|-----------------------------------------------|-------------
UseIgnoreNull | Ignore the actual result of a property if the expected value is null | `{ "a": null }` matches `{”a": xxx}` where `xxx` could be anything
UseNegate | Assert the actual result must not be the expected result if the expected value starts with `!` | Given an expectation of `{ "a": "!value"}` , `{"a": "value"} should fail
UseRegex | The value must match a regex if the expectation looks like `/{regex}/` | `{"a": "/^.*$/" }` uses `^.*$` to match the actual value
UseWildcard | Use wildcard match if the expectation contains `*` | `{"a": "a*" }` uses wildcard `a*` to match against actual value
UseJson | The value must be a JSON string and its value is verified using the current pipeline, all the above rules will apply if they are configured in the pipeline  | 
UseHtml | The value must be an HTML string and its value is normalized before verification |

Docfx tests implements it own log file comparison using extensibility model provided by `JsonDiff`.

### Breaking changes to existing docfx tests:

- Assert a JSON property does not exist using `{ "key": undefined }` over `{ "!key": null }`
- All output verification including files verification uses **contain equals** semantic, just like how we currently verify output json file. To assert a file does not exist, use ` { “filename": undefined }`
- Drop yaml as json for output verification
- Add strict assertion to verify we never touch input files in any case. Updated some dependency lock tests accordingly.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4972)


src/Microsoft.DocAsTest/JsonDiff/JsonDiffExtensions.cs
```diff
@@ -0,0 +1,134 @@
...
+                }
+                return (expected, actual);
+            });
+        }
+
+        public static JsonDiffBuilder UseJson(this JsonDiffBuilder builder, JsonDiffOptions options = default)
+            => UseJson(builder, IsFile(".json"), options);
+
+        public static JsonDiffBuilder UseJson(this JsonDiffBuilder builder, JsonDiffPredicate predicate, JsonDiffOptions options = default)
+        {
```
mingwli: no need to overload?

```
return builder.Use(IsFile(".json"), (expected, actual, name, diff) =>
            {
```

src/Microsoft.DocAsTest/JsonDiff/JsonDiffExtensions.cs
```diff
@@ -80,23 +104,72 @@ public static JsonDiffBuilder UseWildcard(this JsonDiffBuilder builder)
...
+            });
+        }
+
+        /// <summary>
+        /// Assert the actual value must be a JSON string that matches the expected JSON string.
+        /// </summary>
+        /// <example>
+        /// Given the expectation "{ \"a\": 1 }", ""{ \"a\": 1 }"" pass but "{ \"a\": 2 }" fail.
+        /// </example>
+        public static JsonDiffBuilder UseJson(this JsonDiffBuilder builder, JsonDiffPredicate predicate = null, JsonDiff jsonDiff = null)
```
live1206: Is there a usage of `predicate` and `jsonDiff`? same as `UseHtml`

test/Microsoft.DocAsTest.Test/JsonDiffTest.cs
```diff
@@ -36,10 +36,7 @@ public class JsonDiffTest
         [InlineData("{'a': 1, 'b': { 'c': 2 }}", "{'a': 1, 'b': { 'c': 3 }}", @"
 -    'c': 2
 +    'c': 3")]
-        [InlineData("{'a': 1}", "{'a': 1, 'b': 2}", @"
--  'a': 1
-+  'a': 1,
-+  'b': 2")]
+        [InlineData("{'a': 1}", "{'a': 1, 'b': 2}", "")]
```
live1206: shall we have some tests for `JsonDiffBuilder` without `UseAdditionalProperties`?
Or can we construct the tests into different ones, each test against one `Use` extension method? That would be easier to see what is failing as well

yufeih: yeah, that's better

PR #4976, opened 2019-08-09
Change log level to warning for internal-bookmark-not-found

v-pegao:
Related workitem is [here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/108157/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4976)

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -342,7 +342,7 @@ outputs:
   docs/d.json:
   .errors.log: |
     ["warning","external-bookmark-not-found","Cannot find bookmark '#title-3' in 'docs/a.md', did you mean '#title-1'?","docs/b.md",2,1]
-    ["suggestion","internal-bookmark-not-found","Cannot find bookmark '#title-3' in 'docs/a.md', did you mean '#title-1'? NOTE: This Suggestion will be elevated to a Warning on 8/26/2019. Please fix invalid bookmarks as soon as possible.","docs/a.md",4,1]
+    ["warning","internal-bookmark-not-found","Cannot find bookmark '#title-3' in 'docs/a.md', did you mean '#title-1'?","docs/a.md",4,1]
     ["warning","external-bookmark-not-found","Cannot find bookmark '#title-3' in 'docs/a.md', did you mean '#title-1'?","docs/d.md",1,1]
 ---
 # only validate bookmarks when the referenced file existed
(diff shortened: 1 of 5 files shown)


docs/specs/markdown.yml
```diff
@@ -342,7 +342,7 @@ outputs:
   docs/d.json:
   .errors.log: |
     ["warning","external-bookmark-not-found","Cannot find bookmark '#title-3' in 'docs/a.md', did you mean '#title-1'?","docs/b.md",2,1]
-    ["suggestion","internal-bookmark-not-found","Cannot find bookmark '#title-3' in 'docs/a.md', did you mean '#title-1'? NOTE: This Suggestion will be elevated to a Warning on 8/26/2019. Please fix invalid bookmarks as soon as possible.","docs/a.md",4,1]
+    ["warning","internal-bookmark-not-found","Cannot find bookmark '#title-3' in 'docs/a.md', did you mean '#title-1'?","docs/a.md",4,1]
```
yufeih: If they are both warnings now, can we consolidate the error code be the same `bookmark-not-found`?

v-pegao: Updated.

mingwli: This requires the diff tool to be updated too or it may influence out monitored repo states?

src/docfx/build/page/BookmarkValidator.cs
```diff
@@ -59,9 +59,7 @@ public void Validate()
                 if (bookmarks.Contains(bookmark))
                     continue;
 
-                var error = isSelfBookmark
-                    ? Errors.InternalBookmarkNotFound(source, file, bookmark, bookmarks)
-                    : Errors.ExternalBookmarkNotFound(source, reference, bookmark, bookmarks);
+                var error = Errors.BookmarkNotFound(source, isSelfBookmark ? file : reference, bookmark, bookmarks);
```
yufeih: remove isSelfBookmark here and in line 25. In `DependencyResolver.cs` line 82, pass `(declaringFile, file, fragment, path)`, also remove `// NOTE: bookmark validation result depend on current inclusion stack`

```suggestion
                var error = Errors.BookmarkNotFound(source, reference, bookmark, bookmarks);
```

docs/specs/markdown.yml
herohua: This is **behavior** change - we may need to confirm with PM first before running it. My personal suggestion would be still using two different log codes as separate log codes could be helpful to BI.

v-pegao: @yufeih @mingwli @herohua Reverted the behavior change on log code, We can open a new PR for this after confirmation.

PR #4978, opened 2019-08-09
Add back workqueue assertions

yufeih:
The assertion now take place after duplication detection #4764 

Make `run` a parameter of `Drain` to express the intent that we only execute and item when calling `Drain`, it is not executed upon `Enqueue`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4978)

### DIFF
--- a/src/docfx/lib/collections/WorkQueue.cs
+++ b/src/docfx/lib/collections/WorkQueue.cs
@@ -13,16 +13,10 @@ namespace Microsoft.Docs.Build
     {
         private static readonly int s_maxParallelism = Math.Max(8, Environment.ProcessorCount * 2);
 
-        private readonly Func<T, Task> _run;
         private readonly ConcurrentQueue<T> _queue = new ConcurrentQueue<T>();
         private readonly ConcurrentHashSet<T> _duplicationDetector = new ConcurrentHashSet<T>();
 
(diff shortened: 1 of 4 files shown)


src/docfx/lib/collections/WorkQueue.cs
```diff
@@ -54,13 +48,18 @@ public void Enqueue(T item)
...
+            }
+
             _queue.Enqueue(item);
 
             Interlocked.Increment(ref _totalCount);
             Interlocked.Increment(ref _remainingCount);
         }
 
-        public Task Drain(Action<int, int> progress = null)
+        public Task Drain(Func<T, Task> run, Action<int, int> progress = null)
```
OsmondJiang: When I call Drain more than one time, they will execute different actions to the same queue, which will brings unexpected result.

Unless we need make sure the Drain can be called only once here.

src/docfx/lib/collections/WorkQueue.cs
```diff
@@ -54,17 +50,22 @@ public void Enqueue(T item)
...
         {
-            if (_queue.Count == 0)
+            if (Interlocked.CompareExchange(ref _drained, 1, 0) == 1)
```
OsmondJiang: 👍

PR #4979, opened 2019-08-09
Stop calling github in case of non-transient fatel errors

yufeih:
Fixes build perf problem found in azure-docs-pr test: https://opbuildstoragesandbox2.blob.core.windows.net/report/2019%5C8%5C9%5C0c07d4eb-0562-6512-9227-12dbf89d9a0f%5CCommit%5C201908090540344152-juchen-v3-perf-5%5Cworkflow_report.html?sv=2016-05-31&sr=b&sig=ztKxP1q1qvQIiu8LqnQbdjHavcWxj5UAMJre3hCM8gk%3D&st=2019-08-09T06%3A05%3A44Z&se=2019-09-09T06%3A10%3A44Z&sp=r

The root cause is we've triggered GitHub abuse detection, an error that is unknown to the current `GitHubAccessor`, thus we keep calling github even if every call is a failure.

This PR makes the following changes to GitHubAccessor:

- Stop calling GitHub in case of __any__ error
- Add a guard to ensure we only call GitHub once for each login or email
- Add both `author.email` and `author.user.email` to `GetUserByCommit` response. `author.email` is the email associated with the local commit, `author.user.email` is the email address associated with GitHub.
- Turn resolveUsers to true by default, and only resolve when GitHub auth token is provided. Since graphql always needed an auth token.
- Unifies graphql error handling

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4979)


src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -58,87 +58,35 @@ public async Task<(Error, GitHubUser)> GetUserByLogin(string login)
...
+                Name = string.IsNullOrEmpty(data.user.name) ? data.user.login : data.user.name,
+                Emails = new[] { data.user.email }.Where(email => !string.IsNullOrEmpty(email)).ToArray(),
+            });
+        }
 
-            if (_unauthorizedError != null)
+        public async Task<(Error, IEnumerable<GitHubUser>)> GetUsersByCommit(string owner, string name, string commit, string authorEmail = null)
+        {
+            // Stop calling GitHub whenever any error is returned from github
+            if (_error != null)
```
OsmondJiang: that's too aggressive, one single network issue error doesn't means we can't use github anymore

yufeih: transient errors are covered by retry, the errors here are non-transient errors.

OsmondJiang: we only retry for three times, can't cover all temporary transient errors

yufeih: updated to only cache non-transient errors

docs/specs/localization.yml
```diff
@@ -745,9 +745,6 @@ repos:
           localization:
             mapping: Branch
             bilingual: true
-          gitHub:
-            userCache: github-user.json
```
OsmondJiang: don't we need userCache config anymore?

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -58,87 +56,33 @@ public async Task<(Error, GitHubUser)> GetUserByLogin(string login)
...
+            });
+        }
 
-            if (_unauthorizedError != null)
+        public async Task<(Error, IEnumerable<GitHubUser>)> GetUsersByCommit(string owner, string name, string commit, string authorEmail = null)
+        {
+            if (_fatalError != null)
             {
-                return default;
+                return (_fatalError, default);
```
OsmondJiang: >_fatalError [](start = 24, length = 11)

that will cause too many same errors reported

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -162,127 +107,97 @@ public async Task<(Error, IEnumerable<GitHubUser>)> GetUsersByCommit(string owne
...
                         {
-                            githubUsers.Add(new GitHubUser
+                            switch (error.type)
                             {
-                                Id = node.author.user?.databaseId,
-                                Login = node.author.user?.login,
-                                Name = string.IsNullOrEmpty(node.author.user?.name) ? node.author.user?.login : node.author.user?.name,
-                                Emails = new[] { node.author.email },
-                            });
+                                case "NOT_FOUND":
```
OsmondJiang: >NOT_FOUND [](start = 38, length = 9)

NOT_FOUND for repository and user need different ways to handle.

src/docfx/lib/github/GitHubAccessor.cs
yufeih: Same as before, `github-api-failed` is a warning, @superyyrrzz originally wants to let users know which articles are affected by github failure by report the error for each affected article.

docs/specs/localization.yml
yufeih: there is no such config now, it is a known location in cache folder.

src/docfx/lib/github/GitHubAccessor.cs
yufeih: hmm, seems we've updated the behavior to only report the first error.

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -38,16 +41,13 @@ public void Dispose()
...
-                return (_rateLimitError, null);
-            }
-
-            if (_unauthorizedError != null)
+            if (_fatalError != null)
             {
                 return default;
             }
 
+            Log.Write($"Calling GitHub user API to resolve {login}");
```
OsmondJiang: Do we trace too much? We already trace them in GitHubUserCache?

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -162,127 +114,102 @@ public async Task<(Error, IEnumerable<GitHubUser>)> GetUsersByCommit(string owne
...
+                    .Or<OperationCanceledException>()
+                    .Or<IOException>()
+                    .RetryAsync(3)
+                    .ExecuteAsync(() => _httpClient.PostAsync(s_url, new StringContent(request, Encoding.UTF8, "application/json"))))
+                {
+                    if (response.StatusCode == HttpStatusCode.Unauthorized || response.StatusCode == HttpStatusCode.Forbidden)
                     {
-                        _rateLimitError = Errors.GitHubApiFailed(rateLimitError.message);
+                        Log.Write(await response.Content.ReadAsStringAsync());
+                        _fatalError = Errors.GitHubApiFailed(response.StatusCode.ToString());
```
OsmondJiang: Is this message clear enough to expose to end user?

src/docfx/lib/github/GitHubAccessor.cs
yufeih: I moved them from GitHubUserCache to here because the log at GitHubUserCache does not always end up in a call to GitHub due to the return in line 46. I want this log to reflect the actual calls we made to GitHub.

test/docfx.Test/lib/GitHubAccessorTest.cs
```diff
@@ -33,23 +33,21 @@ public async Task GetUserByLogin(string login, int? expectedId, string expectedN
...
 
             // skip check if the machine exceeds the GitHub API rate limit
-            if (!string.IsNullOrEmpty(_token))
+            if (!string.IsNullOrEmpty(s_token))
             {
-                if (hasError)
-                    Assert.NotNull(error?.Message);
-                else
-                    Assert.Null(error?.Message);
+                Assert.Null(error);
```
OsmondJiang: why `repo not found` doesn't return error here? is that a break change?

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -152,6 +103,7 @@ public async Task<(Error, IEnumerable<GitHubUser>)> GetUsersByCommit(string owne
               user {
                 databaseId
                 name
+                email
```
OsmondJiang: do we cover this change in our test?

src/docfx/lib/github/GitHubAccessor.cs
yufeih: yep

src/docfx/lib/github/GitHubAccessor.cs
yufeih: Yes, there are more descriptions in the error message template.

src/docfx/lib/github/GitHubAccessor.cs
yufeih: Added a test case where `author.email` is different from `author.user.email`

test/docfx.Test/lib/GitHubAccessorTest.cs
yufeih: reverted

PR #4985, opened 2019-08-13
remove bookmark for uid resoved href

OsmondJiang:
before we implement Reference SDP, we don't need generate bookmark for resolved href of UID. Current implementation is not consistent with v2 template.

fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/112063

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4985)

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -302,7 +302,7 @@ inputs:
 outputs:
   docs/a.json:
   docs/b.json: |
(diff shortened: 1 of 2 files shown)


OsmondJiang:
Closing it because it's only related to DevSandbox repo, not Samples repo, sorry about that.

PR #4986, opened 2019-08-13
use query text as display fallback for xref resolving

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/112036

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4986)

### DIFF
--- a/src/docfx/build/xref/XrefMap.cs
+++ b/src/docfx/build/xref/XrefMap.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Docs.Build
             _externalXrefMap = ExternalXrefMapLoader.Load(docset, restoreFileMap);
         }
 
-        public (Error error, string href, IXrefSpec xrefSpec) Resolve(string uid, SourceInfo<string> href, string displayPropertyName, Document relativeTo)
+        public (Error error, string href, string display, IXrefSpec xrefSpec) Resolve(string uid, SourceInfo<string> href, string displayPropertyName, string text, Document relativeTo)
         {
             if (t_recursionDetector.Value.Contains((uid, displayPropertyName, relativeTo)))
             {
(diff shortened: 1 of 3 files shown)


src/docfx/build/xref/XrefMap.cs
```diff
@@ -70,8 +70,10 @@ public XrefMapModel ToXrefMapModel(Context context)
             var displayPropertyValue = spec.GetXrefPropertyValue(displayPropertyName);
 
             // fallback order:
-            // xrefSpec.displayPropertyName -> xrefSpec.name -> uid
-            var display = !string.IsNullOrEmpty(displayPropertyValue) ? displayPropertyValue : (!string.IsNullOrEmpty(name) ? name : uid);
+            // text -> xrefSpec.displayPropertyName -> xrefSpec.name -> uid
+            var display = !string.IsNullOrEmpty(text)
```
yufeih: can the fallback be handled in `DependencyResolver` ? XrefMap only need to return `xrefSpec`

live1206: sounds good, then we do not need to pass along `text`

live1206: We can not calculate display in `DependencyResolver`, because we have `t_recursionDetector` within `XrefMap.Resolve`, and we need to load the lazy value by `var displayPropertyValue = spec.GetXrefPropertyValue(displayPropertyName);` within `XrefMap.Resolve` to be able to detect the cyclic reference.
If we do it in `DependencyResolver`, it will end up into `StackOverflow`.
We can clean up this when we remove the handling of cyclic reference.

yufeih: cool, #3916

PR #4996, opened 2019-08-15
try to handle generated html at one place

OsmondJiang:
1. Return HTMLNODE instead of STRING for markup.ToHtml
2. Re-use HTMLNODE returned from markup as much as possible
3. remove tags and iframes are processed inside ToHtml
4. bookmarks and linktypes are process in one place in BuildPage

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4996)


src/docfx/build/page/BuildPage.cs
```diff
@@ -313,6 +312,16 @@ private static (TemplateModel model, JObject metadata) CreateTemplateModel(Conte
             return (model, metadata);
         }
 
+        private static string CreateContent(Context context, HtmlNode html, Document file, HashSet<string> bookmarks = null)
+        {
+            // add bookmark validation
```
live1206: why do we hide `bookmark validation` within `CreateContent`? If we invoke this explicitly, we do not need to handle null bookmarks.

docs/specs/schema.yml
```diff
@@ -522,3 +522,31 @@ outputs:
...
+    markdown: |
+      # h1
+      <script>alert(0);</script>
+      <link rel="stylesheet" href="https://a.com/b.css" />
+      <div style="display:hidden"></div>
+      <style>h1 {color:red;}</style>
+  _themes/ContentTemplate/schemas/TestData.schema.json: |
+    {
+      "properties": {
+        "markdown": {
```
yufeih: can omit `type`, the suggested contentType should be `markdown` but somehow `Markdown` also works due to how JSON.NET handle enums.

```suggestion
        "markdown": { "contentType": "markdown" }
```

src/docfx/build/page/BuildPage.cs
```diff
@@ -313,10 +312,21 @@ private static (TemplateModel model, JObject metadata) CreateTemplateModel(Conte
...
         }
 
+        private static string CreateHtmlContent(Context context, HtmlNode html, Document file, HashSet<string> bookmarks = null)
+        {
+            // add bookmark validation
+            bookmarks = bookmarks ?? HtmlUtility.GetBookmarks(html);
+            context.BookmarkValidator.AddBookmarks(file, bookmarks);
+
+            // get conceptual from htmldom
+            return LocalizationUtility.AddLeftToRightMarker(file.Docset.Culture, HtmlUtility.AddLinkType(html, file.Docset.Locale).WriteTo());
```
yufeih: if we can make `AddLeftToRightMarker` return an `HtmlNode`, than it could be chained with `StripTags` and we don't need a `CreateHtmlContent` method

OsmondJiang: What about AddLinkType?

yufeih: yep, missed that

src/docfx/build/page/BuildPage.cs
yufeih: I also think calling `AddBookmarks` explicitly seems better than passing null here

src/docfx/build/page/BuildPage.cs
```diff
@@ -266,9 +264,11 @@ private static async Task<(List<Error> errors, JObject model)> LoadJson(Context
                 var (deserializeErrors, landingData) = JsonUtility.ToObject<LandingData>(pageModel);
                 errors.AddRange(deserializeErrors);
 
+                var htmlDom = HtmlUtility.LoadHtml(await RazorTemplate.Render(file.Mime, landingData)).StripTags().RemoveRerunCodepenIframes();
+                ValidateBookmarks(context, file, htmlDom);
```
yufeih: consider inllne

```suggestion
            context.BookmarkValidator.AddBookmarks(file, HtmlUtility.GetBookmarks(htmlDom));
```

OsmondJiang:
@fenxu  the diff related to the CordovaDocs are not expected, would you please help to take a look?

fenxu:
Because of html content change, published case will detect difference for content hash. It's expected

PR #4997, opened 2019-08-16
fetch without depth option

OsmondJiang:
when fetch with depth option, the `branch` build and `pr` build use different options(1 for branch, 9999999 for pr). This will cause switching which cause performance issue.

With this change, the branch build without CRR cache will slow.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4997)

### DIFF
--- a/src/docfx/config/AppData.cs
+++ b/src/docfx/config/AppData.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Docs.Build
         // For testing purpose
         internal static Func<string> GetCachePath;
 
-        public static string GitRoot => Path.Combine(s_root, "git2");
+        public static string GitRoot => Path.Combine(s_root, "git3");
 
         public static string DownloadsRoot => Path.Combine(s_root, "downloads2");
 
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -104,16 +104,16 @@ namespace Microsoft.Docs.Build
         /// </summary>
         public static void InitFetchCheckout(string path, string url, string committish, Config config = null)
         {
-            InitFetch(path, url, new[] { committish }, bare: false, depthOne: false, prune: true, config);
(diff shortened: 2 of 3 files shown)


OsmondJiang:
image

yufeih:
Can you test locally if this is a breaking change? I 'm not sure if you need to pass `--unshallow` to turn a repo cloned using `--depth 1` into a repo with full branch info.

If this is a breaking change, you'll need to change the directory name.

OsmondJiang:
> Can you test locally if this is a breaking change? I 'm not sure if you need to pass `--unshallow` to turn a repo cloned using `--depth 1` into a repo with full branch info. (That is why I use --depth 999999 instead of not specifying a --depth)
> 
> If this is a breaking change, you'll need to change the directory name.

Good point, I am going to test it locally. But actually it's a fetch operation, so I don't think there is a big problem.

OsmondJiang:
image


The failed test has nothing to do with this PR, @928PJY  would you please help to take a look?

yufeih:
@herohua seems PAT for https://github.com/VSC-Service-Account has expired, we need a new one

928PJY:
you have re-trigger the CI, I cannot see the build log anymore, but from the screenshot, the error message is a little bit confused, @yufeih@live.com any idea about it?

---
In reply to: [521908700](https://github.com/dotnet/docfx/pull/4997#issuecomment-521908700) [](ancestors = 521908700)

OsmondJiang:
> you have re-trigger the CI, I cannot see the build log anymore, but from the screenshot, the error message is a little bit confused, @[yufeih@live.com](mailto:yufeih@live.com) any idea about it?
> 
> In reply to: [521908700](https://github.com/dotnet/docfx/pull/4997#issuecomment-521908700) [](ancestors = 521908700)

It's caused by Razor file watching issue, let's create a bug for tracking: https://ceapex.visualstudio.com/Engineering/_workitems/edit/113897

PR #4998, opened 2019-08-16
Extract template tests to template.yml to make test clear

v-pegao:
Workitem is [here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/107430/)

|Request|Status|
|---|---|
|extract all template specific tests into a new `template.yml` for the tests to be clearer|in progress|
|based on above, check if we missed any import cases and add them|Not started|

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/4998)


mingwli:
@v-pegao please add the following too:
https://github.com/dotnet/docfx/blob/v3/docs/specs/schema.yml#L356
https://github.com/dotnet/docfx/blob/v3/docs/specs/schema.yml#L406
https://github.com/dotnet/docfx/blob/v3/docs/specs/schema.yml#L526

v-pegao:
@mingwli Updated the template.yml

yufeih:
@v-pegao @mingwli @OsmondJiang does this work includes removing TestData.schema.json?

mingwli:
@yufeih @OsmondJiang No I think this issue only address extracting the apply JS/mustache/liquid part into a single yaml file.

yufeih:
@mingwli , ok, can we have another task that git rid of TestData.schema.json?

mingwli:
> @mingwli , ok, can we have another task that git rid of TestData.schema.json?

will track in DevOps here https://ceapex.visualstudio.com/Engineering/_workitems/edit/115502

PR #5002, opened 2019-08-19
Refine Circular xref detection

live1206:
#3916 
- The content type of `displayProperty` is should be plain text with syntax `@uid?displayProperty` since the title of URL should be plain text
- We can not remove `Lazy` from xref spec property, because if multiple xref spec properties needs uid resolving, each of them would require xref map to be loaded, which is single instance for build. And it will cause 
`(ValueFactory attempted to access the Value property of this instance.)`
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5002)
- Handle circular reference caused by JsonSchemaTransformer for SDP reference

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -25,7 +25,7 @@ inputs:
     xref: 
       - 1.xrefmap.json
       - 2.xrefmap.json
-  docs/a.md: Link to @System.String
+  docs/a.md: Link to @System.String?displayProperty=fullName
   1.xrefmap.json: |
     {
       "references":[{
@@ -40,7 +40,7 @@ inputs:
     { "references":[{"uid": "a", "name": "a", "fullName": "a", "href": "https://docs.microsoft.com/en-us/a", "nameWithType": "a"}]}
 outputs:
   docs/a.json: |
-    { "conceptual": "<p>Link to <a href=\"https://docs.microsoft.com/en-us/dotnet/api/system.string\">String</a></p>\n"}
+    { "conceptual": "<p>Link to <a href=\"https://docs.microsoft.com/en-us/dotnet/api/system.string\">System.String</a></p>\n"}
 ---
 # Restore xref map from configured url and use the YAML file for resolving
 inputs:
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -554,7 +554,106 @@ outputs:
       }
     }
 ---
-# display property of markdown content type will be ignored
+# circular reference within SDP UID definition a -> a
+inputs:
+  docfx.yml:
+  docs/a.yml: |
+    #YamlMime:TestData
+    uid: a
+    xref: a
+  _themes/ContentTemplate/schemas/TestData.schema.json: |
+    {
+      "properties": {
+          "uid": {
+            "type": "string",
+            "contentType": "uid"
+          },
+          "xref": {
+              "type": "string",
+              "contentType": "Xref"
+          }
+      },
+      "xrefProperties": [
+          "xref"
+      ]
+    }
+outputs:
+  .errors.log: |
+    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/a.yml'","docs/a.yml"]
+---
+# circular reference within SDP UID definition a -> b -> a
+inputs:
+  docfx.yml:
+  docs/a.yml: |
+    #YamlMime:TestData
+    uid: a
+    xref: b
+  docs/b.yml: |
+    #YamlMime:TestData
+    uid: b
+    xref: a
+  _themes/ContentTemplate/schemas/TestData.schema.json: |
+    {
+      "properties": {
+          "uid": {
+            "type": "string",
+            "contentType": "uid"
+          },
+          "xref": {
+              "type": "string",
+              "contentType": "Xref"
+          }
+      },
+      "xrefProperties": [
+          "xref"
+      ]
+    }
+outputs:
+  .errors.log: |
+    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/a.yml' --> 'docs/b.yml'","docs/a.yml"]
+    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/b.yml' --> 'docs/a.yml'","docs/b.yml"]
+---
+# circular reference within SDP UID definition a -> b -> a -> c
+inputs:
+  docfx.yml:
+  docs/a.yml: |
+    #YamlMime:TestData
+    uid: a
+    xref: b
+    anotherXref: c
+  docs/b.yml: |
+    #YamlMime:TestData
+    uid: b
+    xref: a
+  docs/c.yml: |
+    #YamlMime:TestData
+    uid: c
+  _themes/ContentTemplate/schemas/TestData.schema.json: |
+    {
+      "properties": {
+          "uid": {
+            "type": "string",
+            "contentType": "uid"
+          },
+          "xref": {
+              "type": "string",
+              "contentType": "Xref"
+          },
+          "anotherXref": {
+              "type": "string",
+              "contentType": "Xref"
+          }
+      },
+      "xrefProperties": [
+          "xref"
+      ]
+    }
+outputs:
+  .errors.log: |
+    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/a.yml' --> 'docs/b.yml'","docs/a.yml"]
+    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/b.yml' --> 'docs/a.yml'","docs/b.yml"]
+---
+# display property of non-plain text content type will be ignored
 inputs:
   docfx.yml: |
     template: {APP_BASE_PATH}data/template
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -554,43 +554,6 @@ outputs:
       }
     }
 ---
-# SDP non-cyclic xref reference should be resolved
-# whose display properties are plain text
-inputs:
-  docfx.yml: |
-    template: {APP_BASE_PATH}data/template
-  docs/a.json: |
-    {
-      "$schema": "https://raw.githubusercontent.com/dotnet/docfx/v3/schemas/TestData.json",
-      "uid": "a",
-      "name": "name for a",
-      "inlineDescription": "description for a",
-      "summary": "Link to @b?displayProperty=inlineDescription",
-    }
-  docs/b.json: |
-    {
-      "$schema": "https://raw.githubusercontent.com/dotnet/docfx/v3/schemas/TestData.json",
-      "uid": "b",
-      "name": "name for b",
-      "inlineDescription": "description for b",
-      "summary": "Link to @a",
-    }
-outputs:
-  docs/a.json: |
-    {
-      "name": "name for a",
-      "summary": "<p>Link to <a href=\"b.json\">description for b</a></p>\n",
-      "inlineDescription": "description for a",
-      "uid": "a"
-    }
-  docs/b.json: |
-    {
-      "name": "name for b",
-      "summary": "<p>Link to <a href=\"a.json\">name for a</a></p>\n",
-      "inlineDescription": "description for b",
-      "uid": "b"
-    }
----
 # display property of markdown content type will be ignored
 inputs:
   docfx.yml: |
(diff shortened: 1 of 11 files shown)


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -115,7 +115,25 @@ internal class DependencyResolver
...
-            var (error, resolvedHref, display, xrefSpec) = _xrefMap.Value.Resolve(Uri.UnescapeDataString(uid), href, displayProperty, text, declaringFile);
+            var (error, resolvedHref, xrefSpec) = _xrefMap.Value.Resolve(Uri.UnescapeDataString(uid), href);
+
+            var name = xrefSpec?.GetXrefPropertyValue("name");
+
+            // for internal UID, the display property can not be markdown or inline markdown
+            // because the display text should be plain text
+            string displayPropertyValue = null;
+            if (xrefSpec is InternalXrefSpec internalSpec
+                && !new JsonSchemaContentType[] { JsonSchemaContentType.Markdown, JsonSchemaContentType.InlineMarkdown }
```
yufeih: avoid the array allocation here using `contentType != Markdown && contentType != InlineMarkdown`

src/docfx/build/xref/XrefMap.cs
```diff
@@ -15,34 +13,24 @@ internal class XrefMap
...
-        private static ThreadLocal<Stack<(string uid, string propertyName, Document parent)>> t_recursionDetector = new ThreadLocal<Stack<(string, string, Document)>>(() => new Stack<(string, string, Document)>());
-
         public XrefMap(Context context, Docset docset, RestoreFileMap restoreFileMap)
         {
             _internalXrefMap = InternalXrefMapBuilder.Build(context);
             _externalXrefMap = ExternalXrefMapLoader.Load(docset, restoreFileMap);
         }
 
-        public (Error error, string href, string display, IXrefSpec xrefSpec) Resolve(string uid, SourceInfo<string> href, string displayPropertyName, string text, Document relativeTo)
+        public (Error error, string href, IXrefSpec xrefSpec) Resolve(string uid, SourceInfo<string> href)
```
yufeih: can JsonSchemaTransformer.cs line 236 be updated to use this method instead because it does not need a link?

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -115,7 +115,25 @@ internal class DependencyResolver
...
+                .Contains(internalSpec.GetXrefPropertyContentType(displayProperty)))
+            {
+                displayPropertyValue = xrefSpec.GetXrefPropertyValue(displayProperty);
+            }
+
+            // fallback order:
+            // text -> xrefSpec.displayPropertyName -> xrefSpec.name -> uid
```
yufeih: any property in this fallback chain could in theory have `contentType == markdown`, so implement the filter in `GetXrefPropertyValue`? consider name `GetXrefPropertyAsString` to indicate there is a string conversion and properties might be dropped if the not a literal.

`GetXrefPropertyAsString` could return null for empty strings so we can use `??` here for fallback

src/docfx/build/xref/XrefMap.cs
live1206: Not really, because `XrefMap.Resolve` takes `uid` as input, which `JsonSchemaTransformer` only has `href`. And also `DependencyResolver` seems to be a better entry point for resolving.

yufeih: the content there MUST be an `uid`, not an `href`. The `xref` content type only cares about retrieving an `XrefSpec`, not building up an URL or display name

live1206: ok, if content can only be `uid` there

live1206: hmm... We also handled the dependency map within `DependencyResolver.ResolveAbsoluteXref`, so it is a better entry point.

src/docfx/build/xref/XrefMap.cs
```diff
@@ -3,46 +3,24 @@
...
         public XrefMap(Context context, Docset docset, RestoreFileMap restoreFileMap)
         {
             _internalXrefMap = InternalXrefMapBuilder.Build(context);
             _externalXrefMap = ExternalXrefMapLoader.Load(docset, restoreFileMap);
         }
 
-        public (Error error, string href, string display, IXrefSpec xrefSpec) Resolve(string uid, SourceInfo<string> href, string displayPropertyName, string text, Document relativeTo)
+        public IXrefSpec Resolve(SourceInfo<string> uid)
         {
-            if (t_recursionDetector.Value.Contains((uid, displayPropertyName, relativeTo)))
```
OsmondJiang: >             [](start = 0, length = 12)

It can be removed totally as we talked offline, but we can simplify the stack, only keep the uid

src/docfx/build/xref/InternalXrefSpec.cs
```diff
@@ -21,52 +27,84 @@ internal class InternalXrefSpec : IXrefSpec
...
-                catch (DocfxException ex)
+
+                foreach (var (key, value) in ExtensionData)
                 {
-                    context.ErrorLog.Write(DeclaringFile, ex.Error);
+                    try
+                    {
+                        spec.ExtensionData[key] = value.Value;
+                    }
+                    catch (DocfxException ex)
```
yufeih: could this catch be the cause of the problem?

live1206: This is used to add `DeclaringFile` for the error, because while outputting xref map, there is no file attached to the error

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -233,11 +239,28 @@ void TraverseObjectXref(JObject obj, Func<JsonSchema, string, JToken, bool> acti
...
+                    var (xrefError, _, _, xrefSpec) = context.DependencyResolver.ResolveAbsoluteXref(content, file);
 
                     if (xrefSpec is InternalXrefSpec internalSpec)
                     {
-                        xrefSpec = internalSpec.ToExternalXrefSpec(context, forXrefMapOutput: false);
+                        if (t_recursionDetector.Value.Contains((content, file)))
+                        {
+                            var referenceMap = t_recursionDetector.Value.Select(x => x.declaringFile).ToList();
+                            referenceMap.Insert(0, file);
+                            throw Errors.CircularReference(referenceMap).ToException();
```
yufeih: move to JsonSchemaTransformer line 93

```suggestion
                            throw Errors.CircularReference(referenceMap, file).ToException();
```

src/docfx/build/xref/InternalXrefSpec.cs
```diff
@@ -61,12 +69,22 @@ public ExternalXrefSpec ToExternalXrefSpec(Context context, bool forXrefMapOutpu
                 {
                     spec.ExtensionData[key] = value.Value;
                 }
-                catch (DocfxException ex)
+
+                // circular-reference is handled in JsonSchemaTransformer
+                // no need to throw it again for xref map output
+                catch (DocfxException ex) when (forXrefMapOutput && ex.Error.Code == "circular-reference")
```
yufeih:
```suggestion
                catch (Exception ex) when (DocfxException.Is(xxx))
                {
                    context.ErrorLog.Write(DeclaringFile, ex.Error);
                }
```

live1206: Once there is always a file in the error, we do not need to rewrite it again here.

src/docfx/Errors.cs
```diff
@@ -73,8 +73,8 @@ public static Error ConfigNotFound(string docsetPath)
         ///   - a.md references b.json's property with xref syntax, and b.json includes a.md reversely
         /// </summary>
         /// Behavior: ✔️ Message: ✔️
-        public static Error CircularReference<T>(IEnumerable<T> dependencyChain)
-            => new Error(ErrorLevel.Error, "circular-reference", $"Build has identified file(s) referencing each other: {string.Join(" --> ", dependencyChain.Select(file => $"'{file}'"))}");
+        public static Error CircularReference<T>(IEnumerable<T> dependencyChain, Document declaringFile)
+            => new Error(ErrorLevel.Error, "circular-reference", $"Build has identified file(s) referencing each other: {string.Join(" --> ", dependencyChain.Select(file => $"'{file}'"))}", file: declaringFile.FilePath);
```
```
{string.Join(" --> ", dependencyChain.Select(file => $"'{file}'")) + $"--> {declaringFile}"}"
```

live1206: This would not work, the dependencyChain should be handled by the caller. We have usage in TOC and Xref, they are constructed in different ways

yufeih: we should be using the same way to detect circular reference using a thread local stack

live1206: updated

docs/specs/xref.yml
```diff
@@ -554,104 +554,111 @@ outputs:
...
+    xref: a
+  _themes/ContentTemplate/schemas/TestData.schema.json: |
     {
-      "$schema": "https://raw.githubusercontent.com/dotnet/docfx/v3/schemas/TestData.json",
-      "uid": "b",
-      "name": "name for b",
-      "description": "description for b",
-      "summary": "Link to @a?displayProperty=summary",
+      "properties": {
+          "uid": {
```
yufeih:
```suggestion
          "uid": { "contentType": "uid" }
```

docs/specs/xref.yml
```diff
@@ -554,104 +554,111 @@ outputs:
...
+          "xref"
+      ]
     }
 outputs:
-  docs/c.json: |
-    {"name":"name for c","summary":"<p>Link to <a href=\"a.json\">name for a</a></p>\n","inlineDescription":"description for c","uid":"c"}
   .errors.log: |
-    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/a.json' --> 'docs/b.json' --> 'docs/a.json'","docs/a.json"]
-    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/b.json' --> 'docs/a.json' --> 'docs/b.json'","docs/b.json"]
+    ["error","circular-reference","Build has identified file(s) referencing each other: 'docs/a.yml' --> 'docs/b.yml' --> 'docs/a.yml'","docs/a.yml"]
```
yufeih: consider use `$"{uid} ({file}"` because there might be multiple uids in the same file
```suggestion
    ["error","circular-reference","Build has identified file(s) referencing each other: 'a (docs/a.yml)' --> 'docs/b.yml' --> 'docs/a.yml'","docs/a.yml"]
```

docs/specs/xref.yml
live1206: think it is better to define string type here as well

docs/specs/xref.yml
live1206: sounds reasonable

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -88,25 +90,31 @@ private static (List<Error>, TableOfContentsModel) LoadTocModel(JToken tocToken)
             Document rootPath,
             List<Document> referencedFiles,
             List<Document> referencedTocs,
-            List<Document> parents,
             string content = null)
         {
             // add to parent path
-            if (parents.Contains(file))
+            if (t_recursionDetector.Value.Contains(file))
```
OsmondJiang: why do we change toc parts?

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -233,11 +254,12 @@ void TraverseObjectXref(JObject obj, Func<JsonSchema, string, JToken, bool> acti
                     break;
 
                 case JsonSchemaContentType.Xref:
-                    var (xrefError, xrefLink, _, xrefSpec) = context.DependencyResolver.ResolveAbsoluteXref(content, file);
+                    // TODO: the content here must be an UID, not href
+                    var (xrefError, _, _, xrefSpec) = context.DependencyResolver.ResolveAbsoluteXref(content, file);
```
OsmondJiang: There is no reason to expose internal Xref Spec here, we can directly resolve External Xref Spec. And then hide a lot of internal Xref Spec logic like when to get the lazy value.

live1206: Will refine this further in the next PR

src/docfx/build/toc/parser/TableOfContentsParser.cs
live1206: Trying to use the same way to detect recursion

PR #5003, opened 2019-08-19
Resolve dependency using entry config

yufeih:
Fixes https://dev.azure.com/ceapex/Engineering/_workitems/edit/113864

The root cause is we use `Docset.DependencyDocsets` to lookup dependencies, which is empty when `Docset` itself is a dependency docset. This PR moves `Docset.DependencyDocsets` to dependency resolver and uses the same `_dependencies` dictionary to resolve everything.

Due to this change, we always use entry docset config for everything including dependencies, so some restore tests are marked as skip because they specify a config in dependency docset which is currently a not supported scenario in the docs world. We'll revisit these scenarios when doing `dotnet-docs` which uses _content from CRR_.

Dependency resolve and fallback resolve is consolidated into the same place at `DependencyResolver.TryResolveFile` to clarify the logic.

Simplified unnecessary Docset constructor parameters.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5003)

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -20,7 +20,7 @@ outputs:
   docs/a.json: |
     {"conceptual":"<p><a href=\"%7E/docfx-test-dependencies1/docs/a.md\">link to 1</a></p>\n<p>a1</p>\n"}
 ---
-# [Skip]Restore dependencies azure repos with their children
+# Restore dependencies azure repos with their children
 repos:
   https://docascode.visualstudio.com/project/_git/a:
     - files:
@@ -118,6 +118,7 @@ inputs:
 outputs:
   .errors.log: |
     ["error","committish-not-found","Can't find branch, tag, or commit 'bad-branch' for repo https://github.com/docascode/docfx-test-dependencies."]
+    ["error","need-restore","Cannot find dependency 'https://github.com/docascode/docfx-test-dependencies#bad-branch', did you forget to run `docfx restore`?"]
 ---
 # Show error when dependent repo can't be found
 restore: false
@@ -129,7 +130,7 @@ inputs:
     [test](dep/abc)
 outputs:
   .errors.log: |
-    ["error","need-restore","Cannot find dependency 'https://github.com/docascode/docfx-test-dependencies#bad-branch', did you forget to run `docfx restore`?","docs/a.md"]
+    ["error","need-restore","Cannot find dependency 'https://github.com/docascode/docfx-test-dependencies#bad-branch', did you forget to run `docfx restore`?"]
 ---
 # Show error when dependent URL can't be found
 restore: false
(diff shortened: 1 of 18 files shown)


OsmondJiang:
Can we just remove the error code "Link Is Dependency" and use  LinkOutOfScope only? Then we can easily fix this problem at:

image

yufeih:
For resources, we currently build resources from main repo but is outside build scope, and so we don't build resources from dependency repo (outside build scope), so need a flag to tell if a file is from a dependency repo or not.

OsmondJiang:
> For resources, we currently build resources from main repo but is outside build scope, and so we don't build resources from dependency repo (outside build scope), so need a flag to tell if a file is from a dependency repo or not.

Then we can just use the _buildScope to tell us if it's dependency docset or not. we don't need change so much ....

OsmondJiang:
As we talked offline, use the same _dependencyDocsets to resolve everything is not correct, the _dependencyDocsets need follows the Docset, no matter it's entry docset or dependency docset.

docs/specs/restore.yml
```diff
@@ -184,7 +185,7 @@ repos:
 outputs:
   docs/a.json:
 ---
-# restore with dependency lock
+# [skip] restore with dependency lock
```
OsmondJiang: still skipped?

src/docfx/build/context/BuildScope.cs
```diff
@@ -27,19 +25,23 @@ internal class BuildScope
 
         public RedirectionMap Redirections { get; }
 
+        public Docset Docset { get; }
+
+        public Docset FallbackDocset { get; }
```
OsmondJiang: I try to hide FallbackDocset from outside of build scope, can we still keep it hidden?

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -280,38 +283,83 @@ private string ResolveToDocsetRelativePath(string path, Document referencingFile
             return docsetRelativePath;
         }
 
-        private Document TryResolveResourceFromHistory(
-            Document referencingFile, GitCommitProvider gitCommitProvider, string pathToDocset, TemplateEngine templateEngine)
+        private Document TryResolveFile(Document declaringFile, string pathToDocset)
```
```suggestion
        private Document TryResolveFile(Document referencingFile, string pathToDocset)
```

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -18,15 +18,17 @@ internal class DependencyResolver
         private readonly DependencyMapBuilder _dependencyMapBuilder;
         private readonly GitCommitProvider _gitCommitProvider;
         private readonly IReadOnlyDictionary<string, string> _resolveAlias;
+        private readonly IReadOnlyDictionary<string, Docset> _dependencies;
```
OsmondJiang: >_dependencies [](start = 61, length = 13)

Is it better to put it in BuildScope?

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -280,38 +283,83 @@ private string ResolveToDocsetRelativePath(string path, Document referencingFile
             return docsetRelativePath;
         }
 
-        private Document TryResolveResourceFromHistory(
-            Document referencingFile, GitCommitProvider gitCommitProvider, string pathToDocset, TemplateEngine templateEngine)
+        private Document TryResolveFile(Document referencingFile, string pathToDocset)
```
OsmondJiang: >Document [](start = 16, length = 8)

Also for this one

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -196,7 +197,7 @@ string GetContentGitUrl(string branchUrlTemplate)
                     (branchUrlTemplate, _) = GetContentGitUrlTemplate(contributionRemote, pathToRepo);
 
                     (editRemote, editBranch) = (contributionRemote, hasRefSpec ? contributionBranch : editBranch);
-                    if (context.BuildScope.GetFallbackDocset(document.Docset) != null)
```
OsmondJiang: >context [](start = 24, length = 7)

FilePath.From

PR #5005, opened 2019-08-21
 report a error when the dest to redirection url does not match any files's publish URL, but the redirect_with_id flag has been set as true

928PJY:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/112321

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5005)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Docs.Build
         /// The dest to redirection url does not match any files's publish URL, but the redirect_with_id flag has been set as true
         /// </summary>
         public static Error RedirectionUrlNotExisted(SourceInfo<string> source)
-            => new Error(ErrorLevel.Error, "redirection-url-not-existed", $"The redirect url '{source}' does not match any file's publish URL, but the redirect_with_id flag has been set as true", source);
+            => new Error(ErrorLevel.Warning, "redirection-url-not-existed", $"The redirect url '{source}' does not match any file's publish URL, but the redirect_with_id flag has been set as true", source);
 
         /// <summary>
         /// Used invalid glob pattern in configuration.
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -44,6 +44,12 @@ namespace Microsoft.Docs.Build
             => new Error(ErrorLevel.Warning, "redirection-url-conflict", $"The '{source}' appears twice or more in the redirection mappings", source);
 
         /// <summary>
+        /// The dest to redirection url does not match any files's publish URL, but the redirect_with_id flag has been set as true
+        /// </summary>
+        public static Error RedirectionUrlNotExisted(SourceInfo<string> source)
+            => new Error(ErrorLevel.Error, "redirection-url-not-existed", $"The redirect url '{source}' does not match any file's publish URL, but the redirect_with_id flag has been set as true");
+
+        /// <summary>
         /// Used invalid glob pattern in configuration.
         /// Examples:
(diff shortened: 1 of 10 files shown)


src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using Microsoft.Graph;
```
live1206: remove

src/docfx/Errors.cs
```diff
@@ -43,6 +43,12 @@ public static Error RedirectionUrlInvalid(SourceInfo<string> source)
         public static Error RedirectionUrlConflict(SourceInfo<string> source)
             => new Error(ErrorLevel.Warning, "redirection-url-conflict", $"The '{source}' appears twice or more in the redirection mappings", source);
 
+        /// <summary>
+        /// The dest to redirection url does not match any files's publish URL, but the redirect_with_id flag has been set as true
+        /// </summary>
+        public static Error RedirectionUrlNotExisted(SourceInfo<string> source)
+            => new Error(ErrorLevel.Error, "redirection-url-not-existed", $"The redirect url '{source}' does not match any file's publish URL, but the redirect_with_id flag has been set as true", source);
```
live1206: it says report `warning` in https://dev.azure.com/ceapex/Engineering/_workitems/edit/112321, upgraded to `error`?

src/docfx/build/redirection/RedirectionMap.cs
yufeih: ~~using Microsoft.Graph;~~

src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -54,12 +56,11 @@ public static RedirectionMap Create(ErrorLog errorLog, Docset docset, Func<strin
 
             var redirectionsBySourcePath = redirections.ToDictionary(file => file.FilePath.Path, PathUtility.PathComparer);
 
+            CheckInvalidRedrectUrl();
             return new RedirectionMap(redirectionsBySourcePath, redirectionsByRedirectionUrl);
 
             void AddRedirections(Dictionary<string, SourceInfo<string>> items, bool redirectDocumentId = false)
```
yufeih: nit: consider turn this into a member function, this method is getting big enough

src/docfx/build/redirection/RedirectionMap.cs
```diff
@@ -111,6 +113,24 @@ void AddRedirections(Dictionary<string, SourceInfo<string>> items, bool redirect
                     }
                 }
             }
+
+            void CheckInvalidRedrectUrl()
+            {
+                var redirectUrls = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+
+                var publishUrls = buildFiles.Select(file => file.SiteUrl).Union(redirections.Select(redirection => redirection.SiteUrl)).ToHashSet();
```
yufeih: Union creates a hashset internally, there is already a ToHashSet at the end, so Concat is faster in this case
```suggestion
                var publishUrls = buildFiles.Select(file => file.SiteUrl).Concat(redirections.Select(redirection => redirection.SiteUrl)).ToHashSet();
```

src/docfx/Errors.cs
928PJY: Yes, since it should break the build of this file, so upgraded to `error`

PR #5008, opened 2019-08-21
Fix code style violation of rule MEN008 and MEN012

v-pegao:
|Rule|Fix|
|---|---|
|MEN008: File name should match type|Use type name for file name|
|MEN012: Flags should be powers of two|Use C# binary literal|

Workitem is [here](https://dev.azure.com/ceapex/Engineering/_workitems/edit/107202?src=WorkItemMention&src-action=artifact_link)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5008)


yufeih:
@v-pegao need to resolve conflict

PR #5013, opened 2019-08-22
Refine xref

live1206:
- For error `xref-not-found`, it should report only `UID` not `href`
- For `DependencyResolver.ResolveXref`, it should return `resolvedLink` and `declaraingFile`, not entire `xrefSpec`
- `XrefMap.Resolve` will return `xrefSpec` with lazy properties, the consumer needs to invoke`ToExternalXrefSpec` if want them to be loaded since the loading is not required for `DependencyResolver` usage
- `XrefMap.Resolve` will add dependencies during resolving
- Add `ToExternalXrefSpec` to `IXrefSpec` to avoid check of `InternalXrefSpec`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5013)


src/docfx/build/xref/XrefMap.cs
```diff
@@ -11,22 +11,41 @@ internal class XrefMap
...
-        public XrefMap(Context context, Docset docset, RestoreFileMap restoreFileMap)
+        public XrefMap(Context context, Docset docset, RestoreFileMap restoreFileMap, DependencyMapBuilder dependencyMapBuilder)
         {
             _internalXrefMap = InternalXrefMapBuilder.Build(context);
             _externalXrefMap = ExternalXrefMapLoader.Load(docset, restoreFileMap);
+            _dependencyMapBuilder = dependencyMapBuilder;
         }
 
-        public IXrefSpec Resolve(SourceInfo<string> uid)
+        public (Error, IXrefSpec) Resolve(SourceInfo<string> uid, Document referenceFile)
```
OsmondJiang: How about Resolve<T> where T : IXrefSpec?

live1206: Do not quite follow here, T is the return type?

OsmondJiang: edited, the html tag is not showed :(

live1206: I guess `T` would be used for the return type? What is the difference?

live1206: The main problem here is, we need to provide two types of IXrefSpec, one for `DependencyResolver.ResolveXref` which only needs lazy properties, another one for `JsonSchemaTransformer` which needs properties loaded.

OsmondJiang: What I means is to totally hide Internal Xref since you don't want others to directly use it expected DependencyResolver.

src/docfx/build/xref/IXrefSpec.cs
```diff
@@ -14,5 +14,7 @@ internal interface IXrefSpec
         HashSet<string> Monikers { get; }
 
         string GetXrefPropertyValueAsString(string propertyName);
+
+        ExternalXrefSpec ToExternalXrefSpec();
```
OsmondJiang: suggest to remove

live1206: with this, we do not need to check if `InternalXrefSpec`

PR #5015, opened 2019-08-22
 Model dependency as PackageUrl

yufeih:
Now every place that process dependencies knows what a dependency is not a simple string but a package that can be of various forms.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5015)


src/docfx/lib/path/PackageUrl.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    /// <summary>
+    /// Define the location of a package.
+    /// A package is any of the following:
+    ///   a) A folder
+    ///   b) A git url in the form of `{remote_url}#{commit-ish}`, when cloned, results in a)
+    /// The commit-sh can be any tag, sha, or branch. The default commit-ish is master.
+    /// </summary>
+    [JsonConverter(typeof(ShortHandConverter))]
+    internal readonly struct PackageUrl
```
OsmondJiang: PackageUrl is too general, especially we also have files to download？

OsmondJiang: since it's related to git, how about GitPackageUrl?

yufeih: A package is just a folder of files, git is one way to retrieve these files, we may in the future support other types of packages like zip files.

The idea came from https://docs.npmjs.com/about-packages-and-modules#about-package-formats

OsmondJiang: But this PackageUrl has "Remote" and "Branch"? Is that common for all kinds of resource we want to restore?

yufeih: This is a union type, Remote and Branch only has value if the type is Git, Path only has value if the type is Folder, C# currently don't have a good way to express union types

yufeih: https://github.com/dotnet/csharplang/issues/113 for a detailed discussion around discriminated unions

PR #5018, opened 2019-08-23
Implement require function using GlobalObject

yufeih:
so we don't build a custom script based on input script, this gives us the correct line, column info.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5018)


yufeih:
This change does not confirm to module spec...

PR #5020, opened 2019-08-23
Fix code style violation of rule MEN002

v-pegao:
Workitem is [here](https://dev.azure.com/ceapex/Engineering/_workitems/edit/107202?src=WorkItemMention&src-action=artifact_link)

|Rule|Fix|
|---|---|
|MEN002: Lines too long<br/>(current line limit is 140 characters)|- If it is caused by too many nested brackets, try  turn local functions into member functions or use early return<br/>- If it is long parameter, try put parameters to a new line, if it is still too long, break each parameter into its own line<br/>- If parameter is a long expression, try extract a variable<br/>- For long member variable declarations with initialization, try put the default initializer to a new line<br/>- If the return value is a tuple ant the tuple definition is long, try put the return value definition in its own line<br/>- Suppress this warning in Errors.cs|

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5020)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 
 namespace Microsoft.Docs.Build
 {
+    [SuppressMessage("Layout", "MEN002", Justification = "Suppress MEN002 for Errors.cs")]
     internal static class Errors
     {
         /// <summary>
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -30,39 +30,47 @@ namespace Microsoft.Docs.Build
                     return;
 
                 var gitMap = fallbackRestoreGitMap ?? restoreGitMap;
-                var (docset, fallbackDocset) = GetDocsetWithFallback();
+                var (docset, fallbackDocset) = GetDocsetWithFallback(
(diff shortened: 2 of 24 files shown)


src/docfx/build/Build.cs
```diff
@@ -38,17 +38,25 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Erro
                 (Docset docset, Docset fallbackDocset) GetDocsetWithFallback()
```
yufeih: try turn inner function into member function first

v-pegao: Turned inner function `GetDocsetWithFallback()` and `ShouldBuildFile()` to member function, and resolved the line too long issue.

src/docfx/Errors.cs
```diff
@@ -3,10 +3,12 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 
 namespace Microsoft.Docs.Build
 {
+    [SuppressMessage("Layout", "MEN002", Justification = "Suppress MEN002 for Errors.cs")]
```
yufeih: 👍 want to help replace places where we use `#pragma warning disable` with Attribute?

v-pegao: Sure. Will update in other PR later.

PR #5026, opened 2019-08-26
Disable razor file watcher

yufeih:
Fixes random UT failure

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5026)

### DIFF
--- a/src/docfx/template/RazorTemplate.cs
+++ b/src/docfx/template/RazorTemplate.cs
@@ -56,6 +56,7 @@ namespace Microsoft.Docs.Build
             void ConfigureServices(IServiceCollection services)
             {
                 services.AddMvc()
+                        .AddRazorOptions(options => options.AllowRecompilingViewsOnFileChange = false)
                         .ConfigureApplicationPartManager(parts =>
                         {
                             // Ensure we only have one private TemplateController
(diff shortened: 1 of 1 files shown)


src/docfx/template/RazorTemplate.cs
```diff
@@ -56,6 +56,7 @@ private static TestServer StartServer()
             void ConfigureServices(IServiceCollection services)
             {
                 services.AddMvc()
+                        .AddRazorOptions(options => options.AllowRecompilingViewsOnFileChange = false)
```
live1206: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.razor.razorviewengineoptions.allowrecompilingviewsonfilechange?view=aspnetcore-2.2
The default value is true if the version is Version_2_1 or earlier. If the version is later and EnvironmentName is Development, the default value is true. Otherwise, the default value is false.

I saw our version is 2.2. Is `EnvironmentName` set to `Development` during the CI run?

yufeih: There is no default value set for `EnvironmentName` and thus ASPNET core assume it is development, setting `EnvironmentName` to production sounds like a better option

PR #5027, opened 2019-08-26
Optimize GetCommitHistory

yufeih:
Optimize `GetCommitHistory`:

- Eliminate `LoadCommit` call when commit cache present.
- Pool objects to avoid excessive memory usage when commit cache missing

legend  | peak memory usage without cache    | build time without cache | build time with cache | cache size
---|---|---|---|---
Before | 7.8G | 5:33 | 1:13 | 3MB
After | 7G | 5:50 | 1:03 | 6MB

https://dev.azure.com/ceapex/Engineering/_workitems/edit/113226/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5027)

### DIFF
--- a/src/docfx/lib/git/FileCommitProvider.cs
+++ b/src/docfx/lib/git/FileCommitProvider.cs
@@ -17,15 +17,16 @@ namespace Microsoft.Docs.Build
         private readonly string _repoPath;
         private readonly Lazy<GitCommitCache> _commitCache;
 
+        // Commit history and a lookup table from commit hash to commit.	
+        // Use `long` to represent SHA2 git hashes for more efficient lookup and smaller size.	
+        private readonly ConcurrentDictionary<long, NativeGitCommit> _commits = new ConcurrentDictionary<long, NativeGitCommit>();
+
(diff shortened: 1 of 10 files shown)


src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -61,63 +60,91 @@ public List<GitCommit> GetCommitHistory(string file, string committish = null)
...
+
+        private unsafe GitCommit[] GetCommitHistory(GitCommitCache.FileCommitCache commitCache, string file, string committish = null)
+        {
+            if (git_revparse_single(out var pHead, _repo, committish ?? "HEAD") != 0)
             {
-                return new List<GitCommit>();
+                throw Errors.CommittishNotFound(_repoPath, committish).ToException();
             }
 
+            var (commits, parents, closeNodes, openNodes) = _objectPool.TryTake(out var pooled)
```
live1206: if we just want to reduce the allocation, can we initialize `commits`, `closeNodes`, `openNodes` as private field?

PR #5028, opened 2019-08-26
Fix a ChakraCore source context thread bug

yufeih:
`JavaScriptSourceContext++` is not thread safe, this can lead to random crashes in very rare case.

https://dev.azure.com/ceapex/Engineering/_build/results?buildId=79335&view=logs&jobId=133bd042-0fac-58b5-e6e7-01018e6dc4d4

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5028)


yufeih:
@928PJY , after redirection warning is enabled, these repos have reached max warning count, thus there is diff for every PR.

PR #5030, opened 2019-08-26
Directly resolve link/content from dependency git without checking out

OsmondJiang:
1. Remove dependency slot :)
2. Remove worktree management for Dependencies
3. No need checkout worktrees for CRRs
4. Only one worktree for Template/Localization dependencies
5. Share Lock will be applied against dependency "remote" during build
6. Exclusive Lock will be applied against dependency "remote" during restore
7. GetGitRestorePath becomes really simple
8. GetDocumentFromGit perf is good.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5030)


OsmondJiang:
The result is good, without CRR checkout, the build time is still keeping aroun 1 minutes on my machine:

image


We will do more in this PR, to remove CRR restore checking out steps, then we can reduce a lot of code complexities and save times.

docs/specs/localization.yml
```diff
@@ -6,8 +6,8 @@ repos:
     - files:
         docfx.yml: |
           dependencies:
-            restore-child: https://docs.com/restore1#child
-  https://docs.com/restore1#child:
+            restore-child: https://docs.com/restore-child1#child
```
OsmondJiang: one repo can be referenced more than one times by different repos?

src/docfx/docset/Document.cs
```diff
@@ -141,34 +143,31 @@ internal class Document : IEquatable<Document>, IComparable<Document>
...
             Debug.Assert(IsValidRelativePath(SitePath));
 
             Debug.Assert(SiteUrl.StartsWith('/'));
             Debug.Assert(!SiteUrl.EndsWith('/') || Path.GetFileNameWithoutExtension(SitePath) == "index");
         }
 
-        /// <summary>
-        /// Reads this document as stream, throws if it does not exists.
-        /// </summary>
-        public Stream ReadStream()
```
OsmondJiang: can't be removed, used for reading YAML header

src/docfx/lib/git/LibGit2.cs
```diff
@@ -84,6 +84,9 @@ static LibGit2()
         [DllImport(LibName)]
         public static unsafe extern int git_revparse_single(out IntPtr @out, IntPtr repo, [MarshalAs(UnmanagedType.LPUTF8Str)]string spec);
 
+        [DllImport(LibName)]
+        public static unsafe extern int git_revparse_ext(out IntPtr @object, out IntPtr @reference, IntPtr repo, [MarshalAs(UnmanagedType.LPUTF8Str)]string spec);
```
OsmondJiang: remove

src/docfx/localization/LocalizationUtility.cs
```diff
@@ -48,14 +48,15 @@ public static (string remote, string branch) GetLocalizedRepo(LocalizationMappin
             return (newRemote, newBranch);
         }
 
-        public static bool TryGetLocalizedDocsetPath(Docset docset, RestoreGitMap restoreMap, Config config, string locale, out string localizationDocsetPath, out string localizationBranch)
+        public static bool TryGetLocalizedDocsetPath(DependencyGitLock gitLock, Docset docset, Config config, string locale, out string localizationDocsetPath, out string localizationBranch, out string localizationCommit)
```
OsmondJiang: TryGetLocalizationRepository

src/docfx/restore/RestoreGitMap.cs
```diff
@@ -5,171 +5,79 @@
...
-
-            Debug.Assert(released);
         }
 
-        /// <summary>
-        /// Acquired all shared git based on dependency lock
-        /// The dependency lock must be loaded before using this method
-        /// </summary>
-        public static RestoreGitMap Create(DependencyLockModel dependencyLock)
+        public static (string path, string commit) GetRestoredRepository(DependencyGitLock gitLock, string remote, string branch, string docsetPath, bool bare)
```
OsmondJiang: Repository?

docs/specs/localization.yml
OsmondJiang: I don't think it's a valid scenario

src/docfx/build/context/Cache.cs
```diff
@@ -41,6 +44,9 @@ internal class Cache
                 new Lazy<(List<Error>, TableOfContentsModel, List<Document>, List<Document>)>(
                     () => TableOfContentsParser.Load(context, file))).Value;
 
+        public Document GetDependentDocument(Docset docset, string pathToDocset, Func<Document> getDocumentFromGit)
```
live1206: Since `getDocumentFromGit` is expensive, otherwise we wouldn't cache here. Is it possible multiple threads trying to create the same key? Or to say, do we need `ConcurrentDictionary<(Docset docset, string pathToDocset), Lazy<Document>>`?

src/docfx/docset/Document.cs
```diff
@@ -141,21 +144,28 @@ internal class Document : IEquatable<Document>, IComparable<Document>
 
             _id = new Lazy<(string docId, string versionId)>(() => LoadDocumentId());
             _repository = new Lazy<Repository>(() => Docset.GetRepository(FilePath.Path));
+            _content = content;
 
+            Debug.Assert(!isFromHistory || _content != null);
```
live1206: seems like the only case `isFromHistory` as `true` is get content from CRR git history? can we remove `IsFromHistory`, only keep `_content`?

src/docfx/lib/process/SharedAndExclusiveLock.cs
```diff
@@ -0,0 +1,54 @@
...
+namespace Microsoft.Docs.Build
+{
+    public struct SharedAndExclusiveLock : IDisposable
+    {
+        private static readonly TimeSpan _defaultTimeout = TimeSpan.FromSeconds(30);
+        private readonly bool _shared;
+        private readonly string _acquirer;
+        private readonly string _lockName;
+
+        public SharedAndExclusiveLock(string lockName, bool shared, TimeSpan? timeout = null)
```
live1206: remove `timeout` input? seems it is unused

yufeih:
Would it be simpler if there is only one global lock that is shared for `build` and exclusive for `restore`. It solves the bug of conflicts when we restore and build the same file.

src/docfx/build/Build.cs
```diff
@@ -73,12 +76,12 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Erro
...
             Docset fallbackDocset,
-            RestoreGitMap restoreGitMap,
             CommandLineOptions options,
             ErrorLog errorLog,
-            string outputPath)
+            string outputPath,
+            GitLock gitLock)
         {
-            using (var context = new Context(outputPath, errorLog, docset, fallbackDocset, restoreGitMap))
+            using (var context = new Context(outputPath, errorLog, docset, fallbackDocset, gitLock))
```
yufeih: Changing from RestoreMap to GitLock is a leak of abstraction. GitLock is an internal data structure that should only be known by the restore component. Other parts of the system query restore formation through the `RestoreMap` interface.

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -318,7 +305,55 @@ private Document TryResolveFile(Document referencingFile, string pathToDocset)
...
+            }
+
+            if (File.Exists(Path.Combine(docset.DocsetPath, filePath.Path)))
+            {
+                return Document.Create(docset, filePath, _templateEngine);
+            }
+
+            var (repo, pathToRepo, commits) = _gitCommitProvider.GetCommitHistory(docset, filePath.Path, docset.Repository.Commit);
+
+            var commit = deleted && commits.Length > 1 ? commits[1] : (!deleted && commits.Length > 0 ? commits[0] : default);
```
yufeih: GetCommitHistory cannot tell you if a file exist in the HEAD commit, this can potentially resolve a deleted dependency file

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -295,14 +285,11 @@ private Document TryResolveFile(Document referencingFile, string pathToDocset)
...
-                var pathToDependencyDocset = pathToDocset.Substring(dependencyName.Length);
-                if (File.Exists(Path.Combine(dependencyDocset.DocsetPath, pathToDependencyDocset)))
-                {
-                    return Document.Create(dependencyDocset, new FilePath(pathToDependencyDocset, dependencyName), _templateEngine);
-                }
+                return _cache.GetDependentDocument(dependentDocset, pathToDocset, () => CreateFromGit(dependentDocset, new FilePath(pathToDocset.Substring(dependencyName.Length), dependencyName)));
             }
 
             // resolve from entry docset
+            pathToDocset = PathUtility.NormalizeFile(pathToDocset);
```
yufeih: pathToDocset is already normalized

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -255,37 +257,25 @@ internal class DependencyResolver
...
-                    {
-                        return PathUtility.NormalizeFile(aliasPath + path.Substring(alias.Length));
-                    }
-                }
-            }
-            return docsetRelativePath;
-        }
+            Debug.Assert(referencingFile != null);
+            Debug.Assert(!string.IsNullOrEmpty(pathToDocset));
+            Debug.Assert(!Path.IsPathRooted(pathToDocset));
```
yufeih: these assertion are not necessary because it is now a private method

src/docfx/build/context/Cache.cs
OsmondJiang: Good suggestion.

src/docfx/build/Build.cs
OsmondJiang: good point

src/docfx/build/dependency/DependencyResolver.cs
OsmondJiang: That's what I met in the integration test, will find a way to know that.

src/docfx/docset/Document.cs
OsmondJiang: It's used in other place as well

OsmondJiang: But it's a good suggestion, we can remove it.

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -318,7 +300,65 @@ private Document TryResolveFile(Document referencingFile, string pathToDocset)
...
+            }
+
+            if (File.Exists(Path.Combine(docset.DocsetPath, filePath.Path)))
+            {
+                return Document.Create(docset, filePath, _templateEngine);
+            }
+
+            var (repo, pathToRepo, commits) = _gitCommitProvider.GetCommitHistory(docset, filePath.Path, docset.Repository.Commit);
+
+            commits = deleted && commits.Count() > 1 ? commits.Skip(1).ToList() : (!deleted && commits.Count() > 0 ? commits : default);
```
live1206: `commits` is `List`, can just use `Count`

OsmondJiang:
the azure diffs are caused by different file name cases. Git operation works when the file path name case matches, we need find a way to resolve it.

image

image

OsmondJiang:
> the azure diffs are caused by different file name cases. Git operation works when the file path name case matches, we need find a way to resolve it.
> 
> image
> 
> image

Need implement the ListFiles for dependency docsets and map the real file path with user write file path

OsmondJiang:
The diffs are expected, we are using real file case for dependency files

src/docfx/localization/LocalizationUtility.cs
```diff
@@ -48,14 +48,14 @@ public static (string remote, string branch) GetLocalizedRepo(LocalizationMappin
             return (newRemote, newBranch);
         }
 
-        public static bool TryGetLocalizedDocsetPath(Docset docset, RestoreGitMap restoreMap, Config config, string locale, out string localizationDocsetPath, out string localizationBranch)
+        public static bool TryGetLocalizationDocset(RestoreGitMap restoreGitMap, Docset docset, Config config, string locale, out string localizationDocsetParh, out Repository localizationRepository)
```
mingwli: `localizationDocsetParh` typo

src/docfx/build/context/Input.cs
```diff
@@ -127,11 +126,12 @@ public FilePath[] ListFilesRecursive(FileOrigin origin, string dependencyName =
                         .ToArray();
 
                 case FileOrigin.Dependency:
-                    var dependencyPath = _restoreMap.GetGitRestorePath(_config.Dependencies[dependencyName], _docsetPath);
-                    return Directory
-                        .GetFiles(dependencyPath, "*", SearchOption.AllDirectories)
+                    var (dependencyPath, commit) = _restoreMap.GetRestoreGitPath(_config.Dependencies[dependencyName], _docsetPath, true);
+
+                    // todo: get tree list from repository
```
mingwli: 👌

test/docfx.Test/lib/ProcessUtilityTest.cs
```diff
@@ -113,113 +113,99 @@ public static void ParallelNestedRunInMutexWithSameNameTest()
...
-                        break;
-                    case "a-e": // acquire exclusive lock
-                        (acquired, acquirer) = ProcessUtility.AcquireExclusiveLock(guid + parts[1]);
-                        Debug.Assert(acquired == results[i], $"{i}");
-                        if (acquired)
+                    }),
+                    Task.Run(() =>
+                    {
+                        barrier.SignalAndWait();
+                        using (InterProcessReaderWriterLock.AcquireWriterLock(name))
```
live1206: From the test name, I guess it should be `InterProcessReaderWriterLock.AcquireReaderLock` here?

src/docfx/lib/process/InterProcessReaderWriterLock.cs
```diff
@@ -0,0 +1,85 @@
...
+                using (InterProcessMutex.Create(_name))
+                {
+                    _fileLock.Dispose();
+                }
+            }
+        }
+
+        public static InterProcessReaderWriterLock AcquireReaderLock(string name)
+        {
+            name = HashUtility.GetMd5Hash(name);
```
yufeih: hash is calculated twice in line 34/35 and 43/44

test/docfx.Test/lib/ProcessUtilityTest.cs
OsmondJiang: I followed the old rule, all test related to the process in here, like InterProcessMutex

src/docfx/lib/process/InterProcessReaderWriterLock.cs
```diff
@@ -0,0 +1,85 @@
...
+                    Directory.CreateDirectory(Path.GetFullPath(Path.GetDirectoryName(path)));
+
+                    using (InterProcessMutex.Create(name))
+                    {
+                        return new FileStream(path, FileMode.OpenOrCreate, access, share);
+                    }
+                }
+                catch
+                {
+                    Thread.Sleep(TimeSpan.FromMilliseconds(200));
```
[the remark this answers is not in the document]
OsmondJiang: image

PR #5034, opened 2019-08-27
📨 Add a basic input abstraction

yufeih:
Depending on #5003 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5034)

### DIFF
--- a/src/docfx/build/context/Input.cs
+++ b/src/docfx/build/context/Input.cs
@@ -23,6 +23,23 @@ namespace Microsoft.Docs.Build
             _fallbackPath = fallbackPath is null ? null : Path.GetFullPath(fallbackPath);
         }
 
+        public bool Exists(FilePath file)
+        {
+            var (basePath, path, commit) = ResolveFilePath(file);
+
+            if (basePath is null)
+            {
+                return false;
+            }
+
+            if (commit is null)
+            {
+                return File.Exists(Path.Combine(basePath, path));
+            }
+
+            throw new NotSupportedException("Checking if a file exists in a git repo");
+        }
+
         public bool TryGetPhysicalPath(FilePath file, out string physicalPath)
         {
             var (basePath, path, commit) = ResolveFilePath(file);
--- a/src/docfx/build/context/Input.cs
+++ b/src/docfx/build/context/Input.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.IO;
 
 namespace Microsoft.Docs.Build
@@ -12,6 +13,7 @@ namespace Microsoft.Docs.Build
         private readonly string _fallbackPath;
         private readonly Config _config;
         private readonly RestoreGitMap _restoreMap;
+        private readonly ConcurrentDictionary<FilePath, byte[]> _gitBlobCache = new ConcurrentDictionary<FilePath, byte[]>();
 
         public Input(string docsetPath, string fallbackPath, Config config, RestoreGitMap restoreMap)
         {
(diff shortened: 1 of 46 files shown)


src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -288,34 +276,32 @@ private string ApplyResolveAlias(string path, Document referencingFile)
             return docsetRelativePath;
         }
 
-        private Document TryResolveFile(Document referencingFile, string pathToDocset)
+        private Document TryResolveFile(Document declaringFile, string pathToDocset, bool lookupFallbackCommits)
```
```suggestion
        private Document TryResolveFile(Document referencingFile, string pathToDocset, bool lookupFallbackCommits)
```

src/docfx/lib/git/GitUtility.cs
```diff
@@ -380,5 +371,33 @@ string GetSecretFromHeader(KeyValuePair<string, string> header)
                 return header.Value;
             }
         }
+
+        private class GitBlobStream : UnmanagedMemoryStream
+        {
+            private bool _disposed;
+            private IntPtr _repo;
```
live1206:
```suggestion
            private readonly IntPtr _repo;
```

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -326,56 +312,23 @@ private Document TryResolveFile(Document referencingFile, string pathToDocset)
...
-                var (repo, pathToRepo, commits) = gitCommitProvider.GetCommitHistory(_fallbackDocset, pathToDocset);
-                if (repo != null)
+                // resolve from fallback docset git commit history
+                if (lookupFallbackCommits)
                 {
-                    var repoPath = PathUtility.NormalizeFolder(repo.Path);
-                    if (commits.Length > 1)
+                    var (repo, _, commits) = _gitCommitProvider.GetCommitHistory(_fallbackDocset, pathToDocset);
+
+                    // TODO: need to get the first commit that contains the data
```
live1206: Could you elaborate a bit more about this TODO, does this mean `commits[1]` might not be the one containing data?

src/docfx/build/context/Input.cs
```diff
@@ -0,0 +1,115 @@
...
+                ?? throw new InvalidOperationException($"Error reading '{file}'");
+
+            return new StreamReader(new MemoryStream(bytes, writable: false));
+        }
+
+        private (string basePath, string path, string commit) ResolveFilePath(FilePath file)
+        {
+            switch (file.Origin)
+            {
+                case FileOrigin.Default when file.Commit is null:
```
OsmondJiang: >case [](start = 16, length = 4)

From input abstraction view of point, it's better avoid to invloving too many business logic in Input.

Default, Dependency, Fallback, Template can all map to Physical, Git and Memory.

Considering that in the future we add new business dependency model, we don't want to change Input to support it.

yufeih: The design goal of `FilePath` is serializable and roamable (machine independent), thus it is designed this way. Using __physical/git/memory__ is machine dependent.

OsmondJiang: physical/git/memory is not always be machine dependent, and also, we should go details about the puporse of serializable/roadmable.

---
In reply to: [320084723](https://github.com/dotnet/docfx/pull/5034#discussion_r320084723) [](ancestors = 320084723)

src/docfx/build/context/Input.cs
```diff
@@ -0,0 +1,167 @@
...
+            {
+                return false;
+            }
+
+            if (commit is null)
+            {
+                return File.Exists(Path.Combine(basePath, path));
+            }
+
+            throw new NotSupportedException($"{nameof(Exists)}: {file}");
```
OsmondJiang: TODO?

src/docfx/build/context/Input.cs
```diff
@@ -0,0 +1,167 @@
...
+            switch (file.Origin)
+            {
+                case FileOrigin.Default when file.Commit is null:
+                    return (_docsetPath, file.Path, null);
+
+                case FileOrigin.Dependency when file.Commit is null:
+                    var (dependencyPath, _) = _restoreMap.GetGitRestorePath(_config.Dependencies[file.DependencyName], _docsetPath);
+                    return (dependencyPath, file.Path, null);
+
+                case FileOrigin.Fallback when file.Commit is null && _fallbackPath != null:
```
OsmondJiang: a little duplicated

src/docfx/build/context/Input.cs
```diff
@@ -0,0 +1,167 @@
...
+            if (commit is null)
+            {
+                return File.Exists(Path.Combine(basePath, path));
+            }
+
+            throw new NotSupportedException($"{nameof(Exists)}: {file}");
+        }
+
+        /// <summary>
+        /// Try get the absolute path of the specified file if it exit physically on disk.
```
live1206:
```suggestion
        /// Try get the absolute path of the specified file if it exists physically on disk.
```

PR #5044, opened 2019-08-29
flatten restored git lock and only restore first level dependencies

OsmondJiang:
1. only restore the first level of dependencies
2. flatten the dependency lock model
3. handle fallback with overwrite config specifically

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5044)


src/docfx/restore/Restore.cs
```diff
@@ -19,83 +19,79 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Erro
...
+                var (extendConfigErrors, extendedConfig) = ConfigLoader.TryLoad(restoreFallbackResult?.Path ?? docsetPath, options, localeToRestore, extend: true);
+                errorLog.Write(extendConfigErrors);
 
                 // restore urls except extend url
                 var restoreUrls = extendedConfig.GetFileReferences().Where(UrlUtility.IsHttp).ToList();
                 await RestoreFile.Restore(restoreUrls, extendedConfig);
 
-                var generatedLock = new DependencyLockModel
+                // restore git repos includes dependency repos, theme repo and loc repos
+                var restoredDependencyLock = RestoreGit.Restore(extendedConfig, localeToRestore, repository, DependencyLockProvider.Load(docsetPath, extendedConfig.DependencyLock));
```
OsmondJiang: restoredResults?

src/docfx/restore/Restore.cs
```diff
@@ -19,83 +19,77 @@ public static async Task Run(string docsetPath, CommandLineOptions options, Erro
...
-                            rootRepository,
-                            dependencyLock,
-                            root);
-                    });
-                }
-            }
+                var (fallbackErrors, fallbackConfig, restoreFallbackResult) = RestoreFallbackRepo(docsetPath, config, repository, localeToRestore, options);
+                errors.AddRange(fallbackErrors);
+
+                config = fallbackConfig ?? config;
```
mingwli: One suggestion:
the returned `fallbackConfig` would be `null` if current repo has config. 
But `config = fallbackConfig ?? config;` is more like saying `fallbackConfig` has higher priority.

Can we get rid of this misleading by making `RestoreFallbackRepo` always return a `fallbackConfig`, and uses
`config = config ?? fallbackConfig;`?

docs/specs/restore.yml
```diff
@@ -204,11 +204,7 @@ repos:
 cache:
   .lock.json: |
     {
-      "git":{
-         "https://docs.com/lock1/c#master":{
-            "commit":"8720d784fb99abcbdfa922025b8bcc060f0f622a"
-         }
-      }
+      "https://docs.com/lock1/c#master":"8720d784fb99abcbdfa922025b8bcc060f0f622a"
```
yufeih: Use array of objects in persistent data model, this give us ways to upgrade in the future without breaking changes. Most of existing persistent data model follows this principle:

{
   "git": [
       { "url": "", "branch": "", "commit": "" },
       { "url": "", "branch": "", "commit": "" },
    ]
}

src/docfx/restore/RestoreFile.cs
```diff
@@ -77,6 +77,8 @@ public static IEnumerable<string> GetFileReferences(this Config config)
 
             yield return config.MonikerDefinition;
 
+            yield return config.DependencyLock;
```
yufeih: consider define this as local file only

src/docfx/restore/RestoreFlags.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace Microsoft.Docs.Build
+{
+    [Flags]
+    internal enum GitFlags
```
yufeih: name does not match file name

OsmondJiang: my mistake, but why no warnings reported?

src/docfx/restore/RestoreFile.cs
OsmondJiang: We can do it in another PR, will add a todo here.

src/docfx/restore/Restore.cs
OsmondJiang: Good suggestion, but I think for transparent purpose, we should move this logic out of RestoreFallbackRepo

    the returned fallbackConfig would be null if current repo has config

docs/specs/restore.yml
OsmondJiang: array would make it hard for lookup. how about use remote url as key

docs/specs/restore.yml
```diff
@@ -204,10 +204,14 @@ repos:
 cache:
   .lock.json: |
     {
-      "git":{
-         "https://docs.com/lock1/c#master":{
-            "commit":"8720d784fb99abcbdfa922025b8bcc060f0f622a"
-         }
+      "git":
+      {
+        "https://docs.com/lock1/c#master":
```
yufeih: use array over object, the object key contains duplicated info
```suggestion
        "git": [ { url, branch, commit }, { url, branch, commit } ]
```

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -23,13 +25,15 @@ namespace Microsoft.Docs.Build
 
         public readonly string Branch;
 
+        public readonly bool HasRefSpec;
```
yufeih: remove this property from `PackageUrl`

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -13,7 +15,7 @@ namespace Microsoft.Docs.Build
     /// The commit-sh can be any tag, sha, or branch. The default commit-ish is master.
     /// </summary>
     [JsonConverter(typeof(ShortHandConverter))]
-    internal readonly struct PackageUrl
+    internal readonly struct PackageUrl : IEquatable<PackageUrl>, IComparable<PackageUrl>
```
yufeih: when will you need to compare equality for PackageUrl

src/docfx/lib/path/PackageUrl.cs
OsmondJiang: sure

docs/specs/restore.yml
OsmondJiang: will change that later because it requires backward compatibility

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -60,5 +65,85 @@ public override string ToString()
...
+
+            switch (Type)
+            {
+                case PackageType.Folder:
+                    return string.Equals(Path, other.Path, PathUtility.PathComparison);
+
+                case PackageType.Git:
+                    return string.Equals(Remote, other.Remote) && string.Equals(Branch, other.Branch);
+
+                default:
```
yufeih: consider full equals comparison without throwing and switch case

PR #5056, opened 2019-09-02
Take zip file as external xref source

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/116338

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5056)

### DIFF
--- a/src/docfx/build/xref/ExternalXrefMapLoader.cs
+++ b/src/docfx/build/xref/ExternalXrefMapLoader.cs
@@ -24,9 +24,23 @@ namespace Microsoft.Docs.Build
                 {
                     LoadZipFile(result, new SourceInfo<string>(Path.Combine(docset.DocsetPath, url), url.Source));
                 }
+                else if (url.Value.EndsWith(".yml", StringComparison.OrdinalIgnoreCase))
+                {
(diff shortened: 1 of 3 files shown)


src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -64,6 +69,35 @@ public static List<(string, Lazy<ExternalXrefSpec>)> Load(string filePath)
             return result;
         }
 
+        private static void LoadZipFile(Dictionary<string, Lazy<ExternalXrefSpec>> result, SourceInfo<string> url)
+        {
+            using (var archive = ZipFile.OpenRead(url.Value))
```
928PJY: >Value [](start = 54, length = 5)

so the zip file cannot be an external resource?

live1206: The zip file is used as external resource

live1206: There is no real usage we have a remote url of zip file. Even we have it, think it should be handled by restore

PR #5061, opened 2019-09-04
Make RestoreGitMap.DependencyGitLock and docsetPath private

yufeih:
So these internal states don't leak to other components.

Print the content of lock file to stdout

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5061)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -192,8 +192,7 @@ namespace Microsoft.Docs.Build
             var dependencyLockPath = string.IsNullOrEmpty(config.DependencyLock)
                 ? new SourceInfo<string>(AppData.GetDependencyLockFile(docsetPath, locale)) : config.DependencyLock;
 
-            var dependenyGitLock = DependencyLockProvider.LoadGitLock(docsetPath, dependencyLockPath) ?? new Dictionary<PackageUrl, DependencyGitLock>();
-            return RestoreGitMap.Create(dependenyGitLock);
+            return RestoreGitMap.Create(docsetPath, dependencyLockPath);
         }
 
         private static Repository GetFallbackRepository(
@@ -206,16 +205,14 @@ namespace Microsoft.Docs.Build
 
             if (LocalizationUtility.TryGetFallbackRepository(repository, out var fallbackRemote, out string fallbackBranch, out _))
             {
-                var fallbackPackageUrl = new PackageUrl(fallbackRemote, fallbackBranch);
-                if (restoreGitMap.DependencyGitLock.GetGitLock(fallbackPackageUrl) == null
-                    && restoreGitMap.DependencyGitLock.GetGitLock(new PackageUrl(fallbackRemote, "master")) != null)
+                foreach (var branch in new[] { fallbackBranch, "master" })
                 {
-                    // fallback to master branch
-                    fallbackPackageUrl = new PackageUrl(fallbackRemote, "master");
+                    var fallbackRepoPath = restoreGitMap.TryGetGitRestorePath(new PackageUrl(fallbackRemote, branch));
+                    if (fallbackRepoPath != null)
+                    {
+                        return Repository.Create(fallbackRepoPath, branch, fallbackRemote);
+                    }
                 }
-
-                var fallbackRepoPath = restoreGitMap.GetGitRestorePath(fallbackPackageUrl);
-                return Repository.Create(fallbackRepoPath, fallbackBranch, fallbackRemote);
             }
 
             return default;
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -205,9 +205,9 @@ namespace Microsoft.Docs.Build
             {
                 foreach (var branch in new[] { fallbackBranch, "master" })
                 {
-                    var fallbackRepoPath = restoreGitMap.TryGetGitRestorePath(new PackageUrl(fallbackRemote, branch));
-                    if (fallbackRepoPath != null)
+                    if (restoreGitMap.BranchExists(fallbackRemote, branch))
                     {
+                        var fallbackRepoPath = restoreGitMap.GetGitRestorePath(new PackageUrl(fallbackRemote, branch));
                         return Repository.Create(fallbackRepoPath, branch, fallbackRemote);
                     }
                 }
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -189,10 +189,8 @@ namespace Microsoft.Docs.Build
             Debug.Assert(!string.IsNullOrEmpty(docsetPath));
 
             var (_, config) = ConfigLoader.TryLoad(docsetPath, commandLineOptions);
-            var dependencyLockPath = string.IsNullOrEmpty(config.DependencyLock)
-                ? new SourceInfo<string>(AppData.GetDependencyLockFile(docsetPath, locale)) : config.DependencyLock;
 
-            return RestoreGitMap.Create(docsetPath, dependencyLockPath);
+            return RestoreGitMap.Create(docsetPath, config, locale);
         }
 
         private static Repository GetFallbackRepository(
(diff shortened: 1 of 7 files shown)


src/docfx/restore/RestoreGitMap.cs
```diff
@@ -11,67 +11,77 @@ namespace Microsoft.Docs.Build
...
     internal class RestoreGitMap : IDisposable
     {
+        private readonly string _docsetPath;
+        private readonly IReadOnlyDictionary<PackageUrl, DependencyGitLock> _dependencyGitLock;
         private readonly IReadOnlyDictionary<(PackageUrl packageUrl, string commit), (string path, DependencyGit git)> _acquiredGits;
 
-        public Dictionary<PackageUrl, DependencyGitLock> DependencyGitLock { get; private set; }
-
-        public RestoreGitMap(IReadOnlyDictionary<(PackageUrl packageUrl, string commit), (string path, DependencyGit git)> acquiredGits = null)
+        public RestoreGitMap(
```
live1206: Make this constructor private and call `RestoreGitMap.Create` in `TocTest`?

yufeih: unfortunately `Create` does a lot extra logic like reading and serializing a file. That test badly needs improvements.

live1206: I think we should handle `null` in `Create` instead

src/docfx/restore/RestoreGitMap.cs
```diff
@@ -89,13 +99,16 @@ public void Dispose()
...
         /// </summary>
-        public static RestoreGitMap Create(Dictionary<PackageUrl, DependencyGitLock> dependencyLock)
+        public static RestoreGitMap Create(string docsetPath, Config config, string locale)
         {
             var acquired = new Dictionary<(PackageUrl packageUrl, string commit), (string path, DependencyGit git)>();
 
             try
             {
-                Debug.Assert(dependencyLock != null);
+                var dependencyLockPath = string.IsNullOrEmpty(config.DependencyLock)
```
live1206:
```suggestion
                var dependencyLockPath = string.IsNullOrEmpty(config?.DependencyLock)
```

src/docfx/restore/RestoreGitMap.cs
```diff
@@ -89,13 +99,16 @@ public void Dispose()
...
+                    ? new SourceInfo<string>(AppData.GetDependencyLockFile(docsetPath, locale)) : config.DependencyLock;
```
live1206:
```suggestion
                    ? new SourceInfo<string>(AppData.GetDependencyLockFile(docsetPath, locale)) : config?.DependencyLock ?? new SourceInfo<string>(string.Empty);
```

src/docfx/template/TemplateEngine.cs
```diff
@@ -149,9 +149,7 @@ public static TemplateEngine Create(Docset docset, RestoreGitMap restoreGitMap)
             }
 
             var theme = LocalizationUtility.GetLocalizedTheme(docset.Config.Template, docset.Locale, docset.Config.Localization.DefaultLocale);
-            var themePath = restoreGitMap.GetGitRestorePath(theme, docset.DocsetPath);
```
OsmondJiang: don't remove this, it's really helpful for debugging if the restore with dependency lock works well.

src/docfx/build/context/Input.cs
```diff
@@ -127,7 +127,7 @@ public FilePath[] ListFilesRecursive(FileOrigin origin, string dependencyName =
                         .ToArray();
 
                 case FileOrigin.Dependency:
-                    var dependencyPath = _restoreMap.GetGitRestorePath(_config.Dependencies[dependencyName], _docsetPath);
```
OsmondJiang: it may be not always relative to current root docset

OsmondJiang: like fallback repo has it's own template relative path, and loc is using overwrite config

src/docfx/template/TemplateEngine.cs
yufeih: It is now printed here: https://github.com/dotnet/docfx/pull/5061/files#diff-77d5963be746079df8655d36a6a64e91R49

src/docfx/restore/RestoreGitMap.cs
```diff
@@ -11,67 +11,77 @@ namespace Microsoft.Docs.Build
...
-                    throw new NotSupportedException($"Unknown package url: '{url}'");
+                    throw new NotSupportedException($"Unknown package url: '{packageUrl}'");
             }
         }
 
-        /// <summary>
-        /// The dependency lock must be loaded before using this method
-        /// </summary>
-        public string GetGitRestorePath(PackageUrl packageUrl)
+        public string TryGetGitRestorePath(PackageUrl packageUrl)
```
OsmondJiang: Then we have two methods with same name "TryGetRestorePath".

With Dependency Lock, we should avoid to provide TryXXX because every package should map to a restored path, otherwise we throw need-restore errors.

src/docfx/restore/RestoreGitMap.cs
```diff
@@ -11,67 +11,77 @@ namespace Microsoft.Docs.Build
 {
     internal class RestoreGitMap : IDisposable
     {
+        private readonly string _docsetPath;
+        private readonly IReadOnlyDictionary<PackageUrl, DependencyGitLock> _dependencyGitLock;
```
OsmondJiang: I don;t think it's internal model, it's the contract between docfx and docs pipeline, we are using it for tracking versioning of dependency, not only for restore.

OsmondJiang: And we still used it a lot during restore

src/docfx/build/Build.cs
```diff
@@ -206,16 +203,14 @@ private static bool ShouldBuildFile(Context context, Document file)
...
             if (LocalizationUtility.TryGetFallbackRepository(repository, out var fallbackRemote, out string fallbackBranch, out _))
             {
-                var fallbackPackageUrl = new PackageUrl(fallbackRemote, fallbackBranch);
-                if (restoreGitMap.DependencyGitLock.GetGitLock(fallbackPackageUrl) == null
-                    && restoreGitMap.DependencyGitLock.GetGitLock(new PackageUrl(fallbackRemote, "master")) != null)
+                foreach (var branch in new[] { fallbackBranch, "master" })
                 {
-                    // fallback to master branch
-                    fallbackPackageUrl = new PackageUrl(fallbackRemote, "master");
+                    if (restoreGitMap.BranchExists(fallbackRemote, branch))
```
OsmondJiang: IsRestored?

src/docfx/restore/RestoreGitMap.cs
```diff
@@ -11,67 +11,87 @@ namespace Microsoft.Docs.Build
...
         private readonly IReadOnlyDictionary<(PackageUrl packageUrl, string commit), (string path, DependencyGit git)> _acquiredGits;
 
-        public Dictionary<PackageUrl, DependencyGitLock> DependencyGitLock { get; private set; }
-
-        public RestoreGitMap(IReadOnlyDictionary<(PackageUrl packageUrl, string commit), (string path, DependencyGit git)> acquiredGits = null)
+        public RestoreGitMap(
```
OsmondJiang: private it and always use Create()?

test/docfx.Test/build/TocTest.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Docs.Build
 {
     public static class TocTest
     {
-        private static readonly RestoreGitMap s_restoreGitMap = new RestoreGitMap();
+        private static readonly RestoreGitMap s_restoreGitMap = new RestoreGitMap(Directory.GetCurrentDirectory(), null, null);
```
OsmondJiang: RestoreGitMap.Create(Directory.GetCurrentDirectory(), new Config(), "en-us");

PR #5062, opened 2019-09-05
output for links design spec

OsmondJiang:
as title

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5062)


docs/designs/link-output.md
```diff
@@ -0,0 +1,97 @@
+# Link And Bookmarks Output
+
+Docfx generate `links` and `bookmarks` information of each files into one single file `links.json` for future data utilization of downstream service of docfx.
```
yufeih:
```suggestion
Docfx generate `links` and `bookmarks` information of each files into one single file `.links.json` for future data utilization of downstream service of docfx.
```

yufeih: do we need to a config flag to control the generation of this file (as well as other files like .xrefmap.json) since it is only useful for English master/live commit build

docs/designs/link-output.md
```diff
@@ -0,0 +1,97 @@
...
+  |----------- |--------------------- |
+  | `id`       | title-h1             |
+
+Example:
+
+```json
+[
+    {
+        "path": "dotnet/articles/a.md",
+        "site-url": "/dotnet/articles/a",
```
yufeih: Consider use `url` here, the meaning it the same as the `url` in `.publish.json`
```suggestion
        "url": "/dotnet/articles/a",
```

docs/designs/link-output.md
```diff
@@ -0,0 +1,97 @@
...
+  |            |                      |
+  |----------- |--------------------- |
+  | `id`       | title-h1             |
+
+Example:
+
+```json
+[
+    {
+        "path": "dotnet/articles/a.md",
```
yufeih: consider remove `path` so we only exposed what is needed at first. 
```suggestion
        ~~"path": "dotnet/articles/a.md",~~
```

docs/designs/link-output.md
```diff
@@ -0,0 +1,97 @@
...
+                "link": "dotnet/articles/b"
+            },
+            {
+                "link": "./c.md"
+            },
+            {
+                "link": "https://docs.com/y"
+            }
+        ],
+        "bookmarks":
```
yufeih: I'm not sure if we can simply `bookmarks` into an array, link is an object because we might very much likely add additional properties like link type, validate or not. I cannot currently think of a useful property that is tied to a bookmark.
```suggestion
        "bookmarks": [ "title-h1", "title-2" ]
```

herohua: `links` can also be an array?

---
In reply to: [321062224](https://github.com/dotnet/docfx/pull/5062#discussion_r321062224) [](ancestors = 321062224)

docs/designs/link-output.md
```diff
@@ -0,0 +1,97 @@
...
+`links` includes all the links which has been processed by docfx:
+
+- Relative link in source content will be processed by docfx and convert it to absolute link if it's valid, otherwise keep it untouched.
+
+- Absolute link in source content will NOT processed by docfx
+
+| Links in source content | Examples              | Output in `links.json` |
+| ----------------------- | --------------------- | ---------------------- |
+| Relative Link           | ./a.md (valid)        | /basepath/article/a    |
+|                         | a/b.md (invalid)      | a/b.md                 |
```
superyyrrzz: Is `a/b.md` input or output? I think even an invalid input `a/b.md` will be `a/b` in output, then the output in `links.json` is `a/b`.

docs/designs/link-output.md
```diff
@@ -0,0 +1,97 @@
...
+| Absolute Link           | /basepath/articles/b  | /basepath/articles/b   |
+|                         | www.github.com/x      | www.github.com/x       |
+|                         | https://docs.com/y    | https://docs/com/y     |
+
+> NOTE: the absolute links converted from relative links by docfx don't contains `locale` and `versioning` information
+
+##  Bookmarks
+
+`bookmarks` includes all bookmarks ids generated/defined from source content, includes its inclusions(token/code-snippets)
```
superyyrrzz: What about bookmarks ids generated/defined from template? Do we need to include or exclude them? I think Ying is confirming with Adam about whether template can export this.

docs/designs/link-output.md
```diff
@@ -0,0 +1,97 @@
...
+| `<a id="with:colon"></a>`      | `<a id="with:colon"></a>`                | with:colon             |
+| `[!include[](d.md)]`           | `<h1 id="title-h1">Title H1</h1>`        | title-h1               |
+
+> The content in d.md is `# Title H1` in above case
+
+## Output
+
+`links.json` file will be generated by docfx to contains all the links and bookmarks information about all the page files in build.
+
+The output is **an array** of below data model
```
superyyrrzz: consider making root level object instead of array, other it will be hard to insert a global scope data into it in the future.

docs/designs/link-output.md
OsmondJiang: Even we have this config, we still need handle non-English cases, and we don't want different config in different branches.

docs/designs/link-output.md
OsmondJiang: it's input.
In v3, we don't touch any links which are invalid.

docs/designs/link-output.md
OsmondJiang: We don't need generate bookmark info anymore. Deleted this part

docs/designs/link-output.md
OsmondJiang: @yufeih

docs/designs/link-output.md
OsmondJiang: @herohua , for extensionality. we want to make it as object in case we want to add more properties like `type`

docs/designs/link-output.md
```diff
@@ -0,0 +1,61 @@
...
+  |----------- |-----------------------|
+  | `links`    | Array of *link model* |
+
+*link model*:
+
+  |                   |                      |
+  |-------------------|----------------------|
+  | `source_url`      | /dotnet/articles/a   |
+  | `target_url`      | /dotnet/articles/b   |
+  | `source_monikers` | [`dotnet-1.1`]       |
```
yufeih: Is this `source_monikers` required for now? can we start without this property?

OsmondJiang: We can tell after meeting with Jason

docs/designs/link-output.md
```diff
@@ -0,0 +1,64 @@
...
+    "links":
+    [
+        {
+            "source_url": "/dotnet/articles/a",
+            "source_moniker_group": "218c13d0",
+            "target_url": "/dotnet/articles/b"
+        },
+        {
+            "source_url": "/dotnet/articles/a",
+            "source_moniker_group": "218c13d0",
```
XiaofuHuang: Can you give me the moniker name list too, we want to ping the page with moniker.

PR #5065, opened 2019-09-05
support file groups

mingwli:
Support file groups.
FileGroups can be defined in
1. in docfx.yml root;
2. in fileGroups as array or object;

When defining FileGroup in both root and fileGroups property, fileGroups have higher priority.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5065)

### DIFF
--- a/src/docfx/build/context/BuildScope.cs
+++ b/src/docfx/build/context/BuildScope.cs
@@ -15,7 +15,7 @@ namespace Microsoft.Docs.Build
         // https://github.com/dotnet/corefx/issues/1086
         // This lookup table stores a list of actual filenames.
         private readonly HashSet<string> _fileNames;
-        private readonly Func<string, bool> _glob;
+        private readonly Func<string, bool>[] _globs;
 
         private readonly Input _input;
         private readonly TemplateEngine _templateEngine;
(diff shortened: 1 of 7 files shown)


mingwli:
DevOps [#118457](https://dev.azure.com/ceapex/Engineering/_workitems/edit/118457)

src/docfx/config/ConfigLoader.cs
```diff
@@ -86,6 +88,34 @@ string GetBranch()
             }
         }
 
+        private static void UpdateFileScope(JObject configObject)
```
yufeih: instead do if here, keep the raw config unchanged and merge it at `BuildScope.CreateGlob` method

src/docfx/build/context/BuildScope.cs
```diff
@@ -76,11 +76,17 @@ public bool GetActualFileName(string fileName, out string actualFileName)
             }
         }
 
-        private static Func<string, bool> CreateGlob(Config config)
+        private static Func<string, bool>[] CreateGlob(Config config)
```
yufeih: The return value can still be `Func<string, bool>`

```suggestion
        private static Func<string, bool> CreateGlob(Config config)
```

mingwli:
off-line synced with @yufeih 

1. Intentionally keep `files` and `exclude` inside `Config.cs`, for this may explicitly declare that such properties are supported.
2. Prefer not to add merge logic inside `ConfigLoader`, this brings out point 3 that we can't tell if user explicitly defines `files` or `exclude` from an initialized `Config` instance.
3. Since we support default `files` include glob in `Config` for its own simplicity, if users really defines `fileGroups` and ignoring root-level `files` or `exclude`, and we still merge them together, the behavior would be against user's will. And hence we use `fileGroups` first if it exists and ignore root-level build-scope, rather than merging them together.

cc @OsmondJiang @live1206 @928PJY

PR #5069, opened 2019-09-05
extend image Docs Markdown extension to support icons

adunndevster:
Work item: https://ceapex.visualstudio.com/Engineering/_workitems/edit/107014/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5069)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ChromelessFormExtension.cs
```diff
@@ -14,7 +14,7 @@ public class ChromelessFormExtension : ITripleColonExtensionInfo
     {
         public string Name => "form";
         public bool SelfClosing => true;
-        public bool EndingTripleColons => false;
+        public bool EndingTripleColons {get; set;} = true;
```
bharney0: Is this supposed to be false?

adunndevster: You are wonderful. Thanks for the catch :).

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
```diff
@@ -50,31 +57,71 @@ public bool TryProcessAttributes(IDictionary<string, string> attributes, out Htm
...
+
             if (string.IsNullOrEmpty(src))
             {
-                logError($"source is a required attribute. Please ensure you have specified a source attribute");
+                logError($"source is a required attribute. Please ensure you have specified a source attribute.");
             }
-            if (string.IsNullOrEmpty(alt))
+            if (string.IsNullOrEmpty(type))
+            {
+                logError($"type is a required attribute. Please ensure you have specified a type attribute. Options (\"comples\", \"content\", \"icon\")");
```
bharney0: Is this a typo "comples"? Should it be "complex"?

superyyrrzz:
@yishengjin1413 Could also help review this?

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
adunndevster: @bharney0 - Good catch... I made a small change, and removed this error altogether. Now "content" is defaulted if no "type" attribute is present.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
```diff
@@ -8,16 +8,19 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
...
     using System.Security.Cryptography;
     using System.Text;
+    using System.Text.RegularExpressions;
 
     public class ImageExtension : ITripleColonExtensionInfo
     {
         public string Name => "image";
         public bool SelfClosing => true;
-        public bool EndingTripleColons => true;
+        public bool EndingTripleColons { get; set; } = false;
```
yufeih: In docfx v3, there is only one global markdown engine, thus `ImageExtension` is also a global singleton that are used by multiple threads concurrently.

Changing this to a settable property leaks internal state across markdown documents and produces wrong result.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
```diff
@@ -8,16 +8,19 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
...
         public Func<HtmlRenderer, TripleColonBlock, bool> RenderDelegate { get; private set; }
+        private List<ImageProperties> imagesList = new List<ImageProperties>();
```
yufeih: Same as above reason, this list as a member variable will have threading problems in v3, as different threads are trying to add/use the same list when building multiple document at the same time. It could be turned into a local variable inside `Render` method.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
yufeih: `EndingTripleColons` flag is consumed in`Parser` stage, (before `Render`), changing the value in `Render` stage may not give you desired behavior

adunndevster: Thank you for the insight @yufeih ! I'll work on it!

adunndevster: @yufeih -Want to take a look again?

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
adunndevster: @yufeih - This should be resolved.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
yufeih: This looks correct now

yufeih:
@yishengjin1413 could you help take a look?

yishengjin1413:
Looks good to me.

PR #5060, opened 2019-09-04
design spec for using content in CRR

OsmondJiang:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5060)

PR #5070, opened 2019-09-06
support content in CRR(dependency) repositories

OsmondJiang:
#5060 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5070)

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -274,3 +274,54 @@ cache:
 outputs:
   docs/a.json: |
     {"conceptual":"<p>233333</p>"}
+---
+# build content in dependency repository
+# content in dependency repository:
+#   1. by default will not be built
+#   2. will be built when:
+#      1. explicitly turn on the inScope flag
+#      2. specify the dependency name in `build scope` and `routes`
+repos:
+  https://docs.com/crrcontent/a:
+    - files:
+        docfx.yml: |
+          files:
+            - "**/*"
+            - "crrb/**/*"
+          dependencies:
+            crrb:
+              url: https://docs.com/crrcontent/b
+              inScope: true
+            crrc: https://docs.com/crrcontent/c
+        a.md:
+  https://docs.com/crrcontent/b:
+    - files:
+        b.md:
+  https://docs.com/crrcontent/c:
+    - files:
+        c.md:
+outputs:
+  a.json:
+  crrb/b.json:
+---
+# routes files in dependency repository
+repos:
+  https://docs.com/crrcontent/1:
+    - files:
+        docfx.yml: |
+          files:
+            - "**/*"
+            - "crrb/**/*"
+          routes:
+            "crrb/": "xyz"
+          dependencies:
+            crrb:
+              url: https://docs.com/crrcontent/2
+              inScope: true
+        a.md:
+  https://docs.com/crrcontent/2:
+    - files:
+        b.md:
+outputs:
+  a.json:
+  xyz/b.json:
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -280,7 +280,7 @@ outputs:
 #   1. by default will not be built
 #   2. will be built when:
 #      1. explicitly turn on the inScope flag
-#      2. specify the dependency name in `build scope` and `routes`
+#      2. specify the dependency name in `build scope`
 repos:
   https://docs.com/crrcontent/a:
     - files:
@@ -304,7 +304,8 @@ outputs:
   a.json:
   crrb/b.json:
 ---
-# routes files in dependency repository
+# apply routes to the files in dependency repository
+# if no route applied, crr files will be built to `{dependencyName}/` folder
 repos:
   https://docs.com/crrcontent/1:
     - files:
@@ -318,10 +319,17 @@ repos:
             crrb:
               url: https://docs.com/crrcontent/2
               inScope: true
+            crrc:
+              url: https://docs.com/crrcontent/3
+              inScope: true
         a.md:
   https://docs.com/crrcontent/2:
     - files:
         b.md:
+  https://docs.com/crrcontent/3:
+    - files:
+        c.md:
 outputs:
   a.json:
   xyz/b.json:
+  crrc/c.json:
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -16,7 +16,7 @@ inputs:
     [!include[include to 2](~/docfx-test-dependencies2/docs/a.md)
 outputs:
   .errors.log: |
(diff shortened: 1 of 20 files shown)


src/docfx/config/DependencyPackageUrl.cs
```diff
@@ -0,0 +1,22 @@
...
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
+{
+    [JsonConverter(typeof(ShortHandConverter))]
+    internal readonly struct DependencyPackageUrl
+    {
+        public readonly PackageUrl Url;
+
+        public readonly bool InScope;
```
mingwli:
```
dependencies:
    _csharplang: 
        remoteUrl: [https://github.com/dotnet/csharplang](https://github.com/dotnet/csharplang)
        branch: master
        inScope: true
```
Can this be serialized?

OsmondJiang: no, unless we change packageurl to class and inherit it, @yufeih  what do you think?

src/docfx/docset/Document.cs
```diff
@@ -343,19 +343,21 @@ string ReplaceLast(string source, string find, string replace)
             }
         }
 
-        private static string ApplyRoutes(string path, IReadOnlyDictionary<string, string> routes, string siteBasePath)
+        private static string ApplyRoutes(FilePath path, IReadOnlyDictionary<string, string> routes, string siteBasePath)
         {
             // the latter rule takes precedence of the former rule
+            var pathToMatch = Path.Combine(path.DependencyName ?? "", path.Path).Replace("\\", "/");
+            var matchedPath = pathToMatch;
```
live1206: can we just use `pathToMatch`?

src/docfx/config/DependencyPackageUrl.cs
live1206: add a test for this case?

OsmondJiang: it's not supported, so if we want to support it, we need change `PackageUrl` to class, then I will also change the test correspondingly.

live1206: or do we really need to support the format above? the user can just specify branch in the fragment of the git url

yufeih: @OsmondJiang , I think turning `PackageUrl` to a class is a good option

src/docfx/build/context/Context.cs
```diff
@@ -61,17 +64,36 @@ public Context(string outputPath, ErrorLog errorLog, Docset docset, Docset fallb
...
                 BuildQueue,
                 GitCommitProvider,
                 BookmarkValidator,
-                restoreGitMap,
                 DependencyMapBuilder,
                 _xrefResolver,
                 TemplateEngine);
         }
 
+        private static Dictionary<string, (Docset docset, bool inScope)> LoadDependencies(Docset docset, RestoreGitMap restoreGitMap)
```
yufeih: consider not putting logic inside `Context`, `BuildScope` and `DependencyResolver` can build their own private data structure.

src/docfx/build/dependency/DependencyResolver.cs
```diff
@@ -152,9 +152,9 @@ internal class DependencyResolver
             }
 
             // Link to dependent repo, don't build the file, leave href as is
-            if (file.FilePath.Origin == FileOrigin.Dependency)
+            if (file.FilePath.Origin == FileOrigin.Dependency && !_buildScope.InScopeDependencyNames.Contains(file.FilePath.DependencyName))
```
yufeih: Checking if a dependency is `in scope` can be done by DependencyResolver itself without asking BuildScope.

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -14,23 +15,26 @@ namespace Microsoft.Docs.Build
     /// The commit-sh can be any tag, sha, or branch. The default commit-ish is master.
     /// </summary>
     [JsonConverter(typeof(ShortHandConverter))]
-    internal readonly struct PackageUrl
+    internal class PackageUrl : IEquatable<PackageUrl>, IComparable<PackageUrl>
```
yufeih: Consider changing `RestoreGitMap.DependencyGitLock` to use `(url, branch)` as key because `PackageUrl` represents more than git and it does not need to implement IEquatable or IComparable

src/docfx/build/context/Context.cs
OsmondJiang: they are sharing data structure, so they don;t need to build their own. what's the concern here? It's just a local variable

src/docfx/build/dependency/DependencyResolver.cs
OsmondJiang: Don't want to make `BuildScope` and `DependencyResolver` both understand `in scope`, logically it belongs to `BuildScope`

src/docfx/lib/path/PackageUrl.cs
OsmondJiang: RestoreGitMap can also get local path right now, so Package Url is perfect for current situation

src/docfx/build/context/Context.cs
OsmondJiang: I moved it out of context, treat it like `docset`, `fallbackdocset`.

src/docfx/lib/path/PackageUrl.cs
OsmondJiang: I will do it in another PR.

docs/specs/restore.yml
```diff
@@ -274,3 +274,65 @@ cache:
...
+  https://docs.com/crrcontent/a:
+    - files:
+        docfx.yml: |
+          files:
+            - "docs/**/*"
+            - "crrb/**/*"
+          dependencies:
+            crrb:
+              type: git
+              remoteUrl: https://docs.com/crrcontent/b
```
yufeih: `msftbot` merged this too soon, consider:

- url: https://docs.com/crrcontent/b
- buildFiles: true

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -14,23 +15,26 @@ namespace Microsoft.Docs.Build
     /// The commit-sh can be any tag, sha, or branch. The default commit-ish is master.
     /// </summary>
     [JsonConverter(typeof(ShortHandConverter))]
-    internal readonly struct PackageUrl
+    internal class PackageUrl : IEquatable<PackageUrl>
```
yufeih: it does not need to implement `IEquatable`

docs/specs/restore.yml
OsmondJiang: already did in another PR for URL

src/docfx/lib/path/PackageUrl.cs
OsmondJiang: Already did in another PR

PR #5071, opened 2019-09-06
Remove host name if sharing with referencing xref

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/117912

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5071)

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -46,7 +46,6 @@ namespace Microsoft.Docs.Build
                 return (xrefError, null, null, null);
             }
 
-            var resolvedHref = RemoveSharingHost(xrefSpec.Href, referencingFile.Docset.HostName);
             var name = xrefSpec.GetXrefPropertyValueAsString("name");
             var displayPropertyValue = xrefSpec.GetXrefPropertyValueAsString(displayProperty);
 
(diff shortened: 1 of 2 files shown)


src/docfx/build/xref/XrefResolver.cs
```diff
@@ -57,8 +58,8 @@ public XrefResolver(Context context, Docset docset, RestoreFileMap restoreFileMa
             {
                 queries["view"] = moniker;
             }
-            var resolvedHref = UrlUtility.MergeUrl(
-                xrefSpec.Href,
+            resolvedHref = UrlUtility.MergeUrl(
+                resolvedHref,
```
mingwli:
```suggestion
                RemoveSharingHost(xrefSpec.Href, referencingFile.Docset.HostName),
```

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -130,6 +131,16 @@ public XrefMapModel ToXrefMapModel()
             return model;
         }
 
+        private string RemoveSharingHost(string url, string hostName)
+        {
+            if (url.StartsWith(hostName, StringComparison.OrdinalIgnoreCase))
```
yufeih:
```suggestion
            if (url.StartsWith(hostName + '/', StringComparison.OrdinalIgnoreCase))
```

live1206: why do we need appending `/` with hostName here?

yufeih: `https://a.com/b` starts with `https://a.co`, but you shouldn't remove in this case 😁

live1206: Thanks for the explanation, fair enough

PR #5072, opened 2019-09-06
xref zip file loading order

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/118436

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5072)

### DIFF
--- a/src/docfx/build/xref/ExternalXrefMapLoader.cs
+++ b/src/docfx/build/xref/ExternalXrefMapLoader.cs
@@ -82,7 +82,7 @@ namespace Microsoft.Docs.Build
                             var xrefMap = YamlUtility.Deserialize<XrefMapModel>(sr.ReadToEnd(), new FilePath(url));
                             foreach (var spec in xrefMap.References)
                             {
-                                result.TryAdd(spec.Uid, new Lazy<ExternalXrefSpec>(() => spec));
+                                result[spec.Uid] = new Lazy<ExternalXrefSpec>(() => spec);
                             }
                         }
                         else if (entry.FullName.EndsWith(".json", StringComparison.OrdinalIgnoreCase))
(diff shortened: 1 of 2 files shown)


mingwli:
>                 foreach (var entry in archive.Entries)

is order stable here?

---
Refers to: src/docfx/build/xref/ExternalXrefMapLoader.cs:76 in fe43f0d. [](commit_id = fe43f0d683253658d216e7442472c6c652bc4596, deletion_comment = False)

live1206:
> > ```
> >             foreach (var entry in archive.Entries)
> > ```
> 
> is order stable here?
> 
> Refers to: src/docfx/build/xref/ExternalXrefMapLoader.cs:76 in [fe43f0d](https://github.com/dotnet/docfx/commit/fe43f0d683253658d216e7442472c6c652bc4596). [](commit_id = [fe43f0d](https://github.com/dotnet/docfx/commit/fe43f0d683253658d216e7442472c6c652bc4596), deletion_comment = False)

I think so since the type of `arvhive.Entries` is `ReadOnlyCollection<ZipArchiveEntry>`

src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -82,15 +82,15 @@ private static void LoadZipFile(Dictionary<string, Lazy<ExternalXrefSpec>> resul
                             var xrefMap = YamlUtility.Deserialize<XrefMapModel>(sr.ReadToEnd(), new FilePath(url));
                             foreach (var spec in xrefMap.References)
                             {
-                                result.TryAdd(spec.Uid, new Lazy<ExternalXrefSpec>(() => spec));
+                                result[spec.Uid] = new Lazy<ExternalXrefSpec>(() => spec);
```
OsmondJiang: this means the latter one has higher priority, does it apply to all or only works in zip files?

OsmondJiang: for the other parts, we are still using TryAdd

live1206: It only applies for zip files as v2 does. And yes, the behavior is inconsistent in v2

OsmondJiang: great to know

live1206:
#4450

PR #5073, opened 2019-09-06
Resolving monikerzone in token should base on root file's moniker

928PJY:
Fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/119189

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5073)

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -16,6 +16,7 @@ namespace Microsoft.Docs.Build
             "_site/**",             // Default output location
             "localization/**",      // Localization file when using folder convention
             "_themes/**",           // Default template location
+            "**/includes/**",       // Default includes folder
         };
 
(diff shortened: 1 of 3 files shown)


src/docfx/config/Config.cs
```diff
@@ -16,6 +16,7 @@ internal sealed class Config
             "_site/**",             // Default output location
             "localization/**",      // Localization file when using folder convention
             "_themes/**",           // Default template location
+            "**/includes/**",       // Default includes folder
```
OsmondJiang: is that a intentional change?

928PJY: removed

PR #5076, opened 2019-09-07
Docker support

baez90:
@superyyrrzz: I added:

* `.dockerignore` to keep the build context small
* one Dockerfile for a pure Mono based distribution
* one Dockerfile for an extended image including the .NET Core SDK (for ASP.NET Core apps)
* a wrapper script to mimic a `docfx` executable

I haven't tried to build the images on a Windows machine and therefore a few small improvements (e.g. explicit access rights of the wrapper script) might be necessary to build a Docker image on a Windows machine.

I modified the original Dockerfiles from [kns-it/docker-docfx](/kns-it/docker-docfx) to copy the resulting `target/Release/docfx` folder of the release task instead of downloading the artifact from GitHub to avoid unnecessary script steps. I would prefer to build the `target/Release` directory within the Docker image build process but for now this seems to be more overhead and repetitive work than a "cleaner" solution.

This is still _work in progress_ so any suggestions/proposals/whatever are very welcome!

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5076)


superyyrrzz:
@baez90 Thank you for the PR! The changes look good for me. Just busy on other things, and I will test it in the weekend to ensure I understand it. (I am not so familiar with docker...)

baez90:
If you have any questions or need further explanations, just let me know :wink: i'd be glad to help if I can!

jeffwilcox:
I haven't had a chance to try this yet, but I'm interested in the performance etc.

My team is just starting to experiment with docfx and so are using a 2-stage Azure Pipeline right now... the first stage is just Windows + chocolatey + docfx to build, and the 2nd stage is our typical team Ubuntu process for building an image and then eventually releasing that to our Kubernetes cluster.

I'd love to use this but probably can't, since we actually bundle a TypeScript app runtime on top of the static generated site to provide advanced auth and other capabilities... hmm.

baez90:
@jeffwilcox performance differs between the two versions. The pure Mono version has a shorter startup time in contrary to the ASP.NET Core including version (haven't measured exact numbers but I could benchmark it if anyone wants some details).

If you can provide me a sample project I will try to benchmark the performance of a native **docfx** CLI running on a Windows VM versus running it in a container.

According your concerns about the TypeScript app: it's hard to say if your workflow would work with a container without any details about what you're trying to achieve. But I'd be glad to help if you can tell me more :wink:

superyyrrzz:
I just try building and using the mono image, but does not make it work. I have some questions:
* It needs to upate `.gitattributes` to set these files using `lf`. I need this in my Windows machine, to avoid the wrong line endings going to docker image.
* I still cannot make it work after got the local image. Say I have a https://github.com/docascode/docfx-seed already cloned locally. What command should I run to generate site for this? `docker run -ti --rm  mono-docfx:latest --help` can print out the message, but I still cannot pass files from local machine into container.
* Under what circumstance should user choose net-core image instead of pure mono image?

baez90:
Yes, line endings and file permissions are known problems with Docker. To be honest, I haven't tried to build any Docker image on Windows for at least a year.

According your problem how to build the docs:

1. enter the directory where you cloned the project
2. on Linux I'd run the DocFX build like this: ``docker run --rm -ti -v `pwd`:/work -w /work -u `id -u` mono-docfx build`` which translates (more or less) to `docker run --rm -ti -v "$(Get-Location)":/work -w /work mono-docfx build` in Powershell. I'm not sure how to translate the  `` `id -u` `` on Windows to run the DocFX process as the current user. Probably the best equivalent is to run it like this: `docker run --rm -ti -v "$(Get-Location)":/work -w /work -u root mono-docfx build`.  The `-v` parameter mounts the current working directory to `/work` within the container and the `-w /work` parameter sets the working directory of the container to `/work` to ensure that everying you run in the container is actually happening on your local directory

To get a little bit more convenience on Linux I like to define an alias like the following:
```bash
alias docfx='docker run -ti --rm -p 8080:8080 --name docfx -u `id -u` -v `pwd`:/work -w /work mono-docfx'
```
I haven't tried something similar on Windows but I'm pretty sure it should be possible to add a custom command to the Powershell profile.

Now to your last question: A user wants to chose the ASP.NET Core (including .NET Core) based image if she or he wants to be able to compile the project with the `metadata` task. I just played around with different configurations and for some reason everything worked also without .NET Core SDK present in the container but I'm not sure if I really cleaned everything correctly. I'll have a more thorough look in the next days and will update the PR as soon as I can tell if the .NET Core image is actually required.

superyyrrzz:
Hi, @baez90 I will have a vacation for 2 weeks, so I am not likely to try this recently. I will let you know when I have an update on this. You can also update this PR if you have any new findings. Thanks!

superyyrrzz:
@baez90 I am closing this as releasing Docker image is not on the plan list. Thank you for your contribution!

BTW, I am happy to help if you need support from docfx for your released docker image.

PR #5084, opened 2019-09-10
remove host name for review

live1206:
Host name of `.xrefmap.json` in v3 is always `docs.microsoft.com` even for review site
Host name of `.xrefmap.json` in v2 is `review.docs.microsoft.com` for review site

Apply the workaround to remove the host name, it can be removed when all xref related repos migrated to v3

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5084)

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -137,6 +137,13 @@ namespace Microsoft.Docs.Build
                 return url.Substring(hostName.Length);
             }
 
+            // TODO: this workaround can be removed when all xref related repos migrated to v3
+            if (hostName.Equals("https://docs.microsoft.com", StringComparison.OrdinalIgnoreCase)
+                        && url.StartsWith($"https://review.docs.microsoft.com/", StringComparison.OrdinalIgnoreCase))
+            {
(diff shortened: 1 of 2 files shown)


live1206:
#4450

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -137,6 +137,13 @@ private string RemoveSharingHost(string url, string hostName)
                 return url.Substring(hostName.Length);
             }
 
+            // TODO: this workaround can be removed when all xref related repos migrated to v3
+            if (hostName.Equals("https://docs.microsoft.com", StringComparison.OrdinalIgnoreCase)
+                        && url.StartsWith($"https://review.docs.microsoft.com/", StringComparison.OrdinalIgnoreCase))
+            {
+                return url.Substring("https://review.docs.microsoft.com".Length);
```
yufeih: when url starts with `https://docs.microsoft.com` this substring gives wrong result.

live1206: Do you mean url starts with `https://docs.microsoft.com` and host name is also `https://docs.microsoft.com`?

PR #5089, opened 2019-09-10
allow style attribute in table

OsmondJiang:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/114684?src=WorkItemMention&src-action=artifact_link

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5089)

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -297,20 +297,13 @@ outputs:
   docs/a.json: |
     { "conceptual": "<div></div><p>body</p>" }
 ---
-# table with styles is allowed
+# markdown table with styles is allowed
 inputs:
   docfx.yml:
   docs/a.md: |
     a     | b       | c 
     :-----|:-------:| ----: 
     0     | 1       | 2 
-
-    <th style="text-align: left;">
-    <th style="text-align: center;">
-    <th style="text-align: right;">
-    <td style="text-align: left;">
-    <td style="text-align: center;">
-    <td style="text-align: right;">
 outputs:
   docs/a.json: |
     {
@@ -331,13 +324,6 @@ outputs:
             </tr>
           </tbody>
         </table>
-
-        <th style=\"text-align: left;\"></th>
-        <th style=\"text-align: center;\"></th>
-        <th style=\"text-align: right;\"></th>
-        <td style=\"text-align: left;\"></td>
-        <td style=\"text-align: center;\"></td>
-        <td style=\"text-align: right;\">\n</td>"
     }
 ---
 # do not process auto links
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -323,7 +323,7 @@ outputs:
               <td style=\"text-align: right;\">2</td>
             </tr>
           </tbody>
-        </table>
+        </table>"
     }
 ---
 # do not process auto links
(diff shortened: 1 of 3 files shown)


OsmondJiang:
Why this diff is not found in migration test? Seems v2 always has the style attributes

image


@yufeih  @928PJY

OsmondJiang:
> Why this diff is not found in migration test? Seems v2 always has the style attributes
> 
> image
> 
> @yufeih @928PJY

This need to be fixed in Migration Tool normalizing html

docs/specs/markdown.yml
```diff
@@ -297,6 +297,49 @@ outputs:
   docs/a.json: |
     { "conceptual": "<div></div><p>body</p>" }
 ---
+# table with styles is allowed
+inputs:
+  docfx.yml:
+  docs/a.md: |
```
yufeih: consider put the test in a `HtmlUtilityTest.HtmlStripTags` UT

src/docfx/lib/HtmlUtility.cs
```diff
@@ -15,6 +16,8 @@ internal static class HtmlUtility
         private static readonly Func<HtmlAttribute, int> s_getValueStartIndex =
             ReflectionUtility.CreateInstanceFieldGetter<HtmlAttribute, int>("_valuestartindex");
 
+        private static readonly Regex s_styleCellAlignRegEx = new Regex("text-align: (?:(?:left)|(?:center)|(?:right));", RegexOptions.Compiled);
```
yufeih: handle `style="text-align: left; background-color: yellow"` case

docs/specs/markdown.yml
OsmondJiang: it's the case like spec test `do not allow custom js, styles and css`, so keep it in spec test may be ok?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -15,6 +16,8 @@ internal static class HtmlUtility
         private static readonly Func<HtmlAttribute, int> s_getValueStartIndex =
             ReflectionUtility.CreateInstanceFieldGetter<HtmlAttribute, int>("_valuestartindex");
 
+        private static readonly Regex s_styleCellAlignRegEx = new Regex("^text-align: (?:(?:left)|(?:center)|(?:right));$", RegexOptions.Compiled);
```
yufeih: Looks like this regex means it must be exactly `text-align: left;` or `text-align: right;` or `text-align: center;`, so may be a `Contains` is enough.

On the other hand, `text-align:right` is also a valid style, but I don't think we need to go that far.

docs/specs/markdown.yml
yufeih: Yeah, the table is fine, just those other tests

test/docfx.Test/lib/HtmlUtilityTest.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using Markdig.Syntax.Inlines;
```
yufeih: remove

src/docfx/lib/HtmlUtility.cs
OsmondJiang: `Contains` or `Equals`?

yufeih: I mean `Array.Contains`, you have an array of `['text-align: left', 'text-align: right', etc]`

PR #5098, opened 2019-09-12
Refine url merge

live1206:
use `UrlUtility.MergeUrl` instead of string concating

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5098)

### DIFF
--- a/src/docfx/lib/UrlUtility.cs
+++ b/src/docfx/lib/UrlUtility.cs
@@ -37,11 +37,11 @@ namespace Microsoft.Docs.Build
             var fragmentIndex = url.IndexOf('#');
             if (fragmentIndex >= 0)
             {
-                fragment = url.Substring(fragmentIndex + 1);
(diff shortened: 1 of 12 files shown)


src/docfx/lib/UrlUtility.cs
```diff
@@ -159,8 +162,14 @@ public static string GetRelativeUrl(string relativeToUrl, string url)
                 result.Append(path, segmentIndex, path.Length - segmentIndex);
             }
 
-            result.Append(query);
-            result.Append(fragment);
+            if (!string.IsNullOrEmpty(query))
```
yufeih: what's the difference? `query` and `fragment` should never be null

live1206: There was 1 commit `query` and `fragment` could be null, they cannot now.

PR #5092, opened 2019-09-11
Add test to verify git url and contributor info for CRR build content

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/118459

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5092)

PR #5104, opened 2019-09-16
refine git contribution for CRR content

live1206:
refine #5092 
Set `original_content_git_url_template` the same value as `original_content_git_url` to avoid the fallback
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5104)


OsmondJiang:
will fallback cause problem?

live1206:
> will fallback cause problem?

no, it is just more convenient to not fall back.

PR #5108, opened 2019-09-17
try use file share and access to implement reader and writer lock

OsmondJiang:
refine reader and writer lock using file access and file share

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5108)


OsmondJiang:
> wrap `ReleaseMutex` in `finally` block

Just did :)

src/docfx/lib/process/InterProcessReaderWriterLock.cs
```diff
@@ -0,0 +1,108 @@
...
+            if (_fileStream != null)
+            {
+                using (new GlobalMutex(_filePath))
+                {
+                    _fileStream.Dispose();
+                }
+            }
+        }
+
+        private class GlobalMutex : IDisposable
```
yufeih:
```suggestion
        private struct GlobalMutex : IDisposable
```

src/docfx/lib/process/InterProcessReaderWriterLock.cs
```diff
@@ -0,0 +1,108 @@
...
+            var filePath = Path.Combine(AppData.MutexRoot, HashUtility.GetMd5Hash(lockName));
+            var fileLock = WaitFile(lockName, filePath, FileAccess.Write, FileShare.None);
+
+            return new InterProcessReaderWriterLock(fileLock, filePath);
+        }
+
+        private static FileStream WaitFile(string name, string path, FileAccess access, FileShare fileShare)
+        {
+            var start = DateTime.UtcNow;
+            while (DateTime.UtcNow - start <= TimeSpan.FromMinutes(1))
```
yufeih: suggest to repeat forever until we grab the lock, similar to `InterProcessMutex`

src/docfx/lib/process/InterProcessReaderWriterLock.cs
```diff
@@ -0,0 +1,108 @@
...
+            var fileLock = WaitFile(lockName, filePath, FileAccess.Read, FileShare.Read);
+
+            return new InterProcessReaderWriterLock(fileLock, filePath);
+        }
+
+        public static InterProcessReaderWriterLock CreateWriterLock(string lockName)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(lockName));
+
+            var filePath = Path.Combine(AppData.MutexRoot, HashUtility.GetMd5Hash(lockName));
```
live1206: There is already a hash operation in `GlobalMutex`, do we need a nested hash?

OsmondJiang: image

OsmondJiang: lock name can be remote url

yufeih: `HashUtility.GetMd5Hash(lockName)` can be shared though

PR #5113, opened 2019-09-18
Handle AbandonedMutexException

yufeih:
- Catch `AbandonedMutexException`
- Calculate MD5 hash once

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5113)

### DIFF
--- a/src/docfx/lib/process/InterProcessMutex.cs
+++ b/src/docfx/lib/process/InterProcessMutex.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Docs.Build
             }
             catch (AbandonedMutexException)
             {
-                // When another process/thread exited without relasing its mutex,
+                // When another process/thread exited without releasing its mutex,
                 // this exception is thrown and we've successfully acquired the mutex.
             }
 
--- a/src/docfx/lib/process/InterProcessMutex.cs
+++ b/src/docfx/lib/process/InterProcessMutex.cs
@@ -26,16 +26,24 @@ namespace Microsoft.Docs.Build
 
             var mutex = new Mutex(initiallyOwned: false, $"Global\\{HashUtility.GetMd5Hash(mutexName)}");
 
-            while (!mutex.WaitOne(TimeSpan.FromSeconds(30)))
+            try
             {
-#pragma warning disable CA2002 // Do not lock on objects with weak identity
-                lock (Console.Out)
-#pragma warning restore CA2002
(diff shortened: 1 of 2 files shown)


src/docfx/lib/process/InterProcessMutex.cs
```diff
@@ -26,17 +26,25 @@ public static InterProcessMutex Create(string mutexName)
...
-                    Console.WriteLine($"Waiting for another process to access '{mutexName}'");
-                    Console.ResetColor();
+                    {
+                        Console.ForegroundColor = ConsoleColor.Yellow;
+                        Console.WriteLine($"Waiting for another process to access '{mutexName}'");
+                        Console.ResetColor();
+                    }
                 }
             }
+            catch (AbandonedMutexException)
```
live1206: This would only throw when the other thread exited without releasing its mutex. Why would it happen in the first place or did it happen during the testing?

src/docfx/lib/process/InterProcessMutex.cs
```diff
@@ -26,17 +26,25 @@ public static InterProcessMutex Create(string mutexName)
...
+            {
+                // When another process/thread exited without relasing its mutex,
```
live1206:
```suggestion
                // When another process/thread exited without releasing its mutex,
```

src/docfx/lib/process/InterProcessReaderWriterLock.cs
```diff
@@ -76,32 +77,35 @@ private static FileStream WaitFile(string name, string path, FileAccess access,
...
             }
 
             public void Dispose()
             {
-                if (_mutex != null)
-                {
-                    _mutex.ReleaseMutex();
-                    _mutex.Dispose();
-                }
+                _mutex.ReleaseMutex();
```
live1206: what if it throws during instancing a new `Mutex`?

src/docfx/lib/process/InterProcessMutex.cs
yufeih: Like a process crash while holding a lock

src/docfx/lib/process/InterProcessReaderWriterLock.cs
yufeih: AbandonedMutex exception only throws in `WaitOne` method

PR #5114, opened 2019-09-18
Add content validation extension

LowEntropyBody:
Work item: https://ceapex.visualstudio.com/Engineering/_workitems/edit/118012

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5114)


src/docfx/lib/markdown/validation/ContentValidationContext.cs
```diff
@@ -0,0 +1,52 @@
...
+namespace Microsoft.Docs.Build
+{
+    internal class ContentValidationContext : IValidationContext
+    {
+        private const string ValidationApiBaseUrl = "validationApiBaseUrl";
+        private const string DefaultValidationApiBaseUrl = "https://docs.microsoft.com";
+        private const string ValidationRequestTimeoutInSeconds = "validationRequestTimeoutInSeconds";
+        private const int DefaultValidationRequestTimeoutInSeconds = 10;
+        private const string ValidationMaxTries = "validationMaxTries";
+        private const int DefaultValidationMaxTries = 4;
```
kimizhu: Why retry 4 times? Usually this number is 3

src/docfx/lib/markdown/validation/ContentValidationContext.cs
```diff
@@ -0,0 +1,52 @@
...
+        public int RetryDelaySeconds { get; set; }
+
+        public int MaxTransientFailureCount { get; set; }
+
+        public ContentValidationContext(Config config)
+        {
+            MarkdownRulesFilePath = config.ContentValidationRulesFilePath;
+            RepositoryUrl = EnvironmentVariable.RepositoryUrl;
+            Branch = EnvironmentVariable.RepositoryBranch;
+            ApiBase = config.GlobalMetadata.TryGetValue(ValidationApiBaseUrl, out var apiBase) ? apiBase.ToObject<string>() : DefaultValidationApiBaseUrl;
```
yufeih: _global metadata_ is for user configuration, these are systems settings, consider hard code it here.

src/docfx/config/Config.cs
```diff
@@ -179,6 +179,11 @@ internal sealed class Config
         /// </summary>
         public readonly SourceInfo<string> MonikerDefinition = new SourceInfo<string>(string.Empty);
 
+        /// <summary>
+        /// Get the file path of content validation rules
+        /// </summary>
+        public readonly SourceInfo<string> ContentValidationRulesFilePath = new SourceInfo<string>(string.Empty);
```
yufeih: add this to `RestoreFile.GetFileReferences`

yufeih:
```suggestion
        public readonly SourceInfo<string> MarkdownValidationRules = new SourceInfo<string>(string.Empty);
```

src/docfx/lib/markdown/validation/ContentValidationExtension.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public static MarkdownPipelineBuilder UseContentValidation(this MarkdownPipelineBuilder builder, MarkdownContext markdownContext, Config config)
+        {
+            var validatorProvider = new OnlineServiceMarkdownValidatorProvider(
+                new ContentValidationContext(config),
+                new ContentValidationLogger(markdownContext));
+
+            return builder.Use(document =>
+            {
+                foreach (var validator in validatorProvider.GetValidators())
```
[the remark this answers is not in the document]
LowEntropyBody: No, we will download rules in docs build, this function won't call any api

src/docfx/lib/markdown/validation/ContentValidationContext.cs
LowEntropyBody: same as v2

LowEntropyBody: In addition, all config parameters are same as v2.

src/docfx/lib/markdown/validation/ContentValidationContext.cs
LowEntropyBody: make sense

src/docfx/lib/markdown/validation/ContentValidationContext.cs
```diff
@@ -0,0 +1,47 @@
...
+
+        public TimeSpan RequestTimeout { get; set; }
+
+        public int MaxTries { get; set; }
+
+        public int RetryDelaySeconds { get; set; }
+
+        public int MaxTransientFailureCount { get; set; }
+
+        public ContentValidationContext(Config config)
```
yufeih:
```suggestion
        public ContentValidationContext(string markdownValidationRules)
```

src/docfx/lib/markdown/MarkdownUtility.cs
```diff
@@ -143,14 +145,19 @@ private static MarkdownPipeline CreateTocMarkdownPipeline()
             return builder.Build();
         }
 
+        private static Config GetConfig()
+        {
+            return t_status.Value.Peek().Context.Config;
```
yufeih: to get the actual path of rule file, use `context.RestoreFileMap.GetRestoredFilePath`

LowEntropyBody: sure

yufeih:
@LowEntropyBody , can you add a basic spec test to illustrate and cover docfx end to end scenario?

src/docfx/lib/markdown/validation/ContentValidationExtension.cs
```diff
@@ -1,23 +1,39 @@
...
+        private static object locker = new object();
 
+        public static MarkdownPipelineBuilder UseContentValidation(this MarkdownPipelineBuilder builder, MarkdownContext markdownContext, Func<string> getMarkdownValidationRulesPath)
+        {
             return builder.Use(document =>
             {
-                foreach (var validator in validatorProvider.GetValidators())
+                if (_validators.IsDefaultOrEmpty)
+                {
+                    lock (locker)
```
yufeih: 😱 😱 😱 😱 😱 😱 😱 can you fix the thread bug instead

LowEntropyBody: not the package bug...

LowEntropyBody: the locker is used to avoid repeated initialization...

yufeih: Duplicated initialization can be solved by using `Lazy<T>`

src/docfx/lib/markdown/validation/ContentValidationContext.cs
superyyrrzz: This value is different in PPE and PROD. Instead of hard code, we need a way to pass system settings.

yufeih: @LowEntropyBody , is this URL for metadata validation only? markdown validation should never call external api, thus how about leave this setting and other network call related settings to default so it does not confuse people?

src/docfx/lib/markdown/validation/ContentValidationExtension.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    internal static class ContentValidationExtension
+    {
+        private static ImmutableArray<IMarkdownObjectValidator> _validators;
+        private static object locker = new object();
+
+        public static MarkdownPipelineBuilder UseContentValidation(this MarkdownPipelineBuilder builder, MarkdownContext markdownContext, Func<string> getMarkdownValidationRulesPath)
+        {
```
yufeih: init validators here (don't use static)

yufeih:
Can you do a performance testing locally on `azure-docs-pr` and paste the time difference between markdown validation on and off? I would like to see the numbers before merge it.

PR #5116, opened 2019-09-18
no checkout for restored fallback repository

OsmondJiang:
Fallback repo now is not needed to be checked out.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5116)


src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -47,10 +47,16 @@ internal class ExternalXrefMapLoader
             return result;
         }
 
-        public static List<(string, Lazy<ExternalXrefSpec>)> Load(string filePath)
+        public static List<(string, Lazy<ExternalXrefSpec>)> Load(Input input, FilePath filePath)
         {
             var result = new List<(string, Lazy<ExternalXrefSpec>)>();
-            var content = File.ReadAllBytes(filePath);
+            byte[] content;
```
yufeih: this is a special optimization if we can read the file directory.

In line 36, try use call the previous `Load` method if the input is a physical file, and use `JsonUtility.Deserialize` otherwise.

src/docfx/build/context/Input.cs
```diff
@@ -28,6 +28,11 @@ public Input(string docsetPath, RepositoryProvider repositoryProvider)
         /// </summary>
         public bool Exists(FilePath file)
         {
+            if (Path.IsPathRooted(file.Path))
```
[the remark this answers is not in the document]
OsmondJiang: it can't, resolve file path is only handling relative path case, check the return results

src/docfx/build/xref/ExternalXrefMapLoader.cs
live1206: agreed

yufeih:
Close stale pull requests

PR #5117, opened 2019-09-18
Generate interactive_type

live1206:
Extract logic from docs-ui:
https://ceapex.visualstudio.com/Engineering/_git/docs-ui?path=%2Fpackages%2Fdocfx-templates%2Fchrome.common.js&version=GBdevelop&line=393&lineStyle=plain&lineEnd=394&lineStartColumn=1&lineEndColumn=1
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5117)

### DIFF
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -17,6 +17,9 @@ repos:
         ms.author: yufeih
         ---
         [d](d.png)
+        ```azurepowershell-interactive
+          New-AzResourceGroup -Name myResourceGroup -Location westeurope
+        ```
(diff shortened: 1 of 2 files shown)


live1206:
The diff is expected, the content is the same, just the order of properties. And there is no need to sort the properties since the output is stable.

live1206:
we decided to keep the logic in template code since it is not a general feature for docfx

PR #5101, opened 2019-09-16
monikers of toc file should be the aggregation of monikers of children

928PJY:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/120652
1. the monikers to toc file is the aggregation of monikers of children
2. the monikers of absolute link/external link/non-versioning page node in toc file should be empty.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5101)

PR #5119, opened 2019-09-19
monikers of external link/absolute link/non-versioning page node should be null

928PJY:
separated from #5101 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5119)

### DIFF
--- a/src/docfx/build/toc/parser/TableOfContentsParser.cs
+++ b/src/docfx/build/toc/parser/TableOfContentsParser.cs
@@ -160,7 +160,7 @@ namespace Microsoft.Docs.Build
                 }
 
                 // resolve monikers
-                newItem.Monikers = GetMonikers(context, rootPath, newItem, errors);
+                newItem.Monikers = GetMonikers(context, newItem, errors);
                 newItems.Add(newItem);
 
                 // validate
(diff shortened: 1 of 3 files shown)


src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -7,6 +7,7 @@
 using System.IO;
 using System.Linq;
 using System.Threading;
+using Microsoft.Graph;
```
yufeih:
```suggestion
~~~using Microsoft.Graph;~~~
```

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -179,19 +177,12 @@ private static (List<Error>, TableOfContentsModel) LoadTocModel(JToken tocToken)
         private static List<string> GetMonikers(Context context, Document rootPath, TableOfContentsItem currentItem, List<Error> errors)
```
live1206:
```suggestion
        private static List<string> GetMonikers(Context context, TableOfContentsItem currentItem, List<Error> errors)
```

PR #5121, opened 2019-09-20
support '#' in PackageUrl.url

mingwli:
[DevOps#62243](https://dev.azure.com/ceapex/Engineering/_workitems/edit/62243)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5121)

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -585,3 +585,23 @@ outputs:
     }
   docs/path1.json: |
     {"document_id": "780484e9-a367-97ab-ba7d-a19f3b378a80", "document_version_independent_id": "e9b3f4e2-1a78-4b5d-fbb4-03e0a33e38f5",}
+---
+# dependencyPackageUrl's branch defined in url has higher priority
+repos:
+  https://docs.com/branch-priority/1:
+    - files:
+        docfx.yml: |
+          dependencies:
+            crr:
+              url: https://docs.com/branch-priority/2#master
+              branch: unused
+              buildFiles: true
+  https://docs.com/branch-priority/2#master:
+    - files:
+        b.md: used
+  https://docs.com/branch-priority/2#unused:
+    - files:
+        c.md: unused
+outputs:
+  crr/b.json: |
+    {"conceptual": "<p>used</p>"}
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -605,3 +605,42 @@ repos:
 outputs:
   crr/b.json: |
     {"conceptual": "<p>used</p>"}
+---
+# [skip] dependencyPackageUrl can use relative path by setting path or url
+# path has higher priority than url
+# todo: un-skip after handle dependency is relative folder case
+inputs:
+  docfx.yml: |
+    files:
+    - crr1/**/*.md
+    - crr2/**/*.md
+    dependencies:
+      crr1:
+        path: crr-by-path
+        includeInBuild: true
+      crr2:
+        url: crr-priority-url
+        path: crr-priority-path
+        includeInBuild: true
+  crr-by-path/crr-by-path.md:
+  crr-priority-url/crr-priority-url.md:
+  crr-priority-path/crr-priority-path.md:
+outputs:
+  crr1/crr-by-path.json:
+  crr2/crr-priority-path.json:
+---
+# dependencyPackageUrl's path using git url won't restore the git repo
+repos:
+  https://docs.com/crr-url-using-path/1:
+  - files:
+      docfx.yml: |
+        dependencies:
+          crr:
+            path: https://docs.com/crr-url-using-path/2
+            includeInBuild: true
+  https://docs.com/crr-url-using-path/2:
+  - files:
+      b.md:
+outputs:
+  .errors.log: |
+    ["error","need-restore","Cannot find dependency 'https://docs.com/crr-url-using-path/2', did you forget to run `docfx restore`?"]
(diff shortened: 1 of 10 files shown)


src/docfx/restore/RestoreGitMap.cs
```diff
@@ -67,7 +67,7 @@ private RestoreGitMap(DependencyLockProvider dependencyLockProvider, string docs
                     return (path, gitLock.Commit);
 
                 default:
-                    throw new NotSupportedException($"Unknown package url: '{packageUrl}'");
+                    throw new NotSupportedException($"Unknown package url: '{packageUrl.Url}', type: {packageUrl.Type}");
```
yufeih: consider modify `PackageUrl.ToString` method

docs/specs/config.yml
```diff
@@ -585,3 +585,23 @@ outputs:
...
     {"document_id": "780484e9-a367-97ab-ba7d-a19f3b378a80", "document_version_independent_id": "e9b3f4e2-1a78-4b5d-fbb4-03e0a33e38f5",}
+---
+# dependencyPackageUrl's branch defined in url has higher priority
+repos:
+  https://docs.com/branch-priority/1:
+    - files:
+        docfx.yml: |
+          dependencies:
+            crr:
+              url: https://docs.com/branch-priority/2#master
```
yufeih: Do we want to throw here for expanded form? as the `url` here means repository url.

mingwli: Still it feels that:
```json
{
    "url": "https://github.com/owner/repo#branch",
    "includeInBuild": true
}
```
comes more convenient

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -77,9 +72,25 @@ private static (string remote, string refspec) SplitGitUrl(string remoteUrl)
...
+            {
+                Type = PackageType.Git;
+                var (url, branch) = SplitGitUrl(Url);
+                Url = url;
+                Branch = branch ?? Branch;
+            }
+            else
+            {
+                Type = PackageType.Folder;
+                Path = Url;
```
OsmondJiang: it has problem when we just use `{Path: xxx}`

OsmondJiang: add a test for that

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -77,9 +72,25 @@ private static (string remote, string refspec) SplitGitUrl(string remoteUrl)
...
+
+        [OnDeserialized]
+        private void OnDeserialized(StreamingContext context)
+        {
+            if (UrlUtility.IsHttp(Url))
+            {
+                Type = PackageType.Git;
+                var (url, branch) = SplitGitUrl(Url);
+                Url = url;
+                Branch = branch ?? Branch;
```
OsmondJiang: it's a little inconsistent, the user specified `Branch` should has higher priority than branch in url?

src/docfx/lib/path/PackageUrl.cs
```diff
@@ -31,16 +34,8 @@ public PackageUrl()
...
-                Type = PackageType.Git;
-                (Url, Branch) = SplitGitUrl(url);
-            }
-            else
-            {
-                Type = PackageType.Folder;
-                Path = url;
-            }
+            Url = url;
+            OnDeserialized(s_streamingContext);
```
yufeih: if populate calls OnDeserialized, is it still necessary to call here?

test/docfx.Test/restore/RestoreTest.cs
```diff
@@ -4,33 +4,42 @@
...
+        [InlineData("https://visualstudio.com/dotnet/docfx", PackageType.Git, "https://visualstudio.com/dotnet/docfx", "master", null)]
+        [InlineData("https://github.com/dotnet/docfx#master", PackageType.Git, "https://github.com/dotnet/docfx", "master", null)]
+        [InlineData("https://github.com/dotnet/docfx#live", PackageType.Git, "https://github.com/dotnet/docfx", "live", null)]
+        [InlineData("https://github.com/dotnet/docfx#", PackageType.Git, "https://github.com/dotnet/docfx", "master", null)]
+        [InlineData("https://github.com/dotnet/docfx#986127a", PackageType.Git, "https://github.com/dotnet/docfx", "986127a", null)]
+        [InlineData("https://github.com/dotnet/docfx#a#a", PackageType.Git, "https://github.com/dotnet/docfx", "a#a", null)]
+        [InlineData("https://github.com/dotnet/docfx#a\\b/d<e>f*h|i%3C", PackageType.Git, "https://github.com/dotnet/docfx", "a\\b/d<e>f*h|i%3C", null)]
+        [InlineData("crr/local-path", PackageType.Folder, "crr/local-path", "master", "crr/local-path")]
+        public static void PackageUrlTest(
+            string remote,
```
yufeih: consider test json deserialize in this method

src/docfx/lib/path/PackageUrl.cs
mingwli: This one is called by  
image
which share the same logic with OnDeserialized. And OnDeserialized needs to be manually invoked here.

yufeih: I see, streaming context could then be null

mingwli: I would very much like to. But it's a `struct` 😂.

mingwli: `StreamingContext.Context` is always null in both case

PR #5129, opened 2019-09-24
fix bug of copying resource from crr

OsmondJiang:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/124081

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5129)

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -313,14 +313,14 @@ repos:
   https://docs.com/crrcontent/b:
     - files:
         b.md:
-        image.png:
+        image.png: xxx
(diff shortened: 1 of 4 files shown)


docs/specs/restore.yml
```diff
@@ -313,12 +313,14 @@ repos:
   https://docs.com/crrcontent/b:
     - files:
         b.md:
+        image.png:
```
yufeih: add some text content here to verify that `image.png` is actually copied.

PR #5134, opened 2019-09-24
Removed an inaccurate warning for image-end

adunndevster:
The warning in this PR was inaccurate, so I made a small adjustment to the logging.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5134)


adunndevster:
@yufeih - I have a question... Where can I see the warning output from my local dev environment? I didn't think to go look for warning logs during testing, and it would help me a lot in the future :).

yufeih:
Writing a test is the simplest way for me to verify markdown behavior. We have a workitem to improve markdown testing to: a) make it easier to write markdown test and errors, b) share tests with v2/v3.

@superyyrrzz to see if he has some ideas around testing markdown extensions locally.

For this change, is there a test case to illustrate the scenario?

adunndevster:
@yufeih - That's the funny thing. The actual output of the markdown was great! It's just that it output a warning it shouldn't have. I missed that warning because it wasn't readily visible to me... I'm sure I could find it somewhere if I looked, and watch for it in the future?

yufeih:
@adunndevster with #5140, diagnostics output become a mandatory test aspect. If you merged the latest dev branch, a test will fail with your change (a warning disappeared).

adunndevster:
@yufeih - That's a fantastic upgrade! I merged the latest from dev, and checked out the tests. You were correct, a couple failed, but only because they were attempting to test valid image extensions as being invalid. I've corrected the tests and added a couple of new one's to cover the errors. 

All the tests run now.

PR #5136, opened 2019-09-25
update versioning-related error code

928PJY:
Update versioning-related error code.
- **publish-url-conflict**
  - Report a error when there are multiple files publish to same url, some with version, some without
  - report with more detail versioning info
- **moniker-overlapping**
  - report with more detail message
  - promote from `warning` to `error`
- **moniker-config-missing**
  - rename to `moniker-range-undefined`
  - promote from `warning` to `error`
- **empty-monikers**
  - rename to `moniker-range-out-of-scope`
  - promote from `warning` to `error`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5136)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -287,7 +287,7 @@ namespace Microsoft.Docs.Build
         /// Behavior: ✔️ Message: ❌
         public static Error PublishUrlConflict(string url, IReadOnlyDictionary<Document, List<string>> files, IEnumerable<string> conflictMonikers)
         {
-            var nonVersion = conflictMonikers.Contains("NONE_VERSION");
+            var nonVersion = conflictMonikers.Contains(PublishModelBuilder.NONE_VERSIONING);
             var message = conflictMonikers.Any() && !nonVersion ? $" of the same version({Join(conflictMonikers)})" : null;
             return new Error(
                 ErrorLevel.Error,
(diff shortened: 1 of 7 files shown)


src/docfx/Errors.cs
```diff
@@ -566,18 +570,18 @@ public static Error MonikerRangeInvalid(SourceInfo<string> source, Exception ex)
         /// </summary>
         /// Behavior: ✔️ Message: ❌
         public static Error MonikerConfigMissing(SourceInfo<string> source)
```
```suggestion
        public static Error MonikerRangeUndefined(SourceInfo<string> source)
```

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -43,7 +43,6 @@ public bool TryAdd(Document file, PublishItem item)
             var monikers = item.Monikers;
             if (monikers.Count == 0)
             {
-                // TODO: report a warning if there are multiple files published to same url, one of them have no version
                 monikers = new List<string> { "NONE_VERSION" };
```
yufeih: suggest to replace this magic string `NONE_VERSION` with a boolean flag indicating whether this file has versions

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -63,10 +61,10 @@ public bool TryAdd(Document file, PublishItem item)
                     .Where(group => group.Count() > 1)
                     .Select(group => group.Key);
 
-                if (conflictMoniker.Count() > 0)
+                if (conflictMoniker.Any()
+                    || (files.Count() > 1 && files.Any(file => file.Value.Contains("NONE_VERSION"))))
```
live1206: put "NONE_VERSION" as a private constant field to avoid duplication?

PR #5137, opened 2019-09-25
Add local build spec

yufeih:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/117454/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5137)


docs/designs/local-build.md
```diff
@@ -0,0 +1,188 @@
...
+`docfx` is the entry point to build a repository locally, it calls `docs build service` to retrieve docset specific configuration.
+
+![](./images/local-build-cli-build.png)
+
+[Link to Graph](https://mermaidjs.github.io/mermaid-live-editor/#/view/eyJjb2RlIjoic2VxdWVuY2VEaWFncmFtXG4gICAgZG9jZnggLT4-IGRvY3MgYnVpbGQgc2VydmljZTogR0VUIC9jb25maWcvZG9jZnhcbiAgICBkb2NzIGJ1aWxkIHNlcnZpY2UgLS0-PiBkb2NmeDogSlNPTiBDb25maWdcblxuXG4iLCJtZXJtYWlkIjp7InRoZW1lIjoiZGVmYXVsdCJ9fQ)
+
+#### Server Command Line Build
+
+Server build uses the exact same way as local command line build, the only exceptions are:
+- Server build runs `docfx` as a service account while locale build runs `docfx` as your personal Microsoft account.
```
mingwli: local

docs/designs/local-build.md
```diff
@@ -0,0 +1,142 @@
...
+
+### Trigger Build
+
+Handling build command is divided into two phases given the time constraint.
+
+#### Phase 1
+
+For the initial phase, the vscode extension calls `docfx build` and `docfx restore` directly, parses generated report files and presents them to problems window. It pipelines `docfx` console output to a specific output channel.
+
+#### Phase 2
```
928PJY: >Phase 2 [](start = 5, length = 7)

Here is the VS Code description about Language server:
> Language Server is a special kind of Visual Studio Code extension that powers the editing experience for many programming languages. With Language Servers, you can implement autocomplete, error-checking (diagnostics), jump-to-definition, and many other [language features](https://code.visualstudio.com/api/language-extensions/programmatic-language-features) supported in VS Code

Seem the `build`, `authentication`, `restore` is not a kind of Editing experience of a programming language, and vscode-language-client doesn't provide native support for those features.

on another hand, handling those request may block the language server from responding the real editing experience request.

maybe we need to reconsider whether it is necessary to consolidate all the feature onto the same pipeline using language server.

yufeih: There are two layers, the __transport layer__ (JSON RPC) and the __protocol layer__ (LSP on top of JSON PRC). We are building interactions between the server and the client using the same __transport layer__, leveraging the _LSP protocol_ as much as we can, building custom protocols as needed.

928PJY: so we are in the same line about the `build`/`restore`/'auth' is not a language server provided feature, right?
talk about leveraging the LSP to achieve those scenarios in the docfx serve mode, how can we resolve the second problem?

---
In reply to: [343428643](https://github.com/dotnet/docfx/pull/5137#discussion_r343428643) [](ancestors = 343428643)

yufeih: For the blocking problem,  `JSON RPC` can handle concurrent requests/notifications, one API does not block other APIs. Also, long operations can be modeled as async API, e.g.,`build` triggers a build and immediately returned with a `build_id`, then you can correlate progress and diagnostics associated with that `build_id`.

PR #5138, opened 2019-09-25
fix bug of file map for crr content

OsmondJiang:
docfx use `FilePath` with "path" and "origin" to represent a file localization.

https://ceapex.visualstudio.com/Engineering/_workitems/edit/124152

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5138)

### DIFF
--- a/src/docfx/build/dependency/DependencyMap.cs
+++ b/src/docfx/build/dependency/DependencyMap.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Docs.Build
                 .ToDictionary(
                     d => GetSourcePath(d.Key.FilePath),
                     d => (from v in d.Value
-                          orderby v.To.FilePath.Path, v.Type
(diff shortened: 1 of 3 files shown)


src/docfx/build/dependency/DependencyMap.cs
```diff
@@ -20,14 +21,17 @@ public object ToDependencyMapModel()
         {
             // TODO: Make dependency map a data model once we remove legacy.
             var dependencies = this
-                .OrderBy(d => d.Key.FilePath.Path)
+                .OrderBy(d => GetSourcePath(d.Key.FilePath))
```
yufeih: Can this be computed upfront? calling a method inside `OrderBy` slows things down exponentially

live1206: could this be a property of `Document`, like `SiteUrl`?

OsmondJiang:
The diff are expected

PR #5142, opened 2019-09-25
Path in FilePath represents `path to main docset`

OsmondJiang:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5142)

### DIFF
--- a/docs/specs/restore.yml
+++ b/docs/specs/restore.yml
@@ -16,7 +16,7 @@ inputs:
     [!include[include to 2](~/docfx-test-dependencies2/docs/a.md)]
 outputs:
   .errors.log: |
-    ["warning","link-out-of-scope","File 'docs/a.md [docfx-test-dependencies1]' referenced by link '~/docfx-test-dependencies1/docs/a.md' will not be built because it is not included in build scope","docs/a.md",1,1]
+    ["warning","link-out-of-scope","File 'docfx-test-dependencies1/docs/a.md [docfx-test-dependencies1]' referenced by link '~/docfx-test-dependencies1/docs/a.md' will not be built because it is not included in build scope","docs/a.md",1,1]
   docs/a.json: |
     {"conceptual":"<p><a href=\"%7E/docfx-test-dependencies1/docs/a.md\">link to 1</a></p>\n<p>a1</p>\n"}
 ---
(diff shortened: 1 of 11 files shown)


src/docfx/lib/path/FilePath.cs
```diff
@@ -31,6 +31,12 @@ internal class FilePath : IEquatable<FilePath>, IComparable<FilePath>
         /// </summary>
         public string Commit { get; }
 
+        /// <summary>
+        /// Gets the path relative to docset root or dependency docset root
+        /// </summary>
+        public string PathWithoutDependencyName
```
yufeih:
```suggestion
        public string GetPathToDocset() => origin == dependency ? GetRelativePath : Path
```

docs/specs/restore.yml
```diff
@@ -16,7 +16,7 @@ inputs:
     [!include[include to 2](~/docfx-test-dependencies2/docs/a.md)]
 outputs:
   .errors.log: |
-    ["warning","link-out-of-scope","File 'docs/a.md [docfx-test-dependencies1]' referenced by link '~/docfx-test-dependencies1/docs/a.md' will not be built because it is not included in build scope","docs/a.md",1,1]
+    ["warning","link-out-of-scope","File 'docfx-test-dependencies1/docs/a.md [docfx-test-dependencies1]' referenced by link '~/docfx-test-dependencies1/docs/a.md' will not be built because it is not included in build scope","docs/a.md",1,1]
```
live1206: do we still need `[docfx-test-dependencies1]` since we have it in the file path now?

src/docfx/lib/path/FilePath.cs
```diff
@@ -48,11 +48,25 @@ public FilePath(string path, string commit, FileOrigin origin)
...
-            Path = PathUtility.NormalizeFile(path);
+            Path = PathUtility.NormalizeFile(System.IO.Path.Combine(dependencyName, path));
             DependencyName = dependencyName;
             Origin = FileOrigin.Dependency;
         }
 
+        /// <summary>
+        /// Gets the path relative to docset root or dependency docset root
+        /// </summary>
+        public string GetPathToOrigin()
```
mingwli: what about a property? e.g.
```code
public string PathToOrigin => 
         Origin == FileOrigin.Dependency 
         ? PathUtility.NormalizeFile(System.IO.Path.GetRelativePath(DependencyName, Path))
         : Path;
```

OsmondJiang: I took Yufei's suggestion to change it to a method, avoid Serialization for unused properties.

OsmondJiang:
diff is expected.

PR #5146, opened 2019-09-26
no need to prepend dependency name to path when globing build scope

mingwli:
[DevOps#124841](https://ceapex.visualstudio.com/Engineering/_workitems/edit/124841)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5146)

### DIFF
--- a/src/docfx/build/context/BuildScope.cs
+++ b/src/docfx/build/context/BuildScope.cs
@@ -95,8 +95,7 @@ namespace Microsoft.Docs.Build
 
                 ParallelUtility.ForEach(fileNames, file =>
                 {
-                    var path = Path.Combine(dependencyName ?? "", file.Path).Replace("\\", "/");
-                    if (glob(path))
+                    if (glob(file.Path))
                     {
                         files.Add(Document.Create(docset, file, _input, _templateEngine));
                     }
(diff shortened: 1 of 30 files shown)


src/docfx/lib/path/FilePath.cs
```diff
@@ -104,7 +104,7 @@ public override bool Equals(object obj)
 
         public override int GetHashCode()
         {
-            return HashCode.Combine(PathUtility.PathComparer.GetHashCode(Path), Origin, DependencyName, Commit);
```
OsmondJiang: do we need change compare?

src/docfx/build/context/BuildScope.cs
```diff
@@ -95,8 +95,7 @@ public bool GetActualFileName(string fileName, out string actualFileName)
 
                 ParallelUtility.ForEach(fileNames, file =>
                 {
-                    var path = Path.Combine(dependencyName ?? "", file.Path).Replace("\\", "/");
```
OsmondJiang: ooooops, my mistake

OsmondJiang: forget to delete

mingwli:
> Is there a test to check if metadata are applied correctly to files in CRR repos using the correct path?

was added in this pr [#5142](https://github.com/dotnet/docfx/pull/5142/files#diff-f5af9eb28100acb0f4e35f1a92fafc22R354)

PR #5147, opened 2019-09-26
use IReadOnlyCollection for monikers output

928PJY:
Fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/124851

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5147)

### DIFF
--- a/src/docfx/build/toc/parser/TableOfContentsParser.cs
+++ b/src/docfx/build/toc/parser/TableOfContentsParser.cs
@@ -174,7 +174,7 @@ namespace Microsoft.Docs.Build
             return (errors, newItems);
         }
 
-        private static IReadOnlyCollection<string> GetMonikers(Context context, TableOfContentsItem currentItem, List<Error> errors)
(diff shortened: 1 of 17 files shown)


live1206:
Is it possible to have a test to demonstrate the fix?

src/docfx/build/toc/parser/TableOfContentsParser.cs
```diff
@@ -205,15 +205,15 @@ private static List<string> GetMonikers(Context context, TableOfContentsItem cur
             {
                 foreach (var item in currentItem.Items)
                 {
-                    if (item.Monikers?.Count == 0)
+                    if (!item.Monikers.Any())
```
live1206: https://stackoverflow.com/questions/305092/which-method-performs-better-any-vs-count-0
For `List`, it is better to use `Count` since it is a prepared property. For `IEnumerable`, it would be better to use `Any()` other than `Count()`.

PR #5153, opened 2019-09-27
Implement basic multiple docsets support

yufeih:
Ignore multiple docset on loc cases.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5153)


src/docfx/lib/log/Log.cs
```diff
@@ -32,6 +32,19 @@ public static void Important(string message, ConsoleColor color)
             }
         }
 
+        public static void Error(Error error, ErrorLevel? level = null)
```
live1206: think `ConsoleLog` is better than `Error`

src/docfx/config/ConfigLoader.cs
```diff
@@ -18,13 +19,41 @@ internal class ConfigLoader
...
         }
 
+        public static (string docsetPath, string outputPath)[] FindDocsets(string workingDirectory, CommandLineOptions options)
+        {
+            if (DisableMultipleDocsets != null && DisableMultipleDocsets())
+            {
+                return new[] { (workingDirectory, options.Output) };
+            }
+
+            return Directory.GetFiles(workingDirectory, "docfx.yml", SearchOption.AllDirectories)
```
live1206: Doesn't docfx v3 accept `docfx.json` as well?
Or does this mean multiple-docset should only run after config migration tool for Docs?

yufeih: yes, it should support `docfx.json`, see TODO comments in line 44

src/docfx/lib/log/Log.cs
yufeih: yep, initially I thought all console writes goes to `Log`, turned out to be not true

src/docfx/config/ConfigLoader.cs
```diff
@@ -18,13 +19,41 @@ internal class ConfigLoader
...
+
+                // TODO: look for docfx.json after config migration tool has been merged info docfx
```
live1206:
```suggestion
                // TODO: look for docfx.json after config migration tool generates docfx.yml the same directory as docfx.json
```

PR #5164, opened 2019-10-05
Add persistence to the toc filter control and a clear button

jo3w4rd:
This PR adds persistence to the filter used for the TOC. The string entered into the control is saved to the local browser session storage and reloaded when the page is loaded. This means that if you enter a filter and see four likely results left in the table of contents list, you can visit all four pages without having to retype the filter string.

The PR also adds a clear button to the control since the string is no longer cleared automatically when you navigate to a new page.

Details:
- Update docfx.js to save the TOC filter string. Also, show and hide the clear button.
- Update toc.html.tmpl to add a clear button to the TOC filter control.
- Update docfx.css to style the clear button and make room at the right side of the control for it.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5164)


superyyrrzz:
@jo3w4rd Thank you for contributing the feature. It looks good. I will test it and review this.

superyyrrzz:
Just tried out and I really like this feature! Thank you for your contribution!

One minor issue is that CI is failing. It must due to a bug in dev branch before, which is fixed now. Could you merge the latest dev branch into your PR? Usually I can do this by click `Update branch` button in PR, but it is strange that this button does not appear in this PR.

jo3w4rd:
Which branch should I set the PR to merge into? Currently I have it set to master, but should it be dev? Otherwise merging in the latest dev branch wouldn't be correct since that would also pull those changes into your master branch when this PR is merged.

superyyrrzz:
It should target `dev` branch.

jo3w4rd:
A test is failing because it is looking for:

```
//div[@class='sidetoc']/div[@id='toc']/ul/li[@class='active in']/a
```
 When the element actually has `[@class='hide active in']`

superyyrrzz:
@jo3w4rd These tests have not been updated for long time. It works well on CI machine but not locally. I cannot think of why it has `hide` here.

I will update these tests in separate PR to unblock this.

jo3w4rd:
>  I cannot think of why it has hide here.

The TOC filter logic adds `show` and `hide` css classes as part of the filtering mechanism. It is possible that if an earlier test set the filter and didn't clear it, then the TOC entry the test is looking for might still be hidden. Before the change in this PR, reloading the page would have cleared the filter and removed both the `show1 and `hide` classes from the TOC elements. Now, however, if the test just reloads the page, but doesn't trigger a new browser session, a filter string from one test would be reloaded in a subsequent test. That's a theory anyway, I haven't looked at the tests closely enough to verify.

superyyrrzz:
@jo3w4rd Sounds reasonable guess. Anyway, I removed these tests as they are a bit old and out of maintenance.

Issue #5165, opened 2019-10-05
docfx crash report: System.IO.DirectoryNotFoundException


vermiculus:
# docfx crash report: System.IO.DirectoryNotFoundException

docfx: `3.0.0-beta-01261-52746d1b2+52746d1b2731665a6b9b1220711d8ab87a3bbf5a`
dotnet: `2.2.103`
x64: `True`
git: `git version 2.22.0`

## repro steps

Run `/Users/sean/.dotnet/tools/.store/docfx/3.0.0-beta-01261-52746d1b2/docfx/3.0.0-beta-01261-52746d1b2/tools/netcoreapp2.2/any/docfx.dll build` in `/Users/sean/git/docfx-test`

## callstack

```
System.IO.DirectoryNotFoundException: Could not find a part of the path '/Users/sean/git/docfx-test/_themes/ContentTemplate/.html.liquid'.
   at Interop.ThrowExceptionForIoErrno(ErrorInfo errorInfo, String path, Boolean isDirectory, Func`2 errorRewriter)
   at Microsoft.Win32.SafeHandles.SafeFileHandle.Open(String path, OpenFlags flags, Int32 mode)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.StreamReader..ctor(String path, Encoding encoding, Boolean detectEncodingFromByteOrderMarks, Int32 bufferSize)
   at System.IO.File.InternalReadAllText(String path, Encoding encoding)
   at System.IO.File.ReadAllText(String path)
   at Microsoft.Docs.Build.LiquidTemplate.LoadTemplate(String fullPath) in D:\a\1\s\src\docfx\template\liquid\LiquidTemplate.cs:line 90
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at Microsoft.Docs.Build.LiquidTemplate.Render(String templateName, JObject model)
   at Microsoft.Docs.Build.BuildPage.CreatePageOutput(Context context, Document file, JObject sourceModel) in D:\a\1\s\src\docfx\build\page\BuildPage.cs:line 114
[...]

PR #5166, opened 2019-10-08
Handle .html.liquid file not found

live1206:
#5165 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5166)


src/docfx/template/liquid/LiquidTemplate.cs
```diff
@@ -122,7 +139,18 @@ public Template GetTemplate(DotLiquid.Context context, string templateName)
             {
                 return _templates.GetOrAdd(
                     templateName,
-                    new Lazy<Template>(() => LoadTemplate(Path.Combine(_templateDir, "_includes", templateName + ".liquid")))).Value;
+                    new Lazy<Template>(() =>
+                        {
+                            try
+                            {
+                                return LoadTemplate(Path.Combine(_templateDir, "_includes", templateName + ".liquid"));
+                            }
```
mingwli: add a test for this when inclusion .liquid not found?

live1206: As I can see, this method is not in use yet

live1206: Seems like it is used for inclusion inside the liquid template. Since it is not related to this fix, leave it for now.

docs/specs/markdown.yml
```diff
@@ -608,3 +608,13 @@ outputs:
...
+---
+# Liquid template file missing
+inputs: 
+  docfx.yml: |
+    output:
+      json: false
+  docs/a.md: Hello `docfx`! 
+outputs: 
+  .errors.log: |
+    ["error","liquid-template-missing","The liquid template '.html.liquid' is missing","docs/a.md"]
```
yufeih: We currently treat template as code, failures in template (like javascript error) causes an unhandled exception and crashes docfx. I'm not sure if it is a good idea to introduce this error code. For instance, if a `.mta.js` is missing, we ignore the script and return the input model instead. Similarly, we can have a default implementation of `.html.liquid` that shows JSON input

live1206: sounds reasonable

src/docfx/template/liquid/LiquidTemplate.cs
```diff
@@ -39,7 +39,26 @@ public string Render(string templateName, JObject model)
...
                 templateName,
-                new Lazy<Template>(() => LoadTemplate(Path.Combine(_templateDir, templateName + ".html.liquid")))).Value;
+                new Lazy<Template>(() =>
+                {
+                    var fileName = $"{templateName}.html.liquid";
+                    try
+                    {
+                        return LoadTemplate(Path.Combine(_templateDir, fileName));
+                    }
+                    catch (Exception ex)
```
yufeih: Consider use return value to show that template is optional, exceptions are fragile and implicit. If we modify `LoadTemplate` method to read from git, this wrapping logic will fail to function because git might throw a different exception type and people will forget to change here.

src/docfx/template/liquid/LiquidTemplate.cs
```diff
@@ -122,7 +141,18 @@ public Template GetTemplate(DotLiquid.Context context, string templateName)
             {
                 return _templates.GetOrAdd(
                     templateName,
-                    new Lazy<Template>(() => LoadTemplate(Path.Combine(_templateDir, "_includes", templateName + ".liquid")))).Value;
+                    new Lazy<Template>(() =>
```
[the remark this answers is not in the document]
live1206: The bug reported from community set config `JSON` to false, trying to output HTML by calling liquid template, and there is no template in place. Who should support the template in this case? us or user? or we do not care if there is no template?

src/docfx/template/liquid/LiquidTemplate.cs
live1206: you are right

src/docfx/template/liquid/LiquidTemplate.cs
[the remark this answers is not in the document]
live1206: Got your point now, the inclusion template is explicitly used in liquid template, it should be in place. No matter it is defined by docs or community user.

PR #5185, opened 2019-10-11
fix: normalize sitebasepath when it is . 

partychen:
normalize sitebasepath when it is . to fix canonicalUrlPrefix and filemap basepath

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5185)

### DIFF
--- a/src/docfx/build/legacy/LegacyFileMap.cs
+++ b/src/docfx/build/legacy/LegacyFileMap.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Docs.Build
                 {
                     host = docset.HostName,
                     locale = docset.Locale,
-                    base_path = $"/{docset.SiteBasePath}",
+                    base_path = $"/{PathUtility.NormalizeSiteBasePath(docset.SiteBasePath)}",
                     source_base_path = docset.Config.DocumentId.SourceBasePath,
                     version_info = new { },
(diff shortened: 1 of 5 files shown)


src/docfx/build/legacy/LegacyFileMap.cs
```diff
@@ -61,7 +61,7 @@ public static void Convert(Docset docset, Context context, IEnumerable<(string p
                 {
                     host = docset.HostName,
                     locale = docset.Locale,
-                    base_path = $"/{docset.SiteBasePath}",
+                    base_path = $"/{PathUtility.NormalizeSiteBasePath(docset.SiteBasePath)}",
```
yufeih: Instead of normalize the site base path on each consume, it is better to normalize it in `SpliteBaseUrl` method, the `siteBasePath` property produced by that method should return a normalized URL that always starts with `/`

yufeih: This and #5176 are caused by the same problem and can be fixed using the aforementioned method

yufeih:
Related https://github.com/dotnet/docfx/issues/4115

PR #5193, opened 2019-10-15
Update conceptual-localization.md

live1206:
`RepositoryAndFolder` mapping not supported in v3 anymore

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5193)

### DIFF
--- a/docs/designs/conceptual-localization.md
+++ b/docs/designs/conceptual-localization.md
@@ -49,25 +49,6 @@ Below kinds of mappings are considered to be supported and there is a **strong c
     ```
     
     > NOTE: The loc org name can be different, it should be configurable
-    
-  - **RepositoryAndFolder**, localization files are stored in ONE **different repository** for **all locales** under different **locale folder**
-  
-    Here is an string convention for loc repo name:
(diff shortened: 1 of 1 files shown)


docs/designs/conceptual-localization.md
```diff
@@ -49,25 +49,6 @@ Below kinds of mappings are considered to be supported and there is a **strong c
     ```
     
     > NOTE: The loc org name can be different, it should be configurable
-    
-  - **RepositoryAndFolder**, localization files are stored in ONE **different repository** for **all locales** under different **locale folder**
```
herohua: >  - **RepositoryAndFolder**, localization files are stored in ONE **different repository** for **all locales** under  [](start = 0, length = 117)

so we'd like to deprecate this functionality?

OsmondJiang: this option is not supported by docfx, just forgot to remove it.

PR #5198, opened 2019-10-16
support triple colon image extension

928PJY:
Fix https://ceapex.visualstudio.com/Engineering/_workitems/edit/131339/?src=WorkItemMention&src-action=artifact_link

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5198)

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -624,19 +624,3 @@ outputs:
     { 
       "conceptual": "<img src=\"a.png\" alt=\"Image\" data-linktype=\"relative-path\"><img src=\"b.png\" alt=\"Image\" data-linktype=\"relative-path\">",
     }
-  .links.json: |
-    {
-      "links": [
-        { "source_url": "/docs/a", "target_url": "/docs/a.png" },
-        { "source_url": "/docs/a", "target_url": "/docs/b.png" }
-      ]
-    }
-  .dependencymap.json: |
-    {
-      "dependencies": {
-        "docs/a.md": [
-            { "source": "docs/a.png", "type": "link" },
-            { "source": "docs/b.png", "type": "link" }
-        ]
-      }
-    }
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -608,3 +608,35 @@ outputs:
                           <a href=\"a#\" data-linktype=\"relative-path\">a1</a>
                      </p>"
     }
+---
+# Triple colon markdown extension
+inputs:
+  docfx.yml:
+  docs/a.png:
+  docs/b.png:
+  docs/a.md: |
+    :::image type="content" source="a.png" alt-text="Image":::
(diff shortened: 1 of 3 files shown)


src/docfx/lib/markdown/LinkExtension.cs
```diff
@@ -36,6 +39,13 @@ internal static class LinkExtension
                     {
                         inline.Tag = ResolveLinks(inline.Tag, inline);
                     }
+                    else if (node is TripleColonBlock tripleColonBlock && tripleColonBlock.Extension is ImageExtension)
+                    {
+                        var imageSrc = tripleColonBlock.GetAttributes().Properties.Where(kv => kv.Key == "src").FirstOrDefault().Value;
```
live1206: `.FirstOrDefault()` is there always a "src" property? otherwise, it could throw NRE
If always, then we should use `tripleColonBlock.GetAttributes().Properties.Single(kv => kv.Key == "src").Value`

928PJY: yes there will always a "src" property

docs/specs/markdown.yml
```diff
@@ -608,3 +608,35 @@ outputs:
...
+    :::image type="content" source="a.png" alt-text="Image":::
+    :::image type="content" source="~/docs/b.png" alt-text="Image":::
+outputs:
+  docs/a.png:
+  docs/b.png:
+  docs/a.json: |
+    { 
+      "conceptual": "<img src=\"a.png\" alt=\"Image\" data-linktype=\"relative-path\"><img src=\"b.png\" alt=\"Image\" data-linktype=\"relative-path\">",
+    }
+  .links.json: |
```
yufeih: nit: consider remove the verification of `.links.json` and `.dependencymap.json` to simplify the test case. Verifying output content is probably enough to show that links in triple colon images are resolved as other links. Other resolve tests cases are also written without checking these files. There are dedicated test cases to check `.links.json` and `.dependencymap.json` generation.

928PJY: make sense, I will move the .link.json validation to the specific case

src/docfx/lib/markdown/LinkExtension.cs
```diff
@@ -36,6 +39,13 @@ internal static class LinkExtension
                     {
                         inline.Tag = ResolveLinks(inline.Tag, inline);
                     }
+                    else if (node is TripleColonBlock tripleColonBlock && tripleColonBlock.Extension is ImageExtension)
+                    {
+                        var imageSrc = tripleColonBlock.GetAttributes().Properties.Single(kv => kv.Key == "src").Value;
```
yufeih: nit: this algorithm is a bit in-efficient, the underlying storage of `GetAttributes().Properties` is a mutable list, using a `for` loop is more efficient and readable here.

PR #5199, opened 2019-10-16
🐛fix docset path to root

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5199)


live1206:
offline synced with @928PJY, will remove `docset_path_to_root` from docfx since docfx build does not have the info of repo root

PR #5201, opened 2019-10-16
exclude localization folder based on mapping type

mingwli:
Take docs-help-pr for example, the [localization folder](https://github.com/MicrosoftDocs/docs-help-pr/tree/master/help-content/localization)is meant to be built and published.
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5201)

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -14,7 +14,6 @@ namespace Microsoft.Docs.Build
         public static readonly string[] DefaultExclude = new[]
         {
             "_site/**",             // Default output location
-            "localization/**",      // Localization file when using folder convention
             "_themes/**",           // Default template location
         };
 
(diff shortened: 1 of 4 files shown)


src/docfx/config/Config.cs
```diff
@@ -14,7 +14,6 @@ internal sealed class Config
         public static readonly string[] DefaultExclude = new[]
         {
             "_site/**",             // Default output location
-            "localization/**",      // Localization file when using folder convention
```
OsmondJiang: what about rename it to _localization

mingwli: #5202

PR #5203, opened 2019-10-16
fix: doc siteUrl can be only /

partychen:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5203)


yufeih:
add a test case with `--legacy` to illustrate the bug?

PR #5207, opened 2019-10-17
Clarify the site host within .xrefmap.json

live1206:
For live branch build, the site host within `.xrefmap.json` should be `docs.microsoft.com`
For non-live branch or PR build, the site host within `.xrefmap.json` should be `review.docs.microsoft.com`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5207)


live1206:
> By saying _clarify_ I assume this is the current behavior and no code change needed?

Nope, this is not the current behavior, we should agree on the behavior firstly. And then fix the code

PR #5210, opened 2019-10-17
Move json token cache to input

yufeih:
Manage JSON/YAML token cache in `Input`. 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5210)

### DIFF
--- a/src/docfx/build/context/Cache.cs
+++ b/src/docfx/build/context/Cache.cs
@@ -4,16 +4,11 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.IO;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal class Cache
     {
-        private readonly ConcurrentDictionary<string, Lazy<(List<Error>, JToken)>> _tokenCache
-            = new ConcurrentDictionary<string, Lazy<(List<Error>, JToken)>>();
-
         private readonly ConcurrentDictionary<string,
             Lazy<(List<Error> errors, TableOfContentsModel tocModel, List<Document> referencedFiles, List<Document> referencedTocs)>>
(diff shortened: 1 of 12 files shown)


test/docfx.Test/TestUtility.cs
```diff
@@ -66,7 +66,7 @@ void Throw(string stackTrace, string message, string detailMessage, string info)
 
         private static void CreateZipFile(KeyValuePair<string, string> file, string filePath)
         {
-            var token = YamlUtility.ToJToken(file.Value);
+            var token = YamlUtility.ToJToken(new StringReader(file.Value));
```
live1206: why not put `new StringReader()` into `ToJToken` as well, or another overloading for it?

src/docfx/build/context/Cache.cs
```diff
@@ -4,16 +4,11 @@
...
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.IO;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal class Cache
     {
-        private readonly ConcurrentDictionary<string, Lazy<(List<Error>, JToken)>> _tokenCache
```
live1206: what are we benefiting by moving the JSON/YAML cache from `Cache.cs` to `Input.cs`?
Or why do we leave TOC cache in `Cache.cs`?

live1206: Just noticed #5211, so the idea is to get rid of `Cache.cs`

test/docfx.Test/TestUtility.cs
yufeih: Either is ok, `ToJToken` is more of a private method, I initially extract it to share between docfx and docastest. Didn't thought it is used here. This line can call `Deserialize<JToken>` instead

PR #5211, opened 2019-10-17
Constructor inject TOC loader dependencies

yufeih:
This is a change after #5210 so it contains some changes in that PR before it is merged. Review #5210 first.

The goal is to:

- Remove `Context.Cache` and each component manages it's own cache
- Turn TOC loader into an instance to house the cache
- Inject TOC loader dependencies using constructor injection

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5211)


src/docfx/build/toc/TableOfContentsLoader.cs
```diff
@@ -11,50 +12,80 @@
...
         {
-            var referencedFiles = new List<Document>();
-            var referencedTocs = new List<Document>();
+            return _cache.GetOrAdd(file.FilePath, _ =>
+            {
+                var referencedFiles = new List<Document>();
+                var referencedTocs = new List<Document>();
 
-            var (errors, model) = LoadInternal(context, file, file, referencedFiles, referencedTocs);
+                var (errors, model) = LoadInternal(file, file, referencedFiles, referencedTocs);
```
live1206: why is the indent different here?

live1206: I see now

PR #5212, opened 2019-10-18
Try run json transform only once for page

mingwli:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5212)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -253,20 +253,17 @@ namespace Microsoft.Docs.Build
             errors.AddRange(schemaValidationErrors);
 
             // transform model via json schema
-            var (schemaTransformError, transformedToken) = schemaTemplate.JsonSchemaTransformer.TransformContent(file, context, obj);
-            errors.AddRange(schemaTransformError);
-            var pageModel = (JObject)transformedToken;
-
             if (file.IsPage)
             {
                 // transform metadata via json schema
                 var (metadataErrors, inputMetadata) = context.MetadataProvider.GetMetadata(file);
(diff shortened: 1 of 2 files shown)


mingwli:
trying to modify input may influence enumeration. see #4960
close this one

mingwli:
> See #4960 , you cannot change the input token since `JToken` is not immutable and the same reference is shared with / cached in other components.

As adviced by @yufeih, try copy clone before running schema content transform

src/docfx/build/page/BuildPage.cs
```diff
@@ -252,22 +252,21 @@ private static async Task<(List<Error> errors, JObject model)> LoadJson(Context
             var schemaValidationErrors = schemaTemplate.JsonSchemaValidator.Validate(obj);
             errors.AddRange(schemaValidationErrors);
 
-            // transform model via json schema
-            var (schemaTransformError, transformedToken) = schemaTemplate.JsonSchemaTransformer.TransformContent(file, context, obj);
-            errors.AddRange(schemaTransformError);
-            var pageModel = (JObject)transformedToken;
+            obj = (JObject)JsonUtility.ShallowClone(obj);
```
yufeih: nit: consider give it a different name if the variable is mutated:

```suggestion
            var xxx = (JObject)JsonUtility.ShallowClone(obj);
```

mingwli: using JsonUtility.Merge for shallow clone

yufeih:
The diff shows that FileLinkBuilder does not de-dup results

PR #5214, opened 2019-10-18
fix: LandingPage missing title and cardtext section.

partychen:
1. fix LandingPage missing cardtext section when item don't have image src. https://ceapex.visualstudio.com/Engineering/_workitems/edit/132787

2. fix LandingPage <a> missing title https://ceapex.visualstudio.com/Engineering/_workitems/edit/132790

### DIFF
--- a/src/Microsoft.Docs.Template/Views/Template/BuildItemListPartial.cshtml
+++ b/src/Microsoft.Docs.Template/Views/Template/BuildItemListPartial.cshtml
@@ -123,7 +123,7 @@ else if (Model.Style == "cards")
                         <a class="card" href="@listItem.Href">
                             @if (listItem.Image?.Src != null)
                             {
-                                <img class="cardImage" role="presentation" src="@listItem.Image.Src">
+                            <img class="cardImage" role="presentation" src="@listItem.Image.Src">
                             }
                             <div class="cardText">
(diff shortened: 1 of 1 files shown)


src/Microsoft.Docs.Template/Views/Template/BuildItemListPartial.cshtml
```diff
@@ -124,18 +124,18 @@ else if (Model.Style == "cards")
...
-                                    {
-                                        <p>@listItem.Text</p>
-                                    }
-                                    else if (!string.IsNullOrEmpty(listItem.Html))
-                                    {
-                                        <p>@Html.Raw(removeTagsAndComments(listItem.Html))</p>
-                                    }
-                                </div>
                             }
+                            <div class="cardText">
```
live1206: indent is different here

yufeih:
@partychen could you update the title a little bit to show that it is for LandingData?

src/Microsoft.Docs.Template/Views/Template/BuildItemListPartial.cshtml
partychen: from the html i think the indent is correct. And i change the indent of  to let it looks pretty.

PR #5216, opened 2019-10-21
remove source base path

mingwli:
it's always '.' in docset, all relative|combile operation can be removed
document id can use filePath directly

[DevOps#132567](https://dev.azure.com/ceapex/Engineering/_workitems/edit/132567)
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5216)

### DIFF
--- a/docs/specs/metadata.yml
+++ b/docs/specs/metadata.yml
@@ -73,18 +73,13 @@ outputs:
 # document id = md5("Win.sccm|mdt/release-notes.md")
 # version independent id = md5("Win.sccm|mdt/release-notes")
 inputs:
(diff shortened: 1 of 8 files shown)


live1206:
Before cleaning in docfx, there are some external dependencies need to be removed firstly

mingwli:
> Before cleaning in docfx, there are some external dependencies need to be removed firstly

do you mean migration-ci? this change may not bring any output changes I suppose.

PR #5222, opened 2019-10-23
Explicitly specify config in loc tests

yufeih:
Drop support for loc config fallback for added complexity

Related #5215 #5221 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5222)


docs/specs/localization.yml
```diff
@@ -697,6 +770,10 @@ repos:
             bilingual: true
   https://github.com/ss/c.zh-cn#master-sxs:
     - files:
+        docfx.yml: |
+          # TODO: should be able to figure out bilingual automatically using convention
```
yufeih: This case illustrate that when build form loc, we currently aren't using convention to figure out if it is sxs

PR #5227, opened 2019-10-24
Replace RestoreFileMap.GetRestoredFilePath with stream based methods

yufeih:
Read from stream if xrefmap is not a physical file.

Related #5116 #5215 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5227)

### DIFF
--- a/src/docfx/build/xref/ExternalXrefMapLoader.cs
+++ b/src/docfx/build/xref/ExternalXrefMapLoader.cs
@@ -86,11 +86,11 @@ namespace Microsoft.Docs.Build
             {
                 foreach (var entry in archive.Entries)
                 {
-                    using (var sr = new StreamReader(entry.Open()))
+                    using (var entryStream = entry.Open())
                     {
                         if (entry.FullName.EndsWith(".yml", StringComparison.OrdinalIgnoreCase))
                         {
-                            var xrefMap = YamlUtility.Deserialize<XrefMapModel>(sr, path);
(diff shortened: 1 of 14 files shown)


src/docfx/template/TemplateSchema.cs
```diff
@@ -21,7 +21,7 @@ public TemplateSchema(string schemaName, string schemaDir, string contentTemplat
         {
             Debug.Assert(!string.IsNullOrEmpty(schemaName));
 
-            SchemaName = SchemaName;
+            SchemaName = schemaName;
```
yufeih: The updated FxCop detects _this_ bug 😄  @mingwli

src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -20,34 +20,45 @@ internal class ExternalXrefMapLoader
...
+                        // Fast pass for JSON xref files
+                        foreach (var (uid, spec) in LoadJsonFile(fileStream.Name))
+                        {
+                            // for same uid with multiple specs, we should respect the order of the list
+                            result.TryAdd(uid, spec);
+                        }
+                    }
+                    else
+                    {
+                        var xrefMap = YamlUtility.Deserialize<XrefMapModel>(new StreamReader(stream), path);
```
live1206: should this be `JsonUtility.Deserialize`?

PR #5229, opened 2019-10-25
Fix PackagePath json serialize

yufeih:
Clarifies that `url` means repository url (without any branch info), `path` means a local folder.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5229)


docs/specs/config.yml
```diff
@@ -586,47 +586,6 @@ outputs:
   docs/path1.json: |
     {"document_id": "780484e9-a367-97ab-ba7d-a19f3b378a80", "document_version_independent_id": "e9b3f4e2-1a78-4b5d-fbb4-03e0a33e38f5",}
 ---
-# branch property has higher priority than branch defined in url
```
yufeih: Removed this tests because they are covered in JSON deserialize UT

PR #5234, opened 2019-10-28
Add yaml mime to document and enable SDP content validation 

hejingkan2005:
- for validation purpose, add root file yaml mime into document 
- enable .yml files validation

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5234)

### DIFF
--- a/src/Microsoft.DocAsCode.Build.SchemaDriven/Processors/MarkdownInterpreter.cs
+++ b/src/Microsoft.DocAsCode.Build.SchemaDriven/Processors/MarkdownInterpreter.cs
@@ -35,7 +35,7 @@ namespace Microsoft.DocAsCode.Build.SchemaDriven.Processors
         {
             var host = context.Host;
 
-            var mr = host.Markup(content, context.GetOriginalContentFile(path));
+            var mr = host.Markup(content, context.GetOriginalContentFile(path), false, true);
             (context.FileLinkSources).Merge(mr.FileLinkSources);
             (context.UidLinkSources).Merge(mr.UidLinkSources);
             (context.Dependency).UnionWith(mr.Dependency);
--- a/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
@@ -32,7 +32,11 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
                 && (string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yml", StringComparison.OrdinalIgnoreCase) 
                 || string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yaml", StringComparison.OrdinalIgnoreCase)))
                 {
-                    document.SetData("SchemaName", YamlMime.ReadMime(InclusionContext.RootFile?.ToString()));
+                    var schemaName = YamlMime.ReadMime(InclusionContext.RootFile?.ToString());
+                    if (!string.IsNullOrEmpty(schemaName))
+                    {
+                        document.SetData("SchemaName", schemaName);
+                    }
                 }
 
                 visitor.Visit(document);
--- a/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
@@ -41,6 +41,7 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
 
         public static void SetSchemaName(MarkdownDocument document)
         {
+            // TODO: add this detection logic in terms of performance optimization, should remove once mime is available in context
             if (InclusionContext.IsInclude
                 && (string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yml", StringComparison.OrdinalIgnoreCase)
                 || string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yaml", StringComparison.OrdinalIgnoreCase)))
--- a/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
@@ -32,7 +32,7 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
                 && (string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yml", StringComparison.OrdinalIgnoreCase) 
                 || string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yaml", StringComparison.OrdinalIgnoreCase)))
                 {
-                    document.SetData("rootFileYamlMime", YamlMime.ReadMime(InclusionContext.RootFile?.ToString()));
+                    document.SetData("SchemaName", YamlMime.ReadMime(InclusionContext.RootFile?.ToString()));
                 }
 
                 visitor.Visit(document);
--- a/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
@@ -7,6 +7,7 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
     using System.IO;
     using Markdig;
     using Markdig.Renderers;
+    using Markdig.Syntax;
     using Microsoft.DocAsCode.Common;
 
     public class ValidationExtension : IMarkdownExtension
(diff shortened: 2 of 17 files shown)


src/Microsoft.DocAsCode.Build.SchemaDriven/Processors/MarkdownInterpreter.cs
```diff
@@ -35,7 +35,7 @@ private static string MarkupCore(string content, IProcessContext context, string
         {
             var host = context.Host;
 
-            var mr = host.Markup(content, context.GetOriginalContentFile(path));
+            var mr = host.Markup(content, context.GetOriginalContentFile(path), false, true);
```
herohua: this actually enables validation - is this your purpose for this PR?

hejingkan2005: Yes, this PR is targeting to enable validation and add YamlMime to the document, then validation library could test it. 
Btw, I understand the next build is scheduled to Nov. 18, I think I get plenty of time for testing before going to production.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -25,6 +28,13 @@ public void Setup(MarkdownPipelineBuilder pipeline)
 
             pipeline.DocumentProcessed += document =>
             {
+                if (InclusionContext.IsInclude 
+                && (string.Equals(Path.GetExtension((RelativePath)InclusionContext.RootFile), ".yml", StringComparison.OrdinalIgnoreCase)
```
superyyrrzz: No need to convert to `(RelativePath)`? Or convert it to `string` as to what `Path.GetExtension` needs?

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -25,6 +28,13 @@ public void Setup(MarkdownPipelineBuilder pipeline)
 
             pipeline.DocumentProcessed += document =>
             {
+                if (InclusionContext.IsInclude
```
superyyrrzz: Could validation library handle this? Other component does not need this, and validation library has enough knowledge to calculate this.

hejingkan2005: The purpose I add "IsInclude" condition here is because the Yaml file could contain 'inline' (not included) markdown property, and in such case, we actually don't have to tag 'yamlMime' as we would skip validation when markdown inline.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
hejingkan2005: Sure

src/Microsoft.DocAsCode.Build.SchemaDriven/Processors/MarkdownInterpreter.cs
hejingkan2005: I have a offline talk with Renze, I will complete this PR after validation library is ready

hejingkan2005: Update title to reflect the real purpose of this pr.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -25,6 +28,13 @@ public void Setup(MarkdownPipelineBuilder pipeline)
 
             pipeline.DocumentProcessed += document =>
             {
+                if (InclusionContext.IsInclude 
+                && (string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yml", StringComparison.OrdinalIgnoreCase)
```
herohua: can `Path.GetExtension(InclusionContext.RootFile?.ToString())` handle `Path.GetExtension(null)`?

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -25,6 +28,13 @@ public void Setup(MarkdownPipelineBuilder pipeline)
 
             pipeline.DocumentProcessed += document =>
             {
+                if (InclusionContext.IsInclude 
+                && (string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yml", StringComparison.OrdinalIgnoreCase) 
+                || string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yaml", StringComparison.OrdinalIgnoreCase)))
+                {
+                    document.SetData("yamlMime", YamlMime.ReadMime(InclusionContext.RootFile?.ToString()));
```
herohua: yamlMime may not be a good name as it is actually the MIME of the **root file**

shall we consider `rootFileYamlMime`?

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
hejingkan2005: Yes, Path.GetExtension(null) returns null here.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
hejingkan2005: Sure, rename it.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -25,6 +28,13 @@ public void Setup(MarkdownPipelineBuilder pipeline)
 
             pipeline.DocumentProcessed += document =>
             {
+                if (InclusionContext.IsInclude 
+                && (string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yml", StringComparison.OrdinalIgnoreCase) 
+                || string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yaml", StringComparison.OrdinalIgnoreCase)))
+                {
+                    document.SetData("rootFileYamlMime", YamlMime.ReadMime(InclusionContext.RootFile?.ToString()));
```
yufeih:
```suggestion
                    document.SetData("SchemaName", YamlMime.ReadMime(InclusionContext.RootFile?.ToString()));
```

superyyrrzz:
Add some tests.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -25,6 +28,13 @@ public void Setup(MarkdownPipelineBuilder pipeline)
 
             pipeline.DocumentProcessed += document =>
             {
+                if (InclusionContext.IsInclude 
+                && (string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yml", StringComparison.OrdinalIgnoreCase) 
+                || string.Equals(Path.GetExtension(InclusionContext.RootFile?.ToString()), ".yaml", StringComparison.OrdinalIgnoreCase)))
+                {
+                    document.SetData("SchemaName", YamlMime.ReadMime(InclusionContext.RootFile?.ToString()));
```
superyyrrzz: `null.ToString()` will throw exception. Skip `SetData` when RootFile is null.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
herohua: This looks very much like the logic that should be handled on validation library side - SDP should be general purposed that the insertion of YamlMime should not depend on whether the Markdig piece is inline or not.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
herohua: +1

---
In reply to: [339871951](https://github.com/dotnet/docfx/pull/5234#discussion_r339871951) [](ancestors = 339871951)

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
hejingkan2005: InclusionContext.RootFile should not be null, or it would drop into this line. But you are right, that ReadMime could be null, so let me address this.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
hejingkan2005: Renze and I talk over this offline, the reason I add the logic in docfx side is because validation service could not retrieve root file yaml mime with the file path provided. It is a relative path, validation library could not retrieve yaml mime directly due to lack of docfx file locale in the context.

hejingkan2005:
> Add some tests.

Done

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
hejingkan2005: Agree

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -25,6 +28,17 @@ public void Setup(MarkdownPipelineBuilder pipeline)
 
             pipeline.DocumentProcessed += document =>
             {
+                if (InclusionContext.IsInclude
```
superyyrrzz: You can extract a method for this delegate, then test can also use this method.  `public void Validate(MarkdownDocument document)`

hejingkan2005: Thanks for the suggestion, done with extraction.

src/Microsoft.DocAsCode.MarkdigEngine/Validation/ValidationExtension.cs
```diff
@@ -33,5 +38,20 @@ public void Setup(MarkdownPipeline pipeline, IMarkdownRenderer renderer)
         {
 
         }
+
+        public static void SetSchemaName(MarkdownDocument document)
```
herohua: >public [](start = 8, length = 6)

can this be `internal`?

PR #5235, opened 2019-10-28
[.links.json] source_url should be root file for markdown token

yufeih:
Follow up on #5232 .

This is a dirty quick fix

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5235)


docs/specs/output.yml
```diff
@@ -490,3 +490,24 @@ outputs:
...
+---
+# source_url is the includer for links and xrefs in markdown token
+inputs:
+  docfx.yml:
+  a.md: |
+    ---
+    uid: a
+    ---
+    [!include[](b.md)]
+  b.md: '[](/a-link), @a'
```
live1206: shall we exclude `b.md` from build?

PR #5237, opened 2019-10-28
🐛Git read history from fallback repo with multiple docset

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/134116

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5237)

PR #5238, opened 2019-10-28
🧪 Add test to support build without a git repo

yufeih:
#5237 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5238)

### DIFF
--- a/src/docfx/docset/Docset.cs
+++ b/src/docfx/docset/Docset.cs
@@ -129,9 +129,9 @@ namespace Microsoft.Docs.Build
                     return Repository.Create(fullPath, branch: null);
                 }
 
-                var parent = Path.GetDirectoryName(fullPath);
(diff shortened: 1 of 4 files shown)


src/docfx/docset/Docset.cs
```diff
@@ -129,9 +129,9 @@ Repository GetRepositoryInternal(string fullPath)
                     return Repository.Create(fullPath, branch: null);
                 }
 
-                var parent = Path.GetDirectoryName(fullPath);
+                var parent = PathUtility.NormalizeFile(Path.GetDirectoryName(fullPath));
```
live1206: should it be `PathUtility.NormalizeFolder`?

PR #5247, opened 2019-10-30
Refine RepositoryProvider

live1206:
- remove config fallback
- remove all the config methods for `RepositoryProvider`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5247)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -86,8 +86,8 @@ namespace Microsoft.Docs.Build
             Config config,
             RepositoryProvider repositoryProvider)
         {
-            Docset currentDocset;
             var (currentDocsetPath, currentRepo) = repositoryProvider.GetRepositoryWithDocsetEntry(FileOrigin.Default);
+            var currentDocset = new Docset(currentDocsetPath, locale, config, currentRepo);
             if (!string.IsNullOrEmpty(locale) && !string.Equals(locale, config.Localization.DefaultLocale))
             {
                 var (fallbackDocsetPath, fallbackRepo) = repositoryProvider.GetRepositoryWithDocsetEntry(FileOrigin.Fallback);
@@ -99,7 +99,6 @@ namespace Microsoft.Docs.Build
                 }
             }
 
-            currentDocset = new Docset(currentDocsetPath, locale, config, currentRepo);
             return (currentDocset, default);
         }
 
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -42,7 +42,7 @@ namespace Microsoft.Docs.Build
 
                     using (var restoreGitMap = RestoreGitMap.Create(docsetPath, locale))
                     {
-                        var repositoryProvider = new RepositoryProvider(docsetPath, options, repository, restoreGitMap);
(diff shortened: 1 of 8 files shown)


src/docfx/docset/RepositoryProvider.cs
```diff
@@ -13,60 +13,26 @@ internal class RepositoryProvider
             = new ConcurrentDictionary<(FileOrigin origin, string dependencyName), Lazy<(string docset, Repository repository)>>();
 
         private readonly string _locale;
+        private readonly Func<RestoreGitMap> _restoreGitMap;
+        private readonly Func<Config> _config;
+        private readonly Lazy<(string fallbackDocsetPath, Repository fallbackRepo)> _fallbackDocset;
+
         private string _docsetPath;
```
yufeih: since this is mutable, it'll produce different result depending on when GetRepository is called.

src/docfx/build/Build.cs
```diff
@@ -36,33 +37,23 @@ private static async Task BuildDocset(string docsetPath, string outputPath, Comm
                 try
                 {
                     // load and trace entry repository
-                    var repositoryProvider = new RepositoryProvider(docsetPath, options);
+                    var repositoryProvider = new RepositoryProvider(docsetPath, options, () => restoreGitMap, () => config);
```
OsmondJiang: config can be always loaded from current repository if we don;t need to consider the case that loc repo dones't have the config.

src/docfx/build/Build.cs
```diff
@@ -36,33 +37,23 @@ private static async Task BuildDocset(string docsetPath, string outputPath, Comm
...
                     var repository = repositoryProvider.GetRepository(FileOrigin.Default);
                     Telemetry.SetRepository(repository?.Remote, repository?.Branch);
                     var locale = LocalizationUtility.GetLocale(repository, options);
 
-                    using (var restoreGitMap = RestoreGitMap.Create(docsetPath, locale))
+                    using (restoreGitMap = RestoreGitMap.Create(docsetPath, locale))
```
OsmondJiang: if we can use fulll path of docset path as the restore file path hash, we can avoid the locale, then we can break the cycle.

src/docfx/build/Build.cs
```diff
@@ -36,33 +37,23 @@ private static async Task BuildDocset(string docsetPath, string outputPath, Comm
...
                     {
-                        // load configuration from current docset and fallback docset
                         var input = new Input(docsetPath, repositoryProvider);
                         var configLoader = new ConfigLoader(docsetPath, input, repositoryProvider);
```
OsmondJiang: Config loader depends on DocsetPath, not input

src/docfx/build/Build.cs
live1206: https://github.com/dotnet/docfx/blob/b135d604cc5bc0f778cba42e3f0c9074cd6bde3b/src/docfx/config/ConfigLoader.cs#L111
`ConfigLoader` still depends on `RepositoryProvider` though

src/docfx/build/Build.cs
live1206: thanks for the hints, will try these ideas out in later PR

PR #5248, opened 2019-10-31
Support resolving links starting with ~

yufeih:
With true multi-docset support, we can replace `resolveAlias` with true `~` support.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5248)

### DIFF
--- a/src/docfx/build/dependency/LinkResolver.cs
+++ b/src/docfx/build/dependency/LinkResolver.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using Microsoft.DocAsCode.MarkdigEngine.Extensions;
 
 namespace Microsoft.Docs.Build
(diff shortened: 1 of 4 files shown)


src/docfx/build/dependency/LinkResolver.cs
```diff
@@ -226,9 +224,19 @@ internal class LinkResolver
         private Document TryResolveRelativePath(Document referencingFile, string relativePath, bool lookupFallbackCommits)
         {
             FilePath path;
+            string pathToDocset;
 
-            // apply resolve alias
-            var pathToDocset = ApplyResolveAlias(referencingFile, relativePath);
+            if (relativePath.StartsWith("~/") || relativePath.StartsWith("~\\"))
```
live1206: add a test with `~\\`, `relativePath.Substring(2)` would not trim properly

PR #5250, opened 2019-10-31
metadata value should be directly passed through

928PJY:
Fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/121851

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5250)

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -175,20 +175,29 @@ outputs:
   docs/a.json: |
     { "key": "another value" }
 ---
-# String values are trimmed in config
+# Metadatata value should be directly passed through
+# Redirect URL should be trimmed
 inputs:
   docfx.yml: |
     globalMetadata:
       key: ' value   '
     redirectionsWithoutId:
       docs/redirect.md: '/absolute/path '
-  docs/a.md: |
-    [redirect](redirect.md?a#b)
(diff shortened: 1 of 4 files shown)


docs/specs/config.yml
```diff
@@ -175,20 +175,29 @@ outputs:
...
   docs/a.json: |
     {
-      "key": "value",
-      "conceptual": "<p><a href=\"redirect?a#b\">redirect</a></p>"
+      "key": " value   "
+    }
+  .publish.json: |
+    {
+      "files": [
+        {
```
yufeih: nit: single line
```suggestion
        { source_path: xxx, redirect_url: xxx },          
```

src/docfx/config/ConfigLoader.cs
```diff
@@ -153,8 +153,6 @@ private static (List<Error>, JObject) LoadConfigObject(string fileName, string c
                 (errors, config) = JsonUtility.Parse(content, new FilePath(fileName));
             }
```
yufeih: also remove the implementation of `TrimStringValues`

PR #5249, opened 2019-10-31
Support .openpublishing.redirection.json

yufeih:
Support `.openpublishing.redirection.json` config at repo root.

For the existing config, moved `redirections` + `redirectionsWithoutId` outside `docfx.yml` to a dedicated file at docset path (`redirections.yml`). There is no such file in DHS so we can safely specialize this file name. This file shares the same schema as `.openpublishing.redirection.json`. A shorthand object form is added to make it less verbose. `redirections/renames` pair is chosen over the existing `redirectionsWithoutId/redirections` pair.

`source_path` is defined and rebased relative to the location of declaring file.

Source info can be reported correct for `.openpublishing.redirection.json` now.
 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5249)

PR #5253, opened 2019-11-01
Fix JsonUtility.Deserialize not having source info

yufeih:
Found in #5249 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5253)


src/docfx/lib/json/JsonUtility.cs
```diff
@@ -125,6 +125,11 @@ public static T Deserialize<T>(TextReader json, FilePath file)
             {
                 try
                 {
+                    var errors = new List<Error>();
```
mingwli: >var errors = new List<Error>(); [](start = 20, length = 31)

errors?

PR #5259, opened 2019-11-05
update markdig version

yishengjin1413:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5259)


yishengjin1413:
@yufeih  Checked V3 code, seems V3 did not use markdig directly. So may be we can update markdigexntension package aftrer V2 have a new package published?

yufeih:
> @yufeih Checked V3 code, seems V3 did not use markdig directly. So may be we can update markdigexntension package aftrer V2 have a new package published?

yep, that'll do

PR #5260, opened 2019-11-05
[Breaking] Move global metadata, file metadata config migration to docfx

yufeih:
https://dev.azure.com/ceapex/Engineering/_git/Docs.Build/pullrequest/17864?_a=overview

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5260)

PR #5268, opened 2019-11-07
Populate source info to extension data

yufeih:
The source info for property keys used to be stored in `JProperty`, this creates a challenge to propagate property name source info for extension data due to the limitation of `ExtensionDataSetter` (`setExtensinoData(string key, object value)`) does not have a place to store property name source info, it only has room for property value source info).

This PR moves property key source info from `JProperty` to `JProperty.Value`, to retrieve key source info, use `JsonUtility.GetKeySourceInfo`. 

It also fixed a bug where source info was reported incorrectly for `fileMetadata` due to property name transformation.

This is needed for #5260 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5268)

### DIFF
--- a/src/docfx/lib/json/JsonUtility.cs
+++ b/src/docfx/lib/json/JsonUtility.cs
@@ -56,7 +56,7 @@ namespace Microsoft.Docs.Build
 
         internal static JsonSerializer Serializer => s_serializer;
 
-        internal static Status State => t_status.Value.Peek();
+        internal static Status State => t_status.Value.TryPeek(out var result) ? result : null;
 
         static JsonUtility()
         {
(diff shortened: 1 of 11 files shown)


src/docfx/lib/json/JsonUtility.cs
```diff
@@ -350,6 +345,21 @@ internal static JToken SetSourceInfo(JToken token, SourceInfo source)
             return token;
         }
 
+        public static SourceInfo GetKeySourceInfo(JToken token)
+        {
+            return token.Annotation<SourceInfo>()?.KeySourceInfo;
+        }
+
+        public static JToken SetKeySourceInfo(JToken token, SourceInfo source)
```
live1206: is it possible to limit the input to `JProperty` instead of `JToken`?

PR #5272, opened 2019-11-08
⚒ RedirectionMap --> RedirectionProvider

yufeih:
Refactor only change to turn `RedirectionMap` into a self contained, constructor inject able component that takes `FilePath` in public interface.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5272)


src/docfx/build/document/Document.cs
```diff
@@ -320,6 +301,17 @@ internal static string PathToRelativeUrl(string path, ContentType contentType, s
             }
         }
 
+        private static ContentType GetContentType(Docset docset, FilePath path)
```
928PJY: confuse with above function with the same name `GetContentType`, maybe you can merge them into one function, or make the function name more specific

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -0,0 +1,222 @@
...
+                    if (!renameHistory.TryAdd(candidate.FilePath, file))
+                    {
+                        _errorLog.Write(Errors.RedirectionUrlConflict(item.RedirectUrl));
+                    }
+                }
+            }
+            return renameHistory;
+        }
+
+        private static string RemoveTrailingIndex(string redirectionUrl)
```
928PJY: >RemoveTrailingIndex [](start = 30, length = 19)

you have removed the query and fragment, not only trailing index

src/docfx/build/document/Document.cs
yufeih: yep, it could be combined to one

PR #5270, opened 2019-11-08
[Breaking] Support dependent repos config in docfx

yufeih:
https://dev.azure.com/ceapex/Engineering/_git/Docs.Build/pullrequest/17983?_a=overview

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5270)

PR #5276, opened 2019-11-11
Turn dependencies to use PathString

yufeih:
Add `PathString` to represent a normalized file path, directory path or URL path. This strong class ensures a path is normalized across the board.

Turns dependencies and `FilePath` to use `PathString`, other properties like `Routes`, `docsetPath` are covered later to make this PR short.

This is to systematically solve the path normalization problem through strong class declaration. The problem is found when doing #5270 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5276)


yufeih:
diffs are expected `.dependencymap` are now sorted using case sensitive comparison.

PR #5277, opened 2019-11-11
Skip yaml file validation

hejingkan2005:
- validate .md file only

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5277)


yufeih:
Can you add a spec test that covers:

1. yaml files are not validated
2. markdown files included by yaml files are validated

hejingkan2005:
> Can you add a spec test that covers:
> 
> 1. yaml files are not validated
> 2. markdown files included by yaml files are validated

Sure, done.

docs/specs/validation.yml
```diff
@@ -340,3 +340,74 @@ outputs:
...
+    {
+      "headings": {
+        "name": "Headings",
+        "description": "Validates H1 and other heading content",
+        "aliases": null,
+        "rules": [
+          {
+            "type": "H1Missing",
+            "message": "H1 is required. Use a single hash (#) followed by a space to create your top-level heading.",
+            "exclusions": [ "toc", "includes", "hubpage", "landingpage" ],
```
yufeih: Do we want respect this `exclusions` property or blindly exclude all markdown files? If there is no intend to support this property, consider removing it from the test case, as it brings confusion.

hejingkan2005: > Do we want respect this `exclusions` property or blindly exclude all markdown files? If there is no intend to support this property, consider removing it from the test case, as it brings confusion.

Done.

PR #5278, opened 2019-11-11
Refine exit code

live1206:
0 for `no error`
1 for `user error`
-99999 for `system error`

https://dev.azure.com/ceapex/Engineering/_workitems/edit/135737

Docs pipeline:
https://ceapex.visualstudio.com/Engineering/_git/Docs.Build/pullrequest/18026?_a=overview

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5278)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -22,10 +22,10 @@ namespace Microsoft.Docs.Build
             }
 
             var result = await Task.WhenAll(docsets.Select(docset => BuildDocset(docset.docsetPath, docset.outputPath, options)));
-            return result.All(x => x == 0) ? 0 : 1;
+            return result.All(x => x) ? 0 : 1;
         }
 
-        private static async Task<int> BuildDocset(string docsetPath, string outputPath, CommandLineOptions options)
+        private static async Task<bool> BuildDocset(string docsetPath, string outputPath, CommandLineOptions options)
         {
             List<Error> errors;
             Config config = null;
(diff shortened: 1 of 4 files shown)


src/docfx/build/Build.cs
```diff
@@ -12,19 +12,20 @@ namespace Microsoft.Docs.Build
 {
     internal static class Build
     {
-        public static Task Run(string workingDirectory, CommandLineOptions options)
+        public static async Task<int> Run(string workingDirectory, CommandLineOptions options)
```
yufeih: consider use `bool` here to represent `hasError`
```suggestion
        public static async Task<bool> Run(string workingDirectory, CommandLineOptions options)
```

live1206: The exit code can be used directly for the caller, will update the inner method to return `Task<bool>` to identify execution succeeded.

src/docfx/cli/Docfx.cs
```diff
@@ -35,8 +35,6 @@ internal static async Task<int> Main(params string[] args)
                 {
                 }
 
-                // An exit code of 0 means docfx completed successfully, including user errors.
-                // A non zero exit code means docfx encountered expected errors due to a bug.
                 return 1;
```
fenxu: >1; [](start = 23, length = 2)

This should return -1 if it's a system error.

PR #5283, opened 2019-11-12
Use PathString for routes

yufeih:
Add `PathString.StartsWithPath` method to replace `PathUtility.Match` and use it on `routes`.

Based on #5279 

Follow up on #5276 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5283)


docs/specs/config.yml
```diff
@@ -202,7 +202,7 @@ outputs:
 inputs:
   docfx.yml: |
     routes:
-      /: outputB
+      ./: outputB
```
yufeih: `/` is the root drive on linux (similar to `C:/` on windows), thus this needs to be `.` or `./`

src/docfx/lib/path/PathUtility.cs
```diff
@@ -52,7 +53,7 @@ public static (bool match, bool isFileMatch, string remainingPath) Match(this st
                 return (true, false, file.Substring(matcher.Length).Replace('\\', '/'));
             }
 
-            return default;
+            return (false, false, "");
```
live1206: nit: can `remainingPath` just be `null`? same as  `return default`

src/docfx/lib/path/PathString.cs
```diff
@@ -13,19 +14,32 @@ namespace Microsoft.Docs.Build
...
+        /// A nullable string that can never contains
+        ///     - backslashes
+        ///     - consegtive dots
+        ///     - consegtive forward slashes
+        ///     - leading ./
+        /// </summary>
+        private string _value;
 
-        public PathString(string value) => Value = PathUtility.Normalize(value);
+        public string Value => _value ?? "";
```
live1206: it says `nullable` string above, why do we need the null fallback here?

src/docfx/lib/path/PathString.cs
```diff
@@ -35,13 +49,83 @@ namespace Microsoft.Docs.Build
 
         public static implicit operator string(PathString value) => value.Value;
 
+        /// <summary>
+        /// Concat two <see cref="PathString"/>s together.
+        /// </summary>
+        public static PathString operator +(PathString a, PathString b)
+        {
+            if (string.IsNullOrEmpty(a._value))
```
live1206: nit: can just use `a.IsEmpty`, same for b

src/docfx/lib/path/PathString.cs
```diff
@@ -35,13 +49,83 @@ namespace Microsoft.Docs.Build
...
+
+        /// <summary>
+        /// Check if the file is the same as matcher or is inside the directory specified by matcher.
+        /// </summary>
+        public bool StartsWithPath(PathString basePath, out PathString remainingPath)
+        {
+            var basePathValue = basePath.Value;
+            var pathValue = Value;
+
+            if (basePathValue.Length == 0)
```
live1206: can `basePathValue` be null? how about just use `basePath.IsEmpty`?

src/docfx/lib/path/PathString.cs
```diff
@@ -35,13 +49,83 @@ namespace Microsoft.Docs.Build
...
+        }
+
         private class PathStringJsonConverter : JsonConverter
         {
             public override bool CanConvert(Type objectType) => objectType == typeof(PathString);
 
             public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
             {
-                var value = reader.ReadAsString();
+                var value = serializer.Deserialize<string>(reader);
```
live1206: what is the difference compared to the original?

yufeih: The original one throws when `PathString` is a dictionary value, because `ReadAsString`  throws when the current JsonReader is reading a property. `serializer.Deserialize<T>` resolves that.

PR #5290, opened 2019-11-14
[.links.json]Exclude non-build toc files from .links.json

live1206:
Follow up on #5282 

The included TOC files are not explicitly excluded from build in config, but excluded during the build process:
https://github.com/dotnet/docfx/blob/b6fb6518d42359c37443e173e0a54a018461c9f7/src/docfx/build/Build.cs#L208

https://dev.azure.com/ceapex/Engineering/_workitems/edit/139450

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5290)

### DIFF
--- a/docs/specs/output.yml
+++ b/docs/specs/output.yml
@@ -515,8 +515,6 @@ outputs:
 # source_url should be the includer of TOC
 inputs:
   docfx.yml: |
-    exclude: 
-      - docs/a/TOC.yml
     globalMetadata:
       "breadcrumb_path": "/docs/breadcrumbs/toc.json"
   docs/breadcrumbs/TOC.yml: |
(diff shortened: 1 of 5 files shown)


yufeih:
👍 This leads me to think that we need to exclude all files that are not published, whether it is due to TOC been include or markdown errors. We have a `PublishModelBuilder.MarkError` method that reliably marks content that is build but not published (except for included TOCs), may be we can use the same pipeline here.

live1206:
> 👍 This leads me to think that we need to exclude all files that are not published, whether it is due to TOC been include or markdown errors. We have a `PublishModelBuilder.MarkError` method that reliably marks content that is build but not published (except for included TOCs), may be we can use the same pipeline here.

Will refine the process in later PR

PR #5295, opened 2019-11-15
Refine how to filter outputs for .links.json

live1206:
Use `PublishModelBuilder.ExcludeFromOutput`

Follow up on #5290 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5295)


src/docfx/build/Build.cs
```diff
@@ -207,6 +206,7 @@ private static bool ShouldBuildFile(Context context, Document file)
             {
                 if (!context.TocMap.Contains(file))
                 {
+                    context.PublishModelBuilder.ExcludeFromOutput(file);
```
yufeih: should this line be moved to line 161?

live1206: sounds good

PR #5301, opened 2019-11-19
Redirect url support starting with https://hostName

mingwli:
While redirect_document_id, support redirect-url starting with `https://hostName`.

Fixing: [DevOps#141966](https://dev.azure.com/ceapex/Engineering/_workitems/edit/141966)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5301)

### DIFF
--- a/docs/specs/redirection.yml
+++ b/docs/specs/redirection.yml
@@ -67,6 +67,18 @@ inputs:
       b.md: a
 outputs:
 ---
+# redirect to can start with https and hostname
+inputs:
+  docfx.yml: |
+    baseUrl: https://docs.microsoft.com/
+  b.md:
+  redirections.yml: |
+    renames:
+      a.md: https://docs.microsoft.com/b
+outputs:
+  b.json: |
+   {"document_id": "780484e9-a367-97ab-ba7d-a19f3b378a80", "document_version_independent_id": "e9b3f4e2-1a78-4b5d-fbb4-03e0a33e38f5"}
+---
 # Redirection files have conflicted ids
 # all redirection files will be built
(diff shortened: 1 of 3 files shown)


src/docfx/build/context/Context.cs
```diff
@@ -57,7 +57,7 @@ public Context(string outputPath, ErrorLog errorLog, Docset docset, Docset fallb
             DocumentProvider = new DocumentProvider(docset, fallbackDocset, BuildScope, input, repositoryProvider, TemplateEngine);
             MetadataProvider = new MetadataProvider(docset, Input, MicrosoftGraphCache, restoreFileMap, DocumentProvider);
             MonikerProvider = new MonikerProvider(Config, BuildScope, MetadataProvider, restoreFileMap);
-            RedirectionProvider = new RedirectionProvider(docset.DocsetPath, ErrorLog, BuildScope, DocumentProvider, MonikerProvider);
+            RedirectionProvider = new RedirectionProvider(docset.DocsetPath, docset.HostName, ErrorLog, BuildScope, DocumentProvider, MonikerProvider);
```
live1206: is this `HostName` correct here? should it be the same as publish host name?

docs/specs/redirection.yml
```diff
@@ -67,6 +67,18 @@ inputs:
       b.md: a
 outputs:
 ---
+# redirect to can start with https and hostname
```
live1206: is `https` part of `hostname`?

mingwli: Yes, it is.

PR #5303, opened 2019-11-19
change the error level of yaml-duplicate-key moniker-range-undefined and merge-conflict to suggestion

928PJY:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/141952

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5303)

### DIFF
--- a/docs/specs/validation.yml
+++ b/docs/specs/validation.yml
@@ -12,7 +12,7 @@ inputs:
 outputs:
   docs/a.json: |
     {
-      "conceptual": "<p>bar</p><blockquote><blockquote><blockquote><blockquote><blockquote><blockquote><blockquote><p>cb1abc6bd98cfc84317f8aa95a7662815417802d</p></blockquote></blockquote></blockquote></blockquote></blockquote></blockquote></blockquote>"
+      "conceptual": ""
     }
   .errors.log: |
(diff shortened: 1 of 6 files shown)


docs/specs/validation.yml
```diff
@@ -10,9 +10,12 @@ inputs:
     bar
     >>>>>>> cb1abc6bd98cfc84317f8aa95a7662815417802d
 outputs:
-  docs/a.json:
+  docs/a.json: |
+    {
+      "conceptual": "<p>bar</p><blockquote><blockquote><blockquote><blockquote><blockquote><blockquote><blockquote><p>cb1abc6bd98cfc84317f8aa95a7662815417802d</p></blockquote></blockquote></blockquote></blockquote></blockquote></blockquote></blockquote>"
```
live1206: do we need to verify the content here?

docs/specs/validation.yml
```diff
@@ -10,9 +10,12 @@ inputs:
     bar
     >>>>>>> cb1abc6bd98cfc84317f8aa95a7662815417802d
 outputs:
-  docs/a.json:
+  docs/a.json: |
+    {
+      "conceptual": ""
```
yufeih: Should we still have output content for this test?

928PJY: yes, we still have output, but we don't need to verify that

PR #5307, opened 2019-11-20
Trim leading locale segment in redirect url

mingwli:
- try trim leading locale when redirect document id

fixing [DevOps#141966](https://dev.azure.com/ceapex/Engineering/_workitems/edit/141966)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5307)

### DIFF
--- a/src/docfx/build/redirection/RedirectionProvider.cs
+++ b/src/docfx/build/redirection/RedirectionProvider.cs
@@ -230,14 +230,15 @@ namespace Microsoft.Docs.Build
                 return redirectionUrl;
             }
 
-            if (redirectPath.IndexOf('/') is int slashIndex && slashIndex < 0)
+            int slashIndex = redirectPath.IndexOf('/');
+            if (redirectPath.IndexOf('/') < 0)
             {
                 return $"/{redirectPath}";
             }
 
             var firstSegment = redirectPath.Substring(0, slashIndex);
             return LocalizationUtility.IsValidLocale(firstSegment)
-                ? $"{redirectPath.Substring(firstSegment.Length)}"
+                ? $"/{redirectPath.Substring(firstSegment.Length)}"
                 : $"/{redirectPath}";
         }
     }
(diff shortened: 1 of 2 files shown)


src/docfx/localization/LocalizationUtility.cs
```diff
@@ -193,6 +193,29 @@ public static bool TryRemoveLocale(string name, out string nameWithoutLocale, ou
             return false;
         }
 
+        public static bool TryTrimPathLeadingLocale(PathString path, out PathString pathWithoutLocale, out string locale)
```
yufeih:
```suggestion
        public static bool TryTrimPathLeadingLocale(string path, out string pathWithoutLocale, out string locale)
```

src/docfx/localization/LocalizationUtility.cs
```diff
@@ -193,6 +193,29 @@ public static bool TryRemoveLocale(string name, out string nameWithoutLocale, ou
...
+        {
+            pathWithoutLocale = default;
+            locale = null;
+            if (string.IsNullOrEmpty(path))
+            {
+                return false;
+            }
+
+            var isAbsolute = path.Value[0] == '/';
```
yufeih: when will path start with `/`?

mingwli: Currently it won't, will limit this functionality inside redirection provider

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -225,5 +221,24 @@ private static string RemoveTrailingIndex(string redirectionUrl)
...
+
+        private static string RemoveLeadingLocale(string redirectionUrl, string hostName)
+        {
+            var (redirectHostName, redirectPath) = UrlUtility.SplitBaseUrl(redirectionUrl);
+            if (!string.Equals(redirectHostName, hostName, StringComparison.OrdinalIgnoreCase))
+            {
+                return redirectionUrl;
+            }
+
+            if (redirectPath.IndexOf('/') is int slashIndex && slashIndex < 0)
```
live1206: isn't `redirectPath.IndexOf('/')` always `int`?

mingwli: 'slashIndex' is used later, and IndexOf doesn't have a 'out int index' overload.

PR #5317, opened 2019-11-25
Fix TOC inclusion fallback order

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/145742

Fallback order for folder reference of TOC inclusion:
TOC.md -> TOC.json -> TOC.yml

If the build is for localization repo, the existing file should have higher priority than the deleted file from git history.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5317)

### DIFF
--- a/src/docfx/lib/path/FilePath.cs
+++ b/src/docfx/lib/path/FilePath.cs
@@ -12,6 +12,11 @@ namespace Microsoft.Docs.Build
     internal class FilePath : IEquatable<FilePath>, IComparable<FilePath>
     {
         /// <summary>
+        /// Indicate if the file is from git commit history.
+        /// </summary>
+        public bool IsFromHistory => Commit != null;
+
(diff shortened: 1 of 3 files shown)


src/docfx/lib/path/FilePath.cs
```diff
@@ -11,6 +11,11 @@ namespace Microsoft.Docs.Build
     /// </summary>
     internal class FilePath : IEquatable<FilePath>, IComparable<FilePath>
     {
+        /// <summary>
+        /// Indicate if the file is from git commit history.
+        /// </summary>
+        public bool IsFromHistory => Commit != null;
```
yufeih: nit: suggest to put auto property after field definition

PR #5321, opened 2019-11-26
Add non-blocking JsonDiskCache and use it for msalias validation

yufeih:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/144985/

The current msgraph and github cache are sync blocking cache, when items expires, the caller is blocked until the underlying API returns, this could starve threadpool.

This PR creates a `JsonDiskCache` that internally is a non-blocking async cache, and uses this cache against msgraph API. Due to the complexity of github user cache, it'll be handled in next PR.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5321)


src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -0,0 +1,117 @@
...
+            {
+                if (value.Expiry != null && value.Expiry < DateTime.UtcNow)
+                {
+                    // When the item expired, trigger background update but don't wait for the result
+                    _backgroundUpdates.GetOrAdd(key, UpdateDelegate(key, valueFactory)).Value.GetAwaiter();
+                }
+                return (default, value);
+            }
+
+            return _initialRequests.GetOrAdd(key, UpdateDelegate(key, valueFactory)).Value.GetAwaiter().GetResult();
```
live1206: seems `_initialRequests` is not needed? just `return UpdateDelegate(key, valueFactory).Value.GetAwaiter().GetResult();`

src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -0,0 +1,117 @@
...
+                    foreach (var cacheKey in item.GetKeys())
+                    {
+                        _cache.TryAdd(cacheKey, item);
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Gets or update the cache asynchroniously.
```
live1206:
```suggestion
        /// Get or update the cache asynchronously.
```

src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -0,0 +1,117 @@
...
+            if (_needUpdate)
+            {
+                var content = JsonUtility.Serialize(new { items = _cache.Values });
+
+                Directory.CreateDirectory(Path.GetDirectoryName(_cachePath));
+                ProcessUtility.WriteFile(_cachePath, content);
+                _needUpdate = false;
+            }
+
+            return result.Select(item => item.error).ToArray();
```
live1206: filter out the null errors from result

src/docfx/lib/cache/JsonDiskCache.cs
yufeih: This is to ensure only one call to `valueFactory` is made when multiple concurrent calls to `GetOrAdd` burst in for the same key.

But `_initialRequests` and `_backgroundUpdates` can be merged into one

src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -0,0 +1,121 @@
...
+                    foreach (var cacheKey in item.GetKeys())
+                    {
+                        _cache.TryAdd(cacheKey, item);
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Gets an item from the cache asynchroniously, or creates the value if it does not exist.
```
live1206: typo: asynchroniously -> asynchronously

src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -0,0 +1,121 @@
...
+        /// The <paramref name="valueFactory"/> can be a long running asynchronious call,
```
live1206: typo

PR #5324, opened 2019-11-27
142096 code extension

adunndevster:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5324)


test/Microsoft.DocAsCode.MarkdigEngine.Tests/CodeTest.cs
```diff
@@ -0,0 +1,603 @@
+// Copyright (c) Microsoft. All rights reserved.
```
yufeih: Tests in `MarkdigEngine.Extensions.Tests` project are shared between docfx v2 and docfx v3, while tests in `MarkdigEngine.Tests` only covers docfx v2. Consider move this test to `MarkdigEngine.Extensions.Tests` unless it is tightly coupled with v2.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+                htmlAttributes.AddProperty("data-interactive", language);
+                htmlAttributes.AddProperty("data-interactive-mode", interactive);
+            }
+            if (!string.IsNullOrEmpty(highlight)) htmlAttributes.AddProperty("highlight-lines", highlight);
+
+            RenderDelegate = (renderer, obj) =>
+            {
+                renderer.WriteLine("<pre>");
+                renderer.Write("<code").WriteAttributes(obj).Write(">");
+                renderer.WriteLine(updatedCodes.FirstOrDefault());
```
yufeih: In v3, this `CodeExtension` will only have one global instance, and the `ITripleColonExtensionInfo.Render`, `ITripleColonExtensionInfo.TryProcessAttributes` methods will be called on the same instance for many articles in parallel from different threads. The current implementation of `RenderDelegate` captures and mutates the `updatedCodes` variable, this can cause bugs where `updatedCodes` of one article is used to render another article

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+            range = range.Replace(" ", "");
+            var ranges = range.Split(',');
+            foreach (var codeRange in ranges)
+            {
+                if (codeRange.Contains("-"))
+                {
+                    var rangeParts = codeRange.Split('-');
+                    if (!string.IsNullOrEmpty(rangeParts[1]))
+                    {
+                        var beg = Convert.ToInt16(rangeParts[0]) - 1;
```
yufeih: `Convert.ToInt16` throws an exception and causes build crash if the input is not a number, try use `int.TryParse` and log a error if the user writes something like `abc-def`

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+
+                return true;
+            };
+
+            return true;
+        }
+
+        private string InferLanguageFromFile(string source, Action<string> logError)
+        {
+            var fileExtension = source.Split('.').ToList().LastOrDefault();
```
yufeih:
```suggestion
            var fileExtension = Path.GetExtension(source);
```

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+            if(fileExtension == null)
+            {
+                logError($"Language is not set, and your source has no file type. Cannot infer language.");
+            }
+            var language = HtmlCodeSnippetRenderer.LanguageAlias.Where(oo => oo.Value.Contains(fileExtension) || oo.Value.Contains($".{fileExtension}")).FirstOrDefault();
+            return language.Key;
```
yufeih: `FirstOrDefault` returns `null` and can cause a `NullReferenceException` in line 137 if the source extension does not exist in our supported list. Consider adding a test case to verify that.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+            }
+            return dedentedSections;
+        }
+
+        private static List<string> GetCodeSectionsFromRange(string range, string code, List<string> codeSections, Action<string> logError)
+        {
+            var codeLines = code.Split('\n');
+
+            range = range.Replace(" ", "");
+            var ranges = range.Split(',');
```
yufeih: with `RemoveEmptyEntries`, line 209 can be removed?
```suggestion
            var ranges = range.Split(new[] { "," }, StringSplitOptions.RemoveEmptyEntries)
```

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+                return false;
+            }
+            var (code, codePath) = _context.ReadFile(source, InclusionContext.File, null);
+            if(string.IsNullOrEmpty(code))
+            {
+                logError($"The code snippet \"{source}\" could not be found.");
+                return false;
+            }
+            
+            var updatedCode = GetCodeSnippet(range, id, code, logError).TrimEnd('\r').TrimEnd('\n');
```
yufeih:
```suggestion
            var updatedCode = GetCodeSnippet(range, id, code, logError).TrimEnd();
```

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+            if (!string.IsNullOrEmpty(range))
+            {
+                codeSections = GetCodeSectionsFromRange(range, code, codeSections, logError);
+            }
+            else if (!string.IsNullOrEmpty(id))
+            {
+                var strRegexStart = @"((<|</|#region)\s*" + id + @"(>|(\s*>)))";
+                var idRegexStart = new Regex(strRegexStart, RegexOptions.IgnoreCase);
+                var idRegexEnd = new Regex(strRegexStart + @"|#endregion", RegexOptions.IgnoreCase);
+                var codeLines = code.Split('\n').ToList();
```
yufeih: The `ToList` call duplicates the array and is not needed:

```suggestion
                var codeLines = code.Split('\n');
```

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,258 @@
...
+
+            if (!string.IsNullOrEmpty(range))
+            {
+                codeSections = GetCodeSectionsFromRange(range, code, codeSections, logError);
+            }
+            else if (!string.IsNullOrEmpty(id))
+            {
+                var strRegexStart = @"((<|</|#region)\s*" + id + @"(>|(\s*>)))";
+                var idRegexStart = new Regex(strRegexStart, RegexOptions.IgnoreCase);
+                var idRegexEnd = new Regex(strRegexStart + @"|#endregion", RegexOptions.IgnoreCase);
```
yufeih: Creating a Regex is a relatively expensive operation, consider promoting it to a static variable so it is created only once. Static members usually starts with `s_` prefix.

```suggestion
                private static readonly Regex s_idRegexEnd = new Regex(strRegexStart + @"|#endregion", RegexOptions.IgnoreCase);
```

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
yufeih: The `code` variable is split both here and in line 207, consider splitting it once because this variable can potentially be verify big, and splitting a big strings wasts memory and degrades performance.

yufeih: Do we need to consider `\r\n` line endings here?

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: It turns out that with int.TryParse, it's totally unnecessary to remove spaces :). I just needed to remove 209.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: It looks to me like ToList is needed, because it gives me FindIndex and FindLastIndex?

adunndevster: No, this works with both \n and \r\n endings.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: Yeah, I thought about that, but if you look at line 155, you will see that I need to dynamically create it with the id value. Do you know of a better way to do that?

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: Done!

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: Actually, I removed the need to use Regex :).

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: I found out about Array.FindIndex(). No more "ToList()"!

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: Done!

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: Done!

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: Thanks!

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: Done!

test/Microsoft.DocAsCode.MarkdigEngine.Tests/CodeTest.cs
adunndevster: @yufeih - fantastic catch! I moved and corrected the tests!

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
adunndevster: @yufeih - Really good insight, thank you. I made it so things are resolved at the right time and place inside the RenderDelegate.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,287 @@
...
+
+            RenderDelegate = (renderer, obj) =>
+            {
+                var currentId = string.Empty;
+                var currentRange = string.Empty;
+                var currentSource = string.Empty;
+                obj.Attributes.TryGetValue("id", out currentId); //it's okay if this is null
+                obj.Attributes.TryGetValue("range", out currentRange); //it's okay if this is null
+                obj.Attributes.TryGetValue("source", out currentSource); //source has already been checked above
+                var (code, codePath) = _context.ReadFile(currentSource, InclusionContext.File, null);
```
yufeih: Pass `obj` to `ReadFile` to produce the correct source info for warnings raised while resolving external files.
```suggestion
                var (code, codePath) = _context.ReadFile(currentSource, InclusionContext.File, obj);
```

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -0,0 +1,287 @@
...
+            htmlAttributes = new HtmlAttributes();
+            htmlAttributes.AddProperty("class", $"lang-{language}");
+            if (!string.IsNullOrEmpty(interactive))
+            {
+                htmlAttributes.AddProperty("data-interactive", language);
+                htmlAttributes.AddProperty("data-interactive-mode", interactive);
+            }
+            if (!string.IsNullOrEmpty(highlight)) htmlAttributes.AddProperty("highlight-lines", highlight);
+
+            RenderDelegate = (renderer, obj) =>
```
yufeih: I can see this is already a pattern used in triple colon parser, but putting the render logic inside a delegate makes it hard to ensure no variables are captured to this lambda.

The `logError` method in line 111 internally captures `processor.LineIndex` and `slice.Text` which could produce unexpected result in mult-threaded environment.

Try putting the logic directly inside the `Render` method.

PR #5326, opened 2019-11-27
Add dependency API interface spec

928PJY:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5326)


herohua:
@jipe@microsoft.com : can you add a comment in the documentation to summarize the conclusion we agreed upon from last review meeting?

928PJY:
Here is our conclusion:
1. `/buildconfig`: we will leverage the current OPS API `/v2/Queries/Docsets` to get the docset info in docfx, here are the remaining tasks:
   - [x] call API `/v2/Queries/Docsets` to fetch the docset info in docfx(@yufeih)
   - [x] remove the API in config migration tool. (@yufeih )
2. `/allfamiliesproductsmonikers`: as designed, this is an existed API, here are the remaining tasks:
   - [ ] add `monikerDefinition` item when loading the OPS config in docfx.
   - [ ] remove `monikerDefinition` when generating `docfx.global.json` in docs pipeline.
3. `/xrefmap/:tag`: we need to add new API to return xref maps by tags, the API interface should follow this design spec, and here are the remaining tasks:
   - [ ] Add a new API `/xrefmap/:tag`.(@partychen)
   - [ ] Consume this API in docfx
   - [ ] remove the `xref` config generate logic in docs pipeline
4. `/metadataValidation` & `/markdownValidation`: we will leverage the current API provided by metadata service. here is the remaining task:
   - [ ] Add a step in docfx to fetch the validation rules and convert it to json schema.
   - [ ] remove the logic in docs pipeline

5. According to the feedback of PM and content team, we need to support building the forked repository. so we need another API to get the source repository URL(the github API `/repos/:owner/:repo` can get the source repository information, but since we don't have github token, we need to add new OPS API). here is the remaining tasks:
   - [ ] Add a new OPS API to get the source repository URL.
   - [ ] call this API in extension to get the repository URL and pass it to docfx.

yufeih:
@928PJY , are you still working on this PR or should it be abandoned or merged?

PR #5333, opened 2019-12-02
⚒Refine localization

live1206:
- Add LocalizationProvider to save localization related status
- Break cyclic dependencies
- No need to switch build docset and fallback docset

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5333)

### DIFF
--- a/src/docfx/localization/LocalizationProvider.cs
+++ b/src/docfx/localization/LocalizationProvider.cs
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
 using System;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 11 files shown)


src/docfx/localization/LocalizationProvider.cs
```diff
@@ -0,0 +1,162 @@
...
+        private readonly bool _buildFromEnglish;
+        private readonly string _entryDocsetPath;
+        private readonly Repository _entryRepository;
+
+        private string _englishDocsetPath;
+        private Repository _englishRepository;
+        private string _localizationDocsetPath;
+        private Repository _localizationRepository;
+
+        public bool IsLocalizationBuild { get; private set; } = false;
```
yufeih:
```suggestion
        public bool IsLocalizationBuild { get; } = false;
```

PR #5334, opened 2019-12-02
⚒ Consolidate config load with FileResolver

yufeih:
Expose only 1 public method `ConfigLoader.Load`, it is internally a 2 stage load. A `PreloadConfig` object is created to represent properties needed for 2 stage load.

Refactored RestoreFileMap into `FileResolver` that is itself a self contained component and are shared between build and restore.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5334)


src/docfx/lib/markdown/MarkdownEngine.cs
```diff
@@ -47,7 +47,7 @@ internal class MarkdownEngine
             _markdownValidationRules = config.MarkdownValidationRules;
             if (!string.IsNullOrEmpty(_markdownValidationRules))
             {
-                using (var stream = restoreFileMap.ReadStream(config.MarkdownValidationRules))
+                using (var stream = fileDownloader.DownloadStream(config.MarkdownValidationRules))
```
live1206: `_noFetch` in `fileDownloader` is `true` here, it will skip the actual downloading?
Just wondering would it be better to separate the actual downloading and reading from local files?

src/docfx/config/ops/OpsConfigLoader.cs
```diff
@@ -10,7 +10,7 @@ namespace Microsoft.Docs.Build
 {
     internal static class OpsConfigLoader
     {
-        public static bool TryLoad(string docsetPath, string branch, out JObject result)
+        public static JObject TryLoad(string docsetPath, string branch)
```
live1206: should this method be renamed to `Load` since the return type is not `bool` anymore

yufeih: yep, that sounds reasonable.

PR #5349, opened 2019-12-09
PublishModel contains baseUrl

mingwli:
Pass baseUrl for migration tool to consume

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5349)

### DIFF
--- a/docs/specs/output.yml
+++ b/docs/specs/output.yml
@@ -21,6 +21,15 @@ outputs:
       ]
     }
 ---
+# Publish manifest contains config baseUrl
+inputs:
+  docfx.yml:
+outputs:
+  .publish.json: |
+    {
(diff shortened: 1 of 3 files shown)


docs/specs/output.yml
```diff
@@ -21,6 +21,15 @@ outputs:
       ]
     }
 ---
+# Publish manifest contains config baseUrl
+inputs:
+  docfx.yml:
+outputs:
+  .publish.json: |
+    {
+      "base_url": "https://docs.com"
```
yufeih: nit: is there a test case for these additional properties in .publish.json like name.

PR #5350, opened 2019-12-09
Turn circular-reference and include-not-found into error

yufeih:
Fixes https://ceapex.visualstudio.com/Engineering/_workitems/edit/150364?src=WorkItemMention&src-action=artifact_link

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5350)

### DIFF
--- a/src/docfx/lib/log/ErrorLog.cs
+++ b/src/docfx/lib/log/ErrorLog.cs
@@ -112,6 +112,11 @@ namespace Microsoft.Docs.Build
                 level = ErrorLevel.Error;
             }
 
+            if (error.Code == "circular-reference" || error.Code == "include-not-found")
+            {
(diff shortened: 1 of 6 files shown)


src/docfx/lib/log/ErrorLog.cs
```diff
@@ -112,6 +112,11 @@ public bool Write(Error error, bool isException = false)
                 level = ErrorLevel.Error;
             }
 
+            if (error.Code == "circular-reference" || error.Code == "include-not-found")
```
live1206: how about maintaining a static array of error codes instead?

PR #5346, opened 2019-12-06
[Breaking] Support docfx.json

yufeih:
Paired change https://dev.azure.com/ceapex/Engineering/_git/Docs.Build/pullrequest/18701?_a=overview

Related #5315

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5346)

PR #5351, opened 2019-12-09
Explicitly control which docsets to build

yufeih:
image

#5346 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5351)

### DIFF
--- a/test/docfx.Test/lib/PathUtilityTest.cs
+++ b/test/docfx.Test/lib/PathUtilityTest.cs
@@ -78,13 +78,9 @@ namespace Microsoft.Docs.Build
         [InlineData("a", "/", false, "")]
         public static void PathMatch(string file, string matcher, bool expectedMatch, string expectedRemainingPath)
         {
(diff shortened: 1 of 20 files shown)


test/docfx.Test/lib/PathUtilityTest.cs
```diff
@@ -78,17 +78,25 @@ public static void GetRelativePathToFile(string relativeTo, string path, string
...
-            Assert.Equal(expectedRemainingPath, remainingPath);
-
             var matches = new PathString(file).StartsWithPath(new PathString(matcher), out var remaining);
             Assert.Equal(expectedMatch, matches);
-            Assert.Equal(expectedRemainingPath, remaining);
+            Assert.Equal(new PathString(expectedRemainingPath), remaining);
+        }
+
+        [Theory]
+        [InlineData("", "", true)]
```
live1206: add some negative case here, otherwise can just use `Assert.True`

PR #5352, opened 2019-12-09
Pass basePath in publish model instead of baseUrl

mingwli:
1. break baseUrl docfx config into hostName and basePath
2. rename siteBasePath to basePath to be consistent with OP
3. pass basePath inside publish model instead of baseUrl to avoid repeated combine/split

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5352)

### DIFF
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -101,7 +101,9 @@ repos:
     files:
       docfx.yml: |
         template: https://docs.com/template1
-        baseUrl: https://docs.com/rawmetadata
+        hostName: https://docs.com
+        basePath: rawmetadata
+
       redirections.yml: |
         redirections:
           a.md: /b
@@ -178,7 +180,7 @@ outputs:
         "content_git_url": "https://github.com/a/rawmetadata/blob/master/c.md",
         "original_content_git_url": "https://github.com/a/rawmetadata/blob/master/c.md",
         "original_content_git_url_template": "{repo}/blob/{branch}/c.md",
-        "gitcommit": "https://github.com/a/rawmetadata/blob/e8f058141f57d856be5a0c86b55cba07e0a8bace/c.md",
+        "gitcommit": "https://github.com/a/rawmetadata/blob/9549588dee7108d93d0bee2c9bb1c5659dd3c842/c.md",
         "site_name": "Docs",
         "depot_name": ".",
         "search.ms_docsetname": "",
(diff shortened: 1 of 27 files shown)


yufeih:
I'd like to keep the input term consistent with the output term, for input we currently use `base_url`

live1206:
> I'd like to keep the input term consistent with the output term, for input we currently use `base_url`

That means we need to split `base url` while using it in Docs.Build and migration tool?

yufeih:
> > I'd like to keep the input term consistent with the output term, for input we currently use `base_url`
> 
> That means we need to split `base_url` while using it in Docs.Build and migration tool?

Or change input `base_url` to `base_path` and `host`

live1206:
> > > I'd like to keep the input term consistent with the output term, for input we currently use `base_url`
> > 
> > 
> > That means we need to split `base_url` while using it in Docs.Build and migration tool?
> 
> Or change input `base_url` to `base_path` and `host`
`host` is not consumed further, I agree with `base_path`

docs/specs/legacy.yml
```diff
@@ -326,7 +328,8 @@ inputs:
     monikerRange:
       'docs/v1/**': '< netcore-2.0'
       'docs/v2/**': 'netcore-2.0'
-    baseUrl: https://docs.com/docs
+    hostName: https://docs.com
```
yufeih: Technically `hostName` does not contain `https://`
```suggestion
    hostName: docs.com
```

docs/specs/legacy.yml
```diff
@@ -326,7 +328,8 @@ inputs:
...
+    basePath: docs
```
yufeih: `basePath` in docs starts with `/` but does not ends with `/` unless it is `/`, can ensure that in `Config.OnDeserialized` method

```suggestion
    basePath: /docs
```

docs/specs/output.yml
```diff
@@ -21,15 +21,17 @@ outputs:
       ]
     }
 ---
-# Publish manifest contains following properties: name, product, baseUrl
+# Publish manifest contains following properties: name, product, site_base_path
```
live1206:
```suggestion
# Publish manifest contains following properties: name, product, base_path
```

PR #5353, opened 2019-12-09
fix cardF className

v-pegao:
Compare with [v2 card list template](https://github.com/microsoft/templates.docs.msft/blob/master/ContentTemplate/LandingData.html.primary.js#L49-L53), we need add `panelContent ` to `cardF className` and avoid `Model.ClassName` being added to `className` twice

Find related workitem [here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/143283?src=WorkItemMention&src-action=artifact_link)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5353)

### DIFF
--- a/src/Microsoft.Docs.Template/Views/Template/BuildItemListPartial.cshtml
+++ b/src/Microsoft.Docs.Template/Views/Template/BuildItemListPartial.cshtml
@@ -91,15 +91,19 @@ else if (Model.Style == "cards")
             {
                 className = "panelContent " + Model.ClassName;
             }
+            else
+            {
+                className = Model.ClassName;
+            }
             if (Model.Columns is JValue value)
             {
                 if (int.TryParse(value.ToString(), out var i))
                 {
-                    className = Model.ClassName + " cols cols" + i.ToString();
+                    className += " cols cols" + i.ToString();
                 }
                 else
                 {
-                    className += Model.ClassName + " cols cols4";
+                    className += " cols cols4";
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


src/Microsoft.Docs.Template/Views/Template/BuildItemListPartial.cshtml
```diff
@@ -91,15 +91,19 @@ else if (Model.Style == "cards")
             {
                 className = "panelContent " + Model.ClassName;
             }
+            else
+            {
+                className = Model.ClassName;
+            }
             if (Model.Columns is JValue value)
```
v-pegao: @live1206 @928PJY I can see another difference between v2 and v3 here: if `Model.Columns` is undefined
- v2 will take `4` by default and append `cols cols4` to className
- v3 wil not append `cols cols` string to `className`

If we want to make the behavior same with v2, maybe we can merge the if conditions:
`if (Model.Columns is JValue value && int.TryParse(value.ToString(), out var i))`

live1206: looks good to me

v-pegao: @live1206 Updated. Could you please help review again?

live1206: The code looks good to me, deployed to PPE already, could you help verify the diff is gone? Thanks

PR #5356, opened 2019-12-10
[Breaking] Move validation, moniker definition config to docfx

yufeih:
https://dev.azure.com/ceapex/Engineering/_git/Docs.Build/pullrequest/18794

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5356)

### DIFF
--- a/src/docfx/config/ops/OpsConfigAdapter.cs
+++ b/src/docfx/config/ops/OpsConfigAdapter.cs
@@ -2,8 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Generic;
+using System.IO;
 using System.Linq;
+using System.Net;
 using System.Net.Http;
+using System.Threading.Tasks;
+using System.Web;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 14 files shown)


yufeih:
The azure-docs-pr diff is expected because of a bug in the previous implementation

src/docfx/restore/FileResolver.cs
```diff
@@ -178,7 +180,13 @@ private Task<HttpResponseMessage> GetAsync(string url, EntityTagHeaderValue etag
 
                 _provideCredential?.Invoke(message);
 
-                return s_httpClient.SendAsync(message);
+                var response = await OpsConfigAdapter.InterceptHttpRequest(_errorLog, message);
```
mingwli: >OpsConfigAdapter.InterceptHttpRequest( [](start = 37, length = 38)

minor: It feels that this bring extra buisiness into a lower-level functionality.

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -2,40 +2,78 @@
...
+        private const string MarkdownValidationRulesApi = "https://ops/markdownvalidationrules/";
+
+        private static readonly string s_opsToken = Environment.GetEnvironmentVariable("DOCS_OPS_TOKEN");
+        private static readonly IReadOnlyDictionary<string, string> s_opsHeaders = new Dictionary<string, string>
+        {
+            { "X-OP-BuildUserToken", s_opsToken },
+        };
+
+        private static readonly bool s_isProduction = !string.Equals(
             "PPE", Environment.GetEnvironmentVariable("DOCS_ENVIRONMENT"), StringComparison.OrdinalIgnoreCase);
```
mingwli: suggestion: better change to equals("prod") for checking is_production, considering passing 'sandbox' by mistake.

yufeih: The intention is to be `prod` by default, the logic is correct but not intuitive, maybe `is_sandbox` expressed it better.

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -47,18 +78,100 @@ public static JObject Load(FileResolver fileResolver, SourceInfo<string> name, s
...
+                {
+                    return new HttpResponseMessage { Content = new StringContent(await rule(errorLog, request.RequestUri)) };
+                }
+            }
+            return null;
+        }
+
+        private static Task<string> GetMonikerDefinition(ErrorLog errorLog, Uri url)
+        {
+            return Fetch($"{s_buildServiceEndpoint}/v2/monikertrees/allfamiliesproductsmonikers", s_opsHeaders);
```
live1206: why no exception handling for this call since we have for the others?

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -47,18 +78,100 @@ public static JObject Load(FileResolver fileResolver, SourceInfo<string> name, s
...
+                return "{}";
+            }
+        }
+
+        private static async Task<string> GetMetadataSchema(ErrorLog errorLog, Uri url)
+        {
+            try
+            {
+                var headers = GetValidationServiceHeaders(url);
+                var rules = Fetch($"{s_validationServiceEndpoint}/rules", headers);
```
live1206: is it possible to have two explicit get URL calls instead of having two implicit calls within one public get URL?

src/docfx/config/ops/OpsConfigAdapter.cs
yufeih: Validation service calls don't block build, when we failed to retrieve validation rule, there is a warning in the build report and validation is skipped. This behavior is the same as in Docs.Build

src/docfx/config/ops/OpsConfigAdapter.cs
yufeih: That would require changes outside of this adapter layer because docfx core only recognize one JSON file, the adapter feels the right place to bridge the gap between validation service APIs and docfx.

PR #5366, opened 2019-12-16
fix: changed post processor manager to use updatedMetadata

Onnno:
each call to postProcessor.PrepareMetadata(metadata) was overriding metadata manipulations of the former postProcessor

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5366)


Onnno:
This fix is necessary to enable me to publish the 'nested namespaces post processor' that generates
yaml files for upper level namespaces so the html output namespace nav bar shows a tree instead of a flat namespace list.

Onnno:
> `PostProcessorsManagers` changes makes sense to me, but I'm not sure if `DocumentBuilder.cs` changes are intentional or accidental. `AddNestedNamespaceFiles` seems to be a method from plugin.

Yes, you'r right. DocumentBuilder.cs was accidental. I reverted it.
That's another pull request I'll make in a moment.

PR #5369, opened 2019-12-17
Use extend config to restore ops service config

yufeih:
Without this, we call ops service to retrieve service config in both restore and build. All ops calls are gated behind the proxy behind `OpsConfigAdapter`, and use `FileResolver` can cache HTTP response for build.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5369)

### DIFF
--- a/src/docfx/config/ops/OpsConfigAdapter.cs
+++ b/src/docfx/config/ops/OpsConfigAdapter.cs
@@ -10,12 +10,12 @@ using System.Net.Http;
 using System.Threading.Tasks;
 using System.Web;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
     internal class OpsConfigAdapter : IDisposable
     {
+        public const string BuildConfigApi = "https://ops/buildconfig/";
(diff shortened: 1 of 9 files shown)


src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -166,35 +166,28 @@ private async Task<string> GetMetadataSchema(Uri url)
...
-                    }
-                    return await response.EnsureSuccessStatusCode().Content.ReadAsStringAsync();
+                if (nullOn404 && response.StatusCode == HttpStatusCode.NotFound)
+                {
+                    return null;
                 }
-            }
-            catch (Exception ex)
-            {
-                throw Errors.DownloadFailed(url).ToException(ex);
```
live1206: why is this exception handling removed?

yufeih: It will be handled by `FileResolver`, I added this catch in https://github.com/dotnet/docfx/pull/5365 because back then `GetBuildConfig` was called directly instead of hijacking http pipeline.

PR #5370, opened 2019-12-17
get default locale from ops api

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/153152

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5370)

### DIFF
--- a/src/docfx/config/ops/OpsConfigAdapter.cs
+++ b/src/docfx/config/ops/OpsConfigAdapter.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Docs.Build
 
             var docsets = JsonConvert.DeserializeAnonymousType(
                 docsetInfo,
-                new[] { new { name = "", base_path = "", site_name = "", product_name = "" } });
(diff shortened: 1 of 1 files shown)


src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -87,7 +87,7 @@ public async Task<JObject> GetBuildConfig(SourceInfo<string> name, string reposi
                 ["xrefHostName"] = GetXrefHostName(docset.site_name, branch),
                 ["localization"] = new JObject
                 {
-                    ["defaultLocale"] = GetDefaultLocale(docset.site_name),
+                    ["defaultLocale"] = docset.locale,
```
mingwli: this will overwrite all loc builds' defaultLocale.
Say we build 'azure-docs-pr.ja-jp' with defaultLocale 'ja-jp'

live1206: you are right

PR #5371, opened 2019-12-17
Postprocessor can add yml files before rendering html

Onnno:
This a feature request.

Background is the following:
Docfx generates a flat namespace navigation tree for html output, e.g.
System.IO and System.IO.Hardware will be listed as two separate namespaces.

Me and other users prefer a navigation tree instead, like this:
- System.IO
-- System.IO.Hardware

So I changed docfx to do so. It is the best way I could find, described below.
I want to publish this now so other users can use it.

My goal: Simple install of nested namespaces for everyone
1. update the docfx nuget package to next release including this pull request
2. install the post processor: https://github.com/billbeeio/docfx-namespace-tree as a nuget package (to be prepared)
3. if possible, the post processor will install the custom template I prepared, otherwise there will be instructions.

To achieve this goal, I need this pull request to get approved.
Please let me know if I should e.g. add additional checks for valid file pathes
or even if there is a better way to let docfx know about the additionaly yml files to be converted.

What I basically do to get nested namespaces:
1. I wrote a postprocessor plugin that inspects the yml files generated by docfx before
  they are converted to html/output format. Since docfx generates one yml file for each
  namespace containing a class or something, but not for namespaces containing only namespaces,
  I generate these myself and add them to the files generated by docfx. You can see my plugin code here: https://github.com/billbeeio/docfx-namespace-tree/blob/master/NamespacePostprocessor/PostProcessor.cs
2. To add those new files to the docfx internal list of yml files to be converted to html, I need
  the code from this pull request.
3. Finally, I override some of the renderer files for html as described here: https://dotnet.github.io/docfx/tutorial/howto_create_custom_template.html#merge-template-with-default-template

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5371)


Onnno:
Not done yet - I will reissue the pull request in some days.

PR #5376, opened 2019-12-18
Add ConfigMonikerRange to publish model

928PJY:
Fix: https://dev.azure.com/ceapex/Engineering/_workitems/edit/153166

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5376)


docs/specs/moniker.yml
```diff
@@ -194,8 +206,14 @@ outputs:
         {"url":"/docs/a","moniker_group":"9d4e15fd"}
       ],
       "moniker_groups":{
-        "4667fedf":["netcore-2.0"],
-        "9d4e15fd":["netcore-1.0","netcore-1.1"]
+        "4667fedf":{
+          "monikers": ["netcore-2.0"],
+          "config_moniker_range": ">= netcore-2.0"
```
herohua: `config` -> `configured`?

src/docfx/publish/PublishItem.cs
```diff
@@ -20,6 +20,8 @@ internal class PublishItem
 
         public string MonikerGroup { get; set; }
 
+        public string ConfigMonikerRange { get; set; }
```
yufeih:
```suggestion
        public string MonikerRange { get; set; }
```

src/docfx/publish/PublishModel.cs
```diff
@@ -18,6 +18,6 @@ internal class PublishModel
 
         public PublishItem[] Files { get; set; }
 
-        public IReadOnlyDictionary<string, IReadOnlyList<string>> MonikerGroups { get; set; } = new Dictionary<string, IReadOnlyList<string>>();
+        public IReadOnlyDictionary<string, MonikerGroup> MonikerGroups { get; set; } = new Dictionary<string, MonikerGroup>();
```
yufeih: If `moniker_range` is already in `PublishItem`, why duplicating it here?

928PJY: add a `[JsonIgnore]` attribute to PublishItem `ConfigMonikerRange` property

src/docfx/publish/PublishItem.cs
928PJY: `ConfigMonikerRange` should be better and no confuse since there are several source of monikerRange: config, metadata and yaml header.

docs/specs/moniker.yml
928PJY: the monikerRange from yaml header can also be treated as `configured`?

PR #5378, opened 2019-12-19
Use http config to pass ops token

yufeih:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5378)

### DIFF
--- a/src/docfx/config/PreloadConfig.cs
+++ b/src/docfx/config/PreloadConfig.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net.Http;
 using Newtonsoft.Json;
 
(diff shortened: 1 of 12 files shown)


src/docfx/config/PreloadConfig.cs
```diff
@@ -28,5 +29,21 @@ internal class PreloadConfig
         /// Gets the authorization keys for required resources access
         /// </summary>
         public readonly Dictionary<string, HttpConfig> Http = new Dictionary<string, HttpConfig>();
+
+        public void ProvideCredential(HttpRequestMessage message)
+        {
+            var url = message.RequestUri.ToString();
+            foreach (var (baseUrl, rule) in Http)
```
fenxu: >baseUrl [](start = 26, length = 7)

Consider to use longest match to avoid some conflicts and confusing to users?

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -20,12 +20,6 @@ internal class OpsConfigAdapter : IDisposable
         private const string MetadataSchemaApi = "https://ops/metadataschema/";
         private const string MarkdownValidationRulesApi = "https://ops/markdownvalidationrules/";
 
-        private static readonly string s_opsToken = Environment.GetEnvironmentVariable("DOCS_OPS_TOKEN");
```
live1206: Does this require docs-pipeline to set the http header for OPS urls?
Also remove https://ceapex.visualstudio.com/Engineering/_git/Docs.Build?path=%2Fsrc%2Fdocs-pipeline%2FDocsPipeline.cs&version=GBdevelop&line=56&lineEnd=57&lineStartColumn=1&lineEndColumn=1&lineStyle=plain

live1206: Test should be updated
https://github.com/dotnet/docfx/blob/91b3e55f1a3d28552baed293253542496944c418/docs/specs/config.yml#L527

yufeih: yep, need a new PR

yufeih:
Prefer https://github.com/dotnet/docfx/pull/5379

PR #5380, opened 2019-12-19
Output config base path as it is

live1206:
when base_path is `/`, the output `base_path` in `.publish.json` should be `/`

Found the issue in this [PR](https://ceapex.visualstudio.com/Engineering/_git/Docs.Build/pullrequest/18812)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5380)

### DIFF
--- a/src/docfx/lib/path/BasePath.cs
+++ b/src/docfx/lib/path/BasePath.cs
@@ -7,19 +7,24 @@ using Newtonsoft.Json;
 namespace Microsoft.Docs.Build
 {
     [JsonConverter(typeof(BasePathJsonConverter))]
-    internal class BasePath
+    internal readonly struct BasePath
     {
-        public string Original { get; set; }
+        private readonly string _original;
 
-        private string RelativePath
-            => Original.StartsWith('/') ? Original.TrimStart('/') : Original;
+        private readonly string _relativePath;
+
+        public string Original => _original ?? "";
+
+        // It is either an empty string, or a path without leading /
+        public string RelativePath => _relativePath ?? "";
 
         public BasePath(string value)
         {
-            Original = value;
+            _original = value;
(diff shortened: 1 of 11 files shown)


src/docfx/lib/path/BasePath.cs
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using Newtonsoft.Json;
+
+namespace Microsoft.Docs.Build
+{
+    [JsonConverter(typeof(BasePathJsonConverter))]
+    internal class BasePath
```
yufeih:
```suggestion
    internal struct BasePath
```

src/docfx/lib/path/BasePath.cs
```diff
@@ -0,0 +1,38 @@
...
+    {
+        public string Original { get; set; }
+
+        private string RelativePath
+            => Original.StartsWith('/') ? Original.TrimStart('/') : Original;
```
yufeih: return "." for empty path so callers can pass this to `Path` functions, set the value in constructor for better performance

yufeih: consider override `ToString() => Original`

live1206: > return "." for empty path so callers can pass this to `Path` functions, set the value in constructor for better performance

`Path.Combine(".", "test")` would output `.\test`

src/docfx/lib/path/BasePath.cs
live1206: Line 219 in `Config.cs` we are changing the value of `Original`

Issue #5011, opened 2019-08-21
Failing with UWP and Visual Studio 2019 (16.*)


phraemer:
**Operation System**: Windows

**DocFX Version Used**: 2.44.0.0

**Template used**: `default`

**Steps to Reproduce**:

I made a clean UWP project with Visual Studio and beside it did `docfx init -q` and made the few changes I believed necessary to point it at the project.

I have a repo of the entire thing here:
https://github.com/phraemer/TestDocfx

Not sure if the `TargetFramework` should be `net46` but previously it has worked with VS 2017 projects and older docfx versions like 2.40 or so.

**Expected Behavior**:

Generates the site.

**Actual Behavior**:

Errors out with:

```
λ docfx .\doc\docfx.json --serve
[19-08-21 01:21:47.893]Info:[MetadataCommand.ExtractMetadata]Environment variable VSINSTALLDIR is set to C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\, it is used as the inner compiler.
[19-08-21 01:21:48.059]Info:[MetadataCommand.ExtractMetadata]Loading projects...
[19-08-21 01:21:48.420]Warning:[MetadataCommand.ExtractMetadata](C:/Users/james/Source/Repos/TestDocfx/TestDocfx/TestDocfx.csproj)Error opening project C:/Users/james/Source/Repos/TestDocfx/TestDocfx/TestDocfx.csproj: Unable to load one or more of the requested types. Retrieve the LoaderExceptions property for more information.. Ignored.
[19-08-21 01:21:48.420]Info:[MetadataCommand]Completed Scope:MetadataCommand in 546.8394 milliseconds.
[19-08-21 01:21:48.420]Error:Error extracting metadata for C:/Users/james/Source/Repos/TestDocfx/TestDocfx/TestDocfx.csproj: System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.DocAsCode.Metadata.ManagedReference.ExtractMetadataWorker.<SaveAllMembersFromCacheAsync>d__13.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
[...]

PR #5382, opened 2019-12-20
fix: VisualStudioVersion including minor version breaks UWP project

superyyrrzz:
#5011
root cause: https://developercommunity.visualstudio.com/content/problem/862455/wrong-import-path-for-microsoftwindowsuixamlcsharp.html#comment-863292

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5382)


herohua:
I'd suggest adding a comment inline alongside the code.

PR #5385, opened 2019-12-20
refactor: one method to gather git dependencies

yufeih:
- Gather all git dependencies in `GetGitDependencies` method, the call just for each restore the result
- always `--prune`

Prep work for https://dev.azure.com/ceapex/Engineering/_workitems/edit/121829/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5385)

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Docs.Build
         /// </summary>
         public static void InitFetchCheckout(string path, string url, string committish, Config config = null)
         {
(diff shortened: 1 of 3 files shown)


live1206:
>         internal static IReadOnlyList<RestoreGitResult> Restore(

make method `public`

---
Refers to: src/docfx/restore/RestoreGit.cs:15 in 06b506d. [](commit_id = 06b506d9019b554f04b8340b4bb0e4adfa893187, deletion_comment = False)

PR #5388, opened 2019-12-23
Look up remote commit in case a branch does not have a local tracking branch

yufeih:
When docfx build runs a against a remote branch that does not have a local tracking branch, build throws `committish-not-found` error. This can happen when building an ops loc repo using sxs, where the sxs branch has a local tracking branch, but the non-sxs branch does not have a local branch name.

This is a regression of https://github.com/dotnet/docfx/pull/5385, before that change, there is an [explicit fetch](https://github.com/dotnet/docfx/pull/5385/files#diff-0e50288f9abd09bdcec034b69428527fL39)  that ensures none sxs branch also have a local tracking branch to workaround the problem. This PR solves the root cause by fallback to check remote commit if there is no local tracking branch.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5388)

### DIFF
--- a/src/docfx/lib/git/FileCommitProvider.cs
+++ b/src/docfx/lib/git/FileCommitProvider.cs
@@ -198,7 +198,8 @@ namespace Microsoft.Docs.Build
             git_revwalk_new(out var walk, _repo);
             git_revwalk_sorting(walk, 1 << 0 | 1 << 1 /* GIT_SORT_TOPOLOGICAL | GIT_SORT_TIME */);
 
-            if (git_revparse_single(out var headCommit, _repo, committish) != 0)
+            if (git_revparse_single(out var headCommit, _repo, committish) != 0 &&
+                git_revparse_single(out headCommit, _repo, $"origin/{committish}") != 0)
             {
                 git_object_free(walk);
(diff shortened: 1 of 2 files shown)


src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -198,7 +198,8 @@ private unsafe (List<Commit>, Dictionary<long, Commit>) LoadCommits(string commi
             git_revwalk_new(out var walk, _repo);
             git_revwalk_sorting(walk, 1 << 0 | 1 << 1 /* GIT_SORT_TOPOLOGICAL | GIT_SORT_TIME */);
 
-            if (git_revparse_single(out var headCommit, _repo, committish) != 0)
+            if (git_revparse_single(out var headCommit, _repo, committish) != 0 &&
+                git_revparse_single(out headCommit, _repo, $"origin/{committish}") != 0)
```
live1206: does `headCommit` always come from the remote branch then?

mingwli:
[DevOps#156319](https://ceapex.visualstudio.com/Engineering/_workitems/edit/156319)

PR #5391, opened 2019-12-24
fix: Fetch loc live branch to generate the correct contributor list

yufeih:
This is to fix a regression caused by https://github.com/dotnet/docfx/pull/5385, when building the sxs branch of a loc repo, we need the none-sxs branch history to compute the correct contributor list, but the cloning process of the loc repo is managed outside of docfx, and it may only clone the sxs branch content for performance reasons.

https://github.com/dotnet/docfx/pull/5388 appeared to have fixed the problem because we are in the transition to a more performant clone process, some test machines still preserve full history but in the future only the branch to build will be cloned.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5391)

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -131,13 +131,13 @@ namespace Microsoft.Docs.Build
 
             try
             {
-                ExecuteNonQuery(path, $"{httpConfig} fetch --tags --progress --update-head-ok --prune \"{url}\" {refspecs}", secrets);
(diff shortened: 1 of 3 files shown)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -283,14 +282,6 @@ private static void InitFetch(string path, string url, IEnumerable<string> commi
             // - git remote set url
             // - git fetch
             // - git checkout (if not a bar repo)
```
live1206: seems there is no `checkout` step after `fetch` ?

PR #5392, opened 2019-12-24
Fetch git dependencies with --depth 1

yufeih:
`--depth 1` was originally enabled in https://github.com/dotnet/docfx/pull/3668 to speed up dependency repo restore speed, but was removed in #4997 due to performance problem on the server.

This PR adds back `--depth 1` to improve local build speed, and add an option `gitShallowFetch` for server to opt out of the behavior.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5392)

### DIFF
--- a/src/docfx/restore/RestoreGit.cs
+++ b/src/docfx/restore/RestoreGit.cs
@@ -13,23 +13,13 @@ namespace Microsoft.Docs.Build
     internal static class RestoreGit
     {
         internal static IReadOnlyList<RestoreGitResult> Restore(
-            Config config,
-            string locale,
-            Repository repository,
-            DependencyLockProvider dependencyLockProvider)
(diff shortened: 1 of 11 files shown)


src/docfx/restore/RestoreGit.cs
```diff
@@ -38,101 +28,79 @@ internal static class RestoreGit
             return results.ToList();
         }
 
-        internal static IReadOnlyList<RestoreGitResult> RestoreGitRepo(
-            Config config,
-            string remote,
-            List<(string branch, RestoreGitFlags flags)> branches,
-            DependencyLockProvider dependencyLockProvider)
+        internal static RestoreGitResult RestoreGitRepo(
```
live1206: make the method public

PR #5393, opened 2019-12-24
Use updated_at to represent when a cache entry will expire

yufeih:
The original cache uses `expiry` to represent when a cache item expires, this is a time into the future. The problem with that approach is that changes to `expiration` config does not take effect immediately, the `expiry` is only updated slowly after they have actually expired calculated using the old `expiration` value.

This PR stores `updated_at` to the cache file. To ensure even distribution of cache items across the whole expiration period, we randomize the `milliseconds` component of `updated_at`, and use it as a deterministic random factor to check if the item has expired. `milliseconds` range from 0 to 999, creating 1000 bucket of ~10 minutes if the total expiration is a week. This precision satisfies the requirements of docs.

The change is backward compatible, all existing cache entries are treated as updated just now.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5393)

### DIFF
--- a/src/docfx/lib/cache/JsonDiskCache.cs
+++ b/src/docfx/lib/cache/JsonDiskCache.cs
@@ -143,7 +143,7 @@ namespace Microsoft.Docs.Build
         private bool HasExpired(TValue value)
         {
             var updatedAt = value.UpdatedAt ?? ((DateTime)(value.UpdatedAt = GetRandomUpdatedAt()));
-            var expiry = updatedAt.Millisecond / 1000.0 * _expirationInSeconds;
+            var expiry = (0.5 + (updatedAt.Millisecond / 1000.0)) * _expirationInSeconds;
 
             return updatedAt.AddSeconds(expiry) < DateTime.UtcNow;
         }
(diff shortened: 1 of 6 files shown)


src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -135,14 +135,17 @@ private Lazy<Task<TError>> UpdateDelegate(TKey key, Func<TKey, Task<(TError, IEn
...
-            return (value / 2) + (t_random.Value.NextDouble() * value / 2);
+            return DateTime.UtcNow.AddMilliseconds(1000.0 * t_random.Value.NextDouble());
         }
 
-        private static bool HasExpired(TValue value)
+        private bool HasExpired(TValue value)
         {
-            return value.Expiry != null && value.Expiry < DateTime.UtcNow;
+            var updatedAt = value.UpdatedAt ?? ((DateTime)(value.UpdatedAt = GetRandomUpdatedAt()));
+            var expiry = (0.5 + (updatedAt.Millisecond / 1000.0)) * _expirationInSeconds;
```
mingwli: should be 2000 ?

yufeih: I see, the original was 0.5~1, I thought it was 0.5~1.5

yufeih:
`GitHubAccessor` produces un-deterministic result in some rare cases, that caused the test diffs recently.

PR #5405, opened 2019-12-27
Simplify git restore with PackageResolver

yufeih:
Similar to `FileResolver`, `PackageResolver` is used to resolve package dependencies, this is currently a folder or git.

Removed existing features:

- Dependency lock
- bare repo
- git work tree

This enables the ability to restore packages on demand during build, for better local authoring experience.

https://dev.azure.com/ceapex/Engineering/_workitems/edit/121829/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5405)


docs/specs/git.yml
```diff
@@ -403,6 +403,7 @@ repos:
         a.md:
 outputs:
   .errors.log: |
+    ["error","committish-not-found","Can't find branch, tag, or commit 'live' for repo *"]
```
yufeih: This error was raised from `docfx restore`, previous implementation does not raise this error which should be an bug

live1206: Is this error duplicated with the one below?

yufeih: Nop, the error from `docfx build` contains file info

src/docfx/restore/PackageFetchOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    [Flags]
+    internal enum PackageFetchOptions
+    {
+        /// <summary>
+        /// Default fetch option.
+        /// </summary>
+        None = 0,
```
live1206: how about just renaming this to `DepthOne`, `FullHistory` and `NotThrow`?

src/docfx/restore/Restore.cs
```diff
@@ -81,5 +75,115 @@ private static async Task<bool> RestoreDocset(string docsetPath, string outputPa
...
+
+        private static void RestorePackages(string docsetPath, Config config, string locale, Repository repository)
+        {
+            using (var packageResolver = new PackageResolver(docsetPath, config))
+            {
+                ParallelUtility.ForEach(
+                    GetPackages(config, locale, repository).Distinct(),
+                    item => packageResolver.DownloadPackage(item.package, item.flags),
+                    Progress.Update,
+                    maxDegreeOfParallelism: 8);
```
live1206: should this be related to `Environment.ProcessorCount` instead of a fixed number like `ParallelUtility`?

yufeih: https://github.com/dotnet/docfx/pull/5009

PR #5413, opened 2019-12-30
Add external links to .links.json

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/152392

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5413)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -269,9 +269,9 @@ namespace Microsoft.Docs.Build
             return result.ToString();
         }
 
-        public static HtmlNode AddLinkType(this HtmlNode html, string locale)
+        public static HtmlNode AddLinkType(this HtmlNode html, string locale, Action<string> action = null)
         {
-            AddLinkType(html, "a", "href", locale);
+            AddLinkType(html, "a", "href", locale, action);
             AddLinkType(html, "img", "src", locale);
             return html;
         }
@@ -335,7 +335,7 @@ namespace Microsoft.Docs.Build
             return html;
         }
 
-        private static void AddLinkType(this HtmlNode html, string tag, string attribute, string locale)
+        private static void AddLinkType(this HtmlNode html, string tag, string attribute, string locale, Action<string> action = null)
         {
             foreach (var node in html.Descendants(tag))
             {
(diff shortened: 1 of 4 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -269,9 +269,9 @@ public static string CreateHtmlMetaTags(JObject metadata, ICollection<string> ht
             return result.ToString();
         }
 
-        public static HtmlNode AddLinkType(this HtmlNode html, string locale)
+        public static HtmlNode AddLinkType(this HtmlNode html, string locale, Action<string> addLinkAction = null)
         {
-            AddLinkType(html, "a", "href", locale);
+            AddLinkType(html, "a", "href", locale, addLinkAction);
```
OsmondJiang: will this cause duplicated links?

live1206: Yes, could do a fix for it, ignore `external` link-type during resolving link. But the inclusion of `external` links can not be fixed.

herohua:
would it make more sense to put the logic in `LoadMarkdown`?

yufeih:
See LinkExtension.cs

live1206:
> would it make more sense to put the logic in `LoadMarkdown`?

This is not only for `.md` files, consider a `JSON/YAML` file with a field which is `Markdown`.

live1206:
> This might not be the best place to process the link. Based on the test case, it seems markdown autolinks are not processed by LinkResolver, I feel like capturing autolinks and have it go though the same link resolve process is a better fix.

Consider a user written link(`<a href="link"></a>`) within HTML block, would it be parsed as autolinks during markup?

yufeih:
html links are not markdown autolinks, but html links are also parsed and resolved using the same link resolve process, so you can write **a href=a.md** and we'll resolve a.md to the correct url using link resolver.

These logics are all handled in LinkExtension.cs

yufeih:
https://github.com/dotnet/docfx/blob/v3/src/docfx/lib/markdown/LinkExtension.cs#L29

Autolinks are explicitly excluded, can try remove that check. I dont recall exactly why it is there in the first place, might be a perf optimization where we can skip external urls.

PR #5426, opened 2020-01-07
Handle AggregatedException

yufeih:
`Parallel.ForEach` generates `AggregatedException` that contains error for each item. Make `DocfxException.IsDocfxException` friendly to `AggregatedException`.

Removed file override, error source info are populated when the error was created.

https://dev.azure.com/ceapex/Engineering/_workitems/edit/160455/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5426)

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -406,7 +406,7 @@ repos:
 outputs:
   .errors.log: |
     ["error","committish-not-found","Can't find branch, tag, or commit 'live' for repo *"]
-    ["error","committish-not-found","Can't find branch, tag, or commit 'live' for repo *","a.md"]
+    ["error","committish-not-found","Can't find branch, tag, or commit 'live' for repo *"]
 ---
 # Use current repo's branch when edit branch isn't specified in contribution config,
 # and contribution config's repository equals to current repo's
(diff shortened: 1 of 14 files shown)


docs/specs/git.yml
```diff
@@ -406,7 +406,7 @@ repos:
 outputs:
   .errors.log: |
     ["error","committish-not-found","Can't find branch, tag, or commit 'live' for repo *"]
-    ["error","committish-not-found","Can't find branch, tag, or commit 'live' for repo *","a.md"]
+    ["error","committish-not-found","Can't find branch, tag, or commit 'live' for repo *"]
```
live1206: why file info is missing for build error?

src/docfx/lib/log/DocfxException.cs
```diff
@@ -22,20 +23,43 @@ public DocfxException(Error error, Exception innerException = null, ErrorLevel?
             OverwriteLevel = overwriteLevel;
         }
 
-        public static bool IsDocfxException(Exception ex, out DocfxException docfxException)
+        public static bool IsDocfxException(Exception ex, out IEnumerable<DocfxException> exceptions)
         {
-            while (ex != null)
+            List<DocfxException> result = null;
```
live1206: how about initializing with `new List<DocfxException>()` to avoid the null fallback afterwards?

docs/specs/git.yml
yufeih: It is a global error and was mistakenly reported as per file error due to we automatically append file info before

PR #5427, opened 2020-01-08
Upgrade Csharp version

live1206:
This PR mostly updates [the `using` declaration](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#using-declarations).

Since the declared using will be disposed in the end of the enclosing scope, if we want to dispose it earlier, the block scope is needed. 
For instance, https://github.com/dotnet/docfx/blob/9f7612933a83b079537c747d3801a3dd8b1af33e/test/docfx.Test/TestUtility.cs#L74
The archive creation should be ended before the file creation.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5427)

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>Latest</LangVersion>
+    <LangVersion>8.0</LangVersion>
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)stylecop.ruleset</CodeAnalysisRuleSet>
     <NoWarn>NU1701</NoWarn>
     <TreatWarningsAsErrors Condition="'$(Configuration)'=='Release'">true</TreatWarningsAsErrors>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>Latest</LangVersion>
     <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)stylecop.ruleset</CodeAnalysisRuleSet>
     <NoWarn>NU1701</NoWarn>
     <TreatWarningsAsErrors Condition="'$(Configuration)'=='Release'">true</TreatWarningsAsErrors>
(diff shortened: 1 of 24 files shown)


src/Directory.Build.props
```diff
@@ -1,7 +1,7 @@
 <Project>
 
   <PropertyGroup>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>Latest</LangVersion>
```
yufeih: Use explicit version in case when the next language upgrade came and we haven't upgrade dotnet to match that language version
```suggestion
    <LangVersion>8.0</LangVersion>
```

src/docfx/build/Build.cs
```diff
@@ -29,52 +29,47 @@ private static async Task<bool> BuildDocset(string docsetPath, string outputPath
...
-            {
-                var stopwatch = Stopwatch.StartNew();
-
-                try
-                {
-                    // load and trace entry repository
-                    var repository = Repository.Create(docsetPath);
-                    Telemetry.SetRepository(repository?.Remote, repository?.Branch);
-                    var locale = LocalizationUtility.GetLocale(repository, options);
+            using var errorLog = new ErrorLog(docsetPath, outputPath, () => config, options.Legacy);
```
yufeih: would be nice if you add a list of new language syntax changed in this PR in the description.

src/docfx/lib/path/PathUtility.cs
```diff
@@ -256,5 +256,10 @@ private static bool GetIsCaseSensitive()
                 return false;
             }
         }
+
+        private static string NewMethod(string pathWithUpperCase)
```
yufeih: revert?

src/Directory.Build.props
live1206: why not use the latest?

src/docfx/lib/path/PathUtility.cs
live1206: whoops

src/docfx/build/Build.cs
live1206: updated description

PR #5429, opened 2020-01-08
perf: Process redirections in parallel

yufeih:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/160455/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5429)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Docs.Build
 {
     internal static class Build
     {
-        public static async Task<int> Run(string workingDirectory, CommandLineOptions options)
+        public static int Run(string workingDirectory, CommandLineOptions options)
         {
             var docsets = ConfigLoader.FindDocsets(workingDirectory, options);
             if (docsets.Length == 0)
@@ -21,11 +21,18 @@ namespace Microsoft.Docs.Build
                 return 1;
             }
 
-            var result = await Task.WhenAll(docsets.Select(docset => BuildDocset(docset.docsetPath, docset.outputPath, options)));
-            return result.All(x => x) ? 0 : 1;
+            var hasError = false;
+            Parallel.ForEach(docsets, docset =>
+            {
+                if (BuildDocset(docset.docsetPath, docset.outputPath, options))
+                {
+                    hasError = true;
+                }
+            });
+            return hasError ? 1 : 0;
         }
 
-        private static async Task<bool> BuildDocset(string docsetPath, string outputPath, CommandLineOptions options)
+        private static bool BuildDocset(string docsetPath, string outputPath, CommandLineOptions options)
         {
             List<Error> errors;
             Config config = null;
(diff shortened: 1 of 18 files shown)


src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -167,26 +169,44 @@ private static IEnumerable<string> ProbeRedirectionFiles(string docsetPath)
...
+
+            var publishUrlMap = files
+                .GroupBy(file => file.siteUrl)
                 .ToDictionary(group => group.Key, group => group.ToList(), PathUtility.PathComparer);
 
-            foreach (var item in redirections.Where(item => item.RedirectDocumentId))
+            var renameHistory = new ConcurrentDictionary<FilePath, FilePath>();
+
+            // Convert the redirection target from redirect url to file path according to the version of redirect source
+            Parallel.ForEach(redirections, item =>
```
mingwli: >Parallel [](start = 12, length = 8)

How do we decide whether using `Parallel` or `ParallelUtility`?

yufeih: - The action can throw `DocfxException`
- This function should continue in case action throws `DocfxException`

mingwli: `GetFileLevelMonikers` is also invoked in this parallel scope, though logically it wouldn't throw. That's why I was confused.

src/docfx/build/Build.cs
```diff
@@ -55,20 +62,20 @@ private static async Task<bool> BuildDocset(string docsetPath, string outputPath
 
                 // run build based on docsets
                 outputPath ??= Path.Combine(docsetPath, docset.Config.Output.Path);
-                await Run(docset, fallbackDocset, options, errorLog, outputPath, input, repositoryProvider, localizationProvider);
+                Run(docset, fallbackDocset, options, errorLog, outputPath, input, repositoryProvider, localizationProvider).GetAwaiter().GetResult();
```
live1206: what do we gain by changing the method chain to synchronous instead of asynchronous all the way?

yufeih: `await Task.WhenAll` wasn't true parallel.

live1206: can we use `ParallelUtility.ForEach` to not use `.GetAwaiter().GetResult()`?

yufeih: split it to a separate PR.

test/docfx.Test/DocfxTest.cs
```diff
@@ -142,16 +142,16 @@ private async static Task RunCore(TestData test, string docsetPath, string outpu
...
             }
             else
             {
-                await RunBuild(docsetPath, outputPath, spec);
+                await RunBuild(docsetPath, outputPath, spec, spec.Locale);
             }
         }
 
-        private static async Task RunBuild(string docsetPath, string outputPath, DocfxTestSpec spec, bool dryRun = false)
+        private static async Task RunBuild(string docsetPath, string outputPath, DocfxTestSpec spec, string locale, bool dryRun = false)
```
live1206: merge from v3 branch? These have been removed

yufeih: done

PR #5433, opened 2020-01-09
feat: track telemetry on log

herohua:
corresponding change in docs pipeline: https://dev.azure.com/ceapex/Engineering/_git/Docs.Build/pullrequest/19674?_a=overview

https://review.docs.microsoft.com/en-us/engineering/projects/ops/build/telemetry/build-telemetry?branch=master#build-log-errorwarningsuggestion

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5433)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -17,26 +17,41 @@ internal static class Telemetry
...
 
+        public static void SetTelemetryConfig(TelemetryConfig telemetryConfig)
+        {
+            foreach (var dimension in telemetryConfig.Dimensions)
+            {
+                if (dimension.Key.Equals("CorrelationId", StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(dimension.Value))
+                {
+                    s_correlationId = dimension.Value;
+                }
+                s_dimensions[dimension.Key] = dimension.Value;
```
928PJY: >s_dimensions [](start = 16, length = 12)

never used?

src/docfx/lib/log/Telemetry.cs
```diff
@@ -17,26 +17,41 @@ internal static class Telemetry
...
+        private static readonly Metric s_buildItemCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Count", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"));
 
         private static readonly string s_version = typeof(Telemetry).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "<null>";
         private static readonly string s_os = RuntimeInformation.OSDescription ?? "<null>";
 
         private static string s_repo = "<null>";
         private static string s_branch = "<null>";
 
+        private static Dictionary<string, string> s_dimensions = new Dictionary<string, string>();
+        private static string s_correlationId = Guid.NewGuid().ToString("N");
```
yufeih: This leads me to think it is simpler to just pass `telemetryCorrelationId` because metric identifier is kind of fixed.

src/docfx/lib/log/Telemetry.cs
herohua: it will be used later for events

src/docfx/lib/log/Telemetry.cs
```diff
@@ -17,26 +17,42 @@ internal static class Telemetry
...
+        private static readonly Metric s_cacheCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Cache", "Name", "State", "OS", "Version", "Repo", "Branch", "CorrelationId"));
+        private static readonly Metric s_buildItemCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Count", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"));
 
         private static readonly string s_version = typeof(Telemetry).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "<null>";
         private static readonly string s_os = RuntimeInformation.OSDescription ?? "<null>";
 
         private static string s_repo = "<null>";
         private static string s_branch = "<null>";
 
+        private static Dictionary<string, string> s_eventDimensions = new Dictionary<string, string>();
```
yufeih: nit: can simply assign here to avoid the copying.
```suggestion
        private static readonly TelemetryConfig s_telemetryConfig;
```

yufeih: There was a potential bug here for multiple docsets, since telemetry config, repo, branch is tied to a docset in docfx, consider use `AsyncLocal<T>` for `s_repo`, `s_branch`, `s_telemetryConfig` as `AsyncLocal<T>` so each docset can have different values.

herohua: are we now building multiple docset in parallel?

928PJY: why is it tied to docset? should it be the same between docsets?

---
In reply to: [365082539](https://github.com/dotnet/docfx/pull/5433#discussion_r365082539) [](ancestors = 365082539)

herohua: I'd like to keep `eventDimensions` and `correlationId` separate because there will be a special handling on `correlationId` (if passed in value is null or empty, don't run the set). Meanwhile, I will reuse `TelemetryConfig.EventDimensions` to avoid the copy.

herohua: I see we are running docset build parallelly. Please ignore my last question.

yufeih: oh, that you can initialize it directly in `TelemetryConfig` like `public readonly string CorrelationId = Guid.NewGuid()`

src/docfx/lib/log/Telemetry.cs
```diff
@@ -17,26 +18,35 @@ internal static class Telemetry
...
         private static readonly string s_version = typeof(Telemetry).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "<null>";
         private static readonly string s_os = RuntimeInformation.OSDescription ?? "<null>";
 
-        private static string s_repo = "<null>";
-        private static string s_branch = "<null>";
+        private static AsyncLocal<string> s_repo = new AsyncLocal<string>();
+        private static AsyncLocal<string> s_branch = new AsyncLocal<string>();
+
+        private static AsyncLocal<Dictionary<string, string>> s_eventDimensions = new AsyncLocal<Dictionary<string, string>>();
+        private static AsyncLocal<string> s_correlationId = new AsyncLocal<string>();
```
fenxu: >s_correlationId [](start = 42, length = 15)

This one is shared I think, like os and version info.

PR #5434, opened 2020-01-09
fix xunit asserts

SimonCropp:
also reduces the noise of build warnings

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5434)


yufeih:
I'll fix the UT failure, it isn't related to your change.

SimonCropp:
can we get a merge here?

PR #5436, opened 2020-01-09
remove redundant net462 checks

SimonCropp:
seems no projects target net462, so these checks should be redundant?

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5436)


superyyrrzz:
You are right that they are outdated. Thank you for the contribution!

PR #5437, opened 2020-01-10
Remove build from en-us for LOC

live1206:
Build from en-us for localization repos are not in use, remove it to simplify the code.

While build from en-us, the local repo is en-us and the loc repo needs to be restored(including all branches). For the case that `live-sxs` branch needs to [get contributor information from `live` branch](https://github.com/dotnet/docfx/blob/aa164bcc41046447a47582f4dfde03e6062f3ba8/src/docfx/lib/git/FileCommitProvider.cs#L201), we would need to get data from `live` based on the restored `live-sxs` branch. 

But these two are restored as totally separated folders, the current implementation of restoring Git repo is to fetch a specific [Commitish(would be a branch in this case)](https://github.com/dotnet/docfx/blob/aa164bcc41046447a47582f4dfde03e6062f3ba8/src/docfx/restore/PackageResolver.cs#L128), not all the branches for better performance.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5437)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Docs.Build
                     return false;
 
                 using var packageResolver = new PackageResolver(docsetPath, config, noFetch: true);
-                var localizationProvider = new LocalizationProvider(packageResolver, options, config, locale, docsetPath, repository);
+                var localizationProvider = new LocalizationProvider(packageResolver, config, locale, docsetPath, repository);
                 var repositoryProvider = new RepositoryProvider(docsetPath, repository, options, config, packageResolver, localizationProvider);
                 var input = new Input(docsetPath, repositoryProvider, localizationProvider);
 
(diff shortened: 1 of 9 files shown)


src/docfx/build/Build.cs
```diff
@@ -46,7 +46,7 @@ private static async Task<bool> BuildDocset(string docsetPath, string outputPath
                     return false;
```
yufeih: can also remove `CommandLineOptions.Locale`

live1206: right

yufeih:
can also remove `LocalizationMapping` I guess because `LocalizationMapping.Folder` only works for _build from english_ and `LocalizationMapping.Branch` isn't a done/live feature.

live1206:
> can also remove `LocalizationMapping` I guess because `LocalizationMapping.Folder` only works for _build from english_ and `LocalizationMapping.Branch` isn't a done/live feature.

Let's refine this later with documentation update

PR #5441, opened 2020-01-10
feat: track docfx config in telemetry

herohua:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5441)

### DIFF
--- a/src/docfx/config/ConfigLoader.cs
+++ b/src/docfx/config/ConfigLoader.cs
@@ -89,13 +89,7 @@ namespace Microsoft.Docs.Build
             errors.AddRange(configErrors);
 
             // Track docfx config telemetry
-            Telemetry.TrackEvent(
-                docfxConfigFileName,
-                new Dictionary<string, string>
-                {
-                    ["Config"] = docfxConfigFileContent,
-                    ["DocsetName"] = docsetName,
-                });
+            Telemetry.TrackDocfxConfig(docfxConfigFileName, docfxConfigFileContent, docsetName);
 
             return (errors, config);
         }
--- a/src/docfx/config/ConfigLoader.cs
+++ b/src/docfx/config/ConfigLoader.cs
@@ -65,7 +65,7 @@ namespace Microsoft.Docs.Build
             var docfxConfigFileName = Path.GetFileName(configPath);
             var docfxConfigFileContent = File.ReadAllText(configPath);
             var docfxConfig = LoadConfig(errors, docfxConfigFileName, docfxConfigFileContent);
-            var (docsetName, xrefEndpoint, xrefQueryTags, opsConfig) = OpsConfigLoader.LoadDocfxConfig(docsetPath, _repository?.Branch ?? "master");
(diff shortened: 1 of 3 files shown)


src/docfx/config/ConfigLoader.cs
```diff
@@ -62,8 +62,10 @@ where glob(configPath)
             var envConfig = LoadEnvironmentVariables();
             var cliConfig = new JObject();
             JsonUtility.Merge(unionProperties, cliConfig, options?.StdinConfig, options?.ToJObject());
-            var docfxConfig = LoadConfig(errors, Path.GetFileName(configPath), File.ReadAllText(configPath));
-            var (xrefEndpoint, xrefQueryTags, opsConfig) = OpsConfigLoader.LoadDocfxConfig(docsetPath, _repository?.Branch ?? "master");
+            var docfxConfigFileName = Path.GetFileName(configPath);
+            var docfxConfigFileContent = File.ReadAllText(configPath);
+            var docfxConfig = LoadConfig(errors, docfxConfigFileName, docfxConfigFileContent);
+            var (docsetName, xrefEndpoint, xrefQueryTags, opsConfig) = OpsConfigLoader.LoadDocfxConfig(docsetPath, _repository?.Branch ?? "master");
```
yufeih: `docsetName` is the same as `preloadConfig.Name`

src/docfx/config/ConfigLoader.cs
```diff
@@ -86,6 +88,15 @@ where glob(configPath)
             var (configErrors, config) = JsonUtility.ToObject<Config>(configObject);
             errors.AddRange(configErrors);
 
+            // Track docfx config telemetry
+            Telemetry.TrackEvent(
+                docfxConfigFileName,
+                new Dictionary<string, string>
+                {
+                    ["Config"] = docfxConfigFileContent,
```
yufeih: `docfxConfigFileContent` is YAML for `docfx.yml`, do you want to always produce json?
```suggestion
                    ["Config"] = docfxConfig.ToString(),
```

src/docfx/lib/log/Telemetry.cs
```diff
@@ -69,6 +69,11 @@ public static void TrackBuildCommitCount(int count)
             s_buildItemCountMetric.TrackValue(count, TelemetryName.BuildCommits.ToString(), s_os, s_version, s_repo, s_branch);
         }
 
+        public static void TrackEvent(string name, Dictionary<string, string> properties, Dictionary<string, double> metrics = null)
```
yufeih: nit: consider a more descriptive method like `TrackConfig(config)`

herohua: so we prefer to hide telemetry specific logic in `Telemetry`? I think that's a good idea.

src/docfx/config/ConfigLoader.cs
herohua: I think it is okay to track different format for different input source - plus the event name is different if the config file is actually `docfx.yml`

PR #5440, opened 2020-01-10
Build implicitly restore dependencies if not cached.

yufeih:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5440)

PR #5456, opened 2020-01-13
Reduce tests execution time

live1206:
Based on #5440 
- Turn `noFetch` flag into enum `FetchOption`:
  - None(default: always fetch latest)
  - NoFetch
  - UseCache(only fetch latest if cache does not exist)
- Add `--use-cache` to specify `FetchOption.UseCahce` for `restore`
- This reduces tests execution time with cache locally from 1:30 to 00:30

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5456)

### DIFF
--- a/src/docfx/cli/Docfx.cs
+++ b/src/docfx/cli/Docfx.cs
@@ -110,6 +110,7 @@ namespace Microsoft.Docs.Build
                     // Restore command
                     syntax.DefineCommand("restore", ref command, "Restores dependencies before build.");
                     syntax.DefineOption("o|output", ref options.Output, "Output directory in which to place restore log.");
+                    syntax.DefineOption("use-cache", ref options.UseCache, "ONly fetch latest content if it does not exist or read from disk cache.");
                     DefineCommonOptions(syntax, ref workingDirectory, options);
 
                     // Build command
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -29,6 +29,7 @@ namespace Microsoft.Docs.Build
         {
             List<Error> errors;
             Config config = null;
+            options.NoFetch = true;
 
             using var errorLog = new ErrorLog(docsetPath, outputPath, () => config, options.Legacy);
             var stopwatch = Stopwatch.StartNew();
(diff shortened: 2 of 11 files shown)


src/docfx/cli/Docfx.cs
```diff
@@ -110,6 +110,7 @@ private static (string command, string workingDirectory, CommandLineOptions opti
                     // Restore command
                     syntax.DefineCommand("restore", ref command, "Restores dependencies before build.");
                     syntax.DefineOption("o|output", ref options.Output, "Output directory in which to place restore log.");
+                    syntax.DefineOption("use-cache", ref options.UseCache, "ONly fetch latest content if it does not exist or read from disk cache.");
```
herohua: `Only`

yufeih: This IMO could be the default behavior when executing `docfx build`. Rebuilding contents fast from the command line is the most frequent ask.

The same pattern could be found for `dotnet run`, `npm run`, etc. These commands don't automatically fetch latest dependencies each time, if you absolutely want the latest dependency, explicitly execute `dotnet restore` and `npm restore`. For us it is the same story. `docfx build` alone ensures the docset is build and build runs fast. To force latest dependency, execute `docfx restore` explicitly.

See `dotnet run --no-cache --no-dependencies` as example patterns. The reason I didn't use `--no-cache` term for docfx is that we have other cache files like github user cache, it could be called `--no-restore-cache`

live1206: Basically I did not change the behavior of `docfx build` in this PR, do not fetch anything during `build`. `Fetch` only happens during `docfx restore`, for now seems like the default behavior of `docfx restore` is to always fetch the latest. That is why I added `--uses-cache` for `restore` here

yufeih: I see, the --no-cache flag looks more like --force flag described here:https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore?tabs=netcore2x

Also this need to update docs-pipeline as this is a behavior difference.

live1206: Right, I will open another breaking PR paired with docs-pipeline change

PR #5468, opened 2020-01-14
use pattern matching

SimonCropp:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5468)

### DIFF
--- a/src/Microsoft.DocAsCode.Build.Common/ModelAttributeHandlers/Handlers/UniqueIdentityHandler.cs
+++ b/src/Microsoft.DocAsCode.Build.Common/ModelAttributeHandlers/Handlers/UniqueIdentityHandler.cs
@@ -44,22 +44,19 @@ namespace Microsoft.DocAsCode.Build.Common
                     return null;
                 }
 
-                var val = currentObj as string;
-                if (val != null)
+                if (currentObj is string val)
                 {
                     context.LinkToUids.Add(val);
                 }
                 else
                 {
-                    var list = currentObj as IEnumerable;
-                    if (list != null)
+                    if (currentObj is IEnumerable list)
                     {
                         foreach (var i in list)
                         {
                             if (i != null)
                             {
-                                var item = i as string;
-                                if (item != null)
+                                if (i is string item)
                                 {
                                     context.LinkToUids.Add(item);
                                 }
--- a/src/Microsoft.DocAsCode.Build.Engine/HostService.cs
+++ b/src/Microsoft.DocAsCode.Build.Engine/HostService.cs
@@ -553,8 +553,7 @@ namespace Microsoft.DocAsCode.Build.Engine
 
         private void HandleUidsChanged(object sender, PropertyChangedEventArgs<ImmutableArray<UidDefinition>> e)
         {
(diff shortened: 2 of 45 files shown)


superyyrrzz:
@SimonCropp Out of curiosity, do you do these manually or with tools?

SimonCropp:
@superyyrrzz I use a combo of things to cleanup/refactor large codebases

 * compiler warnings
 * VS code hints
 * Some VS addins
 * R#
 * some r# addins
 * mono cecil to look for probs in the resulting IL.
 * ad-hoc unit tests to looks for probs using reflection
 * simple regex text search/replaces my preferred tool is sublime text

SimonCropp:
@superyyrrzz shall i keep on going?

superyyrrzz:
@SimonCropp Absolute welcome! It helps.

As we are currently putting more engineering resources on vNext (see v3 branch), you can also try improving it if you are interested. v3's code is more modern that is based on .NET Core 3.1 so that it could have some different potential improvements. /cc @herohua @yufeih 

Thank you for keep improving DocFX!

SimonCropp:
@superyyrrzz @herohua @yufeih ok will have a look at the v3 branch. on thing that could be problematic is keeping the v3 brach in sync with def. what is your approach for handling that? my preference would be for at least a daily "merge dev into v3". I can help with this, but it would be more complicated through the external contributor restrictions of GH, so perhaps nomineere someone on the team?

superyyrrzz:
@SimonCropp No need to sync as it is a total rewrite.

SimonCropp:
@superyyrrzz so does that mean dev will be abandoned? is do how are you reconciling fixes and improvements in dev?

superyyrrzz:
v3 still has works to be done before releasing a usable version for the community. Even released, it is not likely to abandon v2 in a short time, as v3 does not support all the features in v2. It would be a long term decision whether or when to abandon v2, so no worry for now 😄

SimonCropp:
but that still seems that efforts on code cleanup an perf should be focused on v3. and dev should be marked as "only critical bug fix PRs". given that I would suggest changing the default branch to v3.

SimonCropp:
As an aside, while i am happy to help out, i am now  pretty disenfranchised, since it seems none of my recent work will benefit the codebase long term. I wonder how many other people (who have contributed improvements) would feel the same way.

superyyrrzz:
[The roadmap](https://github.com/dotnet/docfx/blob/dev/Roadmap.md#v2-strategy) mentioned it, but I think this document is not easy to find. Sorry for the confusing...

> but that still seems that efforts on code cleanup an perf should be focused on v3. and dev should be marked as "only critical bug fix PRs".

I agree that if you want to contribute some refactor or code clean up, I would recommend to contribute to v3 with higher priority. However, your effort deserves it, which can help other bug fix and features in dev branch. Dev branch is still active for new feature and bug fix.

> given that I would suggest changing the default branch to v3.

As dev branch still reflect the code that all the community users are using now, we are still using it as default.

PR #5478, opened 2020-01-15
fix for invalid ending json

live1206:
- workaround for invalid ending JSON de-serialization
- space should be valid ending

The root cause is from `JSON.NET`, filed an [issue](https://github.com/JamesNK/Newtonsoft.Json/issues/2256) for it

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5478)

### DIFF
--- a/test/docfx.Test/lib/JsonUtilityTest.cs
+++ b/test/docfx.Test/lib/JsonUtilityTest.cs
@@ -445,6 +445,14 @@ namespace Microsoft.Docs.Build
             Assert.Equal("{\"b\":1,\"d\":false}", result);
         }
 
+        [Theory]
+        [InlineData("{}---")]
+        [InlineData("{}{}")]
+        public void TestInvalidJson(string json)
+        {
+            Assert.Throws<DocfxException>(() => JsonUtility.Deserialize<ClassWithSourceInfo>(json, new FilePath("path")));
+        }
+
         [Fact]
         public void TestDeserializeWithSourceInfo()
         {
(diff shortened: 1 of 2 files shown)


test/docfx.Test/lib/JsonUtilityTest.cs
```diff
@@ -459,7 +467,7 @@ public void TestDeserializeWithSourceInfo()
         [Fact]
         public void TestExtensionDataWithSourceInfo()
         {
-            var (_, result) = DeserializeWithValidation<ClassWithExtensionData>("{\"a\": 1}");
+            var (_, result) = DeserializeWithValidation<ClassWithExtensionData>("{\"a\": 1} ");
```
herohua: how about add a specific test for valid JSON

live1206: There are many tests with valid JSon within this test file

yufeih:
There is a [CheckAdditionalContent](https://github.com/JamesNK/Newtonsoft.Json/blob/c5f210347055c43970199d4611993200d00977fd/Src/Newtonsoft.Json/JsonConvert.cs#L832) property you could use for this.

fenxu:
> 
> 
> There is a [CheckAdditionalContent](https://github.com/JamesNK/Newtonsoft.Json/blob/c5f210347055c43970199d4611993200d00977fd/Src/Newtonsoft.Json/JsonConvert.cs#L832) property you could use for this.

By default that's true. I think it already checked...

yufeih:
And I think it make sense to _NOT_ check additional content for `Stream` input by default, because performing such check leads to over-read.

We currently reads the first command line input as JSON config when `--stdio` is specified, a more user friendly version is to read the first complete JSON as config in case of `--stdio`. _NOT_ checking additional content for std input stream enables this behavior.

live1206:
> And I think it make sense to _NOT_ check additional content for `Stream` input by default, because performing such check leads to over-read.
> 
> We currently reads the first line of standard input as JSON config when `--stdio` is specified, a more user friendly version is to read the first complete JSON as config in case of `--stdio`, this allows you to input multi line JSON with intention. _NOT_ checking additional content for std input stream enables this behavior.

Makes sense, thanks for the explanation! 👍👍

PR #5482, opened 2020-01-16
Remove shared *.schema.json in spec tests

v-pegao:
Clean-up usage of TestData.schema.json in spec tests in order that each test is more self-contained and more readable.

[Find related workitem here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/115502/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5482)


docs/specs/escape.yml
```diff
@@ -135,6 +134,34 @@ inputs:
     Link to @a&b?displayProperty=description
     Link to @a&c
     Link to <xref:a%2A>
+  _themes/ContentTemplate/schemas/TestData.schema.json: |
+    {
+      "type": "object",
```
yufeih: remove unnecessary properties in this schema json to make the test case short and more readable. E.g., this case only needs:

```
{
    "properties": {
         "uid": { "contentType": "uid" }
    },
    "xrefProperties": ["name", "fullName", "description"]
}
```

v-pegao: seems not working, i'm checking the diff between expected and actual

v-pegao: @yufeih Updated, could you please help review? Let me know if I should do this for all `TestData.schema.json`

yufeih: Yep, do this for all other cases

v-pegao: Done. Could you please help rewiew?

docs/specs/xref.yml
```diff
@@ -338,6 +350,20 @@ repos:
...
+          {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+              "uid": { "contentType": "uid" },
+              "xref": { "contentType": "Xref" },
+              "name": { "type": "string" },
+              "fullName": { "type": "string" },
+              "description": { "type": "string" },
+              "$schema": { "type": "string" }
```
yufeih:
```suggestion
              ~~~"$schema": { "type": "string" }~~~
```

v-pegao: This change will cause test failure, why we need this change?

yufeih: Oops, I meant removing these unnecessary lines like ~"$schema": { "type": "string" }~ and ~"type": "object",~ and ~"additionalProperties": false,~

v-pegao: Updated. Sorry for missing this one.

PR #5486, opened 2020-01-16
fix: remove TrackValue value per dimension limit

herohua:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5486)

### DIFF
--- a/src/docfx/lib/log/Telemetry.cs
+++ b/src/docfx/lib/log/Telemetry.cs
@@ -17,10 +17,14 @@ namespace Microsoft.Docs.Build
     {
         private static readonly TelemetryClient s_telemetryClient = new TelemetryClient(TelemetryConfiguration.CreateDefault());
 
-        private static readonly Metric s_operationTimeMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Time", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"));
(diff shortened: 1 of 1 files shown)


yufeih:
Changing the limit to `int.MaxValue` may have a negative impact on memory usage and performance, because to house the different values, it needs to [buffer](https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/BASE/src/Microsoft.ApplicationInsights/Metrics/Implementation/ConcurrentDatastructures/MultidimensionalCube2.cs#L321) these distinct values in memory.

What metric exceed the 100 limit? Most of the track call uses `TrackValue(1)` so they have only one distinct value.

herohua:
> Changing the limit to `int.MaxValue` may have a negative impact on memory usage and performance, because to house the different values, it needs to [buffer](https://github.com/microsoft/ApplicationInsights-dotnet/blob/develop/BASE/src/Microsoft.ApplicationInsights/Metrics/Implementation/ConcurrentDatastructures/MultidimensionalCube2.cs#L321) these distinct values in memory.
> 
> What metric exceed the 100 limit? Most of the track call uses `TrackValue(1)` so they have only one distinct value.

It is the values on dimensions that would hit the limit, particular the `TrackValue` on `BuildLog`. With current setting, if the number of log code exceeds 100, further log code won't be captured by telemetry.

PR #5487, opened 2020-01-16
Add CheckAdditionalContent option for de-serialization JSON Stream

live1206:
A proper fix for #5478 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5487)

### DIFF
--- a/src/docfx/lib/json/JsonUtility.cs
+++ b/src/docfx/lib/json/JsonUtility.cs
@@ -115,17 +115,17 @@ namespace Microsoft.Docs.Build
         /// De-serialize a data string, which is not user input, to an object
         /// schema validation errors will be ignored, syntax errors and type mismatching will be thrown
         /// </summary>
-        public static T Deserialize<T>(string json, FilePath file, bool checkAddionalContent = true) where T : class, new()
+        public static T Deserialize<T>(string json, FilePath file) where T : class, new()
         {
             using var reader = new StringReader(json);
-            return Deserialize<T>(reader, file, checkAddionalContent);
+            return Deserialize<T>(reader, file, true);
         }
 
         /// <summary>
         /// De-serialize a data string, which is not user input, to an object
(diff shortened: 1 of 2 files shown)


src/docfx/lib/json/JsonUtility.cs
```diff
@@ -105,17 +115,17 @@ public static string Serialize(object graph, bool indent = false)
         /// De-serialize a data string, which is not user input, to an object
         /// schema validation errors will be ignored, syntax errors and type mismatching will be thrown
         /// </summary>
-        public static T Deserialize<T>(string json, FilePath file) where T : class, new()
+        public static T Deserialize<T>(string json, FilePath file, bool checkAddionalContent = true) where T : class, new()
         {
             using var reader = new StringReader(json);
-            return Deserialize<T>(reader, file);
+            return Deserialize<T>(reader, file, checkAddionalContent);
```
yufeih: `string` does not need this switch, this can always be `true`
```suggestion
            return Deserialize<T>(reader, file, true);
```

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -105,17 +115,17 @@ public static string Serialize(object graph, bool indent = false)
...
         }
 
         /// <summary>
         /// De-serialize a data string, which is not user input, to an object
         /// schema validation errors will be ignored, syntax errors and type mismatching will be thrown
         /// </summary>
-        public static T Deserialize<T>(TextReader json, FilePath file) where T : class, new()
+        public static T Deserialize<T>(TextReader json, FilePath file, bool checkAddionalContent = true) where T : class, new()
```
yufeih: typo
```suggestion
        public static T Deserialize<T>(TextReader json, FilePath file, bool checkAdditionalContent = true) where T : class, new()
```

PR #5491, opened 2020-01-17
Improve inheritdoc

### LINKED ISSUE #1306: Support for cref in inheritdoc tags
Following up on #247.

In regards to [this comment]( https://github.com/dotnet/docfx/commit/a2ba6d52432731a26af11c4f532c011fb7a78fe3#diff-a8bc073a917585e3db210c4032ae314cR253), I would like to refer to my comment in the related [roslyn issue](https://github.com/dotnet/roslyn/issues/67#issuecomment-206877691) again.

### LINKED ISSUE #1516: Inheritdoc doesn't copy from generic interfaces
### Title
For simple cases, <inheritdoc /> comments get copied just fine from an interface a class implements.
But if the interface is a generic type, e.g.

    public interface IInterface<TSomething>

it does not work.

### Functional impact
The result is, the documentation summary does not get copied from the interface to the actual implementation

### Minimal repro steps
* create an interface which has a generic type, 
* implement it in a class, 
* document a method of the interface
* add <inheritdoc /> to the implementation

### Expected result
Summary should be copied over to the implementation

### Actual result
Nothing shows up in the generated doc for the implemented class.

### Further technical details
After digging into the code, the issue is that when you iterate the "Implements" collection, the IDs used  are generated slightly differently than the ones compared against (when trying to find the interface's documentation data).

`VisitorHelper.GetCommentId(symbol)` generates IDs like **ClassName´1**
but 
for the implementation lookup, `SpecIdHelper.GetSpecId(symbol, typeGenericParameters, methodGenericParameters);` is used, which generates IDs which look like **ClassName{{T}}**

That's why `context.Members.TryGetValue` [right here](https://github.com/dotnet/docfx/blob/master/src/Microsoft.DocAsCode.Metadata.ManagedReference/Resolvers/CopyInheritedDocumentation.cs#L109) never returns anything.

Hope that makes sense

amirebrahimi:
Adds cref attribute support and fixes copying from templated sources

Resolves #1306
Fixes #1516

Updated existing tests and added an additional test

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5491)


superyyrrzz:
@amirebrahimi  Thank you for contribution!

PR #5493, opened 2020-01-19
fix: handle circular redirection

928PJY:
Fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/164345

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5493)

### DIFF
--- a/docs/specs/redirection.yml
+++ b/docs/specs/redirection.yml
@@ -478,3 +478,34 @@ inputs:
 outputs:
   folder/index.json: |
     { "document_id": "c663555f-8096-38df-325b-181ddc79afd8" }
+---
+# Handle circular redirection with document id
+inputs:
+  docfx.yml:
+  .openpublishing.redirection.json: |
+    {
+      "redirections": [
+        {
+          "source_path": "a.md",
+          "redirect_url": "/a",
+          "redirect_document_id": true
+        },
+        {
+          "source_path": "b.md",
+          "redirect_url": "/b",
+          "redirect_document_id": true
+        }
+      ]
+    }
+  a.md:
+outputs:
+  .publish.json: |
+    {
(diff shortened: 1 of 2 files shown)


docs/specs/redirection.yml
```diff
@@ -478,3 +478,34 @@ inputs:
...
+  .publish.json: |
+    {
+      "files": [
+          { "url": "/a", "redirect_url": "/a", "source_path": "a.md" },
+          { "url": "/b", "redirect_url": "/b", "source_path": "b.md" },
+          { "url": "/a", "source_path": "a.md" }
+      ]
+    }
+  .errors.log: |
+    ["error","publish-url-conflict","Two or more files publish to the same url '/a': 'a.md [redirection]', 'a.md'"]
```
herohua: >publish-url-conflict [](start = 14, length = 20)

maybe later we can refine on this error message

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -45,8 +45,14 @@ public string GetRedirectUrl(FilePath file)
 
         public FilePath GetOriginalFile(FilePath file)
         {
+            var redirectionChain = new HashSet<FilePath>();
             while (_renameHistory.TryGetValue(file, out var renamedFrom))
             {
+                if (redirectionChain.Contains(file))
```
yufeih: `HashSet.Add` method returns `false` if an item already exists in the collection, so can simplify to:
```suggestion
                if (!redirectionChain.Add(file))
                {
                    return file;
                }
```

PR #5496, opened 2020-01-20
fix: path normalize does not combine slash

yufeih:
The current path normalize algorithm merges consecutive slashes into one: `///` --> `/`, this is technically incorrect:
- It breaks url after normalization: `https://a.com` --> `https:/a.com`
- It turns UNC path into absolute path: `//a/b` --> `/a/b`

This PR fix this by not using `string.Split`.

Related test failure #5440 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5496)


live1206:
-      "asset_id": "admin/index",
+      "asset_id": "admin//index",

The asset_id has been changed, which is unexpected

src/docfx/lib/path/PathUtility.cs
```diff
@@ -225,6 +223,28 @@ public static string UrlToShortName(string url)
...
         }
 
+        private static bool IsNormalized(string path)
+        {
+            for (var i = 0; i < path.Length; i++)
+            {
+                var ch = path[i];
+                if (ch == '\\')
+                {
+                    return false;
```
live1206: possible to return the last visited index as the start of the normalization part?

yufeih:
Close this as docs relying on this behavior, and windows seems to allow multiple slashes to accessing the same file.

PR #5498, opened 2020-01-20
Enable more nullable checks

yufeih:
Upgrade stylecop because the old version keeps complaining about spacing problems with `string[]?`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5498)


yufeih:
#5515

PR #5510, opened 2020-02-07
171009 code extention bug

adunndevster:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/171009

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5510)


superyyrrzz:
Thank you @adunndevster for the fix! The change looks good to me. I am curios that what the error message would be like after this change? I just want to make sure that the user can get enough information to locate the error.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonParser.cs
```diff
@@ -40,7 +40,7 @@ public override BlockState TryOpen(BlockProcessor processor)
             var extensionName = "triple-colon";
             Action<string> logError = (string message) => _context.LogError(
                 $"invalid-{extensionName}",
-                $"Invalid {extensionName} on line {processor.LineIndex}. \"{slice.Text}\" is invalid. {message}",
+                $"Invalid {extensionName} on line {processor.LineIndex}. {message}",
```
yufeih: There is a dedicated line number column to show where the error is on build report, and we'll respect that line when showing the warning in vscode. Suggest to remove the line info from error message.

yufeih:
> Thank you @adunndevster for the fix! The change looks good to me. I am curios that what the error message would be like after this change? I just want to make sure that the user can get enough information to locate the error. It would be better to have a test covering it.

@superyyrrzz , all current markdig extension test does not cover error messages.

PR #5519, opened 2020-02-12
Lowercase locale

yufeih:
When repo name is {xyz}.ja-JP, the computed locale is currently `ja-JP`, leading us to clone `template.ja-JP` repo. Since we treat urls as case sensitive, this will not leverage the existing `template.ja-jp` repo cache.

This PR normalizes all locales to lowercase, so the computed locale becomes `ja-jp`.

Since `bilingual` is removed from config, the only localization related config is `defaultLocale`, I've moved this field to the root config node.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5519)


src/docfx/config/Config.cs
```diff
@@ -19,6 +19,11 @@ internal class Config : PreloadConfig
             "_themes/**",           // Default template location
         };
 
+        /// <summary>
+        /// Gets the default locale of this docset.
+        /// </summary>
+        public string DefaultLocale { get; private set; } = "en-us";
```
live1206: Is any repository using this config?

yufeih: nop, this config is not used by docs right now.

PR #5521, opened 2020-02-12
fix github auth

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5521)


yufeih:
The root cause is git matches path on slash boundary, but we use `StartsWith`.

Issue #4569, opened 2019-05-16
Deterministic export


tommikiviniemi-srs:
Generating the same documentation twice will contain a delta as the export is not deterministic, as the generated documentation's manifest.json's content order seems to be random.

It'd be great if generation was deterministic as it'd play far nicer with automated build systems where the artefacts can just be committed back to the source control.

Thanks.

superyyrrzz:
It is ordered by `source_relative_path` now. You can look at this sample: https://github.com/dotnet/docfx/blob/gh-pages/manifest.json

tommikiviniemi-srs:
Hi.

If I build the same documentation twice the entries in the generated .json file are jumping around (didn't check if the hashes etc. also change).

It'd be really nice if running the docfx html generation command twice in a row yielded the same results. Considering that it's not deterministic a commit is generated every time I run that command, even if the documentation didn't change.

For what it's worth I clear the generated documentation between those two runs (because docfx doesn't clean up files that are no longer used).

In other words:
1) Run docfx generate command
2) Save the generated directory
3) rm the generated directory
4) Run docfx generate command again
5) Diff the saved directory with the newly generated directory - they should be identical, but entries in the manifest json are jumping around.

As mentioned, the reason I remove the generated directory is that docfx doesn't remove files that may have been removed from the document sources. I didn't check whether not cleaning up the directory between two runs keeps manifest.json identical.

Thanks!

tommikiviniemi-srs:
To further illustrate the use chase:
If the export was deterministic, then I could configure our CI system to always generate the documentation and commit if there's changes. Right now I have to have a manual trigger for it that I run after editing documentation, because it will always generate a commit because of the delta in the manifest file.

Thanks.

superyyrrzz:
Thank you for your info. With more tests, I find `incremental_info/processors` is unstable in the manifest. Please correct me if that is not your case. It needs some code change to stabilize it as docfx do with `files` in the manifest.

tommikiviniemi-srs:
Hi superyyrrzz.

I double checked what exactly is jumping around, and it's stuff under incremental_info.processors in manifest.json, and the entries under index.json. I am attaching screenshots of the two files that are not stable.

<img width="2555" alt="Screenshot 2019-05-20 at 15 57 10" src="https://user-images.githubusercontent.com/7289960/58023342-3f1ff000-7b18-11e9-8846-954a1e1b6c18.png">
[...]

PR #5524, opened 2020-02-12
Avoid non-deterministic output

fibann:
index.json and incremental_info.processors in manifest.json are now generated
from sorted dictionaries and are now the same on every run.

#4569

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5524)


fibann:
I found the lines relevant to the bug easy enough, but I'll need some guidance to write a test if one is needed.

superyyrrzz:
Thank you @fibann for the contribution! It is fine to me for this code change as long as it does not break any existing tests.

I will merge this PR in next Monday to align with our release model.

PR #5529, opened 2020-02-14
Load OPS redirection config from repository root folder

yufeih:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/164344/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5529)


live1206:
Add a test with multiple redirection configs to demonstrate that the case would not cause issue in https://dev.azure.com/ceapex/Engineering/_workitems/edit/164344/

PR #5531, opened 2020-02-17
Handle the case that relative path outside docset folder

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/150423

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5531)

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -653,37 +653,29 @@ outputs:
     ["warning","file-not-found","Invalid file link: 'b.md/'.","docs/a.md",1,1]
 ---
 # resource inclusion with relative path outside docset
-repos:
-  https://github.com/ops1/resource#live:
-  - files:
-      .openpublishing.publish.config.json: |
-        {
-          "docsets_to_publish": [{ "build_source_folder": "new-hope/new-hope" }]
-        }
-      new-hope/new-hope/docfx.yml: |
-        output:
-          copyResources: false
-      new-hope/new-hope/specs/unify-version/unify-version.md: |
-        ![ready-for-review](../../../new-hope/specs/gauntlet/media/ready-for-review.png)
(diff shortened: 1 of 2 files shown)


src/docfx/build/dependency/LinkResolver.cs
```diff
@@ -173,7 +173,7 @@ internal class LinkResolver
 
                     // resolve file
                     var lookupFallbackCommits = inclusion || _documentProvider.GetContentType(path) == ContentType.Resource;
-                    var file = TryResolveRelativePath(referencingFile.FilePath, path, lookupFallbackCommits);
+                    var file = TryResolveRelativePath(referencingFile.Docset.DocsetPath, referencingFile.FilePath, path, lookupFallbackCommits);
```
yufeih: Add a `string _docsetPath` member field and initialize it from constructor?

docs/specs/markdown.yml
```diff
@@ -651,3 +651,39 @@ outputs:
...
+  new-hope/new-hope/specs/international-content/docs-vscode-validation.json:
+---
+# Link to a file within docset using relative path outside docset
+repos:
+  https://github.com/ops1/md#live:
+  - files:
+      .openpublishing.publish.config.json: |
+        {
+          "docsets_to_publish": [{ "build_source_folder": "docs/docs" }]
+        }
```
yufeih: nit: can add a  `docset.yml:` file to simplify the config

src/docfx/build/dependency/LinkResolver.cs
live1206: It's just a private method, do not feel like a private field is needed for it.

PR #5543, opened 2020-02-20
monikerParser should be case-insensitive

928PJY:
Fix: https://ceapex.visualstudio.com/Engineering/_workitems/edit/148433

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5543)


928PJY:
the moniker used inside docfx is from monikerDefinition, not from the user input, It can be case sensitive.

---
In reply to: [361698313](https://github.com/dotnet/docfx/pull/5543#issuecomment-361698313) [](ancestors = 361698313)

PR #5552, opened 2020-02-21
feat: recognize more environment in docfx OpsConfig

fenxu:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5552)

### DIFF
--- a/src/docfx/config/ops/DocsEnvironment.cs
+++ b/src/docfx/config/ops/DocsEnvironment.cs
@@ -7,8 +7,8 @@ namespace Microsoft.Docs.Build
 {
     public enum DocsEnvironment
     {
-        Production,
-        Sandbox,
+        Prod,
+        PPE,
         Internal,
         Perf,
     }
--- a/src/docfx/config/ops/DocsEnvironment.cs
+++ b/src/docfx/config/ops/DocsEnvironment.cs
@@ -9,7 +9,7 @@ namespace Microsoft.Docs.Build
     {
         Production,
         Sandbox,
-        @Internal,
+        Internal,
         Perf,
     }
 }
--- a/src/docfx/config/ops/DocsEnvironment.cs
+++ b/src/docfx/config/ops/DocsEnvironment.cs
@@ -1,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+namespace Microsoft.Docs.Build
+{
+    public enum DocsEnvironment
+    {
+        Production,
+        Sandbox,
+        @Internal,
+        Perf,
+    }
+}
--- a/src/docfx/config/ops/OpsConfigAdapter.cs
+++ b/src/docfx/config/ops/OpsConfigAdapter.cs
@@ -17,22 +17,24 @@ namespace Microsoft.Docs.Build
 {
     internal class OpsConfigAdapter : IDisposable
     {
-        public static string ValidationServiceEndpoint =>
-            s_isProduction
-            ? "https://docs.microsoft.com"
-            : "https://ppe.docs.microsoft.com";
+        public static string ValidationServiceEndpoint => GetValue(
+            production: "https://docs.microsoft.com",
+            sandbox: "https://ppe.docs.microsoft.com",
+            @internal: "https://ppe.docs.microsoft.com",
+            perf: "https://ppe.docs.microsoft.com");
 
         public const string BuildConfigApi = "https://ops/buildconfig/";
         private const string MonikerDefinitionApi = "https://ops/monikerDefinition/";
         private const string MetadataSchemaApi = "https://ops/metadataschema/";
         private const string MarkdownValidationRulesApi = "https://ops/markdownvalidationrules/";
 
-        private static readonly string? s_environment = Environment.GetEnvironmentVariable("DOCS_ENVIRONMENT");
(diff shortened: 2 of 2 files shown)


src/docfx/config/ops/DocsEnvironment.cs
```diff
@@ -0,0 +1,15 @@
...
+
+#nullable enable
+
+namespace Microsoft.Docs.Build
+{
+    public enum DocsEnvironment
+    {
+        Production,
+        Sandbox,
+        @Internal,
```
herohua: remove `@`?

fenxu: Sure. My original thoughts is keeping this same with paramter name. However seems no necessity.

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -17,22 +17,24 @@ namespace Microsoft.Docs.Build
 {
     internal class OpsConfigAdapter : IDisposable
     {
-        public static string ValidationServiceEndpoint =>
-            s_isProduction
-            ? "https://docs.microsoft.com"
-            : "https://ppe.docs.microsoft.com";
+        public static string ValidationServiceEndpoint => GetValue(
```
yufeih: with `switch expression`, we don't need this `GetValue` helper method and the code is more readable:

```suggestion
        public static string ValidationServiceEndpoint => env switch
        {
               prod => '',
               _ =>
        };
```

fenxu: We need overwrite method to the env, look at method: GetXrefMapApiEndpoint

fenxu: Maybe I can also change that.

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -17,22 +17,30 @@ namespace Microsoft.Docs.Build
...
-            s_isProduction
-            ? "https://docs.microsoft.com"
-            : "https://ppe.docs.microsoft.com";
+        public static string ValidationServiceEndpoint => s_docsEnvironment switch
+        {
+            DocsEnvironment.Production => "docs.microsoft.com",
+            DocsEnvironment.Sandbox => "ppe.docs.microsoft.com",
+            DocsEnvironment.Internal => "ppe.docs.microsoft.com",
+            DocsEnvironment.Perf => "ppe.docs.microsoft.com",
+            _ => string.Empty
```
yufeih: nit: consider `throw` here because if we reached here, there must be some code bugs, better stop immediately.
```suggestion
            _ => throw InvalidOperationException(),
```

src/docfx/config/ops/DocsEnvironment.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#nullable enable
+
+namespace Microsoft.Docs.Build
+{
+    public enum DocsEnvironment
+    {
+        Production,
```
yufeih: should it be `Prod`, `PPE` then?

fenxu: This is only internal name and I think it's OK to have Production and Sandbox which is same with before.

yufeih: I mean will `Enum.TryParse` fail?

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -17,22 +17,30 @@ namespace Microsoft.Docs.Build
...
-        private static readonly string s_buildServiceEndpoint = s_isProduction
-            ? "https://op-build-prod.azurewebsites.net"
-            : "https://op-build-sandbox2.azurewebsites.net";
+        private static readonly string s_buildServiceEndpoint = s_docsEnvironment switch
+        {
+            DocsEnvironment.Prod => "https://op-build-prod.azurewebsites.net",
+            DocsEnvironment.PPE => "https://op-build-sandbox2.azurewebsites.net",
+            DocsEnvironment.Internal => "https://op-build-internal.azurewebsites.net",
+            DocsEnvironment.Perf => "https://op-build-internal.azurewebsites.net",
+            _ => "https://op-build-sandbox2.azurewebsites.net",
```
live1206: should this throw as well?

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -227,24 +240,56 @@ private static string GetHostName(string siteName)
...
         private static bool IsLive(string branch)
         {
             return branch == "live" || branch == "live-sxs";
         }
+
+        private static DocsEnvironment GetDocsEnvironment()
+        {
+            return Enum.TryParse(Environment.GetEnvironmentVariable("DOCS_ENVIRONMENT"), true, out DocsEnvironment docsEnvironment)
+                ? docsEnvironment
+                : DocsEnvironment.PPE;
```
yufeih: When `DOCS_ENVIRONMENT` is not set, fallback to `Prod` environment. Because for docfx shipped to an end user, it should connect to our production environment without user doing anything special.

```suggestion
                : DocsEnvironment.Prod;
```

src/docfx/config/ops/OpsConfigAdapter.cs
live1206: thought s_docsEnvironment was default as "PPE" in line 292, and it should still throw here

PR #5556, opened 2020-02-24
Fields should be private

live1206:
To avoid bugs like #5540, we would like to make fields private only.

And also, public fields are not recommended in [Docs](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/fields)
image

Property is not that tedious as it evolves

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5556)

### DIFF
--- a/src/docfx/build/metadata/GlobalMetadata.cs
+++ b/src/docfx/build/metadata/GlobalMetadata.cs
@@ -13,13 +13,13 @@ namespace Microsoft.Docs.Build
     internal class GlobalMetadata
     {
         // For v2 backward compatibility
-        public HashSet<string> ContributorsToExclude { get; set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        public HashSet<string> ContributorsToExclude { get; private set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
         // For v2 backward compatibility
         [JsonProperty("_op_documentIdPathDepotMapping")]
-        public Dictionary<PathString, DocumentIdConfig> DocumentIdDepotMapping { get; set; }
+        public Dictionary<PathString, DocumentIdConfig> DocumentIdDepotMapping { get; private set; }
 
         [JsonExtensionData]
-        public JObject ExtensionData { get; set; } = new JObject();
+        public JObject ExtensionData { get; private set; } = new JObject();
     }
 }
--- a/src/docfx/build/metadata/GlobalMetadata.cs
+++ b/src/docfx/build/metadata/GlobalMetadata.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Docs.Build
     internal class GlobalMetadata
     {
         // For v2 backward compatibility
-        public HashSet<string> ContributorsToExclude { get; private set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        public HashSet<string> ContributorsToExclude { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
 
         // For v2 backward compatibility
         [JsonProperty("_op_documentIdPathDepotMapping")]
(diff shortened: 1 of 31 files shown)


yufeih:
I don't quite follow how this relates to #5540

live1206:
> I don't quite follow how this relates to #5540

If it was a public property, not a public field in the first place. It would not cause problems

yufeih:
can also remove the `JsonUtility` patch if we decide to not use `readonly` field for data models.

live1206:
> can also remove the `JsonUtility` patch if we decide to not use `readonly` field for data models.

not really, we still have `public readonly` fields for `struct`:
https://github.com/dotnet/docfx/blob/31edbda9ca8231fd779ab6b6567611220dc493c8/src/docfx/config/DocumentIdConfig.cs#L14

yufeih:
> > can also remove the `JsonUtility` patch if we decide to not use `readonly` field for data models.
> 
> not really, we still have `public readonly` fields for `struct`:
> https://github.com/dotnet/docfx/blob/31edbda9ca8231fd779ab6b6567611220dc493c8/src/docfx/config/DocumentIdConfig.cs#L14

It could also be turned into property I guess because it falls into the category of data serialization objects?

src/docfx/lib/log/Progress.cs
```diff
@@ -68,10 +68,11 @@ public static string FormatTimeSpan(TimeSpan value)
 
         private class LogScope : IDisposable
```
yufeih: This could be a struct though, thus it can have fields

```suggestion
        private struct LogScope : IDisposable
```

live1206:
> > > can also remove the `JsonUtility` patch if we decide to not use `readonly` field for data models.
> > 
> > 
> > not really, we still have `public readonly` fields for `struct`:
> > https://github.com/dotnet/docfx/blob/31edbda9ca8231fd779ab6b6567611220dc493c8/src/docfx/config/DocumentIdConfig.cs#L14
> 
> It could also be turned into property I guess because it falls into the category of data serialization objects?



> > > can also remove the `JsonUtility` patch if we decide to not use `readonly` field for data models.
> > 
> > 
> > not really, we still have `public readonly` fields for `struct`:
> > https://github.com/dotnet/docfx/blob/31edbda9ca8231fd779ab6b6567611220dc493c8/src/docfx/config/DocumentIdConfig.cs#L14
> 
> It could also be turned into property I guess because it falls into the category of data serialization objects?

Since this is a `readonly` struct, it can not have a property with `private` setter, do we want to remove the struct not `readonly`?

src/docfx/build/metadata/GlobalMetadata.cs
```diff
@@ -13,13 +13,13 @@ namespace Microsoft.Docs.Build
     internal class GlobalMetadata
     {
         // For v2 backward compatibility
-        public HashSet<string> ContributorsToExclude { get; set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
+        public HashSet<string> ContributorsToExclude { get; private set; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```
yufeih: nit: For dictionaries, `{ get; }` can also work with JSON deserialization. Thus I would prefer `{ get; }` to make it clear that JSON deserialization is reusing our dictionary, instead of create a new dictionary and overriding ours. (Note the `StringComparer` here, it has to be our own dictionary otherwise code logic may fail.

```suggestion
        public HashSet<string> ContributorsToExclude { get; } = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
```

yufeih:
> > > > can also remove the `JsonUtility` patch if we decide to not use `readonly` field for data models.
> > > 
> > > 
> > > not really, we still have `public readonly` fields for `struct`:
> > > https://github.com/dotnet/docfx/blob/31edbda9ca8231fd779ab6b6567611220dc493c8/src/docfx/config/DocumentIdConfig.cs#L14
> > 
> > 
> > It could also be turned into property I guess because it falls into the category of data serialization objects?
> 
> Since this is a `readonly` struct, it can not have a property with `private` setter, do we want to make the struct not `readonly`?

yep, removing it is fine for me

PR #5559, opened 2020-02-25
keep using StringComparer.Ordinal for moniker sorting

928PJY:
revert the change in #5543 

monikers `["azuresqldb-current", "azure-sqldw-latest"]`  
by using the default comparer, it will be sorted to  `["azuresqldb-current", "azure-sqldw-latest"]`
by using the `StringComparer.OrdinalIgnoreCase`, it will be sorted to  `["azure-sqldw-latest", "azuresqldb-current"]`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5559)

### DIFF
--- a/src/docfx/lib/moniker/MonikerRangeParser.cs
+++ b/src/docfx/lib/moniker/MonikerRangeParser.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Docs.Build
                     return ExpressionCreator.Create(value)
                         .Accept(_monikersEvaluator)
                         .Select(x => x.MonikerName.ToLowerInvariant())
-                        .OrderBy(_ => _, StringComparer.OrdinalIgnoreCase)
+                        .OrderBy(_ => _, StringComparer.Ordinal)
                         .ToArray();
                 }
                 catch (MonikerRangeException ex)
--- a/src/docfx/lib/moniker/MonikerRangeParser.cs
+++ b/src/docfx/lib/moniker/MonikerRangeParser.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Docs.Build
                     return ExpressionCreator.Create(value)
                         .Accept(_monikersEvaluator)
                         .Select(x => x.MonikerName.ToLowerInvariant())
-                        .OrderBy(_ => _)
+                        .OrderBy(_ => _, StringComparer.OrdinalIgnoreCase)
                         .ToArray();
                 }
                 catch (MonikerRangeException ex)
--- a/test/docfx.Test/lib/MonikerRangeParserTest.cs
+++ b/test/docfx.Test/lib/MonikerRangeParserTest.cs
@@ -41,10 +41,21 @@ namespace Microsoft.Docs.Build
                     Order = 3
                 },
                 new Moniker
-                {
+                 {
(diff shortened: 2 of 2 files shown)


live1206:
can we have a test to cover this?

yufeih:
Can we change v2 and keep the v3 behavior?

This makes moniker hash generation depending on an external file. Which makes the algorithm complex.

FYI, `Docs.Exporter` is also [generating](https://dev.azure.com/ceapex/Engineering/_git/Docs.Exporter?path=%2Fsrc%2FDocsExporter%2FModels%2FExporter%2FDocument.cs&version=GBdevelop&line=36&lineEnd=36&lineStartColumn=13&lineEndColumn=16&lineStyle=plain) this `moniker-list-hash`, because it will be used as a `key` (document identifier) in the new blob based DHS.

928PJY:
Added

---
In reply to: [590695724](https://github.com/dotnet/docfx/pull/5559#issuecomment-590695724) [](ancestors = 590695724)

yufeih:
> Can we change v2 and keep the v3 behavior?
> 
> This makes moniker hash generation depending on an external file. Which makes the algorithm complex.
> 
> FYI, `Docs.Exporter` is also [generating](https://dev.azure.com/ceapex/Engineering/_git/Docs.Exporter?path=%2Fsrc%2FDocsExporter%2FModels%2FExporter%2FDocument.cs&version=GBdevelop&line=36&lineEnd=36&lineStartColumn=13&lineEndColumn=16&lineStyle=plain) this `moniker-list-hash`, because it will be used as a `key` (document identifier) in the new blob based DHS. I am worried if this algorithm depend on `moniker definition`, the logic will spread across multiple locations and is hard to get it right.

I realized that this isn't related to moniker definition, but still, we need to agree on the sort comparer everywhere.

928PJY:
Yes, you are right, but from the current situation, seems this is the only place which involves the sort comparer

---
In reply to: [590705303](https://github.com/dotnet/docfx/pull/5559#issuecomment-590705303) [](ancestors = 590705303)

test/docfx.Test/lib/MonikerRangeParserTest.cs
```diff
@@ -41,11 +41,22 @@ public class MonikerRangeParserTest
                     Order = 3
                 },
                 new Moniker
-                {
+                 {
```
live1206: remove the space

src/docfx/lib/moniker/MonikerRangeParser.cs
```diff
@@ -33,7 +33,7 @@ public string[] Parse(SourceInfo<string> rangeString)
                     return ExpressionCreator.Create(value)
                         .Accept(_monikersEvaluator)
                         .Select(x => x.MonikerName.ToLowerInvariant())
-                        .OrderBy(_ => _)
+                        .OrderBy(_ => _, StringComparer.OrdinalIgnoreCase)
```
yufeih:
```suggestion
                        .OrderBy(_ => _, StringComparer.Ordinal)
```

PR #5560, opened 2020-02-25
fix: Ensure a single key has only one entry in JsonDiskCache

yufeih:
Fixes: https://dev.azure.com/ceapex/Engineering/_workitems/edit/174735/

The root cause is we have two different entries of the same key. This PR updates `JsonDiskCache` to ensure a single key has only a single value.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5560)


src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -133,6 +126,17 @@ public async Task<TError[]> Save()
             });
         }
 
+        private void UpdateCache(TValue item)
+        {
+            foreach (var cacheKey in item.GetKeys())
+            {
+                if (!_cache.TryGetValue(cacheKey, out var existingValue) || HasExpired(existingValue))
```
live1206: This would prevent overwriting the existing non-expired value for the upcoming run.
But what about the corrupted cache file on the build server? Same email could match 2 non-expired entry(1 valid and 1 invalid), shall we clean that up as well?

yufeih: We only update the entry when it is a new entry or the cache has expired, this check does exactly that. Won't trigger update when a cached item hasn't expired.

The code will auto-heal corrupted cache files, so no need for manual clean up.

mingwli: what if the corrupted entry get loaded first?

live1206: Actually, once the cache file loaded and then save, there will be only 1 entry saved to the cache file.
Let's say the invalid entry(email with null user) saved, will that be updated?

And we can remove the filtering expired on save since they should be updated here.
https://github.com/dotnet/docfx/blob/db43f6b6162ad8d6d8120e2b9424b9212482449a/src/docfx/lib/cache/JsonDiskCache.cs#L96

src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -133,6 +127,19 @@ public async Task<TError[]> Save()
             });
         }
 
+        private void UpdateCache(TValue value)
+        {
+            foreach (var cacheKey in value.GetKeys())
+            {
+                if (!_cache.TryGetValue(cacheKey, out var existingValue) || HasExpired(existingValue))
```
mingwli: >_cache.TryGetValue(cacheKey, out var existingValue) [](start = 21, length = 51)

It won't hit here when not expired, right?

src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -133,6 +127,19 @@ public async Task<TError[]> Save()
...
+                {
+                    _cache[cacheKey] = _resolveConflict != null && existingValue != null
```
live1206: `_resolveConflict` should only apply for non-expired existing value

src/docfx/lib/cache/JsonDiskCache.cs
live1206: it can, if value exists or expired(not and)

src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -239,5 +239,17 @@ private async Task<HttpResponseMessage> SendRequest(string api, StringContent re
...
+        private static GitHubUser ResolveGitHubUserConflict(GitHubUser a, GitHubUser b)
+        {
+            // Pick the user with a valid Id
+            if (a.Id is null)
+                return b;
+            if (b.Id is null)
+                return a;
+
+            // otherwise pick a random one
+            return a;
```
mingwli: what about using latest by comparing update_at?

yufeih: It is not expected to enter here, thus I'm using the simplest approach.

mingwli: what about the user rename case?

---
In reply to: [383781441](https://github.com/dotnet/docfx/pull/5560#discussion_r383781441) [](ancestors = 383781441)

Issue #5536, opened 2020-02-18
ArgumentNullException in newer version of docfx NuGet (2.49)


wiwei:
**Operation System**: Windows

**DocFX Version Used**: 2.49

**Template used**: `default`

**Steps to Reproduce**:
Run docfx docfx.json -f on the following project:

https://github.com/microsoft/MixedRealityToolkit-Unity/blob/mrtk_development/docfx.json

https://dev.azure.com/aipmr/MixedRealityToolkit-Unity-CI/_build/results?buildId=9516&view=logs&j=275f1d19-1bd8-5591-b06b-07d489ea915a&t=a4822bc1-f3a9-5984-c941-6c1b7aa2ab81

Note that this seems to be a regression in the latest docfx NuGet package. We previously were using the latest (2.48.1) which was working fine, and after pulling the newer package (2.49) this issue started happening. To work around this we're forcing our infrastructure to use an older version (i.e. 2.48.1) which is still working fine for us.

**Expected Behavior**:
No exception

**Actual Behavior**:
2020-02-18T18:04:54.4278760Z Parameter name: key
2020-02-18T18:04:54.4279171Z    at System.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
2020-02-18T18:04:54.4279350Z    at System.Collections.Generic.Dictionary`2.FindEntry(TKey key)
2020-02-18T18:04:54.4279504Z    at System.Collections.Generic.Dictionary`2.TryGetValue(TKey key, TValue& value)
2020-02-18T18:04:54.4279710Z    at Microsoft.DocAsCode.Metadata.ManagedReference.CopyInherited.Copy(MetadataItem dest, String srcName, ResolverContext context)
2020-02-18T18:04:54.4279934Z    at Microsoft.DocAsCode.Metadata.ManagedReference.CopyInherited.InheritDoc(MetadataItem dest, ResolverContext context)
2020-02-18T18:04:54.4280101Z    at Microsoft.DocAsCode.Metadata.ManagedReference.CopyInherited.<>c__DisplayClass0_0.<Run>b__1(MetadataItem current, MetadataItem parent)
2020-02-18T18:04:54.4280283Z    at Microsoft.DocAsCode.Common.TreeIterator.Preorder[T](T current, T parent, Func`2 childrenGetter, Func`3 action)
[...]

PR #5561, opened 2020-02-25
Safe guard against not existing definition when inheriting doc

Laniusexcubitor:
When using inheritdoc on members or classes where no base documentation can be resolved docfx can throw a ArgumentNullException.

#5536 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5561)

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference/Resolvers/CopyInheritedDocumentation.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference/Resolvers/CopyInheritedDocumentation.cs
@@ -142,7 +142,9 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
                 if (!context.References.TryGetValue(srcName, out var referenceItem) || referenceItem.Definition == null ||
                     !context.Members.TryGetValue(referenceItem.Definition, out src))
                 {
-                    Logger.LogWarning($"Could not resolve base documentation for '{dest.Name}'");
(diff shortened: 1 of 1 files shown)


src/Microsoft.DocAsCode.Metadata.ManagedReference/Resolvers/CopyInheritedDocumentation.cs
```diff
@@ -139,7 +139,7 @@ private static void Copy(MetadataItem dest, string srcName, ResolverContext cont
             if (src == null && !context.Members.TryGetValue(srcName, out src))
             {
                 // Try to resolve any templated references before giving up
-                if (!context.References.TryGetValue(srcName, out var referenceItem) || !context.Members.TryGetValue(referenceItem.Definition, out src))
+                if (!context.References.TryGetValue(srcName, out var referenceItem) || referenceItem.Definition == null || !context.Members.TryGetValue(referenceItem.Definition, out src))
```
Laniusexcubitor: I'd like to generate a warning here telling about the issue so the developer can fix their inheritdoc usage.
Can we create warnings here?

yufeih: yes, `Logger` is a static class that you can use to create warnings.

PR #5562, opened 2020-02-25
fix: handle the case that base path not started by '/'

fenxu:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5562)

### DIFF
--- a/src/docfx/config/ops/OpsConfigAdapter.cs
+++ b/src/docfx/config/ops/OpsConfigAdapter.cs
@@ -105,7 +105,7 @@ namespace Microsoft.Docs.Build
             var xrefMapApiEndpoint = GetXrefMapApiEndpoint(xrefEndpoint);
             if (docset.base_path != "/")
             {
-                xrefQueryTags.Add($"/{docset.base_path.TrimStart('/')}");
+                xrefQueryTags.Add(docset.base_path);
             }
             var xrefMaps = new List<string>();
             foreach (var tag in xrefQueryTags)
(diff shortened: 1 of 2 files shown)


live1206:
we need to add a spec test for this case.
Is there a real usage repo for reference?

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -153,8 +153,7 @@ private async Task<string[]> GetXrefMaps(string xrefMapApiEndpoint, string tag,
         {
             var url = $"{xrefMapApiEndpoint}/v1/xrefmap{tag}{xrefMapQueryParams}";
             var response = await Fetch(url, value404: "{}");
-            return JsonConvert.DeserializeAnonymousType(response, new { links = new[] { "" } }).links
-                ?? Array.Empty<string>();
+            return JsonConvert.DeserializeAnonymousType(response, new { links = new[] { "" } }).links;
```
live1206: revert this line, there was a bug fix for this

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -105,7 +105,7 @@ private async Task<string> GetBuildConfig(Uri url)
             var xrefMapApiEndpoint = GetXrefMapApiEndpoint(xrefEndpoint);
             if (docset.base_path != "/")
             {
-                xrefQueryTags.Add(docset.base_path);
+                xrefQueryTags.Add($"/{docset.base_path.TrimStart('/')}");
```
live1206: we could not have a spec test for this since this is based on OP api docset info, it should be fine then

yufeih: There is a dedicated type `BasePath` for this purpose

PR #5569, opened 2020-02-26
Publish trimmed single file

mingwli:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5569)

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -44,7 +44,7 @@ function publishBinaryPackages() {
     $rids = @("win7-x64", "osx-x64", "linux-x64") # Microsoft.ChakraCore doesn't provide win-x64 runtime build, using win7-x64
     foreach ($rid in $rids) {
         $packageName = "docfx-$rid-$version"
-        exec "dotnet publish src\docfx\docfx.csproj -c release -r $rid -o $packagesBasePath/$rid /p:Version=$version /p:InformationalVersion=$version /p:PackAsTool=false"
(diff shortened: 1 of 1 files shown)


mingwli:
The trimmed single executable lack some libraries, e.g. `LibGit2Sharp.NativeBinaries`

May need more trimmer settings documented [here](https://github.com/mono/linker/blob/master/src/ILLink.Tasks/README.md)

yufeih:
There are more to it, a lot of code will be removed and it'll cause runtime exceptions like `MissingMethodException`.

To publish trimmed binaries, we need to adjust our test strategy to catch any tree shaking bugs.

For this experiment, the goal is first get a draft idea of how much size benefit out of this.

mingwli:
> There are more to it, a lot of code will be removed and it'll cause runtime exceptions like `MissingMethodException`.
> 
> To publish trimmed binaries, we need to adjust our test strategy to catch any tree shaking bugs.
> 
> For this experiment, the goal is first get a draft idea of how much size benefit out of this.

Agreed.
|     | untrimmed | trimmed | ratio |
| ---| --- | --- | --- |
|win| 46.3m | 24.5m | 52.9% |
|osx| 42.0m | 22.4m | 53.3% |

PR #5570, opened 2020-02-26
Support xrefmap fallback for localization

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/178612

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5570)

### DIFF
--- a/src/docfx/build/xref/ExternalXrefMapLoader.cs
+++ b/src/docfx/build/xref/ExternalXrefMapLoader.cs
@@ -14,11 +14,38 @@ namespace Microsoft.Docs.Build
     {
         private static readonly byte[] s_uidBytes = Encoding.UTF8.GetBytes("uid");
 
-        public static IReadOnlyDictionary<string, Lazy<ExternalXrefSpec>> Load(Config config, FileResolver fileResolver)
+        public static IReadOnlyDictionary<string, Lazy<ExternalXrefSpec>> Load(Config config, FileResolver fileResolver, Docset fallbackDocset, Context context)
         {
             var result = new Dictionary<string, Lazy<ExternalXrefSpec>>();
 
             foreach (var url in config.Xref)
+            {
+                try
+                {
+                    LoadXref(url);
+                }
+                catch (DocfxException ex) when (ex.Error.Code == "file-not-found")
+                {
+                    FilePath path;
+                    if (fallbackDocset != null && context.Input.Exists(path = new FilePath(url, FileOrigin.Fallback)))
+                    {
+                        using var stream = context.Input.ReadStream(path);
+                        var xrefMap = JsonUtility.Deserialize<XrefMapModel>(new StreamReader(stream), path);
+                        foreach (var spec in xrefMap.References)
+                        {
+                            result.TryAdd(spec.Uid, new Lazy<ExternalXrefSpec>(() => spec));
+                        }
+                    }
+                    else
+                    {
+                        throw;
+                    }
+                }
+            }
+
+            return result;
+
+            void LoadXref(SourceInfo<string> url)
             {
                 using var stream = fileResolver.ReadStream(url);
                 var path = new FilePath(url);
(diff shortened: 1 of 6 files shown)


src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -14,11 +14,42 @@ internal class ExternalXrefMapLoader
...
+                {
+                    LoadXref(url);
+                }
+                catch (DocfxException ex) when (ex.Error.Code == "file-not-found")
+                {
+                    FilePath path;
+                    if (fallbackDocset != null && input.Exists(path = new FilePath(url, FileOrigin.Fallback)))
+                    {
+                        using var stream = input.ReadStream(path);
+                        var xrefMap = JsonUtility.Deserialize<XrefMapModel>(new StreamReader(stream), path);
```
mingwli: >var xrefMap = JsonUtility.Deserialize<XrefMapModel>(new StreamReader(stream), path); [](start = 24, length = 84)

Are we assuming that xref map file in fallback repo is in json format?
(I'm not quite familiar with the context)

src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -14,11 +14,42 @@ internal class ExternalXrefMapLoader
...
         {
             var result = new Dictionary<string, Lazy<ExternalXrefSpec>>();
 
             foreach (var url in config.Xref)
+            {
+                try
+                {
+                    LoadXref(url);
+                }
+                catch (DocfxException ex) when (ex.Error.Code == "file-not-found")
```
yufeih: Consider handle this at `FileResolver` layer to enable loc fallback for all contents referenced by config.

src/docfx/build/xref/ExternalXrefMapLoader.cs
live1206: you are right, it can be other formats

src/docfx/build/xref/ExternalXrefMapLoader.cs
live1206: Thought about that as well, did this the same behavior as `LinkResolver`
Now in `FileResolver`, it only holds `_docsetPath`, which can be default or fallback. And it is not singleton in our app, maybe we can make it singleton and hold fallback docset as well.

`DocumentProvider` is singleton in our app and holds default and fallback docset.
Maybe we can resolve this in `LinkResolver` as well.
https://github.com/dotnet/docfx/blob/8c969bfa998d5566e76ebf1c565fe7b0bcb148a8/src/docfx/build/dependency/LinkResolver.cs#L274

Prefer not to refactor them in this PR though

live1206: `FileResolver` is used for resolving external link in config, move the fallback logic within it and would come up with a better naming for it, probably make it single instance as well later

src/docfx/restore/FileResolver.cs
```diff
@@ -50,10 +60,16 @@ public Stream ReadStream(SourceInfo<string> file)
             if (!UrlUtility.IsHttp(file))
             {
                 var localFilePath = Path.Combine(_docsetPath, file);
+                FilePath fallbackFilePath;
                 if (File.Exists(localFilePath))
                 {
                     return File.OpenRead(localFilePath);
                 }
+                else if (_fallbackDocset != null && _input != null
```
yufeih: `Path.Combine(_fallbackDocset.DocsetPath, file)` already produced the full path, why not use it directly w/o Input?

The original intention of `FilePath` is to been able to construct it without using absolute path (maybe it needs a check in the constructor), so you can `new FilePath("a.md", Origin.Fallback)`.

live1206: Good point, we only have files in the fallback docset, just a different local file path.
Adding a check for `FilePath`

live1206: the check would cause more code change, fix it in a new PR instead

Issue #5515, opened 2020-02-11
Enable nullable reference types


yufeih:
https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

yufeih:
Done 👏

PR #5573, opened 2020-02-27
Enable nullable checks for metadata

yufeih:
#5515 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5573)

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -529,7 +529,7 @@ repos:
           Osmond creates this, without email
 inputs:
   docfx.yml: |
-    editRepositoryUrl: https://docascode.visualstudio.com/contribution1/_git/a
+    editRepositoryUrl: https://github.com/a/b
 cache:
   github-users.json: |
       { "items": [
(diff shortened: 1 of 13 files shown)


mingwli:
Cache is being weird :P

yufeih:
> Cache is being weird :P

This is probably due to my change

docs/specs/git.yml
```diff
@@ -552,6 +552,30 @@ outputs:
       "updated_at": "2018-10-30 12:00 AM"
     }
 ---
+# Do not resolve contributors if repo and editRepositoryUrl are not GitHub repository
```
live1206: could this be a separated PR?

yufeih: This bug is actually detected by nullability check, see https://github.com/dotnet/docfx/pull/5573/files#diff-63c67683598f3abbc11bdf942d99d29eL82, the old implementation assumes `editRepositoryUrl` is correct and can potentially leak `null` to `GitHubAccessor`, after enable nullable check, the compiler help us detect this, so I'd like to include it in this PR to illustrate how nullable check helps write robust code.

PR #5574, opened 2020-02-27
Gracefully handle git crash

yufeih:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/171333/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5574)

### DIFF
--- a/src/docfx/restore/PackageResolver.cs
+++ b/src/docfx/restore/PackageResolver.cs
@@ -135,6 +135,10 @@ namespace Microsoft.Docs.Build
             Directory.CreateDirectory(cwd);
             GitUtility.Init(cwd);
 
+            // Remove git lock files if previous build was killed during git operation
+            DeleteIfExist(Path.Combine(cwd, ".git/index.lock"));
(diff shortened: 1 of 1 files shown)


src/docfx/restore/PackageResolver.cs
```diff
@@ -152,6 +156,8 @@ private void DownloadGitRepositoryCore(string cwd, string url, string committish
                 }
             }
 
+            GitUtility.Clean(cwd, "-xdff");
```
live1206: in which case will there be untracked files?

yufeih: I originally thought if something bad happens during checkout, `git clean` may help us. But most likely git index file will also be corrupted by then so this isn't likely to help.

This PR still cannot 100% handle git crash, but would mitigate most of the common scenarios. That is good enough for now.

PR #5580, opened 2020-02-28
Sanitize HTML

yufeih:
[AB#178608](https://dev.azure.com/ceapex/Engineering/_workitems/edit/178608)

HTML sanitizer spec.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5580)

### DIFF
--- a/docs/designs/sanitize-html.md
+++ b/docs/designs/sanitize-html.md
@@ -57,7 +57,7 @@ Use an _allow list_ to sanitize HTML tags and attributes. HTML tags can contain 
 
 Each allowed HTML tag can have additional attributes as follows:
 
-> NOTE: This list may be adjustmented based on docs usage collected from telemetry
+> NOTE: This list may be adjusted based on docs usage collected from telemetry
 
 ```html
 <!-- Content sectioning -->
--- a/docs/designs/sanitize-html.md
+++ b/docs/designs/sanitize-html.md
@@ -51,17 +51,17 @@ Use an _allow list_ to sanitize HTML tags and attributes. HTML tags can contain 
 - Standard HTML5 global attributes:
 
   ```html
-  id, itemid, itemprop, itemref, itemscope, itemtype,
+  name, id, itemid, itemprop, itemref, itemscope, itemtype,
   part, slot, spellcheck, title
   ```
 
 Each allowed HTML tag can have additional attributes as follows:
 
+> NOTE: This list may be adjustmented based on docs usage collected from telemetry
 
 ```html
 <!-- Content sectioning -->
 <address>: 
-<h1>-<h6>: 
 <section>:
 
 <!-- Text content -->
@@ -80,7 +80,7 @@ Each allowed HTML tag can have additional attributes as follows:
 <ul>:
 
 <!-- Inline text semantics -->
-<a>: download, href, hreflang, ping, rel, target, type
+<a>: href
 <abbr>:
 <b>:
 <bdi>:
@@ -88,7 +88,6 @@ Each allowed HTML tag can have additional attributes as follows:
 <br>:
 <cite>:
 <code>:
-<data>: value
 <dfn>:
 <em>:
 <i>:
@@ -106,7 +105,7 @@ Each allowed HTML tag can have additional attributes as follows:
 <var>:
 
 <!-- Image and multimedia -->
-<img>: alt, decoding, height, intrinsicsize, loading, sizes, src, width
+<img>: alt, height, src, width
 
 <!-- Demarcating edits -->
 <del>: cite, datetime
@@ -118,14 +117,14 @@ Each allowed HTML tag can have additional attributes as follows:
 <colgroup>:
 <table>:
 <tbody>:
-<td>: colspan, headers, rowspan
+<td>: colspan, rowspan
 <tfoot>:
-<th>: abbr, colspan, headers, rowspan, scope
+<th>: colspan, rowspan, scope
 <thead>:
 <tr>:
 
 <pre>:
-<iframe>: allow, allowfullscreen, allowpaymentrequest, height, name, referrerpolicy, sandbox, src, srcdoc, width
+<iframe>: allowfullscreen, height, src, width
 ```
 
 > The above table is based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element. It
--- a/docs/designs/sanitize-html.md
+++ b/docs/designs/sanitize-html.md
@@ -1,0 +1,144 @@
+# Sanitize HTML
+
+[Feature 178608](https://dev.azure.com/ceapex/Engineering/_workitems/edit/178608/)
+
+## Overview
+
+Content authors can embed arbitrary HTML tags to markdown documents. These HTML tags could contain `<script>` tag that allows arbitrary script execution, or `style` attribute that compromises site visual experience.
+
+### Goals
+
+- Remove HTML tags and attributes from user content that may alter site style
+- Remove HTML tags and attributes from user content that may allow javascript execution
+- Apply HTML sanitization consistently for all user content
+
+### Out of scope
+
+- Report warnings for removing disallowed HTML tags and attributes
+- Refactor HTML processing pipeline
+
+## Technical Design
+
+The sanitization is performed against user HTML, that is in most cases HTML tags inside markdown files. It does not sanitize HTML produced from the system, including markdown extension, jint, mustache or liquid template.
+
+#### Sanitize URL
+
+To prevent script execution like `<a href="javascript://void">`, URL schemas are sanitized using a _disallow list_. Restricting URL schemas to `http:` or `https:` only is not the scope of this feature. The _disallow list_ is:
+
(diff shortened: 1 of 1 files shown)


docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Appendix
+
+#### HTML5 attribute name allowlist
+
+- This list is based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element.
+- This list exclude experimental and obsolete attributes.
+- This list exclude user interactive DOM elements and attributes (like `<button>`, `<menu>`).
+- This list exclude media DOM elements (like `<video>`, `<audio>`).
+- This list exclude DOM elements that are affect main site structure (like `<body>` and `<header>`)
```
mingwli:
```suggestion
- This list excludes DOM elements that may affect main site structure (e.g. `<body>` and `<header>`)
```

test/docfx.Test/lib/HtmlUtilityTest.cs
```diff
@@ -60,7 +64,7 @@ public void HtmlStripTags(string input, string output)
         [InlineData("<td style='text-align: left;'>", "<td style='text-align: left;'>")]
         [InlineData("<td style='text-align: center;'>", "<td style='text-align: center;'>")]
         [InlineData("<td style='text-align: right;'>", "<td style='text-align: right;'>")]
```
mingwli: add a test ?
```
 [InlineData("<td style='text-align: right; blahblahblah?'>", "<td style='text-align: right;'>")]
```

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Overview
+
+Content authors can embed arbitrary HTML tags to markdown documents. These HTML tags could contain `<script>` tag that allows arbitrary script execution, or `style` attribute that compromises site visual experience.
+
+### Goals
+
+- Remove HTML tags and attributes from user content that may alter site style
+- Remove HTML tags and attributes from user content that may allow javascript execution
+- Apply HTML sanitization consistently for all user content
```
zhiliangxu: Does that include all reference contents and YAML contents?

zhiliangxu: If including reference and YAML contents cost much effort, can we consider only including Markdown files as first phase?

yufeih: The current HTML post processor runs against all user contents, including YAML or reference, this is 0 cost 😄

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Out of scope
+
+- Report warnings for removing disallowed HTML tags and attributes
```
jdanyow, line 17: Why is this out of scope? Perf? 
1. It would be awesome if these violations were visible to authors.
1. It would be super duper awesome if these violations went into some reporting sink so we could identify and chip away at content using unsupported markup.

From the template/ux perspective we really don't want folks using html at all.

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,144 @@
+# Sanitize HTML
```
jdanyow, line 1: What do you think about introducing sanitization "modes" that we can refer to in our yaml schemas?
1. "Loose": the default, applies to markdown files and yaml properties strings with `contentType: markdown`.
2. "Strict": disallows html entirely. We'd start to reference this in [our schemas](https://dev.azure.com/ceapex/Engineering/_git/docs-ui/?path=%2Fpackages%2Fschemas%2Fsrc%2FModule.schema.json&version=GBmatthew%2Fsystemic-focus-colors&line=21&lineEnd=26&lineStartColumn=1&lineEndColumn=5&lineStyle=plain) via `contentType: markdown-strict` or something.

Overall we'd like to discourage use of html in content as it's a constant source of accessibility bugs, layout issues, design regressions.

adkinn: We might also need an special permissive mode for the currently archived content. I don't believe anyone wants to go back and update those pages unless they have to. The permissive mode should not be accessible to other repos though.

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,144 @@
...
+- Based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element.
+- Excludes experimental and obsolete attributes.
+- Excludes user interactive DOM elements and attributes (like `<button>`, `<menu>`).
+- Excludes media DOM elements (like `<video>`, `<audio>`).
+- Excludes DOM elements that affect main site structure (like `<body>` and `<header>`)
+
+**Global attributes**
+
+```html
+class, dir, hidden, id,
```
adkinn, line 69: I think some of these attributes should not be in the allowlist. "class", "dir", "hidden", "lang" and "tabindex" all have UI and Accessibility implications that we don't want the authors customizing.

docs/designs/sanitize-html.md
yufeih: It was in a separate feature. But it make sense to include both in this spec.

docs/designs/sanitize-html.md
jdanyow: How about:
1. "Loose": backwards compat... archive repos, certain page layouts like the old hub and landing pages.
2. "Standard": essentially this spec in it's current form
3. "Strict": disallow html

All modes can be activated by metadata/schema contentType.

yufeih: We can automatically detect hub/landing/archived contents to save us from changing the configs for these repos/files,

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,134 @@
...
+<figure>:
+<hr>:
+<li>: value
+<ol>: reversed, start, type
+<p>:
+<pre>:
+<ul>:
+
+<!-- Inline text semantics -->
+<a>: download, href, hreflang, ping, rel, target, type
```
jdanyow, line 83: `href` only in both loose and strict modes

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,134 @@
...
+<abbr>:
+<b>:
+<bdi>:
+<bdo>:
+<br>:
+<cite>:
+<code>:
+<data>: value
```
jdanyow, line 91: not needed?

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,134 @@
...
+<span>:
+<strong>:
+<sub>:
+<sup>:
+<time>: datetime
+<u>:
+<var>:
+
+<!-- Image and multimedia -->
+<img>: alt, decoding, height, intrinsicsize, loading, sizes, src, width
```
jdanyow, line 109: `alt`, `src`, `width`, `height`

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,134 @@
...
+
+<!-- table -->
+<caption>:
+<col>:
+<colgroup>:
+<table>:
+<tbody>:
+<td>: colspan, headers, rowspan
+<tfoot>:
+<th>: abbr, colspan, headers, rowspan, scope
```
jdanyow, line 123: `colspan`, `rowspan`, `scope`

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,134 @@
...
+<thead>:
+<tr>:
+
+<pre>:
+<iframe>: allow, allowfullscreen, allowpaymentrequest, height, name, referrerpolicy, sandbox, src, srcdoc, width
```
jdanyow, line 128: - disallow in standard mode.
- `name`, `src`, `width`, `height`, `allowfullscreen`

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,134 @@
...
+  part, slot, spellcheck, title
+  ```
+
+Each allowed HTML tag can have additional attributes as follows:
+
+
+```html
+<!-- Content sectioning -->
+<address>: 
+<h1>-<h6>:
```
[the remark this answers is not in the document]
yufeih: There is a validation feature to block h1-h6 entirely for user written HTML, I'm removing them entirely from here.

docs/designs/sanitize-html.md
```diff
@@ -0,0 +1,133 @@
...
+- Standard HTML5 global attributes:
+
+  ```html
+  name, id, itemid, itemprop, itemref, itemscope, itemtype,
+  part, slot, spellcheck, title
+  ```
+
+Each allowed HTML tag can have additional attributes as follows:
+
+> NOTE: This list may be adjustmented based on docs usage collected from telemetry
```
Paul-Dempsey, line 60: "adjustmented" is not an English word. Should be "adjusted"

yufeih: oops, my silly typo, Thanks!

PR #5588, opened 2020-03-03
Enable nullable checks for publish

yufeih:
#5515 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5588)

### DIFF
--- a/src/docfx/build/redirection/BuildRedirection.cs
+++ b/src/docfx/build/redirection/BuildRedirection.cs
@@ -18,23 +18,17 @@ namespace Microsoft.Docs.Build
             var (monikerError, monikers) = context.MonikerProvider.GetFileLevelMonikers(file.FilePath);
             errors.AddIfNotNull(monikerError);
 
-            var publishItem = new PublishItem
-            {
-                Url = file.SiteUrl,
-                SourcePath = file.FilePath.Path,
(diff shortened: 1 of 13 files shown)


src/docfx/build/redirection/BuildRedirection.cs
```diff
@@ -18,23 +18,17 @@ internal static List<Error> Build(Context context, Document file)
...
-            }
+            var publishItem = new PublishItem(
+                file.SiteUrl,
+                context.Config.Legacy ? context.DocumentProvider.GetOutputPath(file.FilePath, monikers) : null,
+                file.FilePath.Path,
+                context.LocalizationProvider.Locale,
+                monikers,
+                context.MonikerProvider.GetConfigMonikerRange(file.FilePath));
+
+            publishItem.RedirectUrl = context.RedirectionProvider.GetRedirectUrl(file.FilePath);
```
live1206: why can not `RedirectUrl` be in the constructor as well?

yufeih: It is not mandatory, same as other properties like Metadata, and in cases like metadata, it needs to be set after the publish item is created.

PR #5591, opened 2020-03-03
Enable nullable reference type for xref and build page

yufeih:
#5515 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5591)


src/docfx/build/xref/ExternalXrefSpec.cs
```diff
@@ -5,29 +5,29 @@
...
 using Newtonsoft.Json.Linq;
 
+#nullable enable
+
 namespace Microsoft.Docs.Build
 {
     internal class ExternalXrefSpec : IXrefSpec
     {
-        public string Uid { get; set; }
+        public string Uid { get; set; } = "";
```
live1206: what is the default value used for? and not in `InternalXrefSpec`

yufeih: in case a item in `.xrefmap.json` does not have a `uid` property

live1206: how could that happen? user define a uid with null or something wrong with the xrefmap output.
If it is a valid case, shall we add `""` to `InternalXrefSpec` as well, and add a test to cover this case

yufeih: It is not possible in our scenario, but a user could handwrite an `.xrefmap.json` with `uid` or `xref` set or `null` or `undefined`, that'll cause `NullReferenceException` and this default value here prevents it from happening.

`InternalXrefSpec` does not have this problem because we are not creating an instance of `InternalXrefSpec` for empty uids.

PR #5596, opened 2020-03-05
ensure contribution branch for included in build CRR

live1206:
https://ceapex.visualstudio.com/Engineering/_workitems/edit/184255

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5596)

### DIFF
--- a/src/docfx/restore/PackageResolver.cs
+++ b/src/docfx/restore/PackageResolver.cs
@@ -46,22 +46,20 @@ namespace Microsoft.Docs.Build
             return _packagePath.GetOrAdd(package, key => new Lazy<string>(() => ResolvePackageCore(key, options))).Value;
         }
 
-        public void DownloadPackage(PackagePath path, PackageFetchOptions options, Config? config = null)
+        public void DownloadPackage(PackagePath path, PackageFetchOptions options)
         {
             try
             {
                 if (path.Type == PackageType.Git)
                 {
                     var repoPath = DownloadGitRepository(path.Url, path.Branch, options.HasFlag(PackageFetchOptions.DepthOne));
-                    if (config is null)
-                        return;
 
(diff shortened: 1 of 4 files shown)


src/docfx/restore/Restore.cs
```diff
@@ -91,7 +91,18 @@ private static void RestorePackages(string docsetPath, Config config, string? lo
             using var packageResolver = new PackageResolver(docsetPath, config, fetchOptions);
             ParallelUtility.ForEach(
                 GetPackages(config, locale, repository).Distinct(),
-                item => packageResolver.DownloadPackage(item.package, item.flags),
+                item =>
```
yufeih: Is it possible to put these code inside `PackageResolver` instead of handling it here? This way the interface stays the same as before. I'd like to keep the entry code cleaner so it's easier to see the structure. `EnsureLocalizationContributionBranch` can possibly be extracted to a `LocUtility` though to share.

live1206: sounds good to me

src/docfx/restore/PackageResolver.cs
```diff
@@ -46,15 +46,23 @@ public string ResolvePackage(PackagePath package, PackageFetchOptions options)
             return _packagePath.GetOrAdd(package, key => new Lazy<string>(() => ResolvePackageCore(key, options))).Value;
         }
 
-        public void DownloadPackage(PackagePath path, PackageFetchOptions options)
+        public void DownloadPackage(PackagePath path, PackageFetchOptions options, Config? config = null)
```
yufeih: _config?

PR #5601, opened 2020-03-07
Add retry logic for fetch method

hejingkan2005:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5601)


hejingkan2005:
- Intermittent IO exception/operation exception when fetching allowlists
System.Threading.Tasks.TaskCanceledException: The operation was canceled.
     ---> System.IO.IOException: Unable to read data from the transport connection: The I/O operation has been aborted because of either a thread exit or an application request..
     ---> System.Net.Sockets.SocketException (995): The I/O operation has been aborted because of either a thread exit or an application request.

hejingkan2005:
> There is already retry [here](https://github.com/dotnet/docfx/blob/v3/src/docfx/restore/FileResolver.cs#L157)

Thanks Yufei. But I don't find the retry logic in build raw log. I think the issue is that the exception was in catched [here](https://github.com/dotnet/docfx/blob/4358288a1dd0082321399dae4f753e4885308bef/src/docfx/config/ops/OpsConfigAdapter.cs#L173). 

Do you think we should re-throw the exception to make sure retry logic would work?

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -202,6 +186,30 @@ private async Task<string> GetMetadataSchema(Uri url)
             };
         }
 
+        private async Task<string> FetchRetry(string url, IReadOnlyDictionary<string, string> headers = null, bool nullOn404 = false)
```
yufeih: This better be named `FetchValidationRules` as it writes a validation specific error message on failure, to prevent other people from misusing this for other purposes.

```suggestion
        private async Task<string> FetchValidationRules(string url)
```

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -202,6 +186,30 @@ private async Task<string> GetMetadataSchema(Uri url)
...
+            try
+            {
+                using var response = await HttpPolicyExtensions
+                   .HandleTransientHttpError()
+                   .Or<OperationCanceledException>()
+                   .Or<IOException>()
+                   .RetryAsync(3, onRetry: (_, i) => Log.Write($"[{i}] Retrying '{url}'"))
+                   .ExecuteAsync(async () =>
+                   {
+                       var content = await Fetch(url, headers);
```
yufeih: Consider use `_http.SendAsync` directly here and move line 229 - 232 here as that one is validation specific.

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -202,6 +186,30 @@ private async Task<string> GetMetadataSchema(Uri url)
             };
         }
 
+        private async Task<string> FetchRetry(string url, IReadOnlyDictionary<string, string> headers = null, bool nullOn404 = false)
+        {
+            try
+            {
+                using var response = await HttpPolicyExtensions
+                   .HandleTransientHttpError()
+                   .Or<OperationCanceledException>()
```
OsmondJiang: the default http client time is 100 seconds, with 3 times retry, the total waiting time maybe 400 seconds, maybe we need consider to reduce the timeout waiting time. 
from the AI data, most of the API is finished within 1 second, so 5 seconds is enough?
image

src/docfx/config/ops/OpsConfigAdapter.cs
hejingkan2005: Done

src/docfx/config/ops/OpsConfigAdapter.cs
hejingkan2005: Done

src/docfx/config/ops/OpsConfigAdapter.cs
yufeih: This is used for both server and local, so I'll keep the default timeout.

I prefer we improve validation service stability to fix the root problem. (If the validation service is stable enough, there's probably no need to skip build in case of validation error)

src/docfx/config/ops/OpsConfigAdapter.cs
yufeih: I didn't see anything changed here, did you forget to push?

OsmondJiang: me either

src/docfx/config/ops/OpsConfigAdapter.cs
OsmondJiang: You are right, we have completed our new allow list service work, but it takes time to switch to our new service. Let's just add a simple retry logic here to make it works. Transient error can not be avoided anyway.

src/docfx/config/ops/OpsConfigAdapter.cs
hejingkan2005: There was an issue with my desktop to push code, and now it is fixed. Sorry for confusion.

src/docfx/config/ops/OpsConfigAdapter.cs
hejingkan2005: I don't change the default timeout for time being, the network connectivity timeout should not  happen very frequent, and it occurs around 8 times in the past week. So I think it is still acceptable to wait Maximum 400 seconds.

PR #5608, opened 2020-03-10
publish to another 2 nuget feeds

superyyrrzz:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5608)

### DIFF
--- a/tools/Deployment/config_gulp.json
+++ b/tools/Deployment/config_gulp.json
@@ -22,8 +22,10 @@
     "masterUrl": "https://www.myget.org/F/docfx/api/v2/package"
   },
   "azdevops": {
-    "devUrl": "https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-build-v2-ppe/nuget/v3/index.json",
-    "masterUrl": "https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-build-v2-prod/nuget/v3/index.json"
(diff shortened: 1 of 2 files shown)


tools/Deployment/config_gulp.json
```diff
@@ -22,8 +22,10 @@
     "masterUrl": "https://www.myget.org/F/docfx/api/v2/package"
   },
   "azdevops": {
-    "devUrl": "https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-build-v2-ppe/nuget/v3/index.json",
-    "masterUrl": "https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-build-v2-prod/nuget/v3/index.json"
+    "perfUrl": "https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-build-v2-perf/nuget/v3/index.json",
```
sharanya-rao: @fenxu @superyyrrzz 
Shall we use v2 feed Urls ?
Sample: https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-build-v2-ppe/nuget/v2

superyyrrzz: I think we should use v3 as it is latest API version and can work.

PR #5609, opened 2020-03-10
Flatten metadata for .publish.json

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5609)

### DIFF
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -574,7 +574,7 @@ inputs:
     }
   _themes/ContentTemplate/TestPage.html.primary.js: |
     exports.transform = function (model) {
-      return model.prop;
+      return model;
     }
 outputs:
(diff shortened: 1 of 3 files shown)


live1206:
The "metadata" layer will be used inside the [template](https://dev.azure.com/ceapex/Engineering/_git/docs-ui?path=%2Fpackages%2Fdocfx-templates%2Fchrome.common.js&version=GBdevelop&line=191&lineEnd=192&lineStartColumn=1&lineEndColumn=1&lineStyle=plain), and it will be removed afterwards.

So the metadata will be flat inside `.publish.json` as well, nothing needs to be changed in docfx v3 for now.

@fenxu @yufeih @herohua

PR #5617, opened 2020-03-13
convert unique within docset metadata rules

OsmondJiang:
convert new metadata rule: unique within docset

Feature: https://ceapex.visualstudio.com/Engineering/_workitems/edit/166621

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5617)

### DIFF
--- a/src/docfx/config/ops/OpsMetadataRuleConverter.cs
+++ b/src/docfx/config/ops/OpsMetadataRuleConverter.cs
@@ -23,7 +23,7 @@ namespace Microsoft.Docs.Build
 
             var schema = new
             {
-                uniquewithindocset = new List<string>(),
+                docsetUnique = new List<string>(),
                 properties = new Dictionary<string, dynamic>(),
                 strictRequired = new List<string>(),
                 dependencies = new Dictionary<string, List<string>>(),
@@ -66,7 +66,7 @@ namespace Microsoft.Docs.Build
 
                 if (rulesInfo.ContainsKey("Uniqueness"))
                 {
-                    schema.uniquewithindocset.Add(attribute);
+                    schema.docsetUnique.Add(attribute);
                 }
 
                 if (rulesInfo.ContainsKey("Required"))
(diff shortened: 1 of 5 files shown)


src/docfx/config/ops/OpsMetadataRuleConverter.cs
```diff
@@ -117,6 +123,7 @@ private static bool TryGetAttributeAdditionalErrors(Dictionary<string, OpsMetada
                 { "Date", new string[] { "date-format-invalid", "date-out-of-range" } },
                 { "MicrosoftAlias", new string[] { "ms-alias-invalid" } },
                 { "Deprecated", new string[] { "attribute-deprecated" } },
+                { "Uniqueness", new string[] { "attribute-duplicated-within-docset" } },
```
yufeih: Given the error code in feature spec is `duplicate-h1`, `duplicate-title`, how about calling it `duplicate-attribute`? 
```suggestion
                { "Uniqueness", new string[] { "duplicate-attribute" } },
```

src/docfx/config/ops/OpsMetadataRuleConverter.cs
```diff
@@ -23,6 +23,7 @@ public static string GenerateJsonSchema(string rulesContent, string allowlistsCo
 
             var schema = new
             {
+                uniquewithindocset = new List<string>(),
```
yufeih: JsonSchema uses camelCase convention. 

Consider `docsetUnique`? This feels more aligned if there are other properties like `globalUnique`.

```suggestion
                docsetUnique = new List<string>(),
```

src/docfx/config/ops/OpsMetadataRuleConverter.cs
OsmondJiang: just follow the existing error code defined in docfx, like attribute-deprecated, I am OK with "duplicate-attribute"

src/docfx/config/ops/OpsMetadataRuleConverter.cs
OsmondJiang: if we already have globalUnique, "docsetUnique" is surely a good idea.

src/docfx/config/ops/OpsMetadataRuleConverter.cs
yufeih: It is called `attribute-deprecated` because the actual error code is like `title-deprecated` or `msalias-deprecated`, see the pattern 🤣

OsmondJiang: the pattern is no pattern here 🤣🤣🤣

PR #5618, opened 2020-03-13
Consolidate repository provider

yufeih:
Move repository outside docset

#4259 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5618)


src/docfx/build/document/RepositoryProvider.cs
```diff
@@ -27,70 +29,93 @@ internal class RepositoryProvider
...
         }
 
+        public (Repository? repository, PathString? pathToRepository) GetRepository(FilePath path)
+        {
+            return path.Origin switch
+            {
+                FileOrigin.Default => GetRepository(Path.Combine(_docsetPath, path.Path)),
+                FileOrigin.Fallback when _localizationProvider != null
+                    => (_localizationProvider.GetFallbackRepositoryWithDocsetEntry().fallbackRepository,
+                        _docsetPathToDefaultRepository is null ? null : _docsetPathToDefaultRepository + path.Path),
```
live1206: can just be `_docsetPathToDefaultRepository`

src/docfx/build/document/RepositoryProvider.cs
```diff
@@ -27,70 +29,93 @@ internal class RepositoryProvider
...
 
-            throw new InvalidOperationException();
+            return (repository, new PathString(Path.GetRelativePath(repository.Path, fullPath)));
         }
 
-        private (string path, Repository?) GetTemplateRepository()
+        private Repository? GetRepositoryCore(string directory)
         {
-            if (_config is null || _packageResolver is null)
+            var repository = GitUtility.FindRepository(directory);
```
live1206: consider rename `repository` to `repoPath`

src/docfx/build/document/RepositoryProvider.cs
yufeih: This is to get the relative path of a _file_ to repository, used to calculate GitHub edit URL and such, it is different from `GetRepositoryWithDocsetEntry`

src/docfx/build/document/RepositoryProvider.cs
yufeih: make sense

src/docfx/build/document/RepositoryProvider.cs
live1206: I mean `_docsetPathToDefaultRepository is null ? null : _docsetPathToDefaultRepository` can just be `_docsetPathToDefaultRepository`, it is nullable anyway.

yufeih: Now I get it :)

live1206: :)

PR #5621, opened 2020-03-16
Remove PublishModelBuilder.ExcludeFromOutput

yufeih:
Make `PublishModelBuilder` read file with errors from `ErrorLog`, writers to `ErrorLog` don't need to worry about excluding the file from publish model.

Makes it easier for #5617 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5621)

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -914,6 +914,7 @@ inputs:
     - name: Invalid TocHref Reference
       tocHref: index.md
 outputs:
+  docs/toc.json:
   .errors.log: |
     ["warning","file-not-found","Invalid file link: 'index.md'.","docs/f1/TOC.yml",2,9]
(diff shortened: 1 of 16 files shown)


docs/specs/toc.yml
```diff
@@ -914,6 +914,7 @@ inputs:
     - name: Invalid TocHref Reference
       tocHref: index.md
 outputs:
+  docs/toc.json:
```
yufeih: This diff is a quirk. Before the change, the error of included file is marked on the includer [here](https://github.com/dotnet/docfx/pull/5621/files#diff-9a6c1c1d834540f46bd38fdef1f759adL166), now the logic is more consistent: exclude files that originate the error.

live1206: From the test title, the error should not be assigned to its parent, but seems that it did before the change, now it's better

yufeih: yep

PR #5623, opened 2020-03-16
Make build sync

yufeih:
Docfx build is very compute intensive. The async programming model does not provide much benefit here. This PR makes build sync, to eliminate async overhead, and open up room to leverage [ref struct](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/ref#ref-struct-types) for better error collection.

The things in build that are currently async are:

- `GitHubAcessor` is an HTTP call but it is synchronized internally anyway because of API rate limit
- `MicrosoftGraphAccessor` is an HTTP call but it is calling `GetAwaiter().GetResult()` at JSON schema layer anyway.
- `RazorTemplate` is async but it is a legacy component to be retired.

Use `Parallel` to build docsets in true parallel. Before this change, `Task.WhenAll` isn't truly parallel.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5623)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -12,7 +12,7 @@ namespace Microsoft.Docs.Build
 {
     internal static class Build
     {
-        public static async Task<int> Run(string workingDirectory, CommandLineOptions options)
+        public static int Run(string workingDirectory, CommandLineOptions options)
         {
             options.UseCache = true;
             var docsets = ConfigLoader.FindDocsets(workingDirectory, options);
@@ -22,11 +22,18 @@ namespace Microsoft.Docs.Build
                 return 1;
             }
 
-            var result = await Task.WhenAll(docsets.Select(docset => BuildDocset(docset.docsetPath, docset.outputPath, options)));
-            return result.All(x => x) ? 0 : 1;
+            var hasError = false;
+            Parallel.ForEach(docsets, docset =>
+            {
+                if (BuildDocset(docset.docsetPath, docset.outputPath, options))
+                {
+                    hasError = true;
+                }
+            });
+            return hasError ? 1 : 0;
         }
 
-        private static async Task<bool> BuildDocset(string docsetPath, string? outputPath, CommandLineOptions options)
+        private static bool BuildDocset(string docsetPath, string? outputPath, CommandLineOptions options)
         {
             List<Error> errors;
             Config? config = null;
@@ -57,7 +64,8 @@ namespace Microsoft.Docs.Build
 
                 // run build based on docsets
                 outputPath ??= Path.Combine(docsetPath, config.OutputPath);
-                await Run(config, docset, fallbackDocset, options, errorLog, outputPath, input, repositoryProvider, localizationProvider, packageResolver);
+                Run(config, docset, fallbackDocset, options, errorLog, outputPath, input, repositoryProvider, localizationProvider, packageResolver);
+                return true;
             }
             catch (Exception ex) when (DocfxException.IsDocfxException(ex, out var dex))
             {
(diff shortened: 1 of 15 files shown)


src/docfx/build/Build.cs
```diff
@@ -22,11 +22,18 @@ public static async Task<int> Run(string workingDirectory, CommandLineOptions op
...
             }
 
-            var result = await Task.WhenAll(docsets.Select(docset => BuildDocset(docset.docsetPath, docset.outputPath, options)));
-            return result.All(x => x) ? 0 : 1;
+            var hasError = false;
+            Parallel.ForEach(docsets, docset =>
+            {
+                if (BuildDocset(docset.docsetPath, docset.outputPath, options))
+                {
+                    hasError = true;
```
live1206: Since this is shared between threads, would there be race condition?
But we only care about the `true` case, overwrite would not happen, it should be fine...

src/docfx/build/page/BuildPage.cs
```diff
@@ -296,7 +294,8 @@ private static async Task<(List<Error> errors, JObject model)> LoadJson(Context
                 var (deserializeErrors, landingData) = JsonUtility.ToObject<LandingData>(pageModel);
                 errors.AddRange(deserializeErrors);
 
-                var htmlDom = HtmlUtility.LoadHtml(await RazorTemplate.Render(file.Mime, landingData)).PostMarkup(context.Config.DryRun);
+                var razorHtml = RazorTemplate.Render(file.Mime, landingData).GetAwaiter().GetResult();
```
live1206: seems like `task.GetAwaiter().GetResult()` and `task.Wait()` are essentially the same? Both calling `task.InternalWait`
Check the [call](https://github.com/dotnet/runtime/blob/bedc8d505aa4c42d22914ffb811d7224d871dec7/src/libraries/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs#L2693) within `Task.Wait()`
and the [call](https://github.com/dotnet/runtime/blob/bedc8d505aa4c42d22914ffb811d7224d871dec7/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/TaskAwaiter.cs#L142) within `TaskAwaiter`

src/docfx/lib/cache/JsonDiskCache.cs
```diff
@@ -78,14 +78,14 @@ public async Task<(TError? error, TValue? value)> GetOrAdd(TKey key, Func<TKey,
                 return (default, value);
             }
 
-            var error = await Update(key, valueFactory);
+            var error = Update(key, valueFactory).GetAwaiter().GetResult();
```
live1206: Can we reduce the call of `.GetAwaiter().GetResult()`? like putting this to a higher layer

src/docfx/build/page/BuildPage.cs
yufeih: They will [throw different exceptions](https://stackoverflow.com/questions/17284517/is-task-result-the-same-as-getawaiter-getresult)

live1206: How about [`task.WaitAndUnwrapException()`](https://stackoverflow.com/questions/9343594/how-to-call-asynchronous-method-from-synchronous-method-in-c?answertab=votes#tab-top) then?

yufeih: That is a third party helper function. The established pattern is to use `GetAwaiter().GetResult()`

live1206: right

src/docfx/lib/cache/JsonDiskCache.cs
yufeih: Lifting it outside is gonna increase the # of places that calls `GetAwaiter().GetResult()`, because both `GitHubAccessor` and `MsAliasValidator` calls this method?

PR #5624, opened 2020-03-16
assign code DocumentError to DocumentException

superyyrrzz:
In OPS, it treats `FatalError` as system error rather than user error. This is not true for `DocumentException`, which is actually user error.

I also removed all other `LogError`s when throwing `DocumentException`, except for few places where `LogError` is used to fetch detailed file info from `LoggerScope`.

https://ceapex.visualstudio.com/Engineering/_workitems/edit/188506/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5624)

### DIFF
--- a/src/docfx/Program.cs
+++ b/src/docfx/Program.cs
@@ -92,9 +92,9 @@ namespace Microsoft.DocAsCode
                 command.Exec(context);
                 return 0;
             }
-            catch (Exception e) when (e is DocumentException)
+            catch (Exception e) when (e is DocumentException de)
             {
-                Logger.LogError(e.Message, code: ErrorCodes.Build.DocumentError);
+                Logger.LogError(e.Message, code: de.Code ?? ErrorCodes.Build.DocumentError);
                 return 1;
             }
             catch (Exception e) when (e is DocfxException)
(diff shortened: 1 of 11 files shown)


src/docfx/Program.cs
```diff
@@ -92,7 +92,12 @@ internal static int ExecSubCommand(string[] args)
                 command.Exec(context);
                 return 0;
             }
-            catch (Exception e) when (e is DocumentException || e is DocfxException)
+            catch (Exception e) when (e is DocumentException)
+            {
+                Logger.LogError(e.Message, code: ErrorCodes.Build.DocumentError);
+                return 1;
```
herohua: would this cause duplicate error logs?

superyyrrzz: Yes it will... This PR only makes this line system error, but the code inside is still system error as it has no code assigned... Let me fix it.

superyyrrzz: After second thoughts, I kept the outer `LogError`, but `DocumentException` can carry information like `Code`. I have to update all the code that throws `DcumentException` if I had kept the inner `LogError`...

If this is good, Sharanya's PR can use the same pattern to throw `DocumentException`.

herohua: I see... Would the duplicate error logs be a problem?

PR #5616, opened 2020-03-12
Do not hardcode remote name, prefer origin

yufeih:
Fixes #5547 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5616)

PR #5625, opened 2020-03-16
Calculate repository url based on remote tracking origin

yufeih:
Addresses @KalleOlaviNiemitalo 's comment in #5616

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5625)


src/docfx/lib/git/GitUtility.cs
```diff
@@ -47,49 +47,76 @@ public static bool IsGitRepository(string path)
...
+                var result = (string?)null;
+                var remotes = default(git_strarray);
+                if (git_remote_list(&remotes, pRepo) == 0)
+                {
+                    var pstr = remotes.strings;
+                    for (var i = 0; i < remotes.count; i++)
+                    {
+                        var pName = *pstr++;
+                        var name = Marshal.PtrToStringUTF8(pName);
+                        if (git_remote_lookup(out var pRemote, pRepo, pName) == 0)
```
mingwli: >git_remote_lookup(out var pRemote, pRepo, pName) == 0 [](start = 28, length = 53)

if ((result == null || name == remoteName) && git_remote_lookup(out var pRemote, pRepo, pName) == 0)
// only need to store the first found remote, and then search for `remoteName`

yufeih: I was lazy

PR #5630, opened 2020-03-17
Fix JsonSchemaTransfomer traversal algorithm

yufeih:
Fixes a bunch of problems with the current SDP transform in _JsonSchemaTransformer_:

- Export all `xrefProperties` regardless if it is defined in schema
- Any `contentType` can be referenced by `@uid?displayName=xrefPropertyName` as long as the content transform result is a string
- Cache only nodes marked as `xrefProperties`. Before the change, all nodes are cached.

This PR removes inner functions in _JsonSchemaTransformer_ to clarify the traverse process. Big, recursive inner functions makes bugs hard to spot.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5630)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -16,199 +16,180 @@ namespace Microsoft.Docs.Build
     {
         private readonly JsonSchema _schema;
         private readonly JsonSchemaDefinition _definitions;
-        private readonly ConcurrentDictionary<JToken, (List<Error>, JToken)> _transformedProperties;
+        private readonly ConcurrentDictionary<JToken, (List<Error>, JToken)> _xrefPropertiesCache =
+                     new ConcurrentDictionary<JToken, (List<Error>, JToken)>(ReferenceEqualsComparer.Default);
+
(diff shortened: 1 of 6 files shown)


yufeih:
`DevSandbox` diff is expected due to traversal order change. Before this PR, the traverse is based on `properties`, now it is order by `xrefProperties`

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -16,199 +16,180 @@ internal class JsonSchemaTransformer
     {
         private readonly JsonSchema _schema;
         private readonly JsonSchemaDefinition _definitions;
-        private readonly ConcurrentDictionary<JToken, (List<Error>, JToken)> _transformedProperties;
+        private readonly ConcurrentDictionary<JToken, (List<Error>, JToken)> _xrefPropertiesCache =
```
live1206: does this only contain xref properties?

yufeih: YES!

PR #5640, opened 2020-03-19
DocumentException should not re-throw DocfxException.

sharanya-rao:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5640)

### DIFF
--- a/src/docfx/SubCommands/DocumentBuilderWrapper.cs
+++ b/src/docfx/SubCommands/DocumentBuilderWrapper.cs
@@ -89,7 +89,7 @@ namespace Microsoft.DocAsCode.SubCommands
                 }
                 catch (AggregateException agg) when (agg.InnerException is DocumentException)
                 {
-                    throw new DocumentException(agg.InnerException.Message);
+                    throw;
                 }
                 catch (DocfxException e)
                 {
@@ -97,7 +97,7 @@ namespace Microsoft.DocAsCode.SubCommands
                 }
                 catch (DocumentException e)
                 {
-                    throw new DocumentException(e.Message);
+                    throw;
                 }
                 catch (Exception e)
                 {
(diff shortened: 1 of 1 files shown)


src/docfx/SubCommands/DocumentBuilderWrapper.cs
```diff
@@ -83,14 +83,22 @@ public void BuildDocument()
...
                 {
                     throw new DocfxException(agg.InnerException.Message);
                 }
+                catch (AggregateException agg) when (agg.InnerException is DocumentException)
+                {
+                    throw new DocumentException(agg.InnerException.Message);
+                }
                 catch (DocfxException e)
                 {
                     throw new DocfxException(e.Message);
```
superyyrrzz: directly `throw`, to avoid call stack altered

src/docfx/SubCommands/DocumentBuilderWrapper.cs
```diff
@@ -83,14 +83,22 @@ public void BuildDocument()
...
                 }
+                catch (DocumentException e)
+                {
+                    throw new DocumentException(e.Message);
```
superyyrrzz: `throw`

src/docfx/SubCommands/DocumentBuilderWrapper.cs
```diff
@@ -83,14 +83,22 @@ public void BuildDocument()
...
                     BuildDocument(_config, _manager, _baseDirectory, _outputDirectory, _pluginDirectory, _templateDirectory);
                 }
-                catch (AggregateException agg) when (agg.InnerException is DocfxException || agg.InnerException is DocumentException)
+                catch (AggregateException agg) when (agg.InnerException is DocfxException)
                 {
                     throw new DocfxException(agg.InnerException.Message);
                 }
+                catch (AggregateException agg) when (agg.InnerException is DocumentException)
+                {
+                    throw;
```
herohua: this directly throws the aggregated exception - is this what you desire?

superyyrrzz: Yes, the outer code will look into the details: https://github.com/dotnet/docfx/pull/5634/files#diff-1413287a5e06fba6e15aab6147f23a15R95

PR #5645, opened 2020-03-20
fix for SDP with non-root uid, which is the only one

live1206:
https://dev.azure.com/ceapex/Engineering/_workitems/edit/188567

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5645)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -38,6 +38,12 @@ namespace Microsoft.Docs.Build
             var errors = new List<Error>();
             var xrefSpecs = new List<InternalXrefSpec>();
             LoadXrefSpecsCore(file, context, _schema, token, errors, xrefSpecs);
+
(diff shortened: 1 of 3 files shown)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -38,6 +38,12 @@ public JsonSchemaTransformer(JsonSchema schema)
             var errors = new List<Error>();
             var xrefSpecs = new List<InternalXrefSpec>();
             LoadXrefSpecsCore(file, context, _schema, token, errors, xrefSpecs);
+
+            // if only one uid defined in the file, remove the bookmark from href if any
+            if (xrefSpecs.Count == 1)
```
yufeih: why not move line 83 here now that `href` is settable?

live1206: Because it would be easier to identify if the UID is at the top level during the traversal.

PR #5647, opened 2020-03-20
Restore dependencies in parallel for implicity restore

yufeih:
The current implementation of implicit restore is restore on demand, while this has the benefit of only download content when needed, the dependencies are downloaded sequentially. Running `docfx build` against a new repo takes a lot longer.

This PR changes implicit restore to call `docfx restore` instead, allowing more parallelism and speed up things significantly.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5647)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -57,7 +57,7 @@ namespace Microsoft.Docs.Build
                 var configLoader = new ConfigLoader(repository, errorLog);
                 (errors, config) = configLoader.Load(docsetPath, locale, options);
                 if (errorLog.Write(errors))
-                    return false;
+                    return true;
 
                 using var packageResolver = new PackageResolver(docsetPath, config, options.FetchOptions);
                 var localizationProvider = new LocalizationProvider(packageResolver, config, locale, docsetPath, repository);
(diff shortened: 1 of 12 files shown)


928PJY:
@yufeih do you mean the `docfx restore` will always be called when running `docfx build`? if so the performance of extension will decrease a lot. so I have to use the flag `--no-restore` then?

yufeih:
> @yufeih do you mean the `docfx restore` will always be called when running `docfx build`? if so the performance of extension will decrease a lot. so I have to use the flag `--no-restore` then?

No, there are 3 fetch options: latest, no fetch, prefer cache. `docfx restore` uses `latest` which means pull the latest even if you have a local copy, `docfx build` uses `prefer cache` option, which means no pull (no network) if there is a local copy, `docfx build --no-restore` uses `no fetch`, which means throw an error if there is no local copy.

This PR does no alter this definition, it just speed up `docfx build` by pulling repositores (if there's no local copy) in parallel.

live1206:
>                     return false;

should follow the pattern of [`restore`](https://github.com/dotnet/docfx/blob/7563ff904f2d3b2fac24dc9129f148e4e9c3f7a1/src/docfx/restore/Restore.cs#L59), return true when error

---
Refers to: src/docfx/build/Build.cs:60 in 4833922. [](commit_id = 4833922ad5068b92c55c62ea2ebd6131531a8912, deletion_comment = False)

src/docfx/build/Build.cs
```diff
@@ -25,6 +25,12 @@ public static int Run(string workingDirectory, CommandLineOptions options)
             var hasError = false;
             Parallel.ForEach(docsets, docset =>
             {
+                if (!options.NoRestore && Restore.RestoreDocset(docset.docsetPath, docset.outputPath, options))
+                {
+                    hasError = true;
+                    return;
+                }
```
live1206: should `return 1` to skip build when `restore` hasError?

PR #5651, opened 2020-03-21
Fix statictoc in Edge and IE

Blake-Eryx:
If you build with statictoc then the table of contents won't work properly in Edge and IE. The filter box is broken and when you click on a different page the TOC loses its place (i.e., the books all collapse and don't re-expand).  This is because of this line:

`if (typeof(Storage) !== "undefined")`

This is not a valid check in Edge because even if DOM is disabled, `typeof(Storage) `always returns "function" (even if `sessionStorage` doesn't exist). Hence, the calls to `sessionStorage` throw an exception and the subsequent TOC code doesn't run.

This patch wraps calls to `sessionStorage `in try blocks to prevent the exceptions that break the TOC in Edge/IE.  Works fine in Edget/IE/Chrome/Firefox after applying this patch.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5651)

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -447,7 +447,11 @@ $(function () {
         var val = this.value;
         //Save filter string to local session storage
         if (typeof(Storage) !== "undefined") {
-          sessionStorage.filterString = val;
+          try {
+            sessionStorage.filterString = val;
+            }
+          catch(e)
+            {}
         }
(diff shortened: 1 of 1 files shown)


superyyrrzz:
Thank you @Blake-Eryx for contribution!

PR #5656, opened 2020-03-23
integrate with docs.validation extension

OsmondJiang:
[AB#187822](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/187822)

The test will failed caused by the Apex Content Extension handling unknown rules, will work on that project to solve this problem.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5656)

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,5 +4,6 @@
         <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab" />
         <add key="docfx-dev" value="https://www.myget.org/F/docfx-dev/api/v3/index.json" />
         <add key="op-external" value="https://www.myget.org/F/op-external/api/v2" />
+        <add key="doc-validation" value="https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-validation-sandbox%40Local/nuget/v3/index.json" />
     </packageSources>
 </configuration>
--- a/NuGet.config
+++ b/NuGet.config
@@ -1,9 +1,9 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <configuration>
     <packageSources>
         <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab" />
         <add key="docfx-dev" value="https://www.myget.org/F/docfx-dev/api/v3/index.json" />
         <add key="op-external" value="https://www.myget.org/F/op-external/api/v2" />
-        <add key="doc-validation" value="https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-validation-sandbox%40Local/nuget/v3/index.json" />
+        <add key="doc-validation" value="https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-validation-sandbox/nuget/v3/index.json" />
     </packageSources>
 </configuration>
--- a/src/docfx/build/context/Context.cs
+++ b/src/docfx/build/context/Context.cs
@@ -2,8 +2,6 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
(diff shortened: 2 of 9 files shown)


src/docfx/build/context/Context.cs
```diff
@@ -64,6 +66,8 @@ internal sealed class Context : IDisposable
 
         public LocalizationProvider LocalizationProvider { get; }
 
+        public Validator DocsValidator { get; }
```
yufeih: Instead of having logic [here](https://github.com/dotnet/docfx/pull/5656/files#diff-d898f1c3e2ee7e44ae96aa08cccd7a49R132), [here](https://github.com/dotnet/docfx/pull/5656/files#diff-c329d0f4c7d5076b957537c2818b4a03R383) and [here](https://github.com/dotnet/docfx/pull/5656/files#diff-16e9cfd7374b164c5e3c3043be71202eR65), consider creating a wrapper class like `ContentValidator` that house the adaption logic. 
```suggestion
        public ContentValidator ContentValidator { get; }
```

src/docfx/build/page/BuildPage.cs
```diff
@@ -375,5 +379,26 @@ private static string CreateContent(Context context, Document file, JObject page
...
+                        Level = 1,
+                        Content = title,
+
+                        // todo: get title precise line info
+                        SourceInfo = new SourceInfo(file.FilePath, 0, 0),
+                    },
+                };
+
+            var validationContext = new ValidationContext { DocumentType = string.IsNullOrEmpty(file.Mime) ? "conceptual" : file.Mime.Value! };
+            context.DocsValidator.ValidateHeadings(headings, validationContext).GetAwaiter().GetResult();
```
yufeih: consider let validation library _not_ returning `Task`, I thought the design guideline for validation library is these methods are compute only, you are never suppose to do any network IO inside these methods. A non-task returning method signature helps convey that principle.

src/docfx/build/page/BuildPage.cs
```diff
@@ -375,5 +379,26 @@ private static string CreateContent(Context context, Document file, JObject page
...
+
+            var headings = new List<Heading>
+                {
+                    new Heading
+                    {
+                        Level = 1,
+                        Content = title,
+
+                        // todo: get title precise line info
+                        SourceInfo = new SourceInfo(file.FilePath, 0, 0),
```
yufeih:
```suggestion
                        SourceInfo = new SourceInfo(file.FilePath, 1, 1),
```

src/docfx/build/page/BuildPage.cs
yufeih: Should `ValidateHeadings` return a validation results? If so, write the errors here.

src/docfx/build/context/Context.cs
OsmondJiang: good suggestion

src/docfx/build/page/BuildPage.cs
OsmondJiang: Just in case we need do some IO validation, like the current `Microsoft alias validation`.

NuGet.config
```diff
@@ -4,5 +4,6 @@
         <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab" />
         <add key="docfx-dev" value="https://www.myget.org/F/docfx-dev/api/v3/index.json" />
         <add key="op-external" value="https://www.myget.org/F/op-external/api/v2" />
+        <add key="doc-validation" value="https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-validation-sandbox%40Local/nuget/v3/index.json" />
```
yufeih: What is this `%40Local`?

PR #5657, opened 2020-03-23
Downgrade output-path-conflict to warning

live1206:
Right now, `output-path-conflict` is defined as `error`.

Consider this case, in the en-us repository a file has been added to the `.redirection` file, and the original source file renamed to some other name or might just be deleted.
Since there is no guarantee that, LOC repository will always sync with en-us repository. There would be a chance that, LOC repository has the `.redirection` file synced from en-us but the source file of the same name still exists in the LOC repository. 
The build would output `output-path-conflict` error, but LOC team could do nothing about it, except waiting for the sync from en-us to be completed.

So we would like to downgrade `output-path-conflict` to not block LOC build.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5657)


src/docfx/build/page/BuildPage.cs
```diff
@@ -12,6 +12,8 @@ namespace Microsoft.Docs.Build
 {
     internal static class BuildPage
     {
+        private static readonly object s_lock = new object();
```
yufeih: consider let `Output` manage per output path locks, to ensure outputs to the same outputPath is saved sequentially.

live1206: I think `context.PublishModelBuilder.TryAdd` and `context.Output.Write` should be an atom operation here.
Let's say, we have `redirection file` A and `source file` B with the same output.
A and B are building simultaneously, B call `TryAdd` firstly and then `write`, A call `TryAdd` later and should rewrite the previous `write` from B. If `TryAdd` and `write` are different operations, it could happen that the later B rewrite finishes before A write.

yufeih: Right, the current implementation does not achieve this either. I'm thinking about sth. like below where the lock is hidden inside a disposable:

```csharp
using (publishManifestBuilder.TryAdd(publishItem, out var writeOutput))
{
    if (writeOutput)
    {
        // write output
    }
}
```

live1206: The case I mentioned above happened during our UT, it is fixed now. Why the current implementation does not achieve this either?

live1206: The return value of `publishManifestBuilder.TryAdd` is the same as `out var writeOutput`

src/docfx/build/context/Context.cs
```diff
@@ -92,7 +92,7 @@ public Context(string outputPath, ErrorLog errorLog, CommandLineOptions options,
             RedirectionProvider = new RedirectionProvider(docset.DocsetPath, Config.HostName, ErrorLog, BuildScope, repositoryProvider, DocumentProvider, MonikerProvider);
             GitHubAccessor = new GitHubAccessor(Config);
             GitCommitProvider = new GitCommitProvider(repositoryProvider);
-            PublishModelBuilder = new PublishModelBuilder(outputPath, Config, Output, ErrorLog);
+            PublishModelBuilder = new PublishModelBuilder(this, outputPath, Config, Output, ErrorLog);
```
yufeih: Pass the actual dependency over the whole context.
```suggestion
            PublishModelBuilder = new PublishModelBuilder(outputPath, Config, Output, ErrorLog, MonikerProvider);
```

src/docfx/lib/collections/ConcurrentHashSet.cs
```diff
@@ -21,6 +21,8 @@ public ConcurrentHashSet(IEnumerable<T> source)
 
         public bool Contains(T value) => _dictionary.ContainsKey(value);
 
+        public HashSet<T> ToHashSet() => _dictionary.Keys.ToHashSet();
```
yufeih: never used?

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -156,5 +218,57 @@ private bool IsInsideOutputFolder(string path)
...
+            PublishUrlConflicts,
+        }
+
+        private class PublishItemKey : IEquatable<PublishItemKey>
+        {
+            public string[] Monikers { get; } = Array.Empty<string>();
+
+            public string? OutputPath { get; }
+
+            public string SiteUrl { get; } = "";
```
yufeih: Minor, this and Monikers does not need init as they are init in constructor
```suggestion
            public string SiteUrl { get; }
```

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -156,5 +218,57 @@ private bool IsInsideOutputFolder(string path)
...
+
+            public FilePath FilePath { get; }
+
+            public PublishItemKey(PublishItem item, FilePath file)
+            {
+                Debug.Assert(item != null);
```
yufeih: With nullable enabled, can remove this line
```suggestion
                ~~~Debug.Assert(item != null);~~~
```

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -156,5 +218,57 @@ private bool IsInsideOutputFolder(string path)
...
+            }
+
+            public bool Equals(PublishItemKey? other)
+            {
+                if (other is null)
+                {
+                    return false;
+                }
+
+                return OutputPathConflicts(OutputPath, other.OutputPath) || PublishUrlConflicts(SiteUrl, other.SiteUrl, Monikers, other.Monikers);
```
yufeih: consider call this `OutputPathEquals || PublishUrlEquals`?

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -140,12 +105,109 @@ orderby monikerGroup
...
+            => (file, item, ConflictingType.None, new ConcurrentDictionary<FilePath, IReadOnlyList<string>>()));
+        }
+
+        private Lazy<(FilePath, PublishItem, ConflictingType, ConcurrentDictionary<FilePath, IReadOnlyList<string>>)> Update(FilePath file, PublishItem item, Action? writeOutput, Lazy<(FilePath, PublishItem, ConflictingType, ConcurrentDictionary<FilePath, IReadOnlyList<string>>)> existingValue, PublishItemKey newKey)
+        {
+            var (existingFile, existingPublishItem, conflictingTYpe, conflictingFiles) = existingValue.Value;
+            conflictingFiles.TryAdd(existingFile, existingPublishItem.Monikers);
+            conflictingFiles.TryAdd(file, item.Monikers);
+
+            if (PublishItemKey.OutputPathConflicts(newKey.OutputPath, existingPublishItem.Path))
```
yufeih: consider let `PublishItem` implement `IComparer<PublishItem>` interface

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -1,28 +1,31 @@
...
+        private readonly Context _context;
         private readonly string _outputPath;
         private readonly Config _config;
         private readonly Output _output;
         private readonly ErrorLog _errorLog;
-        private readonly ConcurrentDictionary<string, ConcurrentBag<FilePath>> _outputPathConflicts = new ConcurrentDictionary<string, ConcurrentBag<FilePath>>(PathUtility.PathComparer);
-        private readonly ConcurrentDictionary<string, ConcurrentDictionary<FilePath, IReadOnlyList<string>>> _filesBySiteUrl = new ConcurrentDictionary<string, ConcurrentDictionary<FilePath, IReadOnlyList<string>>>(PathUtility.PathComparer);
-        private readonly ConcurrentDictionary<string, FilePath> _filesByOutputPath = new ConcurrentDictionary<string, FilePath>(PathUtility.PathComparer);
-        private readonly ConcurrentDictionary<FilePath, PublishItem> _publishItems = new ConcurrentDictionary<FilePath, PublishItem>();
+        private readonly ConcurrentDictionary<PublishItemKey, Lazy<(FilePath, PublishItem, ConflictingType, ConcurrentDictionary<FilePath, IReadOnlyList<string>>)>> _outputConflicts
```
yufeih: The key could still be `PublishItem`, as long as the class is `PublishItemDuplicateComparer : IEqualityComparer<PublishItem>`

```suggestion
        private readonly ConcurrentDictionary<PublishItemKey, Lazy<(FilePath, PublishItem, ConflictingType, ConcurrentDictionary<FilePath, IReadOnlyList<string>>)>> _outputConflicts
```

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -156,5 +218,57 @@ private bool IsInsideOutputFolder(string path)
...
         }
+
+        private enum ConflictingType
+        {
+            None,
+            OutputPathConflicts,
+            PublishUrlConflicts,
+        }
+
+        private class PublishItemKey : IEquatable<PublishItemKey>
```
yufeih: consider extract this into a separate file

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -140,12 +105,109 @@ orderby monikerGroup
...
+                    return new Lazy<(FilePath, PublishItem, ConflictingType, ConcurrentDictionary<FilePath, IReadOnlyList<string>>)>(
+                        () => (file, item, ConflictingType.PublishUrlConflicts, conflictingFiles));
+                }
+                return new Lazy<(FilePath, PublishItem, ConflictingType, ConcurrentDictionary<FilePath, IReadOnlyList<string>>)>(
+                    () => (existingFile, existingPublishItem, ConflictingType.PublishUrlConflicts, conflictingFiles));
+            }
+            return existingValue;
+        }
+
+        private int CompareMonikers(string[] monikers, string[] otherMonikers)
```
yufeih: How about comparing `PublishItem.MonikerGroup` property instead?

src/docfx/publish/PublishModelBuilder.cs
live1206: would like to keep `output-path-conflict` and `publish-url-conflict`

PR #5659, opened 2020-03-24
trim the href before resolve link

928PJY:
[AB#148782](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/148782)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5659)

### DIFF
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -120,12 +120,15 @@ outputs:
 # Resolve <a> and <img> tags embedded inside markdown
 inputs:
   docfx.yml:
-  docs/a.md: Link to <a href="b.md" target="_blank"><em>b</em></a>
+  docs/a.md: |
+    <a href="">Empty string</a>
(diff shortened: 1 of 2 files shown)


928PJY:
like a `HrefString`?

---
In reply to: [380006526](https://github.com/dotnet/docfx/pull/5659#issuecomment-380006526) [](ancestors = 380006526)

yufeih:
> like a `HrefString`?
> 
> In reply to: [380006526](https://github.com/dotnet/docfx/pull/5659#issuecomment-380006526) [](ancestors = 380006526)

yeah, something like that, but it could be a different change after this PR.

I'm think about using `PathString` to represent links as well, they are after all sharing a lot of similarities. So far, this PR LGTM.

PR #5661, opened 2020-03-24
Preprocess xref tag in mustache template

mingwli:
[AB#102271](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/102271)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5661)


src/docfx/template/MustacheTemplate.cs
```diff
@@ -72,7 +71,13 @@ private class PartialLoader : IStubbleLoader
 
             public string Load(string name) => _cache.GetOrAdd(
                 name,
-                key => new Lazy<string>(() => File.ReadAllText(Path.Combine(_dir, name + ".tmpl.partial")))).Value;
+                key => new Lazy<string>(() =>
+                {
+                    var fileName = Path.Combine(_dir, name);
+                    if (File.Exists(fileName))
+                        return File.ReadAllText(fileName);
+                    return File.ReadAllText(Path.Combine(_dir, name + ".tmpl.partial"));
```
mingwli: NetMember has full-name partial path defination
```
<td>
  <xref uid="{{ uid }}" template="partials/dotnet/xref-name.tmpl" />
</td>
```

src/docfx/template/XrefTagParser.cs
```diff
@@ -0,0 +1,58 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Text.RegularExpressions;
+
+namespace Microsoft.Docs.Build
+{
+    internal static class XrefTagParser
```
yufeih: Consider prefix it with `Mustache` to indicate this is not a general xref tag parser

```suggestion
    internal static class MustacheXrefTagParser
```

test/docfx.Test/template/XrefTagParserTest.cs
```diff
@@ -0,0 +1,66 @@
...
+            "{{/href}}" +
+            "{{^href}}" +
+            "  <span> {{.}} </span>" +
+            "{{/href}}"
+            )]
+        public void ProcessXrefTag(string template, string expected)
+        {
+            template = template.Replace('\'', '"');
+            Assert.Equal(
+                Regex.Replace(expected.Replace('\'', '"'), @"(?<=\}) +| +(?=\{)", ""),
```
yufeih: is there a simpler way of doing the validation without regex? like striping out all newlines and spaces. Guessing the meaning of `(?<=\}) +| +(?=\{)` is no fun :(

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -237,7 +237,7 @@ private JToken LoadXrefProperty(Document file, Context context, SourceInfo<strin
                         return (errors, specObj);
                     }
 
-                    return (errors, JValue.CreateNull());
+                    return (errors, JValue.CreateString(content));
```
yufeih:
```suggestion
                    return (errors, value);
```

src/docfx/template/XrefTagParser.cs
```diff
@@ -0,0 +1,58 @@
...
+                else if (string.Equals(attribute.Name, "template", StringComparison.OrdinalIgnoreCase))
+                    partialName = attribute.Value.Trim(s_trimChars);
+            }
+
+            var resolvedTag = partialName == null
+                ? "<a href=\"{{href}}\"> {{uid}} </a>"
+                : "{{> " + partialName + "}}";
+            var resultTemplate = XrefTagTemplate.Replace("@resolvedTag", resolvedTag);
+
+            if (uidName == ".")
```
yufeih: what if `uidName` is null?

src/docfx/template/MustacheTemplate.cs
yufeih: Do you prefer to change `docs-ui` or here?

mingwli: I'd like to, this will make docs-ui partial style more consistent.
But this also requires v2  to support it, which may need some tests first.

src/docfx/template/XrefTagParser.cs
mingwli: Nice catch !
Currently there's no such case, we can simply throw here to avoid it.

There's however one exception for `ProfileList`
```html
<xref href="{{uid}}" title="{{name}}">
  {{#image}}
    <img class="profile-list-image" src="{{src}}" alt="">
  {{/image}}
</xref>
```
Current regex won't match it, so it's safe here. We need to add a TODO for `ProfileList` thougth.

src/docfx/template/MustacheTemplate.cs
yufeih: So then we'll make then change later as we onboard .NET?

src/docfx/template/MustacheXrefTagParser.cs
```diff
@@ -0,0 +1,64 @@
...
+                ? "<a href=\"{{href}}\"> {{uid}} </a>"
+                : "{{> " + partialName + "}}";
+            var resultTemplate = XrefTagTemplate.Replace("@resolvedTag", resolvedTag);
+
+            if (uidName == null)
+            {
+                // TODO: uid may fallback to href in ProfileList
+                throw new NotSupportedException($"<xref> defined without uid in templte: {templateFileName}");
+            }
+            else if (uidName == ".")
```
yufeih: I thought it was simply:

```suggestion
            if (string.IsNullOrEmpty(uidName) || uidName == ".")
```

Throw here is probably not a good approach here as templates are some how user content?

src/docfx/template/MustacheTemplate.cs
mingwli: @yufeih Talked with @superyyrrzz , This will break v2 xref process.
Also perform a test without '.tmpl' ext and rename the .tmpl file.
[Report](https://opbuildstoragesandbox2.blob.core.windows.net/report/2020%5C3%5C24%5Cdde987a0-8603-7cdc-14a2-f3823991e8d9%5CCommit%5C202003240833006394-template-xref-partial%5Cworkflow_report.html?sv=2016-05-31&sr=b&sig=ZTCH0QPBZt16HsxdsPjYWkugF0HN3TkpjaZdfyUTfdY%3D&st=2020-03-24T08%3A30%3A41Z&se=2020-04-24T08%3A35%3A41Z&sp=r)

[template](https://github.com/microsoft/templates.docs.msft/commits/mingwli/xref-partial)
[test-repo](https://github.com/mingwli/test/blob/template-xref-partial/.openpublishing.publish.config.json#L69)
```diff
<tr>
    <td>
-        <a data-linktype="relative-path" href="catlibrary.cat-2.catchjerry#catlibrary_cat_2_catchjerry">catchjerry()</a>
+        <a class="xref" href="catlibrary.cat-2.catchjerry#CatLibrary_Cat_2_CatchJerry" data-linktype="relative-path">CatchJerry()</a>
    </td>
    <td>
        <p>this one can't.</p>
    </td>
</tr>
```

src/docfx/template/MustacheXrefTagParser.cs
mingwli: What about make it default `uid`?

yufeih: sure

mingwli:
> ⚠[DevSandbox](https://github.com/OPS-E2E-PPE/DevSandbox)(00:00:15.5578313, 422 more diff)

Expected change due to unresolved xref remains what user writes

PR #5663, opened 2020-03-24
Add robots metadata for customized 404 page

928PJY:
[AB#190733](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/190733)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5663)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -45,12 +45,6 @@ namespace Microsoft.Docs.Build
             errors.AddRange(outputErrors);
             publishItem.ExtensionData = metadata;
 
-            if (IsCustomized404Page(file))
-            {
-                // custom 404 page is not supported
-                errors.Add(Errors.Content.Custom404Page(file));
-            }
-
             if (shouldWriteOutput && !context.Config.DryRun)
             {
                 if (output is string str)
@@ -190,6 +184,7 @@ namespace Microsoft.Docs.Build
             if (IsCustomized404Page(file))
             {
                 systemMetadata.Robots = "NOINDEX, NOFOLLOW";
+                errors.Add(Errors.Content.Custom404Page(file));
             }
 
             if (context.Config.OutputPdf)
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -385,7 +385,7 @@ namespace Microsoft.Docs.Build
 
         private static bool IsCustomized404Page(Document file)
         {
-            return Path.GetFileName(file.FilePath.Path).Equals("404.md", PathUtility.PathComparison);
+            return Path.GetFileNameWithoutExtension(file.FilePath.Path).Equals("404", PathUtility.PathComparison);
         }
     }
 }
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Docs.Build
             errors.AddRange(outputErrors);
             publishItem.ExtensionData = metadata;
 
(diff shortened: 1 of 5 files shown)


src/docfx/build/page/BuildPage.cs
```diff
@@ -187,6 +187,11 @@ private static (List<Error>, SystemMetadata) CreateSystemMetadata(Context contex
             systemMetadata.SearchProduct = context.Config.Product;
             systemMetadata.SearchDocsetName = context.Config.Name;
 
+            if (IsCustomized404Page(file))
+            {
+                systemMetadata.Robots = "NOINDEX, NOFOLLOW";
```
yufeih: Do we have to add the metadata pragmatically? The true page https://github.com/MicrosoftDocs/DocsRoot/blob/a5d336d517b20b65af10d57ce831818767b22adb/DocsCoreContent/404.md already has configured `noindex, nofollow` as metadata.

928PJY: At least, it is feature of v2, and there are some 404 files([bagie/docs](https://bagie.visualstudio.com/DefaultCollection/tools/_git/docs?path=%2FCTAS%2F404.md),  [MicrosoftDocs/msrd-dev-center-pr](https://github.com/MicrosoftDocs/msrd-dev-center-pr/blob/master/articles/404.md)) which doesn't have this metadata

yufeih: I see. How about moving line 51 down here so the check, report, action happens at the same place?

src/docfx/build/page/BuildPage.cs
```diff
@@ -377,5 +382,10 @@ private static string CreateContent(Context context, Document file, JObject page
             ValidateBookmarks(context, file, htmlDom);
             return CreateHtmlContent(context, htmlDom);
         }
+
+        private static bool IsCustomized404Page(Document file)
+        {
+            return Path.GetFileName(file.FilePath.Path).Equals("404.md", PathUtility.PathComparison);
```
yufeih: Keep the original behavior, cause a user could write a 404.yml and published to 404

928PJY: For v2, only the `404.md` is treated as a customized 404 page. do we want to keep the behavior?

yufeih: That was a bug (hole) in v2 🤣 ,

src/docfx/build/page/BuildPage.cs
928PJY: Yeah... I have thought about that, but do you think that is a little bit wired to report this warning in a function called `CreateSystemMetadata`. I can do the move if you think that would be better~

src/docfx/build/page/BuildPage.cs
928PJY: Got you, the `404.md` and `404.yml` will both be consumed as 404 file in DHS, I will do the change

src/docfx/build/page/BuildPage.cs
yufeih: Yes, it is better, `CreateSystemMetadata` already reports tons of warnings.

src/docfx/build/page/BuildPage.cs
```diff
@@ -187,6 +181,12 @@ private static (List<Error>, SystemMetadata) CreateSystemMetadata(Context contex
             systemMetadata.SearchProduct = context.Config.Product;
             systemMetadata.SearchDocsetName = context.Config.Name;
 
+            if (IsCustomized404Page(file))
```
yufeih: Move these before line 151 so make dry run work

PR #5666, opened 2020-03-25
<xref> default use name as display text

mingwli:
[AB#192954](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/192954)

### DIFF
--- a/src/docfx/template/MustacheXrefTagParser.cs
+++ b/src/docfx/template/MustacheXrefTagParser.cs
@@ -41,7 +41,7 @@ namespace Microsoft.Docs.Build
             }
 
             var resolvedTag = partialName == null
-                ? "<a href=\"{{href}}\"> {{uid}} </a>"
+                ? "<a href=\"{{href}}\"> {{name}} {{^name}}{{uid}}{{/name}} </a>"
                 : "{{> " + partialName + "}}";
             var resultTemplate = XrefTagTemplate.Replace("@resolvedTag", resolvedTag);
 
--- a/docs/specs/template.yml
+++ b/docs/specs/template.yml
@@ -260,7 +260,6 @@ inputs:
         name: uid1-name
         href: https://docs.com/uid1
       - uid: uid2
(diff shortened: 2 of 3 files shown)


yufeih:
@OsmondJiang , we had quite a bit of diffs due to validation recently, is it possible to create a stable set of rules for `v3-integration-test` branch?

src/docfx/template/MustacheXrefTagParser.cs
```diff
@@ -41,7 +41,7 @@ private static string ReplaceXrefTag(Match match)
             }
 
             var resolvedTag = partialName == null
-                ? "<a href=\"{{href}}\"> {{uid}} </a>"
+                ? "<a href=\"{{href}}\"> {{name}} {{^name}}{{uid}}{{/name}} </a>"
```
yufeih: instead of changing the template, how about having a `name` property for every xrefspec just like `href`? We have a couple of `GetXrefPropertyValueAsString("name")`, seems to me that `name` is just as special as `href` for xref.

@live1206 , what do you think?

mingwli: @yufeih Cool for me, sounds like a good idea. And can simply used `{{name}}` without a literal fallback.

OsmondJiang:
> @OsmondJiang , we had quite a bit of diffs due to validation recently, is it possible to create a stable set of rules for `v3-integration-test` branch?

I'd like to, but for this case, it's caused by allow list changes, not rule set, allow list are global level, which means it controls all repos and all branches.

Yesterday in PPE environment, the allow list data was broken accidentally, we fixed it today and add some test sto make sure it wouldn't happened again. cc @fkpwolf

src/docfx/template/MustacheXrefTagParser.cs
mingwli: #5668

PR #5680, opened 2020-03-26
Transform xref in HTML block should not auto close the parent tag

928PJY:
[AB#151947](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/151947)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5680)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -155,14 +155,22 @@ namespace Microsoft.Docs.Build
             var doc = new HtmlDocument();
             doc.LoadHtml(html);
 
+            var pos = 0;
+            var result = new StringBuilder(html.Length + 64);
+
             // TODO: remove this column offset hack while we have accurate line info for link in HTML block
             var columnOffset = 0;
-            var replacingNodes = new List<(HtmlNode, HtmlNode)>();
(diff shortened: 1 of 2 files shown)


928PJY:
The fix cannot handle the case `<xref href="a" />`, it will be parsed  to `<xref href="a"></xref>` and lost the original HTML info

src/docfx/lib/HtmlUtility.cs
```diff
@@ -155,16 +155,24 @@ public static string TransformXref(string html, Func<string, bool, int, (string?
...
             var columnOffset = 0;
-            var replacingNodes = new List<(HtmlNode, HtmlNode)>();
             foreach (var node in doc.DocumentNode.Descendants())
             {
                 if (node.Name != "xref")
                 {
                     continue;
                 }
 
+                var valueStartIndex = node.StreamPosition;
```
yufeih: maybe `AngleSharp` can come to rescue

yufeih: @928PJY Seems like `AngleSharp` also does not satisfy our need.

PR #5685, opened 2020-03-27
warning when enum dependency missed

fkpwolf:
for current test case, the enum value is always mandatory.
@OsmondJiang 

[AB#190363](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/190363)


src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -553,6 +553,25 @@ private void PostValidateDocsetUnique(List<(string name, Error)> errors)
                         }
                     }
                 }
+                else
+                {
+                    if (string.IsNullOrEmpty(dependentFieldNameWithIndex))
```
yufeih: may be refactor this chunk of code in to a separate method? This seems dup with ln 538 ~  552

yufeih:
@herohua , would you help add @fkpwolf to `dotnet-docfx-contrib` so his PR can auto trigger build? I don't seem to have team management permission

src/docfx/lib/schema/JsonSchemaValidator.cs
fkpwolf: not exactly same. Refactored code to reflect this. And if extract code to another method, that would result in a long parameters list for the method..

yufeih:
@fkpwolf Let'me know if the diffs are expected and I'm force merge this PR. You don't seem to have the permission to force merge.

OsmondJiang:
@fkpwolf  we need carefully check the differences showed before, otherwise many repositories' PR will get these warnings unexpectedly

OsmondJiang:
image
most of the current allow list allows empty value in the second level

herohua:
> @herohua , would you help add @fkpwolf to `dotnet-docfx-contrib` so his PR can auto trigger build? I don't seem to have team management permission

Sure. @fkpwolf to add you as one of the contributors, would you go to `opensource.microsoft.com` and join `dotnet` organization? Thanks!

OsmondJiang:
@yufeih and @fkpwolf , I've change the error reporting logic:
1. report the correct file source info
2. don't report `invalid-paired-attribute` when
    - first level key is missing
    - second level has {"": null} dependency

please reference the changed test cases for detail.

PR #5686, opened 2020-03-27
Allow "Monikers" for user metadata

live1206:
[AB#194115](https://dev.azure.com/ceapex/Engineering/_workitems/edit/194115)

- `MonikerRange` takes precedence over `Monikers` since the conceptual user is more likely to config it
- warn `duplicate-moniker-config` on `MonikerRange` and `Monikers` both defined
- `Monikers` should not out range `MonikerRange` from `Config` file


src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -191,8 +191,8 @@ private static IEnumerable<string> ProbeRedirectionFiles(string docsetPath, Repo
                     continue;
                 }
 
-                var (error, redirectionSourceMonikers) = _monikerProvider.GetFileLevelMonikers(file);
-                if (error != null)
+                var (errors, redirectionSourceMonikers) = _monikerProvider.GetFileLevelMonikers(file);
+                foreach (var error in errors)
```
yufeih:
```suggestion
                _errorLog.Write(errors);
```

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,49 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
                 {
-                    return (Errors.Versioning.MonikerRangeUndefined(metadata.MonikerRange), configMonikers);
+                    errors.Add(Errors.Versioning.MonikerRangeUndefined(metadata.MonikerRange));
+                    return (errors, configedMonikers);
                 }
 
                 var fileMonikers = _rangeParser.Parse(metadata.MonikerRange);
-                var intersection = configMonikers.Intersect(fileMonikers).ToArray();
+
+                // monikerRagen takes precedence over monikers since it is more likely from user configuration
```
yufeih: I recall we agreed `monikers` should take precedence since it is more precise.

docs/specs/moniker.yml
```diff
@@ -1595,3 +1595,76 @@ outputs:
...
+    {
+      "monikers": [
+        { "moniker_name": "netcore-1.0", "product_name": ".NET Core" },
+        { "moniker_name": "netcore-2.0", "product_name": ".NET Core" },
+      ]
+    }
+outputs:
+  .errors.log: |
+    ["error","moniker-range-out-of-scope","No moniker intersection between docfx.yml/docfx.json and file metadata. Config moniker range '>= netcore-2.0' is 'netcore-2.0', while file moniker range 'netcore-1.0' is 'netcore-1.0'","docs/v1/TOC.md",3,15]
+    ["warning","duplicate-moniker-config","Only 'monikers' or 'monikerRange' should be configed","docs/v1/TOC.md"]
```
yufeih: Consider tweak the message to say that when both are present, monikerRange will be ignored? Also position the error message at `monikerRange`?

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,49 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
+                if (fileMonikers.Length > 0 && metadata.Monikers.Length > 0)
```
yufeih: Prefer using _existance_ over _length_ to check if `monikers` or `monikerRange` is defined. Can change the data model to be nullable.

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,49 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
+
+                var intersection = configedMonikers.Intersect(fileMonikers).ToArray();
 
                 // With non-empty config monikers,
                 // warn if no intersection of config monikers and file monikers
                 if (intersection.Length == 0)
                 {
-                    var error = Errors.Versioning.MonikeRangeOutOfScope(configMonikerRange, configMonikers, metadata.MonikerRange, fileMonikers);
-                    return (error, intersection);
+                    errors.Add(Errors.Versioning.MonikeRangeOutOfScope(configMonikerRange, configedMonikers, fileMonikers, metadata.MonikerRange, metadata.Monikers));
```
yufeih: nit: `fileMonikers` already is `metadata.Monikers`
```suggestion
                    errors.Add(Errors.Versioning.MonikeRangeOutOfScope(configMonikerRange, configedMonikers, fileMonikers, metadata.MonikerRange));
```

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,49 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
-                var intersection = configMonikers.Intersect(fileMonikers).ToArray();
+
+                // monikerRagen takes precedence over monikers since it is more likely from user configuration
+                if (fileMonikers.Length > 0 && metadata.Monikers.Length > 0)
+                {
+                    errors.Add(Errors.Versioning.DuplciateMonikerConfig(new SourceInfo(file, 0, 0)));
+                }
+                if (fileMonikers.Length == 0 && metadata.Monikers.Length > 0)
+                {
+                    fileMonikers = metadata.Monikers.Select(x => x.Value).ToArray();
```
yufeih: Now that `monikers` is an input, need to:
- Check that `monikers` are all valid like `monikerRange` (exist in moniker definition)
- Lowercase `monikers` like `monikerRange` since it is case insensitive

src/docfx/build/moniker/MonikerProvider.cs
live1206: I would prefer `MonikerRange` though as reasoned in description

src/docfx/build/moniker/MonikerProvider.cs
live1206: what is the difference?

src/docfx/build/moniker/MonikerProvider.cs
live1206: Not really since I choose `MonikerRange` to take precedence over `Monikers`

src/docfx/build/moniker/MonikerProvider.cs
live1206: Check the last test from this PR, error will be thrown if `monikers` out range

yufeih: I mean `monikers` can be `[asdf, netcore-1.0]` where `asdf` is an undefined moniker

src/docfx/build/moniker/MonikerProvider.cs
yufeih: A `monikerRange` like `>netcore-2.0 && <netcore-2.0` could evaluate to an empty moniker list, but that doesn't mean a user didn't write it.

src/docfx/build/moniker/MonikerProvider.cs
yufeih: @928PJY @herohua for more inputs

live1206: As I understand, `MonikerRang` and `Monikers` should be equal as long as the moniker definition is correct, why is `monikers` more precise here?
As mentioned in the description, the conceptual users are more likely to use `MonikerRange`, and `Monikers` are only used within combined-versioning, which is from plugin calculation.
@928PJY @herohua

src/docfx/build/moniker/MonikerProvider.cs
live1206: get your point now, that means `Monikers` should be parsed by `MonikerRangeParser`

yufeih: Oh, I'm not saying the should go thought `MonikerRangeParser`, it just needs the same validation and transform (toLower), a simple method is enough to handle this without _parsing_.

live1206: why not just use the parser then? it doesn't contain unnecessary logic

docs/specs/moniker.yml
```diff
@@ -1595,3 +1595,76 @@ outputs:
...
+    # Node
+  monikerDefinition.json: |
+    {
+      "monikers": [
+        { "moniker_name": "netcore-1.0", "product_name": ".NET Core" },
+        { "moniker_name": "netcore-2.0", "product_name": ".NET Core" },
+      ]
+    }
+outputs:
+  4667fedf/docs/toc.json:
```
928PJY: >4667fedf/docs/toc.json: [](start = 2, length = 23)

add the detail moniker list make the expected output more intuitive

live1206: sure

src/docfx/build/moniker/MonikerProvider.cs
live1206: Actually, we do not need to change the Model type to nullable, just check on `metadta.MonikerRange` would be enough.

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,53 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
+                    return (errors, configedMonikers);
                 }
 
                 var fileMonikers = _rangeParser.Parse(metadata.MonikerRange);
-                var intersection = configMonikers.Intersect(fileMonikers).ToArray();
+
+                // monikerRagen takes precedence over monikers since it is more likely from user configuration
+                if (metadata.MonikerRange != null && metadata.Monikers != null)
+                {
+                    var source = metadata.Monikers.First().Source;
```
928PJY: >First() [](start = 51, length = 7)

would this throw exception if there is no item in monikers?

live1206: reverting the nullable change

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,53 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
-                    return (Errors.Versioning.MonikerRangeUndefined(metadata.MonikerRange), configMonikers);
+                    errors.Add(Errors.Versioning.MonikerRangeUndefined(metadata.MonikerRange));
+                    return (errors, configedMonikers);
+                }
+
+                // monikerRagen takes precedence over monikers since it is more likely from user configuration
+                if (!string.IsNullOrEmpty(metadata.MonikerRange) && metadata.Monikers.Length > 0)
+                {
+                    var source = metadata.Monikers.First().Source;
+                    if (source != null)
```
928PJY: >if (source != null) [](start = 20, length = 19)

when would thi be null?

live1206: Just to make nullable check happy

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,53 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
+                    {
+                        errors.Add(Errors.Versioning.DuplciateMonikerConfig(source));
+                    }
                 }
 
                 var fileMonikers = _rangeParser.Parse(metadata.MonikerRange);
-                var intersection = configMonikers.Intersect(fileMonikers).ToArray();
+                if (fileMonikers.Length == 0 && metadata.Monikers.Length > 0)
```
928PJY: >if (fileMonikers.Length == 0 && metadata.Monikers.Length > 0) [](start = 16, length = 61)

Hah, think about this case:
```
---
monikerRange: >1.0 < 1.0
monikers: 1.0
---
the final monikers will be `1.0`, and a `duplciate-moiniker` warning is reported with msg says `we take the monikerRange`?

**Comment by live1206** [comment]
["monikers"](https://github.com/MicrosoftDocs/DevSandbox/blob/f8310cacebcc4939d1a04608ce7d491509838210/devsandbox/reference/System.Web.HttpTaskAsyncHandler.yml#L46) is used in `DevSandbox` already since it contains reference SDP files

File: src/docfx/build/moniker/MonikerProvider.cs (hunk 10 above)
Comment (live1206) [reply src/docfx/build/moniker/MonikerProvider.cs]: Nope, the final monikers will be empty, which is the output of `monikerRange`.
Take a look at the [check](https://github.com/dotnet/docfx/pull/5686/files#diff-afafc49d45fa0ff0eb00d49f86fc7084L91) in this PR

Comment (live1206) [reply src/docfx/build/moniker/MonikerProvider.cs]: The check here should be `if (!string.IsNullOrEmpty(metadata.MonikerRange) && metadata.Monikers.Length > 0)`.

Think about your case here, if `monikers` takes precedence, and `monikers` only contains invalid values. It would be the same

Comment (928PJY) [reply src/docfx/build/moniker/MonikerProvider.cs]: if `monikers` takes precedence, and monikers only contains invalid values, then we take the invalid values and report `undefined` warnings

File: docs/specs/moniker.yml [hunk 11]
```diff
@@ -1595,3 +1595,128 @@ outputs:
...
+    {
+      "monikers": [
+        { "moniker_name": "netcore-1.0", "product_name": ".NET Core" },
+        { "moniker_name": "netcore-2.0", "product_name": ".NET Core" },
+      ]
+    }
+outputs:
+  .errors.log: |
+    ["warning","duplicate-moniker-config","Both 'monikers' and 'monikerRange' are defined, 'monikers' is ignored","docs/v1/TOC.md",2,11]
+    ["error","moniker-range-out-of-scope","No moniker intersection between docfx.yml/docfx.json and file metadata. Config moniker range '>= netcore-2.0' is 'netcore-2.0', while file moniker range '>netcore-1.0 <netcore-1.0' is ","docs/v1/TOC.md",3,15]
```
928PJY: >["error","moniker-range-out-of-scope","No moniker intersection between docfx.yml/docfx.json and file metadata. Con [](start = 4, length = 114)

we can use a in-scope monikers to make this case more focus on this scenario? so as the case `# Warn on moniker config duplication`

src/docfx/build/moniker/MonikerProvider.cs
live1206: `moniker-range-invalid` is an error.
And check the test added in my last commit, error `moniker-range-out-of-scope` is thrown, which is similar

docs/specs/moniker.yml
live1206: Sure, `monikers` can be 2.0

live1206: But `monikers` is never parsed in this case, so `moniker-range-invalid` is not thrown, maybe it also covers the point that it is not in use.
And `moniker-range-out-of-scope` is actually thrown on `monikerRange`

src/docfx/Errors.cs
```diff
@@ -450,7 +453,7 @@ public static Error MonikerRangeInvalid(SourceInfo<string> source, Exception ex)
             /// which used monikerRange in its yaml header or used moniker-zone syntax.
             /// </summary>
             /// Behavior: ✔️ Message: ❌
-            public static Error MonikerRangeUndefined(SourceInfo? source)
+            public static Error MonikerRangeUndefined(SourceInfo source)
```
```suggestion
            public static Error MonikerRangeUndefined(SourceInfo? source)
```

src/docfx/build/metadata/UserMetadata.cs
```diff
@@ -19,6 +20,9 @@ internal class UserMetadata
         [JsonProperty("monikerRange")]
         public SourceInfo<string> MonikerRange { get; private set; } = new SourceInfo<string>("");
 
+        [JsonConverter(typeof(OneOrManyConverter))]
+        public SourceInfo<string>[] Monikers { get; private set; } = Array.Empty<SourceInfo<string>>();
```
yufeih:
```suggestion
        public SourceInfo<SourceInfo<string>[]> Monikers { get; private set; }
```

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,57 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
+                    {
+                        errors.Add(Errors.Versioning.DuplciateMonikerConfig(source));
+                    }
                 }
 
                 var fileMonikers = _rangeParser.Parse(metadata.MonikerRange);
-                var intersection = configMonikers.Intersect(fileMonikers).ToArray();
+                if (string.IsNullOrEmpty(metadata.MonikerRange) && metadata.Monikers.Length > 0)
+                {
+                    fileMonikers = metadata.Monikers.SelectMany(x => _rangeParser.Parse(x)).ToArray();
```
yufeih: consider do validation only

src/docfx/Errors.cs
```diff
@@ -430,6 +430,9 @@ public static Error UidPropertyConflict(string uid, string propertyName, IEnumer
 
         public static class Versioning
         {
+            public static Error DuplciateMonikerConfig(SourceInfo source)
```
herohua: >Duplciate [](start = 32, length = 9)

typo: `Duplicate`

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -81,36 +83,57 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
...
 
-        private (Error? error, string[] monikers) GetFileLevelMonikersCore(FilePath file)
+        private (List<Error> errors, string[] monikers) GetFileLevelMonikersCore(FilePath file)
         {
+            var errors = new List<Error>();
             var (_, metadata) = _metadataProvider.GetMetadata(file);
 
             var configMonikerRange = GetConfigMonikerRange(file);
-            var configMonikers = _rangeParser.Parse(configMonikerRange);
+            var configedMonikers = _rangeParser.Parse(configMonikerRange);
```
herohua: >configed [](start = 16, length = 8)

nit: `configed` -> `configured`

src/docfx/build/metadata/UserMetadata.cs
live1206: If change the type as suggested, `[JsonConverter(typeof(OneOrManyConverter))]` would not work

yufeih: OK, then I'm good with the current approach

src/docfx/build/moniker/MonikerProvider.cs
live1206: Since we do not have a place to easily get `MonikerDefinition`, which is the valid moniker list.
It only exists within [EvaluatorWithMonikersVisitor](https://github.com/dotnet/docfx/blob/cff6a06bbc9c9ae6a8a08f5bc96e591fe0acc1ab/src/docfx/build/moniker/MonikerProvider.cs#L39), then ties to `MonikerRangeParser`.
I would leave this logic as it is since we need to refactor the whole moniker parsing logic quite a bit to split the validation only logic out. I suspect that we would have that much side effect of reusing the `MonikerRangeParser`

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -67,7 +68,8 @@ public SourceInfo<string> GetConfigMonikerRange(FilePath file)
             // User should not define it in moniker zone
             if (fileLevelMonikers.Length == 0)
             {
-                return (Errors.Versioning.MonikerRangeUndefined(rangeString), Array.Empty<string>());
+                var source = rangeString.Source;
```
yufeih: revert

src/docfx/Errors.cs
```diff
@@ -430,6 +430,9 @@ public static Error UidPropertyConflict(string uid, string propertyName, IEnumer
 
         public static class Versioning
         {
+            public static Error DupicateMonikerConfig(SourceInfo? source)
```
herohua: >Dupicate [](start = 32, length = 8)

typo: `Duplicate`

live1206:
The diff of DevSandbox here is expected.
Now, we accept `Monikers` from YAML header, and there is no moniker group configured in docfx.json.
It throws `moniker-range-out-of-scope` error for this article.
We will skip this validation for `reference` docset in later dev task.

yufeih:
`DevSandbox` is technically _NOT_ a reference docset

PR #5687, opened 2020-03-27
Relocate error file path, git urls based on source map

yufeih:
[AB#192500](https://dev.azure.com/ceapex/Engineering/_workitems/edit/192500/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5687)

### DIFF
--- a/docs/specs/output.yml
+++ b/docs/specs/output.yml
@@ -614,3 +614,29 @@ outputs:
           }
       ]
     }
+---
+# Relocate error file path, git urls based on source map
+repos:
+  https://github.com/relocate/source-file:
+    - files:
+        docfx.yml: |
+          sourceMap: a/sourceMap.json
+        a/sourceMap.json: |
+          {
+            "../a.yml": "original.xml",
+            "..\\b.md": "original.xml",
+            "..\\c.md": "original.xml",
+          }
+        a/original.xml:
+        a.yml:
+        b.md: '[](c.md)'
+        c.md:
+outputs:
+  b.json: |
+    {
+      "conceptual": "<p><a href='c'></a></p>",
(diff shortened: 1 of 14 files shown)


src/docfx/lib/path/FilePath.cs
```diff
@@ -37,31 +36,72 @@ internal class FilePath : IEquatable<FilePath>, IComparable<FilePath>
...
-            Debug.Assert(origin != FileOrigin.Dependency);
+        /// <summary>
+        /// Gets the original file path specified in source map.
+        /// </summary>
+        public PathString? OriginalPath { get; }
 
+        /// <summary>
+        /// Creates an unknown file path.
+        /// </summary>
+        public FilePath(string path)
```
live1206: Is it possible to consolidate the constructors? Then I would not need to think about which one to choose for different origin...

yufeih: I'm consolidating it here so that each type goes to a different constructor, the reason is that the required field of each type is different.

docs/specs/output.yml
```diff
@@ -614,3 +614,29 @@ outputs:
           }
       ]
     }
+---
+# Relocate error file path, git urls based on source map
+repos:
+  https://github.com/relocate/source-file:
+    - files:
+        docfx.yml: |
+          sourceMap: a/sourceMap.json
```
live1206: Will the extensibility team be OK with this config file update?

yufeih: yes, I checked with @TianqiZhang and we agreed to use a dedicated file over yaml header for this. The file format is the same as the current mapping file produced from `ecma2yml`.

PR #5688, opened 2020-03-27
Add BuildFileType metric telemetry

928PJY:
[AB#193620](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/193620)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5688)

### DIFF
--- a/src/docfx/lib/log/Telemetry.cs
+++ b/src/docfx/lib/log/Telemetry.cs
@@ -2,8 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading.Tasks;
@@ -24,7 +24,8 @@ namespace Microsoft.Docs.Build
         private static readonly Metric s_operationTimeMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Time", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
         private static readonly Metric s_errorCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"BuildLog", "Code", "Level", "Type", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
         private static readonly Metric s_cacheCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Cache", "Name", "State", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
-        private static readonly Metric s_buildItemCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Count", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
+        private static readonly Metric s_buildCommitCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Count", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
+        private static readonly Metric s_buildFileTypeCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, "BuildFileType", "FileExtension", "DocuemntType", "MimeType", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
 
         private static readonly string s_version = typeof(Telemetry).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "<null>";
         private static readonly string s_os = RuntimeInformation.OSDescription ?? "<null>";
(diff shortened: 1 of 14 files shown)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -65,26 +75,25 @@ public static void TrackCacheMissCount(TelemetryName name)
             s_cacheCountMetric.TrackValue(1, name.ToString(), "miss", s_os, s_version, s_repo, s_branch, s_correlationId);
         }
 
-        public static void TrackBuildItemCount(ContentType contentType)
+        public static void TrackBuildFileTypeCount(Document file)
         {
-            s_buildItemCountMetric.TrackValue(1, $"{TelemetryName.BuildItems}-{contentType}", s_os, s_version, s_repo, s_branch, s_correlationId);
+            var fileExtension = (Path.GetExtension(file.FilePath.Path) ?? "<null>").ToLowerInvariant();
```
yufeih:
```suggestion
            var fileExtension = file.FilePath.Format.ToString();
```

src/docfx/lib/log/Telemetry.cs
```diff
@@ -24,7 +24,8 @@ internal static class Telemetry
         private static readonly Metric s_operationTimeMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Time", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
         private static readonly Metric s_errorCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"BuildLog", "Code", "Level", "Type", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
         private static readonly Metric s_cacheCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Cache", "Name", "State", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
-        private static readonly Metric s_buildItemCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Count", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
+        private static readonly Metric s_buildCommitCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Count", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
```
yufeih: If `os`, `version`, `repo`, `branch` become global properties, can they be removed from dimensions?

928PJY: There is a tricky thing in the telemetry library:
The global property will not be applied to metric if we use the metric container to do the pre-aggregate.

src/docfx/lib/log/Telemetry.cs
928PJY: the format just support `json`, `md` and `yml`, what about resource? like `.jpg`

src/docfx/lib/log/Telemetry.cs
yufeih: hmmm, then seems the original code is more clear?

src/docfx/lib/log/Telemetry.cs
yufeih: yep you are right.
```suggestion
            var fileExtension = Path.GetExtension(file.FilePath.Path)?.ToLowerInvariant() ?? "<null>";
```

src/docfx/lib/log/Telemetry.cs
yufeih:
```suggestion
        private static readonly Metric s_buildCommitCountMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"BuildCommitCount", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
```

928PJY: The global property still works for `TrackEvent` `TrackMetric` and `TrackException`

src/docfx/lib/log/Telemetry.cs
```diff
@@ -34,10 +35,19 @@ internal static class Telemetry
 
         private static string s_correlationId = EnvironmentVariable.CorrelationId ?? Guid.NewGuid().ToString("N");
 
+        static Telemetry()
+        {
+            s_telemetryClient.Context.GlobalProperties["OS"] = RuntimeInformation.OSDescription ?? "<null>";
+            s_telemetryClient.Context.GlobalProperties["Version"] = typeof(Telemetry).Assembly.GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "<null>";
+            s_telemetryClient.Context.GlobalProperties["CorrelationId"] = EnvironmentVariable.CorrelationId ?? Guid.NewGuid().ToString("N");
```
yufeih: Ok, if this to be put into `GlobalProperties`, ensure it is using the same value:
```suggestion
            s_telemetryClient.Context.GlobalProperties["CorrelationId"] = s_correlationId;
```

Consider doing this for other global properties

PR #5689, opened 2020-03-27
Sdp xref href relative resolve

mingwli:
[AB#192954](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/192954)

In SDP, referencing file is not taken considered during xref transform for calculating relative href.
Make external xrefSpec href private settable

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5689)

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Docs.Build
             }
 
             query = queries.AllKeys.Length == 0 ? "" : "?" + string.Join('&', queries);
-            var fileLink = UrlUtility.MergeUrl( xrefSpec.Href, query, fragment);
+            var fileLink = UrlUtility.MergeUrl(xrefSpec.Href, query, fragment);
             _fileLinkMapBuilder.AddFileLink(inclusionRoot.FilePath, inclusionRoot.SiteUrl, fileLink);
 
             resolvedHref = UrlUtility.MergeUrl(resolvedHref, query, fragment);
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Docs.Build
             queries.Remove("displayProperty");
 
             // need to url decode uid from input content
-            var (xrefError, xrefSpec, resolvedHref) = ResolveXrefSpec(new SourceInfo<string>(uid, href.Source), hrefRelativeTo, inclusionRoot);
+            var (xrefError, xrefSpec, resolvedHref) = ResolveXrefSpec(new SourceInfo<string>(uid, href.Source), inclusionRoot ?? hrefRelativeTo, writeFileLink: true);
             if (xrefError != null || xrefSpec is null || resolvedHref == null)
             {
                 return (xrefError, null, "", null);
@@ -84,7 +84,7 @@ namespace Microsoft.Docs.Build
             return (null, resolvedHref, display, xrefSpec?.DeclaringFile);
         }
 
-        public (Error?, IXrefSpec?, string? href) ResolveXrefSpec(SourceInfo<string> uid, Document referencingFile, Document? inclusionRoot = null)
+        public (Error?, IXrefSpec?, string? href) ResolveXrefSpec(SourceInfo<string> uid, Document referencingFile, bool writeFileLink = false)
         {
             var (error, xrefSpec) = Resolve(uid, referencingFile);
             if (xrefSpec == null)
@@ -92,10 +92,7 @@ namespace Microsoft.Docs.Build
 
             var href = RemoveSharingHost(xrefSpec.Href, _config.HostName);
 
-            // NOTE: this should also be relative to root file
-            // Using inclusionRoot to decide whether write link
-            referencingFile = inclusionRoot ?? referencingFile;
-            if (inclusionRoot != null)
+            if (writeFileLink)
                 _fileLinkMapBuilder.AddFileLink(referencingFile.FilePath, referencingFile.SiteUrl, href);
 
             if (xrefSpec?.DeclaringFile != null && referencingFile != null)
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -136,7 +136,7 @@ namespace Microsoft.Docs.Build
                 .OrderBy(xref => xref.Uid).ToArray();
 
             var model = new XrefMapModel { References = references };
-            if (basePath != null)
+            if (basePath != null && references.Length > 0)
             {
                 var properties = new XrefProperties();
                 properties.Tags.Add(basePath);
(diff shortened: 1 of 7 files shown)


src/docfx/build/xref/XrefResolver.cs
```diff
@@ -93,35 +93,44 @@ internal class XrefResolver
...
-
                     // DHS appends branch infomation from cookie cache to URL, which is wrong for UID resolved URL
                     // output xref map with URL appending "?branch=master" for master branch
-                    var (_, _, fragment) = UrlUtility.SplitUrl(xref.Href);
-                    var path = $"https://{_xrefHostName}{xref.DeclaringFile.SiteUrl}";
+                    var (href, _, fragment) = UrlUtility.SplitUrl(xref.Href);
+                    var path = $"https://{_xrefHostName}{href}";
                     var query = repositoryBranch == "master" ? "?branch=master" : "";
-                    xrefSpec.Href = UrlUtility.MergeUrl(path, query, fragment);
+                    href = UrlUtility.MergeUrl(path, query, fragment);
```
yufeih: `MergeUrl` will split target URL for you:
```suggestion
                    var href = UrlUtility.MergeUrl($"https://{_xrefHostName}{xref.Href}", query, fragment);
```

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -93,41 +93,50 @@ internal class XrefResolver
             return (null, resolvedHref, display, xrefSpec?.DeclaringFile);
         }
 
-        public (Error?, ExternalXrefSpec?) ResolveXrefSpec(SourceInfo<string> uid, Document referencingFile)
+        public (Error?, ExternalXrefSpec?, string? href) ResolveXrefSpec(SourceInfo<string> uid, Document referencingFile)
```
yufeih: If this method returns `href` and `IXrefSpec` (without calling `ToExternal`), would line 62 be changed to call this method instead and avoid duplicating URL processing logic?
```suggestion
        public (Error?, IXrefSpec?, string? href) ResolveXrefSpec(SourceInfo<string> uid, Document referencingFile)
```

mingwli: @yufeih  No, that will make break xref spec-level circular reference.
ResolveXrefSpec will visit all xref properties.
One way to avoid duplicate href resolve would be move it inside 
`private (Error?, IXrefSpec?) Resolve(SourceInfo<string> uid, Document referencingFile)`

yufeih: If `ResolveXrefSpec` returns `IXrefSpec` _without calling `ToExternalXrefSpec`_, then the method won't expand all xref properties?

mingwli: @yufeih Without calling ToExternalXrefSpec will lead to xrefProperties lost for internal specs.

mingwli: I've moved shared linkBuilder logic inside Resolve.

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -44,8 +44,8 @@ internal class XrefResolver
             SourceInfo<string> href, Document hrefRelativeTo, Document inclusionRoot)
         {
             var (uid, query, fragment) = UrlUtility.SplitUrl(href);
-            string? moniker = null;
-            string? text = null;
+            var moniker = default(string);
```
live1206: does this really look better :)

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -58,76 +58,58 @@ internal class XrefResolver
...
-
                     // DHS appends branch infomation from cookie cache to URL, which is wrong for UID resolved URL
                     // output xref map with URL appending "?branch=master" for master branch
-                    var (_, _, fragment) = UrlUtility.SplitUrl(xref.Href);
-                    var path = $"https://{_xrefHostName}{xref.DeclaringFile.SiteUrl}";
                     var query = repositoryBranch == "master" ? "?branch=master" : "";
-                    xrefSpec.Href = UrlUtility.MergeUrl(path, query, fragment);
+                    var href = UrlUtility.MergeUrl($"https://{_xrefHostName}{xref.Href}", query);
+
+                    var xrefSpec = xref.ToExternalXrefSpec(href);
```
live1206: If we keep the public setter in `Href` property, we can get rid of this `overwriteHref`?

mingwli: Prefer private setter to avoid unintentional assign. 
I just wasted a long time local debuging a weird behavior. Orzzzzz

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -162,15 +144,26 @@ private string RemoveSharingHost(string url, string hostName)
...
             {
-                return (Errors.Xref.XrefNotFound(uid), null);
+                return (Errors.Xref.XrefNotFound(uid), null, null);
+            }
+
+            var href = UrlUtility.MergeUrl(RemoveSharingHost(xrefSpec.Href, _config.HostName), query);
+
+            // NOTE: this should also be relative to root file
+            var sourceFile = inclusionRoot ?? referencingFile;
+            _fileLinkMapBuilder.AddFileLink(sourceFile.FilePath, sourceFile.SiteUrl, href);
```
live1206: By moving this logic down to `Resolve`, the file link will be added for `ResolveXrefSpec` as well, was there a missing dependency for `xref` node in SDP before?

src/docfx/build/xref/XrefResolver.cs
live1206: IMHO, this looks weird as well... what kind of unintentional assignment did you get?

yufeih: A multi-threading bug caused by shared state mutation. Our best practice is to use _immutable_, _readonly_ object model as much as you can.

src/docfx/build/xref/XrefResolver.cs
yufeih: Maybe some folks like to see `var` alignments. (P.S., I'm one of them, but I didn't tell @mingwli to make this change 🤣 )

src/docfx/build/xref/XrefResolver.cs
live1206: I had another look, maybe this add file link and calculating relative herf should belong to method `ResolveInternalXrefSpec`, we should not have them for external UID, right?

src/docfx/build/xref/XrefResolver.cs
live1206: Agree with the best practice, my point is we are mutating it anyway.

src/docfx/build/xref/XrefResolver.cs
mingwli: It's always added before.
https://github.com/dotnet/docfx/blob/v3/src/docfx/build/xref/XrefResolver.cs#L86

mingwli: @live1206  Move relative/removeHost logic inside resolve internal/external

mingwli:
> ⚠[DevSandbox](https://github.com/OPS-E2E-PPE/DevSandbox)(00:00:18.4223231, 132 more diff)

Expected, xref href is now correctly resolved

Issue #5156, opened 2019-09-29
Improve readability of the default theme


djee-ms:
> _Disclaimer:_ This issue is similar to #3283, but #3283 seems more subjective/vague and geared toward parity with docs.microsoft.com. It is also tagged as v3. This issue is more focused and practical, proposes some immediate change to the existing system, with simple CSS tweaks and nothing else.

I would like to propose some changes to the default theme. As mentioned in #3283, the current layout is small and cramped. It focuses on secondary items, while missing primary ones.

## tl;dr

The changes can be summarized with this commit : microsoft/MixedReality-WebRTC@a1c758fccfbce7850228d1d0d425c2f819109e3d

Here is a simple before / after showing the API reference of that project. There are still some issues, but this is I think much better from a readability point of view.

![image](https://user-images.githubusercontent.com/49064175/65826376-ebd85e00-e2be-11e9-9de5-7bf791a96939.png)

### Formatting of inline code blocks

![image](https://user-images.githubusercontent.com/49064175/65826437-e7607500-e2bf-11e9-8a1c-6606bc0fc8a6.png)

❌ Default code style looks like an invalid HMTL link, as it uses roughly the same red color as broken links in Wikipedia
❌ Similarly, code links (that is, `<a>` inside `<code>`) don't display any difference from regular `<code>` blocks (second `<code>` in screenshot is a link, first is not).

✔ Use of red should probably be avoided unless intending to carry a negative meaning
✔ Links should have a consistent color and hovering effect whether inline code or not.

### Header emphasizing in API

![image](https://user-images.githubusercontent.com/49064175/65826510-63a78800-e2c1-11e9-9b5f-4270af392194.png)

❌ The current focus, based on font size and weight, is put on items not of primary interest for an API reference.
❌ The lack of spacing makes critical information lost in a sea of text : class summary, information related to a particular property (and where that block of information stops and the next one start). Compare with _e.g._ https://www.php.net/manual/en/mysqli.commit.php or even https://docs.python.org/3/library/bisect.html
❌ The border on code blocks makes them a visual separator, splitting the member block into two parts. Compare to _e.g._ https://doc.rust-lang.org/book/ch03-02-data-types.html

✔ Spacing should be added to make the class summary stand out; this is the first and most critical piece of information for the class.
✔ Block of information about members (properties, events, etc.) must also be more clearly delimited with spacing.
✔ Emphasize on member names (bold) to start the block of information and delimit it from the previous one

## Remaining issues

[...]

PR #5693, opened 2020-03-29
Improve readability of default theme

djee-ms:
Change docfx.css to increase the readability of the default theme by
highlighting headers and groups (C#: fields, properties, methods, etc.)
and tuning down focus on less important items.

Bug: #5156

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5693)

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.css
+++ b/src/docfx.website.themes/default/styles/docfx.css
@@ -1007,7 +1007,6 @@ div.embeddedvideo iframe {
 
 /* Major sections, increase visibility **/
 #fields, #properties, #methods, #events {
-    color: #337ac4;
     font-weight: bold;
     margin-top: 2em;
 }
--- a/src/docfx.website.themes/default/styles/docfx.css
+++ b/src/docfx.website.themes/default/styles/docfx.css
@@ -71,6 +71,14 @@ h6 mark {
     margin-left: 5em;
 }
 
+.level0.summary {
+  margin: 2em 0 2em 0;
+}
+
+.level1.summary {
+  margin: 1em 0 1em 0;
+}
+
 span.parametername,
 span.paramref,
 span.typeparamref {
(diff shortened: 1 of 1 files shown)


src/docfx.website.themes/default/styles/docfx.css
```diff
@@ -960,3 +997,17 @@ div.embeddedvideo iframe {
...
+/* Color theme */
+
+/* These are not important, tune down **/
+.decalaration, .fieldValue, .parameters, .returns {
+  color: #a2a2a2;
+}
+
+/* Major sections, increase visibility **/
+#fields, #properties, #methods, #events {
+    color: #337ac4;
```
superyyrrzz: This color is almost same with `#337ab7`, which is used in links in other places. This may mislead the user to try clicking on it. What about keep it black?

djee-ms: Right, because I committed just a subset of the full `main.css` we use, see _e.g._ [this page](https://microsoft.github.io/MixedReality-WebRTC/api/Microsoft.MixedReality.WebRTC.PeerConnection.html) using the proposed style. Here we're missing the header override we use, which takes precedence over this color (which should therefore be removed):

```css
/** Content **/

#_content h1 {
  color: rgb(68, 68, 68);
  font-size: 40px;
  font-weight: 100;
  margin-top: 0;
}

#_content h2 {
  color: rgb(68, 68, 68);
  font-size: 30px;
  font-weight: 100;
  padding-top: 70px;
  margin-top: -50px;
}

#_content h3 {
  color: rgb(68, 68, 68);
  font-size: 22px;
  font-weight: 100;
  padding-top: 70px;
  margin-top: -50px;
}
```

superyyrrzz: Thank you for the explanation. Would you remove this line of color change in this PR?

djee-ms:
Thanks @superyyrrzz for fixing. Sorry I am quite busy at the moment and was doing this on my spare time.

superyyrrzz:
@djee-ms Never mind! I updated it to include this in next Monday's release. Thank you for contribution!

PR #5706, opened 2020-04-01
Use standalone appdata path for each test case

yufeih:
Fixes test framework problems due to shared appdata path: 

- Test won't pick up changes in dependency repo due to the cache in appdata
- Test fails if two cases uses the same repo url or file url due to the shared appdata cache

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5706)

### DIFF
--- a/test/docfx.Test/DocfxTest.cs
+++ b/test/docfx.Test/DocfxTest.cs
@@ -69,12 +69,12 @@ namespace Microsoft.Docs.Build
             CreateDocset(TestData test, DocfxTestSpec spec)
         {
             var testName = $"{Path.GetFileName(test.FilePath)}-{test.Ordinal:D2}-{HashUtility.GetMd5HashShort(test.Content)}";
-            var basePath = Path.GetFullPath(Path.Combine(spec.Temp ? Path.GetTempPath() : "docfx-test", testName));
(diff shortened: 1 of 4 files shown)


test/docfx.Test/DocfxTest.cs
```diff
@@ -64,20 +61,20 @@ public static async Task Run(TestData test, DocfxTestSpec spec)
...
             }
         }
 
-        private static (string docsetPath, string cachePath, string statePath, string outputPath, Dictionary<string, string> repos)
+        private static (string docsetPath, string appDataPath, string outputPath, Dictionary<string, string> repos)
             CreateDocset(TestData test, DocfxTestSpec spec)
         {
             var testName = $"{Path.GetFileName(test.FilePath)}-{test.Ordinal:D2}-{HashUtility.GetMd5HashShort(test.Content)}";
-            var basePath = Path.GetFullPath(Path.Combine(spec.Temp ? Path.GetTempPath() : "docfx-test", testName));
+            var basePath = Path.GetFullPath(Path.Combine(spec.Temp ? Path.GetTempPath() : "docfx-tests", testName));
```
yufeih, line 72: This change is needed to due to cache folder structure change. It allows tests to succeed on server.

PR #5707, opened 2020-04-01
Move TOC deserialization to TOC parser

yufeih:
[192497](https://dev.azure.com/ceapex/Engineering/_workitems/edit/192497/)

Refactor TOC pipeline using a set of small changes. This one moves `LoadTocModel` to `TocParser`, because `splitItems` happens between `LoadTocModel` and `ResolveTocModelItems`.

For `TocLoader`, simplifies error handling using args over return value.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5707)

### DIFF
--- a/src/docfx/build/toc/TableOfContentsLoader.cs
+++ b/src/docfx/build/toc/TableOfContentsLoader.cs
@@ -8,7 +8,6 @@ using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Threading;
-using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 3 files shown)


src/docfx/build/toc/TableOfContentsLoader.cs
```diff
@@ -289,12 +232,12 @@ private IReadOnlyList<string> GetMonikers(TableOfContentsNode currentItem, List<
...
-            List<Document> referencedFiles,
-            List<Document> referencedTocs,
-            SourceInfo<string?> tocHref,
-            List<Error> errors)
+                Document filePath,
+                Document rootPath,
+                List<Document> referencedFiles,
+                List<Document> referencedTocs,
+                SourceInfo<string?> tocHref,
+                List<Error> errors)
```
live1206, line 240: do we want to return `List<Error>` instead of passing it along?

yufeih: return value put a huge burden on the caller, so I tend to prefer putting it as and argument. If we have a stack only `ErrorBuilder`, it will also be an argument.

PR #5708, opened 2020-04-01
[Breaking] Support overwrite buildtime through config

mingwli:
Fixing update-time change in integration test.

Since docs-help-pr is always a PR build, the branch state never gets a chance to be updated.
Hence in commit-ci build, thought branch commits moved forward, but no newer state added.

Expecting update-time to be changed to `2020-01-01`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5708)

### DIFF
--- a/src/docfx/config/EnvironmentVariable.cs
+++ b/src/docfx/config/EnvironmentVariable.cs
@@ -21,6 +21,8 @@ namespace Microsoft.Docs.Build
 
         public static string? CorrelationId => GetValue("DOCFX_CORRELATION_ID");
 
+        public static string? BuildTime => GetValue("DOCFX_BUILD_TIME");
+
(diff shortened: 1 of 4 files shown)


src/docfx/config/EnvironmentVariable.cs
```diff
@@ -21,6 +21,8 @@ public static class EnvironmentVariable
 
         public static string? CorrelationId => GetValue("DOCFX_CORRELATION_ID");
 
+        public static string? BuildTime => GetValue("DOCFX_BUILD_TIME");
```
yufeih, line 24: remove

PR #5709, opened 2020-04-01
Handle TOC build scope in TocMap

yufeih:
[AB#192497](https://dev.azure.com/ceapex/Engineering/_workitems/edit/192497/)

Refactor TOC pipeline using a set of small changes. 

This PR moves TOC build scope to `TocMap`, because `splitItems` will produce more TOCs then what's inside `BuildScope`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5709)

### DIFF
--- a/src/docfx/build/document/DocumentProvider.cs
+++ b/src/docfx/build/document/DocumentProvider.cs
@@ -20,7 +20,6 @@ namespace Microsoft.Docs.Build
         private readonly string _depotName;
         private readonly (PathString, DocumentIdConfig)[] _documentIdRules;
         private readonly (PathString src, PathString dest)[] _routes;
-        private readonly HashSet<string> _configReferences;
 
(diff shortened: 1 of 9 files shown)


src/docfx/build/document/DocumentProvider.cs
```diff
@@ -138,8 +107,7 @@ private bool TryGetDocumentIdConfig(PathString path, out DocumentIdConfig result
 
         private Document GetDocumentCore(FilePath path)
         {
-            var contentType = path.Origin == FileOrigin.Redirection ? ContentType.Redirection : GetContentType(path.Path);
-
+            var contentType = _buildScope.GetContentType(path);
```
mingwli, line 110: >GetContentType [](start = 42, length = 14)

minor suggestion: 
Can this be a static helper function and stay in DocumentProvider?(feels more reasonable to me)
`public static GetContentType(FilePath path)`

PR #5712, opened 2020-04-01
Skip moniker range config validation for reference

live1206:
[AB#195160](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/195160)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5712)

### DIFF
--- a/src/docfx/build/moniker/MonikerProvider.cs
+++ b/src/docfx/build/moniker/MonikerProvider.cs
@@ -125,13 +125,13 @@ namespace Microsoft.Docs.Build
             return (errors, configMonikers);
         }
 
-        private static (Error?, string[]) GetMonikerIntersection(UserMetadata metadata, SourceInfo<string?> configMonikerRange, string[] configMonikers, string[] fileMonikers, bool isReference)
+        private static (Error?, string[]) GetMonikerIntersection(UserMetadata metadata, SourceInfo<string?> configMonikerRange, string[] configMonikers, string[] fileMonikers, bool skipMonikerValidation)
         {
             Error? error = null;
 
             // for reference docset, if config monikers is not defined
             // just use file monikers
-            if (isReference && configMonikerRange.Value is null)
+            if (skipMonikerValidation && configMonikerRange.Value is null)
             {
                 return (error, fileMonikers);
             }
--- a/src/docfx/build/moniker/MonikerProvider.cs
+++ b/src/docfx/build/moniker/MonikerProvider.cs
@@ -65,9 +65,10 @@ namespace Microsoft.Docs.Build
             var configMonikerRange = GetConfigMonikerRange(file);
             var (_, fileLevelMonikers) = GetFileLevelMonikers(file);
 
+            // For conceptual docset,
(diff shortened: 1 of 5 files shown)


mingwli:
> ⚠[docs-help-pr](https://github.com/OPS-E2E-PPE/docs-help-pr)(00:00:18.6557215, 584 more diff)

Will be fine after a commit ci build

src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -129,6 +129,7 @@ private async Task<string> GetBuildConfig(Uri url)
                     $"{MetadataSchemaApi}{metadataServiceQueryParams}",
                 },
                 xref = xrefMaps,
+                isReference = docset.use_template,
```
yufeih, line 132: What if `use_template` translate to sth. like `monikerRange { '/': '*' }'` ? I didn't quite like this `isReference` flag.

live1206: #5714 to support `*` in moniker range parser

live1206: @928PJY had a concern that, if the user misuse `*` a non-versioning repo, then every article will output `monikers` with all the monikers within Docs, which is more than 1000

yufeih: So `*` only be used for checking if it is inside a valid moniker range?

live1206: how to limit the usage?

yufeih: A better name should suffice.

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -108,26 +110,33 @@ public MonikerProvider(Config config, BuildScope buildScope, MetadataProvider me
...
+                var (intersectionError, intersection) = GetMonikerIntersection(metadata, configMonikerRange, configMonikers, fileMonikers, _config.SkipMonikerValidation);
                 errors.AddIfNotNull(intersectionError);
                 return (errors, intersection);
             }
 
             return (errors, configMonikers);
         }
 
-        private static (Error?, string[]) GetMonikerIntersection(UserMetadata metadata, SourceInfo<string?> configMonikerRange, string[] configMonikers, string[] fileMonikers)
+        private static (Error?, string[]) GetMonikerIntersection(UserMetadata metadata, SourceInfo<string?> configMonikerRange, string[] configMonikers, string[] fileMonikers, bool isReference)
```
yufeih, line 128:
```suggestion
        private static (Error?, string[]) GetMonikerIntersection(UserMetadata metadata, SourceInfo<string?> configMonikerRange, string[] configMonikers, string[] fileMonikers, bool skipMonikerValidation)
```

PR #5714, opened 2020-04-01
Consume wildcard in moniker range parser

live1206:
- Consume `*`
- The operator should only be `=`, or without operator, which will be default as `=`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5714)

### DIFF
--- a/test/docfx.Test/lib/MonikerRangeParserTest.cs
+++ b/test/docfx.Test/lib/MonikerRangeParserTest.cs
@@ -110,6 +110,8 @@ namespace Microsoft.Docs.Build
         [InlineData(
             "azuresqldb-current || azure-sqldw-latest",
             "azure-sqldw-latest azuresqldb-current")]
+        [InlineData("*", "netcore-1.0 netcore-2.0 netcore-3.0 dotnet-1.0 dotnet-2.0 dotnet-3.0 azure-sqldw-latest azuresqldb-current")]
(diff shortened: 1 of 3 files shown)


test/docfx.Test/lib/MonikerRangeParserTest.cs
```diff
@@ -126,6 +128,7 @@ public void TestEvaluateMonikerRange(string rangeString, string expectedMonikers
         [InlineData(">netcore<-1.0 ||| >netcore-2.0", "Expect a comparator set, but got `| >netcore-2.0`")]
         [InlineData(">netcore<-1.0 || ||", "Expect a comparator set, but got ` ||`")]
         [InlineData(">netcore<-1.0 || <", "Expect a moniker string, but got ``")]
+        [InlineData(">*", "Moniker `*` is not defined")]
```
yufeih, line 131: what if `* && >netcore-1.0` or `* || *`

live1206: the only operator should be `=` as described

live1206: Basically, only `*` is allowed

yufeih: so should `=*` also be disallowed?

live1206: they are essentially the same though... `*` will be `=*` in our parser

live1206:
we do not want user to misuse the wildcard, it would cause the output `monikers` metadata exploding

PR #5744, opened 2020-04-08
Clean up dependency map

yufeih:
The dependency types used for loc and archive is `uid`, `file` and `include`.

- LegacyDependencyType -> DependencyType
- Ignore metadata dependency type: https://apidrop.visualstudio.com/Content%20CI/_git/docfx-v3-migration/pullrequest/874?_a=overview

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5744)


live1206:
We need to consolidate `bookmark` into `file` in the migration tool

PR #5729, opened 2020-04-03
Implement TOC splitItemsBy

yufeih:
[AB#192497](https://dev.azure.com/ceapex/Engineering/_workitems/edit/192497)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5729)

PR #5747, opened 2020-04-09
Move exception handling to ParallelUtility

yufeih:
#5729 

Makes it easier to write parallel looks without try catch.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5747)

### DIFF
--- a/src/docfx/lib/log/ErrorLog.cs
+++ b/src/docfx/lib/log/ErrorLog.cs
@@ -142,8 +142,7 @@ namespace Microsoft.Docs.Build
                                             : WarningCount > 0 ? ConsoleColor.Yellow
                                             : ConsoleColor.Magenta;
                     Console.WriteLine();
-                    Console.WriteLine(
-                        $"  {ErrorCount} Error(s), {WarningCount} Warning(s), {SuggestionCount} Suggestion(s)");
+                    Console.WriteLine($"  {ErrorCount} Error(s), {WarningCount} Warning(s), {SuggestionCount} Suggestion(s)");
                 }
 
                 Console.ResetColor();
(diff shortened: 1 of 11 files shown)


yufeih:
@OsmondJiang , do you know why `azure-docs-pr` has these validation diffs? Does some one changed the ruleset?

yufeih:
`devsandbox` diff is due to `errorCode` fix. The repo has build error.

OsmondJiang:
> @OsmondJiang , do you know why `azure-docs-pr` has these validation diffs? Did some one changed the ruleset?

@yufeih today we synced the latest allow list data from PROD to PPE per Yanjun's requests, seems this sync brings new values for ms.service.

image

Add @kfpwolf.

src/docfx/lib/log/ErrorLog.cs
```diff
@@ -35,7 +35,7 @@ internal sealed class ErrorLog : IDisposable
 
         public IEnumerable<FilePath> ErrorFiles => _errorFiles;
 
-        public ErrorLog(string? outputPath, bool legacy = false)
+        public ErrorLog(string? outputPath = null, bool legacy = false)
```
live1206, line 38: if `outputPath = null` is only for testing, can we remove the default value?

src/docfx/lib/log/ErrorLog.cs
```diff
@@ -240,16 +239,11 @@ private bool ExceedMaxErrors(Config? config, ErrorLevel level)
...
-            {
-                return false;
-            }
-
             return level switch
             {
-                ErrorLevel.Error => Interlocked.Increment(ref _errorCount) > config.MaxErrors,
-                ErrorLevel.Warning => Interlocked.Increment(ref _warningCount) > config.MaxWarnings,
-                ErrorLevel.Suggestion => Interlocked.Increment(ref _suggestionCount) > config.MaxSuggestions,
+                ErrorLevel.Error => Interlocked.Increment(ref _errorCount) > (config?.MaxErrors ?? int.MaxValue),
```
live1206, line 244: why not return earlier?

src/docfx/lib/log/ErrorLog.cs
yufeih: sure

src/docfx/lib/log/ErrorLog.cs
yufeih: `_errorCount` would be incorrect.

live1206:
> `devsandbox` diff is due to `errorCode` fix. The repo has build error.
> 
> @live1206 , do you think the diff is expected?

I am just wondering, maybe we should not return 0 from docs-pipeline when publish is skipped. The build should just fail when there is build error for `en-us` repo

mingwli:
> Why 'base_path' diff in DevSandbox?

It's just a line change.

PR #5753, opened 2020-04-10
Support uid attribute in <xref>

mingwli:
[AB#199521](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/199521)

In MarkdownEngine, pass lambda to resolve xref with both href and uid, href has higher priority

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5753)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -8,6 +8,7 @@ using System.Linq;
 using System.Text;
 using System.Web;
 using HtmlAgilityPack;
+using Markdig.Syntax;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
@@ -144,10 +145,11 @@ namespace Microsoft.Docs.Build
             return result.ToString();
         }
 
-        public static string TransformXref(string html, Func<string, bool, int, (string? href, string display)> transform)
+        public static string TransformXref(string html, MarkdownObject block, Func<SourceInfo<string>?, SourceInfo<string>?, bool, (string? href, string display)> resolveXref)
         {
             // Fast pass it does not have <xref> tag
-            if (!(html.Contains("<xref", StringComparison.OrdinalIgnoreCase) && html.Contains("href", StringComparison.OrdinalIgnoreCase)))
+            if (!(html.Contains("<xref", StringComparison.OrdinalIgnoreCase)
+                && (html.Contains("href", StringComparison.OrdinalIgnoreCase) || html.Contains("uid", StringComparison.OrdinalIgnoreCase))))
             {
                 return html;
             }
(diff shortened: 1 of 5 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -165,24 +167,22 @@ public static string TransformXref(string html, Func<string, bool, int, (string?
...
+                    href == null ? null : (SourceInfo<string>?)new SourceInfo<string>(href, block?.ToSourceInfo(columnOffset: columnOffset)),
+                    uid == null ? null : (SourceInfo<string>?)new SourceInfo<string>(uid, block?.ToSourceInfo(columnOffset: columnOffset)),
+                    isShorthand);
 
                 var resolvedNode = new HtmlDocument();
                 if (string.IsNullOrEmpty(resolvedHref))
-                {
-                    resolvedNode.LoadHtml(raw);
-                }
+                    resolvedNode.LoadHtml(rawSource ?? $"<span class=\"xref\">{(href != null ? UrlUtility.SplitUrl(href).path : uid)}</span>");
```
yufeih, line 183: Consider leave the bracket. There was a stylecop rule for this but it was turned off.

src/docfx/lib/markdown/XrefExtension.cs
```diff
@@ -38,23 +40,22 @@ internal static class XrefExtension
...
             });
 
             string ResolveXref(string html, MarkdownObject block)
-            {
-                return HtmlUtility.TransformXref(
-                    html,
-                    (href, isShorthand, columnOffset) => resolveXref(
-                        new SourceInfo<string>(href, block.ToSourceInfo(columnOffset: columnOffset)), isShorthand));
-            }
+                => HtmlUtility.TransformXref(html, block, resolveXref);
```
yufeih, line 58: Don't remove `columnOffset`, it is used to generate two errors if the same xref appears on the same line

src/docfx/lib/HtmlUtility.cs
```diff
@@ -144,10 +145,11 @@ public static string TransformLinks(string html, Func<string, int, string> trans
             return result.ToString();
         }
 
-        public static string TransformXref(string html, Func<string, bool, int, (string? href, string display)> transform)
+        public static string TransformXref(string html, MarkdownObject block, Func<SourceInfo<string>?, SourceInfo<string>?, bool, (string? href, string display)> resolveXref)
         {
             // Fast pass it does not have <xref> tag
-            if (!(html.Contains("<xref", StringComparison.OrdinalIgnoreCase) && html.Contains("href", StringComparison.OrdinalIgnoreCase)))
+            if (!(html.Contains("<xref", StringComparison.OrdinalIgnoreCase)
+                && (html.Contains("href", StringComparison.OrdinalIgnoreCase) || html.Contains("uid", StringComparison.OrdinalIgnoreCase))))
```
herohua, line 152: ` href`/` uid` (with a leading space) would be better?

mingwli: html attribute can also be `attr = 'value'`, with space between `attr`, `=` and `value`
```
<!DOCTYPE html>↩
<html>↩
<head>↩
<title>xxx</title>↩
</head>↩
<body>↩
↩
<img src = "src" alt = "xxxx"/>↩
↩
↩
</body>↩
</html>
```

src/docfx/lib/HtmlUtility.cs
```diff
@@ -165,24 +167,27 @@ public static string TransformXref(string html, Func<string, bool, int, (string?
...
-                var raw = HttpUtility.HtmlDecode(
-                    node.GetAttributeValue("data-raw-html", null) ?? node.GetAttributeValue("data-raw-source", null) ?? $"<span class=\"xref\">{HttpUtility.HtmlEncode(UrlUtility.SplitUrl(xref).path)}</span>");
+                var href = HttpUtility.HtmlDecode(node.GetAttributeValue("href", null));
+                var uid = href == null ? HttpUtility.HtmlDecode(node.GetAttributeValue("uid", null)) : default;
 
-                var isShorthand = raw.StartsWith("@");
-
-                var (resolvedHref, display) = transform(xref, isShorthand, columnOffset);
+                var (resolvedHref, display) = resolveXref(
+                    href == null ? null : (SourceInfo<string>?)new SourceInfo<string>(href, block?.ToSourceInfo(columnOffset: columnOffset)),
```
live1206, line 177: since `href` is nullable, can just be `(SourceInfo<string>?)new SourceInfo<string>(href, block?.ToSourceInfo(columnOffset: columnOffset)`

src/docfx/lib/HtmlUtility.cs
```diff
@@ -165,24 +167,27 @@ public static string TransformXref(string html, Func<string, bool, int, (string?
...
+                    uid == null ? null : (SourceInfo<string>?)new SourceInfo<string>(uid, block?.ToSourceInfo(columnOffset: columnOffset)),
```
live1206, line 178: `uid` is nullable already

src/docfx/lib/markdown/MarkdownEngine.cs
```diff
@@ -182,11 +183,24 @@ private string GetLink(SourceInfo<string> href)
             return link;
         }
 
-        private (string? href, string display) GetXref(SourceInfo<string> href, bool isShorthand)
+        private (string? href, string display) GetXref(SourceInfo<string>? href, SourceInfo<string>? uid, bool isShorthand)
         {
             var status = t_status.Value!.Peek();
-            var (error, link, display, _) = _xrefResolver.ResolveXrefByHref(
-                href, (Document)InclusionContext.File, (Document)InclusionContext.RootFile);
+            var error = default(Error);
```
live1206, line 189: no need to assign here
```csharp
Error error;
string? link;
string display;
```

src/docfx/lib/HtmlUtility.cs
mingwli: This will make this parameter always not null.

are you suggesting change the [GetXref](https://github.com/dotnet/docfx/pull/5753/files#diff-1d01eb080be0edc3ee154a3655a2df8fR186)'s parameter type?
from `Source<string>?` to `Source<string?>`, and do the check there?

src/docfx/lib/markdown/MarkdownEngine.cs
mingwli: you are right!

src/docfx/lib/HtmlUtility.cs
mingwli: Still let call-site decide what to do when href/uid is null.

PR #5766, opened 2020-04-13
Output document_id for redirection file for BI usage

live1206:
[AB#193606](https://dev.azure.com/ceapex/Engineering/_workitems/edit/193606)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5766)

### DIFF
--- a/docs/specs/redirection.yml
+++ b/docs/specs/redirection.yml
@@ -80,7 +80,7 @@ outputs:
    {"document_id": "780484e9-a367-97ab-ba7d-a19f3b378a80", "document_version_independent_id": "e9b3f4e2-1a78-4b5d-fbb4-03e0a33e38f5"}
 ---
 # redirect to can start with https and hostname with locale
-# output document_id for redirection file as well
+# output document_id and canonical_url for redirection file as well
 inputs:
   docfx.yml: |
     hostName: docs.microsoft.com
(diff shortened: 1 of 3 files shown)


src/docfx/build/redirection/BuildRedirection.cs
```diff
@@ -27,6 +28,11 @@ internal static List<Error> Build(Context context, Document file)
                 file.Mime.Value);
 
             publishItem.RedirectUrl = context.RedirectionProvider.GetRedirectUrl(file.FilePath);
+            var (documentId, _) = context.DocumentProvider.GetDocumentId(context.RedirectionProvider.GetOriginalFile(file.FilePath));
```
yufeih, line 31: Do we also want to include `document_version_independent_id`? They seem to appear hand in hand. Though the bug/feature may not say it is required.

live1206: I can sync with the requester, not required from the description

herohua: Yes, we do.

docs/specs/redirection.yml
```diff
@@ -88,6 +89,8 @@ inputs:
     renames:
       a.md: https://docs.microsoft.com/en-us/b?query#fragment
 outputs:
+  .publish.json: |
```
herohua, line 92: also add a case that `redirect_document_id == false`?

src/docfx/build/redirection/BuildRedirection.cs
```diff
@@ -27,6 +28,13 @@ internal static List<Error> Build(Context context, Document file)
                 file.Mime.Value);
 
             publishItem.RedirectUrl = context.RedirectionProvider.GetRedirectUrl(file.FilePath);
+            var (documentId, documentVersionIndependentId) = context.DocumentProvider.GetDocumentId(context.RedirectionProvider.GetOriginalFile(file.FilePath));
+            publishItem.ExtensionData = new JObject
+            {
+                ["document_id"] = documentId,
+                ["document_version_independent_id"] = documentVersionIndependentId,
+                ["canonical_url"] = file.CanonicalUrl,
```
yufeih, line 36: Is `canonical_url` required?  canonical urls are for HTML page SEO purpose, redirection files doesn't get a chance to show the HTML page.

 https://yoast.com/what-is-a-canonical-url/

jesong: BI team need to know the page url. Provide it from canonical_url or from other metadata both OK to me. But, we need the url.

docs/specs/redirection.yml
live1206: good point, test added

src/docfx/build/redirection/BuildRedirection.cs
yufeih: What does v2 and conceptual documents use for this info?

PR #5772, opened 2020-04-14
Add a forward-only HtmlReader

yufeih:
[AB#200575](https://dev.azure.com/ceapex/Engineering/_workitems/edit/200575/) [AB#151947](https://dev.azure.com/ceapex/Engineering/_workitems/edit/151947/) [AB#201261](https://dev.azure.com/ceapex/Engineering/_workitems/edit/201261/)

### Summary

Implements a high-performance, near-zero-allocation, forward-only, read-only `HtmlReader` that confirms to [W3C HTML parsing spec](https://html.spec.whatwg.org/multipage/parsing.html). Applies `HtmlReader` to link resolve and xref resolve.

Unlike HTML DOM apis like `HtmlAgilityPack` or `AngleSharp`,  `HtmlReader` provides a fast, forward only, read only API at the _tokenizer_ layer. (Think `XDocument` vs `XmlReader`).

But that doesn't mean you cannot use `HtmlReader` to manipulate HTML contents. `HtmlReader` provides accurate source information for all tags and attributes at string index level, to allow efficient HTML content transform using string manipulating. Unlike HTML DOM manipulation APIs, this way of transforming HTML content leaves the reset of the HTML content as is. (It does not attempt to do smart things like tag normalization, auto close, etc).

### Implementation notes

- Uses `ReadOnlySpan<T>` to view the attributes of an existing HTML string without allocation.
- Uses `ArrayPool<T>` to pool the attribute list for start tag as needed.
- Uses `ref struct` for `HtmlReader` and `HtmlAttribute` to avoid allocation and ensure stack only,
- The state machine is modeled against the HTML spec using `switch..case`. Method names corresponds to state names.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5772)


src/docfx/lib/html/HtmlAttribute.cs
```diff
@@ -0,0 +1,37 @@
...
+        private readonly (int start, int length) _valueRange;
+        private readonly (int start, int length) _tokenRange;
+
+        public (int start, int length) NameRange => _nameRange;
+
+        public (int start, int length) ValueRange => _valueRange;
+
+        public (int start, int length) TokenRange => _tokenRange;
+
+        public ReadOnlySpan<char> Name => _html.AsSpan().Slice(_nameRange.start, _nameRange.length);
```
mingwli, line 21: Can I think of `AsSpan()` is like creating a pointer to specific block of memory(here it's still allocated on heap?)

And can `AsSpan()` be called only once and cached, instead of many time?

src/docfx/lib/html/HtmlReader.cs
```diff
@@ -0,0 +1,443 @@
...
+        private HtmlAttribute[]? _attributes;
+        private int _attributesLength;
+
+        public HtmlTokenType Type => _type;
+
+        public (int start, int length) NameRange => _nameRange;
+
+        public (int start, int length) ContentRange => _tokenRange;
+
+        public ReadOnlySpan<char> Name => _html.AsSpan().Slice(_nameRange.start, _nameRange.length);
```
mingwli, line 35: Can I think of `AsSpan()` is like creating a pointer to specific block of memory(here the string's still allocated on heap?)

If so, can `AsSpan()` be called only once and cached, instead of many time?

yufeih: https://docs.microsoft.com/en-us/archive/msdn-magazine/2018/january/csharp-all-about-span-exploring-a-new-net-mainstay

herohua:
This is a large PR - would it be helpful that we schedule a meeting to go through it together?

PR #5781, opened 2020-04-15
Return type.uid if other types don't contain required property

v-pegao:
Check if key exist for all type when map, if key doesn't exist, return uid instead.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5781)

### DIFF
--- a/src/docfx.website.themes/common/UniversalReference.common.js
+++ b/src/docfx.website.themes/common/UniversalReference.common.js
@@ -116,13 +116,15 @@ function handleInheritance(tree) {
 function joinType(parameter) {
   // change type in syntax from array to string
   var joinTypeProperty = function (type, key) {
-    if (!type || !type[0] || !type[0][key]) return null;
+    if (!type) return null;
     var value = type.map(function (t) {
+      if (!t || !t[key]) return null;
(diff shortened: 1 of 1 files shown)


superyyrrzz:
We can update the logic like:
```
joinType(parameter, key):
   if (all items does not contain key) 
       return null
   if (some item contains the key)
       return join(item.key ?? item.uid)
```

src/docfx.website.themes/common/UniversalReference.common.js
```diff
@@ -116,14 +116,19 @@ function handleInheritance(tree) {
...
-    if (!type || !type[0] || !type[0][key]) return null;
-    var value = type.map(function (t) {
-      return t[key][0].value;
-    }).join(' | ');
-    return [{
-      lang: type[0][key][0].lang,
-      value: value
-    }];
+    if (!type) return null;
+    if (type.some((t) => { return t && t[key] })) {
```
superyyrrzz: return early for readability: `if (!...) return null`

PR #5788, opened 2020-04-16
SDP unresolved xref default as object

mingwli:
[AB#204964](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/204964)

Fixing parent-contexted property leaked to child-context.

Current, <xref> mustache parser is using `href` property to decide if spec is resolved, in which case if parent context contains a `href` attribute, the child-context will treat it as uid successfully resolved by mistake.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5788)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -277,19 +277,9 @@ namespace Microsoft.Docs.Build
                     var (xrefError, xrefSpec, href) = context.XrefResolver.ResolveXrefSpec(content, file, file);
                     errors.AddIfNotNull(xrefError);
 
-                    if (xrefSpec != null)
-                    {
-                        var specObj = JsonUtility.ToJObject(xrefSpec.ToExternalXrefSpec(href));
-                        return (errors, specObj);
-                    }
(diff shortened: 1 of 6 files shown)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -282,8 +282,14 @@ private JToken LoadXrefProperty(Document file, Context context, SourceInfo<strin
                         var specObj = JsonUtility.ToJObject(xrefSpec.ToExternalXrefSpec(href));
                         return (errors, specObj);
                     }
-
-                    return (errors, value);
+                    else
+                    {
+                        return (errors, new JObject
```
yufeih, line 287: nit: ternary operator with line 282?

mingwli:
> ⚠[DevSandbox](https://github.com/OPS-E2E-PPE/DevSandbox)(00:00:18.2332273, 21 more diff)

expected, the new result is `more` right

PR #5792, opened 2020-04-16
198685 lightbox bug

adunndevster:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5792)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
```diff
@@ -110,7 +111,12 @@ public bool TryProcessAttributes(IDictionary<string, string> attributes, out Htm
                 if (!string.IsNullOrEmpty(currentLightbox))
                 {
                     var lighboxHtmlAttributes = new HtmlAttributes();
-                    lighboxHtmlAttributes.AddProperty("href", $"{currentLightbox}#lightbox");
+                    var path = currentLightbox;
```
adunndevster, line 114: @yufeih - This logic works, but I'm curious, I noticed I did not need to do this for source, because it seems some magic happens downstream from "WriteAttributes" to help the source resolve to the correct path. Do you know why that is?

yufeih: In v2, links are resolved against HTML after markup, that's why it works. But in v3, link resolve happens as part of markup, thus this code won't work for v3.

There is a `MarkdownContext.ReadFile`  hides this difference for content inclusion. I think we'll need a similar `MarkdownContext.GetLink` to handle link resolve.

adunndevster: Hmmm... @yufeih, in my debugging, it never seems to pass through that method. It causes me to wonder if the transformation might be happening somewhere else?

adunndevster: @yufeih - I figured out that if it only runs through that code during InclusionContext tests, but not the Markdig.Extension unit tests. It looks like it would take a big refactor to change the traverse algorithm. I will at least update the pathing logic to match what we have in ResolveLinkExtension.

PR #5795, opened 2020-04-17
set mimeType of conceptual to 'Conceptual'

928PJY:
[AB#204994](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/204994)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5795)

### DIFF
--- a/src/docfx/template/TemplateEngine.cs
+++ b/src/docfx/template/TemplateEngine.cs
@@ -49,6 +49,11 @@ namespace Microsoft.Docs.Build
             return mime is null || !_schemas.TryGetValue(mime, out var schemaTemplate) || schemaTemplate.Value.IsPage;
         }
 
+        public bool IsConceptual(string? mime)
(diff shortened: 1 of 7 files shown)


src/docfx/template/TemplateEngine.cs
```diff
@@ -49,6 +49,11 @@ public bool IsPage(string? mime)
             return mime is null || !_schemas.TryGetValue(mime, out var schemaTemplate) || schemaTemplate.Value.IsPage;
         }
 
+        public bool IsConceptual(string? mime)
```
yufeih, line 52:
```suggestion
        public static bool IsConceptual(string? mime)
```

PR #5809, opened 2020-04-21
Add HtmlWriter, extract HtmlExtension

yufeih:
[AB#200575](https://dev.azure.com/ceapex/Engineering/_workitems/edit/200575/)

This PR adds `HtmlWriter` and refactors `TransformLink`, `TransformXref` and `RemoveRerunCodepenIframes` to use `HtmlReader/HtmlWriter`.

It extracts an `HtmlExtension` that process user HTML content inside markdown, so that each HTML block or HTML inline is parsed only once.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5809)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -98,7 +98,7 @@ namespace Microsoft.Docs.Build
                     var source = block?.ToSourceInfo(columnOffset: attribute.Range.start);
                     var link = HttpUtility.HtmlEncode(transform(new SourceInfo<string>(HttpUtility.HtmlDecode(attribute.Value.ToString()), source)));
 
-                    attribute.Value = link.AsMemory();
+                    attribute = attribute.WithValue(link);
                 }
             }
         }
@@ -152,7 +152,7 @@ namespace Microsoft.Docs.Build
                 ? rawHtml ?? rawSource ?? $"<span class=\"xref\">{(!string.IsNullOrEmpty(display) ? display : (href != null ? UrlUtility.SplitUrl(href).path : uid))}</span>"
                 : $"<a href='{HttpUtility.HtmlEncode(resolvedHref)}'>{HttpUtility.HtmlEncode(display)}</a>";
 
-            token.RawText = resolvedNode.AsMemory();
+            token = token.WithRawText(resolvedNode);
         }
 
         /// <summary>
@@ -264,7 +264,7 @@ namespace Microsoft.Docs.Build
                 {
                     if (attribute.NameIs("src") && attribute.Value.Span.Contains("codepen.io", StringComparison.OrdinalIgnoreCase))
                     {
-                        attribute.Value = (attribute.Value.ToString() + "&rerun-position=hidden&").AsMemory();
+                        attribute = attribute.WithValue(attribute.Value.ToString() + "&rerun-position=hidden&");
                     }
                 }
             }
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -152,7 +152,7 @@ namespace Microsoft.Docs.Build
                 ? rawHtml ?? rawSource ?? $"<span class=\"xref\">{(!string.IsNullOrEmpty(display) ? display : (href != null ? UrlUtility.SplitUrl(href).path : uid))}</span>"
                 : $"<a href='{HttpUtility.HtmlEncode(resolvedHref)}'>{HttpUtility.HtmlEncode(display)}</a>";
 
-            token = token.WithRawText(resolvedNode);
+            token = new HtmlToken(resolvedNode);
         }
 
         /// <summary>
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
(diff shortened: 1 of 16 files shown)


src/docfx/lib/collections/ArrayBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+        public T[] ToArray()
+        {
+            if (_array is null)
+            {
+                return Array.Empty<T>();
+            }
+
+            if (_length == _array.Length)
+            {
+                return _array;
```
herohua, line 45: >_array [](start = 23, length = 6)

safer to always copy?

src/docfx/lib/html/HtmlWriter.cs
```diff
@@ -0,0 +1,158 @@
...
+
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Docs.Build
+{
+    /// <summary>
+    /// Provides a high-performance API for forward-only, non-cached writing of HTML read from HtmlReader.
+    /// </summary>
+    internal class HtmlWriter
```
herohua, line 12: >HtmlWriter [](start = 19, length = 10)

shall we have unit tests for `HtmlWriter`?

src/docfx/lib/html/HtmlWriter.cs
```diff
@@ -0,0 +1,158 @@
...
+            }
+
+            // Reset
+            _replacedToken = null;
+            _newAttributes.Clear();
+            _replacedAttributes.Clear();
+            _replacedAttributeValues.Clear();
+        }
+
+        public void ReplaceTag(string tag)
```
live1206, line 115: This is the only operation method in use, why are we having the others in this PR?

src/docfx/lib/html/HtmlReader.cs
```diff
@@ -7,16 +7,17 @@
...
     /// </summary>
-    internal ref struct HtmlReader
+    internal class HtmlReader
     {
         private string _html;
         private int _length;
         private int _position;
 
         private HtmlTokenType _type;
+        private bool _isSelfClosing;
```
live1206, line 20: can just use auto property

src/docfx/lib/HtmlUtility.cs
```diff
@@ -17,6 +17,20 @@ internal static class HtmlUtility
     {
         private static readonly string[] s_allowedStyles = new[] { "text-align: right;", "text-align: left;", "text-align: center;" };
 
+        public static string TransformHtml(string html, Action<HtmlReader, HtmlWriter> transform)
+        {
+            var reader = new HtmlReader(html);
+            var writer = new HtmlWriter(html.Length + 64);
```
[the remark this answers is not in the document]
yufeih: It is a magic value to tweak based on real world usage. `html.length *2` would require too much memory usage then needed.

src/docfx/lib/html/HtmlWriter.cs
yufeih: For completeness. The others will be used later.

src/docfx/lib/html/HtmlReader.cs
yufeih: I intentionally use private fields here to ensure everything is a field access.

src/docfx/lib/collections/ArrayBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+namespace System.Collections.Generic
+{
+    internal struct ArrayBuilder<T>
+    {
+        private T[]? _array;
+        private int _length;
+
+        public int Length => _length;
+
+        public ReadOnlySpan<T> Span => _array is null ? ReadOnlySpan<T>.Empty : _array.AsSpan(0, _length);
```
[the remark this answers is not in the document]
yufeih: `AsSpan(0, 0)` produces an empty span, and it does so without extra overhead.

src/docfx/lib/html/HtmlWriter.cs
yufeih: yes

src/docfx/lib/html/HtmlAttribute.cs
```diff
@@ -7,38 +7,52 @@ namespace Microsoft.Docs.Build
...
+            Type = value is null ? HtmlAttributeType.NameOnly : (type ?? HtmlAttributeType.DoubleQuoted);
+            Name = name.AsMemory();
+            Value = value.AsMemory();
+            RawText = default;
+            Range = default;
+            ValueRange = default;
         }
 
-        public bool NameIs(string name)
+        internal HtmlAttribute(
```
live1206, line 37:
```suggestion
        public HtmlAttribute(
```

src/docfx/lib/html/HtmlAttribute.cs
```diff
@@ -7,38 +7,52 @@ namespace Microsoft.Docs.Build
...
     internal readonly struct HtmlAttribute
     {
-        private readonly string _html;
-        private readonly (int start, int length) _nameRange;
-        private readonly (int start, int length) _valueRange;
-        private readonly (int start, int length) _tokenRange;
+        public readonly HtmlAttributeType Type { get; }
 
-        public (int start, int length) NameRange => _nameRange;
+        public readonly ReadOnlyMemory<char> Name { get; }
```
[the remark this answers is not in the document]
yufeih: https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-8#readonly-members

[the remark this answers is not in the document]
mingwli: > Auto-implemented properties are an exception; the compiler will treat all auto-implemented getters as readonly

---
In reply to: [415270829](https://github.com/dotnet/docfx/pull/5809#discussion_r415270829) [](ancestors = 415270829)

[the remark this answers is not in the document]
yufeih: @mingwli , 👍 but I'm adding it anyway to enforce readonly a) in case of future change, b) give a clue that everything in this struct is readonly.

yufeih:
Diff are all expected line number changes. `devsandbox` diff due to bad HTML content.

PR #5810, opened 2020-04-21
F#: Cache project infos

saul:
Project infos are now cached as they're expensive to calculate. The RoslynProjectLoader already does this.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5810)


superyyrrzz:
Thank you for the contribution! @saul

PR #5779, opened 2020-04-15
Export href for uid declaration

mingwli:
[AB#195829](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/195829)

1. traverse before `build xrefmap` & `content transform` to get `uidCount`
2. cached:
   - document uid count
   - jobject -> uid (faster is uid declaration check)
   - jobject -> uid-href

PR #5814, opened 2020-04-22
Warning upon href defined side-by-side with uid

mingwli:
Since we auto inject an `href` property at the same level of `uid` in #5779, we should explicitly forbid user from writing `href` for being prudent.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5814)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -439,7 +439,7 @@ namespace Microsoft.Docs.Build
             /// </summary>
             /// Behavior: ✔️ Message: ❌
             public static Error UidOverwriteHref(SourceInfo source)
-                => new Error(ErrorLevel.Error, "uid-overwrite-href", $"'UID' and 'href' defined at the same level is not allowed, the href value will be overwritten", source);
+                => new Error(ErrorLevel.Warning, "uid-overwrite-href", $"'UID' and 'href' defined at the same level is not allowed, the href value will be overwritten", source);
         }
 
         public static class Versioning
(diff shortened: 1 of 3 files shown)


src/docfx/Errors.cs
```diff
@@ -433,6 +433,13 @@ public static Error DuplicateUid(SourceInfo<string> uid, IEnumerable<SourceInfo>
             /// Behavior: ✔️ Message: ❌
             public static Error UidPropertyConflict(string uid, string propertyName, IEnumerable<string?> conflicts)
                 => new Error(ErrorLevel.Warning, "xref-property-conflict", $"UID '{uid}' is defined with different {propertyName}s: {StringUtility.Join(conflicts)}");
+
+            /// <summary>
+            /// Define href property at the same level with uid, href value will be overwritten.
+            /// </summary>
+            /// Behavior: ✔️ Message: ❌
+            public static Error UidOverwriteHref(SourceInfo source)
+                => new Error(ErrorLevel.Error, "uid-overwrite-href", $"'UID' and 'href' defined at the same level is not allowed, the href value will be overwritten", source);
```
herohua, line 442: 'UID' and 'href' are not allowed to define at the same level. The input 'href' will be ignored.

herohua, line 442: is this an error or warning (build continues)?

PR #5815, opened 2020-04-22
update output design spec to involve two new configs: OutputType, OutputUrlType

928PJY:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5815)

### DIFF
--- a/docs/designs/output.md
+++ b/docs/designs/output.md
@@ -21,7 +21,7 @@ The output URL schema and output path schema are controlled by two config items:
 
 - OutputUrlType:
     - `Docs`: a.md -> `/xxx/a`
-    - `Static`: a.md -> `/xxx/a/`
+    - `StaticPretty`: a.md -> `/xxx/a/`
     - `StaticUgly`: a.md -> `/xxx/a.html`(or `/xxx/a.json`)
 
 ## URL Schema
@@ -66,7 +66,7 @@ https://docs.microsoft.com/en-us/dotnet/api/system.string.html#Instantiation
 
 ## Content Outputs
 
-Each input file transforms to zero or more output files depending on its content type. `docfx` places output files in `output-path` relative to output directory.:
+Each input file transforms to zero or more output files depending on its content type. `docfx` places output files in `output-path` relative to output directory:
 
 ### Docs hosting output path schema:
 
@@ -86,10 +86,6 @@ _site/dotnet/01ddf122/api/system.string.json
   base-path     site-path (relative-to-base-path)
       |--^-| |------------------^-----------------|
 _site/dotnet/01ddf122/api/system.string/index.html
-
-  base-path     site-path (relative-to-base-path)
-      |--^-| |------------------^----------|
-_site/dotnet/01ddf122/api/system.string.html
 ```
 
 Different files can share the same `{site-url}` or `{site-path}` due to versioning and localization.
@@ -175,6 +171,8 @@ Different files can share the same `{site-url}` or `{site-path}` due to versioni
     | `site-path` | dotnet/01ddf122/api/thumbnail.png |
     | `output-path` | dotnet/01ddf122/api/thumbnail.png |
 
+> Note: For static page versioning, the ideal output path should use the human readable `moniker` instead of monikerHash, but that require coping the same content to different output folder of each version. For the time been, the static output is only for generating PDF, the URL does not appear in PDF now, so we can still use the monikerHash.
+
 ## System Generated Outputs
 
 Besides content files, `docfx` generates miscellaneous files during the build. The name of these system generated files starts with `.`. The extension MUST be `.json` for files in json format.
@@ -185,7 +183,16 @@ File name           | Description
 .xrefmap.json        | A manifest of `uid` and xref specs as described in [xref](xref.md)
 .errors.log          | A report file that contains build errors and warnings. Each line is a json array: `[{level}, {code}, {message}, {file?}, {line?}, {column?}]`.
 
-## Copy Resource
+## Dependency Copy
+
+As the requirements mentioned, `Static build output xcopy deployable`, the output folder should be self-contained in static build.
+
+# Resource
 
 To improve docs publish performance, there is feature to skip copy resource files to the output folder.
 With those two new configs(`OutputType` and `OutputUrlType`) involved, whether need to copy resource should be determined  by `OutputUrlType`: skip copy resource files when it is `Docs`.
+
+# Template
+
+For Docs hosting, `_themes` resources is published separately, so we don't need to copy them into the output folder.
+With those two new configs(`OutputType` and `OutputUrlType`) involved, whether need to copy used resources(`.css` files) should be determined  by `OutputUrlType`: skip copy resource files when it is `Docs`.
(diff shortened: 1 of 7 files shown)


docs/designs/output.md
```diff
@@ -54,22 +66,38 @@ https://docs.microsoft.com/en-us/dotnet/api/system.string.html#Instantiation
...
+`{output-dir}/{base-path}/{site-path}`
+
+```
+  base-path     site-path (relative-to-base-path)
+      |--^-| |------------------^-----------------|
 _site/dotnet/01ddf122/api/system.string/index.html
+
+  base-path     site-path (relative-to-base-path)
+      |--^-| |------------------^----------|
+_site/dotnet/01ddf122/api/system.string.html
```
live1206, line 92: static hosting with ugly URL? since the detailed example has been added later. Do we still need it here?

docs/designs/output.md
```diff
@@ -54,22 +66,38 @@ https://docs.microsoft.com/en-us/dotnet/api/system.string.html#Instantiation
 
 ## Content Outputs
 
-Each input file transforms to zero or more output files depending on its content type. `docfx` places output files in `output-path` relative to output directory. **Regardless of static rendering or dynamic rendering, output path shares the same schema**:
+Each input file transforms to zero or more output files depending on its content type. `docfx` places output files in `output-path` relative to output directory.:
```
herohua, line 69: >. [](start = 160, length = 1)

remove `.`

docs/designs/output.md
```diff
@@ -128,4 +185,7 @@ File name           | Description
 .xrefmap.json        | A manifest of `uid` and xref specs as described in [xref](xref.md)
 .errors.log          | A report file that contains build errors and warnings. Each line is a json array: `[{level}, {code}, {message}, {file?}, {line?}, {column?}]`.
 
+## Copy Resource
 
+To improve docs publish performance, there is feature to skip copy resource files to the output folder.
+With those two new configs(`OutputType` and `OutputUrlType`) involved, whether need to copy resource should be determined  by `OutputUrlType`: skip copy resource files when it is `Docs`.
```
herohua, line 191: >OutputUrlType [](start = 127, length = 13)

I would suggest having this as a separate option because even for output type `Html`, we can improve perf by avoid copying images.

928PJY: it is not determined by `OutputType`, but `OutputUrl` type, as the requirements list above, `Static build output xcopy deployable` requires the output to be self-contained.

---
In reply to: [412878126](https://github.com/dotnet/docfx/pull/5815#discussion_r412878126) [](ancestors = 412878126)

docs/designs/output.md
```diff
@@ -11,34 +11,46 @@ This document specifies docfx output file layout. It is designed to satisfy thes
...
+
+The output URL schema and output path schema are controlled by two config items:
+
+- OutputType:
+    - `Html`: apply liquid to generate HTML output
+    - `Json`: output Json file without liquid applied
+
+- OutputUrlType:
+    - `Docs`: a.md -> `/xxx/a`
+    - `Static`: a.md -> `/xxx/a/`
```
yufeih, line 24: How about `docs`, `pretty` and `ugly` ? The initial names are `dynamic`, `staticPretty`, `staticUgly`, since we renamed `dynamic` to `docs`, I guess there is no need to include `static`.

docs/designs/output.md
```diff
@@ -79,44 +107,73 @@ Different files can share the same `{site-url}` or `{site-path}` due to versioni
...
     | `output-path` | dotnet/01ddf122/api/system.string.json |
 
-- Content for static rendering using pretty url
+- Content for static hosting using pretty URL(OutputType: `Html`):
 
     | | |
     |------ |----|
-    | `url` | https://docs.microsoft.com/en-us/dotnet/api/system.string/ |
-    | `site-url` | /dotnet/api/system.string/ |
+    | `url` | https://self-hosting/en-us/dotnet/01ddf122/api/system.string/ |
```
yufeih, line 114: Add a note that for static page versioning, better use _moniker_ in the URL as it is human readable. That may require copying the same content to different output locations.

For the time been, we use moniker hash because the URL does not appear in PDF.
```suggestion
    | `url` | https://self-hosting/en-us/dotnet/v1/api/system.string/ |
```

docs/designs/output.md
yufeih: Keeping it a separate option allows us to create a self contained package for docs that can be re-published later. Maybe changing the name from `copyResources` --> `selfContained` describes the intent better.

928PJY: what does this requirements come from?

---
In reply to: [413451101](https://github.com/dotnet/docfx/pull/5815#discussion_r413451101) [](ancestors = 413451101)

docs/designs/output.md
928PJY: prefer to keep the static to make it more specific, so it be `Docs`, `StaticPretty` and `StaticUgly`

---
In reply to: [413447678](https://github.com/dotnet/docfx/pull/5815#discussion_r413447678) [](ancestors = 413447678)

docs/designs/output.md
928PJY: and in the current implementation, the `CopyResource` config doesn't work when in legacy mode..

---
In reply to: [413454336](https://github.com/dotnet/docfx/pull/5815#discussion_r413454336) [](ancestors = 413454336,413451101)

docs/designs/output.md
```diff
@@ -79,44 +103,75 @@ Different files can share the same `{site-url}` or `{site-path}` due to versioni
...
+- Image in static hosting:
+
+    | | |
+    |------ |----|
+    | `url` | https://docs.microsoft.com/en-us/dotnet/01ddf122/api/thumbnail.png |
+    | `site-url` | /dotnet/01ddf122/api/thumbnail.png |
+    | `site-path` | dotnet/01ddf122/api/thumbnail.png |
+    | `output-path` | dotnet/01ddf122/api/thumbnail.png |
+
+> Note: For static page versioning, the ideal output path should use the human readable `moniker` instead of monikerHash, but that require coping the same content to different output folder of each version. For the time been, the static output is only for generating PDF, the URL does not appear in PDF now, so we can still use the monikerHash.
```
live1206, line 174:
```suggestion
> Note: For static page versioning, the ideal output path should use the human readable `moniker` instead of monikerHash, but that require coping the same content to different output folder of each version. For the time being, the static output is only for generating PDF, the URL does not appear in PDF now, so we can still use the monikerHash.
```

docs/designs/output.md
```diff
@@ -11,34 +11,46 @@ This document specifies docfx output file layout. It is designed to satisfy thes
...
+
+The output URL schema and output path schema are controlled by two config items:
+
+- OutputType:
+    - `Html`: apply liquid to generate HTML output
+    - `Json`: output Json file without liquid applied
+
+- OutputUrlType:
+    - `Docs`: a.md -> `/xxx/a`
+    - `StaticPretty`: a.md -> `/xxx/a/`
```
yufeih, line 24: `pretty`, `ugly`, `docs`

PR #5806, opened 2020-04-21
Support static output

928PJY:
1. change output url schema according to this [Spec](https://github.com/dotnet/docfx/pull/5815)
2. generate toc.html for static putout mode

TODOs: 
1. Handle `redirection-url-not-found` warning for static-output mode
2. Report a warning if the liquid does not exist
3. Whether need to copy resource should be determined by OutputUrlType

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5806)

PR #5820, opened 2020-04-23
Report a warning if the liquid is not found when generating html output

928PJY:
[AB#204994](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/204994)

#5806 TODO-2

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5820)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -88,6 +88,13 @@ namespace Microsoft.Docs.Build
 
             public static Error MetadataValidationRuleset(string ruleset)
                 => new Error(ErrorLevel.Info, "MetadataValidationRuleset", $"Metadata validation ruleset used: {ruleset}");
+
+            /// <summary>
+            /// Liquid is not found for curtrent mime type.
+            /// </summary>
+            /// Behavior: ❌ Message: ❌
+            public static Error LiquidNotFound(SourceInfo<string?> source)
+                => new Error(ErrorLevel.Warning, "liquid-not-found", $"Liquid used to genenrate HTML is not found for mimeType '{source}'", source);
         }
 
         public static class Logging
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -90,11 +90,11 @@ namespace Microsoft.Docs.Build
                 => new Error(ErrorLevel.Info, "MetadataValidationRuleset", $"Metadata validation ruleset used: {ruleset}");
 
             /// <summary>
-            /// Liquid is not found for curtrent mime type.
+            /// Liquid is not found for current mime type.
             /// </summary>
             /// Behavior: ❌ Message: ❌
             public static Error LiquidNotFound(SourceInfo<string?> source)
-                => new Error(ErrorLevel.Warning, "liquid-not-found", $"Liquid used to genenrate HTML is not found for mimeType '{source}'", source);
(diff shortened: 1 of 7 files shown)


src/docfx/Errors.cs
```diff
@@ -88,6 +88,13 @@ public static Error MicrosoftGraphApiFailed(string exMessage)
 
             public static Error MetadataValidationRuleset(string ruleset)
                 => new Error(ErrorLevel.Info, "MetadataValidationRuleset", $"Metadata validation ruleset used: {ruleset}");
+
+            /// <summary>
+            /// Liquid is not found for curtrent mime type.
```
herohua, line 93: typo: `current`

src/docfx/Errors.cs
```diff
@@ -88,6 +88,13 @@ public static Error MicrosoftGraphApiFailed(string exMessage)
...
+            /// </summary>
+            /// Behavior: ❌ Message: ❌
+            public static Error LiquidNotFound(SourceInfo<string?> source)
+                => new Error(ErrorLevel.Warning, "liquid-not-found", $"Liquid used to genenrate HTML is not found for mimeType '{source}'", source);
```
herohua, line 97: `Liquid` -> `Liquid template`
and better to add `The output HTML will not be generated.`

herohua: typo: `generate`

src/docfx/template/liquid/LiquidTemplate.cs
```diff
@@ -52,7 +52,7 @@ public string Render(string templateName, JObject model)
             // if liquid template not found, return the json
```
mingwli, line 52: This comment can be removed : )-

src/docfx/template/liquid/LiquidTemplate.cs
```diff
@@ -52,7 +52,7 @@ public string Render(string templateName, JObject model)
...
             if (template is null)
             {
-                return JsonUtility.Serialize(model);
+                throw Errors.System.LiquidNotFound(mimeType).ToException(isError: false);
```
mingwli, line 55: The liquid name is decided by `layout` property, should `layout` be reported too?

928PJY: I thought about that too, but the `layout` is generated by the template script, not set by the user, so it may bring confusion.

928PJY: I thought about that too, but the layout is generated by the template script, not set by the user, so it may bring confusion.

---
In reply to: [414273833](https://github.com/dotnet/docfx/pull/5820#discussion_r414273833) [](ancestors = 414273833)

herohua: I agree with Jiayin.

---
In reply to: [415189929](https://github.com/dotnet/docfx/pull/5820#discussion_r415189929) [](ancestors = 415189929,414273833)

928PJY:
@yufeih thought that we can make the error code more general for all the template apply warnings, for example, js script crash. @herohua do you have any suggestions?

herohua:
> @yufeih thought that we can make the error code more general for all the template apply warnings, for example, js script crash. @herohua do you have any suggestions?

In this case, I would prefer to have `LiquidNotFound` separate from other template issues considering the impact. Partial PDFs may not be a problem to end users, but crashing is definitely a system error.

PR #5821, opened 2020-04-23
integrate .sourcemap.json

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5821)


docs/specs/ecma2Yaml.yml
```diff
@@ -18,7 +18,8 @@ repos:
             },
           }],
         }
-      docfx.yml:
+      docfx.yml: |
+          sourceMap: .sourceMap.json
```
yufeih, line 22: `OpsConfigLoader` should auto populate this property, otherwise this needs config change in repos.

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -169,6 +169,12 @@ public void Save()
             }
         }
 
+        private PathString GetOriginalFullPath(FilePath file)
+        {
+            var originalPath = _sourceMap.GetOriginalFilePath(file.Path);
```
yufeih, line 174: With this change, should `SourceMap.GetOriginalFilePath` take `FilePath` as input to only apply to `FileOrigin.Main`?

live1206: yes, it should be. Adding an assert for it.

docs/specs/ecma2Yaml.yml
live1206: https://github.com/dotnet/docfx/pull/5687#discussion_r399924939
I thought a config change is needed...

yufeih: Oh, I mean we'll expose a config in docfx as a feature, but for existing OPS repos, it'll have a thin compact layer in OpsConfigLoader. 

We aren't touching any of the user configs during the migration.

live1206: OK, got your point

PR #5823, opened 2020-04-24
Validate html by allowlist 

fkpwolf:
find un-safe html and report warning

Feature https://dev.azure.com/ceapex/Engineering/_workitems/edit/192183

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5823)

PR #5835, opened 2020-04-27
Add and track Error.Name

yufeih:
#5823 #5817

Several cases requires tracking additional dimension in addition to error code. This PR adds a `Name` property to `Error` and use it to:
- Track additional telemetry associated with the error
- Customize error code, error message based on `Name`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5835)

### DIFF
--- a/src/docfx/lib/log/Telemetry.cs
+++ b/src/docfx/lib/log/Telemetry.cs
@@ -25,7 +25,7 @@ namespace Microsoft.Docs.Build
         private static readonly MetricConfiguration s_metricConfiguration = new MetricConfiguration(1000, int.MaxValue, new MetricSeriesConfigurationForMeasurement(false));
 
         private static readonly Metric s_operationTimeMetric = s_telemetryClient.GetMetric(new MetricIdentifier(null, $"Time", "Name", "OS", "Version", "Repo", "Branch", "CorrelationId"), s_metricConfiguration);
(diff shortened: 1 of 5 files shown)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -64,9 +64,9 @@ public static IDisposable TrackingOperationTime(TelemetryName name)
             return new PerfScope(name.ToString());
         }
 
-        public static void TrackErrorCount(string code, ErrorLevel level)
+        public static void TrackErrorCount(string code, ErrorLevel level, string? name)
         {
-            s_errorCountMetric.TrackValue(1, code, level.ToString(), "User", s_os, s_version, s_repo, s_branch, s_correlationId);
+            s_errorCountMetric.TrackValue(1, code, level.ToString(), CoalesceEmpty(name), "User", s_os, s_version, s_repo, s_branch, s_correlationId);
```
OsmondJiang, line 69: how about use Dictionary<string, string> properties, then the caller can track any dimensions they want.

use s_telemetryClient.TrackMetrics()

yufeih: Unfortunately, the dimension count for metric is fixed. (It is reasonable because pre-aggregation needs to accumulate values based on a fixed dimension).

PR #5839, opened 2020-04-28
Split toc wrap into items

mingwli:
[AB#210316](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/210316)

 wrap splitted item root into items, otherwise its properties will be [lost after loading toc as object](https://github.com/dotnet/docfx/blob/v3/src/docfx/build/toc/TableOfContentsParser.cs#L54)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5839)

### DIFF
--- a/docs/specs/toc.yml
+++ b/docs/specs/toc.yml
@@ -1494,26 +1494,43 @@ inputs:
     splitItemsBy: name
     items:
     - name: 1
+      href: a.md
       items:
       - name: 1-1
-        href: a.md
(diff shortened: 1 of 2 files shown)


mingwli:
image

PR #5842, opened 2020-04-28
Only validate user HTML

yufeih:
[AB#212595](https://dev.azure.com/ceapex/Engineering/_workitems/edit/212595)

Validate markdown HTML user input instead of markup result.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5842)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -21,19 +21,16 @@ internal static class HtmlUtility
         private static readonly string[] s_allowedStyles = new[] { "text-align: right;", "text-align: left;", "text-align: center;" };
 
         // todo remove `class` & `hidden` & `tabindex` after we can know it is is generated by ourself
-        private static readonly HashSet<string> s_globalAllowedAttributes = new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "id", "itemid", "itemprop", "itemref", "itemscope", "itemtype", "part", "slot", "spellcheck", "title", "role", "class", "hidden", "tabindex" };
```
fkpwolf, line 24: can we remove class & hidden & tabindex here since currently we can make sure the attributes is in user HTML?

PR #5854, opened 2020-04-30
Detect circular redirection

live1206:
[AB#209225](https://dev.azure.com/ceapex/Engineering/_workitems/edit/209225)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5854)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -355,6 +355,9 @@ namespace Microsoft.Docs.Build
                             "Replace the redirect URL with a valid absolute URL, or set redirect_document_id to false in .openpublishing.redirection.json.";
                 return new Error(ErrorLevel.Suggestion, "redirection-url-not-found", message, source);
             }
+
+            public static Error CircularRedirection(SourceInfo? source, IEnumerable<FilePath> redirectionChain)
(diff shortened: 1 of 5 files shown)


src/docfx/Errors.cs
```diff
@@ -355,6 +355,9 @@ public static Error RedirectionUrlNotFound(string from, SourceInfo<string> sourc
                             "Replace the redirect URL with a valid absolute URL, or set redirect_document_id to false in .openpublishing.redirection.json.";
                 return new Error(ErrorLevel.Suggestion, "redirection-url-not-found", message, source);
             }
+
+            public static Error CircularRedirection(SourceInfo? source, IEnumerable<FilePath> redirectionChain)
```
yufeih, line 359: nit: reuse `CircularReference` as `FilePath.ToString()` already says it is a `redirection`?

live1206: `CircurlarRefernce` is `Error`, and `CirclaurRedirection` is supposed to be `Warning`?

yufeih: Yeah, warning is better.

PR #5863, opened 2020-05-06
always output object style report

928PJY:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5863)

### DIFF
--- a/src/docfx/lib/log/Error.cs
+++ b/src/docfx/lib/log/Error.cs
@@ -69,16 +69,14 @@ namespace Microsoft.Docs.Build
 
         public string ToString(ErrorLevel level, SourceMap? sourceMap)
         {
-            var originalPath = FilePath is null ? null : sourceMap?.GetOriginalFilePath(FilePath);
-            var payload = originalPath is null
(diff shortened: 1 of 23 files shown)


src/docfx/lib/log/Error.cs
```diff
@@ -69,16 +69,14 @@ public Error WithLevel(ErrorLevel level)
...
-            }
-            return JsonUtility.Serialize(payload.Take(i + 1));
+            var message_severity = level;
+            var file = FilePath is null ? FilePath?.Path : sourceMap?.GetOriginalFilePath(FilePath) ?? FilePath?.Path;
+            var date_time = DateTime.UtcNow;
+            var log_item_type = "user";
+            var end_line = EndLine;
+            var end_column = EndColumn;
+
+            return JsonUtility.Serialize(new { message_severity, log_item_type, Code, Message, file, Line, end_line, Column, end_column, date_time });
```
live1206, line 79: we intentionally to omit the line info if it is from source map, because the line info is incorrect

PR #5868, opened 2020-05-07
apply source map for more output

live1206:
[AB#215115](https://dev.azure.com/ceapex/Engineering/_workitems/edit/215115)

Apply source map for more output:
- .manifet.json
- .publish.json
- op_aggregated_file_map_info.json

Leave `filemap.json` as it is, because it is supposed to be 1 to 1 relationship, which would be 1 to many with source map.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5868)

### DIFF
--- a/docs/specs/output.yml
+++ b/docs/specs/output.yml
@@ -672,6 +672,13 @@ outputs:
     {"message_severity":"error","code":"schema-not-found","message":"Unknown schema ''","file":"a/original.xml"}
   .dependencymap.json: |
     {"dependencies":{"a/b.original.xml":[{"source":"a/original.xml","type":"file"}]}}
+  .publish.json: |
+    {
+      "files": [
+        {"url": "/b", "source_path": "a/b.original.xml"},
+        {"url": "/c", "source_path": "a/original.xml"},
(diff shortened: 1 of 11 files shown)


docs/specs/output.yml
```diff
@@ -672,6 +672,13 @@ outputs:
     {"message_severity":"error","code":"schema-not-found","message":"Unknown schema ''","file":"a/original.xml"}
   .dependencymap.json: |
     {"dependencies":{"a/b.original.xml":[{"source":"a/original.xml","type":"file"}]}}
+  .publish.json: |
+    {
+      "files": [
+        {"url": "/b", "source_path": "a/b.original.xml"},
```
yufeih, line 678: Would you add a comment to `PublishItem.SourcePath` to say who is using it? I couldn't remember it immediately and digged into code, it seems to be used by publish output for PR comment?

live1206: yes

PR #5875, opened 2020-05-08
Process toc nodes in parallel

mingwli:
Parse each toc node in parallel
```
LoadTocNodes => {
  // parallel for each
  LoadTocNode
}
```

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5875)


src/docfx/lib/collections/CollectionUtility.cs
```diff
@@ -15,12 +16,28 @@ internal static class CollectionUtility
             }
         }
 
+        public static void AddIfNotNull<T>(this ConcurrentBag<T> @this, T? value) where T : class
+        {
+            if (!(value is null))
```
live1206, line 21:
```suggestion
            if (value != null)
```

mingwli:
> ⚠[mc-docs-pr](https://github.com/opse2epr/mc-docs-pr)(00:00:49.8796627, 219958 more diff)

Caused by FileLinkItem equals check

PR #5879, opened 2020-05-09
fix crash on versioning file redirect to url of non-versioning file

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5879)

### DIFF
--- a/docs/specs/redirection.yml
+++ b/docs/specs/redirection.yml
@@ -514,18 +514,15 @@ inputs:
   docfx.yml: |
     monikerRange:
       'docs/v1/**': '< netcore-2.0'
-      'docs/v2/**': '>= netcore-2.0'
     hostName: docs.com
     basePath: /docs
     routes:
       docs/: .
       docs/v1/: .
-      docs/v2/: .
     monikerDefinition: monikerDefinition.json
   redirections.yml: |
     redirections:
       docs/v1/c.md: /docs/test/a
-      docs/v2/c.md: /docs/b
   docs/test/a.md:
   monikerDefinition.json: |
     {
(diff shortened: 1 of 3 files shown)


docs/specs/redirection.yml
```diff
@@ -508,3 +508,30 @@ outputs:
     {"message_severity":"warning","code":"circular-redirection","message":"Build has identified circular redirection: a.md [redirection] --> b.md [redirection] --> c.md [redirection] --> a.md [redirection]","file":"redirections.yml","line":2,"column":9}
     {"message_severity":"warning","code":"circular-redirection","message":"Build has identified circular redirection: b.md [redirection] --> c.md [redirection] --> a.md [redirection] --> b.md [redirection]","file":"redirections.yml","line":3,"column":9}
     {"message_severity":"warning","code":"circular-redirection","message":"Build has identified circular redirection: c.md [redirection] --> a.md [redirection] --> b.md [redirection] --> c.md [redirection]","file":"redirections.yml","line":4,"column":9}
+---
+# Versioning file redirec to url of non-versioning file
```
yufeih, line 512:
```suggestion
# Versioning file redirection to url of non-versioning file
```

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -213,17 +213,21 @@ private static IEnumerable<string> ProbeRedirectionFiles(string docsetPath, Repo
...
-                    candidates = docs.Where(doc => _monikerProvider.GetFileLevelMonikers(doc).monikers.Length == 0).ToList();
+                    candidates = docs.Where(doc => _monikerProvider.GetFileLevelMonikers(doc).monikers.Length == 0);
                 }
                 else
                 {
-                    candidates = docs.Where(doc => _monikerProvider.GetFileLevelMonikers(doc).monikers.Intersect(redirectionSourceMonikers).Any()).ToList();
+                    candidates = docs.Where(doc => _monikerProvider.GetFileLevelMonikers(doc).monikers.Intersect(redirectionSourceMonikers).Any());
+                }
+
+                if (candidates.Any())
```
yufeih, line 226: This will call the enumerator twice, once in `Any`, once in `OrderBy`

live1206: `Any` will execute only 1 iteration, but still `Orderby` and the later `foreach` will execute the `enumeration` twice

PR #5880, opened 2020-05-09
Move strip tags to HtmlExtension

yufeih:
This way, we only strip tags from user input the same way as validating tags.

[AB#201261](https://dev.azure.com/ceapex/Engineering/_workitems/edit/201261/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5880)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Buffers;
```
live1206, line 5: remove unused references

src/docfx/lib/HtmlUtility.cs
```diff
@@ -16,9 +16,7 @@ namespace Microsoft.Docs.Build
 {
     internal static class HtmlUtility
     {
-        public delegate void TransformHtmlDelegate(ref HtmlToken token);
-
-        private static readonly string[] s_allowedStyles = new[] { "text-align: right;", "text-align: left;", "text-align: center;" };
+        public delegate void TransformHtmlDelegate(ref HtmlReader reader, ref HtmlToken token);
```
live1206, line 19: why is `HtmlToken` not `ref struct` just `struct`?

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -256,18 +257,19 @@ private JToken LoadXrefProperty(Document file, Context context, SourceInfo<strin
                     errors.AddRange(markupErrors);
 
                     // todo: use BuildPage.CreateHtmlContent() when we only validate markdown properties' bookmarks
-                    return (errors, HtmlUtility.LoadHtml(html).PostMarkup(context.Config.DryRun).WriteTo());
+                    return (errors, html);
```
live1206, line 260: is `stripTags` fast enough for `DryRun` now?

src/docfx/lib/HtmlUtility.cs
yufeih: `HtmlToken` and `HtmlAttributes` are bullets (think `JToken`), `HtmlReader`, `HtmlWriter` are guns (think `JsonSerializer`).

src/docfx/lib/html/HtmlReader.cs
```diff
@@ -45,6 +45,25 @@ public HtmlReader(string html)
             _length = html.Length;
         }
 
+        public bool ReadToEndTag(ReadOnlySpan<char> name)
```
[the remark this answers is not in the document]
yufeih: Yes, I kept it in the signature so it could eventually become a self contained library.

src/docfx/lib/schema/JsonSchemaTransformer.cs
yufeih: Yes, I think so.

PR #5895, opened 2020-05-12
Replace HtmlUtility with HtmlReader

yufeih:
[AB#217885](https://dev.azure.com/ceapex/Engineering/_workitems/edit/217885)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5895)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -117,46 +116,24 @@ public static HtmlNode LoadHtml(string html)
...
-            var total = 0L;
-            foreach (var child in node.ChildNodes)
-            {
-                total += CountWord(child);
-            }
-            return total;
         }
 
-        public static HashSet<string> GetBookmarks(this HtmlNode html)
+        public static void GetBookmarks(ref HtmlToken token, HashSet<string> bookmarks)
```
live1206, line 128: why not return `HashSet<string>`?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -117,46 +116,24 @@ public static HtmlNode LoadHtml(string html)
             return doc.DocumentNode;
         }
 
-        public static long CountWord(this HtmlNode node)
+        public static void CountWord(ref HtmlToken token, ref long wordCount)
```
live1206, line 119: consider returning `long`

src/docfx/lib/html/HtmlToken.cs
```diff
@@ -15,7 +15,7 @@ internal struct HtmlToken
 
         public ReadOnlyMemory<char> RawText { get; }
 
-        public Memory<HtmlAttribute> Attributes { get; }
+        public Memory<HtmlAttribute> Attributes { get; private set; }
```
live1206, line 18: how about maintaining a field `private HtmlAttribute[] _attributes` and return `_attributes.AsMemory()` for this readonly property.

PR #5897, opened 2020-05-12
Use inline codesign

sharanya-rao:
[AB#216841](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/216841)

Test build: https://ceapex.visualstudio.com/Engineering/_build/results?buildId=146572&view=results

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5897)


azure-pipelines-master.yml
```diff
@@ -33,12 +33,134 @@ steps:
...
+    MaxConcurrency: '50'
+    MaxRetryAttempts: '5'
+
+- task: EsrpCodeSigning@1
+  displayName: Sign dependent dlls
+  inputs:
+    ConnectedServiceName: 'CodeSigning-APEX'
+    FolderPath: '$(Build.SourcesDirectory)/target/Release'
+    Pattern: |
+        **\DotLiquid.resources.dll
```
superyyrrzz, line 98: Can we use the exclude syntax here? My concern is that we may forget to update this when add a dependency in the future. We should try avoid missing signing files.

sharanya-rao: let me see

sharanya-rao: Updated. Please review.

PR #5903, opened 2020-05-13
214972 code snippet fix

adunndevster:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5903)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/CodeExtension.cs
```diff
@@ -308,11 +308,11 @@ public static int FindIndexOfTag(this string[] codeLines, string id, bool isEnd
                 return Array.FindIndex(codeLines, line => line.IndexOf(id, StringComparison.OrdinalIgnoreCase) > -1) + 2;
             } else
             {
+                Regex endTagRegex = new Regex($"\\s*(?:</\\s*{id}\\s*?>)|(?:(?:;|%)\\s*<\\s*{id}\\s*?>)", RegexOptions.IgnoreCase);
```
adunndevster, line 311: Too expensive?

yufeih: A general question for code snippet extension. There used to be a different [CodeSnippet](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/CodeSnippet/HtmlCodeSnippetRenderer.cs#L123) syntax and it also does code snippet extraction, is there a difference between that and the new triple column code snippet extractor? Could the logic be shared?

adunndevster: @yufeih - Thanks for pushing me on this. I was able to remove a lot of code, and reuse lots of logic. You may notice some of the tests are modified, just a little for small things like minor indent changes.

PR #5910, opened 2020-05-15
Skip logging when error code is missing

live1206:
[AB#219773](https://dev.azure.com/ceapex/Engineering/_workitems/edit/219773)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5910)

### DIFF
--- a/src/docfx/config/ops/OpsPreProcessor.cs
+++ b/src/docfx/config/ops/OpsPreProcessor.cs
@@ -52,13 +52,7 @@ namespace Microsoft.Docs.Build
                         Path.Combine(_buildOptions.DocsetPath, _config.Monodoc.SourceXmlFolder),
                         Path.Combine(_buildOptions.DocsetPath, _config.Monodoc.OutputYamlFolder),
                         fallbackXmlPath,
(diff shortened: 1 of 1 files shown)


src/docfx/config/ops/OpsPreProcessor.cs
```diff
@@ -52,7 +52,13 @@ private void PreProcessMonoDocXml()
                         Path.Combine(_buildOptions.DocsetPath, _config.Monodoc.SourceXmlFolder),
                         Path.Combine(_buildOptions.DocsetPath, _config.Monodoc.OutputYamlFolder),
                         fallbackXmlPath,
-                        item => _errorLog.Write(ConstructError(item)),
+                        item =>
```
yufeih, line 55: nit: consider extract the logic into a method like `void Foo(LogItem)` so there is less logic here. Think if there is powershell and we also need to pass an `Action<LogItem>`
```suggestion
                        WriteError,
```

PR #5941, opened 2020-05-21
Use HtmlReader in MustacheXrefTagParser

yufeih:
[AB#217885](https://dev.azure.com/ceapex/Engineering/_workitems/edit/217885)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5941)

### DIFF
--- a/src/docfx/template/MustacheXrefTagParser.cs
+++ b/src/docfx/template/MustacheXrefTagParser.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Text.RegularExpressions;
 
 namespace Microsoft.Docs.Build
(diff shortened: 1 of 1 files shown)


src/docfx/template/MustacheXrefTagParser.cs
```diff
@@ -28,37 +27,35 @@ public static string ProcessXrefTag(string templateStr)
...
-            var xrefTag = HtmlUtility.LoadHtml(match.Value).ChildNodes.FindFirst("xref");
             var uidName = "uid";
             var partialName = default(string);
-            foreach (var attribute in xrefTag.Attributes)
+            var reader = new HtmlReader(match.Value);
+            while (reader.Read(out var token))
             {
-                // TODO: uid may fallback to href in ProfileList
-                if (string.Equals(attribute.Name, "uid", StringComparison.OrdinalIgnoreCase))
+                if (token.NameIs("xref"))
```
live1206, line 35: would there be multiple `xref` nodes? This would take the last, and the previous implementation took the first.

live1206: or return when found `xref`?

yufeih: not likely, because the regex uses a non greedy match, it'll grab one `<xref>` tag at a time.

PR #5942, opened 2020-05-21
stop publishing to myget, now azdevops only

superyyrrzz:
[AB#223303](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/223303)

I do not add new badge, as Azure DevOps only supports generating badge for release version package (not pre-release).

> A badge cannot be generated until a SemVer released version (e.g. 1.2.3, not 1.2.3-beta) of docfx.console is published to the feed.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5942)

### DIFF
--- a/tools/Deployment/gulpfile.js
+++ b/tools/Deployment/gulpfile.js
@@ -14,7 +14,7 @@ format.extend(String.prototype, {})
 
 let Common = require("./out/common").Common;
 let Guard = require("./out/common").Guard;
-let Myget = require("./out/myget").Myget;
+let Nuget = require("./out/nuget").Nuget;
 let Github = require("./out/github").Github;
 let Chocolatey = require("./out/chocolatey").Chocolatey;
 let SyncBranch = require("./out/syncBranch").SyncBranch;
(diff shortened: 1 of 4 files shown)


tools/Deployment/gulpfile.js
```diff
@@ -126,30 +114,6 @@ gulp.task("publish:azdevops-ppe", () => {
     return Myget.publishToMygetAsync(artifactsFolder, process.env.NUGETEXE, "anything", config.azdevops["ppeUrl"]);
```
sharanya-rao, line 114: remove ?

sharanya-rao: Why do we have Myget.publishToMyGetAsync here ?

superyyrrzz: We need to publish to azdevops. I think the method name is confusing `publishToMygetAsync`... Let me update `myget.ts` to `nuget.ts`.

PR #5965, opened 2020-05-27
Limit info message count

yufeih:
Otherwise there will be too much info messages for `blogs-archive-pr`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5965)

### DIFF
--- a/src/docfx/lib/log/ErrorLog.cs
+++ b/src/docfx/lib/log/ErrorLog.cs
@@ -24,6 +24,7 @@ namespace Microsoft.Docs.Build
         private int _errorCount;
         private int _warningCount;
         private int _suggestionCount;
+        private int _infoCount;
 
         private int _maxExceeded;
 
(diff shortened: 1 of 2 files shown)


src/docfx/lib/log/ErrorLog.cs
```diff
@@ -243,6 +245,7 @@ private bool IncrementExceedMaxErrors(Config? config, ErrorLevel level)
                 ErrorLevel.Error => Interlocked.Increment(ref _errorCount) > (config?.MaxErrors ?? int.MaxValue),
                 ErrorLevel.Warning => Interlocked.Increment(ref _warningCount) > (config?.MaxWarnings ?? int.MaxValue),
                 ErrorLevel.Suggestion => Interlocked.Increment(ref _suggestionCount) > (config?.MaxSuggestions ?? int.MaxValue),
+                ErrorLevel.Info => Interlocked.Increment(ref _infoCount) > (config?.MaxSuggestions ?? int.MaxValue),
```
928PJY, line 248: why not using a separate config?

live1206: +1

PR #6003, opened 2020-06-01
Dev task229732: [MyGet] update feeds to new locations (DocFX v3)

byszhoubin:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6003)

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,8 +2,6 @@
 <configuration>
     <packageSources>
         <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab" />
-        <add key="docs-build-v2-prod" value="https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-build-v2-prod/nuget/v3/index.json" />
-        <add key="OpenPublishingService" value="https://pkgs.dev.azure.com/ceapex/_packaging/OpenPublishingService/nuget/v3/index.json" />
         <add key="doc-public-packages" value="https://docfx.pkgs.visualstudio.com/docfx/_packaging/docs-public-packages/nuget/v3/index.json" />
     </packageSources>
 </configuration>
--- a/NuGet.config
+++ b/NuGet.config
@@ -2,7 +2,8 @@
 <configuration>
     <packageSources>
         <add key="dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab" />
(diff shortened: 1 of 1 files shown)


byszhoubin:
The build error message:
Unable to load the service index for source https://pkgs.dev.azure.com/ceapex/_packaging/OpenPublishingService/nuget/v3/index.json.

Response status code does not indicate success: 401 (Unauthorized)

Looks like docfx v3 can't get package from OpenPublishingService feed.

@herohua Please help confirm if this is permission issue or other issue.

yufeih:
@byszhoubin as an open source project, docfx cannot depend on private NuGet feed. If you fetch the latest v3 branch, there is no `op-external` dependency now.

byszhoubin:
If there is no package dependent on op-external, I'll remove this feed from config

byszhoubin:
I removed op-external from Nuget.config, the build failed of error message:

Unable to find package Microsoft.Docs.LocaleHelper
Unable to find package Microsoft.Docs.MetadataService.Models

For package Microsoft.Docs.LocaleHelper, it's existing in OpenPublishingService feed.
For package Microsoft.Docs.MetadataService.Models, it doesn't exist in OpenPublishingService feed, I think this one still in op-external feed.

Given docfx v3 can't depend on private feed, @herohua  please help confirm if we need to move these 2 packages to public feed?

cc: @yufeih

yufeih:
I can see these are transitive dependencies from `Validation.DocFXAdapter`. We should be able to decouple them.

herohua:
@yufeih Any good suggestion? I am thinking to put those packages into `docs-public-packages`.

byszhoubin:
I just moved All versions of these 2 packages into docs-public-packages feed.

Link: 
[Microsoft.Docs.LocaleHelper](https://docfx.visualstudio.com/docfx/_packaging?_a=package&feed=docs-public-packages&package=Microsoft.Docs.LocaleHelper&protocolType=NuGet&version=2.0.19211.1)

[Microsoft.Docs.MetadataService.Models](https://docfx.visualstudio.com/docfx/_packaging?_a=package&feed=docs-public-packages&package=Microsoft.Docs.MetadataService.Models&protocolType=NuGet&version=0.3.6)

cc: @herohua @yufeih

PR #6006, opened 2020-06-02
Add identify CJK character for word count

hiihellox10:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6006)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -6,6 +6,7 @@ using System.Buffers;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Web;
 using HtmlAgilityPack;
 using Markdig.Syntax;
@@ -94,6 +95,13 @@ namespace Microsoft.Docs.Build
             { "iframe", new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "allowfullscreen", "height", "src", "width" } },
         };
 
+        private static readonly List<string> s_cjkCharUnicodeRangeList = new List<string>
+        {
+            "^[\u2E80-\u9FFF]$",    // CJK character
+            "^[\xAC00-\xD7A3]$",    // Hangul Syllables
+            "^[\uFF00-\uFFEF]$",    // Halfwidth and Fullwidth Forms(Include Chinese punctuation)
+        };
+
         public static string TransformHtml(string html, TransformHtmlDelegate transform)
         {
             var result = new ArrayBufferWriter<char>(html.Length + 64);
@@ -118,7 +126,6 @@ namespace Microsoft.Docs.Build
 
         public static void CountWord(ref HtmlToken token, ref long wordCount)
         {
-            // TODO: word count does not work for CJK locales...
             if (token.Type == HtmlTokenType.Text)
             {
                 wordCount += CountWordInText(token.RawText.Span);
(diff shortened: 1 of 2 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -395,34 +402,52 @@ private static bool IsLink(ref HtmlToken token, in HtmlAttribute attribute)
...
-            if (word)
+            if (isEnglishWord)
             {
                 total++;
             }
 
             return total;
         }
+
+        private static bool IsCJKChar(char c)
```
yufeih, line 448: This part of code is very performance sensitive (we are processing a huge number of markdown files like <https://github.com/MicrosoftDocs/azure-docs> and every character goes through this function), prefer `<` and `>` over regex.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -395,34 +402,52 @@ private static bool IsLink(ref HtmlToken token, in HtmlAttribute attribute)
...
-            var word = false;
+            var isEnglishWord = false;
 
             foreach (var ch in text)
             {
-                if (ch == ' ' || ch == '\t' || ch == '\n' || ch == '\r')
+                if (IsCJKChar(ch))
                 {
-                    if (word)
+                    total++;
```
yufeih, line 411: Add some tests in `HtmlUtilityTest.CountWord` to cover this part.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -395,34 +402,52 @@ private static bool IsLink(ref HtmlToken token, in HtmlAttribute attribute)
         private static int CountWordInText(ReadOnlySpan<char> text)
         {
             var total = 0;
-            var word = false;
+            var isEnglishWord = false;
```
yufeih, line 405: Technically this is not limited to `English`. All non CJK words fall under this category, so keep the existing name?

```suggestion
            var word = false;
```

yufeih:
@hiihellox10 Thank you for working on this. Just curious, are you working on this as a community contributor or is it part of our planed work [here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/139732). I'm asking because you are part of MSFT org (but I cannot tie GitHub account back to MSFT alias) and we just happened to start working on a couple of word count features 😊

hiihellox10:
@yufeih I'm a vendor at Alice's team, our need doc word count metadata to improve docs search api result.

yufeih:
I see and have checked with @killa1218. This might not be directly related to https://ceapex.visualstudio.com/Engineering/_workitems/edit/99452/ (Given that task is still in feature design), but it is never the less a great improvement!

src/docfx/lib/HtmlUtility.cs
hiihellox10: I have no better way than this, CJK character is so many, regex is the most commonly used.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -94,6 +95,13 @@ internal static class HtmlUtility
             { "iframe", new HashSet<string>(StringComparer.OrdinalIgnoreCase) { "allowfullscreen", "height", "src", "width" } },
         };
 
+        private static readonly List<string> s_cjkCharUnicodeRangeList = new List<string>
+        {
+            "^[\u2E80-\u9FFF]$",    // CJK character
```
yufeih, line 100: This regex basically means any character whose char code is between `\u2E80` and `\u9FFF`, so this is equivalent to `c >= '\u2E80' && c <= '\u9FFF'`

src/docfx/lib/HtmlUtility.cs
```diff
@@ -402,42 +404,38 @@ private static bool IsLink(ref HtmlToken token, in HtmlAttribute attribute)
...
                     {
-                        isEnglishWord = false;
+                        word = false;
                         total++;
                     }
                 }
                 else
                 {
-                    if (ch == ' ' || ch == '\t' || ch == '\n' || ch == '\r')
+                    if (s_wordSeparatorChars.Any(i => i == ch))
```
yufeih, line 423: Revert this change because `Enumerable.Any` allocates memory and is much slower than the old version. Same as line 431.

Please also use the same technique to implement `IsCJKChar` by inlining the range check directly in line 411

PR #6022, opened 2020-06-03
Duplicated restore should only be executed once

live1206:
[AB#230952](https://ceapex.visualstudio.com/Engineering/_workitems/edit/230952)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6022)


yufeih:
Close this PR, the root cause seems to be somewhere else.

PR #6025, opened 2020-06-04
Improve code readability for word count.

hiihellox10:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6025)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -403,34 +403,22 @@ namespace Microsoft.Docs.Build
 
             foreach (var ch in text)
             {
-                if (IsCJKChar(ch))
+                if (IsWordEndChar(ch))
                 {
-                    total++;
-
                     if (word)
                     {
                         word = false;
                         total++;
                     }
+
+                    if (IsCJKChar(ch))
+                    {
+                        total++;
+                    }
                 }
-                else
+                else if (!IsPunctuation(ch))
                 {
-                    if (ch == ' ' || ch == '\t' || ch == '\n' || ch == '\r')
-                    {
-                        if (word)
-                        {
-                            word = false;
-                            total++;
-                        }
-                    }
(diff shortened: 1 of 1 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -442,11 +430,22 @@ private static int CountWordInText(ReadOnlySpan<char> text)
             return total;
         }
 
+        private static bool IsWordEndChar(char ch)
```
yufeih, line 433: Does this change caused `IsCJKChar` to be called twice?

hiihellox10: Yes, this is a must repeat. Do you have a better way? Newly added method IsWordEndChar can make the code easier to understand.

yufeih: You can keep the current control flow and only extract methods.

BTW, I was wondering why line 418 and line 427 in the current version behaves differently. If seems to me they are all word separators. Maybe you could help merge them.

hiihellox10: This PR is improve the readability of the code and make the logic of word count statistics easier to understand. The logic has 2 stages: 1, Whether it is the end character of a word(English word). 2, Whether it is a character in a word(English word). The line 418 is stage 1, line 427 is state 2. If I keep the old control flow, it’s readable is not better than now.

yufeih: The current control flow is readable enough to me. For this particular piece of code, I would in favor of avoiding the duplicated check.

hiihellox10: It is enough now. But I like readable code, I have no good way to avoiding the duplicated check now.

PR #6033, opened 2020-06-04
Fix loc build hang due to unreleased disposable

yufeih:
[AB#230952](https://dev.azure.com/ceapex/Engineering/_workitems/edit/230952/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6033)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -41,24 +41,29 @@ namespace Microsoft.Docs.Build
 
         private static bool BuildDocset(string docsetPath, string? outputPath, CommandLineOptions options, FetchOptions fetchOptions)
         {
-            using var errorLog = new ErrorLog(outputPath);
             var stopwatch = Stopwatch.StartNew();
+
+            using var errorLog = new ErrorLog(outputPath);
+            using var disposables = new DisposableCollector();
 
             try
             {
                 var configLoader = new ConfigLoader(errorLog);
-                var (errors, config, buildOptions, packageResolver, fileResolver) = configLoader.Load(docsetPath, outputPath, options, fetchOptions);
-                if (errorLog.Write(errors))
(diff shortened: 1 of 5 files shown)


src/docfx/build/Build.cs
```diff
@@ -41,24 +41,29 @@ public static int Run(string workingDirectory, CommandLineOptions options)
...
             var stopwatch = Stopwatch.StartNew();
 
+            using var errorLog = new ErrorLog(outputPath);
+            using var disposables = new DisposableCollector();
+
             try
             {
                 var configLoader = new ConfigLoader(errorLog);
-                var (errors, config, buildOptions, packageResolver, fileResolver) = configLoader.Load(docsetPath, outputPath, options, fetchOptions);
-                if (errorLog.Write(errors))
```
yufeih, line 51: When building a _loc repo_ with _multiple docsets_ when _some docsets_ are not provisioned, this code base return here, causing `packageResolver` not disposing immediately.

In this case `packageResolver` holds a reader lock of fallback repo, and that blocks all other docsets.

src/docfx/restore/PackageResolver.cs
```diff
@@ -102,23 +102,25 @@ private void DownloadGitRepository(string url, string committish, bool depthOne,
             var gitPath = GetGitRepositoryPath(url, committish);
             var gitDocfxHead = Path.Combine(gitPath, ".git", "DOCFX_HEAD");
 
+            if (_fetchOptions == FetchOptions.UseCache && File.Exists(gitDocfxHead))
```
yufeih, line 105: Check if we can skip the download (use cache) without waiting on the lock.

PR #6035, opened 2020-06-05
support canonical version only metadata validation

OsmondJiang:
[AB#230344](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/230344)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6035)

### DIFF
--- a/src/docfx/lib/log/Error.cs
+++ b/src/docfx/lib/log/Error.cs
@@ -47,12 +47,16 @@ namespace Microsoft.Docs.Build
             Name = name;
         }
 
-        public Error WithCustomError(CustomError customError)
+        public Error WithCustomRule(CustomRule customRule, bool? isCanonicalVersion = null)
         {
+            var level = isCanonicalVersion != null && customRule.CanonicalVersionOnly
+                ? (isCanonicalVersion.Value ? customRule.Severity ?? Level : ErrorLevel.Off)
+                : customRule.Severity ?? Level;
+
             return new Error(
-                customError.Severity ?? Level,
-                string.IsNullOrEmpty(customError.Code) ? Code : customError.Code,
-                string.IsNullOrEmpty(customError.AdditionalMessage) ? Message : $"{Message}{(Message.EndsWith('.') ? "" : ".")} {customError.AdditionalMessage}",
+                level,
+                string.IsNullOrEmpty(customRule.Code) ? Code : customRule.Code,
+                string.IsNullOrEmpty(customRule.AdditionalMessage) ? Message : $"{Message}{(Message.EndsWith('.') ? "" : ".")} {customRule.AdditionalMessage}",
                 FilePath,
                 Line,
                 Column,
--- a/src/docfx/lib/log/Error.cs
+++ b/src/docfx/lib/log/Error.cs
@@ -49,9 +49,10 @@ namespace Microsoft.Docs.Build
 
         public Error WithCustomRule(CustomRule customRule, bool? isCanonicalVersion = null)
         {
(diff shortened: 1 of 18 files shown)


src/docfx/lib/log/Error.cs
```diff
@@ -47,12 +47,16 @@ public Error(ErrorLevel level, string code, string message, FilePath? file = nul
             Name = name;
         }
 
-        public Error WithCustomError(CustomError customError)
+        public Error WithCustomRule(CustomRule customRule, bool? isCanonicalVersion = null)
         {
+            var level = isCanonicalVersion != null && customRule.CanonicalVersionOnly
```
yufeih, line 52: nit: consider split it into 2 lines, one is to apply `customRule.Severity` override, the other to apply `isCanonical` override.

src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -17,34 +18,43 @@ internal class JsonSchemaValidator
         private readonly JsonSchema _schema;
         private readonly JsonSchemaDefinition _definitions;
         private readonly MicrosoftGraphAccessor? _microsoftGraphAccessor;
-        private readonly ListBuilder<(JsonSchema schema, string key, JToken value, SourceInfo? source)> _metadataBuilder;
+        private readonly ListBuilder<(JsonSchema schema, string key, JToken value, SourceInfo? source, bool? isCanonicalVersion)> _metadataBuilder;
+        private static readonly ThreadLocal<bool?> _isCanonicalVersion = new ThreadLocal<bool?>();
```
yufeih, line 22: nit: use `t_` for thread local variables to tell the different between member variables and thread static variables.
```suggestion
        private static readonly ThreadLocal<bool?> t_isCanonicalVersion = new ThreadLocal<bool?>();
```

src/docfx/lib/log/Error.cs
```diff
@@ -47,12 +47,16 @@ public Error(ErrorLevel level, string code, string message, FilePath? file = nul
             Name = name;
         }
 
-        public Error WithCustomError(CustomError customError)
+        public Error WithCustomRule(CustomRule customRule, bool? isCanonicalVersion = null)
```
yufeih, line 50: consider promote `IsCanonicalVersion` into a property (similar to `Name`) so you don't need to use a tuple [here](https://github.com/dotnet/docfx/pull/6035/files#diff-eb444f860ada42127995887c1b00a00aR519)

src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -517,13 +527,19 @@ private void PostValidateDocsetUnique(List<Error> errors)
 
             foreach (var group in validatedMetadataGroups)
             {
-                if (group.Count() > 1)
+                IEnumerable<(JsonSchema schema, string key, JToken value, SourceInfo? source, bool? isCanonicalVersion)> items = group;
+                var (metadataValue, (metadataKey, _)) = group.Key;
+                if (CanonicalVersionOnly(metadataKey, Errors.JsonSchema.DuplicateAttributeCode))
```
yufeih, line 532: how about doing the filtering be done in line 523?

src/docfx/lib/log/Error.cs
OsmondJiang: 👍

PR #6036, opened 2020-06-05
Don't restore duplicated dependencies

yufeih:
[AB#230952](https://dev.azure.com/ceapex/Engineering/_workitems/edit/230952/)

Further optimize multiple docsets restore to avoid re-fetching the same repo multiple times in a single build.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6036)

### DIFF
--- a/src/docfx/restore/FileResolver.cs
+++ b/src/docfx/restore/FileResolver.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.IO;
 using System.Net;
 using System.Net.Http;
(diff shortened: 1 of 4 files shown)


src/docfx/restore/FileResolver.cs
```diff
@@ -83,7 +87,7 @@ public string ResolveFilePath(SourceInfo<string> file)
 
         public async Task Download(SourceInfo<string> file)
         {
-            if (!UrlUtility.IsHttp(file))
+            if (!UrlUtility.IsHttp(file) || s_downloadedUrls.Contains(file))
```
live1206, line 90: There is still chance to download more than once, right?

yufeih: yes, tiny.

live1206: ok, fair trade off to achieve better perf without lock

PR #6044, opened 2020-06-08
231473 image render fix

adunndevster:
I consolidated the ImageExtension code @yufeih !

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6044)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
```diff
@@ -94,23 +94,30 @@ public bool TryProcessAttributes(IDictionary<string, string> attributes, out Htm
             {
                 htmlAttributes.AddProperty("alt", alt);
             }
-            var id = GetHtmlId(block);
+            var id = GetHtmlId(markdownObject);
             if(type == "complex") htmlAttributes.AddProperty("aria-describedby", id);
 
             RenderDelegate = (renderer, obj) =>
             {
+                dynamic blockOrInline;
```
yufeih, line 102: Is there a way to not using `dynamic` here? `dynamic` is likely to hurt the performance of this method: https://stackoverflow.com/questions/7478387/how-does-having-a-dynamic-variable-affect-performance

adunndevster: Does it look better now @yufeih ?

yufeih: Yep, much better. There are two tests failing though.

adunndevster: @yufeih - It looks like those tests are passing now. I noticed the regression diff tests (is that what you call them?) were not passing so I downloaded one of the diff files, and it had lots of this:
image

That is now what the image output should be, so we should be good to merge this, yes?

yufeih: yes

PR #6046, opened 2020-06-09
support canonical-version-only content validation

OsmondJiang:
[AB#230341](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/230341)

Now it's only support combined versioning, will support splited versioning later

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6046)

### DIFF
--- a/src/docfx/lib/markdown/MarkdownVisitContext.cs
+++ b/src/docfx/lib/markdown/MarkdownVisitContext.cs
@@ -20,7 +20,7 @@ namespace Microsoft.Docs.Build
 
         public bool IsInclude => FileStack.Count >= 2;
 
-        public MonikerList? ZoneMoniker => ZoneMonikerStack.Count >= 1 ? ZoneMonikerStack.Peek() : default;
+        public MonikerList ZoneMoniker => ZoneMonikerStack.Count >= 1 ? ZoneMonikerStack.Peek() : default;
 
         public string? Zone => ZoneStack.Count >= 1 ? ZoneStack.Peek() : default;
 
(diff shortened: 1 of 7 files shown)


src/docfx/lib/markdown/MarkdownVisitContext.cs
```diff
@@ -16,9 +18,12 @@ internal class MarkdownVisitContext
 
         public bool IsInclude => FileStack.Count >= 2;
 
+        public MonikerList ZoneMoniker => ZoneMonikerStack.Count >= 1 ? ZoneMonikerStack.Peek() : default;
```
yufeih, line 21: nit: here is now a `TryPeak` method

```suggestion
        public MonikerList ZoneMoniker => ZoneMonikerStack.TryPeek(out var top) ? top : default;
```

src/docfx/lib/markdown/MarkdownVisitContext.cs
```diff
@@ -8,6 +8,8 @@ namespace Microsoft.Docs.Build
 {
     internal class MarkdownVisitContext
     {
+        public Stack<MonikerList> ZoneMonikerStack { get; set; }
```
yufeih, line 11: nit: consider make these 2 stacks private (and pass it through constructor)? (so `Visit` callback cannot see see / change them)

src/docfx/lib/markdown/validation/DocsValidationExtension.cs
```diff
@@ -77,6 +87,28 @@ internal static class DocsValidationExtension
             });
         }
 
+        private static bool? IsCanonicalVersion(MonikerList pageLevelMonikerList, MonikerList fileLevelMonikerList, MonikerList zoneLevelMonikerList)
```
yufeih, line 90: Consider move this as a common method in `MonikerList`

PR #6049, opened 2020-06-09
Support run ecma2yaml multiple times

mingwli:
[AB#233975](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/233975)

1. make `ECMA2Yaml` and `SourceMap` array type
2. pass `SourceMap` to docfx with array value, and merge them in `SourceMap.cs`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6049)

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -287,7 +287,8 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// A file containing a map of file path to the original file path.
         /// </summary>
-        public SourceInfo<string> SourceMap { get; private set; } = new SourceInfo<string>("");
+        [JsonConverter(typeof(OneOrManyConverter))]
+        public SourceInfo<string>[] SourceMap { get; private set; } = Array.Empty<SourceInfo<string>>();
 
         /// <summary>
         /// Determines if validate the moniker configuration.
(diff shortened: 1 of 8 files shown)


src/docfx/config/Config.cs
```diff
@@ -302,7 +303,7 @@ internal class Config : PreloadConfig
         /// <summary>
         /// Determines and configures build to consume XML files produced from monodoc
         /// </summary>
-        public ECMA2YamlRepoConfig? Monodoc { get; private set; }
+        public ECMA2YamlRepoConfig[]? Monodoc { get; private set; }
```
yufeih, line 306: `OneOrManyConverter`?

src/docfx/config/ops/OpsPreProcessor.cs
```diff
@@ -37,28 +37,32 @@ private void PreProcessMonoDocXml()
             {
                 lock (s_lock)
                 {
-                    if (Directory.Exists(_config.Monodoc.OutputYamlFolder))
+                    for (var idx = 0; idx < _config.Monodoc.Length; idx++)
```
yufeih, line 40: nit: spell checker bot does not know `idx`: `i` or `index`?

src/docfx/config/Config.cs
mingwli: Cool, ideally it should be.
(I thought that currently this value was handled by `OpsConfigAdapter`, and it's definitely an `array`...)

PR #6052, opened 2020-06-10
Add TOC and breadcrumb validation

lamebrain:
Integrate validations into existing pipelines for the following:
* toc-missing
* toc-deprecated
* toc-entry-duplicated
* toc-breadcrumb-link-external

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6052)

### DIFF
--- a/src/docfx/build/toc/BuildTableOfContents.cs
+++ b/src/docfx/build/toc/BuildTableOfContents.cs
@@ -31,13 +31,13 @@ namespace Microsoft.Docs.Build
             var (monikerErrors, monikers) = context.MonikerProvider.GetFileLevelMonikers(file.FilePath);
             errors.AddRange(monikerErrors);
 
-            var outputPath = context.DocumentProvider.GetOutputPath(file.FilePath);
-
-            var jsonOutputPath = LegacyUtility.ChangeExtension(context.DocumentProvider.GetOutputPath(file.FilePath, true), ".json");
-            if (string.Compare(jsonOutputPath, metadata.BreadcrumbPath, System.StringComparison.OrdinalIgnoreCase) == 0)
+            var outputUrl = context.DocumentProvider.GetDocsSiteUrl(file.FilePath);
+            if (string.Compare(outputUrl, metadata.BreadcrumbPath, System.StringComparison.OrdinalIgnoreCase) == 0)
             {
                 context.ContentValidator.ValidateBreadcrumbLinkExternal(model.Items, file.FilePath);
             }
+
+            var outputPath = context.DocumentProvider.GetOutputPath(file.FilePath);
 
             // enable pdf
             if (context.Config.OutputPdf)
--- a/src/docfx/build/toc/BuildTableOfContents.cs
+++ b/src/docfx/build/toc/BuildTableOfContents.cs
@@ -31,12 +31,6 @@ namespace Microsoft.Docs.Build
             var (monikerErrors, monikers) = context.MonikerProvider.GetFileLevelMonikers(file.FilePath);
             errors.AddRange(monikerErrors);
 
-            var outputUrl = context.DocumentProvider.GetDocsSiteUrl(file.FilePath);
-            if (string.Compare(outputUrl, metadata.BreadcrumbPath, System.StringComparison.OrdinalIgnoreCase) == 0)
-            {
-                context.ContentValidator.ValidateBreadcrumbLinkExternal(model.Items, file.FilePath);
-            }
-
             var outputPath = context.DocumentProvider.GetOutputPath(file.FilePath);
 
             // enable pdf
--- a/src/docfx/build/toc/BuildTableOfContents.cs
+++ b/src/docfx/build/toc/BuildTableOfContents.cs
@@ -15,6 +15,8 @@ namespace Microsoft.Docs.Build
             // load toc tree
             var (errors, node, _, _) = context.TableOfContentsLoader.Load(file);
 
+            context.ContentValidator.ValidateTocDeprecated(file.FilePath); // toc-deprecated
+
             var (metadataErrors, metadata) = context.MetadataProvider.GetMetadata(file.FilePath);
             errors.AddRange(metadataErrors);
 
@@ -30,6 +32,13 @@ namespace Microsoft.Docs.Build
             errors.AddRange(monikerErrors);
 
             var outputPath = context.DocumentProvider.GetOutputPath(file.FilePath);
+
+            var jsonOutputPath = LegacyUtility.ChangeExtension(context.DocumentProvider.GetOutputPath(file.FilePath, true), ".json");
+
+            if (string.Compare(jsonOutputPath, metadata.BreadcrumbPath, System.StringComparison.OrdinalIgnoreCase) == 0)
+            {
+                context.ContentValidator.ValidateBreadcrumbLinkExternal(model.Items, file.FilePath, metadata.BreadcrumbPath); // toc-breadcrumb-link-external
+            }
 
             // enable pdf
             if (context.Config.OutputPdf)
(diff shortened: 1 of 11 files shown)


docs/specs/validation.yml
```diff
@@ -1568,3 +1568,144 @@ outputs:
...
+            "type": "TocDeprecated",
+            "message": "Markdown TOC files are deprecated. Convert to a YAML TOC.",
+            "code": "toc-deprecated",
+            "severity": "SUGGESTION",
+            "exclusions": ["includes", "landingpage", "hubpage", "conceptual"]
+          }
+        ]
+      }
+    }
+  a/toc.md:
```
yufeih, line 1592: The linux test failed because it is expecting an upper case `TOC.md`.

```suggestion
  a/TOC.md:
```

src/docfx/validation/ContentValidator.cs
```diff
@@ -43,6 +44,89 @@ public void ValidateManifest(FilePath filePath, PublishItem publishItem)
...
+            if (TryGetValidationDocumentType(ContentType.TableOfContents, string.Empty, false, out var documentType))
+            {
+                var validationContext = new ValidationContext { DocumentType = documentType };
+                foreach (var node in nodes)
+                {
+                    var tocItem = new TocItem()
+                    {
+                        FilePath = node.Href.Value,
+                        IsHrefExternal = UrlUtility.GetLinkType(node.Href) == LinkType.External,
+                        SourceInfo = new SourceInfo(filePath, 0, 0),
```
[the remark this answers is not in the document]
OsmondJiang: if it's validate breadcrumb, do we need consider the `TopicHref` and `TocHref`?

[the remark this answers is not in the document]
OsmondJiang: got, that's means the resolved toc model has mapped the `topichref` and `tochref` to `href`

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -29,6 +31,12 @@ public static void Build(Context context, Document file)
 
             var outputPath = context.DocumentProvider.GetOutputPath(file.FilePath);
 
+            var jsonOutputPath = LegacyUtility.ChangeExtension(context.DocumentProvider.GetOutputPath(file.FilePath, true), ".json");
```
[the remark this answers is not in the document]
OsmondJiang: if the file with versioning, the output path would contains group hash value like `2420f88b/docs/toc.json`, @yufeih  would you double confirm?

src/docfx/validation/ContentValidator.cs
```diff
@@ -43,6 +44,89 @@ public void ValidateManifest(FilePath filePath, PublishItem publishItem)
...
+                    };
+                    Write(_validator.ValidateToc(tocItem, validationContext).GetAwaiter().GetResult());
```
yufeih, line 89: If I understand correctly , the `ValidateToc` call here looks through all the TOC validators [here](https://dev.azure.com/ceapex/Engineering/_git/Docs.Validation?path=%2FValidation%2FValidator.cs&version=GBdevelop&line=65&lineEnd=65&lineStartColumn=39&lineEndColumn=67&lineStyle=plain), and so even if this method is called `ValidateBreadcrumbLinkExternal`, it'll still execute all 4 TOC validators against each TOC node. If so, there are a couple of problems I see with this design:

1. It is very easy to introduce bugs. The current tests all checks against a single rule, if there are multiple rules turned on at the same time, e.g., if `TocMissing` and `TocBreadcrumbLinkExternal` are both turned on, will there be a lot of `TocMissing` errors for breadcrumb since the `ValidateTocMissing` validator is using the [HasReferencedTocs](https://dev.azure.com/ceapex/Engineering/_git/Docs.Validation?path=%2FValidation%2FValidateToc%2FValidateTocMissing.cs&version=GBdevelop&line=20&lineEnd=20&lineStartColumn=26&lineEndColumn=43&lineStyle=plain) property for detection, and each breadcrumb TOC node does not set `HasReferencedTocs`?

2. The performance is sub-optimal.

src/docfx/build/toc/TableOfContentsLoader.cs
```diff
@@ -79,6 +82,11 @@ internal class TableOfContentsLoader
 
                 var node = _parser.Parse(file.FilePath, errors);
                 node.Items = LoadTocNodes(node.Items, file, rootPath, referencedFiles, referencedTocs, errors);
+
+                if (file == rootPath)
+                {
+                    _contentValidator.ValidateTocEntryDuplicated(file, node);
```
yufeih, line 88: The `referencedFiles` property contains all the markdown files referenced by this TOC, could use that over traversing the TOC nodes again. And also can change the signature to `List<SourceInfo<Document>>` (and also update insertion logic) to retrieve exactly where the source file is referenced in this TOC file.

src/docfx/validation/ContentValidator.cs
OsmondJiang: I have changed the interface, now it's only run one rule for each toc model type.

src/docfx/build/toc/BuildTableOfContents.cs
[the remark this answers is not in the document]
lamebrain: Changed to GetDocsSiteUrl

src/docfx/build/toc/BuildTableOfContents.cs
```diff
@@ -29,6 +31,12 @@ public static void Build(Context context, Document file)
...
+            if (string.Compare(jsonOutputPath, metadata.BreadcrumbPath, System.StringComparison.OrdinalIgnoreCase) == 0)
```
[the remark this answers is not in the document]
lamebrain: Yes, good catch. I think I mixed up validations with a different rule. I added the check in `LoadTocNode` since `GetHrefType` is static. Please verify.

src/docfx/build/toc/TableOfContentsLoader.cs
lamebrain: This is much better, thanks for the tip. I decided not to add the source info because there could potentially be multiple nodes in the same file that could be duplicated and the source info wouldn't be of much help in that case. The node file names are output in the error for authors to find and fix the problem(s).

docs/specs/validation.yml
```diff
@@ -1568,3 +1568,144 @@ outputs:
     {"message_severity":"suggestion","log_item_type":"user","code":"duplicate-h1","message":"H1 'h1' is duplicated with other articles: 'docs/v1/b.md, docs/v1/token1.md, docs/v1/token2.md(5,1)'","file":"docs/v1/token2.md","line":5,"end_line":5,"column":1,"end_column":1}
   2420f88b/docs/a.json:
   2420f88b/docs/b.json:
+---
```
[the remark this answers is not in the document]
lamebrain: good idea. I split out the toc and breadcrumb validations. @OsmondJiang do you want to split out the remaining validations once this is merged?

src/docfx/docfx.csproj
```diff
@@ -14,7 +14,7 @@
   <ItemGroup>
     <PackageReference Include="Azure.Identity" Version="1.2.0-preview.1" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.0.3" />
-    <PackageReference Include="docs.validation" Version="1.0.0-beta-00129-f257836" />
+    <PackageReference Include="docs.validation" Version="1.0.0-beta-00141-21b1204" />
```
lamebrain, line 17: @OsmondJiang I noticed that we are referencing packages from the Docs.Validation `develop` branch and it is ahead of `master`, even after release. Is there a release process to merge with master or should we consolidate the branches?

OsmondJiang: Docs.Validation is still under fast development process, we currently using the package version released from developer branch directly in docfx v3 which is fast and convenient, later once we are stable, we will use master to control the release process, good point!

PR #6053, opened 2020-06-10
Get canonical version for publish URL

live1206:
[AB#230340](https://dev.azure.com/ceapex/Engineering/_workitems/edit/230340)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6053)

PR #6058, opened 2020-06-11
Fix moniker order

live1206:
Fix for #6053 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6058)

### DIFF
--- a/src/docfx/build/moniker/MonikerProvider.cs
+++ b/src/docfx/build/moniker/MonikerProvider.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Docs.Build
             {
                 return value;
             }
-            return int.MinValue;
+            return 0;
         }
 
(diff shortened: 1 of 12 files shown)


src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -49,7 +49,7 @@ public int GetMonikerOrder(string moniker)
             {
                 return value;
             }
-            return int.MinValue;
+            return 0;
```
live1206, line 52: Since `order` in optional, 0 is default as lowest order.
https://github.com/dotnet/docfx/blob/5b882493aa776dc37106cb6466b6178a81a7196c/src/docfx/lib/moniker/Moniker.cs#L17

PR #6059, opened 2020-06-11
fix bug of invalid monikers

OsmondJiang:
invalid moniker should be considered as `isCanonicalVersion = false`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6059)

### DIFF
--- a/docs/specs/validation.yml
+++ b/docs/specs/validation.yml
@@ -1657,10 +1657,9 @@ inputs:
     }
 outputs:
   .errors.log: |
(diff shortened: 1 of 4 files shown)


yufeih:
Close this as it is a very corner case.

OsmondJiang:
close this pr because we think it's the existing error has already told user that the moniker zone is invalid

PR #6060, opened 2020-06-11
Clarify dependencies for context

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6060)

### DIFF
--- a/src/docfx/build/context/Context.cs
+++ b/src/docfx/build/context/Context.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Net.Mail;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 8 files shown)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -33,47 +33,72 @@ public JsonSchemaTransformer(JsonSchema schema)
...
             var uidCount = _uidCountCache.GetOrAdd(file, GetFileUidCount(_schema, token));
-            return TransformContentCore(file, context, _schema, token, uidCount);
+            return TransformContentCore(file, _schema, token, uidCount, context.MarkdownEngine, context.LinkResolver, context.XrefResolver);
         }
 
-        public (List<Error>, IReadOnlyList<InternalXrefSpec>) LoadXrefSpecs(Document file, Context context, JToken token)
+        public (List<Error>, IReadOnlyList<InternalXrefSpec>) LoadXrefSpecs(
+            Document file,
+            JToken token,
+            MarkdownEngine markdownEngine,
```
yufeih, line 42: Move these services to the constructor of `JsonSchemaTransformer`?

live1206: I could give it a try, ideally we should not pass around the component instances, but `JsonSchemaTransformer` is not hold in `context`.

src/docfx/build/context/Context.cs
```diff
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Net.Mail;
```
yufeih, line 5:
```suggestion
~~using System.Net.Mail;~~
```

PR #6061, opened 2020-06-11
🐛🐛🐛 fix bug of conceptual layout

OsmondJiang:
accidentally find a bug when debugging sql-docs-pr repo, layout is added as global metdata in sql-docs-pr 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6061)

### DIFF
--- a/src/docfx/validation/MetadataValidator.cs
+++ b/src/docfx/validation/MetadataValidator.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Docs.Build
             {
                 // Only validate conceptual files
                 if (contentType == ContentType.Page && mime == "Conceptual" &&
-                    (!metadata.ContainsKey("layout") || string.Equals(metadata.Value<string>(), "conceptual", StringComparison.OrdinalIgnoreCase)))
+                    (!metadata.ContainsKey("layout") || string.Equals(metadata.Value<string>("layout"), "conceptual", StringComparison.OrdinalIgnoreCase)))
                 {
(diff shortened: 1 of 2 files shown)


OsmondJiang:
image

src/docfx/validation/MetadataValidator.cs
```diff
@@ -76,7 +76,8 @@ public List<Error> ValidateMetadata(JObject metadata, FilePath filePath)
             foreach (var schemaValidator in _schemaValidators)
             {
                 // Only validate conceptual files
-                if (contentType == ContentType.Page && mime == "Conceptual" && !metadata.ContainsKey("layout"))
+                if (contentType == ContentType.Page && mime == "Conceptual" &&
+                    (!metadata.ContainsKey("layout") || string.Equals(metadata.Value<string>(), "conceptual", StringComparison.OrdinalIgnoreCase)))
```
yufeih: `Value<string>` will throw if the value is not a string?

OsmondJiang: good catch!

OsmondJiang:
These diff are expected, take Dynamics-365-Operations as example, there is a global metadata in docfx.jon "layout: Conceptual". So the diff are caused by these markdown file finally getting validated. 😎

PR #6072, opened 2020-06-14
document --enable-local-file-access option

Blake-Eryx:
Starting with version 0.12.6, wkhtmlpdf blocks local file access by default. This means that when generating PDF, if you are running on a local file system then the JS and CSS files will be blocked (probably among other files) and the output will look bare bones. Passing "--enable-local-file-access" as an additional option corrects this. This PR documents this in a couple of places.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6072)

### DIFF
--- a/Documentation/tutorial/docfx.exe_user_manual.md
+++ b/Documentation/tutorial/docfx.exe_user_manual.md
@@ -544,7 +544,7 @@ noStdin                  | Do not use `--read-args-from-stdin` for the wkhtmltop
 Key                      | Description
 -------------------------|-----------------------------
 filePath                 | The path and file name of a wkhtmltopdf.exe compatible executable.
-additionalArguments      | Additional arguments that should be passed to the wkhtmltopdf executable.
+additionalArguments      | Additional arguments that should be passed to the wkhtmltopdf executable. For example, pass `--enable-local-file-access` if you are building on a local file system. This will ensure that the supporting *.js and *.css files are loaded when rendering the HTML being converted to PDF.
 
 ## 4. Supported File Mapping Format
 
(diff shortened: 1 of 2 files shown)


Documentation/tutorial/docfx.exe_user_manual.md
```diff
@@ -544,7 +544,7 @@ noStdin                  | Do not use `--read-args-from-stdin` for the wkhtmltop
 Key                      | Description
 -------------------------|-----------------------------
 filePath                 | The path and file name of a wkhtmltopdf.exe compatible executable.
-additionalArguments      | Additional arguments that should be passed to the wkhtmltopdf executable.
+additionalArguments      | Additional arguments that should be passed to the wkhtmltopdf executable. For example, pass `--enable-local-file-access` if you are building on a local file system. This will ensure that the supporting *.js and *.css files are loaded when rendering the HTML being converted to PDF.
```
superyyrrzz, line 547: Should we mention `version >= 0.12.6`, or do you think it is a general suggestion regardless of wkhtmltopdf version?

Blake-Eryx: I think just as a general suggestion would be good, especially since this emulates the behavior prior to 0.12.6.

Blake-Eryx: So, no harm in passing this value if someone is using an older wkhtmlpdf.

superyyrrzz:
@Blake-Eryx Thank you for contribution!

PR #6081, opened 2020-06-15
Mustache <xref> with inner template content

mingwli:
[AB#233985](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/233985)

Found only one sample on prod: https://developer.microsoft.com/en-us/advocates/
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6081)

PR #6083, opened 2020-06-16
Mustache <xref> with inner template content

mingwli:
Another solution for [AB#233985](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/233985)

See #6081
using condition to control opening/closing `<a>` `<span>`, instead of copy the whole middle-stuff.

Keeping self-closing `<xref/>` logic as before for better readability.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6083)

### DIFF
--- a/test/docfx.Test/template/MustacheXrefTagParserTest.cs
+++ b/test/docfx.Test/template/MustacheXrefTagParserTest.cs
@@ -71,10 +71,17 @@ namespace Microsoft.Docs.Build
             "</xref>",
             "{{#uid-from-href}}" +
             "  {{#href}}" +
-            "    <a href='{{href}}' title='{{name}}'> <h3>{{name}}</h3> </a>" +
+            "    <a href='{{href}}' title='{{name}}'>" +
             "  {{/href}}" +
             "  {{^href}}" +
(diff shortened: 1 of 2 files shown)


test/docfx.Test/template/MustacheXrefTagParserTest.cs
```diff
@@ -50,12 +51,60 @@ public class MustacheXrefTagParserTest
...
             )]
+        [InlineData("<xref uid='{{ . }}' title='{{title}}'/>",
+            "{{#href}}" +
+            "  <a href='{{href}}' title='{{title}}'> {{name}} </a>" +
+            "{{/href}}" +
+            "{{^href}}" +
+            "  <span> {{name}} </span>" +
+            "{{/href}}"
+            )]
+        [InlineData(
```
mingwli, line 68: #6081 transformed result reads easier to me, while its process logic maybe not that straight-forward :/

PR #6084, opened 2020-06-16
Remove triple colon extension RenderDelegate

yufeih:
[AB#226125](https://ceapex.visualstudio.com/Engineering/_workitems/edit/226125?src=WorkItemMention&src-action=artifact_link)

Removes `RenderDelegate` pattern used in triple colon extensions.

Each `ITripleColonExtensionInfo` instance is a global singleton instance. The method that it implements will be called by multiple threads at the same time, so it _MUST_ be thread safe and _pure_.

The current pattern of using `RenderDelegate` makes the `Render` function not thread safe. As the `RenderDelegate` can accidentally capture local variables in `TryProcessAttributes` method, the delegate can store states of one article and be used to render a different article with a wrong state. This causes unstable output and other bugs.

In short, the principle is that `ITripleColonExtensionInfo` derived classes _MUST NOT_ contain fields with states.

@adunndevster 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6084)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/ImageExtension.cs
```diff
@@ -90,93 +81,88 @@ public bool TryProcessAttributes(IDictionary<string, string> attributes, out Htm
...
-                    renderer.Write("<a").WriteAttributes(lightboxHtmlAttributes).WriteLine(">");
-                }
-                if(currentType != "complex")
-                {
-                    renderer.Write("<img").WriteAttributes(obj).WriteLine(">");
-                } else
-                {
-                    if (currentType == "complex" && tripleColonObj.Count == 0)
-                    {
-                        logWarning("If type is \"complex\", then descriptive content is required. Please make sure you have descriptive content.");
```
yufeih, line 161: This is an example: though this line seems stateless, `logWarning` implicitly captures [processor.LineIndex](https://github.com/dotnet/docfx/pull/6084/files#diff-1df2207e9d7ec9145d28ccc0e789985dL49) that is tied to the current processor state. When `Render` is called on a different block, there is a chance that `Line` info is mis-reported.

adunndevster: Got it! Thank you for doing this, it looks much better.

PR #6118, opened 2020-06-21
Add option to toggle including the TOC page in PDF output

Blake-Eryx:
A TOC page is always included in the PDF output and there is no way to turn that off.  This PR adds a new option "includeDefaultToc" (defaults to true, to maintain current behavior) that can toggle including or excluding this page in the PDF output.

This feature has been requested in the comments of the main DocFx page, and also I would personally prefer using wkhtmlpdf's built-in toc option instead (via the additional arguments option).

P.S. I call it "DefaultToc" since you could add a TOC yourself via wkhtmlpdf.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6118)

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.js
+++ b/src/docfx.website.themes/default/styles/docfx.js
@@ -422,6 +422,8 @@ $(function () {
       $('#toc a.active').parents('li').each(function (i, e) {
         $(e).addClass(active).addClass(expanded);
         $(e).children('a').addClass(active);
+      })
+      $('#toc a.active').parents('li').each(function (i, e) {
         top += $(e).position().top;
       })
       $('.sidetoc').scrollTop(top - 50);


src/docfx/Models/PdfJsonConfig.cs
```diff
@@ -29,6 +29,9 @@ public class PdfJsonConfig : BuildJsonConfig
         [JsonProperty("keepRawFiles")]
         public bool KeepRawFiles { get; set; }
 
+        [JsonProperty("includeDefaultToc")]
+        public bool IncludeDefaultToc { get; set; }
```
superyyrrzz, line 33: To keep backward compatibility, set default value of this to `true`? The default value is used if neither `docfx.json` nor CLI options sets it.

Or, you can also change it to `ExcludeDefaultToc`, the we need not to explicitly set the default value as it is `false` by default.

PR #6124, opened 2020-06-22
Micro-optimize loading git commit history

yufeih:
[AB#240717](https://dev.azure.com/ceapex/Engineering/_workitems/edit/240717)

Reduces `azure-docs-pr` build time (without commit history) from 7:30 to 4:30 (useful for non-master branch builds, telemetry shows there is a significant amount of non-master branch builds, they can be slow due to the current LRU cache mechanism, which will be addressed after this PR)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6124)

### DIFF
--- a/src/docfx/lib/git/FileCommitProvider.cs
+++ b/src/docfx/lib/git/FileCommitProvider.cs
@@ -7,8 +7,10 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Collections.Extensions;
 
 using static Microsoft.Docs.Build.LibGit2;
 
@@ -16,6 +18,8 @@ namespace Microsoft.Docs.Build
 {
     internal sealed class FileCommitProvider : IDisposable
     {
+        private static readonly DictionarySlim<uint, Tree> s_emptyTree = new DictionarySlim<uint, Tree>();
+
         private readonly Repository _repository;
         private readonly Lazy<GitCommitCache> _commitCache;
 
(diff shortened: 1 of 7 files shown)


src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -193,145 +192,162 @@ public void Dispose()
...
         {
-            var blob = treeId;
+            var node = tree;
 
             for (var i = 0; i < pathSegments.Length; i++)
             {
-                var files = _trees.GetOrAdd(blob.a, _ => LoadTree(blob));
-                if (files is null || !files.TryGetValue(pathSegments[i], out blob))
+                var children = node.Children ?? LoadChildren(node);
+                if (ReferenceEquals(children, s_emptyTree) || !children.TryGetValue(pathSegments[i], out node))
```
live1206, line 283: if `s_emptyTree` is only used for checking no children, why not just `children.Count == 0`?

src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -193,145 +192,162 @@ public void Dispose()
...
                 GitCommit = gitCommit;
+                Parents = new Commit[parentIds.Length];
             }
         }
+
+        private class Tree
+        {
+            public git_oid Id { get; set; }
+
+            public DictionarySlim<uint, Tree>? Children { get; set; }
```
live1206, line 350: don't quite follow the reduction of dictionary lookup here... 
Seems like it needed 1 dictionary lookup before and 1 now

src/docfx/lib/git/FileCommitProvider.cs
yufeih: yeah, both are fine

PR #6134, opened 2020-06-24
Fix Code Style Violation of Rule MEN002(Line too long)

v-pegao:

Changes:
- Suppress MEN002(line too long) for ChakraHost.Hosting in GlobalSuppressions.cs
- Suppress SA1404(Code analysis suppression should have justification) and remove justification in GlobalSuppressions.cs
- Fix some MEN002(line too long) 

Fix for MEN002 is follow below rules:
- If it is caused by too many nested brackets, try turn local functions into member functions or use early return
- If it is long parameter, try put parameters to a new line, if it is still too long, break each parameter into its own line
- If parameter is a long expression, try extract a variable
- For long member variable declarations with initialization, try put the default initializer to a new line
- If the return value is a tuple ant the tuple definition is long, try put the return value definition in its own line
- Suppress this warning in Errors.cs

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6134)


src/docfx/restore/Restore.cs
```diff
@@ -43,7 +43,8 @@ public static bool RestoreDocset(string docsetPath, string? outputPath, CommandL
             {
                 // load configuration from current entry or fallback repository
                 var configLoader = new ConfigLoader(errorLog);
-                var (errors, config, buildOptions, packageResolver, fileResolver) = configLoader.Load(disposables, docsetPath, outputPath, options, fetchOptions);
+                var (errors, config, buildOptions, packageResolver, fileResolver) = configLoader.Load(
```
live1206, line 46: How about putting `configLoader.Load(` into the next line?

v-pegao: The change is follow the rule "If it is long parameter, try put parameters to a new line, if it is still too long, break each parameter into its own line", should I update the rule?

live1206: I guess the rule is specific for method definition?
Take this [change](https://github.com/dotnet/docfx/pull/6134/files#diff-0b7306773d16a50c3dae5a64c992125aL146-L147) for instance, assignment might be a different story

v-pegao: The scope is not specified for the rule, add @yufeih to discuss the change for the case.

v-pegao:
Will category the changes and send the new pull requests, which will make pull request easy to review

PR #6139, opened 2020-06-28
Fix Code Style Violation of Rule MEN002

v-pegao:

Changes:
- Suppress SA1404(Code analysis suppression should have justification) and remove justification in GlobalSuppressions.cs
- Suppress MEN002(line too long) for ChakraHost.Hosting in GlobalSuppressions.cs
- Suppress MEN002 in attribute of class LibGit2
- Suppress violation with attribute instead of `#pragma warning disable {rule}`
  - There are removed suppression without suppress attribute being added, because there is no violation for that rule.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6139)


yufeih:
@v-pegao , I've updated the max line columns to 160, this should reduce the places to update

PR #6140, opened 2020-06-28
Avoid constructing large string exceeding the limit size

live1206:
[AB#241219](https://dev.azure.com/ceapex/Engineering/_workitems/edit/241219)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6140)

### DIFF
--- a/src/docfx/build/context/Output.cs
+++ b/src/docfx/build/context/Output.cs
@@ -84,7 +84,7 @@ namespace Microsoft.Docs.Build
             _queue.Completion.Wait();
         }
 
-        private string GetDestinationPath(string destRelativePath)
+        public string GetDestinationPath(string destRelativePath)
         {
             Debug.Assert(!Path.IsPathRooted(destRelativePath));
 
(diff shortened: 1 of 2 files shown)


src/docfx/build/context/Output.cs
```diff
@@ -84,7 +84,7 @@ public void WaitForCompletion()
             _queue.Completion.Wait();
         }
 
-        private string GetDestinationPath(string destRelativePath)
+        public string GetDestinationPath(string destRelativePath)
```
yufeih, line 87: Rather than make this method public, prefer a `WriteLines` method to:
1. Allow encapsulation of async write operations using a private queue
2. Restrict file operations inside `Input` and `Output` class so we can swap the implementation without caller awareness.

live1206: sounds good

PR #6165, opened 2020-07-01
Fix Code Violation MEN002(Line too long)

v-pegao:

Fix MEN002 for:
- Method Definition:
  - If it is long parameter, try put parameters to a new line, if it is still too long, break each parameter into its own line
- Member Variable Declarations:
  - For long member variable declarations with initialization, try put the default initializer to a new line
  - If it is long parameter, try put parameters to a new line, if it is still too long, break each parameter into its own line 
- Long return:
  - If the return value is a tuple and the tuple definition is long, try put the return value definition in its own line
- Long if conditions and other expressions

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6165)

### DIFF
--- a/src/docfx/build/document/DocumentProvider.cs
+++ b/src/docfx/build/document/DocumentProvider.cs
@@ -46,7 +46,8 @@ namespace Microsoft.Docs.Build
         {
             var file = GetDocument(path);
             var outputPath = file.SitePath;
-            if ((file.ContentType == ContentType.Page && file.IsPage) || file.ContentType == ContentType.Redirection || file.ContentType == ContentType.TableOfContents)
+            if ((file.ContentType == ContentType.Page && file.IsPage) ||
+                file.ContentType == ContentType.Redirection || file.ContentType == ContentType.TableOfContents)
             {
                 var fileExtension = _config.Legacy && file.IsPage
                     ? ".raw.page.json"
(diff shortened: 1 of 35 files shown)


src/docfx/build/document/DocumentProvider.cs
```diff
@@ -46,7 +46,8 @@ public string GetOutputPath(FilePath path)
         {
             var file = GetDocument(path);
             var outputPath = file.SitePath;
-            if ((file.ContentType == ContentType.Page && file.IsPage) || file.ContentType == ContentType.Redirection || file.ContentType == ContentType.TableOfContents)
+            if ((file.ContentType == ContentType.Page && file.IsPage) ||
+                file.ContentType == ContentType.Redirection || file.ContentType == ContentType.TableOfContents)
```
yufeih, line 50: Try break every `||` for `if`

src/docfx/build/xref/InternalXrefSpec.cs
```diff
@@ -32,7 +32,8 @@ public InternalXrefSpec(SourceInfo<string> uid, string href, Document declaringF
 
         public string? GetXrefPropertyValueAsString(string propertyName)
         {
-            return XrefProperties.TryGetValue(propertyName, out var property) && property.Value is JValue propertyValue && propertyValue.Value is string internalStr ? internalStr : null;
+            return XrefProperties.TryGetValue(propertyName, out var property) &&
+                   property.Value is JValue propertyValue && propertyValue.Value is string internalStr ? internalStr : null;
```
yufeih, line 36: try break at `?`

PR #6177, opened 2020-07-02
Add RemoveHostName switch to config, set it from env

live1206:
- `RemoveHostName` as switch to remove host
- redirection URL should always remove host
- set it from env, which is set from server side

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6177)

### DIFF
--- a/src/docfx/config/EnvironmentVariable.cs
+++ b/src/docfx/config/EnvironmentVariable.cs
@@ -19,8 +19,6 @@ namespace Microsoft.Docs.Build
 
         public static string? CorrelationId => GetValue("DOCFX_CORRELATION_ID");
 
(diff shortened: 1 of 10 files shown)


src/docfx/config/EnvironmentVariable.cs
```diff
@@ -19,6 +19,8 @@ public static class EnvironmentVariable
 
         public static string? CorrelationId => GetValue("DOCFX_CORRELATION_ID");
 
+        public static bool RemoveHostName => bool.TryParse(GetValue("DOCFX_REMOVE_HOST_NAME"), out var removeHostName) && removeHostName;
```
yufeih, line 22: This is not needed because the environment var is auto populated during config load

live1206: ok

PR #6195, opened 2020-07-06
Using ordinal comparer to sorting .errors.log

mingwli:
following #6188
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6195)

### DIFF
--- a/test/docfx.RegressionTest/Normalizer.cs
+++ b/test/docfx.RegressionTest/Normalizer.cs
@@ -86,6 +86,7 @@ namespace Microsoft.Docs.Build
             foreach (var log in logs)
             {
                 var obj = JObject.Parse(log);
+                obj.Remove("date_time");
 
                 if (obj.ContainsKey("code")
                     && obj["code"]!.Value<string>() == "yaml-syntax-error"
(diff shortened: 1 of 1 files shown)


mingwli:
`String.CompareTo` default calls [InternalCompareString]
image
(https://referencesource.microsoft.com/#mscorlib/system/globalization/compareinfo.cs,1356715f78447ed6,references).
And I'm not sure about its behavior on different platform.

the `-`, `--` is ignored, and seems default comparer is culture-variant.

test/docfx.RegressionTest/Normalizer.cs
```diff
@@ -80,13 +80,12 @@ private static string NormalizeJsonFile(string path)
 
         private static void PrettifyJsonLog(string logPath)
         {
-            var logs = File.ReadAllLines(logPath).OrderBy(line => line);
+            var logs = File.ReadAllLines(logPath).Select(line => Regex.Replace(line, ",\"date_time\":.*?Z\"", "")).OrderBy(line => line);
```
yufeih, line 83: If in doubt, give it a comparer
```suggestion
            var logs = File.ReadAllLines(logPath).Select(line => Regex.Replace(line, ",\"date_time\":.*?Z\"", "")).OrderBy(line => line, StringComparer.Ordinal);
```

mingwli: right, `Ordinal` should be more secure. I thought both ways are acceptable. Locally tried `Ordinal` against `Regex replace`, no much perf gap.

PR #6200, opened 2020-07-07
Update Jint to the latest version

yufeih:
Move thread handling logic outside of JavaScriptEngine, so it is easier to attach a different JavaScriptEngine per thread using `ThreadLocal`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6200)

### DIFF
--- a/.github/actions/spell-check/dictionary/custom-dictionary.txt
+++ b/.github/actions/spell-check/dictionary/custom-dictionary.txt
@@ -518,3 +518,4 @@ MEMORYSTATUSEX
 sizeof
 IMemory
 uncomment
+Esprima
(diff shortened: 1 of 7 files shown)


lahma:
Out of curiosity, did you run before and after benchmarks? There's a lot of performance work done in Jint 3 so it would be interesting to know if things got faster.

yufeih:
Yes! and plenty so hence this upgrade. (I can share the results later).

There is a noticeable PInvoke overhead for chakra in some of our use cases and it seems like a dead-end, so I'm looking if JINT can be tuned to handle these cases better.

lahma:
Great to hear, please ping back the project if there specific use cases that are slow for you, maybe we can figure out something to make things faster. One thing at least to note is to prefer running the engine in strict mode (via options) and favoring arrow functions over normal ones. Strict enables doing less and arrow functions require less context and are thus faster to dispatch.

Waiting eagerly to see the results 👍

lahma:
@yufeih so did you have the chance to check before/after performance? 🙂

PR #6216, opened 2020-07-09
send docfx config event

928PJY:
[AB#264062](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/264062)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6216)


src/docfx/lib/log/DocfxBuildTelemetryModel.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    [JsonObject(NamingStrategyType = typeof(SnakeCaseNamingStrategy))]
+    internal class DocfxBuildTelemetryModel
+    {
+        public Dictionary<string, Dictionary<string, object>> FileMetadata { get; set; } = new Dictionary<string, Dictionary<string, object>>();
+
+        [JsonExtensionData]
+        public Dictionary<string, object> ExtensionData { get; set; } = new Dictionary<string, object>();
+
+        public bool ShouldSerializeFileMetadata() => false;
```
yufeih, line 18: trim complex leaf nodes when size exceed?

src/docfx/lib/log/Telemetry.cs
```diff
@@ -73,6 +74,28 @@ public static void SetRepository(string? repo, string? branch)
...
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Error happens when deserializing the config file {fileName}: {ex.Message}");
+            }
+
+            var dimensions = new Dictionary<string, string>();
+            dimensions["DocsetName"] = docsetName;
+            dimensions["Config"] = docfxConfigTelemetryValue;
+            TrackEvent(fileName, dimensions);
```
yufeih, line 96: nit: why not directly call `s_telemetryClient.TrackEvent`? this helper method does not seem to add much value.

928PJY: this method is used to construct the eventTelemetry?

---
In reply to: [452600406](https://github.com/dotnet/docfx/pull/6216#discussion_r452600406) [](ancestors = 452600406)

src/docfx/lib/log/Telemetry.cs
```diff
@@ -147,5 +184,50 @@ private static string CoalesceEmpty(string? str)
...
+        private static string SerializeConfigObjectWithMaxLength(JObject graph)
+        {
+            TryRemoveNestedObject(graph);
+            var result = JsonUtility.Serialize(graph);
+            if (result.Length <= MaxEventPropertyLength)
+            {
+                return result;
+            }
+
+            TryRemoveLongArray(graph);
```
live1206, line 197: Feel a bit strange to trim again here and also why not trim JObject as well, how about reduce `MaxChildrenLength` instead?

src/docfx/lib/log/Telemetry.cs
```diff
@@ -73,6 +78,22 @@ public static void SetRepository(string? repo, string? branch)
...
+            var hashCode = HashUtility.GetMd5Hash(docfxConfigTelemetryValue);
+            if (docfxConfigTelemetryValue.Length > MaxEventPropertyLength)
+            {
+                docfxConfigTelemetryValue = SerializeConfigObjectWithMaxLength(docfxConfig);
+            }
+
+            var dimensions = new Dictionary<string, string>();
+            dimensions["DocsetName"] = docsetName;
+            dimensions["Config"] = docfxConfigTelemetryValue;
+            dimensions["ContentHash"] = hashCode;
```
yufeih, line 93: What is the purpose of this hash?

928PJY: `hashcode` will be used by the Feature Utilization team to detect the config change once we remove some nested object from the too-long config

src/docfx/lib/log/Telemetry.cs
928PJY: remove it then

src/docfx/lib/log/Telemetry.cs
```diff
@@ -73,6 +78,24 @@ public static void SetRepository(string? repo, string? branch)
...
         }
 
+        public static void TrackDocfxConfig(string fileName, string docsetName, JObject docfxConfig)
+        {
+            var newValue = JsonUtility.DeepClone(docfxConfig) as JObject;
+            var docfxConfigTelemetryValue = JsonUtility.Serialize(newValue!);
+            var hashCode = HashUtility.GetMd5Hash(docfxConfigTelemetryValue);
+            if (docfxConfigTelemetryValue.Length > MaxEventPropertyLength)
+            {
+                TryRemoveNestedObject(newValue!);
```
yufeih, line 88: The `!` here will cause an `NRE` if the config is not a `JObject`, in which case we can skip the fallback.

src/docfx/lib/log/Telemetry.cs
```diff
@@ -73,6 +78,24 @@ public static void SetRepository(string? repo, string? branch)
             s_telemetryClient.Context.GlobalProperties["Branch"] = s_branch;
         }
 
+        public static void TrackDocfxConfig(string docsetName, JObject docfxConfig)
+        {
+            var docfxConfigTelemetryValue = JsonUtility.Serialize(docfxConfig);
+            var hashCode = HashUtility.GetMd5Hash(docfxConfigTelemetryValue);
+            if (docfxConfigTelemetryValue.Length > MaxEventPropertyLength)
+            {
+                var newValue = JsonUtility.DeepClone(docfxConfig) as JObject;
```
yufeih, line 87:
```suggestion
                var newValue = (JObject)JsonUtility.DeepClone(docfxConfig);
```

KalleOlaviNiemitalo:
Does this now send the whole docfx.json to Application Insights? How will I be able to opt out of that when building proprietary non-Microsoft documentation?

yufeih:
@KalleOlaviNiemitalo Today you have to opt in to collect telemetry using the CLI (by manually setting the AI instrumentation key) and we only collect information Microsoft docs repos.

@928PJY @herohua , if in the future we choose to collect telemetry by default to the public (whether in CLI or vscode extension), we'll need a privacy review.

PR #6224, opened 2020-07-10
Fix for duplicate-uids of multiple versioning SDP

AngryBerryMS:
Fix the duplicate-uid warnings by adding monikers for specs (which is used to determin whether same uids have different monikers) of each file.

[AB#263955](https://dev.azure.com/ceapex/Engineering/_workitems/edit/263955)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6224)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -25,7 +25,12 @@ namespace Microsoft.Docs.Build
         private static ThreadLocal<Stack<SourceInfo<string>>> t_recursionDetector
                  = new ThreadLocal<Stack<SourceInfo<string>>>(() => new Stack<SourceInfo<string>>());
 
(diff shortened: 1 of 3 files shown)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -119,6 +126,8 @@ public JsonSchemaTransformer(MarkdownEngine markdownEngine, LinkResolver linkRes
                     LazyThreadSafetyMode.PublicationOnly);
             }
 
+            xref.Monikers = _monikerProvider.GetFileLevelMonikers(file.FilePath).monikers;
```
yufeih, line 129: consider make `monikers` a required constructor parameter of `InternalXrefSpec` to help avoid this kind of bug.

PR #6230, opened 2020-07-13
Turn git-clone-incomplete into a warning

yufeih:
This is more on local build, if building a repo clone using `--depth 1`, `git-clone-incomplete` will almost always throw. This PR turns it into a warning and tweak the behavior to match what the warning message says.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6230)

### DIFF
--- a/src/docfx/lib/process/InterProcessReaderWriterLock.cs
+++ b/src/docfx/lib/process/InterProcessReaderWriterLock.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Docs.Build
         private static FileStream WaitFile(string name, string lockHash, FileAccess access, FileShare fileShare)
         {
             var path = Path.Combine(AppData.MutexRoot, lockHash);
-            var start = DateTime.UtcNow;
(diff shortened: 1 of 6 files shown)


src/docfx/lib/process/InterProcessReaderWriterLock.cs
```diff
@@ -51,8 +51,10 @@ private static FileStream WaitFile(string name, string lockHash, FileAccess acce
                 }
                 catch
                 {
-                    if (DateTime.UtcNow - start > TimeSpan.FromSeconds(30))
+                    var now = DateTime.UtcNow;
+                    if (now - lastPrintTime > TimeSpan.FromSeconds(30))
                     {
+                        lastPrintTime = now;
```
yufeih, line 57: This avoid printing too much `Waiting for another process to access` to the console.

PR #6233, opened 2020-07-13
Empty path-to-root fallback to "."

live1206:
[AB#264636](https://dev.azure.com/ceapex/Engineering/_workitems/edit/264636)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6233)

### DIFF
--- a/docs/specs/ops.yml
+++ b/docs/specs/ops.yml
@@ -268,11 +268,12 @@ repos:
           "docsets_to_publish": [{ "build_source_folder": "articles" }],
           "dependent_repositories": [{
               "path_to_root": "",
-              "url": "https://github.com/ops1/template",
+              "url": "https://github.com/ops1/template1",
               "branch": "test"
             }]
         }
       articles/docfx.yml:
+  https://github.com/ops1/template1#test:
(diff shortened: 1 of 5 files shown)


live1206:
> IMO the problem is around how we use Path.GetRelativePath in general. With PathString, empty string is treated as '.' but Path.GetRelativePath does not behave the same. A proper fix to me is give PathString the ability to GetRelativePath and use PathString consistently across the board.

Sounds good to me, planned to either fallback to `.` or throw user error, did not recall the fallback logic of `PathString`.
Thanks for the reminding!👍

live1206:
> IMO the problem is around how we use Path.GetRelativePath in general. With PathString, empty string is treated as '.' but Path.GetRelativePath does not behave the same. A proper fix to me is give PathString the ability to GetRelativePath and use PathString consistently across the board.

Use `PathString` as the model field can simply implement the fallback to `.`
Doubt that if we want to turn everything into `PathString`.

PR #6255, opened 2020-07-17
Fix regression test unstable log

yufeih:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6255)

### DIFF
--- a/test/docfx.RegressionTest/RegressionTest.cs
+++ b/test/docfx.RegressionTest/RegressionTest.cs
@@ -73,15 +73,16 @@ namespace Microsoft.Docs.Build
 
             static string GetDocfxConfig(Options opts)
             {
-                // Git token for CRR restore
-                var http = new Dictionary<string, object>();
-                http["https://github.com"] = new { headers = ToAuthHeader(s_githubToken) };
-                http["https://dev.azure.com"] = new { headers = ToAuthHeader(s_azureDevopsToken) };
                 var docfxConfig = JObject.FromObject(new
                 {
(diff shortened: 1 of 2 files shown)


test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -73,15 +73,16 @@ private static (string baseLinePath, string outputPath, string repositoryPath, s
...
                 {
-                    http,
-                    maxWarnings = 5000,
-                    maxInfos = 30000,
+                    http = new Dictionary<string, object>
+                    {
+                        ["https://github.com"] = new { headers = ToAuthHeader(s_githubToken) },
+                        ["https://dev.azure.com"] = new { headers = ToAuthHeader(s_azureDevopsToken) },
+                    },
+                    maxFileWarnings = 1000,
```
mingwli, line 83: 🐶

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -73,15 +73,16 @@ private static (string baseLinePath, string outputPath, string repositoryPath, s
...
+                    maxFileSuggestions = 1000,
```
live1206, line 84: And there is no performance downgrade with such size of error log?

yufeih: seems so

PR #6257, opened 2020-07-17
Track file log count 

XiaofuHuang:
[AB#249218](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/249218)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6257)


src/docfx/config/Config.cs
```diff
@@ -107,6 +107,11 @@ internal class Config : PreloadConfig
         /// </summary>
         public int MaxFileInfos { get; private set; } = 20;
 
+        /// <summary>
+        /// Gets the maximum remove deduplication log count
+        /// </summary>
+        public int MaxRemoveDeduplicationLogCount { get; private set; } = 200;
```
OsmondJiang, line 113: we don't expect user to change that count, so how about we hard code at where it's used?

XiaofuHuang: Got it

PR #6260, opened 2020-07-17
Regression-test send crash stacktrace

mingwli:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6260)

### DIFF
--- a/test/docfx.RegressionTest/RegressionTest.cs
+++ b/test/docfx.RegressionTest/RegressionTest.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Docs.Build
                 Run,
                 _ =>
                 {
-                    SendPullRequestComments(crashed: true, "argument exception");
+                    SendPullRequestComments("regression-test argument exception");
                     return -9999;
                 });
         }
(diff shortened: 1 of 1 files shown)


test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -345,36 +358,34 @@ private static string BasicAuth(string? token)
             return Convert.ToBase64String(Encoding.UTF8.GetBytes($"user:{token}"));
         }
 
-        private static Task SendPullRequestComments()
+        private static Task SendPullRequestComments(bool crashed = false, string message = "")
```
yufeih, line 361: nit: one arg is enough?
```suggestion
        private static Task SendPullRequestComments(string? crashMessage = null)
```

live1206:
@yufeih The error log of the test output is still unstable

yufeih:
@OsmondJiang is working on the update on validation lib

PR #6263, opened 2020-07-17
Strip empty moniker zone HTML from output

yufeih:
[AB#268907](https://dev.azure.com/ceapex/Engineering/_workitems/edit/268907/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6263)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -479,13 +479,16 @@ namespace Microsoft.Docs.Build
                 => new Error(ErrorLevel.Suggestion, "moniker-range-undefined", "Moniker range missing in docfx.yml/docfx.json, user should not define it in file metadata or moniker zone. NOTE: This Suggestion will become a Error on 06/30/2020.", source);
 
             /// <summary>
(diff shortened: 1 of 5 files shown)


src/docfx/Errors.cs
```diff
@@ -483,9 +483,6 @@ public static Error MonikerRangeUndefined(SourceInfo? source)
             /// or moniker-zone defined in article.md has no intersection with file-level monikers.
             /// </summary>
             /// Behavior: ✔️ Message: ❌
-            public static Error MonikeRangeOutOfScope(SourceInfo<string?> source, IEnumerable<string> zoneLevelMonikers, IEnumerable<string> fileLevelMonikers)
```
live1206, line 486: Per offline discussion, the out of scope moniker range can be warning for moniker zone, but it should has its own error code, not sharing with the file level one anymore.

cc @herohua for awareness

PR #6264, opened 2020-07-17
Add pull-request-only property to error

XiaofuHuang:
[AB#249217](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/249217)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6264)


src/docfx/build/Build.cs
```diff
@@ -57,7 +59,10 @@ private static bool BuildDocset(string docsetPath, string? outputPath, CommandLi
 
                 new OpsPreProcessor(config, errorLog, buildOptions).Run();
                 var sourceMap = new SourceMap(new PathString(buildOptions.DocsetPath), config, fileResolver);
-                errorLog.Configure(config, buildOptions.OutputPath, sourceMap);
+                var validationRules = JsonUtility.DeserializeData<Dictionary<string, ValidationRules>>(
```
yufeih, line 62: move this line into `ErrorLog.Configure` method because `ErrorLog.Configure` is also used in `Restore.cs`

XiaofuHuang: As we disscussed offline, now restore read the validationRules may cause some problem, so we only read it in Build.cs

src/docfx/build/Build.cs
```diff
@@ -57,7 +59,13 @@ private static bool BuildDocset(string docsetPath, string? outputPath, CommandLi
 
                 new OpsPreProcessor(config, errorLog, buildOptions).Run();
                 var sourceMap = new SourceMap(new PathString(buildOptions.DocsetPath), config, fileResolver);
-                errorLog.Configure(config, buildOptions.OutputPath, sourceMap);
+
+                var validationRules = !string.IsNullOrEmpty(config.MarkdownValidationRules.Value)
+                    ? JsonUtility.DeserializeData<Dictionary<string, ValidationRules>>(
+                        fileResolver.ReadString(config.MarkdownValidationRules),
+                        config.MarkdownValidationRules.Source?.File)
```
OsmondJiang, line 66: extract to a method to keep the main pipeline clean

src/docfx/lib/log/CustomRule.cs
```diff
@@ -16,8 +16,22 @@ internal class CustomRule
 
         public bool CanonicalVersionOnly { get; private set; }
 
+        public bool PullRequestOnly { get; private set; }
+
         public CustomRule() { }
 
-        public CustomRule(ErrorLevel? severity) => Severity = severity;
```
yufeih, line 21: Keep this constructor to make `ShortHandConverter` work. There is an existing shorthand to turn off a rule as illustrated in the test.

XiaofuHuang: Got it

PR #6269, opened 2020-07-17
Support tilde_path metadata

live1206:
[AB#268911](https://dev.azure.com/ceapex/Engineering/_workitems/edit/268911)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6269)

### DIFF
--- a/src/docfx/build/link/LinkResolver.cs
+++ b/src/docfx/build/link/LinkResolver.cs
@@ -222,7 +222,7 @@ namespace Microsoft.Docs.Build
             if (relativePath.StartsWith("~/") || relativePath.StartsWith("~\\"))
             {
                 var (_, metadata) = _metadataProvider.GetMetadata(referencingFile);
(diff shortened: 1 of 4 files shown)


src/docfx/build/link/LinkResolver.cs
```diff
@@ -218,8 +221,13 @@ internal class LinkResolver
 
             if (relativePath.StartsWith("~/") || relativePath.StartsWith("~\\"))
             {
-                // Treat ~/ as path relative to docset
-                pathToDocset = new PathString(relativePath.Substring(2).TrimStart('/', '\\'));
+                var (_, metadata) = _metadataProvider.GetMetadata(referencingFile);
+                pathToDocset = new PathString(Path.GetRelativePath(
```
yufeih, line 225: Consider only do the `Path.Combine` here and move line 238 - line 242 to line 244

live1206: right

PR #6271, opened 2020-07-17
Prevent NOTE, TIP, etc. sections from being split by a page break in PDF

Blake-Eryx:
NOTE sections can sometimes be cut in half by page break in PDF output. This will help avoid that.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6271)


superyyrrzz:
A reasonable improvement. Thanks @Blake-Eryx !

PR #6283, opened 2020-07-21
fix for ECMA2YAML config in docset level

live1206:
[AB#270694](https://dev.azure.com/ceapex/Engineering/_workitems/edit/270694)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6283)


src/docfx/config/ops/OpsConfigLoader.cs
```diff
@@ -119,20 +119,25 @@ private static (JObject obj, string path, string name)[] GetDependencies(OpsConf
             {
```
yufeih, line 119: use fallback like `docsetConfig?.ECMA2Yaml ?? opsConfig.ECMA2Yaml` to indicate that the semantic is the same regardless of where the config is in?

live1206: sure

PR #6286, opened 2020-07-21
refactor: Remove Error.ToString parameters

yufeih:
[AB#268450](https://dev.azure.com/ceapex/Engineering/_workitems/edit/268450/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6286)

### DIFF
--- a/src/docfx/lib/log/ErrorLog.cs
+++ b/src/docfx/lib/log/ErrorLog.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Docs.Build
             foreach (var exception in exceptions)
             {
                 Log.Write(exception);
-                if (Write(exception.Error, exception.OverwriteLevel))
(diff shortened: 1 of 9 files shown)


src/docfx/lib/log/ErrorLog.cs
```diff
@@ -108,12 +108,19 @@ public bool Write(Error error, ErrorLevel? overwriteLevel = null)
                 return false;
             }
 
+            error = error.WithLevel(level);
```
live1206, line 111: Should this line go after line 99?

yufeih: yep

PR #6289, opened 2020-07-22
Support source_monikers for redirection

live1206:
[AB#268909](https://dev.azure.com/ceapex/Engineering/_workitems/edit/268909)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6289)


docs/specs/ops.yml
```diff
@@ -269,3 +269,24 @@ repos:
...
 outputs:
+---
+# Support source_monikers for rediction
+repos:
+  https://github.com/redirection/source-monikers:
+  - files:
+      docfx.yml: |
+        monikerDefinition: monikerDefinition.json
+      .openpublishing.redirection.json: |
+        { "redirections": [{"source_path": "a.md", "source_monikers": ["netcore-1.0", "netcore-2.0"], "redirect_URL": "/"}] }
```
yufeih, line 280: consider rename `source_monikers` to `monikers`? This feels more aligned with treating redirection properties as YAML header.

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -43,6 +43,8 @@ public MonikerProvider(Config config, BuildScope buildScope, MetadataProvider me
             _monikerOrder = GetMonikerOrder(monikerDefinition);
         }
 
+        public (List<Error>, MonikerList) Validate(SourceInfo<string?>[] monikers) => _rangeParser.Validate(monikers);
```
yufeih, line 46: make `MonikerProvider.GetFileLevelMonikers` return the right moniker list for redirection, so code that depend on that method works. (Like publish url conflict validation, consider a test to cover multiple versions of the same file with different redirect URLs)

live1206: Actually, I do not think the current model could fit this requirement.
`MonikerProvider.GetFileLevelMonikers(FilePath)` and `FilePath` of the redirection will be the same for different monikers, not like the physical files, for them different files would always end up with different monikers

yufeih, line 46: remove?

docs/specs/redirection.yml
```diff
@@ -564,3 +564,53 @@ inputs:
...
 outputs:
+---
+# Support source_monikers for rediction
+repos:
+  https://github.com/redirection/source-monikers:
+  - files:
+      docfx.yml: |
+        monikerDefinition: monikerDefinition.json
+      .openpublishing.redirection.json: |
+        { "redirections": [{"source_path": "a.md", "monikers": ["netcore-1.0", "netcore-2.0"], "redirect_URL": "/b"}, {"source_path": "a.md", "monikers": "netcore-2.1", "redirect_URL": "/c"}] }
```
yufeih, line 575: nit: format it a little bit to show there are two lines of redirections

consider also adding a negative case of two redirections of the same path with overlapping monikers

src/docfx/lib/path/FilePath.cs
```diff
@@ -50,46 +52,47 @@ public FilePath(string path)
...
 
         public static FilePath Content(PathString path)
         {
             Debug.Assert(!System.IO.Path.IsPathRooted(path));
-            return new FilePath(FileOrigin.Main, path, default, default);
+            return new FilePath(FileOrigin.Main, path, default, default, default);
         }
 
-        public static FilePath Redirection(PathString path)
+        public static FilePath Redirection(PathString path, MonikerList monikers = default)
```
yufeih: nit: consider remove the optional constructor parameter to ensure redirection has monikers
```suggestion
        public static FilePath Redirection(PathString path, MonikerList monikers)
```

src/docfx/lib/path/FilePath.cs
```diff
@@ -112,6 +115,10 @@ public override string ToString()
 
                 default:
                     tags += $"[{Origin.ToString().ToLowerInvariant()}]";
+                    if (Monikers.HasMonikers)
```
yufeih: move this to line 124?

src/docfx/lib/path/FilePath.cs
```diff
@@ -169,6 +176,11 @@ public int CompareTo(FilePath? other)
                 result = IsGitCommit.CompareTo(other.IsGitCommit);
             }
 
+            if (result == 0)
+            {
```
yufeih, line 182: consume `Monikers` in `Equals` method

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -42,7 +42,7 @@ internal class PublishModelBuilder
             _sourceMap = sourceMap;
         }
 
-        public void SetPublishItem(FilePath file, JObject? metadata, string? outputPath)
+        public void SetPublishItem(FilePath file, JObject? metadata, string? outputPath = null)
```
yufeih: consider keep existing signature?

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -41,16 +41,16 @@ internal class RedirectionProvider
...
                    new Lazy<(IReadOnlyDictionary<FilePath, FilePath> renameHistory, IReadOnlyDictionary<FilePath, (FilePath, SourceInfo?)> redirectionHistory)>(
                         () => GetRenameAndRedirectionHistory(redirections, _redirectUrls));
-                _publishUrlMap = publishUrlMap;
             }
         }
 
         public bool Contains(FilePath file)
         {
-            return _redirectUrls.ContainsKey(file);
+            return _redirectUrls.Keys.Select(x => x.Path).Contains(file.Path);
```
yufeih, line 53: consider create another dictionary using `PathString` as key to avoid this linear search? The redirection list is big.

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -114,7 +114,15 @@ public FilePath GetOriginalFile(FilePath file)
                 }
 
                 var absoluteRedirectUrl = redirectUrl.Value.Trim();
-                var filePath = FilePath.Redirection(path);
+
+                MonikerList monikers = default;
+                if (item.Monikers != null)
+                {
+                    List<Error> monikerErrors;
+                    (monikerErrors, monikers) = _monikerProvider.Validate(item.Monikers);
```
yufeih, line 122: consider to the validation in `MonikerProvider.GetFileLevelMonikers`

src/docfx/build/redirection/RedirectionProvider.cs
live1206: u r right!

PR #6297, opened 2020-07-23
Create an ErrorBuilder abstraction to split ErrorWriter from ErrorLog

yufeih:
[AB#268450](https://dev.azure.com/ceapex/Engineering/_workitems/edit/268450/)

Split `ErrorLog` into `ErrorLog` and `ErrorWriter`, where `ErrorLog` depends on `Config` and `ErrorWriter` is pure output. Created an `ErrorBuilder` to abstract them.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6297)

### DIFF
--- a/src/docfx/lib/log/ErrorBuilder.cs
+++ b/src/docfx/lib/log/ErrorBuilder.cs
@@ -1,0 +1,47 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Docs.Build
+{
+    internal abstract class ErrorBuilder
+    {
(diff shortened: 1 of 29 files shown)


src/docfx/lib/log/ErrorBuilder.cs
```diff
@@ -0,0 +1,47 @@
...
+
+namespace Microsoft.Docs.Build
+{
+    internal abstract class ErrorBuilder
+    {
+        public abstract bool HasError { get; }
+
+        public abstract void Add(Error error);
+
+        public virtual bool FileHasError(FilePath file) => throw new NotSupportedException();
```
live1206, line 15: could this be abstract method as well?

PR #6322, opened 2020-07-28
Skip moniker validation for YAML files

live1206:
[AB#275017](https://dev.azure.com/ceapex/Engineering/_workitems/edit/275017)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6322)

### DIFF
--- a/src/docfx/build/moniker/MonikerProvider.cs
+++ b/src/docfx/build/moniker/MonikerProvider.cs
@@ -75,7 +75,7 @@ namespace Microsoft.Docs.Build
             // For conceptual docset,
             // Moniker range not defined in docfx.yml/docfx.json,
             // User should not define it in moniker zone
-            if (!_config.SkipMonikerValidation && configMonikerRange.Value is null)
+            if (!file.SkipMonikerValidation && configMonikerRange.Value is null)
             {
                 errors.Add(Errors.Versioning.MonikerRangeUndefined(rangeString));
                 return default;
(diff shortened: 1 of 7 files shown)


src/docfx/lib/path/FilePath.cs
```diff
@@ -40,6 +40,11 @@ internal class FilePath : IEquatable<FilePath>, IComparable<FilePath>
 
         public MonikerList Monikers { get; }
 
+        /// <summary>
+        /// Skip moniker validation for YAML files
+        /// </summary>
+        public bool SkipMonikerValidation { get; }
```
yufeih, line 46: this is not part of the key, so remove this field from this class.

src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -75,7 +75,7 @@ public MonikerList GetZoneLevelMonikers(ErrorBuilder errors, FilePath file, Sour
             // For conceptual docset,
             // Moniker range not defined in docfx.yml/docfx.json,
             // User should not define it in moniker zone
-            if (!_config.SkipMonikerValidation && configMonikerRange.Value is null)
+            if (!file.SkipMonikerValidation && configMonikerRange.Value is null)
```
yufeih, line 78:
```suggestion
            if (file.Format == FileFormat.Markdown && configMonikerRange.Value is null)
```

PR #6323, opened 2020-07-28
Loc consume _repo.en-us in op-config

mingwli:
[AB#274997](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/274997)

added a `FallbackRepository` in `PreLoadConfig` for restoring fallback repository, which is converted from `_repo.en-us`
Priority: `_repo.en-us` > `conventional`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6323)

### DIFF
--- a/src/docfx/lib/LocalizationUtility.cs
+++ b/src/docfx/lib/LocalizationUtility.cs
@@ -43,19 +43,24 @@ namespace Microsoft.Docs.Build
             return s_lrmAdjustment.Replace(text, me => $"{me.Groups[1]}{me.Groups[2]}&lrm;{me.Groups[3]}{me.Groups[4]}");
         }
 
-        public static string? GetFallbackDocsetPath(string docsetPath, Repository? repository, PackageResolver packageResolver)
+        public static string? GetFallbackDocsetPath(string docsetPath, Repository? repository, PackagePath? fallbackRepository, PackageResolver packageResolver)
         {
-            if (repository != null)
+            var docsetSourceFolder = repository == null
+                ? docsetPath.Split('/').Last()
+                : Path.GetRelativePath(repository.Path, docsetPath);
+            var fallbackRemote = fallbackRepository?.Url;
+            var fallbackBranch = fallbackRepository?.Branch;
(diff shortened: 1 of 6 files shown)


src/docfx/lib/LocalizationUtility.cs
```diff
@@ -43,19 +43,24 @@ public static string AddLeftToRightMarker(CultureInfo culture, string text)
             return s_lrmAdjustment.Replace(text, me => $"{me.Groups[1]}{me.Groups[2]}&lrm;{me.Groups[3]}{me.Groups[4]}");
         }
 
-        public static string? GetFallbackDocsetPath(string docsetPath, Repository? repository, PackageResolver packageResolver)
+        public static string? GetFallbackDocsetPath(string docsetPath, Repository? repository, PackagePath? fallbackRepository, PackageResolver packageResolver)
         {
-            if (repository != null)
+            var docsetSourceFolder = repository == null
+                ? docsetPath.Split('/').Last()
```
yufeih, line 49: We currently don't support any fallback if loc content does not belong to a repository. So keep the `if (repository is null) return null` behavior.

src/docfx/lib/LocalizationUtility.cs
```diff
@@ -43,19 +43,24 @@ public static string AddLeftToRightMarker(CultureInfo culture, string text)
...
+                : Path.GetRelativePath(repository.Path, docsetPath);
+            var fallbackRemote = fallbackRepository?.Url;
+            var fallbackBranch = fallbackRepository?.Branch;
+            if (fallbackRemote == null && repository != null)
```
yufeih, line 53: check `PackagePath.Type` to only use `fallbackRepository` if it's type is git.

yufeih:
> Shall we totally stop supporting naming convention?

That would be too _inconvenient_ for both _us_ (changing test cases) as well as people who wants to use loc build.

PR #6334, opened 2020-07-29
fix bug that the altTextDuplicate check is not stable

OsmondJiang:
the alt text in the included files are filtered out incorrectly, this fix should be able to help stable the diff in PR.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6334)


OsmondJiang:
the report is still wrong because we should report on the root file, not the include file, will change that in docs validation lib.

PR #6341, opened 2020-07-31
Reduce memory usage of legacy dependency map

yufeih:
[AB#268456](https://dev.azure.com/ceapex/Engineering/_workitems/edit/268456/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6341)


src/docfx/build/legacy/LegacyDependencyMap.cs
```diff
@@ -16,42 +18,42 @@ internal static class LegacyDependencyMap
...
+                        from_file_path = (basePath + dep.From).Replace('/', '\\'),
+                        to_file_path = (basePath + dep.To).Replace('/', '\\'),
                         version = dep.Version,
-                    })).ToArray();
+                    });
 
-                context.Output.WriteLines("full-dependent-list.txt", dependencyList);
-                context.Output.WriteLines("server-side-dependent-list.txt", dependencyList);
+                // TODO: remove this silly duplicated output
+                context.Output.WriteLines(new[] { "full-dependent-list.txt", "server-side-dependent-list.txt" }, dependencyList);
```
herohua, line 54: 👍👍👍

src/docfx/build/Build.cs
```diff
@@ -108,6 +109,8 @@ private static void Run(Context context)
             // TODO: decouple files and dependencies from legacy.
             var dependencyMap = context.DependencyMapBuilder.Build();
 
+            MemoryCache.Clear();
```
live1206, line 112: This could fix the issue by releasing the memory manually. 
But shall we rethink about the life cycle of the instances within `Context`?
Does everything in this "IoC container" need to keep the state all the time? 
For a real IoC container, we could configure the life cycle of the instance type.

mingwli:
>                         if (fileItem != null)

this will never be null now

---
Refers to: src/docfx/build/legacy/LegacyFileMap.cs:42 in c5eb6c0. [](commit_id = c5eb6c0c3bdb50144c4b3b806e87704f665cccca, deletion_comment = False)

src/docfx/build/legacy/LegacyDependencyMap.cs
```diff
@@ -16,42 +18,42 @@ internal static class LegacyDependencyMap
...
+                            source.Path,
+                            dependencyItem.To.Path,
                             context.MonikerProvider.GetConfigMonikerRange(source),
                             dependencyItem.Type));
                     }
-                }
+                });
+
+                var basePath = Path.GetFullPath(Path.Combine(docsetPath)).Replace("\\", "/") + '/';
+                var jsonOptions = new JsonSerializerOptions { IgnoreNullValues = true };
```
mingwli, line 37: unused?

src/docfx/build/Build.cs
live1206: We will break down the dependencies based on precise usage, that would be further refactoring

PR #6363, opened 2020-08-05
Warn on duplicated xml files

live1206:
[AB#277006](https://ceapex.visualstudio.com/Engineering/_workitems/edit/277006)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6363)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -765,10 +765,10 @@ namespace Microsoft.Docs.Build
                 => new Error(ErrorLevel.Error, "mustache-not-found", $"Mustache template is not found at '{templateFileName}'.");
         }
 
-        public static class ECMA2Yaml
+        public static class SourceMap
         {
-            public static Error DuplicateXml(string yamlFile, IEnumerable<PathString> xmlFiles)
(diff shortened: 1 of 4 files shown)


src/docfx/Errors.cs
```diff
@@ -764,5 +764,11 @@ public static Error LiquidNotFound(SourceInfo<string?> source)
             public static Error MustacheNotFound(string templateFileName)
                 => new Error(ErrorLevel.Error, "mustache-not-found", $"Mustache template is not found at '{templateFileName}'.");
         }
+
+        public static class SourceMap
+        {
+            public static Error DuplicateContent(string key, IEnumerable<PathString> originalFiles)
+                => new Error(ErrorLevel.Warning, "duplicate-content", $"'{key}' is duplicated from {StringUtility.Join(originalFiles)}");
```
herohua, line 771: Do we want to add a hint that which file is ignored?

live1206: Do not think this matters, since the user has the clear action indication already.
Please refer to `duplicate-uid`, which has the same error message.

yufeih: `duplicate-content` seems a broad concept, consider `duplicate-sourcemap-item`

yufeih:
silly @msftbot

PR #6365, opened 2020-08-05
content_git_url, original_content_git_url respect user input

mingwli:
[AB#279561](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/279561)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6365)

### DIFF
--- a/src/docfx/build/page/BuildPage.cs
+++ b/src/docfx/build/page/BuildPage.cs
@@ -166,7 +166,9 @@ namespace Microsoft.Docs.Build
                 = context.DocumentProvider.GetDocumentId(context.RedirectionProvider.GetOriginalFile(file.FilePath));
 
             (systemMetadata.ContentGitUrl, systemMetadata.OriginalContentGitUrl, systemMetadata.OriginalContentGitUrlTemplate)
-                = context.ContributionProvider.GetGitUrl(file.FilePath);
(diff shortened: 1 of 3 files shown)


src/docfx/build/page/BuildPage.cs
```diff
@@ -166,7 +166,9 @@ private static SystemMetadata CreateSystemMetadata(ErrorBuilder errors, Context
                 = context.DocumentProvider.GetDocumentId(context.RedirectionProvider.GetOriginalFile(file.FilePath));
 
             (systemMetadata.ContentGitUrl, systemMetadata.OriginalContentGitUrl, systemMetadata.OriginalContentGitUrlTemplate)
-                = context.ContributionProvider.GetGitUrl(file.FilePath);
```
herohua, line 169: So there won't be a case that only one of the three is specified, but the others are expected to be generated by DocFX?

mingwli: There's no mix-usage case for this. 
Either all **user-defined**: special SDP case, e.g. powershell 
Or all **system-generated**

PR #6369, opened 2020-08-06
Add GetZoneLevelMonikers and GetZone to markdown object

yufeih:
[AB#278675](https://dev.azure.com/ceapex/Engineering/_workitems/edit/278675/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6369)

### DIFF
--- a/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MonikerRange/MonikerRangeBlock.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MonikerRange/MonikerRangeBlock.cs
@@ -9,8 +9,12 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
     public class MonikerRangeBlock : ContainerBlock
     {
         public string MonikerRange { get; set; }
+
+        public object ParsedMonikers { get; set; }
+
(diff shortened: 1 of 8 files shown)


src/Microsoft.DocAsCode.MarkdigEngine.Extensions/MonikerRange/MonikerRangeBlock.cs
```diff
@@ -9,8 +9,12 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
     public class MonikerRangeBlock : ContainerBlock
     {
         public string MonikerRange { get; set; }
+
+        public object ParsedMonikers { get; set; }
```
live1206, line 13: how about just `Monikers`?

yufeih: That would confuse with `MonikerRange` because we have other places where `MonikerRange` and `Monikers` are used at the same level to represent input. This is not an input.

PR #6370, opened 2020-08-06
Learn achievement dependency

mingwli:
when `LearningPath1.achievement` refers to `LearningPath.trophy`, dependency type should be `achievement`.
[AB#280867](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/280867)

Using transient declaring property name to distinguish the referenced xref spec's type

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6370)

### DIFF
--- a/src/docfx/build/xref/InternalXrefSpec.cs
+++ b/src/docfx/build/xref/InternalXrefSpec.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
 namespace Microsoft.Docs.Build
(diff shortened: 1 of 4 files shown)


src/docfx/build/xref/InternalXrefSpec.cs
```diff
@@ -21,12 +22,17 @@ internal class InternalXrefSpec : IXrefSpec
 
         string IXrefSpec.Uid => Uid.Value;
 
-        public InternalXrefSpec(SourceInfo<string> uid, string href, Document declaringFile, MonikerList monikerList)
+        // TODO: change to use xrefSpec type to express what kind of xref spec it is: e.g. achievement, module
+        [JsonIgnore]
+        internal string? DeclaringPropertyName { get; }
```
yufeih, line 27: Technically this is the full JSON token path

```suggestion
        internal string? DeclaringPropertyPath { get; }
```

PR #6371, opened 2020-08-06
Support public template definition file

928PJY:
For now, to run the local validation, we still need to clone the template repository, but it is private, so it is not accessible for the public contributors. but the schemas are already public and organized by this definition file: https://static.docs.com/ui/latest/schemas/schema_config.json.

With this change, in `dry-run` mode and after resolving the following dependencies, a private template repository is no longer needed.

# Dependency
- For SDP, some bookmarks are imported by mustache, but if we support another contentType `bookmark` in schema, and let the schema tell whether the property will be exported as a bookmark, then we will not need the `*.js` and `*.html.tmpl` anymore.
- For SDP, some page are data type and will go through a different flow, for now, we detect that by search whether there is a `{mime}.html.tmpl`. but if we can let the schema/definition file to tell whether it is a data type, we can get rid of this detection.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6371)

### DIFF
--- a/src/docfx/template/TemplateEngine.cs
+++ b/src/docfx/template/TemplateEngine.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Diagnostics;
 using System.IO;
 using System.Runtime.InteropServices;
 using System.Threading;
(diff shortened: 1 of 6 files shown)


src/docfx/template/TemplateEngine.cs
```diff
@@ -12,47 +13,79 @@ namespace Microsoft.Docs.Build
...
+        private readonly LiquidTemplate? _liquid;
+        private readonly ThreadLocal<IJavaScriptEngine>? _js;
+        private readonly MustacheTemplate? _mustacheTemplate;
+        private readonly TemplateDefinition? _templateDefinition;
+        private readonly FileResolver _fileResolver;
 
         private readonly ConcurrentDictionary<string, JsonSchemaValidator?> _schemas
                    = new ConcurrentDictionary<string, JsonSchemaValidator?>(StringComparer.OrdinalIgnoreCase);
 
-        public TemplateEngine(Config config, BuildOptions buildOptions, PackageResolver packageResolver, Lazy<JsonSchemaTransformer> jsonSchemaTransformer)
```
herohua, line 26: We probably want to extract a TemplateProvider and have two implementations for package-based and file-based.

928PJY:
Close this one and create another PR #6550

PR #6372, opened 2020-08-06
Display alt not text when external href is null

AngryBerryMS:
When the external hrefs don't defined (as the picture shows below), the spec.Value is not null but the spec.Value.Href is an empty string, **we should return null for the xrefspec at this kind of cases**. 
After this fix, when encounter the xref quote like `<xref href="SlackAPI.Block?alt=SlackAPI.Block&text=Block" data-throw-if-not-resolved="True"></xref>`, will display the "alt" part but not the "text" part.

Bug: [278744](https://ceapex.visualstudio.com/Engineering/_workitems/edit/278744)

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6372)

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -80,7 +80,7 @@ namespace Microsoft.Docs.Build
 
             // need to url decode uid from input content
             var (xrefError, xrefSpec, resolvedHref) = ResolveXrefSpec(new SourceInfo<string>(uid, href.Source), referencingFile, inclusionRoot);
-            if (xrefError != null || xrefSpec is null || resolvedHref == null || resolvedHref.Length == 0)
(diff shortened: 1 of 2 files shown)


src/docfx/build/xref/XrefResolver.cs
```diff
@@ -210,7 +210,7 @@ private string RemoveSharingHost(string url, string hostName)
             if (_externalXrefMap.Value.TryGetValue(uid, out var spec))
             {
                 var href = RemoveSharingHost(spec.Value.Href, _config.HostName);
-                return (spec.Value, href);
+                return ((spec.Value.Href != null && spec.Value.Href.Length == 0) || (spec.Value.Href == null)) ? default : (spec.Value, href);
```
live1206, line 213: can simply check `string.IsNullOrEmpty(href)`

PR #6377, opened 2020-08-07
Give a warning when source-path not specified in .openpublishing.redi…

AngryBerryMS:
Give a warning when source-path not specified in .openpublishing.redirection.json

[AB#279435](https://dev.azure.com/ceapex/Engineering/_workitems/edit/279435)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6377)

### DIFF
--- a/docs/specs/redirection.yml
+++ b/docs/specs/redirection.yml
@@ -237,6 +237,7 @@ inputs:
 outputs:
   .errors.log: |
     {"message_severity":"error","code":"redirection-is-empty","message":"The key or value of redirection 'sccm/mdm/index.md: ' is null or empty.","file":"redirections.yml","line":2,"column":22}
+    {"message_severity":"error","code":"redirection-is-empty","message":"The key or value of redirection '.: /docfx/test' is null or empty.","file":"redirections.yml","line":3,"column":7}
     {"message_severity":"error","code":"redirection-is-empty","message":"The key or value of redirection 'a.md: ' is null or empty.","file":"redirections.yml","line":4,"column":8}
 ---
 # The dest to redirection url does not match any files's publish URL, but the redirect_with_id flag has been set as true
@@ -657,7 +658,7 @@ outputs:
       ]
     }
 ---
-# Give a warning when source-path not specified in .openpublishing.redirection.json
+# There is no 'source-path' defined for 'redirect_url': '/'
 repos:
   https://docs.com/ops/redirect-1:
   - files:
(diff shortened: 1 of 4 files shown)


docs/specs/redirection.yml
```diff
@@ -656,3 +656,15 @@ outputs:
...
+repos:
+  https://docs.com/ops/redirect-1:
+  - files:
+      docfx.yml:
+      .openpublishing.redirection.json: |
+        { "redirections": [{"redirect_url": "/"}] }
+outputs:
+  .publish.json:
+  .errors.log: |
+    {"message_severity":"warning","code":"source-path-not-specified","message":"There is no 'source-path' defined in .openpublishing.redirection.json file.","line":1,"column":39}
```
live1206, line 670: There could be other redirection containing `source-path` in the redirection file, so it is inaccurate to say `in .openpublishing.redirection.json file`.
How about `There is no 'source-path' defined for 'redirect_url': '/'`?

src/docfx/Errors.cs
```diff
@@ -341,6 +341,16 @@ public static Error RedirectionUrlNotFound(string from, SourceInfo<string> sourc
 
             public static Error CircularRedirection(SourceInfo? source, IEnumerable<FilePath> redirectionChain)
                 => new Error(ErrorLevel.Warning, "circular-redirection", $"Build has identified circular redirection: {string.Join(" --> ", redirectionChain)}.", source);
+
+            /// <summary>
+            /// If the source-path defined in .openpublishing.redirection.json cannot be found,
+            /// should give this warning to users.
+            /// </summary>
+            /// Behavior: ✔️ Message: ✔️
```
live1206, line 349: message is not confirmed with PM yet

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -173,6 +173,18 @@ private static RedirectionItem[] LoadRedirectionModel(ErrorBuilder errors, strin
                     // Rebase source_path based on redirection definition file path
                     var basedir = Path.GetDirectoryName(fullPath) ?? "";
 
+                    // Give a warning when source-path not specified in .openpublishing.redirection.json
+                    if (fullPath.EndsWith(".openpublishing.redirection.json"))
```
live1206, line 177: this warning should include `renames` as well

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -173,6 +173,18 @@ private static RedirectionItem[] LoadRedirectionModel(ErrorBuilder errors, strin
...
+                        foreach (var item in redirections)
+                        {
+                            if (item.SourcePath.IsDefault)
+                            {
+                                errors.Add(Errors.Redirection.SourcePathNotSpecified(item.RedirectUrl));
+                            }
+                        }
+                    }
+
                     return (
```
live1206, line 188: Convert this into a `foreach` loop, warn on missing `item.SourcePath` to avoid another loop for warnings

docs/specs/redirection.yml
AngryBerryMS: Ok. 👌

src/docfx/Errors.cs
AngryBerryMS: Fixed.

live1206: Offline synced with @yufeih, we can use `missing-attribute` for this usage
https://github.com/dotnet/docfx/blob/c9aaefec21a15af46c024e2421c7cbbb83724586/src/docfx/Errors.cs#L563

src/docfx/build/redirection/RedirectionProvider.cs
herohua: Yes, no matter where the redirection rules come from, we should always have this validation.

---
In reply to: [466897413](https://github.com/dotnet/docfx/pull/6377#discussion_r466897413) [](ancestors = 466897413)

src/docfx/build/redirection/RedirectionProvider.cs
AngryBerryMS: Done

src/docfx/build/redirection/RedirectionProvider.cs
AngryBerryMS: Done

src/docfx/Errors.cs
AngryBerryMS: Done

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -173,17 +167,34 @@ private static RedirectionItem[] LoadRedirectionModel(ErrorBuilder errors, strin
...
+                            results.Add(new RedirectionItem
+                            {
+                                SourcePath = new PathString(sourcePath),
+                                Monikers = item.Monikers,
+                                RedirectUrl = item.RedirectUrl,
+                                RedirectDocumentId = item.RedirectDocumentId,
+                            });
+                        }
+
+                        if (item.SourcePath.IsDefault)
```
live1206, line 185: put these two `if` check ahead, `continue` within them.
Then we can skip the previous duplicated `if` check.

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -173,17 +167,34 @@ private static RedirectionItem[] LoadRedirectionModel(ErrorBuilder errors, strin
...
+                        {
+                            // Give a missing-attribute warning when source-path not specified
+                            errors.Add(Errors.JsonSchema.MissingAttribute(item.RedirectUrl, "source_path"));
+                        }
+
+                        if (string.IsNullOrEmpty(item.RedirectUrl.Value))
```
live1206, line 191:
```suggestion
                        if (string.IsNullOrEmpty(item.RedirectUrl))
```

docs/specs/redirection.yml
```diff
@@ -656,3 +657,53 @@ outputs:
...
+  - files:
+      docfx.yml:
+      .openpublishing.redirection.json: |
+        { "redirections": [{"source_path": "a.md"}] }
+outputs:
+  .publish.json:
+  .errors.log: |
+    {"message_severity":"warning","code":"missing-attribute","message":"Missing required attribute: 'redirect_url'.","line":0,"column":0}
+---
+# The 'sourth-path' and 'redirect-url' are both empty
```
live1206, line 685: This test can be removed, duplicated with test `#redirect to empty URL`

PR #6380, opened 2020-08-10
Safe guard MarkdownObject.GetPathToRoot method

yufeih:
Last PR of [AB#278675](https://dev.azure.com/ceapex/Engineering/_workitems/edit/278675/):

Safe guard `GetPathToRoot` method and detect misuse. Fixed a couple of bugs found by the misuse with comments inline

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6380)


docs/specs/output.yml
```diff
@@ -512,7 +512,19 @@ outputs:
...
+      "links":[
+        {"source_url":"/docs/a/toc.json","target_url":"/docs/b","source_git_url":"https://github.com/output-links/markdown/blob/master/docs/a/TOC.yml","source_line":2},
```
yufeih, line 516: The diff in this test is shown in regression test result, the same image alt text suggestion is reported twice with different source locations. This PR fixed it to show only the correct location.

yufeih:
Diffs are all expected now.

src/Microsoft.DocAsCode.MarkdigEngine.Extensions/TripleColon/TripleColonInlineParser.cs
```diff
@@ -37,8 +39,8 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             var inline = new TripleColonInline(this)
             {
                 Closed = false,
-                Column = 0,
-                Line = processor.LineIndex,
+                Line = line,
+                Column = column,
                 Span = new SourceSpan(processor.LineIndex, slice.End),
```
live1206, line 44: Can be removed? being set afterwards.

yufeih: yep

PR #6441, opened 2020-08-25
IsCanonicalVersion bool? -> bool

yufeih:
`bool` is enough to represent `IsCanonicalVersion`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6441)

### DIFF
--- a/src/docfx/validation/MetadataValidator.cs
+++ b/src/docfx/validation/MetadataValidator.cs
@@ -52,14 +52,8 @@ namespace Microsoft.Docs.Build
 
             foreach (var schemaValidator in _schemaValidators)
             {
-                if (file.PageType != null)
+                if (file.PageType != null && file.PageType != "toc")
                 {
-                    if (file.PageType == "conceptual" &&
-                        metadata.ContainsKey("layout") &&
-                        !string.Equals(metadata.GetValue("layout")?.ToString(), "conceptual", StringComparison.OrdinalIgnoreCase))
-                    {
-                        continue;
-                    }
                     errors.AddRange(schemaValidator.Validate(metadata, file.FilePath));
                 }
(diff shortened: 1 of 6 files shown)


yufeih:
diffs are expected

src/docfx/validation/MetadataValidator.cs
```diff
@@ -52,14 +52,8 @@ public void ValidateMetadata(ErrorBuilder errors, JObject metadata, Document fil
 
             foreach (var schemaValidator in _schemaValidators)
             {
-                if (file.PageType != null)
+                if (file.PageType != null && file.PageType != "toc")
```
OsmondJiang, line 55: remove this hard code, every valdiation rules should have contentTypes

OsmondJiang (approved):

src/docfx/validation/MetadataValidator.cs
yufeih: done

OsmondJiang:
the change in sql-docs-pr is strange

yufeih:
> the change in sql-docs-pr is strange, that's something should not be reported for reference page

Gosh, content validation rules uses `exclusions` while `metadata` rules uses `inclusions`. 

image

OsmondJiang:
> > the change in sql-docs-pr is strange, that's something should not be reported for reference page
> 
> Gosh, content validation rules uses `exclusions` while `metadata` rules uses `inclusions`.
> 
> image

Hmmm. yes, it's caused by some history reason

OsmondJiang (approved):

Issue #6448, opened 2020-08-26
Support `main` and `master`as the default branch


928PJY:
[AB#285222](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/285222)
- Branch should be nullable for the current repository (#6449 )
- Branch fallback of dependencies from `main` to `master` (#6453)
- Xrefmap output (#6450 )
- Learn validation (#6451)

PR #6449, opened 2020-08-26
Branch should be nullable

928PJY:
#6448 

[AB#291371](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/291371)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6449)

### DIFF
--- a/src/docfx/build/metadata/ContributionProvider.cs
+++ b/src/docfx/build/metadata/ContributionProvider.cs
@@ -183,7 +183,7 @@ namespace Microsoft.Docs.Build
             return _input.GetFullPath(originalPath is null ? file : new FilePath(originalPath));
         }
 
-        private string? GetContentGitUrl(string repo, string branch, string pathToRepo)
+        private string? GetContentGitUrl(string repo, string? branch, string pathToRepo)
         {
             if (!string.IsNullOrEmpty(_config.EditRepositoryUrl))
             {
--- a/src/docfx/build/metadata/ContributionProvider.cs
+++ b/src/docfx/build/metadata/ContributionProvider.cs
@@ -109,7 +109,7 @@ namespace Microsoft.Docs.Build
         {
             if (fileCommits.Length > 0)
             {
-                if (_config.UpdateTimeAsCommitBuildTime && repository != null)
+                if (_config.UpdateTimeAsCommitBuildTime && repository != null && repository.Branch != null)
                 {
                     return _commitBuildTimeProviders
                         .GetOrAdd(repository.Path, new Lazy<CommitBuildTimeProvider>(() => new CommitBuildTimeProvider(_config, repository))).Value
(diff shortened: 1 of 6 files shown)


src/docfx/config/AppData.cs
```diff
@@ -35,10 +35,11 @@ public static string GetCommitCachePath(string remote)
             return Path.Combine(CacheRoot, "commits", HashUtility.GetMd5Hash(remote));
         }
 
-        public static string GetCommitBuildTimePath(string remote, string branch)
+        public static string GetCommitBuildTimePath(string remote, string? branch)
         {
+            var branchHash = string.IsNullOrEmpty(branch) ? string.Empty : HashUtility.GetMd5Guid(branch).ToString();
```
yufeih, line 40: consider suppress the commit build time feature entirely when there is no branch

928PJY: which feature do you mean? can you add the task here if there is?

src/docfx/lib/git/Repository.cs
```diff
@@ -19,7 +19,7 @@ private Repository(string remote, string? branch, string commit, PathString path
         {
             // remove user name, token and .git from url like https://xxxxx@dev.azure.com/xxxx.git
             Remote = Regex.Replace(remote, @"^((http|https):\/\/)?([^\/\s]+@)?([\S]+?)(\.git)?$", "$1$4");
-            Branch = branch ?? "master";
```
herohua, line 22: Is it possible to simply change to `main`?

928PJY: We can, but it is not necessary, we prefer to keep the real value if we can.

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -183,7 +183,7 @@ private PathString GetOriginalFullPath(FilePath file)
             return _input.GetFullPath(originalPath is null ? file : new FilePath(originalPath));
         }
 
-        private string? GetContentGitUrl(string repo, string branch, string pathToRepo)
+        private string? GetContentGitUrl(string repo, string? branch, string pathToRepo)
```
yufeih, line 186: This would produce an invalid URL like `blob//a.md`. Consider changing line 205 to fallback to commit hash to produce a clickable URL like `blob/a09cd/a.md` in case of detached head.

src/docfx/config/AppData.cs
yufeih: ContributionProvider.cs line 112 is currently controlling when to enable commit build time.

yufeih (approved):

PR #6450, opened 2020-08-26
Only distinguish live and non-live when output xrefmap

928PJY:
#6448 

[AB#291373](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/291373)
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6450)

### DIFF
--- a/docs/specs/xref.yml
+++ b/docs/specs/xref.yml
@@ -1224,9 +1224,9 @@ outputs:
         ]
     }
 ---
-# define UID in master branch
+# define UID in non-live branch
 repos:
-  https://docs.com/test#master:
+  https://docs.com/test#test:
     - files:
         docfx.yml:
         docs/a.md: |
@@ -1237,7 +1237,7 @@ repos:
 outputs:
   docs/a.json:
   .xrefmap.json: |
-    {"references":[{"uid":"a", href: "https://docs.com/docs/a?branch=master"}]}
+    {"references":[{"uid":"a", href: "https://docs.com/docs/a?branch=test"}],"properties": {"tags": ["/","internal"]}}
 ---
 # define UID in live branch
 repos:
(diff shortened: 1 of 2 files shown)


herohua (approved):

docs/specs/xref.yml
```diff
@@ -1237,7 +1237,7 @@ repos:
 outputs:
   docs/a.json:
   .xrefmap.json: |
-    {"references":[{"uid":"a", href: "https://docs.com/docs/a?branch=master"}]}
+    {"references":[{"uid":"a", href: "https://docs.com/docs/a?branch=test"}],"properties": {"tags": ["/","internal"]}}
```
live1206, line 1240: property tags can be omitted from the test

live1206 (approved):

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -145,7 +145,7 @@ public XrefMapModel ToXrefMapModel(bool isLocalizedBuild)
                 {
```
yufeih, line 145: Apply this patch only when `outputUrlType == Docs`?

yufeih (approved):

src/docfx/build/xref/XrefResolver.cs
928PJY: Why should add this logic in this time, I mean the original logic is also applied to all the `outputUrlType`. and in our spec, we already consume the live branch

---
In reply to: [477133335](https://github.com/dotnet/docfx/pull/6450#discussion_r477133335) [](ancestors = 477133335)

yufeih: doesn't need to be in this PR

PR #6453, opened 2020-08-26
Use main as the default branch of dependencies

928PJY:
#6448 

[AB#291372](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/291372)
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6453)

### DIFF
--- a/src/docfx/lib/LocalizationUtility.cs
+++ b/src/docfx/lib/LocalizationUtility.cs
@@ -109,7 +109,12 @@ namespace Microsoft.Docs.Build
                             try
                             {
                                 var defaultBranchFallbackBranch = GitUtility.GetDefaultBranchFallbackBranch(contributionBranch);
-                                GitUtility.Fetch(config, repository.Path, repository.Remote, $"+{defaultBranchFallbackBranch}:{defaultBranchFallbackBranch}", "--update-head-ok");
+                                GitUtility.Fetch(
+                                    config,
+                                    repository.Path,
+                                    repository.Remote,
+                                    $"+{defaultBranchFallbackBranch}:{defaultBranchFallbackBranch}",
+                                    "--update-head-ok");
                             }
                             catch (InvalidOperationException e)
                             {
--- a/src/docfx/lib/LocalizationUtility.cs
+++ b/src/docfx/lib/LocalizationUtility.cs
@@ -56,7 +56,7 @@ namespace Microsoft.Docs.Build
             if (fallbackRemote != null)
             {
                 var docsetSourceFolder = Path.GetRelativePath(repository.Path, docsetPath);
-                foreach (var branch in new[] { fallbackBranch, "master" })
+                foreach (var branch in new[] { fallbackBranch, "main" })
                 {
                     if (packageResolver.TryResolvePackage(new PackagePath(fallbackRemote, branch), PackageFetchOptions.None, out var fallbackRepoPath))
                     {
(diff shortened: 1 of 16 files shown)


src/docfx/lib/path/PackagePath.cs
```diff
@@ -49,7 +49,7 @@ public PackagePath(string remote, string? branch)
         {
             Type = PackageType.Git;
             Url = remote;
-            Branch = branch ?? "master";
+            Branch = AssignBranchWithDefaultBranchFallback(branch);
```
yufeih, line 52: Instead of mutating the `branch` value, consider keep the branch value and use a computed property like `IsMasterOrMain`? This way there is no hardcoded `== "main"` check and `DocfxTest.cs` doesn't need to change.

src/docfx/lib/git/FileCommitProvider.cs
```diff
@@ -274,8 +274,11 @@ private unsafe List<git_oid> LoadCommitIds(string? committish)
 
             if (git_revparse_single(out var headCommit, _repo, committish) != 0)
             {
-                git_object_free(walk);
-                throw Errors.Config.CommittishNotFound(_repository.Remote, committish).ToException();
+                if (committish != "main" || git_revparse_single(out headCommit, _repo, "master") != 0)
```
yufeih, line 277: I wasn't expecting changes in this layer, any reasons?

src/docfx/lib/LocalizationUtility.cs
```diff
@@ -72,11 +72,15 @@ public static string AddLeftToRightMarker(CultureInfo culture, string text)
...
         }
 
-        public static bool TryGetContributionBranch(string branch, [NotNullWhen(true)] out string? contributionBranch)
+        public static bool TryGetContributionBranch(string? branch, [NotNullWhen(true)] out string? contributionBranch)
         {
-            if (branch.EndsWith("-sxs"))
+            if (branch != null && branch.EndsWith("-sxs"))
             {
                 contributionBranch = branch[0..^4];
+                if (contributionBranch == "master")
```
yufeih, line 80: this change is also suspicious. This method is used in a couple of other places and the change seems to be not backward compatible.

src/docfx/lib/git/GitUtility.cs
```diff
@@ -47,6 +47,20 @@ public static bool IsGitRepository(string path)
             return false;
         }
 
+        public static bool IsDefaultBranch(string branch)
```
yufeih, line 50: Consider a different name since this is not technically checking for default branch, e.g., the default branch for docfx is `dev`.

928PJY: 😂

---
In reply to: [478139519](https://github.com/dotnet/docfx/pull/6453#discussion_r478139519) [](ancestors = 478139519)

928PJY: but in docfx, for all our consumed repository, we actually use the `master`/`main` as the default banch. so would this be a concern?

---
In reply to: [478140410](https://github.com/dotnet/docfx/pull/6453#discussion_r478140410) [](ancestors = 478140410,478139519)

yufeih (approved):
👍

PR #6447, opened 2020-08-25
Fix lazy not lazy when retrieving ops user token

yufeih:
https://dev.azure.com/ceapex/Engineering/_build/results?buildId=186480&view=logs&jobId=73369bf2-6897-51f4-4549-2cc8453877a2&j=743bc1cd-037d-5150-fc69-2953b556a708

```
Cannot get 'OPBuildUserToken' from azure key vault, please make sure you have been granted the permission to access: DefaultAzureCredential authentication failed: Object reference not set to an instance of an object.
```

This PR:

- Improves the exception handling
- Fix the wrong lazy

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6447)

PR #6457, opened 2020-08-26
Fix mac test failure

yufeih:
Fix mac test failure caused by #6447, that PR test pass without running mac tests due to manual trigger.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6457)

### DIFF
--- a/src/docfx/config/ops/OpsConfigAdapter.cs
+++ b/src/docfx/config/ops/OpsConfigAdapter.cs
@@ -31,8 +31,8 @@ namespace Microsoft.Docs.Build
         public static readonly DocsEnvironment DocsEnvironment = GetDocsEnvironment();
 
         private static readonly SecretClient s_secretClient = new SecretClient(new Uri("https://docfx.vault.azure.net"), new DefaultAzureCredential());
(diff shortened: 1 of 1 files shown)


src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -30,7 +30,9 @@ internal class OpsConfigAdapter : IDisposable
 
         public static readonly DocsEnvironment DocsEnvironment = GetDocsEnvironment();
 
-        private static readonly SecretClient s_secretClient = new SecretClient(new Uri("https://docfx.vault.azure.net"), new DefaultAzureCredential());
+        private static readonly Lazy<SecretClient> s_secretClient = new Lazy<SecretClient>(
```
live1206, line 33: These two should be both thread safe, only initialized once.

yufeih: now should be the right fix.

live1206 (approved):

PR #6484, opened 2020-08-31
Service page fix

AngryBerryMS:
Discussed with Tianqi, when the "LandingPageType" is `Root`, the generated service page's name should be "index.yml"，and we shouldn't create a folder/directory, either. (Only create folder/directory when the "LandingPageType" is `Service`)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6484)

### DIFF
--- a/src/docfx/build/toc/ServicePageGenerator.cs
+++ b/src/docfx/build/toc/ServicePageGenerator.cs
@@ -43,7 +43,17 @@ namespace Microsoft.Docs.Build
             {
                 var topLevelTOCRelativeDir = Path.GetDirectoryName(_joinTOCConfig.TopLevelToc);
                 var baseDir = string.IsNullOrEmpty(_joinTOCConfig.OutputFolder) ? topLevelTOCRelativeDir : _joinTOCConfig.OutputFolder;
-                var servicePagePath = FilePath.Generated(new PathString($"./{baseDir}/{directoryName}/{filename}.yml"));
+
+                var pageType = node.LandingPageType.Value;
+                FilePath servicePagePath;
+                if (pageType == LandingPageType.Root)
+                {
+                    servicePagePath = FilePath.Generated(new PathString($"./{baseDir}/{directoryName}/index.yml"));
+                }
+                else
+                {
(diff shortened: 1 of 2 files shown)


src/docfx/build/toc/ServicePageGenerator.cs
```diff
@@ -86,15 +96,21 @@ internal class ServicePageGenerator
...
             }
 
             foreach (var item in node.Items)
             {
-                GenerateServicePageFromTopLevelTOC(item, results, $"{directoryName}/{filename}");
+                if (!string.IsNullOrEmpty(node.LandingPageType.ToString()) && node.LandingPageType == LandingPageType.Root)
+                {
+                    GenerateServicePageFromTopLevelTOC(item, results, $"{directoryName}");
+                }
+                else
```
live1206, line 110: if `node.LandingPageType` is empty, shall we generate service page for it?

AngryBerryMS: Although `node.LandingPageType` may be empty, it's sub-sub items may have `LandingPageType`.
Although I haven't found cases like this, I think it deserves to be considered.........

live1206: Then we can just remove `!string.IsNullOrEmpty(node.LandingPageType.ToString())` here

AngryBerryMS: image

AngryBerryMS: Fixed.

live1206 (approved):

.github/actions/spell-check/dictionary/dictionary.txt
```diff
@@ -189829,6 +189829,7 @@ ileosigmoidostomy
 ileostomies
 ileostomy
 ileotomy
+Ilearn
```
yufeih, line 189832: revert

yufeih (approved):

PR #6494, opened 2020-09-01
Fix service page herf when the original href starts with ~/

AngryBerryMS:
Fix service page herf when the original href starts with ~/

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6494)

### DIFF
--- a/src/docfx/build/toc/ServicePageGenerator.cs
+++ b/src/docfx/build/toc/ServicePageGenerator.cs
@@ -72,8 +72,20 @@ namespace Microsoft.Docs.Build
                     {
                         if (topLevelTOCRelativeDir != null)
                         {
+                            string? hrefFileFullPath;
+
                             var topLevelTOCYmlDir = Path.GetFullPath(Path.Combine(_docsetPath, topLevelTOCRelativeDir));
-                            var hrefFileFullPath = Path.GetFullPath(Path.Combine(topLevelTOCYmlDir == null ? "" : topLevelTOCYmlDir, childHref));
+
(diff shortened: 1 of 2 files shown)


src/docfx/build/toc/ServicePageGenerator.cs
```diff
@@ -72,8 +72,20 @@ internal class ServicePageGenerator
                     {
                         if (topLevelTOCRelativeDir != null)
                         {
+                            string? hrefFileFullPath;
+
                             var topLevelTOCYmlDir = Path.GetFullPath(Path.Combine(_docsetPath, topLevelTOCRelativeDir));
-                            var hrefFileFullPath = Path.GetFullPath(Path.Combine(topLevelTOCYmlDir == null ? "" : topLevelTOCYmlDir, childHref));
+
+                            if (childHref.StartsWith("~/") || childHref.StartsWith("~\\"))
```
yufeih, line 79: Is there a reason why service pages are generated before resolving TOC links? If service pages are generated _after_ TOC link resolve, then presumably we don't have this problem?

AngryBerryMS: Service page yml files are generated before resolving. 
But if we don't do this, it will get wrong results. For example, "href: ~/api/doc.md" will be resolved as "../../~/api/doc.md" and result in an invalid file link error.

yufeih (approved):

PR #6496, opened 2020-09-02
Read liquid localized strings from token.json

yufeih:
Read from `yml/Conceptual.yml` --> `token.json` so we only consume one file from the template for localized string. Their contents are the same.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6496)

### DIFF
--- a/src/docfx/template/TemplateEngine.cs
+++ b/src/docfx/template/TemplateEngine.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Docs.Build
             _templateDefinition = PathUtility.LoadYamlOrJson<TemplateDefinition>(errors, _templateDir, "template") ?? new TemplateDefinition();
 
             _global = LoadGlobalTokens();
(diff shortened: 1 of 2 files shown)


mingwli:
> ⚠[AspNetCore.Docs](https://github.com/dotnet/AspNetCore.Docs)(00:00:14.6973453, 6010 more diff)

yml/Conceptual.yml --> token.json seems not totally the same

yufeih:
> > ⚠[AspNetCore.Docs](https://github.com/dotnet/AspNetCore.Docs)(00:00:14.6973453, 6010 more diff)
> 
> yml/Conceptual.yml --> token.json seems not totally the same

Yep and I'm patching pdf template: https://github.com/microsoft/templates.docs.msft.pdf/pull/19

partychen (approved):

PR #6498, opened 2020-09-03
Can cross-version referencing single-version uid

mingwli:
[AB#293823](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/293823)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6498)

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -232,7 +232,7 @@ namespace Microsoft.Docs.Build
             if (_internalXrefMap.Value.TryGetValue(uid, out var specs))
             {
                 var spec = default(InternalXrefSpec);
-                if (!monikers.HasValue || !monikers.Value.HasMonikers)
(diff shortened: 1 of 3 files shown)


live1206 (approved):

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -232,7 +232,7 @@ private static string RemoveSharingHost(string url, string hostName)
             if (_internalXrefMap.Value.TryGetValue(uid, out var specs))
             {
                 var spec = default(InternalXrefSpec);
-                if (!monikers.HasValue || !monikers.Value.HasMonikers)
+                if (specs.Length == 1 || !monikers.HasValue || !monikers.Value.HasMonikers)
```
yufeih, line 235: What exactly is the rule here? If we allow fallback to other versions (or better yet latest version), should line 244 also fallback?

PR #6511, opened 2020-09-04
Remove static variables in Chakra js engine

yufeih:
Makes it easier to abstract file IO for chakra core js engine because the `Require` method is currently `static`, and if we abstract IO operations, `Require` must access an instance of abstracted file system.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6511)

### DIFF
--- a/src/docfx/lib/js/ChakraCoreJsEngine.cs
+++ b/src/docfx/lib/js/ChakraCoreJsEngine.cs
@@ -30,7 +30,6 @@ namespace Microsoft.Docs.Build
         {
             _scriptDir = scriptDir;
             _requireFunction = new JavaScriptNativeFunction(Require);
-            GCHandle.Alloc(_requireFunction, GCHandleType.Pinned);
 
             if (global != null)
             {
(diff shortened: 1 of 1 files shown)


src/docfx/lib/js/ChakraCoreJsEngine.cs
```diff
@@ -145,19 +121,21 @@ private static JavaScriptValue Run(string scriptPath)
...
                         exports,
                         JavaScriptValue.FromString(dirname),
-                        JavaScriptValue.CreateFunction(s_requireFunction),
+                        JavaScriptValue.CreateFunction(_requireFunction),
                         JavaScriptValue.CreateObject(),
                     });
 
-                return modules[scriptPath] = module.GetProperty(exportsProperty);
+                var moduleExports = module.GetProperty(exportsProperty);
+                moduleExports.AddRef();
```
live1206, line 129: keep this comment?
```
// Avoid exports been GCed by javascript garbage collector.
```

live1206 (approved):

PR #6513, opened 2020-09-07
Loc's uid override fallback's when path different

mingwli:
[AB#296459](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/296459)

Loc's file path can be different with fallback's, and loc's should take higher priority.
e.g.
```
en-us: a.yml
loc: trans/a.yml
```
And hence if `a.yml` declares a uid `a`, `duplicate-uid` can be reported, which isn't the desired behavior.
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6513)

### DIFF
--- a/src/docfx/build/xref/InternalXrefMapBuilder.cs
+++ b/src/docfx/build/xref/InternalXrefMapBuilder.cs
@@ -126,6 +126,13 @@ namespace Microsoft.Docs.Build
                 return specsWithSameUid;
             }
 
+            // loc override fallback uid
+            if (specsWithSameUid.Length == 2 &&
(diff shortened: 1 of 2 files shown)


src/docfx/build/xref/InternalXrefMapBuilder.cs
```diff
@@ -126,6 +126,13 @@ private InternalXrefSpec[] AggregateXrefSpecs(string uid, InternalXrefSpec[] spe
                 return specsWithSameUid;
             }
 
+            // loc override fallback uid
+            if (specsWithSameUid.Length == 2 &&
```
yufeih, line 130: consider use filtering and not return directly here, there are more filtering below like moniker check.

yufeih (approved):

PR #6531, opened 2020-09-09
Virtualize file system access for template engine

yufeih:
Promote `Package` to abstract base class to allow different implementations like in memory for test, github for web etc. Apply this new virtualization to template engine.

This also makes it easier to access loose schema files for local validation.

[AB#298773](https://dev.azure.com/ceapex/Engineering/_workitems/edit/298773/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6531)


src/docfx/restore/LocalPackage.cs
```diff
@@ -0,0 +1,53 @@
...
+        public override IEnumerable<PathString> GetFiles()
+        {
+            if (!Directory.Exists(_directory))
+            {
+                throw Errors.Config.DirectoryNotFound(_directory).ToException();
+            }
+
+            return new FileSystemEnumerable<PathString>(_directory, ToPathString, s_enumerationOptions)
+            {
+                ShouldIncludePredicate = (ref FileSystemEntry entry) => !entry.IsDirectory && entry.FileName[0] != '.',
```
mingwli, line 32: >!entry.IsDirectory && entry.FileName[0] != '.' [](start = 72, length = 46)

Can this be shared with `PathUtility`

src/docfx/restore/PackageResolver.cs
```diff
@@ -2,281 +2,18 @@
...
 
 namespace Microsoft.Docs.Build
 {
-    internal class PackageResolver : IDisposable
+    internal abstract class PackageResolver : IDisposable
     {
-        // NOTE: This line assumes each build runs in a new process
-        private static readonly ConcurrentDictionary<(string gitRoot, string url, string branch), Lazy<PathString>> s_gitRepositories
-                          = new ConcurrentDictionary<(string gitRoot, string url, string branch), Lazy<PathString>>();
+        public abstract Package ResolvePackage2(PackagePath package, PackageFetchOptions options);
```
mingwli, line 11: >ResolvePackage2 [](start = 32, length = 15)

emmm, this name :/

test/docfx.Test/InMemoryPackage.cs
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) Microsoft. All rights reserved.
```
928PJY, line 1: maybe we can add this file when we implement the in-memory spec test

src/docfx/restore/Restore.cs
```diff
@@ -79,7 +79,7 @@ private static void RestorePackages(ErrorBuilder errors, BuildOptions buildOptio
             ParallelUtility.ForEach(
                 errors,
                 GetPackages(config).Distinct(),
-                item => packageResolver.DownloadPackage(item.package, item.flags));
+                item => packageResolver.ResolvePackage(item.package, item.flags));
```
928PJY, line 82: >ResolvePackage [](start = 40, length = 14)

no need to change

src/docfx/template/TemplateEngine.cs
```diff
@@ -37,21 +35,19 @@ internal class TemplateEngine : IDisposable
...
-            };
-
-            _contentTemplateDir = Path.Combine(_templateDir, "ContentTemplate");
-            _schemaDir = Path.Combine(_contentTemplateDir, "schemas");
+                template = new PackagePath(Path.Combine(buildOptions.DocsetPath, "_themes"));
+            }
 
-            _templateDefinition = PathUtility.LoadYamlOrJson<TemplateDefinition>(errors, _templateDir, "template") ?? new TemplateDefinition();
+            _package = packageResolver.ResolvePackage2(template, PackageFetchOptions.DepthOne);
+            _templateDefinition = _package.TryReadYamlOrJson<TemplateDefinition>(errors, "template") ?? new TemplateDefinition();
```
928PJY, line 45: >_package.TryReadYamlOrJson<TemplateDefinition>(errors, "template") ?? new TemplateDefinition(); [](start = 34, length = 95)

make this a Lazy load since it is not always needed

src/docfx/restore/LocalPackage.cs
```diff
@@ -0,0 +1,53 @@
...
+                    ? entry.FileName.ToString()
+                    : string.Concat(entry.Directory.Slice(entry.RootDirectory.Length + 1), "/", entry.FileName);
+
+                return PathString.DangerousCreate(result);
+            }
+        }
+
+        public override string? TryGetPhysicalPath(PathString path)
+        {
+            var fullPath = Path.GetFullPath(Path.Combine(_directory, path));
```
928PJY, line 49: >GetFullPath [](start = 32, length = 11)

why use GetFullPath here but not in line#19?

src/docfx/build/context/Output.cs
```diff
@@ -103,6 +103,31 @@ public void Copy(string destRelativePath, FilePath file)
...
+        /// Throws if multiple threads trying to write to the same destination concurrently.
+        /// </summary>
+        public void Copy(string destRelativePath, Package package, PathString sourcePath)
+        {
+            EnsureNoDryRun();
+
+            _queue.Post(() =>
+            {
+                var targetPhysicalPath = GetDestinationPath(destRelativePath);
+                var sourcePhysicalPath = package.TryGetPhysicalPath(sourcePath);
```
928PJY, line 117: in which case you cannot get the physical path but be able to readStream?

src/docfx/restore/LocalPackageResolver.cs
```diff
@@ -0,0 +1,284 @@
...
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+
+namespace Microsoft.Docs.Build
+{
+    internal class LocalPackageResolver : PackageResolver
```
928PJY, line 12: >LocalPackageResolver [](start = 19, length = 20)

should there be only one PackageResolver, which return different kind of Package?

src/docfx/template/MustacheTemplate.cs
```diff
@@ -199,15 +200,15 @@ private static bool IsTruthy(JValue value)
 
         private BlockToken? GetTemplate(string name)
         {
-            return _templates.GetOrAdd(name, key =>
+            return _templates.GetOrAdd(_baseDirectory.Concat(new PathString(name)), key =>
```
928PJY, line 203: >_baseDirectory.Concat(new PathString(name)) [](start = 39, length = 43)

just use the name as cache key to avoid useless computation

src/docfx/restore/LocalPackage.cs
928PJY: seems the `_directory` is already an absolute path, it is necessary to call `GetFullPath` here？

src/docfx/restore/PackageResolver.cs
928PJY: how about make this `ResolvePackage`, and rename the following `ResolvePackage` to `ResolvePackagePath`

---
In reply to: [486133240](https://github.com/dotnet/docfx/pull/6531#discussion_r486133240) [](ancestors = 486133240)

src/docfx/restore/LocalPackage.cs
yufeih: Yes, to normalize path in case path is a relative path. E.g., `Path.Combine("C:/a", "../b")` give `C:/a/../b` but ideally this method should return `C:/b`

src/docfx/restore/Package.cs
```diff
@@ -0,0 +1,60 @@
...
+            using var reader = ReadText(path);
+            return reader.ReadToEnd();
+        }
+
+        public virtual TextReader ReadText(PathString path)
+        {
+            return new StreamReader(ReadStream(path));
+        }
+
+        public T? TryReadYamlOrJson<T>(ErrorBuilder errors, string pathWithoutExtension) where T : class, new()
```
928PJY, line 41: >TryReadYamlOrJson [](start = 18, length = 17)

why not split this to two functions? in which case we need this fallback logic?

src/docfx/template/TemplateEngine.cs
```diff
@@ -37,21 +35,22 @@ internal class TemplateEngine : IDisposable
...
 
-            _templateDefinition = PathUtility.LoadYamlOrJson<TemplateDefinition>(errors, _templateDir, "template") ?? new TemplateDefinition();
+            _global = _package.TryReadYamlOrJson<JObject>(errors, "ContentTemplate/token") ?? new JObject();
 
-            _global = LoadGlobalTokens();
-            _liquid = new LiquidTemplate(_templateDir, _global);
-            _js = new ThreadLocal<JavaScriptEngine>(() => JavaScriptEngine.Create(_contentTemplateDir, _global));
-            _mustacheTemplate = new MustacheTemplate(_contentTemplateDir, _global, jsonSchemaTransformer);
+            _liquid = new LiquidTemplate(_package, _global);
+            _js = new ThreadLocal<JavaScriptEngine>(() => JavaScriptEngine.Create(_package, _global));
```
928PJY, line 52: Consider passing the basePath `contentTemplate` to JS engine when creating it, so no need to prepend it when run the scripts

src/docfx/template/liquid/LiquidTemplate.cs
```diff
@@ -78,17 +66,37 @@ public string Render(string templateName, SourceInfo<string?> mime, JObject mode
...
-            return Path.Combine((string)context["theme_rel"], resourcePath);
+            return t_package?.TryGetPhysicalPath(new PathString(resourcePath));
         }
 
-        private static Template LoadTemplate(string fullPath)
+        private Template LoadTemplate(PathString path)
         {
-            var template = Template.Parse(File.ReadAllText(fullPath));
+            return _templates.GetOrAdd(path, new Lazy<Template?>(() => LoadTemplateCore(path))).Value
+                ?? throw Errors.Template.LiquidNotFound(path).ToException(isError: false);
```
928PJY, line 88: >path [](start = 56, length = 4)

we have discussed this before, if we report a error message like `{layout}.liquid.js` not found, it may be confused to the end user since the layout is transport to the end user, so please keep the original error message based on `mime`

src/docfx/template/liquid/LiquidTemplate.cs
```diff
@@ -78,17 +66,37 @@ public string Render(string templateName, SourceInfo<string?> mime, JObject mode
...
         {
-            return Path.Combine((string)context["theme_rel"], resourcePath);
+            return t_package?.TryGetPhysicalPath(new PathString(resourcePath));
         }
 
-        private static Template LoadTemplate(string fullPath)
+        private Template LoadTemplate(PathString path)
         {
-            var template = Template.Parse(File.ReadAllText(fullPath));
+            return _templates.GetOrAdd(path, new Lazy<Template?>(() => LoadTemplateCore(path))).Value
```
928PJY, line 87: >new Lazy [](start = 45, length = 8)

lazy is not neccessary here?

src/docfx/template/TemplateEngine.cs
```diff
@@ -91,17 +90,15 @@ public JsonSchemaValidator GetSchemaValidator(SourceInfo<string?> mime)
...
         {
             var layout = model.RawMetadata?.Value<string>("layout") ?? mime.Value ?? throw new InvalidOperationException();
-            var themeRelativePath = _templateDir;
 
             var liquidModel = new JObject
             {
                 ["content"] = model.Content,
                 ["page"] = model.RawMetadata,
                 ["metadata"] = model.PageMetadata,
-                ["theme_rel"] = themeRelativePath,
```
928PJY, line 101: >theme_rel [](start = 18, length = 9)

this is not used any more?

src/docfx/restore/Restore.cs
928PJY: then remove the `DownloadPackage` from `localPackageResolver`

---
In reply to: [486173641](https://github.com/dotnet/docfx/pull/6531#discussion_r486173641) [](ancestors = 486173641)

src/docfx/template/MustacheTemplate.cs
928PJY: maybe PathString would be better to keep constant with others

---
In reply to: [486780502](https://github.com/dotnet/docfx/pull/6531#discussion_r486780502) [](ancestors = 486780502)

928PJY (approved):
:shipit:

PR #6550, opened 2020-09-14
Support public template

928PJY:
For now, to run the local validation, we still need to clone the template repository, but it is private, so it is not accessible for the public contributors. but the schemas are already public and organized by this definition file: https://static.docs.com/ui/latest/schemas/schema_config.json.

[AB#288440](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/288440)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6550)

### DIFF
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -485,3 +485,27 @@ outputs:
         {"asset_id": "original_b", "original_type": "Conceptual"},
       ]
     }
+---
+# Support public template
+inputs:
+  docfx.yml: |
+    template: "https://static.docs.com/ui/latest"
+  docs/a.yml: |
+    #YamlMime:TestData
+    test: b.md
+  docs/b.md:
+remoteFiles:
+  https://static.docs.com/ui/latest/schemas/TestData.schema.json: |
+    {
+      "renderType": "component",
+      "type": "object",
+      "properties": {
+        "test": {"contentType": "Href"},
+      }
+    }
+outputs:
+  docs/a.json: |
+    {
+      "test": "b"
+    }
+  docs/b.json:
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -494,7 +494,7 @@ inputs:
     #YamlMime:TestData
     test: b.md
   docs/b.md:
-remoteFiles:
+http:
   https://static.docs.com/ui/latest/schemas/TestData.schema.json: |
     {
       "renderType": "component",
--- a/src/docfx/config/TestQuirks.cs
+++ b/src/docfx/config/TestQuirks.cs
@@ -11,6 +11,8 @@ namespace Microsoft.Docs.Build
 
         public static Func<string, string>? GitRemoteProxy { get; set; }
 
+        public static Func<string, string?>? RemoteFileProxy { get; set; }
+
         public static bool? Verbose { get; set; }
     }
 }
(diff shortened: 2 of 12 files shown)


src/docfx/restore/PackageResolver.cs
```diff
@@ -44,9 +44,13 @@ public bool TryResolvePackage(PackagePath package, PackageFetchOptions options,
             }
         }
 
-        public Package ResolveAsPackage(PackagePath package, PackageFetchOptions options)
+        public Package ResolveAsPackage(PackagePath package, PackageFetchOptions options, FileResolver fileResolver)
```
yufeih, line 47: consider putting `FileResolver` as a member, if there is circular dependency, use `Lazy<FileResolver>`
```suggestion
        public Package ResolveAsPackage(PackagePath package, PackageFetchOptions options)
```

docs/specs/schema.yml
```diff
@@ -485,3 +485,27 @@ outputs:
...
+---
+# Support public template
+inputs:
+  docfx.yml: |
+    template: "https://static.docs.com/ui/latest"
+  docs/a.yml: |
+    #YamlMime:TestData
+    test: b.md
+  docs/b.md:
+remoteFiles:
```
yufeih, line 497: nit: consider `http`?
```suggestion
http:
```

src/docfx/restore/PublicTemplatePackage.cs
```diff
@@ -0,0 +1,51 @@
...
+        public override IEnumerable<PathString> GetFiles()
+        {
+            throw new NotSupportedException();
+        }
+
+        public override Stream ReadStream(PathString path) => _fileResolver.ReadStream(GetPath(path));
+
+        public override PathString? TryGetPhysicalPath(PathString path)
+        {
+            if (_fileResolver.TryResolveFilePath(GetPath(path), out var fullPath))
```
yufeih, line 35: We should never hit this right?
```suggestion
            throw new NotSupportedException();
```

src/docfx/restore/PublicTemplatePackage.cs
```diff
@@ -0,0 +1,51 @@
...
+
+        public PublicTemplatePackage(string baseUrl, FileResolver fileResolver)
+        {
+            _baseUrl = new Uri($"{baseUrl.TrimEnd('/')}/");
+            _fileResolver = fileResolver;
+        }
+
+        public override bool Exists(PathString path)
+        {
+            return _fileResolver.TryResolveFilePath(GetPath(path), out _);
```
yufeih, line 23: We should never hit this right?
```suggestion
    throw new NotSupportedException();
```

928PJY: it has been used when getting the schema

---
In reply to: [488341115](https://github.com/dotnet/docfx/pull/6550#discussion_r488341115) [](ancestors = 488341115)

src/docfx/config/TestQuirks.cs
```diff
@@ -11,6 +11,8 @@ internal static class TestQuirks
 
         public static Func<string, string>? GitRemoteProxy { get; set; }
 
+        public static Func<string, string?>? RemoteFileProxy { get; set; }
```
yufeih, line 14:
```suggestion
        public static Func<string, string?>? HttpProxy { get; set; }
```

yufeih (approved):

PR #6553, opened 2020-09-15
Support DOCS_SERVICE_ENDPOINT

928PJY:
[AB#288435](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/288435)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6553)

### DIFF
--- a/src/docfx/config/ops/OpsAccessor.cs
+++ b/src/docfx/config/ops/OpsAccessor.cs
@@ -113,7 +113,7 @@ namespace Microsoft.Docs.Build
         {
             using var request = new HttpRequestMessage
             {
-                RequestUri = new Uri("/route/mslearnhierarchy/api/OnDemandHierarchyDrySync"),
+                RequestUri = new Uri($"{BuildServiceEndpoint()}/route/mslearnhierarchy/api/OnDemandHierarchyDrySync"),
                 Method = HttpMethod.Post,
                 Content = new StringContent(body, Encoding.UTF8, "application/json"),
             };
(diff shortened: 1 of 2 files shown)


src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -145,22 +165,11 @@ private async Task<string> Fetch(string url, IReadOnlyDictionary<string, string>
...
-                return await _http.SendAsync(request);
-            }
-        }
-
-        private async Task<string> FetchValidationRules(string url, string repositoryUrl = "", string branch = "", DocsEnvironment? environment = null)
+        private async Task<string> FetchValidationRules(string routePath, string repositoryUrl = "", string branch = "", DocsEnvironment? environment = null)
         {
             try
             {
+                var url = $"{BuildServiceEndpoint(environment)}/{routePath.TrimStart('/')}";
```
yufeih, line 172: we can assert the url path always starts with `/`
```suggestion
                var url = $"{BuildServiceEndpoint(environment)}{routePath}";
```

yufeih (approved):

src/docfx/config/ops/OpsAccessor.cs
928PJY: this is a best practice heard from the service side,  it can still be right regardless of that the caller doesn't start with `/`

PR #6541, opened 2020-09-11
error mapping for learn validation

xichan-ms:
Draft solution: error mapping for learn validation

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6541)

PR #6554, opened 2020-09-15
Replace Error.Name with Error.Source.PropertyPath

yufeih:
Today we use JSON property name in `Error.Name` as key to specialize error messages. This works for metadata validation since metadata is a flat list. As we onboard more JSON schema based validations with nested structure, error specialization needs to work for nested properties, so _name_ is now extended to _property path_ in this PR.

Placing the _property path_ in `Error` does not scale well beyond JSON schema validation, other validations against JSON property (E.g. in JsonSchemaTransformer) also needs to access property path, so _property path_ is promoted to `SourceInfo.PropertyPath` and can be accessed from everywhere that produces a validation message.

#6541 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6554)


yufeih:
Close this approach because `SourceInfo.PropertyPath` is not immutable, in case of JSON merge (as seen when computing metadata), this property needs rebase for every node. So I tend to compute the property path during schema validation and transform stage now.

PR #6569, opened 2020-09-17
Default allow fallbacking to public data when the credential is not sufficient to get the required data

928PJY:
[AB#303461](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/303461)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6569)

### DIFF
--- a/src/docfx/config/ops/OpsAccessor.cs
+++ b/src/docfx/config/ops/OpsAccessor.cs
@@ -214,6 +214,9 @@ namespace Microsoft.Docs.Build
         {
             using (PerfScope.Start($"[{nameof(OpsConfigAdapter)}] Executing request '{request.Method} {request.RequestUri}'"))
             {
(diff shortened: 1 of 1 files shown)


live1206:
In the work item, it says "Overwritten to false in server build."
Should there be a corresponding PR in docs.build?

yufeih (commented):
`fallbacking`

928PJY:
> In the work item, it says "Overwritten to false in server build."
> Should there be a corresponding PR in docs.build?

@live1206 task updated, this metadata can always be set as true no matter in server build or local build.

live1206 (approved):

PR #6574, opened 2020-09-21
Ignore ecma2yml config item if source xml folder does not exist

live1206:
[#AB304055](https://dev.azure.com/ceapex/Engineering/_workitems/edit/304055)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6574)

### DIFF
--- a/src/docfx/config/ops/OpsConfigLoader.cs
+++ b/src/docfx/config/ops/OpsConfigLoader.cs
@@ -38,11 +38,11 @@ namespace Microsoft.Docs.Build
             }
 
             var buildSourceFolder = new PathString(Path.GetRelativePath(repository.Path, docsetPath));
-            return ToDocfxConfig(repository.Branch, opsConfig, buildSourceFolder);
+            return ToDocfxConfig(docsetPath, repository.Branch, opsConfig, buildSourceFolder);
         }
 
         private static (string? xrefEndpoint, string[]? xrefQueryTags, JObject config) ToDocfxConfig(
-            string? branch, OpsConfig opsConfig, PathString buildSourceFolder)
+            string docsetPath, string? branch, OpsConfig opsConfig, PathString buildSourceFolder)
         {
             var result = new JObject();
             var dependencies = GetDependencies(opsConfig, branch, buildSourceFolder);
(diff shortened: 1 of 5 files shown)


TianqiZhang (approved):

src/docfx/config/ops/OpsConfigLoader.cs
```diff
@@ -132,7 +132,12 @@ private static (JObject obj, string path, string name)[] GetDependencies(OpsConf
                 foreach (var ecma2Yaml in ecma2YamlConfig)
                 {
                     var ecma2YamlJObject = JsonUtility.ToJObject(ecma2Yaml);
-                    ecma2YamlJObject[nameof(ECMA2YamlRepoConfig.SourceXmlFolder)] = Path.GetRelativePath(buildSourceFolder, ecma2Yaml.SourceXmlFolder);
+                    var sourceXmlFolder = Path.GetRelativePath(buildSourceFolder, ecma2Yaml.SourceXmlFolder);
+                    if (!Directory.Exists(Path.GetFullPath(Path.Combine(docsetPath, sourceXmlFolder))))
```
TianqiZhang, line 136: sourceXmlFolder is not always relative to docsetPath, it could be relative to repo root.

live1206: As offline discussed, we get the relative path to `buildSourceFolder` firstly, which is `docsetPath`.

yufeih, line 136: Consider check existence before calling EcmaToYaml? I would like the config adapter to only to config JSON conversion so if a user writes it in `docfx.yml` this directory check is still there?

live1206: Then we need to check the existence in multiple places, such as `SourceMap`

yufeih: yep, `FileResolver.TryReadString` seems like a reasonable addition to have.

yufeih (approved):
👍

PR #6575, opened 2020-09-21
when href is empty, set it as null

AngryBerryMS:
When childHref is an empty string, set it to null. Or the docs-ui will render some empty table rows for it.

[AB#301651](https://dev.azure.com/ceapex/Engineering/_boards/board/t/DocFX%20v3%20Team/Stories/?workitem=301651)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6575)


src/docfx/build/toc/ServicePageGenerator.cs
```diff
@@ -85,7 +85,7 @@ internal class ServicePageGenerator
                     }
                     else
```
live1206, line 86: This else could be non-empty `childHref` and link type is not `RelativePath`.
For this case, shall we set `childHref` as `null` as well?

AngryBerryMS: No. It should be correct now since the last PR is merged.............🤦‍😔

live1206 (approved):

PR #6576, opened 2020-09-21
index.html: Change from .NET to explicit languages

perlun:
.NET is not a "programming language" per se, and even though I understand that it might be used as an umbrella term here, I think it makes more sense to explicitly mention the languages supported by DocFx in this case.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6576)


yufeih (approved):
Thanks! I also added F#.

PR #6580, opened 2020-09-22
Add name to Error.Tostring

xichan-ms:
[AB#297513](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/297513)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6580)


src/docfx/lib/log/Error.cs
```diff
@@ -109,7 +109,8 @@ public override string ToString()
             {
                 message_severity = Level,
                 Code,
-                Message,
+                property_path = Name,
```
yufeih, line 112: Name --> PropertyPath?

xichan-ms: OK

yufeih (approved):

xichan-ms:
Diff explanation: 
the differences are expected, because we add a property_path to the function ToString of Error.cs.

PR #6584, opened 2020-09-23
Add a basic docfx new command

yufeih:
With `docfx new`, you can scaffold a new project template like `docfx new conceptual`.

There is only one conceptual template at the moment but it can grow to include dotnet API reference in the future.

The default conceptual template makes use of a static template in <https://github.com/docascode/template>.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6584)


src/docfx/cli/New.cs
```diff
@@ -0,0 +1,113 @@
...
+                else
+                {
+                    Directory.CreateDirectory(Path.GetDirectoryName(targetFullPath));
+                    File.Copy(file, targetFullPath, overwrite: options.Force);
+                }
+            }
+
+            if (dryRun)
+            {
+                if (dryRun && overwriteFiles.Count > 0)
```
live1206, line 87:
```suggestion
                if (overwriteFiles.Count > 0)
```

live1206 (approved):

Issue #3357, opened 2018-09-03
Support line info in mixed contents


yufeih:
Line info is currently wrong for mixed contents, e.g., yaml header inside markdown, markdown content embedded inside yaml or json.

Split from #3050

PR #6587, opened 2020-09-24
Report approximate line number for markdown contents inside YAML

yufeih:
Mitigate #3357 as seen in #6541

This fix does not account for different string formatting options in YAML/JSON and new line escape mechanisms, it is a simple approximation not an accurate result. Doing an accurate source map does not seem achievable at the moment (or too complex to do) because it requires carry over source map out of YAML/JSON strings.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6587)


xichan-ms (approved):

yufeih:
The `.links.json` diff is wrong.

yufeih:
Diffs are correct now.

PR #6596, opened 2020-09-25
Content validation with monikers in consideration

bianyifan:
- Remove `File` from and add `Monikers` into `ValidationContext`
- Update `ContentValidator` and other related code to pass file-level and zone-level monikers to Docs Validation
- Remove moniker info from metadata `docsetUnique` validation error message
- Update Docs Validation version to apply new validators
- Cleanup explicit heading validations on included files

[AB#308802](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/308802)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6596)

### DIFF
--- a/src/docfx/validation/ContentValidator.cs
+++ b/src/docfx/validation/ContentValidator.cs
@@ -49,7 +49,12 @@ namespace Microsoft.Docs.Build
             // validate image link and altText here
             if (_links.TryAdd((file, link)) && TryGetValidationDocumentType(file, out var documentType))
             {
-                var validationContext = new ValidationContext { DocumentType = documentType, File = file.Path };
+                var validationContext = new ValidationContext
+                {
+                    DocumentType = documentType,
+                    FileSourceInfo = new SourceInfo(file),
+                    Monikers = _monikerProvider.GetFileLevelMonikers(_errors, file).ToList(),
+                };
                 Write(_validator.ValidateLink(
                     new Link
                     {
@@ -59,6 +64,7 @@ namespace Microsoft.Docs.Build
                         IsInlineImage = origin.IsInlineImage(imageIndex),
                         SourceInfo = link.Source,
                         ParentSourceInfoList = origin.GetInclusionStack(),
+                        Monikers = origin.GetZoneLevelMonikers().ToList(),
                     }, validationContext).GetAwaiter().GetResult());
             }
         }
@@ -67,7 +73,12 @@ namespace Microsoft.Docs.Build
         {
             if (TryGetValidationDocumentType(file, isInclude, out var documentType))
             {
-                var validationContext = new ValidationContext { DocumentType = documentType };
+                var validationContext = new ValidationContext
+                {
+                    DocumentType = documentType,
+                    FileSourceInfo = new SourceInfo(file),
+                    Monikers = _monikerProvider.GetFileLevelMonikers(_errors, file).ToList(),
+                };
                 Write(_validator.ValidateCodeBlock(codeBlockItem, validationContext).GetAwaiter().GetResult());
             }
         }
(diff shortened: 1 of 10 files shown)


yufeih (changes requested):
This change seems to trigger moniker validation for tokens

bianyifan:
@yufeih @OsmondJiang I have checked regression test results, they are all caused by changes of validation rule message or contents.

src/docfx/validation/ContentValidator.cs
```diff
@@ -59,6 +64,7 @@ public void ValidateImageLink(FilePath file, SourceInfo<string> link, MarkdownOb
                         IsInlineImage = origin.IsInlineImage(imageIndex),
                         SourceInfo = link.Source,
                         ParentSourceInfoList = origin.GetInclusionStack(),
+                        Monikers = origin.GetZoneLevelMonikers().ToList(),
```
yufeih, line 67: Are you able to change the contract to `IReadOnlyList<string>` where possible to avoid these `ToList` calls?

yufeih (approved):

src/docfx/validation/ContentValidator.cs
bianyifan: I have updated PR to Docs.Validation which is currently pending review. Will change this accordingly later.

src/docfx/validation/ContentValidator.cs
```diff
@@ -76,7 +87,12 @@ public void ValidateHeadings(FilePath file, List<ContentNode> nodes, bool isIncl
         {
             if (TryGetValidationDocumentType(file, isInclude, out var documentType))
             {
-                var validationContext = new ValidationContext { DocumentType = documentType, FileSourceInfo = new SourceInfo(file) };
+                var validationContext = new ValidationContext
```
OsmondJiang, line 90: can we create a method to create ValdiationContext? so we make sure all validation rules have enough validation context

src/docfx/validation/ContentValidator.cs
```diff
@@ -76,7 +87,12 @@ public void ValidateHeadings(FilePath file, List<ContentNode> nodes, bool isIncl
         {
             if (TryGetValidationDocumentType(file, isInclude, out var documentType))
```
OsmondJiang, line 88: `isInclude` should be removed?

src/docfx/validation/ContentValidator.cs
bianyifan: Not all validations can have monikers, e.g. manifest files don't have monikers, trying to get monikers on them will cause a warning being thrown.

src/docfx/validation/ContentValidator.cs
bianyifan: Retain this to future feature.

src/docfx/validation/ContentValidator.cs
OsmondJiang: so make the moniker as  optional?

src/docfx/validation/ContentValidator.cs
OsmondJiang: for what feature? learn content validation?

OsmondJiang: for heading validation, we can remove it at least.

bianyifan: Removed. `isInclude` is still there for `ValidateCodeBlock`, need cleanup in future.

OsmondJiang:
AB#311777

OsmondJiang (approved):

PR #6597, opened 2020-09-25
Add minItemWhen, maxItemWhen validation support

hejingkan2005:
[AB#297521](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/297521)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6597)


src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -262,6 +264,70 @@ private void ValidateProperties(JsonSchema schema, string propertyPath, JObject
...
+                    if (obj.TryGetValue(prop.Key, StringComparison.OrdinalIgnoreCase, out var value)
+                        && prop.Value != null
+                        && prop.Value.Const != null
+                        && prop.Value.Const.Equals(value))
+                    {
+                        return true;
+                    }
+                }
+            }
+            return false;
```
xichan-ms, line 328: How about we recalling validate function here and checking whether there exist new appearing error to indicate match or not.

This modification will bring the supporting for pattern, type, const and so on, for example, the schema we can use
"minItemsWhen": [
                  {
                    "value": 1,
                    "condition": {"properties": {"isCorrect": {"pattern": "^a.*"}}}
                  }
                ],

hejingkan2005: Very good suggestion, I have updated the code. Thanks @xichan-ms

src/docfx/lib/schema/JsonSchema.cs
```diff
@@ -262,17 +262,29 @@ internal class JsonSchema
...
+        /// <summary>
+        /// Whether content fallback is allowed for loc page
+        /// </summary>
+        public bool ContentFallback { get; set; } = true;
+
+        /// <summary>
+        /// Properties that are used to hold min item count that meet condition.
+        /// </summary>
+        [JsonConverter(typeof(ValueTupleConverter<JsonSchema, int>), "condition", "value")]
+        public (JsonSchema, int)[] MinItemsWhen { get; set; } = Array.Empty<(JsonSchema, int)>();
```
yufeih, line 274: Consider having a dedicated type and apply `OneOrManyConverter`?

src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -262,6 +264,65 @@ private void ValidateProperties(JsonSchema schema, string propertyPath, JObject
...
+                        $"The array must have least {value} matched item(s)."));
+                }
+            }
+        }
+
+        private bool SchemaMatches(JsonSchema schema, JToken map)
+        {
+            if (map is JObject obj)
+            {
+                var errors = Validate(schema, obj);
```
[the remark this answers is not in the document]
hejingkan2005: sure.

src/docfx/lib/schema/JsonSchema.cs
hejingkan2005: Ok, I've created a new type named 'ConditionalCheckSchema' for it.

yufeih (approved):

src/docfx/Errors.cs
```diff
@@ -497,6 +497,9 @@ public static Error FormatInvalid(SourceInfo? source, string value, object type,
             public static Error ArrayLengthInvalid(SourceInfo? source, string propName, string criteria)
                 => new Error(ErrorLevel.Warning, "array-length-invalid", $"Array '{propName}' length should be {criteria}.", source, propName);
 
+            public static Error ConditionalCheckInvalid(SourceInfo? source, string propName, string message)
```
yufeih, line 500: consider change the error code to include `array` to indicate this validation target is array, similar to other `array-xyz` error codes.

hejingkan2005: OK

PR #6623, opened 2020-10-10
Add Theme branch into publish manifest

928PJY:
[AB#319985](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/319985)

### DIFF
--- a/src/docfx/publish/PublishModelBuilder.cs
+++ b/src/docfx/publish/PublishModelBuilder.cs
@@ -79,6 +79,7 @@ namespace Microsoft.Docs.Build
                 _config.Name,
                 _config.Product,
                 _config.BasePath.ValueWithLeadingSlash,
+                _config.Template.IsMainOrMaster ? null : _config.Template.Branch,
                 items,
                 monikerGroups);
 
--- a/docs/specs/output.yml
+++ b/docs/specs/output.yml
@@ -74,6 +74,23 @@ outputs:
       "files": [{ "url": "/a", "ms.my": 1 }]
     }
 ---
+# Publish manifest contains theme_branch if template branch is not main or master
+repos:
+  https://docs.com/markdown-note:
+  - files:
+      docsets.yml:
+      a/docfx.yml: 'template: https://docs.com/template#master'
+      b/docfx.yml: 'template: https://docs.com/template#main'
+      c/docfx.yml: 'template: https://docs.com/template#test'
+  https://docs.com/template#main:
+  - files:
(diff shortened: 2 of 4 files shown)


docs/specs/template.yml
```diff
@@ -15,6 +15,22 @@ repos:
 outputs:
   docs/a.json: |
     { "conceptual": "<div class=\"NOTE\"><p>my note</p><p>note</p></div>" }
+  .publish.json: |
+    {"theme_branch": "main"}
+---
```
herohua, line 20: shall we add a case of none-master/main branch?

928PJY: updated

src/docfx/template/TemplateEngine.cs
```diff
@@ -63,6 +66,19 @@ public RenderType GetRenderType(ContentType contentType, SourceInfo<string?> mim
             };
         }
 
+        public string? GetThemeBranch()
+        {
+            if (_package is LocalPackage package)
+            {
+                var (themeRepo, _) = _repositoryProvider.GetRepository(PathString.DangerousCreate($"{package.PackagePath}/"));
```
herohua, line 73: > _repositoryProvider.GetRepository(PathString.DangerousCreate($"{package.PackagePath}/")); [](start = 36, length = 90)

This looks like internal logic of `LocalPackage`. Would it be better let package returns branch info?

partychen (approved):

src/docfx/template/TemplateEngine.cs
928PJY: but the package is not always a repository, even for localPackage, it is more related to the template, so I put it here.

herohua (approved):
:shipit:

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -79,6 +82,7 @@ orderby monikerGroup
                 _config.Name,
                 _config.Product,
                 _config.BasePath.ValueWithLeadingSlash,
+                _templateEngine.GetThemeBranch(),
```
yufeih, line 85: Is it suffice enough to say:
```suggestion
                _config.Template.Branch,
```

yufeih:
> No pipelines are associated with this pull request.

Hit the same problem described here: https://developercommunity.visualstudio.com/content/problem/1211721/azp-run-does-not-find-associated-pipelines-for-pr.html

yufeih (approved):

PR #6628, opened 2020-10-12
Overview page fix

AngryBerryMS:
- Remove uid and href of service page model
  - [AB#305351](https://dev.azure.com/ceapex/Engineering/_workitems/edit/305351)
  - uids in TopLevelTOC aren't added to InternalXrefBuilder, so remove them only affect the xrefmap.json 
  - href of a service page is useless. We only care about the hrefs of children items.
- Add href to reference toc of overview pages
  - [AB#312579](https://dev.azure.com/ceapex/Engineering/_workitems/edit/312579)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6628)


src/docfx/build/toc/ServicePageGenerator.cs
```diff
@@ -57,41 +51,61 @@ internal class ServicePageGenerator
...
                     if (!string.IsNullOrEmpty(childHref) && (childHrefType == TocHrefType.RelativeFolder || childHrefType == TocHrefType.TocFile))
                     {
                         childHref = null;
                     }
 
-                    if (!string.IsNullOrEmpty(childHref))
+                    if (item.Value.LandingPageType.Value != null)
                     {
-                        if (!(childHref.StartsWith("~/") || childHref.StartsWith("~\\")))
+                        if (!string.IsNullOrEmpty(childHref))
```
live1206, line 80: Would it be clearer to put `if (!string.IsNullOrEmpty(childHref))` as outer check?
The check of `if (item.Value.LandingPageType.Value != null)` only affects the value of `childHref`, maybe we can put it into `GetHrefRelativeToServicePage`?

src/docfx/build/toc/ServicePageGenerator.cs
```diff
@@ -57,41 +51,61 @@ internal class ServicePageGenerator
                     servicePagePath = FilePath.Generated(new PathString($"./{baseDir}/{directoryName}/{filename}.yml"));
                 }
 
+                if (string.IsNullOrEmpty(node.Href.Value))
```
live1206, line 54: `Uid` is not needed for service page, why do we need `Href`?

AngryBerryMS: This is for overview page. Although the service pages themselves don't need href of themselves, the "Overview" item should be added to ReferenceToc to get correct link to Overview page. And the link of "Overview" item is the same as the parent item, so the correct href should be added.

src/docfx/build/toc/ServicePageGenerator.cs
AngryBerryMS: If we do that, we still cannot put `if (item.Value.LandingPageType.Value != null)` into `GetHrefRelativeToServicePage`. Because when `childHref` isn't empty, we also need to generated href instead of using uid when children are also service pages.

live1206 (approved):

PR #6639, opened 2020-10-13
Fix wrong source info for codescnipet in include file

yufeih:
[AB#320919](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/320919)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6639)

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -795,3 +795,20 @@ inputs:
     tel:
 outputs:
   index.json: 
+---
+# Report correct line number for unknown-language-code in token
+inputs:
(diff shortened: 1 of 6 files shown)


yufeih:
The diffs are column changes from 0 -> 1, which is kind of expected. We might consider unify the meaning of 0 and 1 for line numbers.

bianyifan (approved):

PR #6651, opened 2020-10-15
Export schemaType to file xrefmap.json

xichan-ms:
[AB#297519](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/297519)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6651)

### DIFF
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -632,6 +632,6 @@ outputs:
   .xrefmap.json: | 
     {
       "references":[
-        {"uid": "a", "href": "https://docs.com/a?branch=export-schemaType-to-xrefmap", "schemaType": "testdata"}
+        {"uid": "a", "schemaType": "testdata"}
       ]
     }
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -614,3 +614,24 @@ outputs:
   .errors.log: | 
    { "message_severity":"error","code":"array-check-invalid","message":"The array must have least 1 matched item(s).","file":"unit.yml", "line":5,"column":3}
    { "message_severity":"error","code":"array-check-invalid","message":"The array must not have more than 1 matched item(s).","file":"unit.yml", "line":15,"column":3}
+---
+# export schemaType to xrefmap.json
+inputs:
+  docfx.yml:
+  a.yml: |
+    ### YamlMime:TestData
+    uid: a
+  _themes/ContentTemplate/schemas/TestData.schema.json: |
+    {
+      "properties": {
+        "uid" : {"contentType": "uid"}
+      }
+    }
+outputs:
+  a.json:
+  .xrefmap.json: | 
+    {
+      "references":[
+        {"uid": "a", "href": "https://docs.com/a?branch=export-schemaType-to-xrefmap", "schemaType": "testdata"}
+      ]
+    }
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -632,6 +632,6 @@ outputs:
   .xrefmap.json: | 
     {
       "references":[
-        {"uid": "a", "schemaType": "testdata"}
+        {"uid": "a", "schemaType": "TestData"}
       ]
     }
--- a/src/docfx/build/xref/InternalXrefSpec.cs
+++ b/src/docfx/build/xref/InternalXrefSpec.cs
@@ -12,6 +12,8 @@ namespace Microsoft.Docs.Build
     {
         public SourceInfo<string> Uid { get; }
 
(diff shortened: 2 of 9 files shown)


src/docfx/build/xref/InternalXrefSpec.cs
```diff
@@ -12,6 +12,8 @@ internal class InternalXrefSpec : IXrefSpec
     {
         public SourceInfo<string> Uid { get; }
 
+        public string? SchemaType { get; private set; }
```
yufeih, line 15: Add `string? SchemaType { get; }` to `IXrefSpec`?

src/docfx/lib/schema/JsonSchema.cs
```diff
@@ -11,6 +11,11 @@ namespace Microsoft.Docs.Build
     [JsonConverter(typeof(JsonSchemaConverter))]
     internal class JsonSchema
     {
+        /// <summary>
```
yufeih, line 14: nit: This is a custom extension, move it to line 193?

src/docfx/template/TemplateEngine.cs
```diff
@@ -204,6 +204,7 @@ private RenderType GetTocRenderType()
             }
 
             var jsonSchema = JsonUtility.DeserializeData<JsonSchema>(schemaString, new FilePath(schemaFilePath));
+            jsonSchema.SchemaType = mime.ToLowerInvariant();
```
yufeih, line 207: If `SchemaType` is a part of `JsonSchema` definition, that means people can override it in the schema definition. Is it intentional? If so, only assign the value if there is no existing `SchemaType`.

yufeih, line 207: What is the purpose of to lower here? We typically preserve the case of mime.

xichan-ms: OK.

src/docfx/build/xref/InternalXrefSpec.cs
xichan-ms: OK.

src/docfx/lib/schema/JsonSchema.cs
xichan-ms: OK.

docs/specs/schema.yml
```diff
@@ -614,3 +614,24 @@ outputs:
...
+      "properties": {
+        "uid" : {"contentType": "uid"}
+      }
+    }
+outputs:
+  a.json:
+  .xrefmap.json: | 
+    {
+      "references":[
+        {"uid": "a", "schemaType": "testdata"}
```
OsmondJiang, line 635: >schemaType [](start = 22, length = 10)

how about call it "type"?

yufeih: That would likely collide with existing properties. Lots of properties we use is already called `type`

src/docfx/template/TemplateEngine.cs
xichan-ms: OK,  have removed it and added fallback logic.

docs/specs/schema.yml
xichan-ms: OK, consider it in another PR.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -175,7 +174,13 @@ public JToken TransformContent(ErrorBuilder errors, JsonSchema schema, FilePath
...
             var href = GetXrefHref(file, uid, uidCount, obj.Parent == null);
             var monikers = _monikerProvider.GetFileLevelMonikers(errors, file);
-            var xref = new InternalXrefSpec(uid, href, file, monikers, obj.Parent?.Path);
+            var xref = new InternalXrefSpec(
+                uid,
+                href,
+                file,
+                monikers,
+                obj.Parent?.Path,
+                schema == rootSchema ? schema.SchemaType is null ? _documentProvider.GetMime(file).Value : schema.SchemaType : null);
```
yufeih, line 183: Use `schema.SchemaType` when it is not root schema.

```suggestion
                schema.SchemaType ?? (schema == rootSchema ? _documentProvider.GetMime(file).Value : null);
```

xichan-ms: OK.

yufeih (approved):
👍

PR #6658, opened 2020-10-15
Support xrefType Validation

xichan-ms:
[AB#297519](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/297519)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6658)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -392,8 +392,8 @@ namespace Microsoft.Docs.Build
             public static Error XrefNotFound(SourceInfo<string> source)
                 => new Error(ErrorLevel.Warning, "xref-not-found", $"Cross reference not found: '{source}'.", source);
 
-            public static Error InvalidXrefType(SourceInfo<string> xref, string expectedXrefType, string? realXrefType)
-               => new Error(ErrorLevel.Warning, "xref-invalid-type", $"Cross reference '{xref}' invalid type, '{expectedXrefType}' is expected, but now is '{realXrefType}'.", xref);
+            public static Error XrefTypeInvalid(SourceInfo<string> xref, string expectedXrefType, string? actualXrefType)
+               => new Error(ErrorLevel.Warning, "xref-type-invalid", $"Invalid cross reference: '{xref}'. Expected type '{expectedXrefType}' but got '{actualXrefType}'.", xref);
 
             public static Error UidNotFound(string uid, IEnumerable<string?> repositories)
                 => new Error(ErrorLevel.Warning, "uid-not-found", $"UID '{uid}' not found, which is referenced by repository {StringUtility.Join(repositories)}.");
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -392,6 +392,9 @@ namespace Microsoft.Docs.Build
             public static Error XrefNotFound(SourceInfo<string> source)
                 => new Error(ErrorLevel.Warning, "xref-not-found", $"Cross reference not found: '{source}'.", source);
 
+            public static Error InvalidXrefType(SourceInfo<string> xref, string expectedXrefType, string? realXrefType)
+               => new Error(ErrorLevel.Warning, "xref-invalid-type", $"Cross reference '{xref}' invalid type, '{expectedXrefType}' is expected, but now is '{realXrefType}'.", xref);
+
             public static Error UidNotFound(string uid, IEnumerable<string?> repositories)
                 => new Error(ErrorLevel.Warning, "uid-not-found", $"UID '{uid}' not found, which is referenced by repository {StringUtility.Join(repositories)}.");
 
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -174,13 +174,18 @@ namespace Microsoft.Docs.Build
         {
             var href = GetXrefHref(file, uid, uidCount, obj.Parent == null);
             var monikers = _monikerProvider.GetFileLevelMonikers(errors, file);
-            var xref = new InternalXrefSpec(
-                uid,
-                href,
-                file,
-                monikers,
-                obj.Parent?.Path,
-                schema.SchemaType ?? (schema == rootSchema ? _documentProvider.GetMime(file).Value : null));
+
+            string? schemaType = uidSchema.SchemaType;
+            if (schemaType is null && schema.SchemaTypeProperty != null && obj.TryGetValue(schema.SchemaTypeProperty, out var type))
+            {
+                schemaType = type.Value<string>();
+            }
+            else if (schemaType is null && schema == rootSchema)
+            {
+                schemaType = _documentProvider.GetMime(file);
+            }
+
+            var xref = new InternalXrefSpec(uid, href, file, monikers, obj.Parent?.Path, schemaType);
 
             if (uidSchema.UIDGlobalUnique)
             {
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -175,22 +175,21 @@ namespace Microsoft.Docs.Build
             var href = GetXrefHref(file, uid, uidCount, obj.Parent == null);
             var monikers = _monikerProvider.GetFileLevelMonikers(errors, file);
 
-            string? schemaType = uidSchema.SchemaType;
-            if (schemaType is null && schema.SchemaTypeProperty != null && obj.TryGetValue(schema.SchemaTypeProperty, out var type))
+            var schemaType = uidSchema.SchemaType;
+            if (schemaType is null)
             {
-                schemaType = type.Value<string>();
-            }
-            else if (schemaType is null && schema == rootSchema)
-            {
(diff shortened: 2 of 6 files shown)


src/docfx/Errors.cs
```diff
@@ -392,6 +392,9 @@ public static Error AtXrefNotFound(SourceInfo<string> source)
             public static Error XrefNotFound(SourceInfo<string> source)
                 => new Error(ErrorLevel.Warning, "xref-not-found", $"Cross reference not found: '{source}'.", source);
 
+            public static Error InvalidXrefType(SourceInfo<string> xref, string expectedXrefType, string? realXrefType)
```
yufeih, line 395: make the method name same as the error code.

src/docfx/Errors.cs
```diff
@@ -392,6 +392,9 @@ public static Error AtXrefNotFound(SourceInfo<string> source)
...
+               => new Error(ErrorLevel.Warning, "xref-invalid-type", $"Cross reference '{xref}' invalid type, '{expectedXrefType}' is expected, but now is '{realXrefType}'.", xref);
```
yufeih, line 396: Change the error code to `xref-type-invalid` to align with other error code style, and also change the message to match `unexpected-type` (which has been reviewed)
```suggestion
               => new Error(ErrorLevel.Warning, "xref-type-invalid", $"Invalid cross reference: '{xref}'. Expected type '{expectedXrefType}' but got '{actualXrefType}'.", xref);
```

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -174,13 +174,18 @@ public JToken TransformContent(ErrorBuilder errors, JsonSchema schema, FilePath
...
             var monikers = _monikerProvider.GetFileLevelMonikers(errors, file);
-            var xref = new InternalXrefSpec(
-                uid,
-                href,
-                file,
-                monikers,
-                obj.Parent?.Path,
-                schema.SchemaType ?? (schema == rootSchema ? _documentProvider.GetMime(file).Value : null));
+
+            string? schemaType = uidSchema.SchemaType;
```
yufeih, line 178: nit: prefer `var`.
```suggestion
            var schemaType = uidSchema.SchemaType;
```

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -174,13 +174,18 @@ public JToken TransformContent(ErrorBuilder errors, JsonSchema schema, FilePath
...
+            if (schemaType is null && schema.SchemaTypeProperty != null && obj.TryGetValue(schema.SchemaTypeProperty, out var type))
+            {
+                schemaType = type.Value<string>();
+            }
+            else if (schemaType is null && schema == rootSchema)
+            {
+                schemaType = _documentProvider.GetMime(file);
+            }
+
+            var xref = new InternalXrefSpec(uid, href, file, monikers, obj.Parent?.Path, schemaType);
```
yufeih, line 188: rename `UIDGlobalUnique` to `UIDGlobalUnique` because the existing casing convention is `Uid`, make `UidGlobalUnique` a get only property similar to other properties in `InternalXrefSpec`. In general, we use immutable models.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -174,13 +174,18 @@ public JToken TransformContent(ErrorBuilder errors, JsonSchema schema, FilePath
...
-                href,
-                file,
-                monikers,
-                obj.Parent?.Path,
-                schema.SchemaType ?? (schema == rootSchema ? _documentProvider.GetMime(file).Value : null));
+
+            string? schemaType = uidSchema.SchemaType;
+            if (schemaType is null && schema.SchemaTypeProperty != null && obj.TryGetValue(schema.SchemaTypeProperty, out var type))
+            {
+                schemaType = type.Value<string>();
```
yufeih, line 181: Will `Value<string>()` throw when the value is not a `string`? Refer to the implementation of `GetXrefPropertyValueAsString` for a safe fallback.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -174,13 +174,18 @@ public JToken TransformContent(ErrorBuilder errors, JsonSchema schema, FilePath
...
-            var xref = new InternalXrefSpec(
-                uid,
-                href,
-                file,
-                monikers,
-                obj.Parent?.Path,
-                schema.SchemaType ?? (schema == rootSchema ? _documentProvider.GetMime(file).Value : null));
+
+            string? schemaType = uidSchema.SchemaType;
+            if (schemaType is null && schema.SchemaTypeProperty != null && obj.TryGetValue(schema.SchemaTypeProperty, out var type))
```
yufeih, line 179: When `SchemaTypeProperty` exists, should not fallback to mime type in any case, so the structure might look like below. Also consider extracting get schema type into a new method.


```
if (schemaType is null)
{
    if (schemaTypeProperty != null)
    {
    }
    else if (schema == rootSchema)
    {
    }
}
```

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -424,6 +429,15 @@ private string GetXrefHref(FilePath file, string uid, int uidCount, bool isRootL
 
                         errors.AddIfNotNull(xrefError);
 
+                        if (xrefSpec != null &&
+                        schema.XrefType != null &&
+                        schema.ContentType == JsonSchemaContentType.Xref &&
+                        (xrefSpec.SchemaType is null ||
+                        !xrefSpec.SchemaType.Equals(schema.XrefType)))
+                        {
+                            errors.AddIfNotNull(Errors.Xref.InvalidXrefType(content, schema.XrefType!, xrefSpec.SchemaType));
```
yufeih, line 438: Don't use `!` unless you absolutely need to!
```suggestion
                            errors.Add(Errors.Xref.InvalidXrefType(content, schema.XrefType, xrefSpec.SchemaType));
```

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -424,6 +429,15 @@ private string GetXrefHref(FilePath file, string uid, int uidCount, bool isRootL
 
                         errors.AddIfNotNull(xrefError);
 
+                        if (xrefSpec != null &&
+                        schema.XrefType != null &&
```
yufeih, line 433: Use the same indention as line 432, keep them left align (without `if (`)

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -424,6 +429,15 @@ private string GetXrefHref(FilePath file, string uid, int uidCount, bool isRootL
...
+                        schema.ContentType == JsonSchemaContentType.Xref &&
+                        (xrefSpec.SchemaType is null ||
+                        !xrefSpec.SchemaType.Equals(schema.XrefType)))
```
yufeih, line 436: Some checks are not necessary, and also use case insensitive comparision.
```suggestion
                        if (xrefSpec != null && schema.XrefType != null && !schema.XrefType.Equals(xrefSpec.SchemaType, StringComparison.OrdinalIgnoreCase)))
```

src/docfx/Errors.cs
xichan-ms: OK.

src/docfx/Errors.cs
xichan-ms: ok.

src/docfx/lib/schema/JsonSchemaTransformer.cs
xichan-ms: ok.

src/docfx/lib/schema/JsonSchemaTransformer.cs
xichan-ms: OK

src/docfx/lib/schema/JsonSchemaTransformer.cs
xichan-ms: OK.

src/docfx/lib/schema/JsonSchemaTransformer.cs
xichan-ms: OK

src/docfx/lib/schema/JsonSchemaTransformer.cs
xichan-ms: OK

src/docfx/lib/schema/JsonSchemaTransformer.cs
xichan-ms: OK

src/docfx/lib/schema/JsonSchemaTransformer.cs
xichan-ms: ok.

yufeih (approved):

PR #6661, opened 2020-10-16
Split toc default by name

AngryBerryMS:
Split specified toc.yml in ops.config by name by default.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6661)

### DIFF
--- a/src/docfx/build/toc/TableOfContentsMap.cs
+++ b/src/docfx/build/toc/TableOfContentsMap.cs
@@ -221,7 +221,7 @@ namespace Microsoft.Docs.Build
 
         private void SplitToc(FilePath file, TableOfContentsNode toc, ConcurrentBag<FilePath> result)
         {
-            if (Array.Exists<string>(_config.SplitTOC, e => e.Equals(file.Path.Value)) || toc.Items.Count <= 0)
+            if (!Array.Exists(_config.SplitTOC, e => e.Equals(file.Path.Value)) || toc.Items.Count <= 0)
             {
                 result.Add(file);
                 return;
--- a/src/docfx/build/toc/TableOfContentsMap.cs
+++ b/src/docfx/build/toc/TableOfContentsMap.cs
@@ -221,7 +221,7 @@ namespace Microsoft.Docs.Build
 
         private void SplitToc(FilePath file, TableOfContentsNode toc, ConcurrentBag<FilePath> result)
         {
-            if (!Array.Exists(_config.SplitTOC, e => e.Equals(file.Path.Value)) || toc.Items.Count <= 0)
+            if (!_config.SplitTOC.Contains(file.Path) || toc.Items.Count <= 0)
             {
                 result.Add(file);
                 return;
--- a/src/docfx/build/toc/TableOfContentsMap.cs
+++ b/src/docfx/build/toc/TableOfContentsMap.cs
@@ -15,6 +15,7 @@ namespace Microsoft.Docs.Build
     /// </summary>
     internal class TableOfContentsMap
     {
+        private readonly Config _config;
         private readonly Input _input;
         private readonly ErrorBuilder _errors;
         private readonly BuildScope _buildScope;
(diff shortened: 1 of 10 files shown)


src/docfx/config/ops/OpsDocsetConfig.cs
```diff
@@ -22,6 +22,9 @@ internal class OpsDocsetConfig
 
         public Dictionary<string, string[]>? CustomizedTasks { get; private set; }
 
+        [JsonProperty("SplitTOC")]
```
yufeih, line 25: nit: it is a bit inconsistent that "SplitTOC" is a string but JoinTOCPluginand ECMA2Yaml uses `nameof`, would we convert all the `nameof`s to strings?

src/docfx/build/toc/TableOfContentsMap.cs
```diff
@@ -218,7 +221,7 @@ where tocToTocs.ContainsKey(item.Key) && !item.Key.IsExperimental()
 
         private void SplitToc(FilePath file, TableOfContentsNode toc, ConcurrentBag<FilePath> result)
         {
-            if (string.IsNullOrEmpty(toc.SplitItemsBy) || toc.Items.Count <= 0)
```
live1206, line 221: remove `SplitItemsBy` from `TableOfContentsNode` since it is not in use anymore

src/docfx/config/Config.cs
```diff
@@ -320,6 +320,8 @@ internal class Config : PreloadConfig
 
         public JoinTOCConfig[] JoinTOC { get; private set; } = Array.Empty<JoinTOCConfig>();
 
+        public string[] SplitTOC { get; private set; } = Array.Empty<string>();
```
yufeih, line 323: This could be defined as `HashSet<PathString>` to make `SplitToc` become a simple `Contains` check.
```suggestion
        public HashSet<PathString> SplitTOC { get; private set; } = new HashSet<PathString>();
```

src/docfx/config/ops/OpsDocsetConfig.cs
live1206, line 25:
```suggestion
        [JsonProperty(nameof(SplitTOC))]
```

src/docfx/build/toc/TableOfContentsMap.cs
```diff
@@ -218,7 +221,7 @@ where tocToTocs.ContainsKey(item.Key) && !item.Key.IsExperimental()
...
+            if (!Array.Exists(_config.SplitTOC, e => e.Equals(file.Path.Value)) || toc.Items.Count <= 0)
```
live1206, line 224: should `file.Path.Value` be case insensitive?

src/docfx/config/ops/OpsConfigLoader.cs
```diff
@@ -81,6 +81,8 @@ internal static class OpsConfigLoader
                 {
                     ["open_to_public_contributors"] = docsetConfig.OpenToPublicContributors,
                 };
+
+                result["SplitTOC"] = JArray.FromObject(docsetConfig.SplitTOC);
```
live1206, line 85:
```suggestion
                result["splitTOC"] = JArray.FromObject(docsetConfig.SplitTOC);
```

src/docfx/config/Config.cs
live1206: +1, should this be case insensitive?

src/docfx/build/toc/TableOfContentsMap.cs
AngryBerryMS: No.

src/docfx/config/Config.cs
AngryBerryMS: I thought it should be case sensitive for two file named api/toc.yml and API/TOC.YML can be placed into the build scope, they are different files.

yufeih: @AngryBerryMS , the value here is a file path. File path is currently case insensitive on windows and mac but case sensitive on linux (the behavior might change in the future). This logic is encapsulated in `PathString` and it is a good practice to use the strongly typed `PathString` to represent file path wherever you can.

yufeih (approved):

live1206 (approved):

PR #6685, opened 2020-10-23
Support '--template-base-path' to specific the template base path when apply liquid

928PJY:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6685)

### DIFF
--- a/src/docfx/cli/Docfx.cs
+++ b/src/docfx/cli/Docfx.cs
@@ -129,6 +129,10 @@ namespace Microsoft.Docs.Build
                     syntax.DefineOption("no-dry-sync", ref options.NoDrySync, "Do not run dry sync for learn validation.");
                     syntax.DefineOption("no-restore", ref options.NoRestore, "Do not restore dependencies before build.");
                     syntax.DefineOption("no-cache", ref options.NoCache, "Do not use cache dependencies in build, always fetch latest dependencies.");
+                    syntax.DefineOption(
+                        "template-base-path",
+                        ref options.TemplateBasePath,
+                        "The base path used for referencing the template resource file when applying liquid.");
                     DefineCommonOptions(syntax, ref workingDirectory, options);
                 });
 
--- a/src/docfx/template/liquid/JavaScriptTag.cs
+++ b/src/docfx/template/liquid/JavaScriptTag.cs
@@ -10,7 +10,8 @@ namespace Microsoft.Docs.Build
     {
         public override void Render(DotLiquid.Context context, TextWriter result)
         {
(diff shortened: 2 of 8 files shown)


src/docfx/cli/Docfx.cs
```diff
@@ -129,6 +129,10 @@ private static (string command, string workingDirectory, CommandLineOptions opti
                     syntax.DefineOption("no-dry-sync", ref options.NoDrySync, "Do not run dry sync for learn validation.");
                     syntax.DefineOption("no-restore", ref options.NoRestore, "Do not restore dependencies before build.");
                     syntax.DefineOption("no-cache", ref options.NoCache, "Do not use cache dependencies in build, always fetch latest dependencies.");
+                    syntax.DefineOption(
```
yufeih, line 132: move to `DefineCommonOptions` because `--template` is also defined there?

src/docfx/template/liquid/JavaScriptTag.cs
```diff
@@ -10,7 +10,8 @@ internal class JavaScriptTag : Tag
     {
         public override void Render(DotLiquid.Context context, TextWriter result)
         {
-            result.Write($@"<script src=""{LiquidTemplate.GetThemeRelativePath(Markup)}"" ></script>");
+            var templateBasePath = (string?)context.Registers["template_base_path"];
+            result.Write($@"<script src=""{LiquidTemplate.GetThemeRelativePath(templateBasePath, Markup)}"" ></script>");
```
yufeih, line 14: consider pass `context` to `GetThemeRelativePath` method to reuse the template base path retrieval logic?
```suggestion
            result.Write($@"<script src=""{LiquidTemplate.GetThemeRelativePath(context, Markup)}"" ></script>");
```

yufeih (commented):
Add a simple spec test?

src/docfx/cli/Docfx.cs
928PJY: but `--template-base-path` can be used separately, and it is not useful for other commands except `build`.

yufeih: It is not useful for `restore` but is useful for other `build` like commands e.g., `serve` (which we currently don't have). I might be overthinking, we could figure it out later.

928PJY: OK, but it still not a common option for all the command~😃

yufeih (approved):

PR #6691, opened 2020-10-26
By default build referenced resource in CRR dependency package

mingwli:
[AB#326400](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/326400)

- generate & store CRR referenced AdditionalResources in LinkResolver
- parallel build AdditionalResources & add to PublishUrlMap
- restore AdditionalResources PublishItems from PublishUrlMap when building PublishModel

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6691)

### DIFF
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -86,6 +86,11 @@ namespace Microsoft.Docs.Build
                     context.ErrorBuilder,
                     context.PublishUrlMap.GetAllFiles(),
                     file => BuildFile(context, file));
+
+                ParallelUtility.ForEach(
+                    context.ErrorBuilder,
+                    context.LinkResolver.GetAdditionalResources(),
+                    resource => BuildResource.Build(context, resource, true));
             }
 
             Parallel.Invoke(
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Docs.Validation;
 
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -85,8 +85,13 @@ namespace Microsoft.Docs.Build
             {
                 ParallelUtility.ForEach(
                     context.ErrorBuilder,
-                    context.PublishUrlMap.GetAllFiles().Concat(context.LinkResolver.GetAdditionalResources()),
+                    context.PublishUrlMap.GetAllFiles(),
                     file => BuildFile(context, file));
+
+                ParallelUtility.ForEach(
+                    context.ErrorBuilder,
+                    context.LinkResolver.GetAdditionalResources(),
+                    file => BuildResource.Build(context, file));
             }
 
             Parallel.Invoke(
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -90,7 +90,7 @@ namespace Microsoft.Docs.Build
                 ParallelUtility.ForEach(
                     context.ErrorBuilder,
                     context.LinkResolver.GetAdditionalResources(),
-                    resource => BuildResource.Build(context, resource, true));
+                    resource => BuildResource.Build(context, resource));
             }
 
             Parallel.Invoke(
--- a/src/docfx/build/Build.cs
+++ b/src/docfx/build/Build.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.Docs.Validation;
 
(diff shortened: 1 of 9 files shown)


src/docfx/build/resource/BuildResource.cs
```diff
@@ -22,6 +22,10 @@ internal static void Build(Context context, FilePath file)
             }
 
             context.PublishModelBuilder.SetPublishItem(file, metadata: null, outputPath);
+            if (isAdditionalResource)
+            {
+                context.PublishUrlMap.AddAdditionalResource(file);
```
yufeih, line 27: consider  let PublishModelBuilder call LinkResolver.GetAdditionalResources on building publish manifest?

With that this method doesn't need change

mingwli: This is feasible. It's just that `PublishModelBuilder` depending on `LinkResolver` feels a bit weird.

src/docfx/build/link/LinkResolver.cs
```diff
@@ -147,6 +151,11 @@ internal class LinkResolver
 
             if (error is null && _buildScope.OutOfScope(file))
             {
+                if (file.Origin == FileOrigin.Dependency && _buildScope.GetContentType(file) == ContentType.Resource)
```
live1206, line 154: how about updating [BuildScope.cs](https://github.com/dotnet/docfx/blob/a6583ad73bb75f83b4c252a4bcd42c21685d3c60/src/docfx/build/context/BuildScope.cs#L182) instead? Include the dependency resources to the build scope, the pipeline should just work.

mingwli: As discussed, that way all CRR resource will be built and published

live1206: Offline synced with @mingwli , the dynamic scope of dependency resources also breaks the design presumption of `PublishUrlMap`, which works on static build scope.
Consider this case, if the user defines a redirect URL of existed resource file from dependency, docfx would not be able to detect the output path conflict.

live1206: NOT all CRR resource will be published if the user config the right scope.

src/docfx/build/Build.cs
```diff
@@ -86,6 +86,11 @@ private static void Run(Context context)
                     context.ErrorBuilder,
                     context.PublishUrlMap.GetAllFiles(),
```
live1206, line 87: >context.PublishUrlMap.GetAllFiles() [](start = 20, length = 35)

How about concat the additional resource files on the original scope?

src/docfx/publish/PublishUrlMap.cs
```diff
@@ -65,6 +65,13 @@ public IEnumerable<(string url, FilePath sourcePath, MonikerList monikers)> GetP
             return _publishUrlMap.Values.SelectMany(x => x).Select(x => (x.Url, x.SourcePath, x.Monikers));
         }
 
+        public PublishUrlMapItem GeneratePublishUrlMapItem(FilePath path)
```
live1206, line 68: It's a bit strange to expose `GeneratePublishUrlMapItem` to others.
Can we just construct the needed fields in the usage instead?

mingwli: you mean passing those `_documentProvider`, `_monikerProvider` dependencies into `PublishModelBuilder`

src/docfx/build/Build.cs
mingwli: Actually it can't, for the additional resources are ONLY found after the static scope get built.

src/docfx/build/link/LinkResolver.cs
```diff
@@ -107,6 +111,8 @@ internal class LinkResolver
             return (error, link, file);
         }
 
+        public IReadOnlyCollection<FilePath> GetAdditionalResources() => _additionalResources.ToHashSet();
```
yufeih, line 114: Eliminate the unneeded `ToHashSet` call
```suggestion
        public IEnumerable<FilePath> GetAdditionalResources() => _additionalResources;
```

src/docfx/publish/PublishModelBuilder.cs
```diff
@@ -47,7 +50,8 @@ public void SetPublishItem(FilePath file, JObject? metadata, string? outputPath)
         public (PublishModel, Dictionary<FilePath, PublishItem>) Build()
         {
             var publishItems = new Dictionary<FilePath, PublishItem>();
-            foreach (var (url, sourcePath, monikers) in _publishUrlMapBuilder.GetPublishOutput())
+            foreach (var (url, sourcePath, monikers) in _publishUrlMap.GetAllFiles().Concat(_linkResolver.GetAdditionalResources())
```
yufeih, line 53: Consider computing the `url` and `monikers` in the loop body, also remove `PublishUrlMap.GetPublishOutput`?
```suggestion
            foreach (var sourcePath in _publishUrlMap.GetAllFiles().Concat(_linkResolver.GetAdditionalResources())
```

src/docfx/build/Build.cs
live1206: right

live1206 (approved):

yufeih (approved):

PR #6700, opened 2020-10-27
Encode html using FormattableString

yufeih:
Makes it easier to enforce html encoding.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6700)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -298,11 +298,11 @@ namespace Microsoft.Docs.Build
             {
                 if (attribute.NameIs("data-raw-html"))
                 {
-                    rawHtml = HttpUtility.HtmlDecode(attribute.Value.ToString());
+                    rawHtml = attribute.Value.ToString();
                 }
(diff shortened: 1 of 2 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -298,11 +298,11 @@ public static bool IsInlineImage(this HtmlBlock node, int imageIndex)
             {
                 if (attribute.NameIs("data-raw-html"))
                 {
-                    rawHtml = HttpUtility.HtmlDecode(attribute.Value.ToString());
+                    rawHtml = attribute.Value.ToString();
```
mingwli, line 301: why no longer decoding here?

live1206 (approved):

src/docfx/lib/HtmlUtility.cs
yufeih: because it'll be encoded anyway in line 329

mingwli (approved):

PR #6703, opened 2020-10-28
Disable TripleCrown for Default Locale

xichan-ms:
[AB#326519](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/326519)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6703)


xichan-ms:
This PR should be merged after the modification of docs-ui being accepted.

xichan-ms:
> ⚠[learn-m365-pr](https://github.com/MicrosoftDocs/learn-m365-pr)(00:00:15.1096426, 17893 more diff)

This is due to the not merging of docs-ui, and I run it locally with modify docs-ui, it works well.

src/Microsoft.Docs.LearnValidation/Validators/PathValidator.cs
```diff
@@ -26,54 +26,18 @@ public override bool Validate(Dictionary<string, IValidateModel> fullItemsDict)
...
+                // when trophy is defined in another path, but that path has error when SDP validating
+                if (path.Achievement is string achievementUID && !fullItemsDict.ContainsKey(achievementUID))
                 {
                     itemValid = false;
-                    Logger.Log(LearnErrorLevel.Error, LearnErrorCode.TripleCrown_LearningPath_ChildrenNotFound, file: item.SourceRelativePath, string.Join(",", childrenCantFind));
                 }
 
-                if (childrenNotModule.Any())
+                // path has child module, but that module has error when SDP validating, except the shared module
+                var childrenCantFind = path.Modules.Where(m => !fullItemsDict.ContainsKey(m) && !_externalXrefsCheck(m, "Module")).ToList();
```
yufeih, line 37: If SDP check fails, we should be able to pass a list of failed files to here and propagate errors accordingly. I don't quite follow the `_externalXrefsCheck` here.

cc @yishengjin1413

xichan-ms:
The diff in repo docs is due to the dev language 'aspx-csharp' is allowed.

image

src/Microsoft.Docs.LearnValidation/Validators/PathValidator.cs
xichan-ms: Resolve by offline talking.

yufeih (approved):

yishengjin1413:
@xichan-ms  Will the failed item be passed to the validation plugin?

yishengjin1413:
@xichan-ms For the token validation, it will only work for loc. And when the loc token is not existing, even it can fallback to en-us, we will also report a fail.  Does SDP validation did the same thing?

yishengjin1413 (approved):

xichan-ms:
> @xichan-ms Will the failed item be passed to the validation plugin?

not, items failed with sdp validation, can not be passed to plugin.

xichan-ms:
> @xichan-ms For the token validation, it will only work for loc. And when the loc token is not existing, even it can fallback to en-us, we will also report a fail. Does SDP validation did the same thing?

Now, it works for loc and non-loc. 
For non-loc, is also by design, the origin error is include-not-found and severity also is error.
For loc, here has the same logic though can fallback to en-us but throw an error.

PR #6712, opened 2020-10-29
Support JsonSchema anyof/oneof/allof/if/then/else/not

yufeih:
[AB#329209](https://dev.azure.com/ceapex/Engineering/_workitems/edit/329209/)

[AB#329703](https://dev.azure.com/ceapex/Engineering/_workitems/edit/329703)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6712)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaValidator.cs
+++ b/src/docfx/lib/schema/JsonSchemaValidator.cs
@@ -103,6 +103,8 @@ namespace Microsoft.Docs.Build
             ValidateAnyOf(schema, propertyPath, token, errors);
             ValidateAllOf(schema, propertyPath, token, errors);
             ValidateOneOf(schema, propertyPath, token, errors);
+            ValidateIfThenElse(schema, propertyPath, token, errors);
+            ValidateNot(schema, propertyPath, token, errors);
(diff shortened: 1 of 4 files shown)


src/docfx/lib/schema/JsonSchemaValidator.cs
```diff
@@ -578,6 +584,131 @@ private static void ValidateDeprecated(JsonSchema schema, string propertyPath, J
             }
         }
 
+        private void ValidateAnyOf(JsonSchema schema, string propertyPath, JToken token, List<Error> errors)
+        {
+            if (schema.AnyOf.Length <= 0)
```
live1206, line 589: nit: when would this be `<`?

live1206 (approved):

src/docfx/lib/schema/JsonSchemaValidator.cs
yufeih: all the other checks uses `<=`, same here :)

PR #6731, opened 2020-11-04
feat: treat .yml as .md in computation of document id for Architecture

herohua:
[AB#331835](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/331835)

Searched in metadata store - no .yml with YamlMime Architecture exists for live branch for now, so it is safe to apply this change without any side effect to existing document ids.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6731)

### DIFF
--- a/src/docfx/template/TemplateEngine.cs
+++ b/src/docfx/template/TemplateEngine.cs
@@ -74,7 +74,8 @@ namespace Microsoft.Docs.Build
         {
             return "Hub".Equals(mime, StringComparison.OrdinalIgnoreCase) ||
                    "Landing".Equals(mime, StringComparison.OrdinalIgnoreCase) ||
-                   "LandingData".Equals(mime, StringComparison.OrdinalIgnoreCase);
+                   "LandingData".Equals(mime, StringComparison.OrdinalIgnoreCase) ||
+                   "Architecture".Equals(mime, StringComparison.OrdinalIgnoreCase);
         }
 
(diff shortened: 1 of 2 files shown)


live1206:
Update [spec test](https://github.com/dotnet/docfx/blob/050830648f25b894528d8f21d5a97f589a4277dd/docs/specs/redirection.yml#L372) to cover the new mime type.

src/docfx/template/TemplateEngine.cs
```diff
@@ -74,7 +74,8 @@ public static bool IsMigratedFromMarkdown(string? mime)
         {
             return "Hub".Equals(mime, StringComparison.OrdinalIgnoreCase) ||
                    "Landing".Equals(mime, StringComparison.OrdinalIgnoreCase) ||
-                   "LandingData".Equals(mime, StringComparison.OrdinalIgnoreCase);
+                   "LandingData".Equals(mime, StringComparison.OrdinalIgnoreCase) ||
+                   "Architecture".Equals(mime, StringComparison.OrdinalIgnoreCase);
```
live1206, line 78: since this list grows, how about creating a hashset for it and use `Contains` to check instead?

live1206 (approved):

yufeih (approved):

PR #6740, opened 2020-11-06
expose git contributor information

AngryBerryMS:
[AB#331195](https://ceapex.visualstudio.com/Engineering/_workitems/edit/331195)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6740)

### DIFF
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -754,3 +754,60 @@ outputs:
       "original_content_git_url": "https://github.com/user-assigned/original/blob/original/index.md",
       "gitcommit": "https://github.com/testowner/testrepo/blob/450790f8a588ed8d96f4cb931c744d3855b7a809/index.md",
     }
+---
+# Put contributor list to .publish.json (Expose Contributor Information)
+repos:
+  https://github.com/expose-contributor/testrepo#master-file-cache:
+    - author: Charlie
+      email: charlie@contoso.com
+      files:
+        docs/index.md: |
+          ---
+          author: charlie
+          ---
+          Alice creates this. Charlie updates this. Bob updates this. Charlie becomes author.
+    - author: Bob 
+      email: bob@contoso.com
+      files:
+        docs/index.md: |
+          Alice creates this. Charlie updates this. Bob updates this.
+    - author: Charlie 
+      email: charlie@contoso.com
+      files:
+        docs/index.md: |
+          Alice creates this. Charlie updates this.
+    - author: Alice 
+      email: alice@contoso.com
+      files:
+        docs/index.md: |
+          Alice creates this.
+inputs:
+  docfx.yml:
+cache:
+  github-users.json: |
+      { "items": [
+          {"name": "Alice", "login": "alice", "id": 1, "emails": ["alice@contoso.com"]},
+          {"name": "Bob", "login": "bob", "id": 2, "emails": ["bob@contoso.com"]},
+          {"name": "Charlie", "login": "charlie", "id": 3, "emails": ["charlie@contoso.com"]}
+        ]
+      }
+outputs:
+  docs/index.json: |
+    {
+      "_op_gitContributorInformation":
+      {
+        "author":
+          { "name": "charlie", "profile_url": "https://github.com/charlie", "display_name": "Charlie", "id": "3" },
+        "contributors": [
+          { "name": "bob", "profile_url": "https://github.com/bob", "display_name": "Bob", "id": "2" },
+          { "name": "alice", "profile_url": "https://github.com/alice", "display_name": "Alice", "id": "1" }
+        ]
+      }
+    }
+  .publish.json: |
+    {
+      "files": [{
+        "path": "docs/index.json",
+        "git_contributor_information": ["charlie", "bob", "alice"]
+      }]
+    }
--- a/docs/specs/git.yml
+++ b/docs/specs/git.yml
@@ -811,3 +811,50 @@ outputs:
         "git_contributor_information": ["charlie", "bob", "alice"]
       }]
     }
+---
+# Author should NOT exist in git_contribor_information if the author haven't committed anything (Expose Condtributor Information)
+repos:
+  https://github.com/expose-contributor-without-author/testrepo#master-author:
+    - author: Bob 
+      email: bob@contoso.com
+      files:
+        docs/index.md: |
+          ---
+          author: charlie
+          ---
+          Alice creates this. Bob updates this, and thinks Charlie should be the author.
+    - author: Alice 
+      email: alice@contoso.com
+      files:
+        docs/index.md: |
+          Alice creates this.
+inputs:
+  docfx.yml:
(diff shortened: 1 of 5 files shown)


src/docfx/build/metadata/SystemMetadata.cs
```diff
@@ -31,6 +31,9 @@ internal class SystemMetadata
         [JsonProperty("_op_gitContributorInformation")]
         public ContributionInfo? ContributionInfo { get; set; }
 
+        [JsonProperty("git_contributor_information")]
```
mingwli, line 34: >   [JsonProperty("git_contributor_information")] [](start = 5, length = 48)

remove. it's already snake case naming convention

AngryBerryMS: Oh...Sorry...

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -33,7 +33,7 @@ internal class ContributionProvider
             _repositoryProvider = repositoryProvider;
         }
 
-        public ContributionInfo? GetContributionInfo(ErrorBuilder errors, FilePath file, SourceInfo<string> authorName)
+        public ContributionInfo? GetContributionInfo(ErrorBuilder errors, FilePath file, SourceInfo<string> authorName, bool removeAuthor = true)
```
mingwli, line 36: >bool removeAuthor = true [](start = 120, length = 24)

nit: bool includeAuthor = false. let default value be the default value

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -108,6 +108,14 @@ internal class ContributionProvider
             return contributionInfo;
         }
 
+        public string?[]? GitContributorInformation(ErrorBuilder errors, FilePath file)
+        {
+            return GetContributionInfo(errors, file, new SourceInfo<string>(""), true)?.Contributors?.
```
mingwli, line 113: can cache the `contributors` in line#75.
don't have to run the whole process twice. Or you can have the `GetContributionInfo` return the two entities.

docs/specs/git.yml
```diff
@@ -754,3 +754,60 @@ outputs:
...
+  https://github.com/expose-contributor/testrepo#master-file-cache:
+    - author: Charlie
+      email: charlie@contoso.com
+      files:
+        docs/index.md: |
+          ---
+          author: charlie
+          ---
+          Alice creates this. Charlie updates this. Bob updates this. Charlie becomes author.
+    - author: Bob
```
herohua, line 769: can we add a case that author is not one of the contributors?

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -108,6 +108,14 @@ internal class ContributionProvider
...
+                Select(e => e.Name).
```
herohua, line 114: minor: where to put `.`?

docs/specs/git.yml
AngryBerryMS: Added.

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -57,7 +57,7 @@ internal class ContributionProvider
 
             if (!_config.ResolveGithubUsers)
             {
-                return contributionInfo;
+                return (contributionInfo, contributionInfo.Contributors?.Select(e => e.Name).Where(e => !string.IsNullOrEmpty(e)).ToArray());
```
mingwli, line 60: this is always null : )

mingwli (approved):

src/docfx/build/metadata/SystemMetadata.cs
```diff
@@ -31,6 +31,8 @@ internal class SystemMetadata
         [JsonProperty("_op_gitContributorInformation")]
         public ContributionInfo? ContributionInfo { get; set; }
 
+        public string?[]? GitContributorInformation { get; set; }
```
yufeih, line 34: Use `string[]?` as the type, each array item must not be null.

yufeih, line 34: Consider name the metadata `github_contributors`:

- The existing `_op_gitContributorInformation` was named purely for backward compatibility, it does not adhere most of the naming conventions we had for metadata. Call this `git_contributor_information` would give you the false assumption that this is the same as `_op_gitContributorInformation` but it's actually not true, `author` is not included in the list.
- Remove the `_information` suffix because just about everything is a piece of information.
- The value is a list of _GitHub_ users not a list of _git_ users

docs/specs/legacy.yml
```diff
@@ -70,6 +70,7 @@ outputs:
         <meta name=\"depot_name\" content=\".\" />
         <meta name=\"document_id\" content=\"fdb5a0d7-b68e-c465-9c5e-54a5cda94902\" />
         <meta name=\"document_version_independent_id\" content=\"92a5a770-f05f-b8bb-81c9-a8d786a9e608\" />
+        <meta content=\"charlie\" name=\"git_contributor_information\">
```
yufeih, line 73: Add it to `OpsMetadata.json`. `htmlMetaHidden` section, this list does not need to appear in the final HTML.

docs/specs/git.yml
```diff
@@ -754,3 +754,107 @@ outputs:
       "original_content_git_url": "https://github.com/user-assigned/original/blob/original/index.md",
       "gitcommit": "https://github.com/testowner/testrepo/blob/450790f8a588ed8d96f4cb931c744d3855b7a809/index.md",
     }
+---
+# Put contributor list to .publish.json (Expose Contributor Information)
```
yufeih, line 758: This file contains a extensive suite of test cases for `_op_gitContributorInformation`. For this particular change, instead of adding new test cases, prefer add check to the new metadata in addition to `_op_gitContributorInformation` check.

It is optional to verify `.publish.json` since we have a orthogonal test case to ensure `.publish.json` contains all system metadata. Technically this change is about adding a new metadata, not adding a property to `.publish.json`.

src/docfx/build/metadata/SystemMetadata.cs
AngryBerryMS: The type of "Name" property in `Contributor.cs` is `string?`, thus we have to use `string?[]?` here

AngryBerryMS: I will update the spec and create a new PR to fix it.

docs/specs/git.yml
AngryBerryMS: Will create a new pr to do this.

docs/specs/legacy.yml
AngryBerryMS: Will create a new pr to do this.

PR #6746, opened 2020-11-09
Expose contributor fix comments

AngryBerryMS:
# Fix newly added comments in [https://github.com/dotnet/docfx/pull/6740](https://github.com/dotnet/docfx/pull/6740)

- Use "git_contributors" instead of "git_contributor_information"
- The type of "Name" property in Contributor.cs is string?, thus we have to use string?[]? here
- Add "git_contributors" to `htmlMetaHidden`
- Move the 2 spec tests to `metadata.yml`.
- I'd like to hold the `.publish.json` part in spec test to make others be clear with  what this feature does

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6746)


src/docfx/build/metadata/SystemMetadata.cs
```diff
@@ -31,7 +31,7 @@ internal class SystemMetadata
         [JsonProperty("_op_gitContributorInformation")]
         public ContributionInfo? ContributionInfo { get; set; }
 
-        public string?[]? GitContributorInformation { get; set; }
+        public string?[]? GitContributors { get; set; }
```
yufeih, line 34:
```suggestion
        public string[]? GithubContributors { get; set; }
```

AngryBerryMS: The type of "Name" property in Contributor.cs is `string?`, thus we have to use `string?[]?` here

yufeih (approved):

PR #6747, opened 2020-11-09
Load external xrefmap in parallel

yufeih:
Address some of the recent _azure-docs-pr_ perf regressions:

**Before**:

image

**After**:

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6747)

### DIFF
--- a/src/docfx/build/xref/ExternalXrefMapLoader.cs
+++ b/src/docfx/build/xref/ExternalXrefMapLoader.cs
@@ -2,12 +2,13 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IO;
 using System.IO.Compression;
-using System.Linq;
 using System.Text;
 using System.Text.Json;
+using System.Threading.Tasks;
 
 namespace Microsoft.Docs.Build
 {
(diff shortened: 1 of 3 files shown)


src/docfx/build/xref/ExternalXrefMapLoader.cs
```diff
@@ -48,26 +49,23 @@ public static ExternalXrefMap Load(Config config, FileResolver fileResolver, Err
...
-                externalXref.TrimExcess();
+                });
 
-                return new ExternalXrefMap(externalXrefMap, externalXref);
+                return new ExternalXrefMap(externalXrefMap.AsDictionary(), externalXref.AsList());
             }
         }
 
-        public static (List<(string, Lazy<ExternalXrefSpec>)> externalXrefSpec, List<ExternalXref> externalXref) LoadJsonFile(string filePath)
+        internal static (List<(string, Lazy<ExternalXrefSpec>)> externalXrefSpec, List<ExternalXref> externalXref) LoadJsonFile(string filePath)
```
live1206, line 68: nit: the class is `internal` already

live1206 (approved):

src/docfx/build/xref/ExternalXrefMapLoader.cs
yufeih: Yes, all of our class are `internal` by default. I marked the method as `internal` to express that it is only consumed by unit tests and not supposed to be called by other classes.

PR #5762, opened 2020-04-13
Development experience improvement: Fallback to use opBuildUserToken in keyVault if no secret set

928PJY:
[AB#201253](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/201253)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/5762)

PR #6766, opened 2020-11-11
Fix docset-not-provisioned error on keyvault access failure

yufeih:
When there are problems retrieving key vault secret, the `catch` block allows code to continue which leads to a misleading `docset-not-provisioned` error. In general, don't cache exceptions let the them populate to the upper layer, in this case, removing the `catch` block eventually caused a `download-failed` error with the detailed exception message.

https://github.com/dotnet/docfx/pull/5762

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6766)

### DIFF
--- a/src/docfx/config/ops/OpsAccessor.cs
+++ b/src/docfx/config/ops/OpsAccessor.cs
@@ -319,27 +319,19 @@ namespace Microsoft.Docs.Build
         {
             // don't access key vault for osx since azure-cli will crash in osx
             // https://github.com/Azure/azure-cli/issues/7519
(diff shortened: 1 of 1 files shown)


928PJY:
The reason why we catch the exception and let the code continue to run if for the public contributor, they cannot retrieve the secret but they can still call the API to retrieve the public data(Repo has been go-lived).

yufeih:
make sense

PR #6773, opened 2020-11-13
Turn static Build method to a class

yufeih:
[AB#335588](https://dev.azure.com/ceapex/Engineering/_workitems/edit/335588/)

We badly need [positional record](https://devblogs.microsoft.com/dotnet/c-9-0-on-the-record/) in C# 9 to reduce the constructor assignment code;

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6773)


src/docfx/build/DocsetBuilder.cs
```diff
@@ -0,0 +1,186 @@
...
+    internal class DocsetBuilder
+    {
+        private readonly string _workingDirectory;
+        private readonly string _docsetPath;
+        private readonly string? _outputPath;
+        private readonly CommandLineOptions _options;
+
+        public DocsetBuilder(string workingDirectory, string docsetPath, string? outputPath, CommandLineOptions options)
+        {
+            _workingDirectory = workingDirectory;
```
live1206, line 20: As from the work item description
- the member fields can be shared for different builds, which are readonly, **but these are just docset input?**
- dependencies created on-the-fly are not readonly, **but are they still be shared between file build?**

live1206 (approved):

928PJY (approved):
:shipit:

PR #6778, opened 2020-11-17
Should not add locale for api links

live1206:
[AB#270389](https://ceapex.visualstudio.com/Engineering/_workitems/edit/270389)
[AB#286381](https://ceapex.visualstudio.com/Engineering/_workitems/edit/286381)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6778)

### DIFF
--- a/src/docfx/lib/HtmlUtility.cs
+++ b/src/docfx/lib/HtmlUtility.cs
@@ -475,7 +475,7 @@ namespace Microsoft.Docs.Build
                         break;
                     case LinkType.AbsolutePath:
                         token.SetAttributeValue("data-linktype", "absolute-path");
-                        token.SetAttributeValue(attribute.Name.ToString(), AddLocaleIfMissing(href, locale));
+                        token.SetAttributeValue(attribute.Name.ToString(), AddLocaleIfMissingForAbsolutePath(href, locale));
                         break;
                     case LinkType.RelativePath:
                         token.SetAttributeValue("data-linktype", "relative-path");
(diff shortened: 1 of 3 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -487,8 +487,14 @@ internal static void AddLinkType(ref HtmlToken token, string locale)
             }
         }
 
-        private static string AddLocaleIfMissing(string href, string locale)
+        private static string AddLocaleIfMissingForAbsolutePath(string href, string locale)
         {
+            // should not add locale for api links
+            if (href.StartsWith("/api"))
```
mingwli, line 493: nit: could be case-insensitive?

mingwli (approved):

src/docfx/lib/HtmlUtility.cs
```diff
@@ -487,8 +487,14 @@ internal static void AddLinkType(ref HtmlToken token, string locale)
             }
         }
 
-        private static string AddLocaleIfMissing(string href, string locale)
+        private static string AddLocaleIfMissingForAbsolutePath(string href, string locale)
```
mingwli, line 490:
```suggestion
        private static string PrependLocaleIfMissing(string href, string locale)
```

live1206: Actually, this is not `prepend`.
And also, would like to specify this is for `AbosulutePath`.

PR #6784, opened 2020-11-19
Add Watch<T>

yufeih:
[AB#335590](https://dev.azure.com/ceapex/Engineering/_workitems/edit/335590)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6784)

### DIFF
--- a/src/docfx/lib/watch/LeafFunction{T}.cs
+++ b/src/docfx/lib/watch/LeafFunction{T}.cs
@@ -8,14 +8,14 @@ namespace Microsoft.Docs.Build
 {
     internal class LeafFunction<T> : IFunction
     {
-        private readonly Func<T> _valueFactory;
+        private readonly Func<T> _changeTokenFactory;
 
         [MaybeNull]
-        internal T Result { get; set; }
+        internal T ChangeToken { get; set; }
 
-        public LeafFunction(Func<T> valueFactory) => _valueFactory = valueFactory;
+        public LeafFunction(Func<T> changeTokenFactory) => _changeTokenFactory = changeTokenFactory;
 
-        public bool HasChanged() => !Equals(Result, _valueFactory());
+        public bool HasChanged() => !Equals(ChangeToken, _changeTokenFactory());
 
         public void AddChild(IFunction childFunction) { }
     }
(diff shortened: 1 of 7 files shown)


src/docfx/lib/watch/Watcher.cs
```diff
@@ -0,0 +1,76 @@
...
+
+using System;
+using System.Collections.Immutable;
+using System.Threading;
+
+namespace Microsoft.Docs.Build
+{
+    public static class Watcher
+    {
+        private static readonly AsyncLocal<ImmutableStack<IFunction>> t_callstack = new AsyncLocal<ImmutableStack<IFunction>>();
```
928PJY, line 12: >ImmutableStack [](start = 43, length = 14)

question: can we replace the `ImmutableStack` with a `Stack`?

yufeih: No, in case of a parallel call, each thread get the same `Stack` and causes thread safety problem. `ImmutableStack` forks a local copy of the stack on each thread.

src/docfx/lib/watch/Watch{T}.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        [MaybeNull]
+        private T _value;
+
+        private volatile ContainerFunction? _function;
+        private object? _syncLock;
+
+        public Watch(Func<T> valueFactory) => _valueFactory = valueFactory;
+
+        public bool IsValueCreated => _function != null;
```
928PJY, line 22: >IsValueCreated [](start = 20, length = 14)

will this just be used for testing?

yufeih: Yes for now. It exist because `Lazy<T>` also has this public property. I want `Watch<T>` to be a replacement of `Lazy<T>`.

src/docfx/lib/watch/LeafFunction{T}.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Microsoft.Docs.Build
+{
+    internal class LeafFunction<T> : IFunction
+    {
+        private readonly Func<T> _valueFactory;
+
+        [MaybeNull]
+        internal T Result { get; set; }
+
+        public LeafFunction(Func<T> valueFactory) => _valueFactory = valueFactory;
```
928PJY, line 16: >valueFactory [](start = 36, length = 12)

consider renaming this since it already consume the `changeTokenFactory`.

928PJY (approved):
:shipit:

PR #6798, opened 2020-11-23
Update error for redirected-file-not-removed

live1206:
[AB#320574](https://dev.azure.com/ceapex/Engineering/_workitems/edit/320574)
Refine error code & message for redirection file not deleted case.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6798)

### DIFF
--- a/src/docfx/publish/PublishUrlMap.cs
+++ b/src/docfx/publish/PublishUrlMap.cs
@@ -118,7 +118,7 @@ namespace Microsoft.Docs.Build
             }
 
             // redirection file is preferred than source file
-            var redirection = conflicts.SingleOrDefault(x => x.SourcePath.Origin == FileOrigin.Redirection);
+            var redirection = conflicts.FirstOrDefault(x => x.SourcePath.Origin == FileOrigin.Redirection);
             if (redirection != null)
             {
                 _errors.Add(Errors.Redirection.RedirectedFileNotRemoved(redirection.SourcePath));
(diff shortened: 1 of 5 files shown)


src/docfx/publish/PublishUrlMap.cs
```diff
@@ -117,12 +117,17 @@ private PublishUrlMapItem ResolveOutputPathConflicts(IGrouping<string, PublishUr
                 return conflicts.First();
             }
 
-            _errors.Add(Errors.UrlPath.OutputPathConflict(conflicts.First().OutputPath, conflicts.Select(x => x.SourcePath)));
-
             // redirection file is preferred than source file
+            var redirection = conflicts.SingleOrDefault(x => x.SourcePath.Origin == FileOrigin.Redirection);
```
yufeih, line 121: `SingleOrDefault` throws when there are more then 1 items match the condition (which can crash the process).
```suggestion
            var redirection = conflicts.FirstOrDefault(x => x.SourcePath.Origin == FileOrigin.Redirection);
```

live1206: We should not have multiple redirection with same file, it will end with `redirection-conflict`?

yufeih (approved):

PR #6805, opened 2020-11-24
This updates docfx to have basic support C# 9 support.

tannergooding:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6805)


src/Microsoft.DocAsCode.MarkdigEngine.Validators/Microsoft.DocAsCode.MarkdigEngine.Validators.csproj
```diff
@@ -3,7 +3,7 @@
 
   <ItemGroup>
     <PackageReference Include="Markdig" Version="0.20.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="1.5.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="5.0.0" />
```
tannergooding, line 6: `Microsoft.CodeAnalysis.Common` migrated forward to `System.Collections.Immutable >= 5.0.0`: https://www.nuget.org/packages/Microsoft.CodeAnalysis.Common/

src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/CSYamlModelGenerator.cs
```diff
@@ -1077,6 +1077,14 @@ public static ExpressionSyntax GetLiteralExpressionCore(object value, ITypeSymbo
                     return SyntaxFactory.LiteralExpression(
                         SyntaxKind.NumericLiteralExpression,
                         SyntaxFactory.Literal((ulong)value));
+                case SpecialType.System_IntPtr:
```
tannergooding, line 1080: `nint x = 5;` is valid in C# 9. `IntPtr x = 5` is not so there may be some other tweak needed to ensure it is rendered correctly, but I wasn't able to determine where.

src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/ReferenceItemVisitor.cs
```diff
@@ -136,6 +137,99 @@ public override void VisitPointerType(IPointerTypeSymbol symbol)
             });
         }
 
+        public override void VisitFunctionPointerType(IFunctionPointerTypeSymbol symbol)
```
tannergooding, line 140: Function pointer types are also new and look, most often like
`delegate*<void>`, `delegate* unmanaged<void>`, or `delegate* unmanaged[Stdcall]<void>`

The contents of `<>` are `<parameter, parameter, parameter, returnType>` where `returnType` must always be specified.

tannergooding: Notably, it looks like this is typed as `delegate* unmanaged<System.Void>`, which I'm not sure is "desirable" since it isn't "legal" syntax, even though `System.Void` is the qualified name.

src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/ReferenceItemVisitor.cs
```diff
@@ -383,6 +477,99 @@ public override void VisitPointerType(IPointerTypeSymbol symbol)
...
```
tannergooding, line 480: Function pointers, much like regular pointers, are illegal in VB. But it seems there is existing logic to generate "parts" for pointers in VB: <https://github.com/dotnet/docfx/pull/6805/files#diff-f05a1f9271b4d7874485ce55d806b5ef784db97e7544cd7f4353ad7c263df0f0L375>

Is this by design?

src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/VisitorHelper.cs
```diff
@@ -60,7 +60,18 @@ public static string GetId(ISymbol symbol)
...
-            return GetDocumentationCommentId(symbol)?.Substring(2);
+            var id = GetDocumentationCommentId(symbol)?.Substring(2);
+
+            if ((id is null) && (symbol is IFunctionPointerTypeSymbol functionPointerTypeSymbol))
+            {
+                // Roslyn doesn't currently support doc comments for function pointer type symbols
+                // This returns just the stringified symbol to ensure the source and target parts
+                // match for reference item merging.
+
+                return functionPointerTypeSymbol.ToString();
```
tannergooding, line 71: As per the comment, Roslyn doesn't support doc comments for function pointers today. This looks to be more generally tracked by https://github.com/dotnet/roslyn/issues/46674.

This at least gives a type name that should be mostly unique and include all the relevant portions.

tannergooding:
There are also other features of C# 9, like `records`, that weren't validated.

tannergooding:
Will look into the failure. Seems at least one of the tests is failing with:
```
Failed Microsoft.DocAsCode.Metadata.ManagedReference.Tests.GenerateMetadataFromAssemblyTest.TestGenerateMetadataFromAssemblyWithReferences [136 ms]
  Error Message:
   Microsoft.DocAsCode.Exceptions.DocfxException : Unable to generate spec reference for T:System.ValueTuple{System.String,System.String}
---- System.NullReferenceException : Object reference not set to an instance of an object.
```

superyyrrzz:
@tannergooding Thank you for adding the new features!

Could you help add some unit tests in this PR for the new feature supported? I think you are supporting native sized integers and function pointers here, but some tests will help me understand it better.

tannergooding:
> Could you help add some unit tests in this PR for the new feature supported?

Can do. I should be able to get to this in the next few days now that I'm back off vacation.

tannergooding:
Rebased and rerunning tests before adding the new tests.

buehler:
This would be very nice! I really like to document stuff with docfx, but newer code uses `record`s which are not supported :-(

roji:
Any news on this PR? Lack of support for C# 8/9 features is really a blocker for using docfx with modern projects...

tannergooding:
I've updated this to include tests and to cover:
* function pointers
* nint/nuint
* readonly members
* ref and ref readonly returns

This does not cover:
* records
* init only setters

There may be other syntax or languages features that also need improvements.

tannergooding:
@superyyrrzz, sorry for the delay in getting this up, but I believe there is now sufficient tests covering the added functionality. Please let me know if anything further is needed.

tannergooding:
Also updated to cover `ref struct` and `readonly struct` which are somewhat common in the BCL and were missing and an explicit test covering value tuple names

tannergooding:
And for modifiers to match the "preferred" order: https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0036#csharp_preferred_modifier_order

src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/ReferenceItemVisitor.cs
yufeih: I'm not aware of any specific reason to _intentionally support pointers for VB_. Might be the code for pointer is small and is accidentally duplicated here, but this function pointer is a big chunk of code and if we believe it is illegal for VB, leaving it as not implemented or marking it as not supported looks more clear to me.

yufeih (approved):
@tannergooding 
Thanks for you contribution! This is HOT 🔥 🔥 🔥 .
The overall logic LGTM and we'll try to merge and ship it soon.

src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/NameVisitor.cs
```diff
@@ -7,9 +7,10 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Linq;
+    using System.Reflection.Metadata;
     using System.Text;
-
-    using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis;
+using NuGet.Protocol.Plugins;
```
tannergooding, line 13:
```suggestion
    using Microsoft.CodeAnalysis;
    using NuGet.Protocol.Plugins;
```

src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/ReferenceItemVisitor.cs
tannergooding: Thanks for the clarification, I think this should be "good" then with how I've set it up.

If there is something explicit to be done to mark it as "illegal" or to "gracefully fail" please let me know :)

tannergooding:
> The overall logic LGTM and we'll try to merge and ship it soon.

Just wondering, will this automatically flow to v3 and to whatever we power docs.microsoft.com with (if its not one of the mainline branches)?

yufeih:
> Just wondering, will this automatically flow to v3 and to whatever we power docs.microsoft.com with (if its not one of the mainline branches)?

Unfortunately no. docs.microsoft.com uses [monodoc](https://github.com/mono/api-doc-tools) for .NET API which v3 will likely also use to unify on a single pipeline. `monodoc` internally uses `Mono.Cecil` to analyze dlls instead of roslyn.

tannergooding:
> monodoc internally uses Mono.Cecil to analyze dlls instead of roslyn.

That seems unfortunate, especially given how many language features can involve special attributes and other annotations that I don't believe Cecil directly supports (they'll be exposed, but the "understanding" of whether `IntPtr` is an `IntPtr` or a `nint` is no longer as simple as checking `symbol.IsNativeIntegerType`).

In either case, it would be good to get most of these features working there as well, to ensure that things like [System.Span<T>](https://docs.microsoft.com/en-us/dotnet/api/system.span-1?view=net-5.0) or [System.Numerics.Vector4](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.vector4?view=net-5.0) correctly display the relevant semantic information in the docs.

yufeih:
The input to generate .NET APIs on docs is a bunch of DLLs rather than the source code. We've tried to stay away from the source code to decouple from the complex build system of each component, with the downside you mentioned. 

@joelmartinez What's our status on supporting these latest C# language features on docs? @tannergooding might be able to help with his expertise in this area.

joelmartinez:
@yufeih we've recently been working to improve mdoc's support for nullable. I would recommend we work with our PM, @mimisasouvanh, on any feature that needs improvement in mdoc's language formatters.

@tannergooding you're right, cecil only takes us so far, and there's additional logic we often need to implement in mdoc as new language features come out, especially the ones that depend on attributions. Please feel free to ping myself or @TianqiZhang if you have any questions or thoughts about contributing to mdoc :)

PR #6829, opened 2020-12-02
triger diffs after update regression all ruleset

xichan-ms:
[AB#338038](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/338038)

PR #6837, opened 2020-12-03
Add warnings and errors for multuple redirection files

AngryBerryMS:
[AB#341331](https://dev.azure.com/ceapex/Engineering/_workitems/edit/341331)

|Error code|Level|Description|
|---|---|---|
|redirection-file-not-found|Error|A registered file is not found in the repo.|
|source-path-conflict|Error| source_path and source_path_from_root exist at the same time.|
|redirection-path-syntax-error|Warning|Check whether starting with "/" or not.|
|redirection-conflict|Error|Same as before except including sub-redirection files check.|

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6837)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -343,8 +343,8 @@ namespace Microsoft.Docs.Build
             /// A redirection file registered in .openpublishing.publish.json is not found in the repo.
             /// </summary>
             /// Behavior: ✔️ Message: ✔️
-            public static Error FileNotFound(string path)
-                => new Error(ErrorLevel.Error, "file-not-found", $"Redirection file '{path}' registered in .openpublishing.publish.json is not found in the repo.");
+            public static Error RedirectionFileNotFound(string path)
+                => new Error(ErrorLevel.Error, "redirection-file-not-found", $"Redirection file '{path}' registered in .openpublishing.publish.json is not found in the repo.");
 
             /// <summary>
             /// A redirection item cannot contain ‘source_path’ and ‘source_path_from_root’ at the same time.
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -354,10 +354,11 @@ namespace Microsoft.Docs.Build
                 => new Error(ErrorLevel.Error, "source-path-conflict", $"A redirection item cannot contain 'source_path' and 'source_path_from_root' at the same time.", source);
 
             /// <summary>
+            /// Check redirection source path syntax.
             /// </summary>
             /// Behavior: ✔️ Message: ✔️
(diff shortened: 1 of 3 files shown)


src/docfx/Errors.cs
```diff
@@ -338,6 +338,27 @@ public static Error CircularRedirection(SourceInfo? source, IEnumerable<FilePath
             /// Behavior: ✔️ Message: ✔️
             public static Error RedirectedFileNotRemoved(FilePath path)
                 => new Error(ErrorLevel.Warning, "redirected-file-not-removed", $"Redirected file '{path}' is still in the repo. After adding a file to the redirection JSON file, you must delete the original file from the repo.");
+
+            /// <summary>
+            /// A redirection file registered in .openpublishing.publish.json is not found in the repo.
+            /// </summary>
+            /// Behavior: ✔️ Message: ✔️
+            public static Error FileNotFound(string path)
+                => new Error(ErrorLevel.Error, "file-not-found", $"Redirection file '{path}' registered in .openpublishing.publish.json is not found in the repo.");
```
live1206, line 347: we already have `file- not -found`, can we have a different code for this one? Like `redirection-file-not-found`?

src/docfx/Errors.cs
```diff
@@ -338,6 +338,27 @@ public static Error CircularRedirection(SourceInfo? source, IEnumerable<FilePath
...
+
+            /// <summary>
+            /// A redirection item cannot contain ‘source_path’ and ‘source_path_from_root’ at the same time.
+            /// </summary>
+            /// /// Behavior: ✔️ Message: ✔️
+            public static Error SourcePathConflict(SourceInfo<string> source)
+                => new Error(ErrorLevel.Error, "source-path-conflict", $"A redirection item cannot contain 'source_path' and 'source_path_from_root' at the same time.", source);
```
live1206, line 354: It would be better to specify which one is in use in the message.

AngryBerryMS: No one will be used when they exist at the same time.

All error codes and messages are confirmed with PM. I think we shouldn't change them.

src/docfx/Errors.cs
AngryBerryMS: All error codes and messages are confirmed with PM. I think we shouldn't change them.

xichan-ms:
Some diffs in repos **DevSandbox, learn-pr, test** are expected, for more information, please refer to [PR](https://github.com/dotnet/docfx/pull/6829).

live1206 (approved):

yufeih (approved):

Issue #6835, opened 2020-12-03
Build a virtual file system


928PJY:
[AB#342653](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/342653)

- [x] Create the docset package during initialization which provides the interface for accessing the files in the current docset. (#6836, #6853)
- [x] Apply package to config loader (#6836, #6840 )
- [x] Apply package to ContributionProvider (#6841)
- [x] Apply package to RedirectionProvider (#6842)
- [x] Apply package to ExternalXrefMapLoader (#6842)
- [x] Apply package to FileResolver (#6841)
- [x] Apply package to TemplateEngine (#6841)
- [x] ~~Apply package to Validation library (For now, only the ruleset need to access the file system, which is not in the scope of real-time validation)~~
- [x] Support MemoryPackage which hosts all the files in memory without touching file system. (#6857)
- [x] Support CompositePackage which maintaining multiple packages and handle the fallback. (#6857)
- [x] Apply MemoryPackage to our spec test. (#6857)

PR #6840, opened 2020-12-04
Apply package to config loader

928PJY:
#6835 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6840)

### DIFF
--- a/src/docfx/build/Builder.cs
+++ b/src/docfx/build/Builder.cs
@@ -15,14 +15,14 @@ namespace Microsoft.Docs.Build
         private readonly string _workingDirectory;
         private readonly CommandLineOptions _options;
         private readonly Watch<DocsetBuilder[]> _docsets;
-        private readonly Package _docsetPackage;
+        private readonly Package _package;
 
-        public Builder(ErrorBuilder errors, string workingDirectory, CommandLineOptions options, Package docsetPackage)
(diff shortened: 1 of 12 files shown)


src/docfx/build/Builder.cs
```diff
@@ -59,7 +59,7 @@ public void Build(params string[] files)
 
         private DocsetBuilder[] LoadDocsets()
         {
-            var docsets = ConfigLoader.FindDocsets(_errors, _workingDirectory, _docsetPackage, _options);
+            var docsets = ConfigLoader.FindDocsets(_errors, _docsetPackage, _options);
```
yufeih, line 62: The package in `Builder` is not specific to a docset, consider rename it to `_package`.

```suggestion
            var docsets = ConfigLoader.FindDocsets(_errors, _package, _options);
```

yufeih (approved):

src/docfx/build/Builder.cs
928PJY: would `package` be too general for in this context?

---
In reply to: [535802282](https://github.com/dotnet/docfx/pull/6840#discussion_r535802282) [](ancestors = 535802282)

PR #6841, opened 2020-12-04
Apply package to ContributionProvider, FileResolver, PackageResolver, TemplateEngine

928PJY:
#6835 

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6841)

### DIFF
--- a/src/docfx/build/context/Input.cs
+++ b/src/docfx/build/context/Input.cs
@@ -116,7 +116,7 @@ namespace Microsoft.Docs.Build
 
             var (package, path) = ResolveFilePath(_sourceMap.GetOriginalFilePath(file) ?? file);
 
-            return package.GetLastWriteTimeUtc(path);
+            return package.TryGetLastWriteTimeUtc(path) ?? default;
         }
 
         /// <summary>
--- a/src/docfx/build/context/Input.cs
+++ b/src/docfx/build/context/Input.cs
@@ -107,6 +107,18 @@ namespace Microsoft.Docs.Build
             return package.TryGetPhysicalPath(path);
         }
 
+        public DateTime GetLastWriteTimeUtc(FilePath file)
+        {
+            if (file.IsGitCommit || file.Origin == FileOrigin.Generated)
+            {
+                return default;
+            }
+
+            var (package, path) = ResolveFilePath(_sourceMap.GetOriginalFilePath(file) ?? file);
+
+            return package.GetLastWriteTimeUtc(path);
+        }
+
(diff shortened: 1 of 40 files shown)


src/docfx/restore/FileResolver.cs
```diff
@@ -178,14 +182,14 @@ private string DownloadFromUrlCore(string url)
             }
         }
 
-        private static string GetRestorePathFromUrl(string url)
+        private static PathString GetRestorePathFromUrl(string url)
         {
-            return PathUtility.NormalizeFile(Path.Combine(AppData.GetFileDownloadPath(url)));
+            return new PathString(PathUtility.NormalizeFile(Path.Combine(AppData.GetFileDownloadPath(url))));
```
yufeih, line 187: `new PathString` does the path normalization.
```suggestion
            return new PathString(Path.Combine(AppData.GetFileDownloadPath(url))));
```

yufeih (changes requested):
Consider leave FileResolver and PackageResolver as is for now, they are typically external resources and so let's focus on content first.

928PJY:
Resouce used in FileResolver can be content as well.
As for packageResolver, template can also be a part of the repository, we should also handle it for now.

---
In reply to: [544666888](https://github.com/dotnet/docfx/pull/6841#issuecomment-544666888) [](ancestors = 544666888)

src/docfx/build/context/Input.cs
```diff
@@ -107,6 +107,18 @@ public bool Exists(FilePath file)
...
+        public DateTime GetLastWriteTimeUtc(FilePath file)
+        {
+            if (file.IsGitCommit || file.Origin == FileOrigin.Generated)
+            {
+                return default;
+            }
+
+            var (package, path) = ResolveFilePath(_sourceMap.GetOriginalFilePath(file) ?? file);
+
+            return package.GetLastWriteTimeUtc(path);
```
yufeih, line 119:
```suggestion
            return package.TryGetLastWriteTimeUtc(path);
```

src/docfx/template/TemplateEngine.cs
```diff
@@ -41,7 +41,11 @@ internal class TemplateEngine
             };
 
         public TemplateEngine(
```
yufeih, line 43: no change?

yufeih (approved):

PR #6848, opened 2020-12-04
Specify scope while update progress

live1206:
[AB#344730](https://ceapex.visualstudio.com/Engineering/_workitems/edit/344730)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6848)

### DIFF
--- a/src/docfx/build/toc/TocMap.cs
+++ b/src/docfx/build/toc/TocMap.cs
@@ -159,13 +159,13 @@ namespace Microsoft.Docs.Build
         private (FilePath[] tocs, Dictionary<FilePath, FilePath[]> docToTocs, List<FilePath> servicePages)
             BuildTocMap()
         {
-            using (Progress.Start("Loading TOC"))
+            using (var scope = Progress.Start("Loading TOC"))
             {
                 var tocs = new ConcurrentBag<FilePath>();
                 var allServicePages = new ConcurrentBag<FilePath>();
 
                 // Parse and split TOC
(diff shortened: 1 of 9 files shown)


src/docfx/build/toc/TocMap.cs
```diff
@@ -159,21 +159,21 @@ private static (int subDirectoryCount, int parentDirectoryCount) GetRelativeDire
         private (FilePath[] tocs, Dictionary<FilePath, FilePath[]> docToTocs, List<FilePath> servicePages)
             BuildTocMap()
         {
-            using (Progress.Start("Loading TOC"))
+            using (var scope = Progress.Start("Loading TOC"))
```
yufeih, line 162: remove `{` and `}` if we are measuring the function?
```suggestion
            using var scope = Progress.Start("Loading TOC");
```

src/docfx/lib/log/Progress.cs
```diff
@@ -8,17 +8,13 @@
 
 namespace Microsoft.Docs.Build
 {
-    internal static class Progress
+    internal static partial class Progress
```
yufeih, line 11: nit: why partial
```suggestion
    internal static class Progress
```

yufeih (approved):
👍

src/docfx/lib/log/Progress.cs
live1206: Mistakenly added by VS while moving `LogScope` out...

PR #6877, opened 2020-12-10
Don't get full path when --template is an URL

yufeih:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6877)


928PJY (approved):

yufeih:
@928PJY The public template seems to be broken is a different way:

`"Cannot find dependency 'https://static.docs.com/ui/latest/token.yml', did you forget to run `docfx restore`?"`

PR #6880, opened 2020-12-10
added 4 more inheritance levels

georgius:
**Operating System**: Windows

**DocFX Version Used**: 2.56.6

**Template used**: default

**Steps to Reproduce**:

1. inherit classes more than 5 levels from System.Object

**Expected Behavior**:
The expected behavior is that result HTML code will be rendered correctly as on attached picture
inheritance_more_levels

**Actual Behavior**:
The actual behavior is that up to 5th inheritance level from System.Object is inheritance rendered correctly. After 5th level the inheritance is not rendered correctly.
inheritance_2_56_6

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6880)

### DIFF
--- a/src/docfx.website.themes/default/styles/docfx.css
+++ b/src/docfx.website.themes/default/styles/docfx.css
@@ -42,7 +42,11 @@ h6 mark {
 .inheritance .level2:before,
 .inheritance .level3:before,
 .inheritance .level4:before,
-.inheritance .level5:before {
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Thank you for the fix🚀

PR #6884, opened 2020-12-11
Support using OPS token from environment variable

928PJY:
[AB#347501](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/347501)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6884)

### DIFF
--- a/src/docfx/config/EnvironmentVariable.cs
+++ b/src/docfx/config/EnvironmentVariable.cs
@@ -19,6 +19,8 @@ namespace Microsoft.Docs.Build
 
         public static string? CorrelationId => GetValue("DOCFX_CORRELATION_ID");
 
+        public static string? OPSBuildUserToken => GetValue("DOCFX_OPS_BUILD_USER_TOKEN");
+
         private static string? GetValue(string name)
         {
             var value = Environment.GetEnvironmentVariable(name);
--- a/src/docfx/config/EnvironmentVariable.cs
+++ b/src/docfx/config/EnvironmentVariable.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Docs.Build
 
         public static string? CorrelationId => GetValue("DOCFX_CORRELATION_ID");
 
-        public static string? OPSBuildUserToken => GetValue("DOCFX_OPS_BUILD_USER_TOKEN");
+        public static string? DocsOpsToken => GetValue("DOCS_OPS_TOKEN");
 
         private static string? GetValue(string name)
         {
--- a/src/docfx/config/ops/OpsAccessor.cs
+++ b/src/docfx/config/ops/OpsAccessor.cs
@@ -334,17 +334,16 @@ namespace Microsoft.Docs.Build
                 // For development usage
                 try
                 {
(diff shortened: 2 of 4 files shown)


src/docfx/config/EnvironmentVariable.cs
```diff
@@ -19,6 +19,8 @@ public static class EnvironmentVariable
 
         public static string? CorrelationId => GetValue("DOCFX_CORRELATION_ID");
 
+        public static string? OPSBuildUserToken => GetValue("DOCFX_OPS_BUILD_USER_TOKEN");
```
yufeih, line 22:
```suggestion
        public static string? OPSBuildUserToken => GetValue("DOCFX_OPS_TOKEN");
```

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -334,14 +334,20 @@ private static async Task FillOpsToken(HttpRequestMessage request, DocsEnvironme
...
+                    var token = EnvironmentVariable.OPSBuildUserToken;
+                    if (string.IsNullOrEmpty(token))
                     {
-                        DocsEnvironment.Prod => s_opsTokenProd,
-                        DocsEnvironment.PPE => s_opsTokenSandbox,
-                        _ => throw new InvalidOperationException(),
-                    };
-
-                    request.Headers.Add("X-OP-BuildUserToken", await token.Value);
+#pragma warning disable SA1119 // Statement should not use unnecessary parenthesis
```
yufeih, line 340: why disable SA1119 here?

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -334,14 +334,20 @@ private static async Task FillOpsToken(HttpRequestMessage request, DocsEnvironme
                 // For development usage
                 try
                 {
-                    var token = (environment ?? DocsEnvironment) switch
+                    var token = EnvironmentVariable.OPSBuildUserToken;
+                    if (string.IsNullOrEmpty(token))
```
yufeih, line 338: use `??`

src/docfx/config/ops/OpsAccessor.cs
928PJY: seems it doesn't report rightly

src/docfx/config/ops/OpsAccessor.cs
928PJY: seems ?? don't handle the empty string?

yufeih: `EnvironmentVariable.GetValue` internally handles empty string, to make it work with `??`

yufeih (approved):

src/docfx/config/ops/OpsAccessor.cs
928PJY: Hah, great, but seem it cannot be used in this case since we need another variable `tokenTask`

PR #6885, opened 2020-12-11
upgrade docs.validation to support RedirectedLink rule

fkpwolf:
[AB#347121](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/347121)


docs/specs/validation/content-validation.yml
```diff
@@ -136,3 +136,36 @@ outputs:
   file1.json:
   a/b/c/file.json:
 ---
+# Validate redirected link
```
xichan-ms, line 139: >Validate redirected link [](start = 2, length = 24)

how about moving it to file image-validtion.yml and renaming that file to link-validation.yml?
because they call the same validation interface.

fkpwolf:
Diff is excepted. And `azure-doc-pr` will have 3k more Suggestions than before.

docs/specs/validation/content-validation.yml
fkpwolf: good idea! since I don't want to create new link-validation.yml which has only one rule...

docs/specs/validation/link-validation.yml
```diff
@@ -417,3 +417,36 @@ outputs:
...
+            "redirectedLinkTypes": [
+               "aka.ms", 
+               "FWLinks"
+            ]
+          }
+        ]
+      }
+    }
+  file1.md: |
+    [external link](https://go.microsoft.com/fwlink/p/?LinkId=389595)
```
xichan-ms, line 448: do we need to validate http link (<a href=x)? 
if so, how about add a this kind of test case?

xichan-ms:
>         private string GetLink(SourceInfo<string> href)

if validate http link, this function will also call validation api.

---
Refers to: src/docfx/lib/markdown/MarkdownEngine.cs:350 in ef75f33. [](commit_id = ef75f3318fe0de118fe0d7e3fde5a96806d03416, deletion_comment = False)

src/docfx/lib/markdown/MarkdownEngine.cs
```diff
@@ -321,6 +321,7 @@ private static ErrorBuilder GetErrors()
 
         private string GetLink(string path, MarkdownObject origin)
         {
+            _contentValidator.ValidateLink(GetRootFilePath(), new SourceInfo<string>(path, origin.GetSourceInfo()), origin);
```
OsmondJiang, line 324: this GetLink is not used by all code path

docs/specs/validation/link-validation.yml
fkpwolf: added

src/docfx/lib/markdown/MarkdownEngine.cs
```diff
@@ -342,19 +332,22 @@ private string GetImageLink(string path, MarkdownObject origin, string? altText)
         private string GetImageLink(SourceInfo<string> href, MarkdownObject origin, string? altText, int imageIndex)
         {
             _contentValidator.ValidateImageLink(GetRootFilePath(), href, origin, altText, imageIndex);
```
OsmondJiang, line 334: Can we combine ValidateLink and ValidateImageLink together

fkpwolf: done

OsmondJiang (approved):
@fkpwolf we need tell Megan that this validation rule should turn on PR-Only flag...

src/docfx/lib/markdown/HtmlExtension.cs
```diff
@@ -54,7 +54,7 @@ string ProcessHtml(string html, MarkdownObject block, ErrorBuilder errors, bool
                 // The following code ensures we preserve the original html when changing links.
                 return HtmlUtility.TransformHtml(html, (ref HtmlReader reader, ref HtmlWriter writer, ref HtmlToken token) =>
                 {
-                    HtmlUtility.TransformLink(ref token, block, getLink, getImageLink!);
+                    HtmlUtility.TransformLink(ref token, block, getLink!, getImageLink!);
```
yufeih, line 57: remove `!` ?
```suggestion
                    HtmlUtility.TransformLink(ref token, block, getLink, getImageLink);
```

yufeih (approved):

src/docfx/lib/markdown/HtmlExtension.cs
fkpwolf: that will result in lots of code warning:

Severity	Code	Description	Project	File	Line	Suppression State
Warning	CS8620	Argument of type 'Func<SourceInfo<string>, MarkdownObject, string>' cannot be used for parameter 'transformLink' of type 'Func<SourceInfo<string>, MarkdownObject?, string>' in 'void HtmlUtility.TransformLink(ref HtmlToken token, MarkdownObject? block, Func<SourceInfo<string>, MarkdownObject?, string> transformLink, Func<SourceInfo<string>, MarkdownObject?, string?, int, string>? transformImageLink = null)' due to differences in the nullability of reference types.	docfx	C:\Users\fan\source\repos\docfx\src\docfx\lib\markdown\HtmlExtension.cs	57	Active

xichan-ms: Maybe we can not remove it now, and this exclamation mark is an exception according to our original talking. : ) @yufeih

PR #6886, opened 2020-12-11
no need to generate service when href defined

AngryBerryMS:
[AB#347152](https://dev.azure.com/ceapex/Engineering/_workitems/edit/347152?src=WorkItemMention&src-action=artifact_link)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6886)

### DIFF
--- a/src/docfx/build/toc/ServicePageGenerator.cs
+++ b/src/docfx/build/toc/ServicePageGenerator.cs
@@ -46,9 +46,8 @@ namespace Microsoft.Docs.Build
             var filename = name.Replace(" ", "");
             var referenceTOCRelativeDir = Path.GetDirectoryName(_joinTOCConfig.ReferenceToc) ?? ".";
             var referenceTOCFullPath = Path.GetFullPath(Path.Combine(_docsetPath, referenceTOCRelativeDir));
-            var nodeHrefFullPath = Path.GetFullPath(Path.Combine(referenceTOCFullPath, node.Href.Value ?? ""));
 
-            if (string.IsNullOrEmpty(node.Href.Value) || (!node.Href.Value.StartsWith("~") && !File.Exists(nodeHrefFullPath)))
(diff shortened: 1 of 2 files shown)


src/docfx/build/toc/ServicePageGenerator.cs
```diff
@@ -48,7 +48,7 @@ public void GenerateServicePageFromTopLevelTOC(TocNode node, List<FilePath> resu
             var referenceTOCFullPath = Path.GetFullPath(Path.Combine(_docsetPath, referenceTOCRelativeDir));
             var nodeHrefFullPath = Path.GetFullPath(Path.Combine(referenceTOCFullPath, node.Href.Value ?? ""));
 
-            if (string.IsNullOrEmpty(node.Href.Value) || !File.Exists(nodeHrefFullPath))
```
herohua, line 51: This looks hack to me. Is it possible to determine by checking whether there is a file of that name already existing in the repo?

herohua (changes requested):
prevent auto-merge

herohua:
Talked with @AngryBerryMS and we prefer to consolidate the behavior as `We don't generate service page as long as href is being specified in top-level-toc.yml regardless of whether the specified file exists, or the file path starts with ~ or not`.

+ @live1206 @mingwli to comment

AngryBerryMS:
Will fix after xiangwei and peican check the TOPLevelTOC of current repos to confirm whether relative path for item that contain landingpageType property

src/docfx/build/toc/ServicePageGenerator.cs
AngryBerryMS: Fount the way to get fullpath when start with ~

yufeih (approved):

Issue #6852, opened 2020-12-04
metadata filename clash between namespace and class name


farcasclaudiu:
**Operating System**: 
`Windows` or `MacOS`

**DocFX Version Used**: 
docfx 2.56.5.0

**Template used**: 
`default`

**Steps to Reproduce**:

1. checkout the sample project that repro the issue from here - https://github.com/farcasclaudiu/docfx_conflict
2. build project (dotnet build) - should succed
3. run `docfx ./docfx_project/docfx.json --force --debug` command
It fails.

**Expected Behavior**:
to generate the apidoc _site without any error.

**Actual Behavior**:
error in command line:
`[20-12-04 01:01:31.963]Error:[BuildCommand.BuildCore.Build Document.LinkPhaseHandlerWithIncremental.Apply Templates](api/docfx_conflict.yml)Error transforming model "/var/folders/87/y9_3c0qn76g9wlzzg5m8fsl40000gn/T/docfx/rawmodel/api/docfx_conflict.raw.json" generated from "api/docfx_conflict.yml" using "ManagedReference.html.primary.js". Error running Transform function inside template preprocessor: type is undefined`

**Cause of the error**
The error message is not very helpful unfortunately.
The error is caused because in the project coexist a namespace (`docfx_conflict.member`) and a class (`docfx_conflict.Member`) and docfx generated yaml metadata files for both of these, but the OS filesystem is case insensitive. One will be overwritten I guess.

Any idea?

georgius:
I just tested this problem on Windows, docfx 2.56.6.0 and no error occurred. Is there still error in your environment?

georgius:
> just tested on Mac with docfx version 2.56.6.0 and the error is the same.

Can you test it also on Windows (just for confirmation that on Windows it is working)?

farcasclaudiu:
I just tested on Windows and it works fine. But not no Mac.

georgius:
> I just tested on Windows and it works fine. But not no Mac.

In this case you should edit your first post and change Operating System. Unfortunately, I don't have MacOS machine to debug. I try to create virtual machine with MacOS, but this will take some time.

farcasclaudiu:
I updated the https://github.com/farcasclaudiu/docfx_conflict project with another similar case:
Docker containers with mounted volumes on Windows or macOS hosts.

Error:
```
[...]

PR #6888, opened 2020-12-12
fix MacOS case insensitive file system

georgius:
**Operating System**: MacOS

**DocFX Version Used**: 2.56.6

**Template used**: default

**Steps to Reproduce**:

1. use [docfx_conflict github project](https://github.com/farcasclaudiu/docfx_conflict) mentioned in #6852 

**Expected Behavior**:
The documentation is created.

**Actual Behavior**:
The actual behavior is that documentation is not created (the error is in #6852). The problem is caused by MacOS file system, which can or cannot be case sensitive. It depends on how is created partition and file system on MacOS. Also in [Microsoft documentation](https://docs.microsoft.com/is-is/dotnet/api/system.platformid?view=netcore-3.1) is stated that MacOS operating system is returned as PlatformID.Unix. In such case is it much more safe to assume that all file systems are case insensitive.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6888)

### DIFF
--- a/src/Microsoft.DocAsCode.Common/Path/PathUtility.cs
+++ b/src/Microsoft.DocAsCode.Common/Path/PathUtility.cs
@@ -26,7 +26,7 @@ namespace Microsoft.DocAsCode.Common
 
         public static bool IsPathCaseInsensitive()
         {
-            return Environment.OSVersion.Platform < PlatformID.Unix;
+            return true;
         }
 
         public static string ToValidFilePath(this string input, char replacement = '_')
--- a/src/Microsoft.DocAsCode.Common/Path/PathUtility.cs
+++ b/src/Microsoft.DocAsCode.Common/Path/PathUtility.cs
@@ -26,6 +26,11 @@ namespace Microsoft.DocAsCode.Common
 
         public static bool IsPathCaseInsensitive()
         {
+            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Linux))
+            {
+                return false;
+            }
+
             return true;
         }
 


yufeih:
@georgius , the fix may report wrong case sensitivity on linux. .NET core itself detects case sensitivity by creating a random file and see if that file exits using a different case, we could use the same trick here. FYI: https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/Common/src/System/IO/PathInternal.CaseSensitivity.cs#L37

bianyifan:
@yufeih Temporary directory may not be on the same file system with the docfx working directory, e.g. temp directory on a case insensitive file system but docfx works on a case sensitive one, the test result could not be correct.

georgius:
@bianyifan is correct, also output site can be hosted on several web servers on different operating systems and file systems. I think, it is safer to assume that file system is case insensitive.

farcasclaudiu:
I have another very similar case when running docfx into a docker container hosted on Windows or macOS.
https://github.com/dotnet/docfx/issues/6852#issuecomment-744308727

yufeih:
@superyyrrzz on his thoughts

src/Microsoft.DocAsCode.Common/Path/PathUtility.cs
```diff
@@ -26,7 +26,7 @@ public static class PathUtility
 
         public static bool IsPathCaseInsensitive()
         {
-            return Environment.OSVersion.Platform < PlatformID.Unix;
+            return true;
```
superyyrrzz, line 29: What about update to `Environment.OSVersion.Platform < PlatformID.Unix || Environment.OSVersion.Platform == PlatformID.MacOSX`?

Some considerations:
1. DocFX is still using .NET Framework instead of .NET Core, so I think the environment platform should return `MacOSX` instead of `Unix` here.
2. In Unix, I'd like to keep the behavior as it is used to be. If we return `true` here on Unix, users can write broken links with only casing difference in Markdown without getting any warning from DocFX build, but they are indeed broken links when served on Linux server.

georgius: Your solution will not work on MacOS (just right tested), because MacOS is reported as Unix (PlatformID.Unix), so docfx assumes that file system is case sensitive (running under mono with .NET Core 3.1.404).

georgius: > 2. In Unix, I'd like to keep the behavior as it is used to be. If we return `true` here on Unix, users can write broken links with only casing difference in Markdown without getting any warning from DocFX build, but they are indeed broken links when served on Linux server.

I tried to build documentation of [docfx_conflict github project](https://github.com/farcasclaudiu/docfx_conflict) on Linux (Debian 10, mono in docker, docfx 2.56.6). The documentation was built successfully (as expected, because Debian has case sensitive file system). But copying output website to Windows failed, due to "same" file names: docfx_conflict.Member.html and docfx_conflict.member.html.

bianyifan: > Your solution will not work on MacOS (just right tested), because MacOS is reported as Unix (PlatformID.Unix), so docfx assumes that file system is case sensitive (running under mono with .NET Core 3.1.404).

Renze means docfx v2 is targeting .NET Framework instead of .NET Core, and running docfx is not officially supported.

> I tried to build documentation of docfx_conflict github project on Linux (Debian 10, mono in docker, docfx 2.56.6). The documentation was built successfully (as expected, because Debian has case sensitive file system).

To be more accurate, ext4, xfs and some other file systems generally used by Linux distributions are case sensitive, NTFS, FAT32, APFS (by default) and ZFS/JFS (with specific option) are case insensitive.

> But copying output website to Windows failed, due to "same" file names: docfx_conflict.Member.html and docfx_conflict.member.html.

I don't know what file system you are using, but I am assuming NTFS since you said Windows. NTFS is a case-insensitive filesystem, you can't have `docfx_conflict.Member.html` and `docfx_conflict.member.html` in same directory and Explorer or other utility will report an error. It's completely unrelated to docfx.

I think there's two things in question here:

- Should docfx generate different results across platforms? If not, we should just treat all platform as case insensitive, and report error on this.
- How does docfx tell a file system is case-sensitive or not? We know a system can run on different file systems, such as one partition is FAT32 and another one is ext4, simply using operating system to determine this could result in runtime error.

superyyrrzz: > Your solution will not work on MacOS (just right tested), because MacOS is reported as Unix (PlatformID.Unix), so docfx assumes that file system is case sensitive (running under mono with .NET Core 3.1.404).

@georgius Got your ideas. So here we cannot rely on `Environment.OSVersion.Platform` to determine if it is Linux or MacOS. I think it is not fair for Linux users if we always return `true` here though.

What we need here is to determine case sensitivity reliably, so [what @yufeih mentioned](https://github.com/dotnet/corefx/blob/bffef76f6af208e2042a2f27bc081ee908bb390b/src/Common/src/System/IO/PathInternal.CaseSensitivity.cs#L37) seems to be the only valid option here...

> Renze means docfx v2 is targeting .NET Framework instead of .NET Core, and running docfx on Mono is not officially supported.

@bianyifan To clarify, Mono is the cross-platform solution recommended by DocFX currently, as DocFX does not take the effort to migrate to .NET Core.

bianyifan: @superyyrrzz Should that be interpreted as "You can give Mono a try but we don't guarantee everything works fine"?

superyyrrzz: It should work fine unless DocFX stops supporting cross-platform.

georgius: > To be more accurate, ext4, xfs and some other file systems generally used by Linux distributions are case sensitive, NTFS, FAT32, APFS (by default) and ZFS/JFS (with specific option) are case insensitive.

Correct, Debian is on EXT4, Windows on NTFS, MacOS on APFS (case insesitive).

> I don't know what file system you are using, but I am assuming NTFS since you said Windows. NTFS is a case-insensitive filesystem, you can't have `docfx_conflict.Member.html` and `docfx_conflict.member.html` in same directory and Explorer or other utility will report an error. It's completely unrelated to docfx.

Correct, Windows is on NTFS, I expected the conflict, I just want to confirm that it happens.

> * Should docfx generate different results across platforms? If not, we should just treat all platform as case insensitive, and report error on this.

By my opinion, on all platforms should be output the same.

On [docfx webpage](https://dotnet.github.io/docfx/) is stated that: 

> DocFX can run on Linux, macOS, and Windows. The generated static website can be deployed to any host such as GitHub Pages or Azure Websites with no additional configuration.

In current state, it is not true, it depends on circumstances.

In my case, I'm using docfx only on Windows and output websites are hosted on IIS (technically, can be hosted also on other web servers). But Linux or MacOS users output websites cannot be in all cases hosted on IIS.

bianyifan: Okay, sorry about my misunderstanding before (not familiar with v2...) But Mono implementation have some small differences with .NET Framework. As per [`PlatformID` documentation](https://docs.microsoft.com/en-us/dotnet/api/system.platformid?view=netframework-4.8), macOS should be correctly reported as `MacOSX` instead of `Unix` on .NET Core.

georgius: In mono is PlatformID overwritten, see https://github.com/mono/mono/blob/9829195f68d0ff07304c30edef7874f4dd6ba6d6/mcs/class/corlib/System/Environment.cs#L228

I think, that critical question is: **Should docfx generate different results across platforms?**

farcasclaudiu: > In mono is PlatformID overwritten, see https://github.com/mono/mono/blob/9829195f68d0ff07304c30edef7874f4dd6ba6d6/mcs/class/corlib/System/Environment.cs#L228
> 
> I think, that critical question is: **Should docfx generate different results across platforms?**

Hi, I think the docfx results should be consistent across all platforms. I believe it should behave to the lowest denominator (it should generate files for case INSENSITIVE file systems).

Best regards.

yufeih: > I think the docfx results should be consistent across all platforms.

This is something we would hope to achieve, but given [our current v2 strategy](https://github.com/dotnet/docfx/blob/dev/Roadmap.md#v2-strategy) and the complexity of file systems, we are unlikely to make that happen in v2 at least for linux.

Lets instead focus on _"how to get the correct case sensitivity on Mac"_. Obviously `Environment.OSVersion.Platform` does not work on mono, but `RuntimeInformation.IsOSPlatform(OSPlatform.OSX)` should work.

farcasclaudiu: I understand that docfx is already working just fine on Windows (which already has a case insensitive file system).
Is it difficult to have the same behavior as well on other platforms (linux, macOS)?
That would make it consistent across all platforms as results + results files (html) will be portable between various servers (linux, windows) where we wish to host the documentation.

georgius:
```suggestion
            if (System.Runtime.InteropServices.RuntimeInformation.IsOSPlatform(System.Runtime.InteropServices.OSPlatform.Linux))
            {
                return false;
            }

            return true;
```

Is it acceptable?

farcasclaudiu: docker is still linux.
Will return true for all platforms will force the docfx engine to generate unique similar filenames for naming conflict cases?

georgius: > docker is still linux.
> Will return true for all platforms will force the docfx engine to generate unique similar filenames for naming conflict cases?

Will return true for all platforms except Linux, doesn't matter of file system and force the docfx engine to generate unique filenames for naming conflict case.

superyyrrzz: > I understand that docfx is already working just fine on Windows (which already has a case insensitive file system).
Is it difficult to have the same behavior as well on other platforms (linux, macOS)?
That would make it consistent across all platforms as results + results files (html) will be portable between various servers (linux, windows) where we wish to host the documentation.

I totally agree with you about this point, but I will be cautious for this behavior change as it will cause many new warnings complaining invalid link on existing documents.

farcasclaudiu: Is it that existing documents can be regenerated with --force parameter and therefore force the linking to be corrected?

superyyrrzz: Yes, it is possible, but I believe some Windows users will not be happy to be forced to fix it, as all case-insensitive links work well before. This is a breaking change we do not want to take per [current strategy in docfx roadmap](https://github.com/dotnet/docfx/blob/dev/Roadmap.md#v2-strategy).

yufeih (approved):

superyyrrzz:
@georgius Thank you for the contribution and effort to test the fix!

PR #6906, opened 2020-12-17
Support file close scenario

928PJY:
[AB#336937](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/336937)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6906)


src/docfx/build/Builder.cs
```diff
@@ -49,7 +49,6 @@ public void Build(params string[] files)
         {
             try
             {
-                _errors.Clear();
                 Watcher.StartActivity();
```
yufeih, line 52: why remove clear?

src/docfx/serve/lsp/TextDocumentHandler.cs
```diff
@@ -37,31 +38,39 @@ internal class TextDocumentHandler : ITextDocumentSyncHandler
...
 
-            _languageServerBuilder.QueueBuild();
+            _languageServerBuilder.QueueBuild(eventTimeStamp);
             return Unit.Task;
         }
 
         public Task<Unit> Handle(DidCloseTextDocumentParams notification, CancellationToken token)
         {
-            _notificationListener.OnNotificationHandled();
+            var eventTimeStamp = DateTime.UtcNow;
```
yufeih, line 67: We may not need this timestamping according to https://github.com/OmniSharp/csharp-language-server-protocol/issues/249#issuecomment-636491490

> With the current implementation you should never see concurrent threads for "write" (Serial) operations (didchange, rename, etc). You could however have multiple threads "reading" (Parallel) at the same time (for completions and such).

yufeih (approved):

PR #6908, opened 2020-12-18
Event debounce for language server

928PJY:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6908)

### DIFF
--- a/src/docfx/serve/LanguageServerBuilder.cs
+++ b/src/docfx/serve/LanguageServerBuilder.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
+using System.Threading.Channels;
 using System.Threading.Tasks;
 using OmniSharp.Extensions.LanguageServer.Protocol.Models;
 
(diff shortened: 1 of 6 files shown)


yufeih (approved):

src/docfx/serve/LanguageServerBuilder.cs
```diff
@@ -30,28 +31,47 @@ internal class LanguageServerBuilder
             _errorList = new();
             _languageServerPackage = languageServerPackage;
             _builder = new(_errorList, languageServerPackage.BasePath, options, _languageServerPackage);
-            _buildSemaphore = new(0);
+            _buildChannel = Channel.CreateUnbounded<bool>();
             _ = StartAsync();
         }
 
         public void QueueBuild()
```
yufeih, line 38: will different actions have different denounce timeout? E.g. textChange is 1s, but textOpen is immediate?

928PJY: It is nice to have and we can re-consider about it in the future?

yufeih (approved):

PR #6914, opened 2020-12-21
limit contentFallback to markdown

xichan-ms:
[AB#350178](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/350178)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6914)

### DIFF
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -503,7 +503,7 @@ outputs:
     }
   docs/b.json:
 ---
-# Disable content fallback for loc page
+# Disable markdown fallback for loc page
 locale: zh-cn
 repos:
   https://docs.com/page-fallback#master:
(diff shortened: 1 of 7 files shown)


yufeih (changes requested):
I thought we talked about redefining `contentFallback` such that `codesnippet` fallback is not affected by the `contentFallback` setting, this implementation isn't doing that.

Look for all the references to `MarkdownContext.ReadFile` and we can tell the difference between a code snippet include and a markdown include, you'll probably need to adjust the `ReadFileDelegate` delegate to take an additional `bool contentFallback` parameter.

docs/specs/schema.yml
```diff
@@ -529,6 +529,35 @@ outputs:
...
+        _themes/ContentTemplate/schemas/TestData.schema.json: |
+          {
+            "contentFallback": false,
+            "properties": { 
+              "content": {"type":"string","contentType": "markdown"}
+              }
+          }
+        _themes/ContentTemplate/TestData.html.primary.tmpl:
+outputs:
+  unit.json:
```
yufeih, line 559: Assert the `content` property output to ensure that we have the expected HTML code output

xichan-ms: OK

yufeih (approved):

PR #6925, opened 2020-12-22
Support refresh credential through LSP in language server mode

928PJY:

This PR is built on #6929

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6925)

### DIFF
--- a/src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
+++ b/src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using MediatR;
+using OmniSharp.Extensions.JsonRpc;
+
+namespace Microsoft.Docs.Build
+{
+    [Parallel]
+    [Method("docfx/userCredentialRefresh", Direction.ServerToClient)]
+    public record CredentialRefreshParams : IRequest<object>
+    {
+        public CredentialType Type { get; init; }
+    }
+
+    public record CredentialRefreshResponse
+    {
+        public string? Result { get; init; }
+
+        public ResponseError? Error { get; init; }
+    }
+
+    public record CredentialRefreshResult
+    {
+        public string? Token { get; init; }
+    }
+}
--- a/src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
+++ b/src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
@@ -9,14 +9,14 @@ namespace Microsoft.Docs.Build
 {
     [Parallel]
     [Method("docfx/userCredentialRefresh", Direction.ServerToClient)]
-    public record CredentialRefreshParams : IRequest<object>
+    public record CredentialRefreshParams : IRequest<CredentialRefreshResponse>
     {
         public CredentialType Type { get; init; }
     }
 
     public record CredentialRefreshResponse
     {
-        public string? Result { get; init; }
+        public CredentialRefreshResult? Result { get; init; }
 
         public ResponseError? Error { get; init; }
     }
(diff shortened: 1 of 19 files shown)


src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
```diff
@@ -0,0 +1,28 @@
...
+
+using System.Threading.Tasks;
+using MediatR;
+using OmniSharp.Extensions.JsonRpc;
+
+namespace Microsoft.Docs.Build
+{
+    [Parallel]
+    [Method("docfx/userCredentialRefresh", Direction.ServerToClient)]
+    public record CredentialRefreshParams : IRequest<CredentialRefreshResponse>
```
yufeih, line 12: use `record` only when you need _value equality semantic_
```suggestion
    public class CredentialRefreshParams : IRequest<CredentialRefreshResponse>
```

src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public CredentialType Type { get; init; }
+    }
+
+    public record CredentialRefreshResponse
+    {
+        public CredentialRefreshResult? Result { get; init; }
+
+        public ResponseError? Error { get; init; }
```
yufeih, line 21: Is this the standard way of handling errors in LSP (or jsonrpc)? There is a dedicated [Error Object](https://www.jsonrpc.org/specification#error_object) in jsonrpc to represent error and it seems to be at a different level.

```suggestion
       ~~public ResponseError? Error { get; init; }~~
```

src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using MediatR;
+using OmniSharp.Extensions.JsonRpc;
+
+namespace Microsoft.Docs.Build
+{
+    [Parallel]
```
yufeih, line 10: Should this be `serial`? How would client process multiple sign in request at the same time
```suggestion
    [Serial]
```

src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
```diff
@@ -0,0 +1,28 @@
...
+    [Method("docfx/userCredentialRefresh", Direction.ServerToClient)]
+    public record CredentialRefreshParams : IRequest<CredentialRefreshResponse>
+    {
+        public CredentialType Type { get; init; }
```
yufeih, line 14: Instead of using a enum of token types, consider use URLs and HTTP standard response headers at the protocol layer (They are both standard), if we have other credentials in the future, the protocol does not need to change.
```suggestion
        public string BaseUrl { get; init; }
```

src/docfx/serve/lsp/credentialRefresh/CredentialRefreshParams.cs
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+using MediatR;
+using OmniSharp.Extensions.JsonRpc;
+
+namespace Microsoft.Docs.Build
+{
+    [Parallel]
+    [Method("docfx/userCredentialRefresh", Direction.ServerToClient)]
```
yufeih, line 11: consider use a verb (as other lsp server to client protocols like `registerCapability`), and also the term `refresh` is a bit limited because we could theoretically use this protocol to retrieve the credentials for the first time.
```suggestion
    [Method("docfx/getCredential", Direction.ServerToClient)]
```

yufeih (changes requested):

Issue #6926, opened 2020-12-22
Using $schema to get completion causes a build warning


kzu:
**Operating System**: `Windows`

**DocFX Version Used**: 2.56.6

**Steps to Reproduce**:

1. Init sample docfx 
2. Add the following to the top of the docfx.json: 

```json
{
  "$schema": "http://json.schemastore.org/docfx",
  "metadata": [
   ...
}
```

3. Run docfx

**Expected Behavior**:

No warnings

**Actual Behavior**:

`Warning:$schema is not a recognized command name, ignored.`

Enabling a good editing experience is important, IMHO:

![image](https://user-images.githubusercontent.com/169707/102932430-088e2200-447f-11eb-9bcb-6834ae1b18c7.png)

![image](https://user-images.githubusercontent.com/169707/102932453-15ab1100-447f-11eb-8322-8f142cfaab79.png)

PR #6927, opened 2020-12-22
Don't warn for $schema element

kzu:
Adding a `$schema` attribute pointing to the official schema is a nice way to get completion in an editor. This caused a warning during build, however.

This checks for that element when building the commands so we don't warn for what is already a supported scenario.

Fixes Using $schema to get completion causes a build warning #6926

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6927)


yufeih (approved):
Thanks for the fix, looks good to me 🔥

PR #6929, opened 2020-12-23
Extract credential handler and apply to all the http client

928PJY:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6929)


yufeih (changes requested):
We may not need this change if using [Polly Retry to refresh authorization](https://github.com/App-vNext/Polly/wiki/Retry#retry-to-refresh-authorization)

herohua:
>     internal sealed class GitHubAccessor

should we implement `IDisposable`?

---
Refers to: src/docfx/lib/github/GitHubAccessor.cs:22 in c6ae779. [](commit_id = c6ae779dce083f233c2b34b51bb60612aca72438, deletion_comment = False)

src/docfx/config/ops/CredentialHandler.cs
```diff
@@ -0,0 +1,186 @@
...
+using System.Net.Http;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Identity;
+using Azure.Security.KeyVault.Secrets;
+
+namespace Microsoft.Docs.Build
+{
+    internal class CredentialHandler : DelegatingHandler
```
herohua, line 15: >CredentialHandler [](start = 19, length = 17)

shall we call this `OpsCredentialHandler`?

src/docfx/config/ops/CredentialHandler.cs
```diff
@@ -0,0 +1,186 @@
...
+                if (TryExtractDocsEnvironmentFromUrl(request.RequestUri?.AbsoluteUri, out _))
+                {
+                    return await HandleOpsRequest(request, cancellationToken);
+                }
+
+                return await base.SendAsync(request, cancellationToken);
+            }
+        }
+
+        public static string BuildServiceEndpoint(DocsEnvironment? environment = null)
```
herohua, line 61: >BuildServiceEndpoint [](start = 29, length = 20)

It is kind of weird to me that `CredentialManager` exposes a method called `BuildServiceEndpoint`. Shall we consider extracting environment and endpoint info into a separate class, e.g. named `OpsEnvironments`?

src/docfx/config/PreloadConfig.cs
```diff
@@ -57,27 +57,16 @@ internal class PreloadConfig
         /// </summary>
         public string? DocsRepositoryOwnerName { get; init; }
 
-        public Action<HttpRequestMessage> GetCredentialProvider()
+        public CredentialHandler GetCredentialHandler()
         {
-            var rules = Http.OrderByDescending(pair => pair.Key, StringComparer.Ordinal).ToArray();
-
-            return message =>
+            return new Lazy<CredentialHandler>(() =>
```
herohua, line 62: >new Lazy<CredentialHandler> [](start = 19, length = 27)

what is the meaning for this `Lazy` as we immediately invoke `.Value`?

test/docfx.Test/build/OpsConfigAdapterTest.cs
```diff
@@ -36,7 +37,21 @@ public static async Task AdaptOpsServiceConfig(string url, string expectedJson)
                 return;
             }
 
-            var accessor = new OpsAccessor(null, request => request.Headers.Add("X-OP-BuildUserToken", token));
+            var credentialHanlder = new CredentialHandler(
```
herohua, line 40: >Hanlder [](start = 26, length = 7)

typo

src/docfx/restore/CredentialHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Docs.Build
+{
+    internal class CredentialHandler : DelegatingHandler
+    {
+        private readonly List<KeyValuePair<string, HttpConfig>> _credentials;
+
+        public CredentialHandler(List<KeyValuePair<string, HttpConfig>> credentials, HttpMessageHandler innerHandler)
```
herohua, line 16: >List [](start = 33, length = 4)

`IReadOnlyList`

src/docfx/restore/CredentialHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Docs.Build
+{
+    internal class CredentialHandler : DelegatingHandler
+    {
+        private readonly List<KeyValuePair<string, HttpConfig>> _credentials;
```
herohua, line 14: >List [](start = 25, length = 4)

`IReadOnlyList`

src/docfx/restore/CredentialHandler.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        public CredentialHandler(List<KeyValuePair<string, HttpConfig>> credentials, HttpMessageHandler innerHandler)
+            : base(innerHandler)
+        {
+            _credentials = credentials;
+        }
+
+        public CredentialHandler Create(HttpMessageHandler innerHandler)
```
herohua, line 22: >Create [](start = 33, length = 6)

this looks weird - why do we want to new a `CredentialHandler` with exiting credentials

src/docfx/config/ops/OpsCredentialHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+using System.Net.Http;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Identity;
+using Azure.Security.KeyVault.Secrets;
+
+namespace Microsoft.Docs.Build
+{
+    internal class OpsCredentialHandler : DelegatingHandler
```
herohua, line 14: >OpsCredentialHandler [](start = 19, length = 20)

should `OpsCredentialHandler` be a `CredentialHandler`?

src/docfx/config/ops/OpsCredentialHandler.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            // Default header which allows fallback to public data when credential is not provided.
+            request.Headers.TryAddWithoutValidation("X-OP-FallbackToPublicData", true.ToString());
+            await FillOpsToken(request);
+            return await base.SendAsync(request, cancellationToken);
+        }
+
+        private static async Task FillOpsToken(HttpRequestMessage request, DocsEnvironment? environment = null)
+        {
+            if (IsSameDocsEnvironmentRequest(request))
```
yufeih, line 38: This line makes me want to get away from using `DelegatingHandler` for middlewares. Its contract makes passing variables like `environment` here so much harder.

All we need is to turn the existing credential provider from `Action<HttpRequestMessage>` to:

```csharp
delegate Task<HttpResponseMessage> CredentialProvider(HttpRequestMessage request, Func<HttpRequestMessage, Task<HttpResponseMessage>> next);
```

Then you only need to change places that calls `_credentialProvider?.Invoke(request);`

yufeih (changes requested):

src/docfx/config/ops/CredentialHandler.cs
928PJY: it handles all the credentials not only ops

src/docfx/config/ops/CredentialHandler.cs
928PJY: moved back to opsAccessor

src/docfx/restore/CredentialHandler.cs
928PJY: it is to create `credentialHandler` with a different `innerHandler`

src/docfx/config/ops/OpsCredentialHandler.cs
928PJY: They handling different kind of credentials, opsCredentialHanlder doesn't inherit from the `CredentialHandler`, Rename `CredentialHandler` may avoid the confusion, but I don't have a better idea

src/docfx/config/ops/OpsCredentialHandler.cs
928PJY: Then the _credentialProvider.invoke need to be called everywhere, and it looks wired to keep a state in the delegate when we need to keep the refreshedToken.

src/docfx/config/PreloadConfig.cs
928PJY: To get the same instance when it gets called multiple times

src/docfx/lib/HttpRequestUtility.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Net.Http;
+
+namespace Microsoft.Docs.Build
+{
+    internal static class HttpRequestUtility
```
live1206, line 8: not in use?

src/docfx/restore/CredentialProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public Dictionary<string, string> GetCredentials(HttpRequestMessage request)
+        {
+            var credentials = new Dictionary<string, string>();
+            if (request.RequestUri?.ToString() is string url)
+            {
+                GetCredentialsCore(credentials, url, _credentials);
+                GetCredentialsCore(credentials, url, OpsAccessor.FallBackCredentials);
+            }
+            return credentials;
```
live1206, line 27: should this be cached? Not constructed per request.

src/docfx/lib/HttpRequestUtility.cs
928PJY: it is an extension of `HttpRequestMessage `, used in `CredentialHandler.cs`.

live1206 (approved):

PR #6962, opened 2021-01-05
Support refresh credential based on lsp

928PJY:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6962)


src/docfx/restore/CredentialProvider.cs
```diff
@@ -4,36 +4,92 @@
...
         }
 
-        public Dictionary<string, string> GetCredentials(HttpRequestMessage request)
+        public bool FillInCredentials(HttpRequestMessage request)
         {
+            var credentialUpdated = false;
             if (request.RequestUri?.ToString() is string url)
             {
-                return _credentialCache.GetOrAdd(url, _ =>
+                var credentials = GetCredentials(url);
```
herohua, line 38: >GetCredentials [](start = 34, length = 14)

should credential refresher knows whether credentials have been updated?

src/docfx/restore/CredentialProvider.cs
```diff
@@ -4,36 +4,92 @@
...
+        {
+            if (request.RequestUri?.ToString() is string url && _getCredential != null)
+            {
+                await s_credentialRefreshSemaphore.WaitAsync(CancellationToken.None);
+                if (!FillInCredentials(request))
+                {
+                    var newCredentials = await _getCredential.Invoke(url);
+                    foreach (var (key, value) in newCredentials)
+                    {
+                        s_refreshedCredentials[key] = value;
```
herohua, line 70: >s_refreshedCredentials [](start = 24, length = 22)

seems that we don't need `s_refreshedCredentials` and the semaphore. can we build out a local variable `refreshedCredentials` and update `s_credentialCache` directly?

src/docfx/restore/CredentialProvider.cs
```diff
@@ -4,36 +4,92 @@
...
+        public CredentialProvider(
+            IReadOnlyDictionary<string, HttpConfig> configCredentials, Func<string, Task<Dictionary<string, HttpConfig>>>? getCredential = null)
         {
-            _credentials = credentials;
+            _configCredentials = configCredentials;
+            _getCredential = getCredential;
         }
 
-        public Dictionary<string, string> GetCredentials(HttpRequestMessage request)
+        public bool FillInCredentials(HttpRequestMessage request)
```
herohua, line 33: >FillInCredentials [](start = 20, length = 17)

I am thinking keeping original output `Dictionary<string, string>` might be cleaner...

src/docfx/config/ConfigLoader.cs
```diff
@@ -64,7 +72,7 @@ where glob(file)
             var preloadConfig = JsonUtility.ToObject<PreloadConfig>(errors, preloadConfigObject);
 
             // Download dependencies
-            var credentialProvider = preloadConfig.GetCredentialProvider();
+            var credentialProvider = preloadConfig.GetCredentialProvider(getCredential);
```
herohua, line 75: >getCredential [](start = 73, length = 13)

`credentialRefresher` might be a better name

src/docfx/restore/CredentialProvider.cs
928PJY: semaphore is used to avoid multiple thread refresh the credential in the same time

---
In reply to: [551896148](https://github.com/dotnet/docfx/pull/6962#discussion_r551896148) [](ancestors = 551896148)

src/docfx/config/ConfigLoader.cs
928PJY: To make the contract more general, it can be used to get credential in the future not only refresh credential

---
In reply to: [551897867](https://github.com/dotnet/docfx/pull/6962#discussion_r551897867) [](ancestors = 551897867)

src/docfx/restore/CredentialHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+
+                response = await next(request);
+                if (response.StatusCode != HttpStatusCode.Unauthorized)
+                {
+                    break;
+                }
+
+                needRefresh = true;
+                _credentialCache.Clear();
+                request = CopyHttpRequestMessage(request);
```
herohua, line 40: can we just remove the value corresponding to `url`

src/docfx/serve/LanguageServerBuilder.cs
```diff
@@ -60,14 +61,17 @@ public async void Run(CancellationToken cancellationToken = default)
                     await WaitToTriggerBuild(cancellationToken);
 
                     using var progressReporter = await CreateProgressReporter();
+                    await Task.Yield();
```
herohua, line 64: >await Task.Yield(); [](start = 20, length = 19)

what is this for?

src/docfx/restore/CredentialHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+                            }
+                        }
+                    }
+                }
+
+                return credentials;
+            });
+        }
+
+        private static HttpRequestMessage CopyHttpRequestMessage(HttpRequestMessage request)
```
herohua, line 82: looks like this implementation has some risk. how about
[https://stackoverflow.com/questions/25044166/how-to-clone-a-httprequestmessage-when-the-original-request-has-content](https://stackoverflow.com/questions/25044166/how-to-clone-a-httprequestmessage-when-the-original-request-has-content)

src/docfx/serve/LanguageServerBuilder.cs
928PJY: Add a description.

src/docfx/build/Builder.cs
```diff
@@ -16,11 +17,14 @@ internal class Builder
         private readonly CommandLineOptions _options;
         private readonly Watch<DocsetBuilder[]> _docsets;
         private readonly Package _package;
+        private readonly Func<HttpRequestMessage, bool, Task<HttpConfig?>>? _getCredential;
 
-        public Builder(CommandLineOptions options, Package package)
+        public Builder(
+            CommandLineOptions options, Package package, Func<HttpRequestMessage, bool, Task<HttpConfig?>>? getCredential = null)
```
yufeih, line 23: Consider give this Func a dedicated interface or delegate?

src/docfx/restore/CredentialHandler.cs
```diff
@@ -0,0 +1,113 @@
...
+namespace Microsoft.Docs.Build
+{
+    internal class CredentialHandler
+    {
+        private const int RetryCount = 3;
+
+        private readonly Func<HttpRequestMessage, bool, Task<HttpConfig?>>[] _credentialProviders;
+        private readonly ConcurrentDictionary<string, Task<Dictionary<string, string>>> _credentialCache = new();
+
+        public CredentialHandler(params Func<HttpRequestMessage, bool, Task<HttpConfig?>>[] credentialProviders) => _credentialProviders = credentialProviders;
```
yufeih, line 21: consider break the `HttpRequestMessage` in the signature into: `Func<string, HttpConfig?, bool, Task<HttpConfig?>>` because:
- The interface is now symmetric, both input and output speaks `HttpConfig`
- The comparison in `LspCredentialProvider` becomes a reference equals check
- Can eliminate the need of `CopyHttpRequestMessage` because you know what old headers to remove

src/docfx/serve/LanguageServerBuilder.cs
yufeih: Consider move this line right before calling `_builder.Build`, in case some other code that calls lsp API are inserted between lin 69 and line 73

src/docfx/serve/lsp/LanguageServerProgressReporter.cs
```diff
@@ -28,5 +28,10 @@ public void Report(string message)
                 Message = message,
             });
         }
+
+        public void OnCompleted()
+        {
+            _workDoneManager.OnCompleted();
```
yufeih, line 34: move this line to the `Dispose` method to let `using` automatically complete the `_workDoneManager`?

test/docfx.Test/lsp/LanguageServerTestCommand.cs
```diff
@@ -16,6 +16,8 @@ public class LanguageServerTestCommand
 
         public Dictionary<string, JToken> ExpectDiagnostics { get; init; }
 
+        public GetCredentialRequest ExpectGetCredentialRequest { get; init; }
```
yufeih, line 19: Move `Response` outside of `GetCredentialRequest` to simplify the structure, the `Response` field could also potentially be shared with other `expectXYZRequest` calls

```suggestion
        public JToken ExpectGetCredentialRequest { get; init; }
        public JToken Response { get; init; }
```

src/docfx/serve/lsp/getCredential/LanguageServerCredentialProvider.cs
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    var getCredentialResponse = await _languageServer.SendRequest(
+                        new GetCredentialParams() { Url = url },
+                        CancellationToken.None);
+                    foreach (var (key, value) in getCredentialResponse.Http)
+                    {
+                        _credentials[key] = value;
+                    }
+                }
+                catch (Exception ex)
```
yufeih, line 56: Remove this catch to the exception will propagate up the stack and generate a `download-failed` error message and stop the build, so authentication failures are handled the same way as other transient HTTP errors.

test/docfx.Test/lsp/LanguageServerTestClient.cs
```diff
@@ -169,14 +193,26 @@ private async Task<ILanguageClient> InitializeClient(string workingDirectory, Pa
                         _notificationSync.TrySetException(new InvalidOperationException(message.Message));
                     }
                 })
+                .OnRequest("docfx/getCredential", async (GetCredentialParams @params) =>
+                {
+                    if (_credentialRefreshHandler == null)
```
yufeih, line 198: The client may receive `docfx/getCredential` request and hit this line before it had the change to process `expectGetCredentialRequest` and set the delegate there. This could use the same synchronization mechanism as notification, basically this `OnRequest` callback puts a `(JToken request, TaskCompletionSource<JToken> response)` to a member variable, and in the `ProcessCommand` method, first call `await SynchronizeNotifications();`, then return the result through `TaskCompletionSource<JToken>`.

yufeih (approved):

PR #6965, opened 2021-01-05
Disable hierarchy validation on partial build

yufeih:
Learn hierarchy validation requires full hierarchy which we don't currently have for single file validation.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6965)

### DIFF
--- a/src/docfx/build/DocsetBuilder.cs
+++ b/src/docfx/build/DocsetBuilder.cs
@@ -180,7 +180,7 @@ namespace Microsoft.Docs.Build
                     () => _repositoryProvider.Save(),
                     () => _errors.AddRange(_githubAccessor.Save()),
                     () => _errors.AddRange(_microsoftGraphAccessor.Save()),
-                    () => _jsonSchemaTransformer.PostValidate(),
+                    () => _jsonSchemaTransformer.PostValidate(files != null),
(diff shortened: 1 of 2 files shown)


src/docfx/build/DocsetBuilder.cs
```diff
@@ -180,7 +180,7 @@ public void Build(string[]? files)
                     () => _repositoryProvider.Save(),
                     () => _errors.AddRange(_githubAccessor.Save()),
                     () => _errors.AddRange(_microsoftGraphAccessor.Save()),
-                    () => _jsonSchemaTransformer.PostValidate(),
+                    () => _jsonSchemaTransformer.PostValidate(files != null),
```
herohua, line 183: how about sharing the logic with line 165

yufeih: I'd love to, but the nullability flow analysis isn't smart enough to know that when `isPartialBuild` is `true`, `files` is non-nullable

image

yufeih: https://github.com/dotnet/roslyn/issues/50250

OsmondJiang (approved):

PR #6967, opened 2021-01-06
Run all spec tests in single file mode

yufeih:
Single file test build each content file and verifies the union of each file build result is the same as expected output. Implies dryRun.

[AB#356173](https://dev.azure.com/ceapex/Engineering/_workitems/edit/356173/)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6967)

### DIFF
--- a/test/docfx.Test/DocfxTest.cs
+++ b/test/docfx.Test/DocfxTest.cs
@@ -54,9 +54,15 @@ namespace Microsoft.Docs.Build
         {
             yield return "";
 
-            if (!spec.DryRunOnly && !spec.NoDryRun && spec.Outputs.ContainsKey(".errors.log"))
+            var hasError = spec.Outputs.ContainsKey(".errors.log");
+            if (hasError && !spec.DryRunOnly && !spec.NoDryRun)
             {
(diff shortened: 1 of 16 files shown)


test/docfx.Test/DocfxTest.cs
```diff
@@ -281,6 +326,22 @@ private static bool IsRequiredOutput(string name)
                 && fileName != "server-side-dependent-list.txt";
         }
 
+        private static bool IsContentFile(string path)
```
928PJY, line 329: This is neccessary and you cannot distinguish a content file rightly until you parse the config.

928PJY (approved):
:shipit:

test/docfx.Test/DocfxTest.cs
yufeih: It is to skip tests that only has a `docfx.yml` and `a.md`, running single file build on these tests isn't quite meaningful.

PR #6982, opened 2021-01-12
Upgrade System.CommandLine

mingwli:
The original package is removed from [dotnet-corefx](https://dotnet.myget.org/gallery/dotnet-corefxlab)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6982)


src/docfx/cli/CommandLine.cs
```diff
@@ -0,0 +1,125 @@
...
+            command.AddOption(new Option<bool>(
+                "--no-cache", "Always fetch latest dependencies in build."));
+            return command;
+        }
+
+        private void Parse(string command, CommandLineOptions options) => (_command, _options) = (command, options);
+
+        private static void DefineCommonCommands(Command command)
+        {
+            command.AddArgument(new Argument<string>("WorkingDirectory", () => "."));
```
yufeih, line 117: Previously this is called `directory` with a description.

```
syntax.DefineParameter("directory", ref options.WorkingDirectory, "A directory that contains docfx.yml/docfx.json.");
```

yufeih: And https://github.com/dotnet/command-line-api/issues/1156 shows to use `{ Arity = ArgumentArity.ZeroOrOne }` to hide the wired `[default:]` thing.

src/docfx/cli/CommandLine.cs
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+
+        private (string command, CommandLineOptions options) Command
+            => (_command, _options);
+
+        private Command BaseCommand(string name)
+            => new Command(name)
+            {
+                Handler = CommandHandler.Create((CommandLineOptions options) => Parse(name, options)),
```
yufeih, line 55: consider move this to `docfx.cs` and hook the handler to the corresponding function directly.

src/docfx/cli/CommandLine.cs
yufeih: Same is true for `new` command, previously `docfx new` shows a template list, now it causes an error because `templateName` is now a required argument.

yufeih (approved):

PR #6979, opened 2021-01-11
Run dotnet-trace in regression test

yufeih:
Starting from .NET 5, we could use `dotnet-trace` to [collect a trace from app start](https://docs.microsoft.com/en-us/dotnet/core/diagnostics/dotnet-trace#use-diagnostic-port-to-collect-a-trace-from-app-startup), this allows us to profile regression tests on CI server and upload performance trace to azure pipeline artifacts for each code change.

[AB#232284](https://dev.azure.com/ceapex/Engineering/_workitems/edit/232284/)
[AB#372082](https://dev.azure.com/ceapex/Engineering/_workitems/edit/372082)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6979)

PR #6989, opened 2021-01-14
Fix thread starvation due to PLINQ AsOrdered

yufeih:
[AB#359565](https://dev.azure.com/ceapex/Engineering/_workitems/edit/359565/)

This thread pool starvation problems was initially observed from https://github.com/dotnet/docfx/pull/6979, where building _AspNetCore.Docs_ took up to 13 minutes to complete on server after enabling CPU Sampling. Local test shows that the build of _AspNetCore.Docs_ ranges between 10 seconds (the optimal case) and > 1 minute.

The cause is due to thread pool starvation where every task wait on loading external xrefmap. PLINQ `AsOrdered` seem to be the cause of this problem. Switching it to normal `Parallel` loops makes the problem disappear, and now the build consistently finishes under 10s.

image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6989)

### DIFF
--- a/src/docfx/build/xref/ExternalXrefMapLoader.cs
+++ b/src/docfx/build/xref/ExternalXrefMapLoader.cs
@@ -21,10 +21,17 @@ namespace Microsoft.Docs.Build
 
         public static ExternalXrefMap Load(Config config, FileResolver fileResolver, ErrorBuilder errors)
         {
-            using (Progress.Start("Loading xref map"))
-            {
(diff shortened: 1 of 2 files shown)


herohua (approved):

herohua:
@yufeih Do we want to report back the issue to dotnet team?

PR #6997, opened 2021-01-18
Clean regression tests

mingwli:
- replace deprovision vsts-docs-pr repo with [windows-compatibility](https://cpubwin.visualstudio.com/_git/windows-compatibility) to make test pass
- add repos list documentation and source excel
- arrange pipeline order to match documentation for easier reference

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6997)

PR #7002, opened 2021-01-18
Use csv to get rid of binary xlsx

mingwli:
Fix binary file in #6997. 

\* Github support preview of csv file in browser:
image

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/7002)


docs/regression-test/test-repos.csv
```diff
@@ -0,0 +1,26 @@
+﻿page view,repos,version,loc,xref,conceptual,SDP conceptual,SDP Ref,"multi
```
yufeih, line 1: The first column is not the actual page view number, but rather the page view rank

PR #7012, opened 2021-01-21
Fix failing tests

live1206:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/7012)

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -252,10 +252,15 @@ namespace Microsoft.Docs.Build
                 return default;
             }
 
-            var (cmd, secret) = (
-                from header in config.GetHttpConfig(url)?.Headers
-                select (cmd: $"-c http.extraheader=\"{header.Key}: {header.Value}\"", secret: GetSecretFromHeader(header))).FirstOrDefault();
+            var httpConfig = config.GetHttpConfig(url);
+            if (httpConfig is null)
+            {
+                return default;
+            }
 
+            var (cmd, secret) = (
+                from header in httpConfig.Headers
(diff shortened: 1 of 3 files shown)


yufeih (commented):
Need to update this logic to:
- Call `GetHttpConfig` (which internally does sorting, so longest url match)

image

src/docfx/lib/git/GitUtility.cs
```diff
@@ -237,28 +237,31 @@ private static string Execute(string cwd, string commandLineArgs, bool stdout =
...
+                return ProcessUtility.Execute("git", commandLineArgs, cwd, stdout, secret);
             }
             catch (Win32Exception ex) when (ProcessUtility.IsExeNotFoundException(ex))
             {
                 throw Errors.System.GitNotFound().ToException(ex);
             }
         }
 
-        private static (string cmd, string[] secrets) GetGitCommandLineConfig(string url, PreloadConfig config)
+        private static (string cmd, string? secret) GetGitCommandLineConfig(string url, PreloadConfig config)
```
yufeih, line 248: `return default` produces `(null, null)`
```suggestion
        private static (string? cmd, string? secret) GetGitCommandLineConfig(string url, PreloadConfig config)
```

yufeih (approved):

src/docfx/lib/git/GitUtility.cs
live1206: Updated, L252 already broke it

yufeih: Yep. Better late than never.

PR #7039, opened 2021-02-02
Reset the verbose option since the SynchronizationContext doesn't flow by the ASPNetCore

928PJY:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/docfx/pull/7039)

### DIFF
--- a/src/docfx/serve/Serve.cs
+++ b/src/docfx/serve/Serve.cs
@@ -36,15 +36,19 @@ namespace Microsoft.Docs.Build
 
         private static async Task StartLanguageServer(HttpContext context, CommandLineOptions options, Package? package)
         {
-            if (context.WebSockets.IsWebSocketRequest)
+            // The SynchronizationContext is lost here, verbose needs to be reset
+            using (Log.BeginScope(options.Verbose))
             {
-                using var webSocket = await context.WebSockets.AcceptWebSocketAsync();
-                var stream = webSocket.AsStream();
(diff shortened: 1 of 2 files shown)


src/docfx/serve/lsp/LanguageServerHost.cs
```diff
@@ -21,38 +21,40 @@ internal class LanguageServerHost
             ILanguageServerNotificationListener? notificationListener = null)
         {
             using var cts = new CancellationTokenSource();
+            using (Log.BeginScope(commandLineOptions.Verbose))
```
yufeih, line 24: Move it to `StartLanguageServer` method because it is AspNetCore that lost this context. (Which means other APIs there in the future also needs to call `Log.BeginScope`)

src/docfx/serve/Serve.cs
```diff
@@ -36,15 +36,19 @@ void Configure(IApplicationBuilder app)
 
         private static async Task StartLanguageServer(HttpContext context, CommandLineOptions options, Package? package)
         {
-            if (context.WebSockets.IsWebSocketRequest)
+            // The SynchronizationContext is lost here, verbose needs to be reset
```
yufeih, line 39: Technically this is called execution context not synchronization context.
```suggestion
            // The execution context is lost here, verbose needs to be reset
```

yufeih (approved):

Issue #6345, opened 2020-07-31
Optimize display of parameter with default enum flags value


mikernet:
**Operating System**: Windows

**DocFX Version Used**: 2.56.2

**Template used**: N/A

If there is a single enum flags value that captures a default enum parameter value then that should be used instead of spitting out all the individual flags that match. You can see how it is doing that here: 

http://www.singulink.com/Docs/Singulink.IO.FileSystem/api/Singulink.IO.DirectoryPath.html#Singulink_IO_DirectoryPath_Parse_System_ReadOnlySpan_System_Char__Singulink_IO_PathFormat_Singulink_IO_PathOptions_

Instead of having this:
```c#
public static IDirectoryPath Parse(ReadOnlySpan<char> path, PathOptions options = PathOptions.NoReservedDeviceNames | PathOptions.NoLeadingSpaces | PathOptions.NoTrailingSpaces | PathOptions.NoTrailingDots | PathOptions.NoUnfriendlyNames)
```

It should have:
```c#
public static IDirectoryPath Parse(ReadOnlySpan<char> path, PathOptions options = PathOptions.NoUnfriendlyNames)
```

KalleOlaviNiemitalo:
Also described in <https://github.com/dotnet/docfx/issues/5694>.

Issue #5694, opened 2020-03-29
How to omit constructor arguments on attribute?


QFSW:
**Operation System**: Windows

**DocFX Version Used**: 2.51.0.0

Some attributes end up with very unwieldy constructors
![image](https://user-images.githubusercontent.com/11485449/77846639-5a836f00-71af-11ea-9009-7e070164c252.png)

Is it possible to still include the attribute decoration but remove all of its arguments from the api reference?

KalleOlaviNiemitalo:
That output makes me wonder if there is a bug where `Platform.AllPlatforms` already includes all the other bits but DocFX lists them regardless.

QFSW:
If this helps you understand the issue any better
In my code I have a bitwise flags enum `Platform`, so each bit represents one platform
Hence `AllPlatforms` is defined as `~0` to cover all platforms

KalleOlaviNiemitalo:
I think the bug is here: <https://github.com/dotnet/docfx/blob/ac8024bc50bd47d748ff99791a70e6f1ef6152d8/src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/CSYamlModelGenerator.cs#L896-L909>

KalleOlaviNiemitalo:
I mean, if `Platform.AllPlatforms == (Platforms)~0`, and the argument of the attribute constructor is `Platforms.AllPlatforms`, then DocFX should just emit that, rather than a huge bitwise OR expression. If that bug were fixed, then your HTML would get a much shorter syntax for the attribute, which could be good enough for you even though it would not satisfy your stated request.

DocFX could use the same algorithm as in `Enum.ToString`: convert the values of all enum constants to `ulong`, try them one by one from largest to smallest, subtract out the bits that matched, and finally list the corresponding names ordered by ascending value.

PR #7043, opened 2021-02-05
Added enum flags folding.

mikernet:
Resolves https://github.com/dotnet/docfx/issues/6345
Resolves https://github.com/dotnet/docfx/issues/5694

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/docfx/pull/7043)

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/CSYamlModelGenerator.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/CSYamlModelGenerator.cs
@@ -961,7 +961,8 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
             var enumType = GetTypeSyntax(namedType);
             if (EqualityComparer<T>.Default.Equals(value, default))
             {
-                return Array.Empty<ExpressionSyntax>();
+                string defaultFlagName = flags.FirstOrDefault(f => EqualityComparer<T>.Default.Equals(f.Value, default)).Name;
+                return defaultFlagName != null ? new[] { GetFlagExpression(defaultFlagName) } : Array.Empty<ExpressionSyntax>();
             }
             List<(string Name, T Value)> sortedFlags = flags.OrderByDescending(p => p.Value).ToList();
             if (sortedFlags.Count == 0)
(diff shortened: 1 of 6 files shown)


src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/CSYamlModelGenerator.cs
```diff
@@ -935,95 +931,72 @@ where object.Equals(value, pair.ConstantValue)
...
+        {
+            var enumType = GetTypeSyntax(namedType);
+            if (EqualityComparer<T>.Default.Equals(value, default))
+            {
+                string defaultFlagName = flags.FirstOrDefault(f => EqualityComparer<T>.Default.Equals(f.Value, default)).Name;
+                return defaultFlagName != null ? new[] { GetFlagExpression(defaultFlagName) } : Array.Empty<ExpressionSyntax>();
+            }
+            var negativeFlags = flags.Where(p => Comparer<T>.Default.Compare(p.Value, default) < 0);
+            var positiveFlags = flags.Where(p => Comparer<T>.Default.Compare(p.Value, default) > 0);
+            var sortedFlags = negativeFlags.OrderByDescending(p => p.Value).Concat(positiveFlags.OrderByDescending(p => p.Value)).ToList();
```
yufeih, line 969: Is `sortedFlags` sorted by bit count? If so, `0b_1000` > `0b_0111` but has less bit count.

mikernet: It's sorted by descending value, not by bit count. I've yet to run into a situation where this is an issue in practice - unless you have something **very** odd going on with your bit flags, it works out fine because all that matters is that flags that hold multiple values are considered before the individual values, which this accomplishes.

yufeih: This does not affect the output because you would match the higher bits first.

yufeih (approved):
LGTM 👍. Thanks for the pull request 🔥🔥

PR #7074, opened 2021-02-21
feat: use client cert to access Graph API

herohua:
[AB#378857](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/378857)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/docfx/pull/7074)

### DIFF
--- a/src/docfx/lib/msgraph/MicrosoftGraphAccessor.cs
+++ b/src/docfx/lib/msgraph/MicrosoftGraphAccessor.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Graph;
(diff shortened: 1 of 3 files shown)


src/docfx/lib/msgraph/MicrosoftGraphAccessor.cs
```diff
@@ -90,5 +92,21 @@ private async Task<T> SendRequest<T>(string api, Func<Task<T>> func)
...
         }
+
+        private static X509Certificate2 GetCertificateFromStoreByThumbprint(string certThumbprint)
+        {
+            using var store = new X509Store(StoreLocation.LocalMachine);
+            store.Open(OpenFlags.ReadOnly);
+
+            var certCollection = store.Certificates;
+            var currentCerts = certCollection.Find(X509FindType.FindByTimeValid, DateTime.Now, false);
+            var matchedCerts = currentCerts.Find(X509FindType.FindByThumbprint, certThumbprint, false);
```
zhiliangxu, line 103: validOnly should be set to true. Please also make it explicit: `validOnly: true`. With that, we probably don't need another `Find` with `X509FindType.FindByTimeValid`

src/docfx/lib/msgraph/MicrosoftGraphAccessor.cs
```diff
@@ -90,5 +92,21 @@ private async Task<T> SendRequest<T>(string api, Func<Task<T>> func)
...
+            if (matchedCerts.Count == 0)
```
partychen, line 104: should we add store.Close() after all ?

yufeih (approved):

PR #7091, opened 2021-02-25
Enable tests with DOCS_GITHUB_TOKEN

yufeih:
Removed a couple of corner case restore test to reduce the amount of secret management.

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/docfx/pull/7091)


docs/specs/restore.yml
```diff
@@ -365,116 +365,3 @@ outputs:
         "docs/a.md": [{ "source": "docs/b.md", "type": "file" }]
       }
     }
----
-# Show error for system service account SSO issue
```
mingwli, line 369: can we simply `skip`, incase adding them back?

mingwli (approved):

PR #7122, opened 2021-03-04
throw on invalid xrefmap

v-pegao:
[AB#368767](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/368767)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/docfx/pull/7122)


yufeih:
Talked with @live1206, we think the original version gives enough information (including callstack) for diagnostics. `content` is huge (up to hundreds of megabytes), it isn't suitable to print to log. Thus close this change.

PR #7125, opened 2021-03-04
[Security] Certificate revocation check is not done properly

partychen:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/docfx/pull/7125)


src/docfx/lib/github/GitHubAccessor.cs
```diff
@@ -24,6 +24,7 @@ internal sealed class GitHubAccessor
         private static readonly Uri s_url = new("https://api.github.com/graphql");
 
         private readonly HttpClient? _httpClient;
+        private readonly HttpClientHandler _httpClientHandler = new() { CheckCertificateRevocationList = true };
```
live1206, line 27: declare this before L44, only used within that closure, not for entire class

live1206 (approved):

PR #7129, opened 2021-03-04
Onboard to scheduled LGTM run

TianqiZhang:
###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/dotnet/docfx/pull/7129)

### DIFF
--- a/azure-pipelines-LGTM.yml
+++ b/azure-pipelines-LGTM.yml
@@ -32,22 +32,23 @@ jobs:
     vmImage: 'windows-latest'
   steps:
 
-  # Run CredScan
-  - task: CredScan@2
-    displayName: Security - CredScan
-    inputs:
-      toolMajorVersion: 'V2'
-
   # Install .NET Core sdk
   - task: UseDotNet@2
     displayName: 'Install .NET Core sdk $(dotnetVersion)'
     inputs:
       packageType: sdk
       version: $(dotnetVersion)
-
   - task: DotNetCoreCLI@2
+    displayName: dotnet build -c Release
     inputs:
       command: 'build'
+      arguments: '-c Release'
+
+  # Run CredScan
+  - task: CredScan@2
+    displayName: Security - CredScan
+    inputs:
+      toolMajorVersion: 'V2'
 
   # Run SDL tools
   - task: BinSkim@3
--- a/azure-pipelines-LGTM.yml
+++ b/azure-pipelines-LGTM.yml
@@ -40,10 +40,10 @@ jobs:
       version: $(dotnetVersion)
 
   - task: UseDotNet@2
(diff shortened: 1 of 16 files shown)


azure-pipelines-LGTM.yml
```diff
@@ -0,0 +1,72 @@
...
+      packageType: sdk
+      version: $(dotnetVersion)
+  - task: DotNetCoreCLI@2
+    inputs:
+      command: 'restore'
+
+  - task: Semmle@0
+    displayName: Security - LGTM
+    env: 
+      SYSTEM_ACCESSTOKEN: $(MICROSOFT_PAT_TOKEN)
```
yufeih, line 43: Is there a way to get rid of this token?

yufeih: How about `$(System.AccessToken)`

azure-pipelines-LGTM.yml
```diff
@@ -0,0 +1,107 @@
...
+  value: Owner=OPSBuild
+- name: LGTM.UploadSnapshot
+  value: true
+- name: Semmle.SkipAnalysis
+  value: false
+
+jobs:
+
+# Test and deploy on windows
+- job: OnboardLGTM
```
yufeih, line 29: nit:
```suggestion
- job: Security Checks
```

azure-pipelines-LGTM.yml
TianqiZhang: In communication with security team, Denis Levin wrote:
> LGTM Proxy does support GitHub, but since the source code repository is at GitHub and not in ADO, and ADO Pipeline’s SYSTEM_ACCESSTOKEN we receive in LGTM proxy as a parameter is created at ADO, we can’t use it to authenticate against GitHub. Instead (currently) we use this token to check if it has read access to https://dev.azure.com/microsoft account. 
> At the moment, the only fix we can offer is to set up your pipeline with another token (PAT token) scoped to https://dev.azure.com/microsoft that have read access (code_read) to dev.azure.com/Microsoft and pass this PAT token to pipeline:
SYSTEM_ACCESSTOKEN: $(MICROSOFT_PAT_TOKEN)

> We have an alternative authentication method in test environment that will take care of this issue, but it’s a few days away from deployment to production.

So we should be able to get rid of this token in a few days.

yufeih (approved):

herohua:
Is there a particular reason why we'd like to have this as a separate pipeline (yml)?

yufeih:
> Is there a particular reason why we'd like to have this as a separate pipeline (yml)?

It took very long to run LGTM (20+ minutes), a dedicated pipeline allows us to run these as an optional check that does not block normal workflow. (It is configured as per commit, but I see no problem running LGTM once every day).

PR #7164, opened 2021-03-10
Add document_url to error log

yufeih:
[AB#388914](https://dev.azure.com/ceapex/Engineering/_workitems/edit/388914/)

With this change, vscode extension shows error code as a standard clickable link, so we could remove our custom code action provider:

doc-url

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Gets moniker range group configuration for v2 backward compatibility.
         /// </summary>
-        public Dictionary<string, GroupConfig> Groups { get; } = new Dictionary<string, GroupConfig>();
+        public Dictionary<string, GroupConfig> Groups { get; } = new();
 
         /// <summary>
         /// Gets output file type
@@ -112,7 +112,7 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Gets the global metadata added to each document.
         /// </summary>
-        public GlobalMetadata GlobalMetadata { get; init; } = new GlobalMetadata();
+        public GlobalMetadata GlobalMetadata { get; init; } = new();
 
         /// <summary>
         /// Gets the {Schema}://{HostName}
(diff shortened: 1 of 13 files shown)


src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -234,111 +161,94 @@ private async Task<string> FetchValidationRules(string routePath, string reposit
             }
         }
 
-        private async Task<string> FetchTaxonomies(DocsEnvironment? environment = null)
+        private async Task<string> Fetch(string url, Dictionary<string, string>? headers = null, string? value404 = null)
```
yufeih, line 164: I refactored this file a little bit to ensure all GET APIs ended up into this `Fetch` method.

src/docfx/serve/LanguageServerBuilder.cs
```diff
@@ -157,6 +157,9 @@ private static Diagnostic ConvertToDiagnostics(Error error, SourceInfo source)
                      new(ConvertLocation(source.Line), ConvertLocation(source.Column)),
                      new(ConvertLocation(source.EndLine), ConvertLocation(source.EndColumn))),
                 Code = error.Code,
+                CodeDescription = error.DocumentUrl != null && Uri.TryCreate(error.DocumentUrl, UriKind.Absolute, out var href)
```
yufeih, line 160: This is a standard way of showing document urls in language service protocol.

src/docfx/config/Config.cs
```diff
@@ -160,24 +159,29 @@ internal class Config : PreloadConfig
...
         /// <summary>
         /// Gets allow custom error code, severity and message.
         /// </summary>
-        public Dictionary<string, SourceInfo<CustomRule>> Rules { get; } = new Dictionary<string, SourceInfo<CustomRule>>();
+        public Dictionary<string, SourceInfo<CustomRule>> Rules { get; } = new();
+
+        /// <summary>
+        /// Gets a map from error code to document URL.
+        /// </summary>
+        public Dictionary<string, string> DocumentUrls { get; } = new();
```
mingwli, line 184: >DocumentUrls [](start = 42, length = 12)

Do we want to limit this to lsp only? It's not useful inside .errors.log file.

928PJY (approved):

PR #7165, opened 2021-03-10
remove unused properties

yunair:


src/Microsoft.Docs.LearnValidation/Models/ValidatorHierarchyItem.cs
```diff
@@ -29,5 +31,29 @@ public class ValidatorHierarchyItem : HierarchyItem, IValidateModel
 
         [JsonProperty("page_kind")]
         public string PageKind { get; set; } = "";
+
+        [JsonProperty("depotName")]
```
yufeih, line 35: We typically don't put a `JsonProperty` for _camelCase_ properties since that's the default behavior in this project. We only add `JsonProperty` that does not adhere to this convention. 


~~[JsonProperty("depotName")]~~

src/Microsoft.Docs.LearnValidation/Models/ValidatorHierarchyItem.cs
```diff
@@ -17,6 +16,9 @@ public class ValidatorHierarchyItem : HierarchyItem, IValidateModel
 
         public bool IsDeleted { get; set; }
 
+        [JsonProperty("uid")]
+        public string UId { get; set; } = "";
+
         public string Uid => UId;
```
yufeih, line 22: If `UId` is marked as `[JsonProperty("uid")]`, is it equivalent to:

```suggestion
        public string Uid { get; set; } = "";
```

yufeih (commented):
Can we remove the `Microsoft.DocsLearning.Hierarchy.DataContract` package reference after this change?

src/Microsoft.Docs.LearnValidation/Models/ValidatorHierarchyItem.cs
yunair: we can't do this,because `IValidateModel` hava a `Uid` with `get` only

yufeih (approved):

yunair:
> Can we remove the `Microsoft.DocsLearning.Hierarchy.DataContract` package reference after this change?

we can't remove this package reference now.
there also have many dependences.

PR #7170, opened 2021-03-11
Track external link tag name and attribute name

yufeih:
We can only get if a link is _href_, _imageSrc_, or _src_ today, this change allows the telemetry to have more info about the HTML tag and attribute, like is it a `<video>` or `<iframe>`.


src/docfx/lib/markdown/MarkdownEngine.cs
```diff
@@ -349,15 +354,7 @@ private void ValidateLink(LinkInfo link)
...
-                {
-                    ImageLinkType = Enum.TryParse(link.ImageType, true, out ImageLinkType type) ? type : ImageLinkType.Default,
-                    AltText = link.AltText,
-                    IsInline = link.MarkdownObject.IsInlineImage(link.HtmlSourceIndex),
-                },
-                _ => null,
-            };
+                };
 
             if (node is null)
```
mingwli, line 359: node won't be null

mingwli (approved):

PR #7183, opened 2021-03-12
Fix join toc config loader to get correct relative path

AngryBerryMS:
[AB#390472](https://dev.azure.com/ceapex/Engineering/_workitems/edit/390472)


src/docfx/config/ops/OpsConfigLoader.cs
```diff
@@ -105,7 +105,9 @@ internal static class OpsConfigLoader
 
             var joinTOCPluginConfig = docsetConfig?.JoinTOCPlugin ?? opsConfig.JoinTOCPlugin ?? Array.Empty<OpsJoinTocConfig>();
             (result["fileMetadata"], result["joinTOC"]) =
-                GenerateJoinTocMetadataAndConfig(joinTOCPluginConfig, new PathString(buildSourceFolder));
+                GenerateJoinTocMetadataAndConfig(
```
live1206, line 108: suggest to prepare `joinTOCPluginConfig` in advance, and simplify the logic of `GenerateJoinTocMetadataAndConfig`

AngryBerryMS: fixed.

live1206 (approved):

PR #7194, opened 2021-03-15
CI uses its own key-vault

mingwli:
[AB#379834](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/379834)

### DIFF
--- a/azure-pipelines-v3.yml
+++ b/azure-pipelines-v3.yml
@@ -9,6 +9,7 @@ pr:
 
 variables:
 - group: devresourcekeyvault
+- group: DocfxCIKVPub
 - name: dotnetVersion
   value: 5.0.200
 - name: runCodesignValidationInjection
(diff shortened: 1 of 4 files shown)


azure-pipelines-v3.yml
```diff
@@ -9,6 +9,7 @@ pr:
 
 variables:
 - group: devresourcekeyvault
+- group: DocfxCIKVPub
 - name: dotnetVersion
```
herohua, line 13: We should use another KV in a specific tenant. Please check my comment under the feature.

herohua (changes requested):
We should use another KV in a specific tenant. Please check my comment under the feature.

azure-pipelines-v3.yml
mingwli: @herohua KV moved to PME

mingwli: @herohua KV moved to PME

herohua (approved):

PR #7196, opened 2021-03-16
Generate .xrefmap.json for localized repo

live1206:
[AB#386397](https://dev.azure.com/ceapex/Engineering/_workitems/edit/386397)

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -130,7 +130,7 @@ namespace Microsoft.Docs.Build
             return (error, xrefSpec, href);
         }
 
-        public XrefMapModel ToXrefMapModel(bool isLocalizedBuild)
+        public XrefMapModel ToXrefMapModel()
         {
             var repositoryBranch = _repository?.Branch;
             var basePath = _config.BasePath.ValueWithLeadingSlash;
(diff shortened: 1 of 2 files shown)


src/docfx/build/xref/XrefResolver.cs
```diff
@@ -130,35 +130,32 @@ internal class XrefResolver
...
+                })
+                .OrderBy(xref => xref.Uid)
+                .ToArray();
 
-                externalXrefs = _jsonSchemaTransformer().GetValidateExternalXrefs();
-            }
+            externalXrefs = _jsonSchemaTransformer().GetValidateExternalXrefs();
 
             var model =
                 new XrefMapModel { References = references, ExternalXrefs = externalXrefs, RepositoryUrl = _repository?.Url, DocsetName = _config.Name.Value };
```
yufeih, line 161: Is there a way to identify the locale of an xrefmap?

live1206: We should be able to have the info of localization during build.
But what is it used for?

yufeih: It's not useful for now, we can add later if needed.

yufeih (approved):

PR #7199, opened 2021-03-16
Support glob pattern in buildFile parameter

928PJY:
[AB#392994](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/392994)


src/docfx/build/DocsetBuilder.cs
```diff
@@ -164,9 +166,38 @@ public void Build(string[]? files)
                 var tocBuilder = new TocBuilder(_config, _tocLoader, _contentValidator, _metadataProvider, _metadataValidator, _documentProvider, _monikerProvider, publishModelBuilder, _templateEngine, output);
                 var redirectionBuilder = new RedirectionBuilder(publishModelBuilder, _redirectionProvider, _documentProvider);
 
-                var filesToBuild = files != null
-                    ? files.Select(file => FilePath.Content(new PathString(file))).Where(file => _input.Exists(file) && _buildScope.Contains(file.Path)).ToHashSet()
-                    : _publishUrlMap.GetFiles().Concat(_tocMap.GetFiles()).ToHashSet();
+                HashSet<FilePath> filesToBuild = new();
+                if (files == null)
```
yufeih, line 170: Could you also change `Docfx.cs` to remove `--files`? The command line argument should be `--file a --file b`. May also need to rename `CommandLineOptions.Files` -> `CommandLineOptions.File`. cc @mingwli

928PJY: but we used it as an array, so it maybe confused if we rename from `Files` -> `File` /cc @mingwli

mingwli: both `--files` and `--file` can be used to map to Files. https://github.com/dotnet/docfx/blob/ci-kv/src/docfx/cli/Docfx.cs#L92-L93.
But I'm not sure if we remove `--files`, and leaving only `--file`, the command line option can still be mapped to [Files](https://github.com/dotnet/docfx/blob/ci-kv/src/docfx/cli/CommandLineOptions.cs#L42) property.

yufeih (approved):

PR #7205, opened 2021-03-16
fix publish url conflict though whole repo

AngryBerryMS:
[AB#393039](https://dev.azure.com/ceapex/Engineering/_workitems/edit/393039)


yufeih (changes requested):
The `GetAllDocuments` method isn't guaranteed to return all documents. This solution seems complicated, let's figure out a better strategy for url conflict.

PR #7227, opened 2021-03-24
Fix glob on the root repository level in partial build

928PJY:


src/docfx/build/DocsetBuilder.cs
```diff
@@ -236,7 +236,7 @@ private HashSet<FilePath> GetFilesToBuild(string[]? files)
                 {
                     if (GlobUtility.IsGlobString(file))
                     {
-                        globs.Add(GlobUtility.CreateGlobMatcher(Path.Combine(_buildOptions.DocsetPath, file)));
+                        globs.Add(GlobUtility.CreateGlobMatcher(PathUtility.Normalize(Path.Combine(_buildOptions.DocsetPath, file))));
```
yufeih, line 239: It is better to normalize in `CreateGlobMatcher` method? Looking at the code, there are many reference to `CreateGlobMatcher` method, it's hard to tell if the input is a normalized path.

We typically use `PathString` over `string` to indicate that the parameter is a normalized path string. Maybe `CreateGlobMatcher` should take a `PathString` instead, but that seems like a bigger change we could defer later.

yufeih (approved):
👍

PR #7230, opened 2021-03-24
Add more metadata to BuildLog Telemetry

fkpwolf:
[AB#397541](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/397541)

### DIFF
--- a/src/docfx/lib/log/Error.cs
+++ b/src/docfx/lib/log/Error.cs
@@ -25,15 +25,7 @@ namespace Microsoft.Docs.Build
 
         public object?[] MessageArguments { get; init; } = Array.Empty<object?>();
 
-        public string? MsAuthor { get; init; }
-
-        public string? MsProd { get; init; }
-
-        public string? MsTechnology { get; init; }
-
-        public string? MsService { get; init; }
-
-        public string? MsSubservice { get; init; }
+        public AdditionalErrorInfo? AdditonalErrorInfo { get; init; }
 
         public Error(ErrorLevel level, string code, FormattableString message, SourceInfo? source = null, string? propertyPath = null)
         {
@@ -67,11 +59,11 @@ namespace Microsoft.Docs.Build
                 log_item_type = "user",
                 pull_request_only = PullRequestOnly ? (bool?)true : null,
                 property_path = PropertyPath,
-                ms_author = MsAuthor,
-                ms_prod = MsProd,
-                ms_technology = MsTechnology,
-                ms_service = MsService,
-                ms_subservice = MsSubservice,
+                ms_author = AdditonalErrorInfo?.MsAuthor,
+                ms_prod = AdditonalErrorInfo?.MsProd,
+                ms_technology = AdditonalErrorInfo?.MsTechnology,
+                ms_service = AdditonalErrorInfo?.MsService,
+                ms_subservice = AdditonalErrorInfo?.MsSubservice,
                 document_url = DocumentUrl,
                 date_time = DateTime.UtcNow, // Leave data_time as the last field to make regression test stable
             }).Replace("\"ms_", "\"ms.");
--- a/src/docfx/lib/log/Error.cs
+++ b/src/docfx/lib/log/Error.cs
@@ -13,8 +13,6 @@ namespace Microsoft.Docs.Build
 
         public string Message { get; init; }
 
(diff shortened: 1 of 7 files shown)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -151,12 +151,24 @@ public static void TrackExternalLink(string tag, string attribute, string scheme
...
+                error.MsTechnology,
+                error.MsService,
+                error.MsSubservice,
+            };
+            var additoinInfoString = JsonUtility.Serialize(additionalInfo);
             if (!s_isRealTimeBuild.Value)
             {
                 s_errorCountMetric.TrackValue(
-                    1, code, level.ToString(), CoalesceEmpty(name), "User", s_os, s_version, s_repo, s_branch, s_correlationId, s_sessionId);
+                    1, code, level.ToString(), CoalesceEmpty(name), additoinInfoString, s_os, s_version, s_repo, s_branch, s_correlationId, s_sessionId);
```
hejingkan2005: Is it okay to remove type = "User" here?

hejingkan2005: Or we can include 'User' into above additionalInfo object? If yes, I think we need to keep its name as it is.

hejingkan2005 (approved):

src/docfx/lib/log/Telemetry.cs
hejingkan2005: Offline talked to @fkpwolf , he has confirmed with @yufeih that 'User' is hard-coded, and not used. So we can discard it.

src/docfx/lib/log/ErrorLog.cs
```diff
@@ -40,10 +40,13 @@ public override void Add(Error error)
...
-                    if (msAuthor != default)
-                    {
-                        error = error with { MsAuthor = msAuthor };
-                    }
+                    var msProd = MetadataProvider?.GetMetadata(Null, source).MsProd;
+                    var msTechnology = MetadataProvider?.GetMetadata(Null, source).MsTechnology;
+                    var msService = MetadataProvider?.GetMetadata(Null, source).MsService;
+                    var msSubservice = MetadataProvider?.GetMetadata(Null, source).MsSubservice;
+
+                    error = error with
```
herohua, line 48: would default overwrite existing value?

src/docfx/lib/log/ErrorLog.cs
```diff
@@ -40,10 +40,13 @@ public override void Add(Error error)
                 try
                 {
                     var msAuthor = MetadataProvider?.GetMetadata(Null, source).MsAuthor;
-                    if (msAuthor != default)
-                    {
-                        error = error with { MsAuthor = msAuthor };
-                    }
+                    var msProd = MetadataProvider?.GetMetadata(Null, source).MsProd;
```
yufeih, line 43: call `GetMetadata` once.

src/docfx/lib/log/Error.cs
```diff
@@ -27,6 +25,16 @@ internal record Error
 
         public object?[] MessageArguments { get; init; } = Array.Empty<object?>();
 
+        public string? MsAuthor { get; init; }
```
yufeih, line 28: Consider extract a class `AdditonalErrorInfo` and put these `Ms.xyz` into that class.

src/docfx/lib/log/Telemetry.cs
```diff
@@ -151,12 +151,24 @@ public static void TrackExternalLink(string tag, string attribute, string scheme
...
+            var name = error.PropertyPath;
+            var additionalInfo = new
+            {
+                error.MsAuthor,
+                error.MsProd,
+                error.MsTechnology,
+                error.MsService,
+                error.MsSubservice,
+            };
+            var additionalInfoString = JsonUtility.Serialize(additionalInfo);
```
yufeih, line 167: Consider use `ms.author` or `ms_author` as key in the output JSON. As a guideline, try to use the same name and case for the same term, if `.` is not allowed, replace it with `_`.

yufeih (approved):

fkpwolf:
the diff are excepted: now log have more properties like `ms.prod`.  So this PR is ready for merge @yufeih

live1206 (approved):

PR #7232, opened 2021-03-25
Do not generate output files when fail on ecma2yaml or maml2yaml

live1206:
[AB#398267](https://ceapex.visualstudio.com/Engineering/_workitems/edit/398267)

### DIFF
--- a/src/docfx/config/ops/OpsPreProcessor.cs
+++ b/src/docfx/config/ops/OpsPreProcessor.cs
@@ -32,9 +32,7 @@ namespace Microsoft.Docs.Build
 
         public bool Run()
         {
-            // TODO: get failure indicator from MAML2Yaml
-            PreProcessMAML();
-            return PreProcessMonoDocXml();
+            return PreProcessMonoDocXml() & PreProcessMAML();
         }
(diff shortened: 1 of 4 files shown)


src/docfx/config/ops/OpsPreProcessor.cs
```diff
@@ -67,7 +69,7 @@ private void PreProcessMonoDocXml()
                             ? null
                             : Path.GetFullPath(Path.Combine(_buildOptions.DocsetPath, ".fallback", monodocConfig.OutputYamlFolder));
                         var (repository, _) = _repositoryProvider.GetRepository(new PathString(xmlDirectory));
-                        ECMA2YamlConverter.Run(
+                        result &= ECMA2YamlConverter.Run(
```
yufeih, line 72: return false on first failure? There doesn't seem to be value in processing more entries in case of failure.

live1206: The failure is most likely caused by user content, so I think it would be worth to report all content errors.

yufeih (approved):

PR #7233, opened 2021-03-25
Add docs trusted domain list

yufeih:
[AB#389089](https://dev.azure.com/ceapex/Engineering/_workitems/edit/389089/)
[AB#389160](https://dev.azure.com/ceapex/Engineering/_workitems/edit/389160/)


src/docfx/data/docs/trusted-domains.json
```diff
@@ -0,0 +1,191 @@
+{
+  "a": [
```
herohua, line 2: can we communicate with Ke and Megan on this list before it goes to live?

yufeih: Yes, they are already asking for this list in feature comments.

easonyq (approved):

PR #7236, opened 2021-03-25
Remove default branch for regression test, use repo default branch

live1206:


test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -194,10 +198,22 @@ private static void EnsureTestData(Options opts, string workingFolder)
             Exec("git", $"clean -xdff", cwd: workingFolder);
 
             var submoduleUpdateFlags = s_isPullRequest ? "" : "--remote";
-            Exec("git", $"{s_gitCmdAuth} submodule set-branch -b {opts.Branch} {s_repositoryName}", cwd: workingFolder, secrets: s_gitCmdAuth);
+            Exec("git", $"{s_gitCmdAuth} submodule set-branch -b {remoteBranch} {s_repositoryName}", cwd: workingFolder, secrets: s_gitCmdAuth);
```
live1206, line 201: For the first-time run of regression test, `submodule set-branch` will crash.
This was found during the 1ES test.

mingwli (approved):

test/docfx.RegressionTest/Options.cs
```diff
@@ -11,7 +11,7 @@ internal class Options
         public string Repository { get; set; } = "";
 
         [Option("branch")]
-        public string Branch { get; set; } = "master";
+        public string Branch { get; set; } = "";
```
yufeih, line 14: nit: prefer `null` for optional config.
```suggestion
        public string? Branch { get; set; }
```

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -194,10 +198,22 @@ private static void EnsureTestData(Options opts, string workingFolder)
...
+
+        private static string GetRemoteDefaultBranch(string repositoryUrl, string workingDirectory)
+        {
+            var remoteInfo = ProcessUtility.Execute("git", $"{s_gitCmdAuth} remote show {repositoryUrl}", workingDirectory, secret: s_gitCmdAuth);
+            var match = Regex.Match(remoteInfo, "^([\\s\\S]*)\\sHEAD branch: (.*)$");
+            if (match.Success)
+            {
+                return match.Groups[2].Value;
+            }
+            return string.Empty;
```
yufeih, line 216: throw on failure

yufeih (approved):

test/docfx.RegressionTest/Options.cs
live1206: Thought we prefer empty string instead of nullable.
https://github.com/dotnet/docfx/blob/8f581275384d31b250e44c14e1a2c98e3e280325/src/docfx/config/Config.cs#L285

PR #7257, opened 2021-03-31
1ES migration

AngryBerryMS:
[AB#398355](https://dev.azure.com/ceapex/Engineering/_workitems/edit/398355)


test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -171,10 +211,17 @@ private static string EnsureTestData(Options opts, string workingFolder)
...
-                Exec("git", $"remote add origin {TestDataRepositoryUrl}", cwd: workingFolder);
-                Exec("git", $"{s_gitCmdAuth} fetch origin --progress template", cwd: workingFolder, secrets: s_gitCmdAuth);
+                try
+                {
+                    Directory.CreateDirectory(workingFolder);
+                    Exec("git", $"init", cwd: workingFolder);
+                    Exec("git", $"remote add origin {TestDataRepositoryUrl}", cwd: workingFolder);
+                    Exec("git", $"{s_gitCmdAuth} fetch origin --progress template", cwd: workingFolder, secrets: s_gitCmdAuth);
+                }
+                catch
```
yufeih, line 221: When will this throw? This could easily leads to infinite loop.

azure-pipelines-v3.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 # Test and deploy on windows
 - job: WindowsBuild
   pool:
-    vmImage: 'windows-latest'
+    name: 1ES-Pool
```
yufeih, line 24: Is this a stateless pool?

1ES-Provision.yml
```diff
@@ -0,0 +1,127 @@
+trigger:
+  batch: true
```
mingwli, line 2: none

1ES-Provision.yml
```diff
@@ -0,0 +1,127 @@
+trigger:
```
yufeih, line 1: This file seem to do the same thing as `azure-pipeline-v3.yml`, except for some trigger and parameter differences, could it be merged as one?

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -35,13 +35,53 @@ internal static class RegressionTest
 
         private static int Main(string[] args)
         {
-            return Parser.Default.ParseArguments<Options>(args).MapResult(
-                Run,
-                _ =>
+            if (args.Length >= 1 && args[0].Equals("warm-up"))
```
yufeih, line 38: We could have a `--warm-up` command line option, and call this executable many times in pipeline definition.

azure-pipelines-v3.yml
AngryBerryMS: Yes.

1ES-Provision.yml
AngryBerryMS: Actually it's very different from `azure-pipeline-v3.yml` as we should promise all repos be downloaded to each agent.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Sometimes the url `https://dev.azure.com/ceapex/Engineering/_git/docfx.TestData` return a 503 code.

1ES-Provision.yml
AngryBerryMS: Done.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: After exploring, I think the current solution is better to download all repos to every agents. A `--warm-up` command option seems to make the pipeline complex.

1ES-Provision.yml
mingwli: I meant this: [https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#push-trigger](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#push-trigger) `trigger: none` 

---
In reply to: [605334656](https://github.com/dotnet/docfx/pull/7257#discussion_r605334656) [](ancestors = 605334656)

test/docfx.RegressionTest/RegressionTest.cs
yufeih: Consider catch and retry the `fetch` command for a finite time (like 3), delay a second or so between each retry. The current approach may leads to infinite loop for non-retriable exception.

1ES-Provision.yml
yufeih: The bottom line is to share the same list of repos between these two jobs.

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -35,13 +35,53 @@ internal static class RegressionTest
...
+                var workingFolder = Path.Combine(s_testDataRoot, $"regression-test.{s_repositoryName}");
+
+                Console.WriteLine($"Downloading {s_repository} with branch {opts.Branch}");
+                EnsureTestData(opts, workingFolder);
+                Console.WriteLine($"{s_repository} with branch {opts.Branch} is finished!");
+            }
+            catch
+            {
+                throw;
+            }
```
mingwli, line 82: Can be removed, or swallow the exception to continue the WarmUp

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -35,13 +35,53 @@ internal static class RegressionTest
...
+        private static int WarmUpAgents(Options opts)
+        {
+            try
+            {
+                s_repository = opts.Repository;
+                s_repositoryName = $"{(opts.DryRun ? "dryrun." : "")}{Path.GetFileName(opts.Repository)}";
+                var workingFolder = Path.Combine(s_testDataRoot, $"regression-test.{s_repositoryName}");
+
+                Console.WriteLine($"Downloading {s_repository} with branch {opts.Branch}");
+                EnsureTestData(opts, workingFolder);
```
mingwli, line 76: Can be moved inside EnsureTestData to share with Run()

1ES-Provision.yml
```diff
@@ -0,0 +1,126 @@
...
+    https://github.com/MicrosoftDocs/test --branch master --timeout 20 --no-dry-sync
+    https://github.com/MicrosoftDocs/DocsRoot --branch master --timeout 20
+    https://dev.azure.com/cpubwin/windows-compatibility/_git/windows-compatibility --branch main --timeout 20
+    https://github.com/MicrosoftDocs/azure-docs-cli --branch master --timeout 50
+    https://github.com/MicrosoftDocs/dataprep-dotnet-pr --branch master --timeout 30
+    https://github.com/MicrosoftDocs/azure-mediaplayer-typescript --branch master --timeout 30
+    https://github.com/MicrosoftDocs/quantum-docs-private --timeout 30 --branch main
+    https://github.com/microsoftgraph/microsoft-graph-docs --branch main --timeout 30
+    https://github.com/MicrosoftDocs/bot-docs-pr --branch master --timeout 40 --no-dry-sync --dry-run --public-template
+    https://github.com/MicrosoftDocs/azure-docs-pr --branch master --profile --timeout 50 --dry-run --regression-rules
```
mingwli, line 45: Can the parameters irrelevant to WarmUp be removed?

1ES-Provision.yml
AngryBerryMS: Seems difficult to share. Currently azure-pipeline-v3.yml contains no branch info but for warm-up we need branch or the clone will fail.

1ES-Provision.yml
AngryBerryMS: ok, will remove

1ES-Provision.yml
AngryBerryMS: I found that `parameters` for azure pipeline may be useful for this case. Testing...

AngryBerryMS: Done

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: swallowed.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Fixed.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Prefer to have it's own method to make it clear.

azure-pipelines-v3.yml
```diff
@@ -72,99 +143,74 @@ jobs:
...
+    - task: PowerShell@2
+      displayName: create needed directory
+      inputs:
+        targetType: 'inline'
+        powerShellErrorActionPreference: continue
+        script: |
+          if (-not (Test-Path -Path D:\docfx.TestData\.temp)) {
+            New-Item -Path 'D:\docfx.TestData\.temp\' -ItemType Directory
+          } else {
+            echo 'The directory is already exist.'
```
live1206, line 204:
```suggestion
            echo 'The directory already exists.'
```

azure-pipelines-v3.yml
```diff
@@ -72,99 +143,74 @@ jobs:
...
+  dependsOn: WindowsBuild
+  pool:
+    name: Docfx-1ES
+  strategy:
+    matrix:
+      ${{ each agent in parameters.agents }}:
+        ${{ agent.key }}:
+          agentIndex: ${{ agent.value }}
+  steps:
+    - template: prepare.yml
```
live1206, line 191: call it `warmup.yml` to be consistent

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -164,19 +210,35 @@ private static bool Test(Options opts, string workingFolder, string remoteBranch
...
+                try
+                {
+                    Directory.CreateDirectory(workingFolder);
+                    Exec("git", $"init", cwd: workingFolder);
+                    Exec("git", $"remote add origin {TestDataRepositoryUrl}", cwd: workingFolder);
+                    Exec("git", $"{s_gitCmdAuth} fetch origin --progress template", cwd: workingFolder, secrets: s_gitCmdAuth);
+                }
+                catch
+                {
+                    if (retryCount++ < 3)
```
live1206, line 226: How about setting `retryCount` as `3` and `if (retryCount-- == 0)`?
And also, can we use polly here to deal with specific exception?

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -35,13 +35,59 @@ internal static class RegressionTest
...
         private static int Main(string[] args)
         {
-            return Parser.Default.ParseArguments<Options>(args).MapResult(
-                Run,
-                _ =>
+            if (args.Length >= 1 && args[0].Equals("warm-up"))
+            {
+                Console.WriteLine($"warm up starting...");
+
+                var opt = string.Join(" ", args[1..]);
```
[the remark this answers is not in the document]
AngryBerryMS: For warm-up, the argument "warm-up" is the must. Thus only 1 argument is invalid.

azure-pipelines-v3.yml
AngryBerryMS: Fixed.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: fixed

live1206 (approved):

azure-pipelines-v3.yml
```diff
@@ -72,99 +143,74 @@ jobs:
...
+      parameters:
+        dotnetVersion: $(dotnetVersion)
+    # Warm up
+    - task: PowerShell@2
+      displayName: create needed directory
```
yufeih, line 196: Can this be done in `regressiontest.exe`?

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -35,13 +35,52 @@ internal static class RegressionTest
 
         private static int Main(string[] args)
         {
-            return Parser.Default.ParseArguments<Options>(args).MapResult(
-                Run,
-                _ =>
+            if (args.Length > 1 && args[0].Equals("warm-up"))
```
yufeih, line 38: Add `WarnUp` to `Options`, don't manually parse command line args.

azure-pipelines-v3.yml
```diff
@@ -72,99 +143,74 @@ jobs:
...
-    displayName: 'Install .NET Core sdk $(dotnetVersion)'
-    inputs:
-      packageType: sdk
-      version: $(dotnetVersion)
-
-  # Download RegressionTest binary
-  - download: current
-    displayName: Download RegressionTest binary
-    artifact: RegressionTest
+    - template: prepare.yml
```
yufeih, line 157: There doesn't seem to be much value in extracting this `prepare.yml`. The only difference between the two are the _Run Regression Test_ vs _Warmup Regression Test_ steps

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -35,13 +35,52 @@ internal static class RegressionTest
...
+                    Run,
+                    _ =>
+                    {
+                        SendPullRequestComments(new() { CrashMessage = "regression-test argument exception" });
+                        return -9999;
+                    });
+            }
+        }
+
+        private static int WarmUpAgents(Options opts)
```
yufeih, line 73: Remove this method (this has many duplicated code with the `Run` method), instead, return after `EnsureTestData` if `opt.WarmUp` is true.

yufeih (changes requested):

azure-pipelines-v3.yml
```diff
@@ -1,3 +1,72 @@
+parameters:
```
yufeih, line 1: move this section to regression text job variables as it is local to that job only.

AngryBerryMS: No, this is also useful for warm up job.

azure-pipelines-v3.yml
AngryBerryMS: This template is useful for both the 2 jobs. If we don't extract it, we will have duplicate segments in this pipeline.

azure-pipelines-v3.yml
AngryBerryMS: Yes, but I think do it outside of the regressontest.exe should be clear. Or we will try to create this dir multiple times.

azure-pipelines-v3.yml
```diff
@@ -34,6 +103,7 @@ jobs:
   # dotnet test
   - task: DotNetCoreCLI@2
     displayName: dotnet test
+    condition: and(succeeded(), eq(variables['pipeline-option'], 'regression-test'))
```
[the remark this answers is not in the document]
AngryBerryMS: If we treat `pipeline-option` as a bool value, I cannot find a way to specify it when run the pipeline on schedule.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Fixed.

azure-pipelines-v3.yml
AngryBerryMS: Template removed.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Done.

test/docfx.RegressionTest/Options.cs
```diff
@@ -7,6 +7,9 @@ namespace Microsoft.Docs.Build
 {
     internal class Options
     {
+        [Option("warm-up")]
+        public bool WarmUp { get; set; } = false;
```
yufeih, line 11: nit: move it after `[Value]` before other options.

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -35,11 +35,12 @@ internal static class RegressionTest
 
         private static int Main(string[] args)
         {
+            Console.WriteLine("starting...");
```
yufeih, line 38: remove this line.

azure-pipelines-v3.yml
```diff
@@ -72,99 +74,160 @@ jobs:
...
+      ${{ each agent in parameters.agents }}:
+        ${{ agent.key }}:
+          agentIndex: ${{ agent.value }}
+  steps:
+    - task: UseDotNet@2
+      displayName: 'Install .NET Core sdk $(dotnetVersion)'
+      inputs:
+        packageType: sdk
+        version: $(dotnetVersion)
+    # Download RegressionTest binary
```
yufeih, line 138: nit: new line between each step

azure-pipelines-v3.yml
```diff
@@ -72,99 +74,160 @@ jobs:
 
 # Run RegressionTest
 - job: RunRegressionTest
-  condition: and(succeeded(), ne(variables['Build.SourceBranch'], 'refs/heads/v3-release'))
+  condition: and(succeeded(), ne(variables['Build.SourceBranch'], 'refs/heads/v3-release'), ne(variables['pipeline-option'], 'warm-up'))
```
yufeih, line 77: nit: `warmup: true`

yufeih (approved):

test/docfx.RegressionTest/Options.cs
AngryBerryMS: Fixed.

azure-pipelines-v3.yml
AngryBerryMS: Done.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: removed

AngryBerryMS:
Those failed build is due to no enough disk space. I should use agent with larger disk size.

PR #7269, opened 2021-04-02
add perf env

AngryBerryMS:


src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -253,7 +254,9 @@ private static string BuildApi(DocsEnvironment? environment = null)
             return (environment ?? DocsEnvironment) switch
             {
                 DocsEnvironment.Prod => "https://buildapi.docs.microsoft.com",
-                _ => "https://BuildApiPubDev.azurefd.net",
+                DocsEnvironment.PPE => "https://BuildApiPubDev.azurefd.net",
+                DocsEnvironment.Perf => "https://op-build-test.azurewebsites.net",
+                _ => throw new NotSupportedException(),
```
live1206, line 259: throw `InvalidOperationException` like the others

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -232,6 +232,7 @@ private static HttpMiddleware BuildMiddleware(DocsEnvironment? environment = nul
                         {
                             DocsEnvironment.Prod => s_opsTokenPublic,
                             DocsEnvironment.PPE => s_opsTokenPubDev,
+                            DocsEnvironment.Perf => s_opsTokenPubDev,
```
live1206, line 235: Does this need to be a separate source environment?

AngryBerryMS: Updated.

src/docfx/config/ops/OpsAccessor.cs
AngryBerryMS: Done.

live1206 (approved):

PR #7275, opened 2021-04-07
Remove External Link Telemetry

yufeih:
[AB#406544](https://ceapex.visualstudio.com/Engineering/_workitems/edit/406544)


live1206 (approved):

herohua:
Is there a need to capture telemetries on all external links (not only the ones violating allow-list)?

yufeih:
> Is there a need to capture telemetries on all external links (not only the ones violating allow-list)?

I'm not aware of an ongoing need of this data. This data also doesn't seem useful to the graceful shrinking trusted domains list. Removing this telemetry aligns URL sanitization with HTML sanitization.

PR #7294, opened 2021-04-12
release 2.57.1

superyyrrzz:


928PJY (approved):

herohua:
Shall we add a note that we reverted "Sub navbar not fully shows if remove"? Since we cannot publish 2.57 again, so I believe adding the note in 2.57.1 would be better.

herohua:
> Shall we add a note that we reverted "Sub navbar not fully shows if remove"? Since we cannot publish 2.57 again, so I believe adding the note in 2.57.1 would be better.

Confirmed with @superyyrrzz that the release note itself needs to be fixed.

herohua (approved):

superyyrrzz:
> Shall we add a note that we reverted "Sub navbar not fully shows if remove"? Since we cannot publish 2.57 again, so I believe adding the note in 2.57.1 would be better.

That commit has been reverted in 2.57, but I missed to remove it from 2.57's release note...

PR #7299, opened 2021-04-13
add retry logic for regression tests

AngryBerryMS:
[AB#407986](https://dev.azure.com/ceapex/Engineering/_workitems/edit/407986)

### DIFF
--- a/test/docfx.RegressionTest/RegressionTest.cs
+++ b/test/docfx.RegressionTest/RegressionTest.cs
@@ -14,6 +14,7 @@ using System.Threading.Tasks;
 using CommandLine;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using Polly;
 
 namespace Microsoft.Docs.Build
 {
@@ -174,6 +175,7 @@ namespace Microsoft.Docs.Build
 
         private static string EnsureTestData(Options opts, string workingFolder, int retryCount = 3)
         {
+            var retryPolicy = Retry();
             if (!Directory.Exists(workingFolder))
             {
                 try
@@ -181,19 +183,15 @@ namespace Microsoft.Docs.Build
                     Directory.CreateDirectory(workingFolder);
                     Exec("git", $"init", cwd: workingFolder);
                     Exec("git", $"remote add origin {TestDataRepositoryUrl}", cwd: workingFolder);
-                    Exec("git", $"{s_gitCmdAuth} fetch origin --progress template", cwd: workingFolder, secrets: s_gitCmdAuth);
+
+                    retryPolicy.Execute(() =>
+                    {
+                        Exec("git", $"{s_gitCmdAuth} fetch origin --progress template", cwd: workingFolder, secrets: s_gitCmdAuth);
+                    });
                 }
                 catch
                 {
-                    if (retryCount-- > 0)
-                    {
-                        _ = Task.Delay(5000);
-                        EnsureTestData(opts, workingFolder, retryCount);
-                    }
-                    else
-                    {
-                        throw new InvalidOperationException();
-                    }
+                    throw new InvalidOperationException();
                 }
             }
 
@@ -203,7 +201,13 @@ namespace Microsoft.Docs.Build
 
             try
             {
-                Exec("git", $"{s_gitCmdAuth} fetch origin --progress --prune {s_repositoryName}", cwd: workingFolder, secrets: s_gitCmdAuth, redirectStandardError: true);
+                retryPolicy.Execute(
+                    () => Exec(
+                            "git",
+                            $"{s_gitCmdAuth} fetch origin --progress --prune {s_repositoryName}",
+                            cwd: workingFolder,
+                            secrets: s_gitCmdAuth,
+                            redirectStandardError: true));
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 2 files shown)


test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -527,5 +534,8 @@ private static async Task SendGitHubPullRequestComments(int prNumber, string bod
 
             response.EnsureSuccessStatusCode();
         }
+
+        private static Policy Retry(int retryCount = 6)
+            => Policy.Handle<Exception>().WaitAndRetry(retryCount, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryCount)));
```
live1206, line 539: - Are we retrying on all exceptions?
- Why retry 6 times instead of 3 like previously?
- Would `Math.Pow(2, retryCount)` be waiting much longer than before?

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -181,19 +183,15 @@ private static string EnsureTestData(Options opts, string workingFolder, int ret
                     Directory.CreateDirectory(workingFolder);
                     Exec("git", $"init", cwd: workingFolder);
                     Exec("git", $"remote add origin {TestDataRepositoryUrl}", cwd: workingFolder);
-                    Exec("git", $"{s_gitCmdAuth} fetch origin --progress template", cwd: workingFolder, secrets: s_gitCmdAuth);
+
+                    s_retryPolicy.Execute(() =>
```
live1206, line 187: How about calling `Retry()` directly?

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: The maximum waiting time is about 2 mins for each retry, which is about 2 times compared with before.

AngryBerryMS: Since the retry is not that common encountered actually, I think retry more times can make a strong promise without reducing the performance a lot....

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Does it can make the code more clear?

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -181,19 +183,15 @@ private static string EnsureTestData(Options opts, string workingFolder, int ret
...
-                    if (retryCount-- > 0)
-                    {
-                        _ = Task.Delay(5000);
-                        EnsureTestData(opts, workingFolder, retryCount);
-                    }
-                    else
-                    {
-                        throw new InvalidOperationException();
-                    }
+                    throw new InvalidOperationException();
```
yufeih, line 194: remove `try { } catch { throw }`

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Done

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -181,19 +183,15 @@ private static string EnsureTestData(Options opts, string workingFolder, int ret
                     Directory.CreateDirectory(workingFolder);
                     Exec("git", $"init", cwd: workingFolder);
                     Exec("git", $"remote add origin {TestDataRepositoryUrl}", cwd: workingFolder);
-                    Exec("git", $"{s_gitCmdAuth} fetch origin --progress template", cwd: workingFolder, secrets: s_gitCmdAuth);
+
+                    Retry().Execute(() =>
```
yufeih, line 187: It seems all git network operations needs retry, consider a `ExecGitRetry(cmd, ...)` method because the name `Retry()` does not tell that it only checks for git failures and retry.

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -527,5 +534,8 @@ private static async Task SendGitHubPullRequestComments(int prNumber, string bod
 
             response.EnsureSuccessStatusCode();
         }
+
+        private static Policy Retry(int retryCount = 5)
+            => Policy.Handle<HttpRequestException>(r => r.StatusCode == HttpStatusCode.ServiceUnavailable).WaitAndRetry(retryCount, retryAttempt => TimeSpan.FromSeconds(Math.Pow(2, retryAttempt)));
```
yufeih, line 539: When git.exe failed, it didn't raise `HttpRequestException`.

yufeih (changes requested):

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Retry 5 times, and the maximum waiting time is less than 1 min.

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Only retry when 503

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Ok

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Removed

yufeih (approved):

PR #7312, opened 2021-04-16
1ES improvement

AngryBerryMS:
After collecting data for docfx-pipeline for about 1 week, I update the pipeline.

- 8 d8ds v4 agents in Docfx-1ES pool
- use `- task: AzureCLI@2` rather than simple `- script` for the warmup of each agent

### DIFF
--- a/test/docfx.RegressionTest/RegressionTest.cs
+++ b/test/docfx.RegressionTest/RegressionTest.cs
@@ -58,7 +58,6 @@ namespace Microsoft.Docs.Build
                 if (opts.WarmUp)
                 {
                     var (_, outputPath, repositoryPath, docfxConfig) = Prepare(opts, workingFolder, remoteBranch);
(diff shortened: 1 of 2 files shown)


test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -58,6 +58,7 @@ private static int Run(Options opts)
                 if (opts.WarmUp)
                 {
                     var (_, outputPath, repositoryPath, docfxConfig) = Prepare(opts, workingFolder, remoteBranch);
+                    Clean(outputPath);
```
live1206, line 61: Is this needed? `Cleanup` will be called while running test anyway.

AngryBerryMS: In case we want to re-run the warm-up manually.

live1206 (approved):

PR #7319, opened 2021-04-20
Output schema to metadata

live1206:
- [AB#411064](https://ceapex.visualstudio.com/Engineering/_workitems/edit/411064)
- Return before applying template while output type is `json` for data type

### DIFF
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -420,7 +420,7 @@ inputs:
   _themes/ContentTemplate/TestData.html.tmpl:
 outputs:
   a.json: |
-    {"metadata": {"href": "b", "schema": "TestData"}}
+    {"metadata": {"href": "b"}, "schema": "TestData"}
   b.json: |
     {"conceptual": "", "schema": "Conceptual"}
 ---
--- a/docs/specs/schema.yml
+++ b/docs/specs/schema.yml
@@ -420,8 +420,9 @@ inputs:
   _themes/ContentTemplate/TestData.html.tmpl:
 outputs:
   a.json: |
-    {"metadata": {"href": "b"}}
-  b.json:
+    {"metadata": {"href": "b", "schema": "TestData"}}
+  b.json: |
+    {"conceptual": "", "schema": "Conceptual"}
 ---
 # Skip javascript on schema validation failure
 inputs:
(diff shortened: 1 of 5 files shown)


yufeih (approved):

AngryBerryMS (approved):

herohua (approved):

docs/specs/schema.yml
```diff
@@ -420,8 +420,9 @@ inputs:
   _themes/ContentTemplate/TestData.html.tmpl:
 outputs:
   a.json: |
-    {"metadata": {"href": "b"}}
-  b.json:
+    {"metadata": {"href": "b", "schema": "TestData"}}
```
yufeih, line 423:
```suggestion
    {"metadata": {"href": "b"}, "schema": "TestData"}
```

yufeih (changes requested):

src/docfx/build/page/PageBuilder.cs
```diff
@@ -134,6 +134,7 @@ public void Build(ErrorBuilder errors, FilePath file)
 
             var systemMetadataJObject = JsonUtility.ToJObject(systemMetadata);
 
+            outputModel["schema"] = mime.Value;
```
yufeih, line 137: Move this line to line 152 in case this key is overridden by other processes.

live1206: ok

yufeih: Also consider adding `schema` to each entry in `.publish.json`

live1206: Where is this requirement from?
Not related to the current feature.

yufeih: No, this isn't required

yufeih (approved):

src/docfx/build/page/PageBuilder.cs
live1206: It shows up in `.publish.json` already by [Extensiondata](https://github.com/dotnet/docfx/blob/a3751d30b15eea5565167bb3dfc1ede7691cf268/src/docfx/publish/PublishItem.cs#L36)

PR #7324, opened 2021-04-22
refactor metadata for template

AngryBerryMS:


src/docfx/data/docs/metadata.json
```diff
@@ -29,59 +29,6 @@
             "type": "boolean",
             "description": "Indicate whether the depot need to API Scan"
         },
-        "product": {
-            "type": ["array", "null"],
```
yufeih, line 33: only remove `htmlMetaName`. Other properties like `type` is still used for metadata validation

src/docfx/build/metadata/MetadataProvider.cs
```diff
@@ -19,27 +18,60 @@ internal class MetadataProvider
 
         private readonly ConcurrentDictionary<FilePath, Watch<(ErrorList, UserMetadata)>> _metadataCache = new();
 
-        public ICollection<string> HtmlMetaHidden { get; }
-
-        public IReadOnlyDictionary<string, string> HtmlMetaNames { get; }
-
-        public MetadataProvider(Config config, Input input, BuildScope buildScope, JsonSchemaLoader jsonSchemaLoader)
+        public ICollection<string> HtmlMetaHidden { get; } = new List<string>()
```
yufeih, line 21: Move these two properties to `HtmlUtility.cs`, they are only used in `CreateHtmlMetaTags` method.

yufeih, line 21:
```suggestion
        public IReadOnlyList<string> HtmlMetaHidden { get; } = new[]
```

AngryBerryMS: Done.

src/docfx/data/docs/metadata.json
AngryBerryMS: Oh sorry. Fixed.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -18,6 +18,54 @@ internal static class HtmlUtility
...
+            "is_dynamic_rendering",
+            "need_preview_pull_request",
+            "moniker_type",
+            "is_significant_update",
+            "serviceData",
+            "github_contributors",
+            "is_hidden",
+        };
+
+        public static readonly Dictionary<string, string> HtmlMetaNames = new()
```
live1206, line 58: nit: IReadOnlyDictionary

src/docfx/lib/HtmlUtility.cs
```diff
@@ -18,6 +18,54 @@ internal static class HtmlUtility
     {
         public delegate void TransformHtmlDelegate(ref HtmlReader reader, ref HtmlWriter writer, ref HtmlToken token);
 
+        public static readonly string[] HtmlMetaHidden = new[]
```
yufeih, line 21: This doesn't need to be public.
```suggestion
        private static readonly string[] s_htmlMetaHidden = new[]
```

src/docfx/lib/HtmlUtility.cs
AngryBerryMS: Want to be compatible  with https://github.com/dotnet/docfx/blob/2146708759fad1b02b5f6e2bac48e202efecb8e6/src/docfx/lib/HtmlUtility.cs#L39

yufeih (approved):
also remove the properties from JsonSchema.cs

PR #7332, opened 2021-04-26
Support output-type for TOC

AngryBerryMS:
* no template when output-type is "json"
* add a schema named "toc" to the output json

### DIFF
--- a/docs/specs/toc/toc.yml
+++ b/docs/specs/toc/toc.yml
@@ -1524,14 +1524,52 @@ outputs:
     }
 ---
 # TOC with output-type="json"
-inputs:
-  docfx.yml: |
-    outputType: Json
-  TOC.yml: |
-    - name: a
-      href: a.md
-  a.md:
+repos:
+  https://github.com/toc/toc-output-type-is-json:
+  - files:
+      docfx.yml: |
+        outputType: Json
+        template: https://docs.com/theme
+      TOC.yml: |
+        - name: a
+          href: a.md
+      a.md:
+  https://docs.com/theme:
+  - files:
+      ContentTemplate/toc.json.js: |
+        exports.transform = function (model) {
+          model["outputType"] = "json"
+          return {
+            content: JSON.stringify(model)
+          }
+        }
 outputs:
   a.json:
   toc.json: |
     {"items":[{"name":"a","href":"a"}],"_path":"toc.json","schema":"toc"}
+---
+# TOC with output-type="pageJson"
+repos:
+  https://github.com/toc/toc-output-type-is-pageJson:
+  - files:
+      docfx.yml: |
+        outputType: PageJson
+        template: https://docs.com/theme
+      TOC.yml: |
+        - name: a
+          href: a.md
+      a.md:
+  https://docs.com/theme:
+  - files:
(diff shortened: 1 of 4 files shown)


src/docfx/build/toc/TocBuilder.cs
```diff
@@ -85,7 +85,15 @@ public void Build(ErrorBuilder errors, FilePath file)
                     var output = _templateEngine.RunJavaScript("toc.json.js", JsonUtility.ToJObject(model));
                     _output.WriteJson(Path.ChangeExtension(outputPath, ".json"), output);
                 }
-                else
+
+                if (_config.OutputType == OutputType.Json)
+                {
+                    var outputModel = JsonUtility.ToJObject(model);
+                    outputModel["schema"] = "toc";
```
yufeih, line 92: Add a get only property to `TocModel` and init the value to `toc`, to line 79/85/98 all pick the `schema` property.

docs/specs/toc/toc.yml
```diff
@@ -1522,3 +1522,16 @@ outputs:
         "a.md": [{ "source": "TOC.yml", "type": "metadata" }]
       }
     }
+---
```
yufeih, line 1525: Also add a test case with `output-type=pagejson`

AngryBerryMS: I will change the default output-type as "pageJson" and then no need to add new spec tests for it.

docs/specs/toc/toc.yml
```diff
@@ -1522,3 +1522,16 @@ outputs:
...
+inputs:
+  docfx.yml: |
+    outputType: Json
+  TOC.yml: |
+    - name: a
+      href: a.md
+  a.md:
+outputs:
+  a.json:
+  toc.json: |
```
yufeih, line 1536: Also need a test in `static.yml` to cover `outputType=html` + `toc.json.js`

docs/specs/toc/toc.yml
```diff
@@ -1522,3 +1522,16 @@ outputs:
...
+---
+# TOC with output-type="json"
+inputs:
+  docfx.yml: |
+    outputType: Json
+  TOC.yml: |
+    - name: a
+      href: a.md
+  a.md:
+outputs:
```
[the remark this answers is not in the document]
AngryBerryMS: Done.

docs/specs/toc/toc.yml
AngryBerryMS: I add a mustache template file with a js file to generate `.html` and `.json` outputs.

src/docfx/build/toc/TocBuilder.cs
AngryBerryMS: Done.

src/docfx/build/toc/TocModel.cs
```diff
@@ -14,6 +14,8 @@ internal class TocModel
         [JsonProperty("_path")]
         public string Path { get; }
 
+        public string Schema { get; init; } = "toc";
```
yufeih, line 17:
```suggestion
        public string Schema { get; } = "toc";
```

yufeih (approved):

PR #7336, opened 2021-04-26
Extract CreateTemplateModel method into TemplateEngine

AngryBerryMS:


docs/specs/legacy.yml
```diff
@@ -70,6 +70,7 @@ outputs:
         <meta name=\"document_id\" content=\"fdb5a0d7-b68e-c465-9c5e-54a5cda94902\" />
         <meta name=\"document_version_independent_id\" content=\"92a5a770-f05f-b8bb-81c9-a8d786a9e608\" />
         <meta name=\"gitcommit\" content=\"https://github.com/legacy/test/blob/42eb7605fdef9f5e7674a1cae803f86a39ee628a/c.md\" />
+        <meta content=\"true\" name=\"isContentRenderType\"></meta>
```
yufeih, line 73: We probably don't need this in the output. We should be able to determine `renderType` based on `schema`

src/docfx/build/page/PageBuilder.cs
```diff
@@ -150,12 +150,23 @@ public void Build(ErrorBuilder errors, FilePath file)
...
 
-            var (templateModel, templateMetadata) = CreateTemplateModel(errors, file, mime, JsonUtility.SortProperties(outputModel));
+            var (templateModel, templateMetadata) = _templateEngine.CreateTemplateModel(
+                errors,
+                file,
+                mime,
+                JsonUtility.SortProperties(outputModel),
+                _buildOptions.Locale,
+                _buildOptions.Culture,
+                _config.TrustedDomains,
```
yufeih, line 166: Consider inject `_bookmarkValidator`, `_searchIndexBuilder`, `_config` as optional dependencies into `TemplateEngine` constructor.

docs/specs/legacy.yml
AngryBerryMS: Yes, after checking the code, I think we can get the rendertype by mime

src/docfx/build/page/PageBuilder.cs
AngryBerryMS: Done

src/docfx/template/TemplateEngine.cs
```diff
@@ -107,6 +122,110 @@ public void CopyAssetsToOutput(Output output, bool selfContained = true)
...
+            return CreateTemplateModel(
+                file, mime, pageModel, _buildOptions.Locale, _buildOptions.Culture, config.TrustedDomains, config.DryRun);
+        }
+
+        public string ProcessHtml(ErrorBuilder errors, Config config, FilePath file, string html)
+        {
+            return ProcessHtml(errors, file, html, config.TrustedDomains, _buildOptions.Locale, _buildOptions.Culture);
+        }
+
+        public (TemplateModel model, JObject metadata) CreateTemplateModel(
```
yufeih, line 136: We probably don't need to expose this method as public.

yufeih (approved):

yufeih (approved):

mingwli (approved):
:shipit:

PR #7339, opened 2021-04-27
Add an option named "--continue" for "build"  command to apply templates.

AngryBerryMS:
Currently I test it using a real repo and it works well for landing/hub pages.

Will try to modify the now spec test project to support this in another PR. 

Please help review firstly.

### DIFF
--- a/src/docfx/build/Builder.cs
+++ b/src/docfx/build/Builder.cs
@@ -34,17 +34,17 @@ namespace Microsoft.Docs.Build
             {
                 // Apply templates.
                 ApplyTemplates.Run(errors, options);
-                return errors.HasError;
+            }
+            else
+            {
+                var files = options.File?.Select(Path.GetFullPath).ToArray();
+
+                package ??= new LocalPackage(options.WorkingDirectory);
+
+                new Builder(options, package).Build(errors, new ConsoleProgressReporter(), files);
             }
 
-            var files = options.File?.Select(Path.GetFullPath).ToArray();
-
-            package ??= new LocalPackage(options.WorkingDirectory);
-
-            new Builder(options, package).Build(errors, new ConsoleProgressReporter(), files);
-
             operation.Complete();
-
             errors.PrintSummary();
             return errors.HasError;
         }
--- a/src/docfx/build/Builder.cs
+++ b/src/docfx/build/Builder.cs
@@ -30,6 +30,13 @@ namespace Microsoft.Docs.Build
             var operation = Telemetry.StartOperation("build");
             using var errors = new ErrorWriter(options.Log);
 
+            if (options.Continue)
+            {
+                // Apply templates.
+                ApplyTemplates.Run(errors, options);
+                return errors.HasError;
+            }
+
             var files = options.File?.Select(Path.GetFullPath).ToArray();
 
             package ??= new LocalPackage(options.WorkingDirectory);
--- a/src/docfx/build/apply-templates/ApplyTemplates.cs
+++ b/src/docfx/build/apply-templates/ApplyTemplates.cs
@@ -1,0 +1,90 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.Docs.Build
+{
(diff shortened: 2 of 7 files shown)


src/docfx/build/Builder.cs
```diff
@@ -30,6 +30,13 @@ public static bool Run(CommandLineOptions options, Package? package = null)
             var operation = Telemetry.StartOperation("build");
             using var errors = new ErrorWriter(options.Log);
 
+            if (options.Continue)
+            {
+                // Apply templates.
+                ApplyTemplates.Run(errors, options);
+                return errors.HasError;
```
yufeih, line 37: Should also call line 46/48, consider not return here.

src/docfx/template/TemplateEngine.cs
```diff
@@ -60,6 +87,20 @@ internal class TemplateEngine
             _searchIndexBuilder = searchIndexBuilder;
         }
 
+        private TemplateEngine(ErrorBuilder errors, Config config, Package package, string locale)
```
yufeih, line 90: Is it possible to have only one private constructor to remove the duplicated logic?

src/docfx/template/TemplateEngine.cs
```diff
@@ -28,19 +28,46 @@ internal class TemplateEngine
...
+            ErrorBuilder errors,
+            Config? config,
+            PackageResolver? packageResolver,
+            string locale,
+            Package? package,
+            BookmarkValidator? bookmarkValidator = null,
+            SearchIndexBuilder? searchIndexBuilder = null,
+            bool fullBuild = true)
+        {
+            if (fullBuild)
```
yufeih, line 41: Consider use two static factory methods over a bool switch. With 2 static factory methods, we could remove these null checks.

src/docfx/build/apply-templates/ApplyTemplates.cs
```diff
@@ -0,0 +1,90 @@
...
+                DryRun = false,
+                Template = new PackagePath(options.Template),
+            };
+
+            var fileResolver = new FileResolver(package);
+            var jsonSchemaLoader = new JsonSchemaLoader(fileResolver);
+            var jsonSchemaProvider = new JsonSchemaProvider(config, package, jsonSchemaLoader);
+            var templateEngine = TemplateEngine.CreateTemplateEngine(errors, config, null, locale, package, fullBuild: false);
+
+            Parallel.ForEach(
```
yufeih, line 39: Use `ParallelUtility.ForEach`, it has logic to handle exceptions.

src/docfx/build/apply-templates/ApplyTemplates.cs
```diff
@@ -0,0 +1,90 @@
...
+                File.WriteAllText(GetOutPathWithDifferentExtension(inputDir, filePath, outputDir, "json"), JsonUtility.Serialize(model));
+            }
+            else
+            {
+                var isContentRenderType = jsonSchemaProvider.GetRenderType(ContentType.Page, new SourceInfo<string?>(schema)) == RenderType.Content;
+                if (isContentRenderType)
+                {
+                    var (model, _) = templateEngine.CreateTemplateModel(file, schema is null ? string.Empty : schema.ToString(), pageModel);
+
+                    File.WriteAllText(GetOutPathWithDifferentExtension(inputDir, filePath, outputDir), JsonUtility.Serialize(model));
```
yufeih, line 73: Create output directory before calling `File.WriteAllText`

src/docfx/build/Builder.cs
AngryBerryMS: Fixed.

src/docfx/template/TemplateEngine.cs
AngryBerryMS: Done.

src/docfx/build/apply-templates/ApplyTemplates.cs
AngryBerryMS: Done

src/docfx/build/apply-templates/ApplyTemplates.cs
AngryBerryMS: Done

src/docfx/template/TemplateEngine.cs
AngryBerryMS: Done

yufeih (approved):

mingwli (approved):
:shipit:

PR #7341, opened 2021-04-28
Support continue-build in Docfx.Test

AngryBerryMS:
[AB#411062](https://ceapex.visualstudio.com/Engineering/_workitems/edit/411062)
* There are 2 learn.yml spec tests with "noContinue" flag. Because the 1st stage build won't generage "hierarchy.json" files.
* There are 2 leacy.yml spec tests with "noContinue" falg, because the "ApplyTemplates" cannot use the cache info in spec test.

### DIFF
--- a/docs/specs/learn.yml
+++ b/docs/specs/learn.yml
@@ -1,6 +1,5 @@
 ---
 # Draft run learn post validation
-noContinue: true
 noDryRun: true
 repos:
   https://github.com/learn/learn-pr:
--- a/docs/specs/learn.yml
+++ b/docs/specs/learn.yml
@@ -1,5 +1,6 @@
 ---
 # Draft run learn post validation
+noContinue: true
 noDryRun: true
 repos:
   https://github.com/learn/learn-pr:
@@ -47,6 +48,7 @@ repos:
 outputs:
 ---
 # Can generate hierarchy
+noContinue: true
 noDryRun: true
 repos:
   https://github.com/learn/learn-pr:
--- a/docs/specs/learn.yml
+++ b/docs/specs/learn.yml
@@ -48,7 +48,6 @@ repos:
 outputs:
 ---
 # Can generate hierarchy
-noContinue: true
 noDryRun: true
 repos:
   https://github.com/learn/learn-pr:
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -1,5 +1,4 @@
 # Build output as page json
-noContinue: true
 repos:
   https://github.com/legacy/test:
   - author: Charlie
@@ -426,7 +425,7 @@ inputs:
   a.md: '[](https://github.com)'
   b.yml: |
     #YamlMime:TestPage
-    link: https://docs.com
+    link: https://static.docs.com
     markdown: '[](a.md)'
   _themes/ContentTemplate/schemas/TestPage.schema.json: |
     {
@@ -446,7 +445,7 @@ outputs:
   b.raw.page.json: |
     { "content": "
         <a data-linktype='external' href='https://github.com'></a>
-        <img data-linktype='external' src='https://docs.com'></img>
+        <img data-linktype='external' src='https://static.docs.com'></img>
         <p><a data-linktype='relative-path' href='a'></a></p>" }
 ---
 # The site base path is empty.
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -418,7 +418,6 @@ outputs:
   b.raw.page.json: '{"content": "<div></div>"}'
 ---
 # Add data-linktype to urls
-noContinue: true
 inputs:
   docfx.yml: |
     outputType: pageJson
--- a/docs/specs/legacy.yml
+++ b/docs/specs/legacy.yml
@@ -1,4 +1,5 @@
 # Build output as page json
+noContinue: true
 repos:
   https://github.com/legacy/test:
   - author: Charlie
@@ -418,6 +419,7 @@ outputs:
   b.raw.page.json: '{"content": "<div></div>"}'
 ---
 # Add data-linktype to urls
+noContinue: true
 inputs:
   docfx.yml: |
     outputType: pageJson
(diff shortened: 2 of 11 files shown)


src/docfx/template/JsonSchemaProvider.cs
```diff
@@ -75,6 +75,17 @@ public RenderType GetRenderType(ContentType contentType, SourceInfo<string?> mim
             };
         }
 
+        public bool IsContentRenderType(string? mime)
```
yufeih, line 78: Why not using `GetRenderType`, this method seems like a duplication of `GetRenderType`.

test/docfx.Test/DocfxTest.cs
```diff
@@ -206,30 +235,67 @@ private static async Task RunLanguageServer(string docsetPath, DocfxTestSpec spe
...
         {
-            var randomOutputPath = Path.ChangeExtension(outputPath, $".{Guid.NewGuid()}");
+            var randomNormalOutputPath = Path.ChangeExtension(outputPath, $".normal.{Guid.NewGuid()}");
+            var randomFirstStageOutputPath = Path.ChangeExtension(outputPath, $".firststage.{Guid.NewGuid()}");
+            var randomContinueOutputPath = Path.ChangeExtension(outputPath, $".continue.{Guid.NewGuid()}");
 
             docsetPath = Path.Combine(docsetPath, spec.Cwd ?? "");
 
+            var isOutputTypePageJson = IsOutputTypePageJson(spec);
+            var isTwoStagesBuild = isOutputTypePageJson && !dryRun && !spec.NoContinue;
```
yufeih, line 245: Update `ExpandTest` to expand `ContinueBuild` into a new test case. Now 1 stage build of `outputType: pagejson` tests are not running.

test/docfx.Test/DocfxTest.cs
```diff
@@ -103,6 +103,35 @@ public static void Run(TestData test, DocfxTestSpec spec)
...
+                foreach (var item in repo)
+                {
+                    existInRepos = existInRepos || IsExist(item.Files);
+                }
+            }
+
+            return existInInputs || existInRepos;
+        }
+
+        private static bool IsExist(IDictionary<string, string> dic, string expect = "outputType: pageJson")
```
yufeih, line 122: nit: consider move this method as an inner method inside `IsOutputTypePageJson`. Could also use _LINQ_ to simplify `IsExist` and `IsOutputTypePageJson`.

docs/specs/legacy.yml
```diff
@@ -1,4 +1,5 @@
 # Build output as page json
+noContinue: true
```
yufeih, line 2: I don't quite following the _"cannot use the cache info in spec test"_ argument, the second stage presumably don't need cache info?

test/docfx.Test/DocfxTest.cs
```diff
@@ -206,30 +235,67 @@ private static async Task RunLanguageServer(string docsetPath, DocfxTestSpec spe
...
+
+                        var templatePath = GetTemplatePath(package, s_repos.Value);
+                        var continueCommandLine = new[]
+                        {
+                            "build", randomFirstStageOutputPath,
+                            "--output", randomContinueOutputPath,
+                            "--log", Path.Combine(randomContinueOutputPath, ".errors.log"),
+                            "--continue",
+                            "--template", templatePath,
+                            spec.NoRestore ? "--no-restore" : null,
```
yufeih, line 292: Consider explicitly pass `--output-type pageJson`. We should also check that `--continue` must be paired with `--output-type pagejson` in the build command, because output types like HTML is not supported yet.

test/docfx.Test/DocfxTest.cs
```diff
@@ -206,30 +235,67 @@ private static async Task RunLanguageServer(string docsetPath, DocfxTestSpec spe
 
         private static void RunBuild(string docsetPath, string outputPath, bool dryRun, bool singleFile, DocfxTestSpec spec, Package package)
         {
-            var randomOutputPath = Path.ChangeExtension(outputPath, $".{Guid.NewGuid()}");
+            var randomNormalOutputPath = Path.ChangeExtension(outputPath, $".normal.{Guid.NewGuid()}");
```
yufeih, line 238: If `randomContinueOutputPath` is set to `randomOutputPath`, we probably only need one additional `randomIntermediatePath` for 2 stage build.

test/docfx.Test/DocfxTest.cs
```diff
@@ -238,12 +304,74 @@ private static void RunBuild(string docsetPath, string outputPath, bool dryRun,
...
                 : spec.Outputs;
 
-            VerifyOutput(randomOutputPath, outputs);
+            VerifyOutput(randomNormalOutputPath, outputs);
 
-            if (Directory.Exists(randomOutputPath))
+            if (isTwoStagesBuild)
             {
-                Directory.Delete(randomOutputPath, recursive: true);
+                CopyNecessaryFilesForContinue(randomNormalOutputPath, randomContinueOutputPath);
```
yufeih, line 311: Instead of copying files around, we only need to verify JSON files in 2 stage build

docs/specs/learn.yml
```diff
@@ -1,5 +1,6 @@
 ---
 # Draft run learn post validation
+noContinue: true
```
yufeih, line 3: `hieararchy.json` is a special file, we could filter it in test code.

src/docfx/template/JsonSchemaProvider.cs
mingwli: There's no context to get `ContentType` here, the only input left is `schema`

yufeih (approved):
Presumably we don't need `noContinue`. Also please fix spell checker errors.

docs/specs/legacy.yml
```diff
@@ -424,7 +424,7 @@ inputs:
   a.md: '[](https://github.com)'
   b.yml: |
     #YamlMime:TestPage
-    link: https://docs.com
+    link: https://static.docs.com
```
mingwli, line 427: should use value in `trustedDomain` for ContinueBuild to pass

PR #7345, opened 2021-04-30
Report git-clone-failed for fallback clone

mingwli:
[AB#418222](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/418222)

When fallback remote is resolved but failed to clone => report `git-clone-failed` error and `abort`
When fallback remote is not resolved but failed to clone => allow `continue to build` without fallback repo


herohua (approved):

src/docfx/lib/LocalizationUtility.cs
```diff
@@ -65,6 +65,7 @@ public static string AddLeftToRightMarker(CultureInfo culture, string text)
                         return Path.Combine(fallbackRepoPath, docsetSourceFolder);
                     }
                 }
+                throw Errors.System.GitCloneFailed(fallbackRemote, fallbackBranch ?? "").ToException();
```
yufeih, line 68: Consider call `TryResolvePackage` for `fallbackBranch` and `ResolvePackage` for `main`. `ResolvePackage` method may throw different exception types or message in the future and it shouldn't be a concern of this method.

yufeih (approved):

PR #7361, opened 2021-05-13
use include-not-found for empty include link

v-pegao:
[AB#415011](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/415011)

### DIFF
--- a/src/docfx/lib/markdown/IncludeExtension.cs
+++ b/src/docfx/lib/markdown/IncludeExtension.cs
@@ -64,8 +64,12 @@ namespace Microsoft.Docs.Build
         private static void ExpandInclusionBlock(
             MarkdownContext context, InclusionBlock inclusionBlock, MarkdownPipeline pipeline, MarkdownPipeline inlinePipeline, ErrorBuilder errors)
         {
-            if (!string.IsNullOrEmpty(inclusionBlock.IncludedFilePath)
-                && !inclusionBlock.IncludedFilePath.EndsWith(".md", StringComparison.OrdinalIgnoreCase))
+            if (string.IsNullOrEmpty(inclusionBlock.IncludedFilePath))
+            {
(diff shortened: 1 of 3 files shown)


src/docfx/lib/markdown/IncludeExtension.cs
```diff
@@ -64,8 +64,12 @@ private static void RemoveIncludeRenderer(IMarkdownRenderer renderer)
         private static void ExpandInclusionBlock(
             MarkdownContext context, InclusionBlock inclusionBlock, MarkdownPipeline pipeline, MarkdownPipeline inlinePipeline, ErrorBuilder errors)
         {
-            if (!string.IsNullOrEmpty(inclusionBlock.IncludedFilePath)
-                && !inclusionBlock.IncludedFilePath.EndsWith(".md", StringComparison.OrdinalIgnoreCase))
+            if (string.IsNullOrEmpty(inclusionBlock.IncludedFilePath))
```
live1206, line 67: I am wondering, would it be easier to log the error in `linkResolver` directly?
https://github.com/dotnet/docfx/blob/ee2c1aa238e40824789c2b829411f2d1913fe63f/src/docfx/build/link/LinkResolver.cs#L60

v-pegao: Current `TryResolveFile()` will resolve the empty link as SelfBookmark, we need identity the link type(include/markdown link) if we want to write `IncludeNotFound` error here.

Before go into `TryResolveFile()`, in `ExpandInclusionBlock()`, we have already check if the link is empty or not when validating `IncludeInvalid`, I think we can write `IncludeNotFound` here before the process `ReadFile() => ResolveContent() => TryResolveFile()`

live1206: Got your point, makes sense to me now.

live1206 (approved):

PR #7362, opened 2021-05-14
Fix regression test conflict

AngryBerryMS:
[AB#424982](https://dev.azure.com/ceapex/Engineering/_workitems/edit/424982)


live1206 (approved):

test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -51,7 +51,13 @@ private static int Run(Options opts)
             try
             {
                 s_repository = opts.Repository;
-                s_repositoryName = $"{(opts.DryRun ? "dryrun." : "")}{Path.GetFileName(opts.Repository)}";
+
+                s_repositoryName = opts.DryRun
```
mingwli, line 55: rename to `s_testName`, the original name could be confusing in the future.

mingwli (changes requested):

mingwli (approved):

test/docfx.RegressionTest/RegressionTest.cs
AngryBerryMS: Done.

PR #7377, opened 2021-05-24
output moniker group in xrefmap to compress data

live1206:
[AB#411112](https://dev.azure.com/ceapex/Engineering/_workitems/edit/411112)

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -158,10 +158,25 @@ namespace Microsoft.Docs.Build
                 .OrderBy(xref => xref.Uid)
                 .ToArray();
 
+            var monikerGroups = new Dictionary<string, MonikerList>(
+                from item in references
+                let monikerGroup = item.MonikerGroup
+                where !string.IsNullOrEmpty(monikerGroup)
+                orderby monikerGroup
+                group item by monikerGroup into g
+                select new KeyValuePair<string, MonikerList>(g.Key, g.First().Monikers));
+
             externalXrefs = _jsonSchemaTransformer().GetValidateExternalXrefs();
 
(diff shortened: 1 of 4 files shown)


yufeih (approved):

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -158,14 +158,20 @@ public XrefMapModel ToXrefMapModel()
...
                 .ToArray();
 
-            externalXrefs = _jsonSchemaTransformer().GetValidateExternalXrefs();
+            var monikerGroups = new Dictionary<string, MonikerList>(
+                from item in references
+                let monikerGroup = item.MonikerGroup
+                where !string.IsNullOrEmpty(monikerGroup)
+                orderby monikerGroup
+                group item by monikerGroup into g
+                select new KeyValuePair<string, MonikerList>(g.Key, g.First().Monikers));
```
mingwli, line 167: some minor thinking: can this be shared with publish model somehow?

live1206: Thought about it, did not find a proper utility for it, though.

mingwli: What about caching a the moniker map<hash, monikers> here https://github.com/dotnet/docfx/blob/v3/src/docfx/build/moniker/MonikerProvider.cs#L68-L71?

live1206: Actually, there is not much to share in this case.
We have different items, both item happen to have `MonikerGroup` and `Monikers`, and what we are doing is to construct a mapping from `MonikerGroup` to `Monikers`, `MonikerGroup` is computed already. No extra calculation or cache needed.

PR #7380, opened 2021-05-24
fix output path conflict

AngryBerryMS:
[AB#429891](https://dev.azure.com/ceapex/Engineering/_workitems/edit/429891)

### DIFF
--- a/src/docfx/publish/PublishUrlMap.cs
+++ b/src/docfx/publish/PublishUrlMap.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
-using System.IO;
 using System.Linq;
 
 namespace Microsoft.Docs.Build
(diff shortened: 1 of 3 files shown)


src/docfx/publish/PublishUrlMap.cs
```diff
@@ -132,16 +132,34 @@ private PublishUrlMapItem ResolveOutputPathConflicts(IGrouping<string, PublishUr
             }
 
             // redirection file is preferred than source file
-            var redirection = conflicts.FirstOrDefault(x => x.SourcePath.Origin == FileOrigin.Redirection);
-            if (redirection != null)
+            var redirections = conflicts.Where(x => x.SourcePath.Origin == FileOrigin.Redirection).OrderBy(x => x.SourcePath.Path, PathUtility.PathComparer);
+            var nonRedirections = conflicts.Where(x => x.SourcePath.Origin != FileOrigin.Redirection)
+                .OrderBy(x => x.SourcePath.Path, PathUtility.PathComparer)
+                .Select(x => x.SourcePath.Path.Value).ToList();
```
live1206, line 138: Change the input type to `IEnumerable`, `ToList()` can be removed

src/docfx/publish/PublishUrlMap.cs
```diff
@@ -132,16 +132,34 @@ private PublishUrlMapItem ResolveOutputPathConflicts(IGrouping<string, PublishUr
...
+            var redirection = redirections.FirstOrDefault();
+            var redirectionCount = redirections?.Count() ?? 0;
```
live1206, line 140: why is `?` needed here?

src/docfx/publish/PublishUrlMap.cs
```diff
@@ -132,16 +132,34 @@ private PublishUrlMapItem ResolveOutputPathConflicts(IGrouping<string, PublishUr
...
+                return conflicts.OrderBy(x => x.SourcePath.Path, PathUtility.PathComparer).Last();
             }
+            else if (redirectionCount == 1)
+            {
+                _errors.Add(Errors.Redirection.RedirectedFileNotRemoved(nonRedirections));
+                return redirection!;
+            }
+            else
+            {
+                if (nonRedirectionCount > 0)
```
live1206, line 155: Can this be `conflicts.Count() > redirectionCount`? We can save a field declaration

src/docfx/publish/PublishUrlMap.cs
```diff
@@ -132,16 +132,34 @@ private PublishUrlMapItem ResolveOutputPathConflicts(IGrouping<string, PublishUr
             }
 
             // redirection file is preferred than source file
-            var redirection = conflicts.FirstOrDefault(x => x.SourcePath.Origin == FileOrigin.Redirection);
-            if (redirection != null)
+            var redirections = conflicts.Where(x => x.SourcePath.Origin == FileOrigin.Redirection).OrderBy(x => x.SourcePath.Path, PathUtility.PathComparer);
+            var nonRedirections = conflicts.Where(x => x.SourcePath.Origin != FileOrigin.Redirection)
+                .OrderBy(x => x.SourcePath.Path, PathUtility.PathComparer)
+                .Select(x => x.SourcePath.Path.Value).ToList();
+            var redirection = redirections.FirstOrDefault();
```
live1206, line 139: nit: consider put this declaration into a block where `redirectionCount > 0`, then we can get rid of all the nullable handling

AngryBerryMS: There are 2 parts that satisfy `redirectionCount > 0`. I don't want too many nested `if...else...`

src/docfx/publish/PublishUrlMap.cs
AngryBerryMS: Fixed.

src/docfx/publish/PublishUrlMap.cs
AngryBerryMS: Done.

src/docfx/publish/PublishUrlMap.cs
```diff
@@ -132,16 +132,33 @@ private PublishUrlMapItem ResolveOutputPathConflicts(IGrouping<string, PublishUr
...
+                {
+                    _errors.Add(Errors.Redirection.RedirectedFileNotRemoved(nonRedirections));
+                }
 
-            // otherwise, prefer the one based on FilePath
-            _errors.Add(Errors.UrlPath.OutputPathConflict(conflicts.First().OutputPath, conflicts.Select(x => x.SourcePath)));
-            return conflicts.OrderBy(x => x.SourcePath.Path, PathUtility.PathComparer).Last();
+                _errors.Add(Errors.UrlPath.OutputPathConflict(conflicts.First().OutputPath, conflicts.Select(x => x.SourcePath)));
+                return redirection!;
+            }
```
mingwli, line 161:
```c#
            if (redirection != null && conflicts.Any(x => x.SourcePath.Origin != FileOrigin.Redirection))
            {
                // redirection file is preferred than source file
                _errors.Add(Errors.Redirection.RedirectedFileNotRemoved(redirection.SourcePath));
                return redirection;
            }

            return conflicts.OrderBy(x => x.SourcePath.Path, PathUtility.PathComparer).Last();
```

What if always report `output-path-conflict`, and report `redirected-file-not-removed` when conflicts contain both.
This make the logic easier to describe and maintain, and won't cause extra confusion to end user.

@live1206 @herohua @AngryBerryMS

live1206: We need to inform the user that, a redirection has been added and the original file needs to be removed.
This removing action point needs to be stressed out, we had this requirement from user previously and confirmed with PM.

live1206 (approved):

PR #7414, opened 2021-06-09
Dedicated graph client for PubDev env

mingwli:
[AB#432334](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/432334)

### DIFF
--- a/src/docfx/config/Config.cs
+++ b/src/docfx/config/Config.cs
@@ -298,12 +298,12 @@ namespace Microsoft.Docs.Build
         /// <summary>
         /// Tenant id that can be used to access the Microsoft Graph API.
         /// </summary>
-        public string MicrosoftGraphTenantId { get; init; } = "72f988bf-86f1-41af-91ab-2d7cd011db47";
+        public string? MicrosoftGraphTenantId { get; init; }
 
         /// <summary>
         /// Client id that can be used to access the Microsoft Graph API.
         /// </summary>
-        public string MicrosoftGraphClientId { get; init; } = "b6b77d19-e9de-4611-bc6c-4f44640ec6fd";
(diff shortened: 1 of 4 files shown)


src/docfx/config/Config.cs
```diff
@@ -303,7 +303,9 @@ internal class Config : PreloadConfig
         /// <summary>
         /// Client id that can be used to access the Microsoft Graph API.
         /// </summary>
-        public string MicrosoftGraphClientId { get; init; } = "b6b77d19-e9de-4611-bc6c-4f44640ec6fd";
+        public string MicrosoftGraphClientId { get; init; } = EnvironmentVariable.DocsEnvironment == DocsEnvironment.Prod
```
yufeih, line 306: can this be set in `OpsXYZ`? in general try to put docs specific stuff there.

src/docfx/config/EnvironmentVariable.cs
```diff
@@ -21,6 +21,11 @@ public static class EnvironmentVariable
 
         public static string? SessionId => GetValue("DOCFX_SESSION_ID");
 
+        public static readonly DocsEnvironment DocsEnvironment =
```
yufeih, line 24: same here, try to limit `DOCS_` within `OpsXyz`

mingwli: These can all be kept in `OpsAbc`, just one issue, how can `RegressionTest` use the `PubDev graph client`

yufeih: regression test docfx config would override ops accessor

yufeih (approved):

PR #7429, opened 2021-06-17
fix: add ops metadata to FullMetadataSchemaApi

v-pegao:
[AB#439759](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/439759)


yufeih (changes requested):
This doesn't sound the right fix, the metadata rule only applies to stuff configured in the portal. Could you give a  short description of the cause of this bug?

PR #7437, opened 2021-06-28
Allow md property fallback from fallback file

mingwli:
[AB#447530](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/447530)

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -426,24 +426,15 @@ namespace Microsoft.Docs.Build
                     return link;
 
                 case JsonSchemaContentType.Markdown:
+
                     // todo: use BuildPage.CreateHtmlContent() when we only validate markdown properties' bookmarks
-                    return _markdownEngine.ToHtml(
-                        errors,
-                        content,
-                        sourceInfo,
-                        MarkdownPipelineType.Markdown,
-                        null,
-                        file.Origin == FileOrigin.Fallback || rootSchema.ContentFallback);
+                    var html = _markdownEngine.ToHtml(errors, content, sourceInfo, MarkdownPipelineType.Markdown, null, rootSchema.ContentFallback);
+                    return html;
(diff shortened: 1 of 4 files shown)


yufeih (commented):
Update ErrorLog to write the enus error only to raw log on verbose mode in case of fallback error.

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -426,15 +426,24 @@ private string GetXrefHref(FilePath file, string uid, int uidCount, bool isRootL
...
                 case JsonSchemaContentType.InlineMarkdown:
-
                     // todo: use BuildPage.CreateHtmlContent() when we only validate markdown properties' bookmarks
-                    return _markdownEngine.ToHtml(errors, content, sourceInfo, MarkdownPipelineType.InlineMarkdown, null, rootSchema.ContentFallback);
+                    return _markdownEngine.ToHtml(
+                        errors,
+                        content,
+                        sourceInfo,
+                        MarkdownPipelineType.InlineMarkdown,
+                        null,
```
yufeih, line 445: Change LinkResolver.TryResolveRelativePath method instead to:
- Make behavior consistent with markdown include
- Keep the value of contentFallback intact

mingwli:
> Update ErrorLog to write the enus error only to raw log on verbose mode in case of fallback error.

This shouldn't be removed for sometimes loc build fails due to en-us error, and we need to keep it to mark the build failed.

yufeih:
> > Update ErrorLog to write the enus error only to raw log on verbose mode in case of fallback error.
> 
> This shouldn't be removed for sometimes loc build fails due to en-us error, and we need to keep it to mark the build failed.

I mean to log the actual error that caused fallback-error not the fallback error itself. Today this isnt logged anywhere, making it hard to see what caused the fallback-error without a local debug.

yufeih (approved):
💯

PR #7438, opened 2021-06-28
Update spell checker

yufeih:


mingwli (approved):

mingwli:
In the future people only need to edit `expect.txt` then?

yufeih:
> In the future people only need to edit `expect.txt` then?

Yes

PR #7446, opened 2021-06-28
Link fix and clarification of plugins dir

BryanLa:


yufeih (approved):
Thanks for the contribution!

PR #7447, opened 2021-06-28
Clarify log message

BryanLa:


src/Microsoft.DocAsCode.Build.Engine/DocumentBuilder.cs
```diff
@@ -49,7 +49,7 @@ public class DocumentBuilder : IDisposable
             string commitToSHA = null,
             bool cleanupCacheHistory = false)
         {
-            Logger.LogVerbose("Loading plug-in...");
+            Logger.LogVerbose("Loading plug-ins and post-processors...");
```
yufeih, line 52: Line 61 only mentions plugins, maybe this would only be loading plugins since a post processor is technically a part of a plugin.

BryanLa: Thanks @yufeih. I discussed this in email with @superyyrrzz as well. The thought was that this would align better w/the docs, as the docs discuss them as 2 different concepts. 

So the function would say "Loading plug-ins and post-processors" at line 52.
Then line 61 would give the plug-ins count
Then line 72 would give the post-processor info (where PostProcessorsManager is instantiated, which eventually calls [GetPostProcessor](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Build.Engine/PostProcessors/PostProcessorsManager.cs#L65)). 

I was just hoping to make this output more aligned with the way the [plug-ins](https://dotnet.github.io/docfx/tutorial/howto_build_your_own_type_of_documentation_with_custom_plug-in.html) and [post-processors](https://dotnet.github.io/docfx/tutorial/howto_add_a_customized_post_processor.html) are presented in the docs, which presents them as 2 different concepts.

yufeih (approved):

PR #7449, opened 2021-06-29
Template clarifications

BryanLa:


Documentation/tutorial/toc.yml
```diff
@@ -19,9 +19,9 @@
     href: intro_overwrite_files.md
   - name: Links and Cross References
     href: links_and_cross_references.md
-- name: Template
+- name: DocFx Template System
```
yufeih, line 22: Since the whole document is all about DocFX, consider remove it from here and also line 5?
```suggestion
- name: Template System
```

BryanLa: Yes, agree, I like that. I'll do a follow up commit and change those as well..

yufeih (approved):
Thanks you! 🎉🎉🎉

Issue #7422, opened 2021-06-11
Using memberpage results in exception


jamiehankins:
**Operating System**: Windows 21H1

**DocFX Version Used**: 2.58.0

**Template used**: `default`, `memberpage`

**Steps to Reproduce**:

1. `git clone https://github.com/jamiehankins/DocFXBug.git` (this is a very small project just to repro the issue)
2. `cd DocFXBug`
3. `.\build.cmd`

**Expected Behavior**: Build the project.

**Actual Behavior**: DocFX crashes.

This regression happened somewhere between 2.56.2 and 2.58.0.

After you've built the project, you can just run docfx.exe directly like so to see the direct output:
```
docfx --loglevel error TestLib\src\docfx.json
```

Here is the direct output:
```
[21-06-11 01:25:44.865]Error:[BuildCore.Build Document.CompilePhaseHandlerWithIncremental.ManagedReferenceDocumentProcessor.Build.BuildManagedReferenceDocument](../docs/api/TestLib.TestClass.yml)Markup failed: Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040).


Build failed.
[21-06-11 01:25:44.933]Error:[BuildCore.Build Document.CompilePhaseHandlerWithIncremental.ManagedReferenceDocumentProcessor.Build.BuildManagedReferenceDocument](../docs/api/TestLib.TestClass.yml)Markup failed: Could not load file or assembly 'System.Runtime.CompilerServices.Unsafe, Version=4.0.4.1, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040).
```

This is a pretty big deal for us. I can't go backwards because I run into issue [4880](https://github.com/dotnet/docfx/issues/4880). I can't use current because of this.

jamiehankins:
Note also that there's some hackery in the csproj file to copy memberpage from the NuGet cache. This doesn't affect the repro. There's no good documentation that I can find on how I'm supposed to run memberpage. The instructions just talk about installing it from the command line directly into your DocFX project directory. That means either "exec"ing Nuget.exe every time you build, or checking in the binaries.

[...]

PR #7451, opened 2021-07-01
Fix for memberpage reference (dotnet/docfx#7422)

jamiehankins:
The memberpage (Microsoft.DocAsCode.Build.MemberLevelManagedReference) project needed a reference to
System.Runtime.CompilerServices.Unsafe. Also, it needed to publish docfx.plugins.config with a binding redirect.

It also needs System.Collections.Immutable.


yufeih (approved):
Thanks for fixing this!

PR #7455, opened 2021-07-05
handle validation exception

xichan-ms:
[AB#452178](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/452178)

### DIFF
--- a/src/docfx/config/ops/OpsMetadataRuleConverter.cs
+++ b/src/docfx/config/ops/OpsMetadataRuleConverter.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Docs.Build
         };
 
 #pragma warning disable MEN003 // Method is too long
-        public static string GenerateJsonSchema(string rulesContent, string allowlistsContent)
+        public static string GenerateJsonSchema(string rulesContent, string allowlistsContent, ErrorBuilder errors)
 #pragma warning restore MEN003 // Method is too long
         {
             Log.Write(rulesContent);
@@ -46,6 +46,7 @@ namespace Microsoft.Docs.Build
             catch (Exception ex)
             {
                 Log.Write(ex);
+                errors.Add(Errors.System.ValidationIncomplete());
             }
             if (rules == null || rules.Count == 0)
             {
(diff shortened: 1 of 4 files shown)


src/docfx/config/ops/OpsMetadataRuleConverter.cs
```diff
@@ -29,19 +29,29 @@ internal static class OpsMetadataRuleConverter
...
+            Rules? rules = null;
+            Taxonomies taxonomies = new();
+            try
+            {
+                rules = JsonConvert.DeserializeObject<Rules>(rulesContent);
+                taxonomies = JsonConvert.DeserializeObject<Taxonomies>(allowlistsContent) ?? new();
+            }
+            catch (Exception ex)
+            {
+                Log.Write(ex);
```
fkpwolf, line 48: should add `_errors.Add(Errors.System.ValidationIncomplete());`?

xichan-ms: done.

src/docfx/validation/CustomRuleProvider.cs
```diff
@@ -254,9 +256,18 @@ private static ErrorLevel ConvertSeverity(ValidationSeverity severity)
 
         private Dictionary<string, ValidationRules>? GetValidationRules(SourceInfo<string> rules)
         {
-            return !string.IsNullOrEmpty(rules.Value)
-                ? JsonUtility.DeserializeData<Dictionary<string, ValidationRules>>(_fileResolver.ReadString(rules), rules.Source?.File)
-                : null;
+            try
+            {
+                return !string.IsNullOrEmpty(rules.Value)
```
live1206, line 261: Instead of catching in the inner method, how about catching on the outer layer instead?
This time, it fails on JSON deserialization, it could fail with other things next time.
We can think about the build pipeline, where we should not fail to let the build continue.

For instance, in the `[DocsetBuilder](https://github.com/dotnet/docfx/blob/39856aad02896115625e38763b6d36be7bb07bee/src/docfx/build/DocsetBuilder.cs#L90)`, when we can continue without these dependencies generated?

yufeih (changes requested):
Reject this PR for now given the reasons I put in the bug.

Issue #7463, opened 2021-07-09
MSBuildLocator fails to use VS2019 build tools when VS2017 is installed.


ListlessValkyrie:
**Windows**

**DocFX v2.58**: 

**default**

**Steps to Reproduce**:

Go and have a look at:
https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Metadata.ManagedReference/MsBuildEnvironmentScope.cs

On line 68:
```
var instances = MSBuildLocator.QueryVisualStudioInstances().ToList();

// workaround for https://github.com/dotnet/docfx/issues/1969
// FYI https://github.com/dotnet/roslyn/issues/21799#issuecomment-343695700
var latest = instances.FirstOrDefault(a => a.Version.Major >= 15);
if (latest != null)
```

Our build server has multiple MSBuild versions installed, and for whatever reason when you enumerate the above the VS2017 one is enumerated first.

(I rebuilt DocFx and put in some logging to get it to log out the instances, its just logging path and version) to get this:

```
foreach(var instance in instances)
{
    Logger.LogInfo($"{instance.MSBuildPath} {instance.Version.Major}");
}
```

09-Jul-2021 13:49:15 | Running DocFx:
-- | --
09-Jul-2021 13:49:15 | [21-07-09 12:49:15.839]Info:[MetadataCommand.ExtractMetadata]C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\MSBuild\15.0\Bin 15
09-Jul-2021 13:49:15 | [21-07-09 12:49:15.841]Info:[MetadataCommand.ExtractMetadata]C:\Program Files (x86)\Microsoft Visual Studio\2019\BuildTools\MSBuild\Current\Bin 16
09-Jul-2021 13:49:15 | [21-07-09 12:49:15.841]Info:[MetadataCommand.ExtractMetadata]C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\MSBuild\Current\Bin 16

Because you are using `var latest = instances.FirstOrDefault(a => a.Version.Major >= 15);` it is going to pick the 2017 build tools.

**Expected Behavior**:

I need it to use the VS2019 build tools.

**Actual Behavior**:

It always uses the VS2017 build tools.

**Fix**:

You just need to order them by descending first:

Change line 68 to be:
```
var latest = instances
    .OrderByDescending(a => a.Version.Major)
    .FirstOrDefault(a => a.Version.Major >= 15);
```

[...]

PR #7469, opened 2021-07-14
When getting an MSBuild instance, get the latest

jamiehankins:
The issue ticket suggested sorting the instances. Really, though, we just want the latest one, so we can do it in O(n).

Tested to make sure things are fine if instances is null or empty.

Fix for https://github.com/dotnet/docfx/issues/7463.


yufeih (approved):
LGTM. Thank you for this fix 👍

jamiehankins:
Glad to help.

PR #7491, opened 2021-07-26
Add DocFx addin for Cake to list of tools

pascalberger:
Adds addin to run DocFx from Cake builds to the list of tools.


yufeih (approved):
Thank you!

PR #7508, opened 2021-08-03
enable https for sitemap

AngryBerryMS:
[AB#454390](https://dev.azure.com/ceapex/Engineering/_workitems/edit/454390)


AngryBerryMS:
Should keep it as http

PR #7512, opened 2021-08-04
fix tsa

AngryBerryMS:


azure-pipelines-v2-lgtm.yml
```diff
@@ -8,6 +8,16 @@ trigger:
     - main
 pr: none
 
+variables:
+  - name: LGTM.SnapshotIdentifiers
```
live1206, line 12: can be simplified to `key:value`

live1206 (approved):

PR #7517, opened 2021-08-09
AAD auth to call Build API

live1206:
[AB#468064](https://dev.azure.com/ceapex/Engineering/_workitems/edit/468064)


test/docfx.Test/build/OpsConfigAdapterTest.cs
```diff
@@ -44,5 +44,31 @@ public static async Task AdaptOpsServiceConfig(string url, string expectedJson)
                 JToken.Parse(expectedJson.Replace('\'', '"')),
                 JToken.Parse(actualConfig));
         }
+
+        [Theory]
```
yufeih, line 48: The `AdaptOpsServiceConfig` test is skipped via `[SkippableTheory]`, do you know why it is skipped and can it be re-enabled?

If it cannot be re-enabled, remove the test. If it can, enable it and share the `[InlineData]` with the above test using `[MemberData]`.

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -23,17 +22,10 @@ internal class OpsAccessor : ILearnServiceAccessor
...
-
-        private static readonly SecretClient s_secretClientPerf = new(new("https://kv-opbuild-test.vault.azure.net/"), new DefaultAzureCredential());
-
-        private static readonly Lazy<Task<string>> s_opsTokenPublic = new(() => GetSecret("OpsBuildUserToken", DocsEnvironment.Prod));
-        private static readonly Lazy<Task<string>> s_opsTokenPubDev = new(() => GetSecret("OpsBuildUserToken", DocsEnvironment.PPE));
-        private static readonly Lazy<Task<string>> s_opsTokenPerf = new(() => GetSecret("OpsBuildUserToken", DocsEnvironment.Perf));
+        private static readonly CachedTokenCredential s_cachedTokenCredential = new(new DefaultAzureCredential());
+        private static readonly string s_opsClientId = DocsEnvironment == DocsEnvironment.Prod
+                ? "6befca88-4c28-430a-957e-f870b267bcfc"
+                : "6ce33073-a071-4cf9-9936-f5a24d21a089";
```
yufeih, line 28: Is perf using the same id as ppe?

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -189,7 +181,7 @@ async Task<HttpResponseMessage> ValidationMiddleware(HttpRequestMessage request,
 
         private Task<string> FetchBuild(string urlPath, string? value404 = null, DocsEnvironment? environment = null, HttpMiddleware? middleware = null)
         {
-            return Fetch(BuildApi(environment) + urlPath, value404, BuildMiddleware(environment, middleware));
+            return Fetch(BuildApi(environment) + urlPath, value404, BuildMiddleware(middleware));
```
yufeih, line 184: It looks like a bug to remove the `environment` here. The URL is dependent on environment, but the secrets aren't?

src/docfx/lib/CachedTokenCredential.cs
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) Microsoft. All rights reserved.
```
yufeih, line 1: We probably don't need this, its OK to cache the token once per process per environment using `Lazy<T>` just like before.

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -225,43 +217,29 @@ async Task<HttpResponseMessage> SendRequest(HttpRequestMessage request)
...
 
                 return await (middleware != null ? middleware(request, next) : next(request));
             };
         }
 
+        private static async Task<string> GetAccessTokenAsync()
+        {
+            var accessToken = await s_cachedTokenCredential
+                .GetTokenAsync(new TokenRequestContext(new[] { $"{s_opsClientId}/.default" }), default)
+                .ConfigureAwait(false);
```
yufeih, line 239: Generally we don't append `.ConfigureAwait(false)`

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -225,43 +217,29 @@ async Task<HttpResponseMessage> SendRequest(HttpRequestMessage request)
...
-                        Log.Write(ex);
-                    }
+                    request.Headers.TryAddWithoutValidation("Authorization", $"Bearer {await GetAccessTokenAsync()}");
                 }
 
                 return await (middleware != null ? middleware(request, next) : next(request));
             };
         }
 
+        private static async Task<string> GetAccessTokenAsync()
```
yufeih, line 235: Ensure this works for users using vscode extension who does not login. I don't remember exactly in that case did vscode extension not pass `X-OP-BuildUserToken` or it passes a well known public user `X-OP-BuildUserToken`. cc @928PJY

src/docfx/config/ops/OpsAccessor.cs
mingwli: Only two aad apps:
image

src/docfx/config/ops/OpsAccessor.cs
928PJY: Good point, for the public user, We just don't pass the token. We should not throw exceptions if we cannot get the AAD token

yufeih (changes requested):
Wait until we sort out internal developer auth.

test/docfx.Test/build/OpsConfigAdapterTest.cs
live1206: `[SkippableTheory]` is to skip these tests locally since there is no token provide.

Basically, there are 2 kinds of tests, one set is with op token provided, one is not.

yufeih: Try use `[MemberData]` to share the test data.

azure-pipelines-v3.yml
```diff
@@ -40,13 +40,15 @@ jobs:
...
 
   # dotnet test
-  - task: DotNetCoreCLI@2
+  - task: AzureCLI@2
     displayName: dotnet test
     condition: and(succeeded(), ne(variables['warm-up'], 'true'))
     inputs:
-      command: test
-      arguments: -c Release --no-build /p:CollectCoverage=true /p:CoverletOutputFormat=cobertura
-      publishTestResults: true
```
yufeih, line 49: Need a dedicated publish test result step

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -231,37 +224,37 @@ private static HttpMiddleware BuildMiddleware(DocsEnvironment? environment = nul
...
-
-                        request.Headers.Add("X-OP-BuildUserToken", await token.Value);
+                        request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", (await accessToken.Value).Token);
                     }
-                    catch (Exception ex)
+                    catch
                     {
-                        Log.Write($"Cannot get 'OPBuildUserToken' from azure key vault, please make sure you have been granted the permission to access.");
-                        Log.Write(ex);
+                        Log.Write("Fail to get AAD access token");
```
yufeih, line 243: Log the exception details as before.

test/docfx.Test/build/OpsConfigAdapterTest.cs
```diff
@@ -12,19 +13,32 @@ namespace Microsoft.Docs.Build
 {
     public static class OpsConfigAdapterTest
     {
+        public static IEnumerable<object[]> TestData => new[]
```
yufeih, line 16: nit: Use `TheoryData` to use its simpler object construction syntax.

test/docfx.Test/build/OpsConfigAdapterTest.cs
```diff
@@ -12,19 +13,32 @@ namespace Microsoft.Docs.Build
...
+                "{'product':'MSDN','siteName':'Docs','hostName':'ppe.docs.microsoft.com','basePath':'/e2eppe-azure-documents','xrefHostName':'ppe.docs.microsoft.com'}",
+            },
+            new[]
+            {
+                "https://ops/buildconfig/?name=E2E_DocFxV3&repository_url=https://github.com/OPS-E2E-PPE/E2E_DocFxV3/&branch=master",
+                "{'product':'MSDN','siteName':'Docs','hostName':'ppe.docs.microsoft.com','basePath':'/E2E_DocFxV3','xrefHostName':'ppe.docs.microsoft.com'}",
+            },
+        };
+
+        [Theory]
```
live1206, line 40: This should be `SkippableTheory` for local dev experience

mingwli: really?

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -262,6 +249,13 @@ private static HttpMiddleware BuildMiddleware(DocsEnvironment? environment = nul
             };
         }
 
+        private static ValueTask<AccessToken> GetAccessTokenAsync(DocsEnvironment? environment = null)
+        {
+            var defaultAzureCredential = new DefaultAzureCredential();
+            return defaultAzureCredential.GetTokenAsync(
+                new TokenRequestContext(new[] { $"{DocsBuildApiAADClientId(environment)}/.default" }));
```
yufeih, line 256: Will a user successfully build a repo if:
- Didn't login to vscode
- Login to azure CLI or visual studio
- Does not have access to build API

mingwli: lemme try

mingwli: @yufeih 
public repo: ✔
private repo: ❌
image

@928PJY  Is building private repo as public user a valid use case currently? 
e.g.: Trying to build https://github.com/MicrosoftDocs/test(private), but not an OPS user.

live1206: I think this is the expected behavior.

928PJY: no, it is not a valid use case for the public user to build a private

mingwli: for building public repo, it's downloading all config related APIs
image

azure-pipelines-v3.yml
```diff
@@ -40,22 +40,31 @@ jobs:
...
       DOCS_OPS_TOKEN: $(opBuildUserToken)
       DOCFX_MICROSOFT_GRAPH_TENANT_ID: '72f988bf-86f1-41af-91ab-2d7cd011db47'
       DOCFX_MICROSOFT_GRAPH_CLIENT_ID: 'b799e059-9dd8-4839-a39c-96f7531e55e2'
       MICROSOFT_GRAPH_CLIENT_CERTIFICATE: $(Docfx-MS-Graph-Client-Cert-PubDev)
+  
+  - task: PublishTestResults@2
+    inputs:
+      displayName: Publish Test Results
+      testResultsFormat: VSTest
+      testResultsFiles: '**/cloudtest_*.trx'
```
yufeih, line 63: Why `cloudtest_`?
```suggestion
      testResultsFiles: '**/*.trx'
```

yufeih (approved):

azure-pipelines-v3.yml
mingwli: image
Seems all `dotnet test` result in CI run is prefixed with this, lemme try if this can be specified.

mingwli: with `dotnet test -c Release --no-build --logger "trx;LogFileName=TestResults.trx" --results-directory ./_temp`, one result file will overwrite another.
well, may have to `foreach *.csproj` to specify the result file name for each single test project.

Simplified as suggested.

mingwli:
image
image
local run passed

mingwli:
> Wait until we sort out internal developer auth.

resolved

PR #7530, opened 2021-08-16
Use real url for download-failed error

v-pegao:
[AB#451575](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/451575)


src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -206,7 +207,8 @@ private async Task<string> Fetch(Func<HttpRequestMessage> requestFactory, string
                 return value404;
             }
 
-            return await response.EnsureSuccessStatusCode().Content.ReadAsStringAsync();
+            await EnsureSuccessStatusCode(response);
+            return await response.Content.ReadAsStringAsync();
 
             async Task<HttpResponseMessage> SendRequest(HttpRequestMessage request)
             {
```
yufeih, line 214: Consider use an `AsyncLocal<string>` to store and retrieve the current request URL.

v-pegao: To store the request url, we need reference type(e.g. AsyncLocal<_CustomClass_>), is it acceptable to use a local variable to store request url here? It may request lot's of change to pass AsyncLocal<_CustomClass_> to here.

yufeih: Sure, since now the code does not touch `CredentialHandler.cs`

src/docfx/restore/FileResolver.cs
```diff
@@ -235,7 +235,7 @@ private static PathString GetRestoreEtagPath(string url)
             }
             catch (Exception ex) when (!DocfxException.IsDocfxException(ex, out _))
             {
-                throw Errors.System.DownloadFailed(UrlUtility.SanitizeUrl(url)).ToException(ex);
+                throw Errors.System.DownloadFailed(UrlUtility.SanitizeUrl(ex.Data["request-url"]?.ToString() ?? url)).ToException(ex);
```
yufeih, line 238:
```suggestion
                throw Errors.System.DownloadFailed(UrlUtility.SanitizeUrl(ex.Data["RequestUrl]?.ToString() ?? url)).ToException(ex);
```

src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -220,16 +222,29 @@ private async Task<string> Fetch(Func<HttpRequestMessage> requestFactory, string
                 return value404;
             }
 
-            return await response.EnsureSuccessStatusCode().Content.ReadAsStringAsync();
+            await EnsureSuccessStatusCode(response);
```
yufeih, line 225: keep the existing way of throwing the exception:
```suggestion
      try
      {
            return await response.EnsureSuccessStatusCode().Content.ReadAsStringAsync();
      }
      catch (Exception ex)
      {
          ex.Data["RequestUrl"] = requestUrl;
          throw;
      }
```

yufeih (approved):

PR #7541, opened 2021-08-19
Skip AAD tests for release

live1206:
https://github.com/dotnet/docfx/runs/3357210415#step:4:1312


mingwli:
lemme fix in my #7540

live1206:
> lemme fix in my #7540

Let's separate it, and we can have a package immediately.

PR #7543, opened 2021-08-20
Use predefined environment variable for AAD tests

live1206:
Follow up for #7541

### DIFF
--- a/test/docfx.Test/build/OpsConfigAdapterTest.cs
+++ b/test/docfx.Test/build/OpsConfigAdapterTest.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Docs.Build
         public static async Task AdaptOpsServiceConfigWithAAD(string url, string expectedJson)
         {
             Skip.If(
-                bool.TryParse(Environment.GetEnvironmentVariable("IS_GITHUB_ACTION"), out var isGithubAction)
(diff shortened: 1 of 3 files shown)


test/docfx.Test/build/OpsConfigAdapterTest.cs
```diff
@@ -61,8 +60,7 @@ public static async Task AdaptOpsServiceConfigWithAAD(string url, string expecte
         {
             Skip.If(
                 bool.TryParse(Environment.GetEnvironmentVariable("IS_GITHUB_ACTION"), out var isGithubAction)
-                && isGithubAction
-                && !RuntimeInformation.IsOSPlatform(OSPlatform.Windows));
+                && isGithubAction);
```
yufeih, line 63: We don't need to invent our own environment variables, See https://docs.github.com/en/actions/reference/environment-variables for a list of predefined environment variables on github predefined environment varables.

mingwli: @live1206 my bad, didn't find this.

live1206: updated

yufeih (approved):

PR #7560, opened 2021-08-24
Fix wrong NuGet package layout

### LINKED ISSUE #7557: (Regression in 2.58.1) docfx.console.targets is no longer in the build folder. Now, it's in the root folder.
**Operating System**: Windows

**DocFX Version Used**: 2.58.1

**Template used**: default

**Steps to Reproduce**:

1. Create a C# project with a package reference to DocFX, like this:
```XML
<PackageReference
  Include="docfx.console"
  Condition="$(Configuration.Contains('WithDocs'))"
  Version="2.58.1"
  PrivateAssets="all"
  IncludeAssets="runtime; build; native; contentfiles; analyzers; buildtransitive" />
```
2. Configure a DocFX build. 
3. Build

**Expected Behavior**:
docfx.exe is run

**Actual Behavior**:
docfx.exe is not run

In 2.58.0, and every version before, docfx.console.targets was in the build folder of the NuGet package directory, e.g.
`.nuget\packages\docfx.console\2.58.0\build\docfx.console.targets`.

Starting with 2.58.1, it is now in the root directory for the package, e.g.
`.nuget\packages\docfx.console\2.58.0\docfx.console.targets`.

As a result, the targets in docfx.console.targets do not get run.

yufeih:
Fixes #7557 

There is a behavior change in NuGet 5.11.0 when <files> look like this. 

5.10.0 put `build/docfx.console.targets` inside `/build/docfx.console.targets`, but 5.11.0 put it under `/docfx.console.targets`.

```xml
    <file src="build/**" target=""/>
```

Related NuGet issue: https://github.com/NuGet/Home/issues/11125

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -21,6 +21,8 @@ steps:
 
 - task: NuGetToolInstaller@1
   displayName: 'Use NuGet'
+  inputs:
+    versionSpec: 5.10.0
 
 - task: CmdLine@1
   displayName: 'npm install'


superyyrrzz (approved):

jamiehankins:
@yufeih From looking at https://github.com/NuGet/Home/issues/11125, it appears that 5.10.0 is where the regression happened and 5.9.1 was the last good build.

PR #7599, opened 2021-09-06
error message should provide file which is missing moniker range in docfx.json

v-pegao:
ab#479580


src/docfx/build/moniker/MonikerProvider.cs
```diff
@@ -80,7 +81,7 @@ public MonikerList GetZoneLevelMonikers(ErrorBuilder errors, FilePath file, Sour
             // User should not define it in moniker zone
             if (configMonikerRange.Value is null && ValidateMoniker(file))
```
yufeih, line 82: consider rename it to "should validate moniker", do the same for the variable in line 115. Because `ValidateMoniker` is the action to actually validate the moniker.

```suggestion
            if (configMonikerRange.Value is null && ShouldValidateMoniker(file))
```

docs/specs/moniker.yml
```diff
@@ -473,7 +473,7 @@ outputs:
   docs/toc.json:
   .errors.log: |
     {"message_severity":"suggestion","code":"moniker-range-undefined","message":"Moniker range missing in docfx.yml/docfx.json, user should not define it in file metadata or moniker zone. NOTE: This Suggestion will become a Error on 06/30/2020.","file":"docs/a.md","line":2,"column":15}
-    {"message_severity":"suggestion","code":"moniker-range-undefined","message":"Moniker range missing in docfx.yml/docfx.json, user should not define it in file metadata or moniker zone. NOTE: This Suggestion will become a Error on 06/30/2020.","file":"docfx.yml","line":4,"column":16}
+    {"message_severity":"suggestion","code":"moniker-range-undefined","message":"Moniker range missing in docfx.yml/docfx.json, user should not define it in file metadata or moniker zone of 'docs/b.md'. NOTE: This Suggestion will become a Error on 06/30/2020.","file":"docfx.yml","line":4,"column":16}
```
yufeih, line 476: Change the `file` attribute instead, to align with the error in line 475

```suggestion
    {"message_severity":"suggestion","code":"moniker-range-undefined","message":"Moniker range missing in docfx.yml/docfx.json, user should not define it in file metadata or moniker zone. NOTE: This Suggestion will become a Error on 06/30/2020.","file":"docs/b.md","line":4,"column":16}
```

v-pegao: The 2 errors are 2 different cases, for case in L476, the undefined moniker range is defined in L4 of docfx.yml, in this case, User should go to docfx.yml to either build the file with the moniker ot remove the moniker definition from file metadata

yufeih: @v-pegao This `b.md` case doens't  seem like the scenario described the linked bug, can you work with @live1206 to figure out exactly what is the buggy scenario?

v-pegao: @yufeih cc @live1206 

I think the bug is the same scenario with the 'docs/b.md'

Description of the scenario:
1. Both the issue file in linked bug and 'docs/b.md' don't have `version`/`group` set in docfx config.
2. No moniker range is defined in publish config of both repo in linked bug and test repo
3. The undefined moniker range is mentioned in fileMetadata in docfx config for both the issue file in linked bug and 'docs/b.md'

With `#1` and `#2`, the ConfigMonikerRange is empty([GetConfigMonikerRangeCore()](https://github.com/dotnet/docfx/blob/v3/src/docfx/build/moniker/MonikerProvider.cs#L191-L213)). With `#3`, the undefined moniker is used and the error will be thrown([GetFileLevelMonikersCore()](https://github.com/dotnet/docfx/blob/v3/src/docfx/build/moniker/MonikerProvider.cs#L117-L127)).

yufeih: @v-pegao `b.md` in this cases configures `monikerRange` in `docfx.json/fileMetadata`, the bug didn't.

v-pegao: @yufeih The bug defined in docfx.json/fileMetadata as well, check the '[linkedin-api-docs/docfx.json](https://github.com/MicrosoftDocs/linkedin-api-docs/blob/a5a1a94b8178975af22e78e323e13ac067a239ca/linkedin-api-docs/docfx.json#L141-L142)', the original branch is deleted.

image

Issue #7617, opened 2021-09-15
Tabs do not preserve url query string


ivantam-fm:
**Operating System**: (`Windows` or `Linux` or `MacOS`)
Windows 10 20H2
MacOS Big Sur 11.5.1

**DocFX Version Used**: 
2.58.0.0

**Template used**: (`default` or `statictoc` or contain custom template)
default

**Steps to Reproduce**:
1. Navigate to a page with tabbed content. e.g. https://example.com?tabs=tabid-1
2. In the address bar, add a query parameter to the url, e.g. https://example.com?tabs=tabid-1&test=true
3. In the address bar, press enter to reload the page.

**Expected Behavior**:
The added query string is preserved.
i.e. https://example.com?tabs=tabid-1&test=true

**Actual Behavior**:
The added query string is removed, only the tab's query string remains.
i.e. https://example.com?tabs=tabid-1

**Possible Bug Location**:
Inside docfx.js, line 1001, updateTabsQueryStringParam(state):
```
function updateTabsQueryStringParam(state) {
  var qs = parseQueryString();
  qs.tabs = state.selectedTabs.join();
  var url = location.protocol + "//" + location.host + location.pathname + "?" + toQueryString(qs) + location.hash;
  if (location.href === url) {
    return;
  }
  history.replaceState({}, document.title, url);
}
```

`var qs = parseQueryString();` should be `var qs = parseQueryString(window.location.search);`

KalleOlaviNiemitalo:
2.58.5 was released, so I think this could now be closed.

PR #7658, opened 2021-10-12
Preserve url query string when using tabs

ivantam-fm:
Fixes issue #7617


ivantam-fm:
@yufeih Do you have plans on merging and releasing this? We are currently blocked by this.

Issue #7689, opened 2021-11-01
Release 2.58.8 zip is corrupt


niederee:
**Operating System**: Windows

**DocFX Version Used**: 2.58.8

**Template used**: NA

**Steps to Reproduce**:

1.  Downloaded zip https://github.com/dotnet/docfx/releases/download/v2.58.8/docfx.zip
2. Right click and extracted it gives error

**Expected Behavior**:  Unzip should work

**Actual Behavior**: Unzip doesn't work

sidneymarie116:
+1. This version of docfx doesn't seem properly packaged for installation with chocolatey. Please notify us when this is fixed.

v-pegao:
Fixed in v2.58.9: https://github.com/dotnet/docfx/releases/download/v2.58.9/docfx.zip

reyang:
v2.58.9 has another issue #7842 which blocks installation.

PR #7690, opened 2021-11-02
use File.ReadAllBytes() to read docfx.zip

v-pegao:
#7689


yufeih:
Paste screenshot to prove the output is correct with this fix.

v-pegao:
> Paste screenshot to prove the output is correct with this fix.

Created [sample docfx.zip](https://github.com/v-pegao/docfx/releases) in forked repo, will be removed after the fix merged
image

yufeih (approved):

PR #7695, opened 2021-11-05
Add back the timeout for TelemetryClient.Flush to avoid deadlock

928PJY:
#512461

### DIFF
--- a/src/docfx/lib/log/Telemetry.cs
+++ b/src/docfx/lib/log/Telemetry.cs
@@ -8,6 +8,7 @@ using System.IO;
 using System.Reflection;
 using System.Runtime.InteropServices;
 using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.ApplicationInsights;
 using Microsoft.ApplicationInsights.DataContracts;
 using Microsoft.ApplicationInsights.Extensibility;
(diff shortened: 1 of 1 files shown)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -233,7 +234,8 @@ public static void TrackException(Exception ex)
 
         public static void Flush()
         {
-            s_telemetryClient.Flush();
+            // TelemetryClient.Flush may meet deadlocks: https://github.com/microsoft/ApplicationInsights-dotnet/issues/1186
+            Task.WaitAny(Task.Run(s_telemetryClient.Flush), Task.Delay(2000));
```
yufeih, line 238: We can increase the timeout a little bit if 2 seconds is to small to flush all the telemetry.

yufeih (approved):

src/docfx/lib/log/Telemetry.cs
928PJY: like 10 seconds?

yufeih: Sounds good

superyyrrzz (approved):

superyyrrzz:
Does this change need to be reverted? https://github.com/928PJY/docfx/commit/de7590b1e6b647243045f1f57fdd18f9cad145fb#diff-3dc856ced277d6fc688a197540a9b355eca4dca734689ae41a18b342523902dd

Issue #3154, opened 2018-08-04
Internal interface is displayed in API documentation when implemented by a public class


airbreather:
**DocFX Version Used**: docfx.console 2.37.2

**Template used**: `default`

**Steps to Reproduce**:

1. Write a .NET project with a public class that implements an internal interface.
1. Generate API documentation for that project.
1. Browse the API documentation.

**Expected Behavior**:

Syntax for the class, generated near the top in the default template, shows only the public interfaces that implements, since nobody browsing the API documentation should care about what internal interfaces it happens to implement.

**Actual Behavior**:

Syntax for the class shows the internal interface.

**Live Sample**:

https://nettopologysuite.github.io/NetTopologySuite.IO.GPX/api/NetTopologySuite.IO.GpxBoundingBox.html

- "View Source" link works, though it doesn't permalink, so [here's](https://github.com/NetTopologySuite/NetTopologySuite.IO.GPX/blob/6a6063ae5ed35005ee2800ff3d88ffa124ae3ed0/src/NetTopologySuite.IO.GPX/GpxBoundingBox.cs#L13) a permalink for now.
- Source code for that version of `ICanWriteToXmlWriter` is [here](https://github.com/NetTopologySuite/NetTopologySuite.IO.GPX/blob/6a6063ae5ed35005ee2800ff3d88ffa124ae3ed0/src/NetTopologySuite.IO.GPX/ICanWriteToXmlWriter.cs), note that it's internal.

airbreather:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed in a week if no further activity occurs.

Thank you, friendly bot, for doing your part to help keep repositories clean.

This still appears to be an issue in the latest stable release (2.40.12).

superyyrrzz:
It sounds make sense. Mark this as a feature request to adjust API display access level.

mkozlina:
I am also having the same problem and would be grateful if you could include this into a next release.

dotMorten:
I'm hitting this as well, and creates major bloat and confusion in the API reference. Is there a workaround we can apply at least? I couldn't find a way in filters to exclude by accessibility.

dotMorten:
I wonder if this statement just needs something like this added:
`where m.DeclaredAccessibility != Accessibility.Internal && m.DeclaredAccessibility != Accessibility.Private`

[...]

PR #7696, opened 2021-11-05
Only show public interfaces in generated syntax

dotMorten:
This addresses the issue with the generated syntax showing internal interfaces, as described in #3154
The changed method is only used to generate syntax, so should be safe. Since interfaces can only be public or internal, there's no need to check for any of the other accessibilities.


KalleOlaviNiemitalo:
> interfaces can only be public or internal

Not true

```C#
public class Outer
{
    private protected interface IPrivateProtected
    {
    }

    public class Inner : IPrivateProtected
    {
    }
}
```

dotMorten:
@KalleOlaviNiemitalo From what I can tell, the effect is the same though. We still don't want to doc that interface

KalleOlaviNiemitalo:
You'd want to doc protected interfaces though. Does this PR handle those right?

dotMorten:
> Does this PR handle those right?

It does now ;-)

KalleOlaviNiemitalo:
How about this case:

```C#
internal class Outer
{
    public interface INested
    {
    }
}

public class Impl : Outer.INested
{
}
```

Here, I think you wouldn't want to mention Outer.INested in the docs, even though it is public itself.

dotMorten:
@KalleOlaviNiemitalo Ugh I'm starting to not like you very much :-)   J/K

Fixed this case now.

yufeih (approved):

yufeih:
Thanks @dotMorten for the fix and @KalleOlaviNiemitalo for the excellent code review!

PR #7703, opened 2021-11-10
change HierarchyDrySync request retry policy

fkpwolf:
[AB#513212](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/513212)

As to HierarchyDrySync request,
1. not retry on timeout
2. timeout = 300s

### DIFF
--- a/src/docfx/config/ops/OpsAccessor.cs
+++ b/src/docfx/config/ops/OpsAccessor.cs
@@ -3,7 +3,6 @@
 
 using System;
 using System.IO;
-using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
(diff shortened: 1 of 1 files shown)


src/docfx/config/ops/OpsAccessor.cs
```diff
@@ -244,6 +245,46 @@ async Task<HttpResponseMessage> SendRequest(HttpRequestMessage request)
             }
         }
 
+        // timeout is 300s and no retry on timeout
+        private async Task<string> LongFetch(Func<HttpRequestMessage> requestFactory, string? value404 = null, HttpMiddleware? middleware = null)
```
yufeih, line 249: instead of copying this method, dd an optional `HttpClient httpClient = null` parameter to the `Fetch` method and use it over `_http` if present.

yufeih: We could disable retry entirely for `OnDemandHierarchyDrySync`

fkpwolf: ok

yufeih (approved):

PR #7709, opened 2021-11-15
Crash fixes for CopyInherited

Sorpirit:
### Found two bug in the CopyInherited class.

**Operating System**: Windows
**DocFX Version Used**: 2.58.9
**Template used**: default
### 1. NullReferenceException exception crash
**Steps to Reproduce**:
1. Try to generate <inheritdoc /> documentation for constructor of the class without inherited documentation for constructor
```
class A : IDispoable
{
   /// <inheritdoc /> 
   public A() { }
   
   /// <inheritdoc /> 
   public Dispose() { }
}
```
**Expected Behavior**:
Inheritdoc will be ignored(hopefully warning would be thrown).
**Actual Behavior**:
Program crashes with NullReferenceException.
**Suggesting fix**:
Add null check(line 45) before getting last item from Inheritance list(line 49)

### 2. StackOverflow crash
**Steps to Reproduce**:
1. Try to generate <inheritdoc /> documentation for IEnumerator when inheriting both from IEnumerator and IEnumerable.
```
class A : IEnumerator<Other>, IEnumerable<Other>
{
   ...
   /// <inheritdoc />
   IEnumerator<Other> IEnumerable<Other>.GetEnumerator()
   ...

   ...
   /// <inheritdoc />
   IEnumerator IEnumerable.GetEnumerator()
   ...
}
```
**Expected Behavior**:
Documentation would be successfully generated.
**Actual Behavior**:
Program crashes with StackOverflow.
**Suggesting fix**:
Add check(line 131) when copying from src to dest if it is the same in order to stop recursion.


yufeih (commented):
Thank you for providing this fix! The fix looks good to me, but to verify the fix and to prevent future regressions, could you help add some test cases to cover the two scenarios in PR description? #7696 contains some examples to add a test case.

yufeih (approved):

PR #7713, opened 2021-11-16
use DateParseHandling.None for pageMetadata to keep raw date string

v-pegao:
[AB#491479](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/491479)

Input date string: `2021-09-16T12:07:50.2667174Z`
Current date string in pageMetadata: `<meta name=\"loc_version\" content=\"9/16/2012 12:07:50 PM\" />`
Date string in pageMetadata with this change: `<meta name=\"loc_version\" content=\"2021-09-16T12:07:50.2667174Z\" />`

### DIFF
--- a/docs/specs/markdown.yml
+++ b/docs/specs/markdown.yml
@@ -851,3 +851,22 @@ outputs:
       "conceptual": "<p>Check <span class=\"no-loc\"  dir=\"ltr\" lang=\"en-us\">Find</span> for more details.\nThis method can be ran on Windows, and you can also used it in Azure SDK.</p>\n",
       "no-loc": ["method", "SDK"]
     }
+---
+# Keep raw date string
+inputs:
+  docfx.yml: |
+    outputType: pageJson
+  docs/a.md: |
+    ---
+    loc_version: "2021-09-16T12:07:50.2667174Z"
+    other_date_string: "9/16/2012 12:07:50 PM"
+    ---
+    (2021-09-16T12:07:50.2667174Z)Date string should not be parsed by Json.Net.
+outputs:
+  docs/a.raw.page.json: |
+    {
+      "rawMetadata": { "loc_version": "2021-09-16T12:07:50.2667174Z", "other_date_string": "9/16/2012 12:07:50 PM", "conceptual": "<p>(2021-09-16T12:07:50.2667174Z)Date string should not be parsed by Json.Net.</p>\r\n" },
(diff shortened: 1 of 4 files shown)


src/docfx/lib/HtmlUtility.cs
```diff
@@ -319,7 +319,24 @@ public static string CreateHtmlMetaTags(JObject metadata)
...
+                    switch (value.Type)
+                    {
+                        case JTokenType.Boolean:
+                        {
+                            content = (bool)value ? "true" : "false";
+                            break;
+                        }
+                        case JTokenType.Date:
+                        {
+                            content = JsonUtility.Serialize(value).Trim('"');
```
yufeih, line 331: This would still not preserve the original string, it just reformats the date that happened to match loc inputs. The fix should disable JSON.NET automatic Date parsing in the first place.

v-pegao: Updated

yufeih (commented):
Add a YAML spec test that illustrates the bug

docs/specs/markdown.yml
```diff
@@ -851,3 +851,22 @@ outputs:
       "conceptual": "<p>Check <span class=\"no-loc\"  dir=\"ltr\" lang=\"en-us\">Find</span> for more details.\nThis method can be ran on Windows, and you can also used it in Azure SDK.</p>\n",
       "no-loc": ["method", "SDK"]
     }
+---
+# Keep raw date string
```
yufeih, line 855: Move this test to `metadata.yml`.

docs/specs/markdown.yml
```diff
@@ -851,3 +851,22 @@ outputs:
...
+  docfx.yml: |
+    outputType: pageJson
+  docs/a.md: |
+    ---
+    loc_version: "2021-09-16T12:07:50.2667174Z"
+    other_date_string: "9/16/2012 12:07:50 PM"
+    ---
+    (2021-09-16T12:07:50.2667174Z)Date string should not be parsed by Json.Net.
+outputs:
+  docs/a.raw.page.json: |
```
yufeih, line 866: Is checking `a.mta.json` sufficient? Remove checking `.raw.page.json` for brevity.

yufeih (approved):

docs/specs/markdown.yml
v-pegao: No, the test is mainly to test the `pageMetadata`(a html string that built by docfx) in `.raw.page.json`. When build the html string, we will directly call `JValue.ToString()` for `Date` without the `DateParseHandling.None`, we need make sure the date is not get parsed with this test.

We can remove check for `a.mta.json` and `rawMetadata` check in `.raw.page.json` for brevity, because we are calling serializer with `DateParseHandling.None` to write them, they are currently safe.

@yufeih Let me know if we want to remove them(`a.mta.json`, `rawMetadata` in `.raw.page.json` )

yufeih: Yes, remove `a.raw.page.json/rawMetadata` and `a.mta.json/conceptual`, format `pageMetadata` as multilines like  https://github.com/dotnet/docfx/blob/v3/docs/specs/legacy.yml#L65

PR #7720, opened 2021-11-22
Avoid multiple execution for concurrency lazy initialization

live1206:
[AB#503787](https://dev.azure.com/ceapex/Engineering/_workitems/edit/503787)

### DIFF
--- a/src/docfx/restore/FileResolver.cs
+++ b/src/docfx/restore/FileResolver.cs
@@ -8,6 +8,7 @@ using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Text;
+using System.Threading;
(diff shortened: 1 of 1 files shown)


yufeih (changes requested):
The default behavior of `Lazy` is `ExecutionAndPublication`, it must be something else.

mingwli:
> The default behavior of `Lazy` is `ExecutionAndPublication`, it must be something else.

How can this be proved?

yufeih:
> > The default behavior of `Lazy` is `ExecutionAndPublication`, it must be something else.
> 
> How can this be proved?

https://source.dot.net/#System.Private.CoreLib/Lazy.cs,208

PR #7727, opened 2021-11-23
Upgrade to .NET 6

yufeih:
Check to see if anything breaks (It shouldn't).

### DIFF
--- a/src/Microsoft.Docs.Build.Specialized/views/Template/BuildItemListPartial.cshtml
+++ b/src/Microsoft.Docs.Build.Specialized/views/Template/BuildItemListPartial.cshtml
@@ -36,7 +36,7 @@
     }
 }
 
-@if (@Model.Style == "unordered")
+@if (@Model?.Style == "unordered")
 {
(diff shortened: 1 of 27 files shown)


src/Microsoft.Docs.Build.Specialized/views/Template/BuildItemListPartial.cshtml
```diff
@@ -36,7 +36,7 @@
     }
 }
 
-@if (@Model.Style == "unordered")
+@if (@Model?.Style == "unordered")
```
mingwli, line 39: 🙄

mingwli (approved):

src/Microsoft.Docs.Build.Specialized/views/Template/BuildItemListPartial.cshtml
yufeih: yes, there are new `null` annotations that give us new warnings, which is good.

PR #7733, opened 2021-11-28
Fixed copy-paste typo

klacol:
This line seems to be copied from the paragraph above. I have changed it. 

For me, the situation was different, though.

 My `main.css` is empty. But I see an `docfx.css` which i can us, to customize the style of the website. Perhaps this could be outlined in the docs as well.


yufeih (approved):
Thanks for the fix!

PR #7743, opened 2021-12-03
collect telemtry of HTML

shanluo911:
Collect the HTML telemetry. [AB#522386](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/522386)
Tech design: https://microsoft-my.sharepoint.com/:w:/p/shanluo/EZvl-6IB_AFDk5onJzNqXj4B2TyxnawcpErGSfpa4c4b2Q?e=gQuaNw

Test:
Send the telemetry to my personal AI.
image

### DIFF
--- a/src/docfx/lib/HtmlSanitizer.cs
+++ b/src/docfx/lib/HtmlSanitizer.cs
@@ -197,8 +197,14 @@ internal class HtmlSanitizer
         }
     }
 
-    public bool IsAllowedHtml(string tokenName, string? attributeName)
+    public bool IsAllowedHtml(string tokenName, string attributeName)
     {
+        if (attributeName.StartsWith("aria-", StringComparison.OrdinalIgnoreCase) ||
+            attributeName.StartsWith("data-", StringComparison.OrdinalIgnoreCase))
+        {
+            return true;
+        }
+
         if (_allowedHtml.TryGetValue(tokenName, out var allowedGlobalAttributes))
         {
             return string.IsNullOrEmpty(attributeName) || (allowedGlobalAttributes != null && allowedGlobalAttributes.Contains(attributeName));
--- a/src/docfx/lib/HtmlSanitizer.cs
+++ b/src/docfx/lib/HtmlSanitizer.cs
@@ -166,48 +166,42 @@ internal class HtmlSanitizer
         foreach (ref var attribute in token.Attributes.Span)
         {
             var attributeName = attribute.Name.ToString();
-            if (!IsAllowedAttribute(attributeName))
+            if (!IsAllowedAttribute(attributeName, allowedAttributes))
             {
                 errors.Add(Errors.Content.DisallowedHtmlAttribute(obj?.GetSourceInfo()?.WithOffset(attribute.NameRange), tokenName, attributeName));
                 attribute = default;
             }
         }
-
-        bool IsAllowedAttribute(string attributeName)
-        {
-            if (_allowedHtml.TryGetValue("*", out var allowedGlobalAttributes)
-                && allowedGlobalAttributes != null
-                && allowedGlobalAttributes.Contains(attributeName))
-            {
-                return true;
-            }
-
-            if (allowedAttributes != null && allowedAttributes.Contains(attributeName))
-            {
-                return true;
-            }
-
-            if (attributeName.StartsWith("aria-", StringComparison.OrdinalIgnoreCase) ||
-                attributeName.StartsWith("data-", StringComparison.OrdinalIgnoreCase))
-            {
-                return true;
-            }
-
-            return false;
-        }
     }
 
     public bool IsAllowedHtml(string tokenName, string attributeName)
     {
-        if (attributeName.StartsWith("aria-", StringComparison.OrdinalIgnoreCase) ||
-            attributeName.StartsWith("data-", StringComparison.OrdinalIgnoreCase))
+        if (!_allowedHtml.TryGetValue(tokenName, out var allowedAttributes))
+        {
+            return false;
+        }
+
+        return string.IsNullOrEmpty(attributeName) || IsAllowedAttribute(attributeName, allowedAttributes);
+    }
+
+    private bool IsAllowedAttribute(string attributeName, HashSet<string>? allowedAttributes)
+    {
+        if (_allowedHtml.TryGetValue("*", out var allowedGlobalAttributes)
+            && allowedGlobalAttributes != null
+            && allowedGlobalAttributes.Contains(attributeName))
         {
             return true;
         }
 
-        if (_allowedHtml.TryGetValue(tokenName, out var allowedGlobalAttributes))
+        if (allowedAttributes != null && allowedAttributes.Contains(attributeName))
         {
-            return string.IsNullOrEmpty(attributeName) || (allowedGlobalAttributes != null && allowedGlobalAttributes.Contains(attributeName));
+            return true;
+        }
+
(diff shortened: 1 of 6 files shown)


src/docfx/lib/markdown/HtmlExtension.cs
```diff
@@ -52,7 +68,7 @@ string ProcessHtml(string html, MarkdownObject block, ErrorBuilder errors)
                 HtmlUtility.TransformLink(ref token, block, getLink!);
```
yufeih, line 68: We need a dedicated `HtmlUtility.CountHtmlUsage(string html, Dictionary<(string element, string? attribute)>)` method and run it in line 62 before `HtmlUtility.TransformHtml` to ensure the usage collected is true. E.g., running it inside `SanitizeHtml` would ignore HTML removed by `RemoveRerunCodepenIframes` method.

Also add a `HtmlSantizer.IsAllowed(string element, string? attribute)` method to determine if an element is allowed.

src/docfx/lib/HtmlUtility.cs
```diff
@@ -325,6 +325,44 @@ public static string CreateHtmlMetaTags(JObject metadata)
...
+            var tokenName = token.Name.ToString();
+            if (!elementCount.ContainsKey(tokenName))
+            {
+                elementCount.Add(tokenName, new Dictionary<string, int>());
+            }
+
+            var attributeCount = elementCount[tokenName];
+            if (token.Attributes.Span.IsEmpty)
+            {
+                attributeCount[string.Empty] = attributeCount.GetValueOrDefault(string.Empty, 0) + 1;
```
yufeih, line 353: Use `CollectionsMarshal` to same dictionary lookups. Same in line 360.
```suggestion
                CollectionsMarshal.GetValueRefOrAddDefault(attributeCount, "", out _)++;
```

test/docfx.Test/lib/HtmlUtilityTest.cs
```diff
@@ -163,4 +163,46 @@ public static void Encode(string input, string expected)
...
+
+    [Theory]
+    [InlineData("<div style='a'></div>", "{\"div\":{\"style\":1}}")]
+    [InlineData("<div><script></script></div>", "{\"div\": {\"\": 1},\"script\": {\"\": 1}}")]
+    public static void CollectHtmlUsage(string input, string expected)
+    {
+        var expectedCount = JsonUtility.DeserializeData<Dictionary<string, Dictionary<string, int>>>(expected, null);
+        var elementCount = new Dictionary<string, Dictionary<string, int>>();
+        HtmlUtility.CollectHtmlUsage(input, elementCount);
+        Assert.True(CompareDictionaries(expectedCount, elementCount));
```
yufeih, line 175: Instead of write a custom compare dictionary, simply compare the serialized JSON.
```suggestion
        Assert.Equal(expected, JsonSerializer.Serialize(elementCount));
```

src/docfx/lib/HtmlSanitizer.cs
```diff
@@ -196,4 +196,14 @@ bool IsAllowedAttribute(string attributeName)
             return false;
         }
     }
+
+    public bool IsAllowedHtml(string tokenName, string? attributeName)
```
herohua, line 200: It seems there are additional logics on allowed attributes (see line 190 to 194). Should we reuse the same logic there?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -325,6 +325,44 @@ public static string CreateHtmlMetaTags(JObject metadata)
...
+
+        static void CollectHtmlUsageCore(ref HtmlToken token, Dictionary<string, Dictionary<string, int>> elementCount)
+        {
+            if (token.Type != HtmlTokenType.StartTag)
+            {
+                return;
+            }
+
+            var tokenName = token.Name.ToString();
+            if (!elementCount.ContainsKey(tokenName))
```
herohua, line 345: use `TryGetValue`

test/docfx.Test/lib/HtmlUtilityTest.cs
KalleOlaviNiemitalo: Does JsonSerializer.Serialize sort the key-value pairs by key?

yufeih: It doesn't sort, but the result should be deterministic. Most likely it'll use dictionary enumeration order which is dictionary insertion order which is deterministic per input HTML.

src/docfx/lib/HtmlUtility.cs
shanluo911: I thought about using `TryGetValue` here, but if the dictionary does not have contain the key, we still need to add (key, dictionary<>) pair.

herohua: But you can still save one access to the dictionary if the key exists.

yufeih: Maybe we could have a helper like this: https://github.com/dotnet/runtime/issues/15059#issuecomment-954708195

test/docfx.Test/lib/HtmlUtilityTest.cs
```diff
@@ -163,4 +163,14 @@ public static void Encode(string input, string expected)
     {
         Assert.Equal(expected, HtmlUtility.Encode(input));
     }
+
+    [Theory]
+    [InlineData("<div style='a'></div>", "{\"div\":{\"style\":1}}")]
```
herohua, line 168: adding another case with multiple attributes?

yufeih (approved):

src/docfx/lib/HtmlSanitizer.cs
```diff
@@ -196,4 +196,20 @@ bool IsAllowedAttribute(string attributeName)
             return false;
         }
     }
+
+    public bool IsAllowedHtml(string tokenName, string attributeName)
+    {
+        if (attributeName.StartsWith("aria-", StringComparison.OrdinalIgnoreCase) ||
```
yufeih, line 202: Can we share the `IsAllowedAttribute` method?

src/docfx/lib/HtmlUtility.cs
```diff
@@ -325,6 +326,44 @@ public static string CreateHtmlMetaTags(JObject metadata)
...
+    public static void CollectHtmlUsage(string html, Dictionary<string, Dictionary<string, int>> elementCount)
+    {
+        var reader = new HtmlReader(html);
+
+        while (reader.Read(out var token))
+        {
+            CollectHtmlUsageCore(ref token, elementCount);
+        }
+
+        static void CollectHtmlUsageCore(ref HtmlToken token, Dictionary<string, Dictionary<string, int>> elementCount)
```
928PJY, line 338: Remove this nested function or move it outside?

928PJY (approved):

test/docfx.Test/lib/HtmlUtilityTest.cs
KalleOlaviNiemitalo: .NET has randomized string hashing but Dictionary\<TKey, TValue\> disables that until it gets more than 100 hash collisions during the same insertion, so it is still deterministic with two elements.

- <https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L62-L72>
- <https://github.com/dotnet/runtime/blob/v6.0.0/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/Dictionary.cs#L676-L682>

yufeih (approved):

PR #7749, opened 2021-12-09
change UidNotFound error level

fkpwolf:
[AB#530134](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/530134)

"due to the decentralized nature of how content and certifications get created, could we just add a warning"
https://dev.azure.com/ceapex/Engineering/_boards/board/t/Docs%20Content%20Hosting%20and%20Storage%20Team/Stories/?workitem=328267#8939105

If a Module is referenced by another repo, with this PR, error level will become Warning from Error, since Module also uses `validateExternalXrefs` feature.

And I also move UidNotFound to Link category since it has no property path so Validation Build rules can't override on it (propertyPath is mandatory input in Rule edit form).

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -267,6 +267,9 @@ internal static class Errors
         /// Behavior: ✔️ Message: ✔️
         public static Error FileNotFound(SourceInfo<string> source)
             => new(ErrorLevel.Warning, "file-not-found", $"Invalid file link: '{source}'.", source);
+
+        public static Error UidNotFound(string uid, IEnumerable<string?> repositories, string? schemaType)
+            => new(ErrorLevel.Warning, "uid-not-found", $"UID '{uid}' with type '{schemaType}' not found, which is referenced by repository {StringUtility.Join(repositories)}.", null);
     }
 
     public static class UrlPath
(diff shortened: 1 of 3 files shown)


yufeih (approved):

yufeih:
@mingwli The regression test cases all failed

mingwli:
> @mingwli The regression test cases all failed

AB#518489

herohua:
adding several people to review this change

src/docfx/Errors.cs
```diff
@@ -267,6 +267,9 @@ public static Error CircularReference<T>(SourceInfo? source, T current, IEnumera
         /// Behavior: ✔️ Message: ✔️
         public static Error FileNotFound(SourceInfo<string> source)
             => new(ErrorLevel.Warning, "file-not-found", $"Invalid file link: '{source}'.", source);
+
+        public static Error UidNotFound(string uid, IEnumerable<string?> repositories, string? schemaType)
```
herohua, line 271: Would it be a better name of `ExternalUidNotFound`?

src/docfx/Errors.cs
```diff
@@ -421,9 +424,6 @@ public static Error XrefNotFound(SourceInfo<string> source)
         public static Error XrefTypeInvalid(SourceInfo<string> xref, string expectedXrefType, string? actualXrefType)
            => new(ErrorLevel.Warning, "xref-type-invalid", $"Invalid cross reference: '{xref}'. Expected type '{expectedXrefType}' but got '{actualXrefType}'.", xref);
 
-        public static Error UidNotFound(string uid, IEnumerable<string?> repositories, string? schemaType)
```
herohua, line 424: I'd prefer keep it in class `Xref`

herohua (changes requested):
This change seems will break learning path/module check - is this by design?
https://dev.azure.com/ceapex/Engineering/_git/docs-ui?path=/packages/schemas/src/LearningPath.schema.json&version=GBdevelop&line=135&lineEnd=136&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

fkpwolf:
close this PR and replace with `propertyPath` approach.

PR #6431, opened 2020-08-21
Reference metadata validate

fkpwolf:
[AB#288366](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/288366)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/dotnet/docfx/pull/6431)

PR #7750, opened 2021-12-10
Fix none breaking whitespace chars

yufeih:
Introduced in https://github.com/dotnet/docfx/pull/6431/files

@fkpwolf Please be extremely careful and avoid copy and pasting any invisible characters into the code. This type of bug is extremely difficult to diagnose.

cc @OsmondJiang


KalleOlaviNiemitalo:
Could add a validation step that fails if any unusual control characters sneak into the DocFX source tree. That would further mitigate <https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-42574> as well.

live1206 (approved):

PR #7752, opened 2021-12-13
Optimize memory usage by not capturing SchemaMap in xref lazy callback.

yufeih:

The `SchemaMap` is a big object that consumes a lot of memory. It maps `JToken` to a schema. The current code captures the `SchemaMap` in the `xrefProperties` lazy callback function, causing the `SchemaMap` to be pinned by that lazy callback, thus evicting entries from `_schemaDocumentsCache` on low memory conditions does not deallocate the big schema map. 

This fix _uncaptures_ the schema map lambda capture, allowing us to deallocate the schema map on low memory conditions.

As part of the investigation, I'm dumping more `GC` related info as part of memory monitor.

The results shows we can limit memory usage < 10G for _azure-docs-sdk-java_ at xref map build stage, whereas before it was > 15G. 

It was too slow to build the TOC map, so I couldn't capture the memory past that stage. I may need to improve TOC map build first as a next step.

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -119,6 +119,7 @@ internal class JsonSchemaTransformer
         var schemaMap = new JsonSchemaMap(IsContentTransform);
         var schemaErrors = schemaValidator.Validate(token, file, schemaMap);
         errors.AddRange(schemaErrors);
+        schemaMap.TrimExcess();
 
         var uidCount = GetFileUidCount(schemaMap, token);
         return (token, schemaValidator.Schema, schemaMap, uidCount);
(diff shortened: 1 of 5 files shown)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -306,6 +304,7 @@ private string GetXrefHref(FilePath file, string uid, int uidCount, bool isRootL
         try
         {
             recursionDetector.Push(uid);
+            var (token, schema, schemaMap, uidCount) = ValidateContent(ErrorBuilder.Null, file);
```
live1206, line 307:
```suggestion
            var (_, _, schemaMap, uidCount) = ValidateContent(ErrorBuilder.Null, file);
```

src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -119,6 +119,7 @@ public IReadOnlyList<InternalXrefSpec> LoadXrefSpecs(ErrorBuilder errors, FilePa
         var schemaMap = new JsonSchemaMap(IsContentTransform);
         var schemaErrors = schemaValidator.Validate(token, file, schemaMap);
         errors.AddRange(schemaErrors);
+        schemaMap.TrimExcess();
```
live1206, line 122: Why is this needed here?

PR #7755, opened 2021-12-14
Optimize TOC reference buildup

yufeih:

This is to optimize the first stage of non-parallel toc map build up.

This is part 1. Part 2 is to fix the O(n^2) conflict detection algorithm.

### DIFF
--- a/src/docfx/build/toc/TocMap.cs
+++ b/src/docfx/build/toc/TocMap.cs
@@ -157,7 +157,7 @@ internal class TocMap
         using var scope = Progress.Start("Loading TOC");
 
         var allTocFiles = new ConcurrentBag<FilePath>();
-        var allTocs = new List<(FilePath file, HashSet<FilePath> docs, HashSet<FilePath> tocs, bool hasFallbackToc)>();
+        var allTocs = new List<(FilePath file, HashSet<FilePath> docs, HashSet<FilePath> tocs, bool shouldBuildFile)>();
         var includedTocs = new HashSet<FilePath>();
         var allServicePages = new List<FilePath>();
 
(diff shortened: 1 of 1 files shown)


src/docfx/build/toc/TocMap.cs
```diff
@@ -152,71 +152,80 @@ private static (int subDirectoryCount, int parentDirectoryCount) GetRelativeDire
...
-            if (tocToTocs.TryGetValue(file, out var tocReferences) && tocReferences.Any(toc => toc.Origin != FileOrigin.Fallback))
+            if (!tocToTocs.TryGetValue(file, out var value))
             {
-                return true;
+                return false;
             }
 
-            return false;
+            // if A toc includes B toc and only B toc is localized, then A need to be included and built
+            return !value.hasFallbackToc;
```
yufeih, line 228: Precompute fallback logic to get rid of `Any` check inside a hot loop.

src/docfx/build/toc/TocLoader.cs
```diff
@@ -77,15 +77,15 @@ public static TocHrefType GetHrefType(string? href)
         return TocHrefType.RelativeFile;
     }
 
-    public (TocNode node, List<FilePath> referencedFiles, List<FilePath> referencedTocs, List<FilePath> servicePages) Load(FilePath file)
+    public (TocNode node, HashSet<FilePath> referencedFiles, HashSet<FilePath> referencedTocs, List<FilePath> servicePages) Load(FilePath file)
```
yufeih, line 80: Turn `List` to `HashSet` to get rid of `Distinct` calls in toc map builder.

herohua (approved):

src/docfx/build/toc/TocMap.cs
```diff
@@ -152,47 +152,46 @@ private static (int subDirectoryCount, int parentDirectoryCount) GetRelativeDire
...
+        ParallelUtility.ForEach(scope, _errors, allTocFiles, file =>
         {
-            var (_, referencedDocuments, referencedTocs, servicePages) = _tocLoader.Load(file);
-
-            tocReferences.TryAdd(file, (referencedDocuments, referencedTocs));
+            var (_, docs, tocs, servicePages) = _tocLoader.Load(file);
+            var shouldBuildFile = tocs.Any(toc => toc.Origin != FileOrigin.Fallback);
 
-            foreach (var servicePage in servicePages)
+            lock (allTocs)
```
live1206, line 177: Can we leverage [ListBuilder](https://github.com/dotnet/docfx/blob/v3/src/docfx/lib/collections/ListBuilder.cs) here?

yufeih: That will lock/unlock 3 times but here we only need to lock/unlock once.

AngryBerryMS (approved):

PR #7756, opened 2021-12-14
Optimize service page publish url conflict detection

yufeih:

This is to remove the O(n^2) conflict detection algorithm. Depend on #7755

### DIFF
--- a/docs/specs/toc/service-page.yml
+++ b/docs/specs/toc/service-page.yml
@@ -718,7 +718,7 @@ outputs:
   a/test/index.json:
   a/api/bot/toc.json:
   .errors.log: |
-    {"message_severity":"warning","code":"output-path-conflict","message":"Two or more files output to the same path 'test/index.json': 'test/index.md', 'test/index.yml [generated]'."}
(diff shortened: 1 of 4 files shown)


docs/specs/toc/service-page.yml
```diff
@@ -789,7 +789,7 @@ outputs:
     {"title": "hello world"}
   a/api/bot/toc.json:
   .errors.log: |
-    {"message_severity":"warning","code":"publish-url-conflict","message":"Two or more files of the same version('v-0.2') publish to the same url '/docs/': 'test/index.yml [generated]<'v-0.2', 'v-0.3'>', 'test1/index.md<'v-0.1', 'v-0.2'>'."}
```
herohua, line 792: Should we keep the original error message as it contains more specific version information?

yufeih: That would require an extra code path to retrieve the monikers. This is a special code path for generated contents (service page) in a corner case, and I feel the message here already conveys the meaning.

AngryBerryMS (approved):

PR #7759, opened 2021-12-14
Add missing locale for absolute path while applying template instead of loading document

live1206:
[AB#508664](https://ceapex.visualstudio.com/Engineering/_workitems/edit/508664)

We should add locale for absolute paths during applying templates, by then we would have the correct locale:
- Conceptual: Move `adding locale` to template engine
- LandingData: Move entire HTML processing to template engine just as SDP (**This might have Xliff files changed during hand-off**)
- SDP: no change

### DIFF
--- a/docs/specs/metadata.yml
+++ b/docs/specs/metadata.yml
@@ -345,6 +345,21 @@ outputs:
   .errors.log: |
     {"message_severity":"warning","code":"yaml-header-syntax-error","file":"docs/a.md","line":2,"column":5}
 ---
+# Input metadata doesn't overwrite system generated metadata
+inputs:
+  docfx.yml:
+  docs/a.md: |
+    ---
+    wordCount: 15
+    document_id: 123
+    ---
+outputs:
+  docs/a.json: |
+    { "wordCount": 0, "document_id": "3862ef10-61c4-6242-7188-36960566ecb6" }
+  .errors.log: |
+    {"message_severity":"warning","code":"attribute-reserved","message":"Attribute wordCount is reserved for use by Docs.","file":"docs/a.md","line":2,"column":1}
+    {"message_severity":"warning","code":"attribute-reserved","message":"Attribute document_id is reserved for use by Docs.","file":"docs/a.md","line":3,"column":1}
+---
 # Title in file metadata should overwrite H1
 inputs:
   docfx.yml: |
--- a/docs/specs/metadata.yml
+++ b/docs/specs/metadata.yml
@@ -345,21 +345,6 @@ outputs:
   .errors.log: |
     {"message_severity":"warning","code":"yaml-header-syntax-error","file":"docs/a.md","line":2,"column":5}
 ---
-# Input metadata doesn't overwrite system generated metadata
-inputs:
-  docfx.yml:
-  docs/a.md: |
-    ---
-    wordCount: 15
-    document_id: 123
-    ---
-outputs:
-  docs/a.json: |
-    { "wordCount": 0, "document_id": "3862ef10-61c4-6242-7188-36960566ecb6" }
-  .errors.log: |
-    {"message_severity":"warning","code":"attribute-reserved","message":"Attribute wordCount is reserved for use by Docs.","file":"docs/a.md","line":2,"column":1}
-    {"message_severity":"warning","code":"attribute-reserved","message":"Attribute document_id is reserved for use by Docs.","file":"docs/a.md","line":3,"column":1}
----
 # Title in file metadata should overwrite H1
 inputs:
   docfx.yml: |
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -176,7 +176,7 @@ inputs:
   a.md: Link to [absolute path](/a)
 outputs:
   a.json: |
-    { "conceptual": "<p>Link to <a href=\"/en-us/a\">absolute path</a></p>" }
+    { "conceptual": "<p>Link to <a href=\"/a\">absolute path</a></p>" }
 ---
 # Should not append locale for absolute path starting with api, _themes
 inputs:
@@ -194,7 +194,7 @@ outputs:
         "<p>
             Link to <a href=\"/api/search/rss?search=SDK%20release%20notes\">absolute path</a>
             Link to <a href=\"/API/search/rss?search=SDK%20release%20notes\">absolute path</a>
-            Link to <a href=\"/en-us/api-management/search/rss?search=SDK%20release%20notes\">absolute path</a>
+            Link to <a href=\"/api-management/search/rss?search=SDK%20release%20notes\">absolute path</a>
          </p>
          <iframe src=\"/_themes/docs.theme/master/en-us/_themes/global/video-embed.html\"></iframe>"
     }
--- a/docs/specs/resolve.yml
+++ b/docs/specs/resolve.yml
@@ -170,35 +170,6 @@ outputs:
     { "conceptual": "<p><a href=\"c\">c</a></p>" }
   docs/dir/c.json:
 ---
-# Add locale for absolute links
-inputs:
-  docfx.yml:
-  a.md: Link to [absolute path](/a)
-outputs:
-  a.json: |
-    { "conceptual": "<p>Link to <a href=\"/en-us/a\">absolute path</a></p>" }
----
-# Should not append locale for absolute path starting with api, _themes
-inputs:
-  docfx.yml: |
-    hostName: docs.microsoft.com
-    removeHostName: true
-  a.md: |
-    Link to [absolute path](/api/search/rss?search=SDK%20release%20notes)
-    Link to [absolute path](/API/search/rss?search=SDK%20release%20notes)
-    Link to [absolute path](/api-management/search/rss?search=SDK%20release%20notes)
-    <iframe src="https://docs.microsoft.com/_themes/docs.theme/master/en-us/_themes/global/video-embed.html"></iframe>
-outputs:
-  a.json: |
-    { "conceptual": 
-        "<p>
-            Link to <a href=\"/api/search/rss?search=SDK%20release%20notes\">absolute path</a>
-            Link to <a href=\"/API/search/rss?search=SDK%20release%20notes\">absolute path</a>
-            Link to <a href=\"/en-us/api-management/search/rss?search=SDK%20release%20notes\">absolute path</a>
-         </p>
-         <iframe src=\"/_themes/docs.theme/master/en-us/_themes/global/video-embed.html\"></iframe>"
-    }
----
 # Show warning if the link is missing
 inputs:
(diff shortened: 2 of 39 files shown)


docs/specs/xref.yml
```diff
@@ -43,7 +43,7 @@ inputs:
     { "references":[{"uid": "a", "name": "a", "fullName": "a", "href": "https://docs.microsoft.com/a", "nameWithType": "a"}]}
 outputs:
   docs/a.json: |
-    { "conceptual": "<p>Link to <a href=\"/en-us/dotnet/api/system.string\">System.String</a></p>\n"}
+    { "conceptual": "<p>Link to <a href=\"/dotnet/api/system.string\">System.String</a></p>\n"}
 ---
```
herohua, line 47: I am not sure whether I've understood well on this PR, but should we keep the output JSON as unchanged because e2e behavior will still be appending locale.

live1206: The e2e behavior for Docs will be the same, which will be covered by the regression tests, and the output file for Docs should be *.raw.page.json.

For this spec test, this output type is *.json, not pageJson, which have not applied template yet, and we will add locales during the applying template stage.

docs/specs/metadata.yml
```diff
@@ -345,21 +345,6 @@ outputs:
   .errors.log: |
     {"message_severity":"warning","code":"yaml-header-syntax-error","file":"docs/a.md","line":2,"column":5}
 ---
-# Input metadata doesn't overwrite system generated metadata
```
yufeih, line 348: Don't remove this test...

src/docfx/template/TemplateEngine.cs
```diff
@@ -210,6 +217,27 @@ private string CreateContent(FilePath file, string? mime, JObject pageModel)
...
+
+        var result = HtmlUtility.TransformHtml(html, (ref HtmlReader reader, ref HtmlWriter writer, ref HtmlToken token) =>
+        {
+            HtmlUtility.AddLocaleIfMissingForAbsolutePath(ref token, _locale);
+
+            if (token.Type == HtmlTokenType.Text)
+            {
+                if (!_config.DryRun)
+                {
+                    wordCount += WordCount.CountWord(token.RawText.Span);
```
yufeih, line 232: The `WordCount` comment says the logic is shared with loc build. Which does it mean that loc build already has the logic? If so I'd like to keep both the word count patch and the update link locale patch in loc build.

docs/specs/xref.yml
herohua: Got it! Thanks for explanation!

src/docfx/template/TemplateEngine.cs
herohua: If with Wei's PR here, should we consider removing the patch in loc build? That'll make the logic consolidated and avoid future errors.

src/docfx/template/TemplateEngine.cs
```diff
@@ -210,6 +217,27 @@ private string CreateContent(FilePath file, string? mime, JObject pageModel)
...
+                }
+            }
+        });
+
+        conceptual.Conceptual = LocalizationUtility.AddLeftToRightMarker(_cultureInfo, result);
```
herohua, line 237: >AddLeftToRightMarker [](http://example.com/codeflow?start=52&length=20)

Not related to this PR: Is this where we should apply the `<ltr>` to `no-loc`? @AngryBerryMS

src/docfx/build/page/PageBuilder.cs
```diff
@@ -380,7 +375,6 @@ private void ProcessConceptualHtml(ErrorBuilder errors, FilePath file, string ht
         _searchIndexBuilder.SetBody(file, searchText.ToString());
 
         conceptual.Conceptual = LocalizationUtility.AddLeftToRightMarker(_buildOptions.Culture, result);
-        conceptual.WordCount = wordCount;
     }
```
herohua, line 378: Should we remove this line too as I see we are adding it in `TemplateEngine`

src/docfx/template/TemplateEngine.cs
```diff
@@ -195,12 +196,18 @@ public string ProcessHtml(ErrorBuilder errors, FilePath file, string html)
...
     {
-        if (JsonSchemaProvider.IsConceptual(mime) || JsonSchemaProvider.IsLandingData(mime))
+        if (JsonSchemaProvider.IsConceptual(mime))
         {
-            // Conceptual and Landing Data
-            return pageModel.Value<string>("conceptual") ?? "";
+            var conceptual = JsonUtility.ToObject<ConceptualModel>(ErrorBuilder.Null, pageModel)!;
+            ProcessConceptualHtml(pageModel.Value<string>("conceptual") ?? "", conceptual);
+            pageModel = JsonUtility.SortProperties(_config.DryRun ? new JObject() : JsonUtility.ToJObject(conceptual));
+            return conceptual.Conceptual ?? "";
```
herohua, line 206: seems sorting is handled by [https://github.com/dotnet/docfx/blob/a4e88c1a3f974b62d44884f827e6488e2959067d/src/docfx/build/page/PageBuilder.cs#L159-L159](https://github.com/dotnet/docfx/blob/a4e88c1a3f974b62d44884f827e6488e2959067d/src/docfx/build/page/PageBuilder.cs#L159-L159) already

src/docfx/lib/HtmlUtility.cs
```diff
@@ -356,7 +356,12 @@ internal static void RemoveRerunCodepenIframes(ref HtmlToken token)
     }
 
     internal static void AddLinkType(
-        ErrorBuilder errors, FilePath file, ref HtmlToken token, string locale, Dictionary<string, TrustedDomains> trustedDomains)
+        ErrorBuilder errors,
+        FilePath file,
+        ref HtmlToken token,
+        Dictionary<string, TrustedDomains> trustedDomains,
+        string locale = "",
```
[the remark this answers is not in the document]
live1206: That would cost another extra HTML parsing

src/docfx/template/TemplateEngine.cs
live1206: This should be sorted because it will affect the order of page metadata in https://github.com/dotnet/docfx/blob/948f000506002aca42acb0635f25097628ba0d76/src/docfx/template/TemplateEngine.cs#L207.

`pageModel` was sorted before passing in this method before, but it has been changed within this method during this PR.

src/docfx/template/TemplateEngine.cs
live1206: @herohua Yes, we should remove the wordCount logic in loc-build after this PR.

@yufeih Btw, where is the comment?

src/docfx/build/page/PageBuilder.cs
live1206: I think there is no difference to put `AddLeftToRightMarker` here or later in applying template.
Prefer to put it here to keep the behavior less changed.

src/docfx/template/TemplateEngine.cs
live1206: Keep `wordCount` in loading markdown, and recalculate it in loc build.

docs/specs/resolve.yml
```diff
@@ -170,35 +170,6 @@ outputs:
     { "conceptual": "<p><a href=\"c\">c</a></p>" }
   docs/dir/c.json:
 ---
-# Add locale for absolute links
```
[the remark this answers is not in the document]
live1206: Added the tests back

docs/specs/metadata.yml
live1206: Added back

src/docfx/build/page/PageBuilder.cs
live1206: Had an offline discussion with @yufeih, the output before applying temple should be more complete, so we should keep this logic here.

src/docfx/template/TemplateEngine.cs
yufeih: Now we don't need to sort at all here xD

src/docfx/lib/HtmlUtility.cs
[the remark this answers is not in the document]
yufeih: The HTML has already been parsed, it is just another method to mutate an `HtmlToken`.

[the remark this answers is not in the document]
yufeih: It is actually the `AddLocaleIfMissingForAbsolutePath` method you added.

src/docfx/template/TemplateEngine.cs
```diff
@@ -195,12 +196,18 @@ public string ProcessHtml(ErrorBuilder errors, FilePath file, string html)
         return LocalizationUtility.AddLeftToRightMarker(_cultureInfo, result);
     }
 
-    private string CreateContent(FilePath file, string? mime, JObject pageModel)
+    private string CreateContent(FilePath file, string? mime, ref JObject pageModel)
```
yufeih, line 199: This method shouldn't be mutating the `pageModel`.
```suggestion
    private string CreateContent(FilePath file, string? mime, JObject pageModel)
```

src/docfx/template/TemplateEngine.cs
```diff
@@ -195,12 +196,18 @@ public string ProcessHtml(ErrorBuilder errors, FilePath file, string html)
...
     {
-        if (JsonSchemaProvider.IsConceptual(mime) || JsonSchemaProvider.IsLandingData(mime))
+        if (JsonSchemaProvider.IsConceptual(mime))
         {
-            // Conceptual and Landing Data
-            return pageModel.Value<string>("conceptual") ?? "";
+            var conceptual = JsonUtility.ToObject<ConceptualModel>(ErrorBuilder.Null, pageModel)!;
```
yufeih, line 203: We probably don't need to roundtrip the `ConceptualModel` here, this method can return the transformed HTML directly, something like:
```suggestion
            return AddLinkLocale(pageModel.Value<string>("conceptual") ?? "");
```

src/docfx/template/TemplateEngine.cs
live1206: right

src/docfx/template/TemplateEngine.cs
live1206: right

src/docfx/lib/HtmlUtility.cs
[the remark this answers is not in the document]
live1206: got your point

docs/specs/resolve.yml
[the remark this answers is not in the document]
herohua: @live1206 Is there any place that we have coverage for --continue in spec/yaml test?

herohua (approved):
:shipit:

PR #7760, opened 2021-12-14
Reduce JsonSchemaMap memory usage

yufeih:

The `SchemaMap` is a big object that consumes a lot of memory. It maps `JToken` to a schema. The current implementation adds all the `JToken`s to the map as long as it will be transformed, this takes up quite a lot of memory. This optimizes the memory usage by only adding `JToken`s that needs dynamic dispatch (AnyOf/OneOf/IfThenElse).


yufeih:
@live1206 ping.

live1206 (approved):

PR #7769, opened 2021-12-16
[WIP]disable git credential helper prompt

v-pegao:
[AB#376820](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/376820)
image

### DIFF
--- a/src/docfx/lib/git/GitUtility.cs
+++ b/src/docfx/lib/git/GitUtility.cs
@@ -155,7 +155,7 @@ internal static partial class GitUtility
 
         var (http, secret) = GetGitCommandLineConfig(url, secrets);
 
-        ExecuteNonQuery(path, $"{http} -c core.longpaths=true fetch --progress {options} \"{url}\" {refspecs}", secret);
+        ExecuteNonQuery(path, $"{http} -c credential.helper=\"\" -c core.longpaths=true fetch --progress {options} \"{url}\" {refspecs}", secret);
     }
 
     /// <summary>


herohua:
@928PJY @shanluo911 Will this change impact docs-validation VSC extension?

src/docfx/lib/git/GitUtility.cs
```diff
@@ -155,7 +155,7 @@ public static void Fetch(SecretConfig secrets, string path, string url, string r
 
         var (http, secret) = GetGitCommandLineConfig(url, secrets);
 
-        ExecuteNonQuery(path, $"{http} -c core.longpaths=true fetch --progress {options} \"{url}\" {refspecs}", secret);
+        ExecuteNonQuery(path, $"{http} -c credential.helper=\"\" -c core.longpaths=true fetch --progress {options} \"{url}\" {refspecs}", secret);
```
yufeih, line 158: Set environment variable `GIT_TERMINAL_PROMPT` to `0` in azure pipeline yml for running tests only. Changing the source code here breaks experiences that expects user interaction.

v-pegao: The OP build needs disable the prompt, otherwise, build hangs for invalid git access token.

I had tested `$env:GIT_TERMINAL_PROMPT=0`, but it didn't look like work for me, I haven't found a working environment variable for now, otherwise, we may not need change in docfx.

But we can try if installing git without git credential manager can work or not

yufeih: OP already disabled it here: https://dev.azure.com/ceapex/Engineering/_git/OpenPublishing.Build?path=/Applications/LocalRun/WorkerRole.cs&version=GBdevelop&line=54&lineEnd=54&lineStartColumn=49&lineEndColumn=68&lineStyle=plain&_a=contents

v-pegao: Let me confirm with Peng Cui first

v-pegao: @yufeih @herohua Peng is OOF, but the `GIT_TERMINAL_PROMPT` seems not work. The [bug](https://dev.azure.com/ceapex/Engineering/_workitems/edit/376820/) is created on Feb 2021, however, the [`GIT_TERMINAL_PROMPT=0` had been set in OPS on Mar 2020(even early)](https://dev.azure.com/ceapex/Engineering/_git/OpenPublishing.Build/commit/8f67cc5fe41d3d70ea0c46b1925826db78b66266?refName=refs/heads/develop&path=/Applications/LocalRun/WorkerRole.cs&_a=contents).

Should we handle this in Docfx side? What I think is we can try to handle this when installing git, otherwise, we can have a switch to control if we want to disable the prompt in docfx.

yufeih:
Close this as docfx isn't the right place to land a fix.

PR #7771, opened 2021-12-17
Remove usage of JToken.Parent

yufeih:

This is to pave the road of transitioning to `System.Text.Json` for better performance and less memory consumption. `Parent` isn't available in `JsonElement`, and have a `Parent` pointer means pining the whole json node tree even if we only reference a leaf node.

### DIFF
--- a/src/docfx/lib/schema/JsonSchemaTransformer.cs
+++ b/src/docfx/lib/schema/JsonSchemaTransformer.cs
@@ -208,13 +208,12 @@ internal class JsonSchemaTransformer
         string? propertyPath)
     {
         schemaMap.TryGetSchema(obj, out var schema);
-        var href = GetXrefHref(file, uid, uidCount, obj.Parent == null);
+        var href = GetXrefHref(file, uid, uidCount, string.IsNullOrEmpty(propertyPath));
         var monikers = _monikerProvider.GetFileLevelMonikers(errors, file);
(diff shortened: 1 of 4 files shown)


src/docfx/lib/schema/JsonSchemaTransformer.cs
```diff
@@ -208,13 +208,12 @@ private static bool IsContentTransform(JsonSchema schema)
...
     {
         schemaMap.TryGetSchema(obj, out var schema);
-        var href = GetXrefHref(file, uid, uidCount, obj.Parent == null);
+        var href = GetXrefHref(file, uid, uidCount, string.IsNullOrEmpty(propertyPath));
         var monikers = _monikerProvider.GetFileLevelMonikers(errors, file);
         var schemaType = GetSchemaType(uidSchema.SchemaType, schema?.SchemaTypeProperty, propertyPath, obj, file);
 
         var xref = new InternalXrefSpec(uid, href, file, monikers)
         {
-            DeclaringPropertyPath = obj.Parent?.Path,
```
herohua, line 217: Does `DeclaringPropertyPath` always equal to `SchemaType`?

yufeih: No, but the `DeclaringPropertyPath` here is only used to determine if it is an achievement, which could be implemented using `SchemaType`.

yufeih: Or yes for `badge` and `trophy` which are the only things we care.

herohua (approved):

PR #7772, opened 2021-12-17
json: SortProperties -> SerializeStable

yufeih:

Replace `SortProperties` with a `SerializeStable` method that produces a stable output for `JToken`. The `SortProperties` method creates a deep clone of the JSON node graph and isn't memory friendly.

The serializer is using `Utf8JsonWriter`, as part of the effort to move to `System.Text.Json`.

### DIFF
--- a/src/docfx/build/context/Output.cs
+++ b/src/docfx/build/context/Output.cs
@@ -32,8 +32,8 @@ internal class Output
 
         _queue.Post(() =>
         {
-            using var writer = new StreamWriter(EnsureDestinationPath(destRelativePath));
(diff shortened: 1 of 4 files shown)


live1206 (approved):

live1206:
The code looks good to me.
But based on the regression test, we will have all the output changed, seems we will have a full publish to DHS for all repos?

yufeih:
> The code looks good to me. But based on the regression test, we will have all the output changed, seems we will have a full publish to DHS for all repos?

Yes, which happens now and then.

PR #7773, opened 2021-12-17
Move to S.T.J for cached json files

yufeih:
[AB#545933](https://dev.azure.com/ceapex/Engineering/_workitems/edit/545933/)


herohua:
What does `S.T.J` stand for?

KalleOlaviNiemitalo:
> What does `S.T.J` stand for?

System.Text.Json

src/docfx/lib/process/ProcessUtility.cs
```diff
@@ -73,19 +79,19 @@ static void PipeStream(TextReader input, TextWriter output1, TextWriter? output2
...
         using (InterProcessMutex.Create(path))
         {
-            content = File.ReadAllText(path);
+            bytes = File.ReadAllBytes(path);
         }
 
         try
         {
-            return JsonUtility.DeserializeData<T>(content, new FilePath(path));
+            return JsonSerializer.Deserialize<T>(bytes, s_jsonOptions) ?? new();
```
live1206, line 94: Instead of using `JsonSerializer` directly, how about having another central place for `system.text.json`? Like the current `JsonUtility`

src/docfx/lib/process/ProcessUtility.cs
```diff
@@ -5,14 +5,20 @@
...
-/// <summary>
-/// Provide process utility
-/// </summary>
 internal static class ProcessUtility
 {
+    private static readonly JsonSerializerOptions s_jsonOptions = new()
+    {
+        AllowTrailingCommas = true,
+        Converters = { new JsonStringEnumConverter() },
+        PropertyNamingPolicy = new JsonSnakeCaseNamingPolicy(),
```
live1206, line 19: Not sure if we will have all the models with snake case naming.

yufeih: For the consumer of this class, yes. But we use both snake case and camel case, so at some stage later we'll probably need an enum to specify the naming convention.

src/docfx/lib/process/ProcessUtility.cs
yufeih: Yes, I was thinking about consolidating it to `JsonUtility` later, but now that you brought up.

src/docfx/lib/json/JsonSnakeCaseNamingPolicy.cs
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Text;
+using System.Text.Json;
+
+namespace Microsoft.Docs.Build;
+
+internal class JsonSnakeCaseNamingPolicy : JsonNamingPolicy
```
herohua, line 9: >JsonSnakeCaseNamingPolicy [](http://example.com/codeflow?start=15&length=25)

add unit tests for this class?

yufeih: Switched to use a NuGet package

src/docfx/lib/process/ProcessUtility.cs
```diff
@@ -73,19 +80,19 @@ static void PipeStream(TextReader input, TextWriter output1, TextWriter? output2
...
         using (InterProcessMutex.Create(path))
         {
-            content = File.ReadAllText(path);
+            bytes = File.ReadAllBytes(path);
         }
 
         try
         {
-            return JsonUtility.DeserializeData<T>(content, new FilePath(path));
+            return JsonSerializer.Deserialize<T>(bytes, s_jsonOptions) ?? new();
```
live1206, line 95: Is `byte[]` better than `string` here?

yufeih: Absolutely! `byte[]` takes less memory and `JsonSerializer` is internally back by `Utf8JsonReader` which works directly on bytes not strings.

live1206 (approved):

PR #7776, opened 2021-12-20
Fix Unsafe TypeNameHandling

TerryFei:
fix Unsafe TypeNameHandling


TerryFei:
Cancel this PR, because the the main branch is merged from dev.

PR #7778, opened 2021-12-20
enable dependency track

v-pegao:
[AB#531544](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/531544)

### DIFF
--- a/src/docfx/docfx.csproj
+++ b/src/docfx/docfx.csproj
@@ -24,7 +24,7 @@
     <PackageReference Include="Microsoft.DocAsCode.ECMAHelper" Version="1.2.1744.22" />
     <PackageReference Include="Microsoft.DocAsCode.MAML2Yaml.Lib" Version="1.1.1719.2" Aliases="maml" />
     <PackageReference Include="Microsoft.Docs.MetadataService.Models" Version="0.3.6" />
-    <PackageReference Include="Microsoft.Graph" Version="4.10.0" />
+    <PackageReference Include="Microsoft.Graph" Version="4.11.0" />
     <PackageReference Include="Microsoft.Experimental.Collections" Version="1.0.6-e190117-3" />
     <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
     <PackageReference Include="Jint" Version="3.0.0-beta-2035" />
--- a/src/docfx/docfx.csproj
+++ b/src/docfx/docfx.csproj
@@ -20,6 +20,7 @@
     <PackageReference Include="HtmlReaderWriter" Version="1.0.0-preview-0001-gcf3636c5d9" />
     <PackageReference Include="LibGit2Sharp.NativeBinaries" Version="2.0.314" />
     <PackageReference Include="Microsoft.ApplicationInsights" Version="2.19.0" />
+    <PackageReference Include="Microsoft.ApplicationInsights.DependencyCollector" Version="2.19.0" />
     <PackageReference Include="Microsoft.ChakraCore" Version="1.11.24" />
     <PackageReference Include="Microsoft.DocAsCode.ECMAHelper" Version="1.2.1744.22" />
     <PackageReference Include="Microsoft.DocAsCode.MAML2Yaml.Lib" Version="1.1.1719.2" Aliases="maml" />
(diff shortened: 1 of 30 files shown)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -17,47 +18,75 @@ internal static class Telemetry
...
+    private static readonly AsyncLocal<bool> s_isRealTimeBuild = new();
+
+    private static string s_repo = "<null>";
+    private static string s_branch = "<null>";
+
+    static Telemetry()
+    {
+        using var telemetryConfiguration = TelemetryConfiguration.CreateDefault();
+        using var dependencyTrackingTelemetryModule = new DependencyTrackingTelemetryModule();
+        dependencyTrackingTelemetryModule.Initialize(telemetryConfiguration);
```
herohua, line 59: We'll need to sanitize SAS from Azure storage URL. Please check sample code: https://dev.azure.com/ceapex/Engineering/_git/Docs.Build?path=/src/shared/Telemetry.cs&version=GBdevelop&line=21&lineEnd=22&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

src/docfx/lib/log/Telemetry.cs
```diff
@@ -17,47 +18,75 @@ internal static class Telemetry
...
+    private static readonly string s_correlationId = EnvironmentVariable.CorrelationId ?? Guid.NewGuid().ToString("N");
+    private static readonly string s_sessionId = EnvironmentVariable.SessionId ?? "<null>";
+    private static readonly AsyncLocal<bool> s_isRealTimeBuild = new();
+
+    private static string s_repo = "<null>";
+    private static string s_branch = "<null>";
+
+    static Telemetry()
+    {
+        using var telemetryConfiguration = TelemetryConfiguration.CreateDefault();
```
herohua, line 57: Can we really dispose this object here (and the line below) as `Telemetry` may still hold reference to it?

herohua (changes requested):

src/docfx/docfx.csproj
```diff
@@ -34,6 +34,7 @@
     <PackageReference Include="Stubble.Core" Version="1.9.3" />
     <PackageReference Include="System.CommandLine" Version="2.0.0-beta1.20574.7" />
     <PackageReference Include="YamlDotNet" Version="11.2.1" />
+    <PackageReference Include="Microsoft.ApplicationInsights.DependencyCollector" Version="2.19.0" />
```
yufeih, line 37: Sort these lines a to z

src/docfx/lib/log/Telemetry.cs
```diff
@@ -17,47 +18,75 @@ internal static class Telemetry
     // https://github.com/microsoft/ApplicationInsights-Home/blob/master/EndpointSpecs/Schemas/Bond/EventData.bond#L19
     private const int MaxEventPropertyLength = 8192;
     private const int MaxChildrenLength = 5;
-    private static readonly TelemetryClient s_telemetryClient = new(TelemetryConfiguration.CreateDefault());
+
+    private static TelemetryClient TelemetryClient { get; }
```
yufeih, line 22: Why is this turned into properties and not inline initialized as before? If the `TelemetryClient` initialization cannot be done with a single expression, you could use a helper method:
```suggestion
    private static readonly TelemetryClient s_telemetryClient = CreateTelemetryClient();
```

src/docfx/lib/log/Telemetry.cs
v-pegao: Added

src/docfx/lib/log/Telemetry.cs
v-pegao: Use dispose to dispose TelemetryConfiguration and DependencyTrackingTelemetryModule

src/docfx/docfx.csproj
v-pegao: Sorted

src/docfx/lib/log/Telemetry.cs
v-pegao: Updated

src/docfx/lib/log/Telemetry.cs
```diff
@@ -283,6 +301,12 @@ public static void Flush()
         Task.WaitAny(Task.Run(s_telemetryClient.Flush), Task.Delay(10000));
     }
 
+    public static void Dispose()
+    {
```
herohua, line 305: I think we can skip this implementation as all resources are static.

v-pegao: Removed

herohua (approved):

yufeih (approved):

PR #7779, opened 2021-12-20
no-loc when xref-not-found

AngryBerryMS:
[AB#532861](https://ceapex.visualstudio.com/Engineering/_workitems/edit/532861)


src/Microsoft.Docs.MarkdigExtensions/MarkdownExtensions.cs
```diff
@@ -153,6 +153,7 @@ public static MarkdownPipelineBuilder UseTripleColon(this MarkdownPipelineBuilde
     public static MarkdownPipelineBuilder UseNoloc(this MarkdownPipelineBuilder pipeline)
     {
         pipeline.Extensions.AddIfNotAlready(new NolocExtension());
+        pipeline.Extensions.AddIfNotAlready(new NoLocXrefContainerExtension());
         return pipeline;
```
herohua, line 157: Would be better to put under `UseXref`?

herohua (approved):
:shipit:

src/Microsoft.Docs.MarkdigExtensions/MarkdownExtensions.cs
AngryBerryMS: Done.

src/docfx/lib/markdown/XrefExtension.cs
```diff
@@ -27,7 +27,7 @@ internal static class XrefExtension
 
                 if (xrefLink.Href is null)
                 {
-                    return new LiteralInline(raw);
+                    return new NoLocXrefContainer(new LiteralInline(raw));
```
yufeih, line 30: We could use `HtmlInline` to replace this `NoLocXrefContainer`.

AngryBerryMS: `htmlTag`="<span class=no-loc>...</span>"?

src/docfx/lib/markdown/XrefExtension.cs
```diff
@@ -27,7 +27,9 @@ internal static class XrefExtension
 
                 if (xrefLink.Href is null)
                 {
-                    return new LiteralInline(raw);
+                    var nolocInline = new NolocInline();
```
yufeih, line 30: inline initialize.
```suggestion
                    return new NolocInline { Text = raw };
```

yufeih (approved):

src/docfx/lib/markdown/XrefExtension.cs
AngryBerryMS: Use nolocinline now

src/docfx/lib/markdown/XrefExtension.cs
AngryBerryMS: Donw

PR #7783, opened 2021-12-22
Fix dotnet publish

mingwli:
Fix `dotnet publish` [failure](https://github.com/dotnet/docfx/runs/4544110947?check_suite_focus=true)


mingwli:
local verified:
image
self-contained `docfx.exe` can run:
image

live1206 (approved):

PR #7789, opened 2021-12-23
Don't skip later validations if previous one fails in Hierarchy Data Validation

shanluo911:
[AB#532550](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/532550)

The validators run in the order: LearningPath => Module => ModuleUnit => Achievement. The validators are involed by Linq.All which returns false and skips the later delegation funcs if any delegation fails. Say, if learning path validation fails, all other three will be skipped. (Kudos to @mingwli)


herohua (approved):

live1206:
Rename the PR title to a more meaningful one, such as "Don't skip later validations if previous one fails"

live1206 (approved):

yufeih (approved):

PR #7790, opened 2021-12-23
check path invalid chars in redirection path

v-pegao:
[AB#532087](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/532087)

Sample error item:
image


src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -364,4 +377,20 @@ private static (string path, string query) RemoveTrailingIndex(string redirectio
         var (path, query, _) = UrlUtility.SplitUrl(redirectionUrl);
         return (path.EndsWith("/index", PathUtility.PathComparison) ? path[..^"index".Length] : path, query);
     }
+
+    private static bool TryGetInvalidPathChars(string path, out HashSet<char> invalidPathChars)
```
yufeih, line 381: This isn't just for redirection. `PathString` is the core type that represents a file path. We should mark a string as `PathString` in the data model if that string represents a path, we should also be checking for invalid chars for all places that turns user inputs to a `PathString`, this could be done in `PathString` json converter.

yufeih (changes requested):

src/docfx/build/redirection/RedirectionProvider.cs
v-pegao: Move the check to `PathString` json converter, not sure if we need to apply a more strict path validation here. For now, the validation will be applied on `PathString` type in redirection item and configs

herohua (changes requested):
There is 0 spec test for this change. Please add some tests covering both valid chars and invalid chars.

src/docfx/Errors.cs
```diff
@@ -126,6 +126,13 @@ public static Error NullArrayValue(SourceInfo? source, string name)
         /// Behavior: ❌ Message: ❌
         public static Error ViolateSchema(SourceInfo? source, string message)
             => new(ErrorLevel.Error, "violate-schema", $"{message}", source);
+
+        /// <summary>
+        /// Path string contains invalid path chars.
+        /// </summary>
+        /// Behavior: ❌ Message: ❌
+        public static Error PathInvalid(SourceInfo? source, string message)
+            => new(ErrorLevel.Warning, "path-invalid", $"{message}", source);
```
herohua, line 135:
```suggestion
            => new(ErrorLevel.Warning, "path-invalid", message, source);
```

herohua (approved):

src/docfx/Errors.cs
v-pegao: @herohua We use interpolated string here because the parameter of 'Error' is 'FormattableString' type:
image.

I have checked the diffs in failed regression tests results, they are all unrelated diffs. I don't have permission to merge PR if any PR check in failed state, could you please help merge the PR if it's ready to merge?

yufeih: You'll need to move this message template here: '$"Path {path} contains invalid chars {string.Join(", ", invalidPathChars.Select(c => $"'{c}'"))}.'

src/docfx/lib/path/PathUtility.cs
```diff
@@ -85,6 +88,27 @@ public static string NormalizeFolder(string path)
...
+            {
+                if (s_invalidPathChars.Contains(c))
+                {
+                    invalidPathChars.Add(c);
+                }
+            }
+        }
+        if (invalidPathChars.Count > 0)
+        {
+            throw new InvalidDataException($"Path {path} contains invalid chars {string.Join(", ", invalidPathChars.Select(c => $"'{c}'"))}.");
```
yufeih, line 107: throw `DocfxException` here.
```suggestion
            throw Errors.Json.PathInvalid(invalidPathChars).ToException();
```

src/docfx/lib/path/PathUtility.cs
```diff
@@ -85,6 +88,27 @@ public static string NormalizeFolder(string path)
         return str;
     }
 
+    public static string NormalizeFileWithValidation(string path)
```
yufeih, line 91: Keep the function single purpose, the caller can chain these two functions
```suggestion
    public static string CheckInvalidPathString(string path)
```

src/docfx/lib/path/PathUtility.cs
v-pegao: @yufeih Updated as suggested, could you please help review?

src/docfx/lib/json/JsonUtility.cs
```diff
@@ -591,6 +591,13 @@ private static void HandleError(object? sender, Newtonsoft.Json.Serialization.Er
                 state.Errors.Add(Errors.Json.ViolateSchema(state.Reader?.CurrentToken?.GetSourceInfo(), ParseException(args.ErrorContext.Error).message));
                 args.ErrorContext.Handled = true;
             }
+            if (args is not null && DocfxException.IsDocfxException(args.ErrorContext.Error, out var dex))
+            {
+                var state = s_status.Value!.Peek();
+                state.Errors.AddRange(dex.Select(ex => new Error(
```
yufeih, line 597: Use with expression
```suggestion
                state.Errors.AddRange(dex.Select(ex => ex.Error with { propertypath, sourceinfo... }));
```

src/docfx/lib/path/PathUtility.cs
```diff
@@ -85,6 +88,27 @@ public static string NormalizeFolder(string path)
         return str;
     }
 
+    public static string CheckInvalidPathString(string path)
+    {
+        path = NormalizeFile(path);
```
yufeih, line 93: Don't normalize in the check method.

src/docfx/lib/path/PathString.cs
```diff
@@ -170,7 +170,7 @@ private class PathStringJsonConverter : JsonConverter
         public override object ReadJson(JsonReader reader, Type objectType, object? existingValue, JsonSerializer serializer)
         {
             var value = serializer.Deserialize<string>(reader);
-            return value is null ? default : new PathString(PathUtility.NormalizeFile(value));
+            return value is null ? default : new PathString(PathUtility.CheckInvalidPathString(value));
```
yufeih, line 173:
```suggestion
            return value is null ? default : new PathString(PathUtility.NormalizeFile(PathUtility.CheckInvalidPathString(value)));
```

src/docfx/lib/json/JsonUtility.cs
v-pegao: Updated

yufeih (approved):

PR #7793, opened 2021-12-24
set CorrelationId to dependency telemetry

v-pegao:
[AB#535434](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/535434)

### DIFF
--- a/src/docfx/cli/Docfx.cs
+++ b/src/docfx/cli/Docfx.cs
@@ -19,6 +19,7 @@ public static class Docfx
     {
         try
         {
+            using var operation = Telemetry.StartOperation();
             return Run(args);
         }
         catch (Exception ex)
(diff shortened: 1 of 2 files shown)


herohua (approved):

yufeih (commented):
We should strive to use standand way of corrlating distributed traces, for AI, this is done using BeginOperation, all the traces would then follow the same operation_id

v-pegao:
> We should strive to use standand way of corrlating distributed traces, for AI, this is done using BeginOperation, all the traces would then follow the same operation_id

Updated to use TelemetryClient.StartOperation to make all trace follow same operation_id

herohua (approved):

yufeih:
Supersaded by #7797

PR #7803, opened 2022-01-05
Allow environment variable to have nested properties

yufeih:
Allows configuring nested properties using environment variables, specifically the GitHub token using `DOCFX__SECRETS__GITHUB_TOKEN`. Environment variables are handy for [github actions](https://github.com/docascode/template/blob/feature/ghpage/.github/workflows/gh-page.yml#L21).

With this change, a double-underscore (`__`) creates a nested property, a single underscore (`_`) indicates camel casing separator.

This change is backward compatible.

### DIFF
--- a/test/docfx.Test/build/ConfigTest.cs
+++ b/test/docfx.Test/build/ConfigTest.cs
@@ -1,12 +1,25 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections;
+using Newtonsoft.Json;
 using Xunit;
 
 namespace Microsoft.Docs.Build;
 
 public static class ConfigTest
 {
+    [Theory]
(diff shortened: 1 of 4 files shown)


test/docfx.Test/build/ConfigTest.cs
```diff
@@ -1,12 +1,25 @@
...
+using Newtonsoft.Json;
 using Xunit;
 
 namespace Microsoft.Docs.Build;
 
 public static class ConfigTest
 {
+    [Theory]
+    [InlineData("DOCFX_FOO", "bar", "{'foo':'bar'}")]
+    [InlineData("DOCFX_FOO", "a;b;c", "{'foo':['a','b','c']}")]
```
live1206, line 14: How about input the value as Json array? ['a', 'b', 'c'] in this case.
We can simply use `JToken.Parse(value)`

yufeih: This is the existing behavior. `;` is simpler than writing a proper json array xD

PR #7808, opened 2022-01-06
Upload binary to artifact on PR

yufeih:
Allows consuming binaries before it is released

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
     - uses: codecov/codecov-action@v1
       if: ${{ matrix.os == 'windows-latest' }}
 
-    - run: dotnet publish src\docfx\docfx.csproj -warnAsError -c Release -o $RUNNER_TEMP/publish
+    - run: dotnet publish src/docfx/docfx.csproj -warnAsError -c Release -o ${{ runner.temp }}/publish
     - uses: actions/upload-artifact@v2
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -22,4 +22,10 @@ jobs:
...
     - run: dotnet test -c Release /p:CollectCoverage=true /p:CoverletOutputFormat=opencover
     - uses: codecov/codecov-action@v1
-      if: ${{ matrix.os == 'macos-latest' }}
+      if: ${{ matrix.os == 'windows-latest' }}
+
+    - run: dotnet publish src\docfx\docfx.csproj -warnAsError -c Release -o $RUNNER_TEMP/publish
+    - uses: actions/upload-artifact@v2
+      with:
+        name: docfx-${{ matrix.os }}
+        path: $RUNNER_TEMP/publish
```
KalleOlaviNiemitalo, line 31: Seems `$RUNNER_TEMP` was not expanded when this action was executed.

KalleOlaviNiemitalo: Should it be `${{ env.RUNNER_TEMP }}` instead?

KalleOlaviNiemitalo: Or perhaps `${{ runner.temp }}`, see [`runner` context](https://docs.github.com/en/actions/learn-github-actions/contexts#runner-context).

live1206 (approved):

PR #7827, opened 2022-01-11
[DO NOT MERGE] Transform metadata based on metadata schema

yufeih:
[AB#401701](https://dev.azure.com/ceapex/Engineering/_workitems/edit/401701)

Prototype only, just to test diff


yufeih:
Diff looks good

PR #7828, opened 2022-01-11
fix pdf render type

easonyq:
This PR is to fix a potential bug described [here](https://ceapex.visualstudio.com/Engineering/_workitems/edit/537091).

Currently, content team has already added a `Toc.schema.json` to the [template repo](https://github.com/microsoft/templates.docs.msft). In [this schema file](https://github.com/microsoft/templates.docs.msft/blob/master/ContentTemplate/schemas/Toc.schema.json#L7), the `renderType` has been set to `component`.

PDF requires TOC should have a renderType of `content` (Because only `content` will make docfx generate HTML file, which will be consumed by PDF processes). So simply syncing this template to PDF template repo will break the PDF building process, until we push this fix.

### DIFF
--- a/src/docfx/template/JsonSchemaProvider.cs
+++ b/src/docfx/template/JsonSchemaProvider.cs
@@ -103,6 +103,11 @@ internal class JsonSchemaProvider
 
     private RenderType GetTocRenderType()
     {
+        // Refer to https://ceapex.visualstudio.com/Engineering/_workitems/edit/537091
+        // There is a TOC.schema.json file existing in templates.docs.msft repo whose 'renderType' is set to 'component'
+        // (https://github.com/microsoft/templates.docs.msft/blob/master/ContentTemplate/schemas/Toc.schema.json#L7)
+        // This will break the PDF process because PDF asks it to be 'content' (so that docfx can generate 'toc.html' which is necessary for building PDF)
+        // To avoid unexpected breaking PDF process through sync templates, special logic dealing with PDF cases goes first.
         if (_template.Url.EndsWith(".pdf", StringComparison.OrdinalIgnoreCase) ||
             _template.Path.EndsWith(".pdf", StringComparison.OrdinalIgnoreCase))
         {
--- a/src/docfx/template/JsonSchemaProvider.cs
+++ b/src/docfx/template/JsonSchemaProvider.cs
@@ -114,6 +114,13 @@ internal class JsonSchemaProvider
             return RenderType.Content;
         }
 
-        return GetSchema(new SourceInfo<string?>("toc")).RenderType;
+        try
+        {
+            return GetSchema(new SourceInfo<string?>("toc")).RenderType;
+        }
+        catch (Exception ex) when (DocfxException.IsDocfxException(ex, out var _))
+        {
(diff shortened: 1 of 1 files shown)


yufeih (approved):

src/docfx/template/JsonSchemaProvider.cs
```diff
@@ -103,16 +103,12 @@ private JsonSchema GetSchema(SourceInfo<string?> mime)
 
     private RenderType GetTocRenderType()
     {
-        try
-        {
-            return GetSchema(new SourceInfo<string?>("toc")).RenderType;
-        }
-        catch (Exception ex) when (DocfxException.IsDocfxException(ex, out var _))
+        if (_template.Url.EndsWith(".pdf", StringComparison.OrdinalIgnoreCase) ||
```
herohua, line 106: How about adding a comment?

herohua (approved):

herohua:
AB#537091

src/docfx/template/JsonSchemaProvider.cs
KalleOlaviNiemitalo: It would be good to describe the template naming rules in the documentation as well, but I didn't find any obvious place in <https://github.com/dotnet/docfx/tree/v3/docs>.

easonyq: Sure.

Moreover, I still have a hesitation about whether we should keep the `catch` logic. In current situation, TOC.schema.json existed in template repo, so `catch` won't be hit. But if for some reason when that json was removed, this function will throw error because the `catch` logic has also been removed. Do you think we should consider such condition? Or we simply think the TOC.schema.json shouldn't be removed. @herohua

dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/docfx?pullRequest=7828) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/docfx?pullRequest=7828) before we can accept your contribution.<br/><br/>:x: easonyq [sign now](https://cla.dotnetfoundation.org/dotnet/docfx?pullRequest=7828)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/docfx?pullRequest=7828) it.</sub>

src/docfx/template/JsonSchemaProvider.cs
yufeih: @KalleOlaviNiemitalo 
This is more of a temporary backdoor to fix some obscure bugs in docsmsft and would be removed once a proper fix is put together in docsmsft.

yufeih: @easonyq , yes we need the catch block. `TOC.schema.json` isn't mandontory. I wonder if any test fails.

easonyq: I've run all UT locally and they are all green. But I still want to add this `catch` logic back in case something unexpected will happen.

Issue #3387, opened 2018-09-07
Support configuration for Markdig built-in extensions


usigrist:
**DocFX Version Used**: 2.39.1  

Is there a way to use grid tables by enabling the [markdig Grid Table](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/Specs/GridTableSpecs.md) extension when using the markdig engine?

herohua:
@yishengjin1413 @wh-alice

usigrist:
Could this be done using the Property `markdownEngineProperties` in `docfx.json` with a new key and a list of native markdig extensions to be added during pipeline initialization? 
If yes I could implement it and provide a pull request.

superyyrrzz:
It sounds good to me. @Thraka also want this extension (https://github.com/dotnet/docfx/issues/4201#issuecomment-474636674) and may like this idea.

@yishengjin1413 @wh-alice What do you think of this? I know docfx has some custom Markdig extension. My only concern is whether it will cause some conflict if we allow users to choose extensions.

yishengjin1413:
@usigrist Sounds good. And I suggest you to add one method like  builder.UseCustomExtensions(_context, config) in [L143](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine/MarkdigMarkdownService.cs#L143) before builder.UseDocfxExtensions(_context) . Since some extensions are added by order in DocfxExtensions.

superyyrrzz:
Markdig also supports footnotes for #1564

Thraka:
I would love this.

sommmen:
Any update on footnote support?

smokedlinq:
Hey @docs-product or someone else, will be this something we can look forward to in v3?

cillroy:
we are evaluating this and looking into the impact on our design team

tibel:
Enabling [markdig Diagrams](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/Specs/DiagramsSpecs.md) would be great too.

lionelberton:
Enabling [markdig Math formulas](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/Specs/MathSpecs.md) would great too

simonziegler:
I'm also keen to see footnotes, diagrams and maths formulae at some time.

wangkanai:
[...]

PR #7833, opened 2022-01-14
Optionally enable Markdig extensions that are not included by default in DocFX

jakraft:
Taking a first crack at #3387

This change will allow users to optionally enable various Markdig extensions that are disabled by default in DocFX. The following extensions are included in this change (more can be added later):
- Task lists
- Grid tables
- Footnotes
- Mathematics
- Diagrams
- Definition lists

To enable them, simply include them as a property in your `markdownEngineProperties` (make sure Markdig is set as your `markdownEngine`):

```json
{
  "build": {
    ...
    "markdownEngineName": "markdig",
    "markdownEngineProperties": {
      "enableTaskLists": true,
      "enableDefinitionLists": true
    }
  }
}
```

### DIFF
--- a/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Constants.cs
+++ b/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/Constants.cs
@@ -13,40 +13,21 @@ namespace Microsoft.DocAsCode.MarkdigEngine.Extensions
         }
 
         /// <summary>
-        /// Optional extensions that can be enabled from
-        /// the markdownEngineProperties property in the docfx.json.
+        /// Names of properties supported in the markdownEngineProperties
+        /// property in the docfx.json
         /// </summary>
-        public static class OptionalExtensionPropertyNames
+        public static class EngineProperties
         {
             /// <summary>
-            /// Enables the Task List Markdig extension by invoking <see cref="Markdig.MarkdownExtensions.UseTaskLists(Markdig.MarkdownPipelineBuilder)"/>.
+            /// Enables the <see cref="LineNumberExtension"/>.
             /// </summary>
-            public const string EnableTaskLists = "enableTaskLists";
+            public const string EnableSourceInfo = "EnableSourceInfo";
 
             /// <summary>
-            /// Enables the Grid Tables Markdig extension by invoking <see cref="Markdig.MarkdownExtensions.UseGridTables(Markdig.MarkdownPipelineBuilder)"/>.
+            /// Contains a list of optional Markdig extensions that are not
+            /// enabled by default by DocFX.
             /// </summary>
-            public const string EnableGridTables = "enableGridTables";
-
-            /// <summary>
-            /// Enables the Footnotes Markdig extension by invoking <see cref="Markdig.MarkdownExtensions.UseFootnotes(Markdig.MarkdownPipelineBuilder)"/>.
-            /// </summary>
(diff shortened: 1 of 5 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/docfx?pullRequest=7833) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/docfx?pullRequest=7833) before we can accept your contribution.<br/><br/>:x: jakraft [sign now](https://cla.dotnetfoundation.org/dotnet/docfx?pullRequest=7833)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/docfx?pullRequest=7833) it.</sub>

yufeih:
Thank you for the enhancement! 

The tests and parameter passing flow looks good, I wonder if it is possible to use the [configuration vocabularies provided by markdig](https://github.com/xoofx/markdig/blob/master/src/Markdig/MarkdownExtensions.cs#L538), this way we don't have to invent our own configuration options and would pickup latest markdig extensions without code change, something like:

```json
{
  "build": {
    "markdownEngineName": "markdig",
    "markdownEngineProperties": {
      "markdigExtensions": [
          "common",
          "pipetables"
      ]
    }
  }
}
```

We'll also need a special extension name to represent the current docfx setup as the default like `"dfm"`. (Docfx Favored Markdown)

jakraft:
@yufeih Ah that's a much better solution! I wish I had spotted that earlier. Could you please clarify what you mean by this:

> We'll also need a special extension name to represent the current docfx setup as the default like "dfm" (Docfx Favored Markdown).

Is the expectation here that users should always include "dfm" in the list of extensions, and if it's not included then we shouldn't call `UseDocFxExtensions()`? I would have figured we'd want those extensions to always be applied.

Somewhat outside the scope of this change, it may also be worth considering adding a "disabledMarkdigExtensions" property to allow users to opt-out of certain extensions. That would allow new extensions to be enabled by default in the future while maintaining backward compatibility with existing documentation. I don't think it's strictly necessary for this first iteration though.

jakraft:
Updated the implementation to use a list of extensions as per @yufeih's suggestion. It may make sense to add a "disabledMarkdigExtensions" later, but that might be tricky to do. The best way I could think to do that is to make a separate `MarkdownPipelineBuilder` instance, calling `Configure()` on the list of disabled extensions, and then calling `builder.Extensions.RemoveAll(e => disabledExtensionsBuilder.Extensions.Contains(de => de.GetType() == e.GetType())`. Not a very elegant solution.

yufeih:
@jakraft 

How about `markdownExtensions` defaults to `["dfm"]` when absent and we treat all extensions equally, this could addresses the extension removal problem:
- `{}`: fallback to `["dfm"]` with all docfx extensions applied
- `{ markdownExtensions: [] }` : No markdown extensions at all
- `{ markdownExtensions: ['dfm', 'tasklists'] }`: Docfx extensions and tasklists
- `{ markdownExtensions: ['tasklists'] }`: Tasklists only

yufeih:
There doesn't seem to be a strong desire to remove a markdown extension, most of the time you can just avoid using the syntax. So the design implemented in this PR looks good.

yufeih (approved):

PR #7834, opened 2022-01-14
report: include moniker range info for moniker-range-undefined

herohua:
[AB#479182](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/479182)

PR #7843, opened 2022-01-19
report: refine error message for moniker range defined in file metadata or in moniker zone

herohua:
This PR supersedes #7834 and #7839 . For moniker range defined in file metadata or in moniker zone, we'd better let end users know that they need to 1) include the file in a group; 2) the group must have moniker range defined.

[AB#479182](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/479182)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -469,10 +469,9 @@ internal static class Errors
         /// which used monikerRange in its yaml header or used moniker-zone syntax.
         /// </summary>
         /// Behavior: ✔️ Message: ❌
-        public static Error MonikerRangeUndefined(SourceInfo? source, string? monikerRange)
+        public static Error MonikerRangeUndefined(SourceInfo<string?> source, FilePath file)
         {
-            var processedMonikerRange = monikerRange != null ? $"'{monikerRange}' " : string.Empty;
-            return new(ErrorLevel.Suggestion, "moniker-range-undefined", $"Moniker range {processedMonikerRange}is missing from the 'groups' setting in docfx.yml/docfx.json. It should not be defined in file metadata or moniker zone.", source);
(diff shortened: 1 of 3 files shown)


src/docfx/Errors.cs
```diff
@@ -469,10 +469,9 @@ public static Error MonikerRangeInvalid(SourceInfo? source, string message)
         /// which used monikerRange in its yaml header or used moniker-zone syntax.
         /// </summary>
         /// Behavior: ✔️ Message: ❌
-        public static Error MonikerRangeUndefined(SourceInfo? source, string? monikerRange)
+        public static Error MonikerRangeUndefined(SourceInfo<string?> source, FilePath file)
         {
-            var processedMonikerRange = monikerRange != null ? $"'{monikerRange}' " : string.Empty;
-            return new(ErrorLevel.Suggestion, "moniker-range-undefined", $"Moniker range {processedMonikerRange}is missing from the 'groups' setting in docfx.yml/docfx.json. It should not be defined in file metadata or moniker zone.", source);
+            return new(ErrorLevel.Suggestion, "moniker-range-undefined", $"Moniker range '{source}' should not be defined in file metadata of docfx.yml/docfx.json or in moniker zone of file '{file}'. Please check the 'groups' setting in docfx.yml/docfx.json.", source);
```
928PJY, line 474: The source already contains the file info, do we still need to mention it in the message?
If we still need, the source contains the FilePath property we don't need to pass it in? are you considering filemetadata or globalmetadata case?

herohua: Yes, file metadata and global metadata are considered - you can find corresponding yaml test. Thus, here `source` could be `docfx.json`. However, since we'd like to print out the md file where this error happens, we will need to pass in `file` separately.

AngryBerryMS (approved):

yufeih:
Retriggerd mac/linux tests

PR #7846, opened 2022-01-20
Log three types of validation ruleset name

928PJY:
[AB#545067](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/545067)

Fix random ruleset name:
image
you can reference the Bug for more detail

image

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -79,6 +79,12 @@ internal static class Errors
 
         public static Error MetadataValidationRuleset(string ruleset, string url)
             => new(ErrorLevel.Info, "MetadataValidationRuleset", $"Metadata validation ruleset used: {ruleset}. Document url: {url}");
+
+        public static Error ContentValidationRuleset(string ruleset, string url)
+            => new(ErrorLevel.Info, "ContentValidationRuleset", $"Content validation ruleset used: {ruleset}. Document url: {url}");
+
+        public static Error BuildValidationRuleset(string ruleset, string url)
+            => new(ErrorLevel.Info, "BuildValidationRuleset", $"Build validation ruleset used: {ruleset}. Document url: {url}");
     }
 
     public static class Logging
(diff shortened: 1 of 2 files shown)


src/docfx/Errors.cs
```diff
@@ -79,6 +79,12 @@ public static Error MicrosoftGraphApiFailed(string exMessage)
 
         public static Error MetadataValidationRuleset(string ruleset, string url)
             => new(ErrorLevel.Info, "MetadataValidationRuleset", $"Metadata validation ruleset used: {ruleset}. Document url: {url}");
+
+        public static Error ContentValidationRuleset(string ruleset, string url)
```
yufeih, line 83: Consider use only 1 error code. We can add parameters to the message to let it show up in logs.

928PJY: the html report will use the error code to identify the type, if we use the same error code, which one should they use?

yufeih: The distinct union

928PJY: but the build report call it metadata validation ruleset, so it should only display the `metadata ruleset`

herohua: Can we make it something like `{type} validation ruleset used: {ruleset}. Document url: {url}`?

herohua: Or, since reporting side is only a RegEx match, we could change it to whatever string we'd like to have.

yufeih: If these rulesets are not supposed to show up in the report, why log them in the first place?

herohua: It would help troubleshooting I suppose.

yufeih: If for troubleshooting, writing to the console would suffice

928PJY: @yufeih Because the log is the contract between docfx and op, so we need to log the `metadatavalidationruleset`, off course we can dinstinguish the "metadatavalidationruleset", then only log it and ignore the others(or write the others to console), but is that necessary? what is your concern to log them all?
@herohua the report generator ont only match the message, it use the logcode as well, it sounds confused to me to use the same "MetadataValidationRuleset" for all the three types of validation(metadata, build, content).

yufeih (approved):

herohua (approved):

herohua:
There are a lot of diff's in regression tests. Please consult @AngryBerryMS and @live1206 before merging the PR.

AngryBerryMS (approved):
:shipit:

PR #7847, opened 2022-01-20
Optimize _tocRel calculation

yufeih:
The `FindNearestToc` method is the biggest contributor (Nearly 30%) to memory allocation. This PR optimizes the method to become allocation-free.

The method uses LINQ extensively and misuse of LINQ leads to double enumeration. In some reference repos, the `tocs` can be tens of thousands and this leads to excessive allocation in path operations.

[AB#545933](https://dev.azure.com/ceapex/Engineering/_workitems/edit/545933/)

See how the memory usage curve flattens after the change:

Before:
image

After:
image

### DIFF
--- a/src/docfx/build/toc/TocMap.cs
+++ b/src/docfx/build/toc/TocMap.cs
@@ -98,57 +98,76 @@ internal class TocMap
     /// 3. sub-name lexicographical nearest
     /// </summary>
     internal static (T? toc, bool hasReferencedTocs) FindNearestToc<T>(
-        T file, IEnumerable<T> tocs, Dictionary<T, T[]> documentsToTocs, Func<T, string> getPath) where T : class, IComparable<T>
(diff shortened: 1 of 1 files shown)


src/docfx/build/toc/TocMap.cs
```diff
@@ -98,57 +98,76 @@ public IEnumerable<FilePath> GetFiles()
...
         {
-            return default;
+            var chA = pathA[i];
+            var chB = pathB[i];
+            if (chA != chB)
+            {
+                break;
+            }
+
+            if (chA == '/')
```
live1206, line 146: Assuming `pathA` is the shorter one here?

yufeih: No, `chA == chB` when it hit here

yufeih:
No diff, wooolaaa ♨️♨️♨️

live1206 (approved):

PR #7864, opened 2022-01-25
Remove dead code in TocDocumentProcessorBase

johnthcall:
I noticed what appears to be dead code in TocDocumentProcessorBase, if it's intentional to validate that TryGetFiledDetails succeeds we remove setting repoDetail and put a comment.


yufeih (approved):
Thanks! This is indeed dead code.

PR #7887, opened 2022-02-14
add contributors with noreply emails

AngryBerryMS:
[AB#549302](https://dev.azure.com/ceapex/Engineering/_workitems/edit/549302)

image

### DIFF
--- a/src/docfx/build/metadata/ContributionProvider.cs
+++ b/src/docfx/build/metadata/ContributionProvider.cs
@@ -68,6 +68,15 @@ internal class ContributionProvider
             foreach (var commit in contributionCommits)
             {
                 var (error, githubUser) = _githubAccessor.GetUserByEmail(commit.AuthorEmail, repoOwner, repoName, commit.Sha);
+                var username = GetGitHubUsernameWhenCommitWithNoReplyEmail(commit.AuthorEmail);
+                if (githubUser is null && !string.IsNullOrEmpty(username))
+                {
+                    githubUser = new GitHubUser()
+                    {
+                        Name = commit.AuthorName,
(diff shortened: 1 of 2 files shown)


src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -68,6 +68,15 @@ internal class ContributionProvider
             foreach (var commit in contributionCommits)
             {
                 var (error, githubUser) = _githubAccessor.GetUserByEmail(commit.AuthorEmail, repoOwner, repoName, commit.Sha);
+                var username = GetGitHubUsernameWhenCommitWithNoReplyEmail(commit.AuthorEmail);
```
herohua, line 71: move into `GetUserByEmail`?

yufeih: If this email pattern matches, we could save the API call to GitHub.

src/docfx/build/metadata/ContributionProvider.cs
```diff
@@ -210,4 +219,21 @@ private DateTime GetUpdatedAt(FilePath file, Repository? repository, GitCommit[]
             ? _input.GetLastWriteTimeUtc(file)
             : default;
     }
+
+    private static string GetGitHubUsernameWhenCommitWithNoReplyEmail(string email)
```
yufeih, line 223: Prefer `null` over empty string to communicate this method may not always succeed.
Is the first number github id? if so, populate it to `GitHubUser.Id` field

```suggestion
    private static GitHubUser? TryGetGitHubUserFromNoReplyEmail(string email)
```

PR #7892, opened 2022-02-15
Sanitize URL for .links.json

live1206:
[AB#549923](https://dev.azure.com/ceapex/Engineering/_workitems/edit/549923)

### DIFF
--- a/src/docfx/build/link/FileLinkMapBuilder.cs
+++ b/src/docfx/build/link/FileLinkMapBuilder.cs
@@ -11,31 +11,23 @@ internal class FileLinkMapBuilder
     private readonly DocumentProvider _documentProvider;
     private readonly MonikerProvider _monikerProvider;
     private readonly ContributionProvider _contributionProvider;
-    private readonly Dictionary<string, TrustedDomains> _trustedDomains;
 
     private readonly Scoped<ConcurrentHashSet<FileLinkItem>> _links = new();
 
     public FileLinkMapBuilder(
-        ErrorBuilder errors,
-        DocumentProvider documentProvider,
-        MonikerProvider monikerProvider,
-        ContributionProvider contributionProvider,
-        Dictionary<string, TrustedDomains> trustedDomains)
+        ErrorBuilder errors, DocumentProvider documentProvider, MonikerProvider monikerProvider, ContributionProvider contributionProvider)
     {
         _errors = errors;
         _documentProvider = documentProvider;
         _monikerProvider = monikerProvider;
         _contributionProvider = contributionProvider;
-        _trustedDomains = trustedDomains;
     }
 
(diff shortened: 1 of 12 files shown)


AngryBerryMS (approved):
:shipit:

yufeih (changes requested):
Fixing this doesn't require changes to `LinkResolver`:
- Sanitized HTML links should not be resolved in the first place.
- HTML link sanitization only works on markdown content, while `LinkResolver` is for all the links.

Adjust the link processing order in `MarkdownEngine.cs` to sanitize HTML before resolving markdown links.

live1206:
> Fixing this doesn't require changes to `LinkResolver`:
> 
> * Sanitized HTML links should not be resolved in the first place.
> * HTML link sanitization only works on markdown content, while `LinkResolver` is for all the links.
> 
> Adjust the link processing order in `MarkdownEngine.cs` to sanitize HTML before resolving markdown links.

`.links.json` not only contains `markdown content`, how can all links be sanitized by only sanitize them in `MarkdownEngine.cs`?

live1206:
> > Fixing this doesn't require changes to `LinkResolver`:
> > 
> > * Sanitized HTML links should not be resolved in the first place.
> > * HTML link sanitization only works on markdown content, while `LinkResolver` is for all the links.
> > 
> > Adjust the link processing order in `MarkdownEngine.cs` to sanitize HTML before resolving markdown links.
> 
> `.links.json` not only contains `markdown content`, how can all links be sanitized by only sanitizing them in `MarkdownEngine.cs`?

I am going to update tests to verify `.links.json` for `xref` and `SDP` as well

src/docfx/build/link/FileLinkMapBuilder.cs
```diff
@@ -54,4 +62,13 @@ public object Build(PublishModel publishModel)
 
         return new { links };
     }
+
+    private bool IsTrusted(string href, string tagName)
```
yufeih, line 66: I feel like the protocol check should happen at the `LinkResolver` layer

live1206: That's one way to sanitize, but then it would be partially duplicated with what's in place since we are not sanitizing attribute here.

And also, the output links in `.links.json` could come from `XrefResolver` as well, if this is unnecessary, we can put the check before https://github.com/dotnet/docfx/blob/afd9c17016a91275ff76f6d2d62da8c00faea665/src/docfx/build/link/LinkResolver.cs#L111

yufeih: I feel like I should have implemented trusted domains at `LinkResolver` and `XrefResolver` in the first place #7220 . We could consolidate them later.

yufeih (approved):

PR #7894, opened 2022-02-17
Fix warnings of unnecessary assignment

v-caxian:
[AB#559395](https://dev.azure.com/ceapex/Engineering/_workitems/edit/559395)

### DIFF
--- a/src/Microsoft.Docs.MarkdigExtensions/ExtensionsHelper.cs
+++ b/src/Microsoft.Docs.MarkdigExtensions/ExtensionsHelper.cs
@@ -252,14 +252,6 @@ public static class ExtensionsHelper
             }
             else
             {
-                if (title.Length >= 2 && title.First() == '\'' && title.Last() == '\'')
-                {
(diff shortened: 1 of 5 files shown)


superyyrrzz (approved):

src/Microsoft.Docs.MarkdigExtensions/ExtensionsHelper.cs
```diff
@@ -252,14 +252,6 @@ private static bool MatchPath(ref StringSlice slice, ref string path)
             }
             else
             {
```
yufeih, line 254: Why is this section removed? The `else if` is duplicated logic, but the first `if` is doing something meaningful?

superyyrrzz: Looks like no matter `if` or `else if`, the assignment to `title` is useless, as `title` never used below...

yufeih (approved):

PR #7910, opened 2022-02-24
[Fix] Correct link type to external when leading host is not removed

shanluo911:
AB#563013

Discussed with @live1206, if the host is removed, it's AbsolutePath, otherwise, External.

### DIFF
--- a/src/docfx/build/link/LinkResolver.cs
+++ b/src/docfx/build/link/LinkResolver.cs
@@ -142,7 +142,6 @@ internal class LinkResolver
         ValidateLink(inclusionRoot, linkNode);
         if (linkType == LinkType.External)
         {
-            var resolvedHref = _config.RemoveHostName ? UrlUtility.RemoveLeadingHostName(href, _config.HostName) : href;
             if (_config.TrustedDomains.TryGetValue(tagName, out var domains) && !domains.IsTrusted(href, out var untrustedDomain))
             {
                 if (tagName == "img")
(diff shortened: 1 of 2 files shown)


src/docfx/build/link/LinkResolver.cs
```diff
@@ -156,7 +155,13 @@ internal class LinkResolver
                 return (errors, "", fragment, LinkType.AbsolutePath, null, false);
             }
 
-            return (errors, resolvedHref, fragment, LinkType.AbsolutePath, null, false);
+            var resolvedHref = _config.RemoveHostName ? UrlUtility.RemoveLeadingHostName(href, _config.HostName) : href;
+            if (!string.Equals(href, resolvedHref, StringComparison.Ordinal))
```
yufeih, line 159:
```suggestion
            if (href != resolvedHref)
```

src/docfx/build/link/LinkResolver.cs
```diff
@@ -156,7 +155,13 @@ internal class LinkResolver
                 return (errors, "", fragment, LinkType.AbsolutePath, null, false);
```
herohua, line 155: It is still possible for an image to have host name like `docs.microsoft.com`.

shanluo911: Not sure, but `docs.microsoft.com` should be a trusted domain, and it will not enter the `if` statement (line 146)

live1206:
Add or update an existing spec test to cover the code change.

PR #7913, opened 2022-02-28
add link telemetry

shanluo911:
[AB#564511](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/564511)

We are planning to move broken link service to build phase. In order to support that, we will need to know some of the key data, for example: number of the links that are not validated by docfx, average number of links in each content file, and etc. We need to collect the link part in docfx.

design spec: https://microsoft-my.sharepoint.com/:w:/p/shanluo/EXP2FbwFZcJAifEI1iCJAToBNBpAf-EX9IPm6Xg9whLxbw?e=iraqp9

### DIFF
--- a/src/docfx/build/xref/XrefResolver.cs
+++ b/src/docfx/build/xref/XrefResolver.cs
@@ -105,6 +105,7 @@ internal class XrefResolver
 
         query = queries.AllKeys.Length == 0 ? "" : "?" + string.Join('&', queries);
         var fileLink = UrlUtility.MergeUrl(xrefSpec.Href, query, fragment);
+        Telemetry.TrackLink("xref");
         _fileLinkMapBuilder.AddFileLink(inclusionRoot, referencingFile, fileLink, href.Source);
 
         resolvedHref = UrlUtility.MergeUrl(resolvedHref, query, fragment);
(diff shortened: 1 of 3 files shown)


src/docfx/lib/log/Telemetry.cs
```diff
@@ -357,6 +390,33 @@ private static void TryRemoveNestedObject(this JObject graph)
         }
     }
 
+    private static (string repoAccount, string repoName, string repoType) TryGetDetailedRepoInfo (string urlString)
```
yufeih, line 393: Use repo url directly as other telemetries

src/docfx/build/xref/XrefResolver.cs
```diff
@@ -105,6 +105,7 @@ internal class XrefResolver
 
         query = queries.AllKeys.Length == 0 ? "" : "?" + string.Join('&', queries);
         var fileLink = UrlUtility.MergeUrl(xrefSpec.Href, query, fragment);
+        Telemetry.TrackLink("xref");
```
yufeih, line 108: This is likely going to have a perf impact similar to markdown element telemetry as we have many many links. We are basically grouping item count of `FileLinkBuilder` so track the aggregated result in `FileLinkBuilder.Build` method.

PR #7915, opened 2022-03-04
filter out info level in vscode validation extension

shanluo911:
[AB#566595](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/566595)

### DIFF
--- a/docs/specs/lsp.yml
+++ b/docs/specs/lsp.yml
@@ -227,3 +227,14 @@ languageServer:
       .openpublishing.publish.config.json: "{}"
   - expectDiagnostics:
      .openpublishing.publish.config.json: [{"code": "config-not-found"}]
+---
+# Info-level errors will be filtered out.
+inputs:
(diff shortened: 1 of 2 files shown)


herohua (approved):

yufeih (changes requested):
It feels wrong to suppress all info level traces to fix a single inaccurate error code.

928PJY:
> It feels wrong to suppress all info level traces to fix a single inaccurate error code.

The **Info** trace is still logged in the output channel, we just don't convert it into diagnostic, which will be displayed when the user hover, it is noisy to convert all info into disagnostic.

928PJY:
There is a configuration called "trace-level", we should be able to leverage that to control the trace level, but for this fix, it looks good to me.

yufeih (approved):
Looks good for info

PR #7917, opened 2022-03-07
Update template parameter for next build only.

ikondrasovas:
By following this walkthrough as it is, the generated site does not contain template changes and there is the following warning message:
"Warning:[BuildCommand.BuildCore.Build Document.LinkPhaseHandlerWithIncremental.Apply Templates]There is no template processing document type(s): Conceptual,Toc"

By also specifying the "default," template, no warnings. The customization is also successfully applied.

This sintax is also used in the template system doc: https://dotnet.github.io/docfx/tutorial/howto_create_custom_template.html#merge-template-with-default-template


yufeih (approved):
Thanks!

PR #7926, opened 2022-03-10
Add buildTime config for regression test

freewheel70:


test/docfx.RegressionTest/RegressionTest.cs
```diff
@@ -115,6 +115,7 @@ static string GetDocfxConfig(Options opts)
                 maxFileSuggestions = 10000,
                 maxFileInfos = 10000,
                 updateTimeAsCommitBuildTime = true,
+                buildTime = "2022-03-10T08:08:08Z",
```
herohua, line 118: Can we use `DateTime.UtcNow`?

herohua: Using a fix value here may hide some potential bugs from surfacing out.

freewheel70: @herohua  the `buildTime` will be used as `updated_at`, in my understanding,  this config is designed to be fixed so that every time regression test runs, the updated_at value will be the same.

freewheel70: FYI https://github.com/dotnet/docfx/pull/5710/files

freewheel70: on a second thought, this PR would only be work-around. 
if we fix [https://dev.azure.com/ceapex/Engineering/_workitems/edit/568811?src=WorkItemMention&src-action=artifact_link], the issue will be gone.

PR #7928, opened 2022-03-11
CommitBuildTimeProvider read latest state file before save

freewheel70:
1. `CommitBuildTimeProvider` read latest state file before save to avoid later `CommitBuildTimeProvider.Save()` cover the previous one's content
2. initialize `_buildTime` in the global `Config` to guarantee all `CommitBuildTimeProvider` instances share the same value.

### DIFF
--- a/src/docfx/lib/git/CommitBuildTimeProvider.cs
+++ b/src/docfx/lib/git/CommitBuildTimeProvider.cs
@@ -5,6 +5,8 @@ namespace Microsoft.Docs.Build;
 
 internal class CommitBuildTimeProvider
 {
+    private static readonly object s_lock = new();
+
     private readonly DateTime _buildTime;
     private readonly Repository _repo;
     private readonly Config _config;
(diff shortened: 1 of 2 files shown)


freewheel70:
Manually trigger another regression test

src/docfx/lib/git/CommitBuildTimeProvider.cs
```diff
@@ -40,6 +33,7 @@ public void Save()
 
         using (PerfScope.Start($"Saving commit build time for {_repo.Commit}"))
         {
+            _buildTimeByCommit = ReadLatestCacheIfAny();
```
herohua, line 36: Should we lock on this operation (read + write)? Otherwise, the file might be read by two providers but only last data are saved.

src/docfx/lib/git/CommitBuildTimeProvider.cs
```diff
@@ -9,23 +9,16 @@ internal class CommitBuildTimeProvider
...
-    private readonly IReadOnlyDictionary<string, DateTime> _buildTimeByCommit;
+    private IReadOnlyDictionary<string, DateTime> _buildTimeByCommit;
 
     public CommitBuildTimeProvider(Config config, Repository repo)
     {
         _repo = repo;
         _config = config;
         _commitBuildTimePath = AppData.BuildHistoryStatePath;
-        _buildTime = config.BuildTime ?? DateTime.UtcNow;
+        _buildTime = config.BuildTime;
```
yufeih, line 19: Does it fix the problem if we simply promote the now fallback to a static variable?
```suggestion
        _buildTime = config.BuildTime ?? s_now;
```

src/docfx/lib/git/CommitBuildTimeProvider.cs
```diff
@@ -38,22 +34,42 @@ public void Save()
...
         }
 
-        using (PerfScope.Start($"Saving commit build time for {_repo.Commit}"))
+        lock (s_lock)
         {
-            var commits = _buildTimeByCommit.Select(item => new CommitBuildTimeItem { Sha = item.Key, BuiltAt = item.Value }).ToList();
-
-            // TODO: retrieve git log from `GitCommitProvider` since it should already be there.
-            foreach (var diffCommit in GitUtility.GetCommits(_repo.Path, _repo.Commit))
+            using (PerfScope.Start($"Saving commit build time for {_repo.Commit}"))
```
herohua, line 39: nit: Let's log the repo info here too.

herohua (approved):

PR #7979, opened 2022-04-07
Adds support for custom text in cref references

### LINKED ISSUE #7978: Content of see cref tag gets dropped
**Operating System**: Windows

**DocFX Version Used**: 2.59

**Template used**: N/A

**Steps to Reproduce**:
1.
Create two classes with the same member name.
```cs
public class Class1
{
   public string Name { get ;set; }
}
public class Class2
{
   public string Name { get ;set; }
}
```
Reference these two in xml doc. We'll use the content of the tag to fully qualify the member, so we can tell the two `Name` properties apart in the generated documentation (otherwise it'll just say `Name` for both of them):
```cs
/// <summary>Some class</summary>
/// <remarks>See <see cref="Class1.Name">Class1.Name</see> and <see cref="Class2.Name">Class2.Name</see>.</remarks>
public class SomeClass
{
}
```
2. Generate the API yaml
3. Notice the generated yaml dropped the content of the reference, and the xref tag is empty:
```yaml
remarks: "See <xref href=\"MyLib.Class1.Name\" data-throw-if-not-resolved=\"false\"></xref> and <xref href=\"MyLib.Class2.Name\" data-throw-if-not-resolved=\"false\"></xref>.\n"
```
**Expected Behavior**:
Generate xref with the content preserved:
```yaml
remarks: "See <xref href=\"MyLib.Class1.Name\" data-throw-if-not-resolved=\"false\">Class1.Name</xref> and <xref href=\"MyLib.Class2.Name\" data-throw-if-not-resolved=\"false\">Class2.Name</xref>.\n"
```
**Actual Behavior**:
Content of the cref is lost.

**Note**:
Putting the string text inside the xref might not be the correct way. [This doc](https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#advanced-more-options-for-cross-reference) seems to indicate it should be done a different way, but there doesn't seem to be a generic way to create arbitrary text for it (ie just fullname, name etc).

dotMorten:
Fixes #7978

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Parsers/TripleSlashCommentModel.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Parsers/TripleSlashCommentModel.cs
@@ -510,7 +510,7 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
                 var nodes = node.XPathSelectElements(nodeSelector + "[@cref]").ToList();
                 foreach (var item in nodes)
                 {
-                    var cref = item.Attribute("cref").Value;                    
+                    var cref = item.Attribute("cref").Value;
                     var success = false;
 
                     if (resolveCRef != null)
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Parsers/TripleSlashCommentModel.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Parsers/TripleSlashCommentModel.cs
@@ -565,17 +565,7 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
                                 }
                                 else
                                 {
-                                    StringBuilder text = new StringBuilder();
-                                    if(item.HasElements)
-                                    {
-                                        foreach (var textNode in item.Nodes())
-                                            text.Append(textNode.ToString());
-                                    }
-                                    else
-                                    {
-                                        text.Append(item.Value);
-                                    }
-                                    replacement = XElement.Parse($"<xref href=\"{HttpUtility.UrlEncode(id)}?text={HttpUtility.UrlEncode(text.ToString())}\" data-throw-if-not-resolved=\"false\"></xref>");
+                                    replacement = XElement.Parse($"<xref href=\"{HttpUtility.UrlEncode(id)}?text={HttpUtility.UrlEncode(item.Value)}\" data-throw-if-not-resolved=\"false\"></xref>");
                                 }
                                 item.ReplaceWith(replacement);
                             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Parsers/TripleSlashCommentModel.cs
```diff
@@ -558,7 +558,25 @@ private void ResolveCrefLink(XNode node, string nodeSelector, Action<string, str
...
+                                if(string.IsNullOrEmpty(item.Value))
+                                {
+                                    replacement = XElement.Parse($"<xref href=\"{HttpUtility.UrlEncode(id)}\" data-throw-if-not-resolved=\"false\"></xref>");
+                                }
+                                else
+                                {
+                                    StringBuilder text = new StringBuilder();
+                                    if(item.HasElements)
+                                    {
+                                        foreach (var textNode in item.Nodes())
```
yufeih, line 571: What if the node is a tree, like `<cref> 0 <b> 1 <i> 2 </i> 3 </b> 4 </cref>` ?

yufeih (commented):
Looks good except for inner text handling 👍 

@superyyrrzz to take another look

src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Parsers/TripleSlashCommentModel.cs
dotMorten: The test does test this and works, there are a lot of extra spaces added. Would love some direction on solving that a bit better

yufeih: A depth-first traversal of the tree to concat all leafs node text?

dotMorten: I don’t just want their text, but any html in that text

yufeih: We'll likely escape `?text` properties in render, does `?text=<b>a</b>` render as `*b*` or `<b>a</b>` according to your test?

dotMorten: `<b>` gets changed to `<strong>` and this gets encoded into the url. It would have made more sense if text goes inside the xref. ie `<xref ...>text is here</xref>`

dotMorten: You're right it is going to escape the html :( Better to just go with Value then and let it strip the xml

yufeih (approved):
Looks good.

superyyrrzz (approved):

dotMorten:
@yufeih  And thank you for the fast turn around. Is there a release scheduled in the near future?

yufeih:
Likely early next week, we typically don't ship at Friday (GMT+8).

928PJY:
@dotMorten @yufeih v2.59.2 has been released.

PR #8001, opened 2022-05-01
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveensrinivasan <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,7 +8,7 @@ updates:
   ignore:
   - dependency-name: docs.validation
   - dependency-name: docs.validation.analyzer
-  - package-ecosystem: "github-actions"
+- package-ecosystem: "github-actions"
     directory: "/"
     schedule: 
       interval: weekly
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -8,3 +8,8 @@ updates:
   ignore:
   - dependency-name: docs.validation
   - dependency-name: docs.validation.analyzer
+  - package-ecosystem: "github-actions"
```
yufeih, line 11: Thanks! Should we remove the indention here? otherwise this array is attached to the `ignore` object. 
```suggestion
- package-ecosystem: "github-actions"
```

yufeih (approved):
Thank you!

PR #8002, opened 2022-05-02
Added TypeScript link

jmswisher:
TypeScript is no longer "coming soon." It has its own page, so I linked to it.


yufeih (approved):
Thanks!

PR #8003, opened 2022-05-04
Use PdfConvertParallelism to control number of WkHtmlToPdf processes

AustinWise:
I observe that the WkHtmlToPdf.exe processes are not very CPU or memory intensive. This is probably due to the [3 second delay](https://github.com/dotnet/docfx/blob/9a3c8aa16adec1be4ee2e62ef45a9a0c78ccef14/src/Microsoft.DocAsCode.HtmlToPdf/HtmlToPdfOptions.cs#L94) added to the page load times in in DocFX's invocation of WkHtmlToPdf.

This change causes DocFX to uses the `--maxParallelism` to control the number of WkHtmlToPdf processes running at one time. By setting it to a large number (like 64), I am able to greatly speed up the build process.


yufeih (approved):
Thanks!

PR #8004, opened 2022-05-06
Remove SourcePath and add specific properties for redirection files

freewheel70:
[AB#595437](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/595437) 

Remove SourcePath and add specific properties for redirection files in .publish.json


src/docfx/build/redirection/RedirectionBuilder.cs
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
```
herohua, line 4: Is this `using` necessary?

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -154,34 +163,34 @@ public FilePath GetOriginalFile(FilePath file)
         return redirectUrls;
     }
 
-    private RedirectionItem[] LoadRedirectionModel()
+    private RedirectionItem[] LoadRedirectionModel(Dictionary<PathString, string> redirectConfigPaths)
```
herohua, line 166: why not adding `ConfigPath` as a property of `RedirectionItem`?

928PJY (approved):
:shipit:

yufeih:
Close this PR as we don't need these extra outputs.

PR #8014, opened 2022-05-12
Add instructions on how to install on a M1 Mac

sethreidnz:
This adds instructions to get docfx running on a Mac using an M1 chip. The solution is to use Rosetta to install brew and docfx so that it can run on the old architecture.


yufeih (approved):
Thanks Seth! This is perfect 👍

PR #8015, opened 2022-05-12
HTML attribute at div implements

DiCaver:
HTML attribute class at div implements was with one "s" too much. Changed from "classs" in to "class".


herohua (approved):

yufeih (approved):
Thank you!

PR #8038, opened 2022-06-03
Video Extension: Added functionality for allowing Video Entry GUID in Video Extension 

bharney0:
Now allowing users to enter a Video Entry Id (GUID) which will build the full URL.
 - `[!VIDEO 0add8ade-e746-4bf3-932a-94413928d098]` which will be converted to `https://learn-video.azurefd.net/vod/player?id=b63c2133-714a-48d7-9689-2120553664d4`
 - `:::video src="0add8ade-e746-4bf3-932a-94413928d098":::` which will be converted to `https://learn-video.azurefd.net/vod/player?id=b63c2133-714a-48d7-9689-2120553664d4`

@yufeih there were some questions about what URLs are allowed on the Video Extension? I was assuming the TrustedDomain code that pulls the taxonomy for AllowedDomains was still checking the video links? But that does not seem to be the case. I notice iframes get stripped from non-allowed domains but both video extensions allow them. That might be future work for code red.


yufeih:
@bharney0 Only links traversed by the [ResolveLinkExtension](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.MarkdigEngine.Extensions/ResolveLink/ResolveLinkExtension.cs) are checked against trusted domains, that method likely needs to understand some triple colon extension blocks.

yufeih (approved):

PR #8051, opened 2022-06-16
Design doc for Source Path complement 

freewheel70:
AB#617208


docs/designs/source-path-and-complement.md
```diff
@@ -0,0 +1,63 @@
...
+
+For most cases, the same source_path corresponds to the same source file (i.e. One-To-One). And the source_path value is the source file's physical file path.
+
+For some special cases, it is allowed to config mappings to resolve source_path values, named **SourceMap**  
+
+|  key             |  Value            |
+|------------------|-------------------|
+| source file path | source_path value |
+
+Due to **SourceMap**, multiple source files may have the same source_path (i.e. **Many-To-One**). And the source_path value may be NOT equal to the source file's physical file path.
```
herohua, line 18: For PowerShell, the source file is in .md but the actual input file to DocFX core is .yml. Please also take this scenario into consideration for this design.

herohua, line 18: What about service page which might come from either .md in repo, or generated .yml in memory. How are we going to address this scenario?

freewheel70: For powershell repo, MAML2Yaml preprocessor will be executed before building, it also leverages SourceMap.

freewheel70: FYI, https://github.com/dotnet/docfx/blob/c8cb790043ae5b93173f3e28dafc28bf7f305d48/src/docfx/config/ops/OpsPreProcessor.cs#L106

freewheel70: Service pages are generated when build from multiple source files (i.e. Toc files). Virtual file paths are generated for them, refer to code https://github.com/dotnet/docfx/blob/0db0fef6a31cdb5a6cb366abc81922e9df201508/src/docfx/build/toc/ServicePageGenerator.cs#L82 and https://github.com/dotnet/docfx/blob/0db0fef6a31cdb5a6cb366abc81922e9df201508/src/docfx/build/toc/ServicePageGenerator.cs#L167

herohua: So what will be the source_path and source_path_complement in such case?

freewheel70: In such case, source_path  is the virtual path (which is unique), source_path_complement  is null.

KalleOlaviNiemitalo:
When using DocFX v2, I made a preprocessor generate Markdown files with `_gitContribute` YAML metadata in them. That worked fine.

PR #8081, opened 2022-07-05
Support hostname overwritten from env var

freewheel70:
[AB#644583](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/644583)

### DIFF
--- a/src/docfx/config/ops/OpsConfigAdapter.cs
+++ b/src/docfx/config/ops/OpsConfigAdapter.cs
@@ -202,10 +202,6 @@ internal class OpsConfigAdapter
 
     private static string GetHostName(string siteName)
     {
-        if (!string.IsNullOrEmpty(EnvironmentVariable.HostName))
-        {
-            return EnvironmentVariable.HostName!;
-        }
         return siteName switch
         {
             "DocsAzureCN" => OpsAccessor.DocsEnvironment switch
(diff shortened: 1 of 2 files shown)


src/docfx/config/ops/OpsConfigAdapter.cs
```diff
@@ -202,6 +202,10 @@ private static (string repository, string branch) GetValidationServiceParameters
 
     private static string GetHostName(string siteName)
```
herohua, line 203: It is theoretically possible that multiple docsets in a repo belong to different sites. Can we verify whether this is the case by taking a look at existing provisioning data?

herohua: If it is true that we do have such cases, consider passing in the `siteName` -> `hostName` mapping instead of the `hostName` itself.

freewheel70: Since this is only for PPE

Consider rename to DOCFX_PPE_DEFAULT_DOMAIN_HOST_NAME

freewheel70: TODO:

add the env var from upstream

928PJY (approved):

herohua (approved):

PR #8082, opened 2022-07-06
remove domain for altrnative hostname in absolute links

AngryBerryMS:
[AB#645361](https://dev.azure.com/ceapex/Engineering/_workitems/edit/645361)

### DIFF
--- a/src/docfx/build/link/LinkResolver.cs
+++ b/src/docfx/build/link/LinkResolver.cs
@@ -155,6 +155,10 @@ internal class LinkResolver
                 }
                 return (errors, "", fragment, LinkType.AbsolutePath, null, false);
             }
+            if (resolvedHref.Contains(_config.AlternativeHostName, StringComparison.OrdinalIgnoreCase))
+            {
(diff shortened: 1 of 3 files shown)


src/docfx/build/link/LinkResolver.cs
```diff
@@ -155,6 +155,10 @@ internal class LinkResolver
                 }
                 return (errors, "", fragment, LinkType.AbsolutePath, null, false);
             }
+            if (resolvedHref.Contains(_config.AlternativeHostName, StringComparison.OrdinalIgnoreCase))
```
herohua, line 158: `Contains` is not safe enough. Is it possible to determine by its hose name?

herohua, line 158: It seems better to move the logic above line 146.

928PJY (approved):

src/docfx/build/link/LinkResolver.cs
AngryBerryMS: Fixed

herohua (approved):

PR #8085, opened 2022-07-06
remove redirection url hostname when match

freewheel70:
AB#638524

### DIFF
--- a/src/docfx/build/redirection/RedirectionProvider.cs
+++ b/src/docfx/build/redirection/RedirectionProvider.cs
@@ -127,23 +127,23 @@ internal class RedirectionProvider
             var monikers = item.Monikers is null ? default : _monikerProvider.Validate(_errors, item.Monikers);
             var filePath = FilePath.Redirection(path, monikers);
 
-            if (item.RedirectDocumentId)
+            switch (UrlUtility.GetLinkType(absoluteRedirectUrl))
             {
-                switch (UrlUtility.GetLinkType(absoluteRedirectUrl))
-                {
-                    case LinkType.RelativePath:
+                case LinkType.RelativePath:
+                    if (item.RedirectDocumentId)
+                    {
                         var siteUrl = _documentProvider.GetSiteUrl(filePath);
                         absoluteRedirectUrl = PathUtility.Normalize(Path.Combine(Path.GetDirectoryName(siteUrl) ?? "", absoluteRedirectUrl));
-                        break;
-                    case LinkType.AbsolutePath:
-                        break;
-                    case LinkType.External:
-                        absoluteRedirectUrl = UrlUtility.RemoveLeadingHostName(absoluteRedirectUrl, hostName, removeLocale: true);
-                        break;
-                    default:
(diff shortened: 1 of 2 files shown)


src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -127,23 +127,23 @@ public FilePath GetOriginalFile(FilePath file)
...
-            if (item.RedirectDocumentId)
+            switch (UrlUtility.GetLinkType(absoluteRedirectUrl))
             {
-                switch (UrlUtility.GetLinkType(absoluteRedirectUrl))
-                {
-                    case LinkType.RelativePath:
+                case LinkType.RelativePath:
+                    if (item.RedirectDocumentId)
+                    {
                         var siteUrl = _documentProvider.GetSiteUrl(filePath);
```
herohua, line 135: Should we also output absolute path in this case even `redirect_document_id = false`?

freewheel70: Not sure, but wanna keep the same logic as before

freewheel70: Will take investigation if I have bandwidth

freewheel70: @herohua  
After reading some related code, I cannot still make sure whether we should make this change.
The logic could go back to 2 years agon (refer to https://github.com/dotnet/docfx/commit/b458df2aa001b4345c21bb42b62bf8d87f4418b1#diff-e52f79a0fe78059ae18d7615b24eafb6249323757b8a639c9eab32ba4e3a6fc7 ) 

As it may be a different task, let's discuss about it later and unblock current PR first.

What do you think?

928PJY (approved):
:shipit:

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -127,23 +127,23 @@ public FilePath GetOriginalFile(FilePath file)
...
-                        absoluteRedirectUrl = UrlUtility.RemoveLeadingHostName(absoluteRedirectUrl, hostName, removeLocale: true);
-                        break;
-                    default:
-                        _errors.Add(Errors.Redirection.RedirectUrlInvalid(path, redirectUrl));
-                        break;
-                }
+                    }
+                    break;
+                case LinkType.AbsolutePath:
+                    break;
```
928PJY, line 140: let's also cover this case, the condition has been updated.

herohua (changes requested):
Please check the open comments. Thanks!

src/docfx/build/redirection/RedirectionProvider.cs
freewheel70: By checking the documentId generation logic, I don't think this change will affect documentId

https://github.com/dotnet/docfx/blob/fbe1cfd157469bf28f59ddf81f47b3b235b363f7/src/docfx/build/document/DocumentProvider.cs#L166

herohua (approved):

Issue #6791, opened 2020-11-20
Doc generation for TypeScript not working even with the basic example


Ellerbach:
**Operating System**:  `Windows`

**DocFX Version Used**: choco install docfx -y

**Template used**: `default`

**Steps to Reproduce**:

I've done **exactly** the steps from the doc: https://dotnet.github.io/docfx/tutorial/universalreference/gen_doc_for_ts.html
TypeScript 3.9 or TypeScript 4.0 give the same result

**Expected Behavior**:

Generate the docs.

**Actual Behavior**:

The api.json is generated
The yml files are all generated.
Only TOC generated in _site. All the rest is ignored.
Tested also on some other TypeScript, well documented application and same issue. Only TOC is generated.

```plain
[20-11-20 05:47:44.305]Info:[BuildCommand]6 plug-in(s) loaded.
[20-11-20 05:47:44.383]Info:[BuildCommand]Markdown engine is dfm
[20-11-20 05:47:44.548]Info:[BuildCommand.BuildCore.Build Document]Max parallelism is 8.
[...]

PR #8099, opened 2022-07-19
Update gen_doc_for_ts.md

mode777:
Typescript example is broken as discussed here #6791 (https://github.com/dotnet/docfx/issues/6791). Re-verting to older npm package 0.10.5 for type2docfx fixes it for now.


paulushub:
@mode777 Downgrading should not be a solution. How about fixing `type2docfx`?

mode777:
@paulushub I agree but the currently released version of type2docfx on npm already seems to target the yet unreleased docfx v3. The "proper" solution should be unlisting on npm...

herohua (approved):

PR #8108, opened 2022-07-26
auto detect redirection files

AngryBerryMS:
[AB#655666](https://dev.azure.com/ceapex/Engineering/_workitems/edit/655666)


test/docfx.Test/cli/NewTest.cs
```diff
@@ -12,6 +12,7 @@ public static class NewTest
 
     static NewTest()
     {
+        TestQuirks.Verbose = true;
```
herohua, line 15: Do we need this change?

src/docfx/cli/New.cs
```diff
@@ -83,7 +83,15 @@ private static bool CreateFromTemplate(string templateName, CommandLineOptions o
             }
             else
             {
-                Directory.CreateDirectory(Path.GetDirectoryName(targetFullPath) ?? ".");
+                var directory = Path.GetDirectoryName(targetFullPath) ?? ".";
+                if (!Directory.Exists(directory))
+                {
+                    Directory.CreateDirectory(directory);
+                    if (TestQuirks.Verbose ?? false)
```
herohua, line 90: It seems weird to initialize a folder to Git when `Verbose == true`.

AngryBerryMS: It's because we want to scan all redirection files within the repo folder, the `_buildOptions.repopath`  is from [here (link)](https://github.com/dotnet/docfx/blob/6163e1c90c8f316789051d511bb1f8a9636ecae5/src/docfx/lib/git/GitUtility.cs#L17), and for these tests without any repo, it will traverse upwards until find a git directory.

For these tests, it will find docfx itself is a git directory, and then we will get all redirection files in all spec tests, which will fail lots tests. So I just want to make sure the git directory is limited within the test folder.

image

src/docfx/config/ops/OpsConfig.cs
```diff
@@ -12,8 +12,6 @@ internal class OpsConfig
 {
     public OpsDocsetConfig[] DocsetsToPublish { get; init; } = Array.Empty<OpsDocsetConfig>();
 
-    public HashSet<PathString> RedirectionFiles { get; init; } = new HashSet<PathString>();
```
freewheel70, line 15: Should we remind user to remove their RedirectionFiles  config? 
e.g. give some warning/suggestions

src/docfx/cli/New.cs
herohua: If so, would it be better to use a separate flag?

src/docfx/config/ops/OpsConfig.cs
AngryBerryMS: As discussed with @herohua before, it's acceptable not reporting a warning during the build. We will as PM to update the tutorial to inform writers that this config is not needed anymore. But we won't force them to remove this config

src/docfx/cli/New.cs
AngryBerryMS: Done

src/docfx/config/TestQuirks.cs
```diff
@@ -14,4 +14,6 @@ internal static class TestQuirks
     public static Func<string, string?>? HttpProxy { get; set; }
 
     public static bool? Verbose { get; set; }
+
+    public static bool? Initializable { get; set; }
```
herohua, line 18: I tend to think this could be a new parameter to `docfx new`.

AngryBerryMS: Currently we only use it for tests. In our real usage, there won't be such scenario since all of our contents are located in a repository.

yufeih: Why does implementing this feature require such a flag?

docs/specs/git.yml
```diff
@@ -62,13 +62,19 @@ outputs:
...
 usePhysicalInput: true
-inputs:
-  docs/index.md: |
-    ---
-    author: invalid_user
-    ---
-    # git-test
-  docfx.yml:
+repos:
+  https://github.com/testowner/testrepo-with-invalid-user#main:
```
928PJY, line 66: why we need to update this test case?

AngryBerryMS: Because I cannot find an easy to prevent it to finally find the Docfx repo itself when dinIt's because we want to scan all redirection files within the repo folder, the _buildOptions.repopath is from [here (link)](https://github.com/dotnet/docfx/blob/6163e1c90c8f316789051d511bb1f8a9636ecae5/src/docfx/lib/git/GitUtility.cs#L17), and for this test, I cannot find an easy to prevent it from finally finding the Docfx repo itself while let it use the cache.

PR #8116, opened 2022-08-01
dangling redirection files

AngryBerryMS:
[AB#681913](https://dev.azure.com/ceapex/Engineering/_workitems/edit/681913)

### DIFF
--- a/src/docfx/Errors.cs
+++ b/src/docfx/Errors.cs
@@ -367,6 +367,12 @@ internal static class Errors
         /// Behavior: ✔️ Message: ✔️
         public static Error RedirectionPathSyntaxError(SourceInfo<string> source)
             => new(ErrorLevel.Warning, "redirection-path-syntax-error", $"Redirection path syntax is incorrect. 'source_path' should start without '/' and 'source_path_from_root' should start with '/'.", source);
+
+        /// <summary>
+        /// Check whether there are any dangling redirection files in a repository.
+        /// </summary>
+        public static Error DanglingRedirectionFiles(IEnumerable<string> danglingRedirectionFiles)
+            => new(ErrorLevel.Warning, "dangling-redirection-files", $"There are some dangling redirection files: {string.Join(", ", danglingRedirectionFiles)}");
     }
 
     public static class Toc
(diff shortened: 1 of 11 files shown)


src/docfx/Errors.cs
```diff
@@ -367,6 +367,12 @@ public static Error SourcePathConflict(SourceInfo<string> source)
         /// Behavior: ✔️ Message: ✔️
         public static Error RedirectionPathSyntaxError(SourceInfo<string> source)
             => new(ErrorLevel.Warning, "redirection-path-syntax-error", $"Redirection path syntax is incorrect. 'source_path' should start without '/' and 'source_path_from_root' should start with '/'.", source);
+
+        /// <summary>
+        /// Check whether there are any dangling redirection files in a repository.
+        /// </summary>
+        public static Error DanglingRedirectionFiles(IEnumerable<string> danglingRedirectionFiles)
```
herohua, line 374: Let's make this as `ErrorLevel.Info` to avoid impact to end users.

src/docfx/build/redirection/RedirectionProvider.cs
```diff
@@ -167,9 +174,20 @@ private RedirectionItem[] LoadRedirectionModel()
 
         foreach (var fullPath in ProbeSubRedirectionFiles())
         {
+            if (_autoScannedRedirectionFiles.Contains(fullPath))
```
herohua, line 177: nit: no need to check `Contains` first

src/docfx/Errors.cs
AngryBerryMS: done

src/docfx/build/redirection/RedirectionProvider.cs
AngryBerryMS: fixed

src/docfx/cli/New.cs
```diff
@@ -83,6 +83,15 @@ private static bool CreateFromTemplate(string templateName, CommandLineOptions o
             }
             else
             {
+                var directory = Path.GetDirectoryName(targetFullPath) ?? ".";
+                if (!Directory.Exists(directory))
+                {
+                    Directory.CreateDirectory(directory);
+                    if (TestQuirks.Initializable ?? false)
```
herohua, line 90: Would it be better to add this param to `docfx new`?

AngryBerryMS: I don't think so. Initializing a git directory is not required when creating a new docset. This is only for test usage

herohua (approved):

Issue #5343, opened 2019-12-05
Triple slash comment 'item' tag generates an empty list item


martinivanoff:
The following C# XML comment generates a list with two empty list items.

XML:

`///<list type="bullet">`
`///       <item>Item 1</item>`
`///       <item>Item 2</item>`
`///</list>`

Generated YAML:

`remarks: "\n<ul><li></li><li></li></ul>\n"`

vwxyzh:
See: https://docs.microsoft.com/en-us/dotnet/csharp/codedoc#list

Source code:
https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Transform/TripleSlashCommentTransform.xsl#L188

KalleOlaviNiemitalo:
[`list` in Sandcastle XML Comments Guide](http://ewsoftware.github.io/XMLCommentsGuide/html/e433d846-db15-4ac8-a5f5-f3428609ae6c.htm) says that, for `list/@type="bullet"` and `list/@type="number"`, the inner text of the `item` element is rendered if the `term` and `description` child elements are omitted. "Area-SandcastleParity" might be appropriate.

craigfowler:
It is not only Sandcastle which behaves in this way.  VS Code for example (with Omnisharp) will render list contents as described in this issue.  Here's a screenshot of a tool-tip from an editor.

<img src="https://user-images.githubusercontent.com/982047/103692718-17e99100-4f90-11eb-8319-f0e887af3bdb.png" alt="VSCode-list-tag" height="400">

I hadn't even realised that the `<term>` and `<description>` tags existed.  I've been using bare `<item>` for so long (and supported by every tool I'd used) that I was unaware that technically I was doing anything wrong.  So, I'd support including this feature based on the principle of least astonishment.

For now, though (now I know) I'm happy with the workaround that I must use term & description tags.

eli-darkly:
I was also blindsided by this bug, and will now have to edit the docs in quite a few projects to add `<description>` tags.

TermWay:
If Visual Studio support this, I think it's fair to say that docfx should also support it.

dotMorten:
Nice to see a PR has been submitted to address this. It's been quite an issue for us.

PR #8132, opened 2022-08-12
Handle list items without either <term> or <description>

mstefarov:
Items in lists should be rendered even if they lack "term" and "description" tags.  This is needed for parity with SandCastle and IntelliSense.  This only impacts lists of type "bullet" and "number" -- tables are not affected.

The simplest way to implement this _would have been_ with a [single indiscriminate `<xsl:apply-templates />` inside `<li>`](https://gist.github.com/mstefarov/2134915ad38671ff3a519a25ceaf9513/revisions#diff-5abcc86a43ac8e6519682f8069fc76538ddb2f000d6a34ec501765cb5fe92ce7), but that would have changed rendered output for existing items that _do_ contain term/description tags.  Specifically, this would force preservation of insignificant whitespace around term/description tags in the rendered output.

So, I went with the slightly longer XSLT implementation that minimizes changes to output.

#5343

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Transform/TripleSlashCommentTransform.xsl
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Transform/TripleSlashCommentTransform.xsl
@@ -195,8 +195,15 @@
             <ul>
               <xsl:for-each select="item">
                 <li>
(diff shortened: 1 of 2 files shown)


dotMorten (approved):
Nice work!

herohua (approved):

PR #8140, opened 2022-08-19
Generate canonical URL with canonical moniker in DocFX

freewheel70:
[AB#678710](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/678710)

### DIFF
--- a/docs/specs/escape.yml
+++ b/docs/specs/escape.yml
@@ -39,7 +39,7 @@ outputs:
     }
   'a b.json': |
     {
-      "canonical_url": "https://docs.com/en-us/a b",
+      "canonical_url": "https://docs.com/en-us/a%20b",
       "gitcommit": "https://github.com/escape/markdown/blob/c4c07cb69d114fa9ba16e55763d47f2344083dad/a b.md",
       "content_git_url": "https://github.com/escape/markdown/blob/main/a b.md",
       "original_content_git_url": "https://github.com/escape/markdown/blob/main/a b.md",
--- a/src/docfx/build/document/DocumentProvider.cs
+++ b/src/docfx/build/document/DocumentProvider.cs
@@ -196,7 +196,7 @@ internal class DocumentProvider
         var renderType = _jsonSchemaProvider.GetRenderType(contentType, mime);
         var sitePath = FilePathToSitePath(path, contentType, _config.UrlType, renderType);
         var siteUrl = PathToAbsoluteUrl(Path.Combine(_config.BasePath, sitePath), contentType, _config.UrlType, renderType);
-        var canonicalUrl = GetCanonicalUrl(sitePath);
+        var canonicalUrl = GetCanonicalUrl(siteUrl);
         var outputPath = GetOutputPath(path, sitePath, contentType, renderType);
 
         return new Document(sitePath, siteUrl, outputPath, canonicalUrl, contentType, mime, renderType);
@@ -256,6 +256,7 @@ internal class DocumentProvider
             }
             if (urlType == UrlType.Docs && contentType != ContentType.Toc)
             {
+                // remove extension
                 var i = url.LastIndexOf('.');
                 return i >= 0 ? url[..i] : url;
             }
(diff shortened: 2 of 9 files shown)


src/docfx/build/document/DocumentProvider.cs
```diff
@@ -264,12 +264,53 @@ private static string PathToRelativeUrl(string path, ContentType contentType, Ur
...
+        {
+            return "";
+        }
+
+        var canonicalUrlPrefix = UrlUtility.Combine($"https://{_config.HostName}", _buildOptions.Locale, _config.BasePath);
+
+        var canonicalUrl = canonicalUrlPrefix + "/" + RemoveExtension(EncodePath(sitePath));
+        canonicalUrl = canonicalUrl.ToLowerInvariant();
+
+        if (canonicalUrl.EndsWith("/index"))
```
herohua, line 281: It seems this part of logic is already handled by https://github.com/dotnet/docfx/blob/3f0d11c66cc11ce28597f668040e792fd7fcec7c/src/docfx/build/document/DocumentProvider.cs#L241-L241. Any chance to reuse original signature of `siteUrl`?

src/docfx/build/page/PageBuilder.cs
```diff
@@ -261,6 +266,27 @@ private SystemMetadata CreateSystemMetadata(ErrorBuilder errors, FilePath file,
...
+        if (pageType == "profile")
+        {
+            return null;
+        }
+
+        var canonicalUrl = _documentProvider.GetCanonicalUrl(file);
+
+        if (_config.IsCanonicalUrlWithMoniker)
+        {
+            var canonicalVersion = _publishUrlMap.GetCanonicalVersion(file);
```
herohua, line 280: better to URL encode `canonicalVersion`

src/docfx/build/document/DocumentProvider.cs
herohua: If the only concern is encoding, please consider including the encoding logic in the generator of `siteUrl`.

src/docfx/build/metadata/SystemMetadata.cs
```diff
@@ -59,6 +59,9 @@ internal class SystemMetadata
     [JsonProperty("_op_canonicalUrlPrefix")]
     public string? CanonicalUrlPrefix { get; set; }
 
+    [JsonProperty("_op_canonicalUrl")]
+    public string? OpCanonicalUrl { get; set; }
```
928PJY, line 63: what does this metadata used for?

src/docfx/build/page/PageBuilder.cs
```diff
@@ -261,6 +266,27 @@ private SystemMetadata CreateSystemMetadata(ErrorBuilder errors, FilePath file,
         return systemMetadata;
     }
 
+    private string? GetCanonicalUrlWithMoniker(string? pageType, FilePath file)
```
928PJY, line 269: the function name means you will append a moniker, but it may not

docs/specs/escape.yml
```diff
@@ -39,7 +39,7 @@ outputs:
     }
   'a b.json': |
     {
-      "canonical_url": "https://docs.com/en-us/a b",
+      "canonical_url": "https://docs.com/en-us/a%20b",
```
928PJY, line 42: will you add test case to cover the moniker query?

src/docfx/build/metadata/SystemMetadata.cs
freewheel70: `_op_canonicalUrl is in the blob content, which is later used in HTML tags. `

please check description of work item 664227

freewheel70: Before, the _op_canonicalUrl is generated by JINT template js directly but not defined on DocFX

src/docfx/build/document/DocumentProvider.cs
freewheel70: code updated, please review again

src/docfx/build/page/PageBuilder.cs
```diff
@@ -261,6 +266,27 @@ private SystemMetadata CreateSystemMetadata(ErrorBuilder errors, FilePath file,
         return systemMetadata;
     }
 
+    private string? GetCanonicalUrlWithMonikerIfNecessary(string? pageType, FilePath file)
+    {
+        if (pageType == "profile")
```
928PJY, line 271: is this special logic come from docs.ui as well?

src/docfx/build/page/PageBuilder.cs
```diff
@@ -261,6 +266,27 @@ private SystemMetadata CreateSystemMetadata(ErrorBuilder errors, FilePath file,
         return systemMetadata;
     }
 
+    private string? GetCanonicalUrlWithMonikerIfNecessary(string? pageType, FilePath file)
```
herohua, line 269: There is duplicate code between `GetCanonicalUrlWithMonikerIfNecessary` and `GetCanonicalUrl`. How about defining this API as `AppendCanonicalMonikerIfNecessary`?

src/docfx/build/page/PageBuilder.cs
herohua: Can you double confirm that for `Profile`, no canonical URL is needed to generate?

freewheel70: sure

freewheel70: @herohua  
I just check the output result of https://github.com/MicrosoftDocs/DocsRoot/blob/972fb584bff2eb20b14a649fe41ed4950d74c04f/profile/edit.md?plain=1#L3 
there is no canonical_url generated, which means the scenario exists. 

But I am not sure whether the business logic is still valid, but I think that is out of scope of this task. 

I guess we would better to keep the logic to avoid breaking something and we may investigate it later as we have some other urgent tasks in progress.

928PJY (approved):

PR #8146, opened 2022-08-24
Generate canonical URL with canonical moniker in DocFX

freewheel70:
[AB#678710](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/678710)

The diffs shown in regression tests are all due to:
1. Either the asset is redirection rule;
2. Or the output mode is ugly.


herohua:
Is it possible to stabilize the order of metadata in regression test comparison so that we can tell whether this change causes unexpected breaking changes? Currently the number of diffs is too high to be verified.

docs/specs/escape.yml
```diff
@@ -39,28 +41,32 @@ outputs:
...
   a&amp;b.json: |
     {
       "canonical_url": "https://docs.com/en-us/a&amp;b",
-      "gitcommit": "https://github.com/escape/markdown/blob/c4c07cb69d114fa9ba16e55763d47f2344083dad/a&amp;b.md",
+      "gitcommit": "https://github.com/escape/markdown/blob/03cb6e2fa19c69303dd589df63df0c40448b241b/a&amp;b.md",
       "content_git_url": "https://github.com/escape/markdown/blob/main/a&amp;b.md",
       "original_content_git_url": "https://github.com/escape/markdown/blob/main/a&amp;b.md",
       "original_content_git_url_template": "{repo}/blob/{branch}/a&amp;b.md"
     }
+  a – b c/ab.json: |
```
herohua, line 66: Do we know why `-` is encoded? I think it is a valid char for URL per [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986).

src/docfx/build/page/PageBuilder.cs
```diff
@@ -261,6 +266,27 @@ private SystemMetadata CreateSystemMetadata(ErrorBuilder errors, FilePath file,
...
+        }
+
+        var canonicalUrl = _documentProvider.GetCanonicalUrl(file);
+
+        if (_config.IsCanonicalUrlWithMoniker)
+        {
+            var canonicalVersion = _publishUrlMap.GetCanonicalVersion(file);
+            if (!string.IsNullOrEmpty(canonicalVersion))
+            {
+                return $"{canonicalUrl}?view={Uri.EscapeDataString(canonicalVersion)}";
```
herohua, line 283: Should use the same logic as in `EncodePath`

freewheel70: sure

src/docfx/build/document/DocumentProvider.cs
```diff
@@ -256,20 +256,22 @@ private static string PathToRelativeUrl(string path, ContentType contentType, Ur
...
-    private string GetCanonicalUrl(string siteUrl)
+    private string GetCanonicalUrl(string? siteUrl)
     {
-        return $"https://{_config.HostName}/{_buildOptions.Locale}{siteUrl}";
+        if (siteUrl == null)
+        {
+            return "";
+        }
+
+        return $"https://{_config.HostName}/{_buildOptions.Locale}{UrlUtility.EscapeUrlPath(siteUrl).ToLowerInvariant()}";
```
herohua, line 274: Do we want to lower case the query string?

freewheel70: at this moment, the return value still has no query value (i.e. moniker) 
The moniker query is added on PageBuilder

herohua (approved):

PR #8156, opened 2022-08-30
[Don't merge, POC] Link trans poc

shanluo911:
This is a poc of transferring external/ absolute link to relative link if possible. Considering this is only a poc, refactoring work is skipped.

Test:
Created two spec tests for external and absolute link, and they both passed. All other spec tests also passed, so this did not break any other logic.

### DIFF
--- a/src/docfx/build/link/LinkResolver.cs
+++ b/src/docfx/build/link/LinkResolver.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Collections.Concurrent;
+using System.IO;
+using System.Linq;
 using Microsoft.Docs.Validation;
 
 namespace Microsoft.Docs.Build;
@@ -95,7 +97,6 @@ internal class LinkResolver
             inclusionRoot,
             linkNode,
             tagName);
-
         inclusionRoot ??= referencingFile;
         if (!isCrossReference)
         {
@@ -113,6 +114,7 @@ internal class LinkResolver
 
         if (!string.IsNullOrEmpty(link))
         {
+            Telemetry.TrackLink(linkType.ToString());
             _fileLinkMapBuilder.AddFileLink(inclusionRoot, referencingFile, link, href.Source);
         }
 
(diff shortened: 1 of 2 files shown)


src/docfx/build/link/LinkResolver.cs
```diff
@@ -350,4 +392,23 @@
...
         _contentValidator.ValidateLink(file, node);
     }
+
+    private bool TryResolveAbsolutePathAsRelativePath(string absolutePath, FilePath referencingFile, bool lookupFallbackCommits, bool contentFallback, out FilePath? relativePath)
+    {
+        var absolutePathReplaceSlash = absolutePath.TrimStart(new char[] { '\\', '/' }).Replace('\\', '/');
+        var basePath = absolutePathReplaceSlash.Split('/')[0];
+        if (string.Equals(_config.BasePath, basePath, StringComparison.Ordinal))
+        {
+            var path = TryResolveRelativePath(
```
AngryBerryMS, line 402: Seems this method may return `null`, so when the returned `path` is null we still return true?

src/docfx/build/link/LinkResolver.cs
```diff
@@ -350,4 +392,23 @@
 
         _contentValidator.ValidateLink(file, node);
     }
+
+    private bool TryResolveAbsolutePathAsRelativePath(string absolutePath, FilePath referencingFile, bool lookupFallbackCommits, bool contentFallback, out FilePath? relativePath)
+    {
+        var absolutePathReplaceSlash = absolutePath.TrimStart(new char[] { '\\', '/' }).Replace('\\', '/');
+        var basePath = absolutePathReplaceSlash.Split('/')[0];
```
AngryBerryMS, line 399: absolute path may contain locale. Please add a test like `[https://docs.microsoft.com/en-us/a`](https://docs.microsoft.com/en-us/a`)

src/docfx/build/link/LinkResolver.cs
```diff
@@ -157,7 +158,48 @@
...
+            {
+                if (TryResolveAbsolutePathAsRelativePath(resolvedHref, hrefRelativeTo, _buildScope.GetContentType(resolvedHref) == ContentType.Resource, true, out var relativePath))
+                {
+                    if (relativePath == null)
+                    {
+                        errors.Add(Errors.Link.FileNotFound(new SourceInfo<string>(href, href)));
+                    }
+                    else
+                    {
+                        href = new SourceInfo<string>(resolvedHref.TrimStart(new char[] { '/', '\\' }).Remove(_config.BasePath.ToString().Length));
```
AngryBerryMS, line 171: In this scenario, we should also remove locale if it exists

src/docfx/build/link/LinkResolver.cs
```diff
@@ -157,7 +158,48 @@
...
+        {
+            if (TryResolveAbsolutePathAsRelativePath(href, hrefRelativeTo, _buildScope.GetContentType(href) == ContentType.Resource, true, out var relativePath))
+            {
+                if (relativePath == null)
+                {
+                    errors.Add(Errors.Link.FileNotFound(new SourceInfo<string>(href, href)));
+                }
+                else
+                {
+                    href = new SourceInfo<string>(href.Value.TrimStart(new char[] { '/', '\\' }).Remove(_config.BasePath.ToString().Length));
```
AngryBerryMS, line 198: In this scenario, we should also remove locale if it exists

src/docfx/build/link/LinkResolver.cs
```diff
@@ -157,7 +158,48 @@
...
+                {
+                    if (relativePath == null)
+                    {
+                        errors.Add(Errors.Link.FileNotFound(new SourceInfo<string>(href, href)));
+                    }
+                    else
+                    {
+                        href = new SourceInfo<string>(resolvedHref.TrimStart(new char[] { '/', '\\' }).Remove(_config.BasePath.ToString().Length));
+                        file = relativePath;
+                        linkType = LinkType.RelativePath;
```
AngryBerryMS, line 173: Why don't we return href in this case? Do we require the later process in this method?

PR #8164, opened 2022-09-02
Get GitHub access token from build api

irushx:


928PJY (approved):
:shipit:

herohua (changes requested):
temporarily holding the approval

herohua (approved):
You can start testing on PubDev now.

docs/specs/git.yml
```diff
@@ -459,8 +459,6 @@ repos:
           Osmond creates this
 inputs:
   docfx.yml: |
-    secrets:
```
928PJY, line 462: We should cover the two cases in spec test:
- token from secrets
- token from API call.

928PJY (approved):

PR #8197, opened 2022-09-28
feat: throw error if template branch isn't main or master or default

herohua:
[AB#713280](https://dev.azure.com/ceapex/d3d54af3-265a-4f18-95f6-9a46397ca583/_workitems/edit/713280)

Design considerations:
- Override with `main` vs. throwing error: Prefer throwing error so that content team will be fully aware of this change. They will then open a PR for fix and during the PR they can double check whether the rendered results are expected.
- Why do we need to pass in a flag: Local run of DocFX for developers and DocFX run in learn-validation extension will still want to have the flexibility to be able to use non-`main` branch template. Thus, a flag is invented and will only be set to `true` for real server build for `live` branch build and PR builds towards `live` branch.

### DIFF
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -547,3 +547,12 @@ outputs:
     {"document_id":"c871e0da-be84-d8ce-ce09-6d935e69b172","document_version_independent_id":"2c52aff4-0602-cfcd-e12f-269c5629bdd8"}
   src/dir1/a.json: |
     {"document_id":"b998d1b1-aaed-698d-7b82-0bc93b0b7ba5","document_version_independent_id":"2db8730f-5266-6c4f-6a6b-7db32fcff3b3"}
+---
+# template branch must be main or master or default when validation is enabled
+inputs:
+  docfx.yml: |
+    validateTemplateBranch: true
+    template: https://docs.com/template#test
+outputs:
+  .errors.log: |
+    {"message_severity":"error","code":"template-branch-invalid"}
--- a/docs/specs/config.yml
+++ b/docs/specs/config.yml
@@ -552,7 +552,23 @@ outputs:
 inputs:
   docfx.yml: |
     validateTemplateBranch: true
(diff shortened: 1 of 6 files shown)


src/docfx/lib/path/PackagePath.cs
```diff
@@ -28,6 +28,8 @@ internal record PackagePath
 
     public bool IsMainOrMaster => Branch == "main" || Branch == "master";
 
+    public bool IsMainOrMasterOrDefault => Branch == "main" || Branch == "master" || Branch == "";
```
freewheel70, line 31: nit: IsMainOrMasterOrDefault  -> IsMainOrMasterOrEmpty

docs/specs/config.yml
```diff
@@ -547,3 +547,12 @@ outputs:
...
     {"document_id":"b998d1b1-aaed-698d-7b82-0bc93b0b7ba5","document_version_independent_id":"2db8730f-5266-6c4f-6a6b-7db32fcff3b3"}
+---
+# template branch must be main or master or default when validation is enabled
+inputs:
+  docfx.yml: |
+    validateTemplateBranch: true
+    template: https://docs.com/template#test
+outputs:
+  .errors.log: |
+    {"message_severity":"error","code":"template-branch-invalid"}
```
freewheel70, line 558: nit: include error message

freewheel70 (commented):
a few minor suggestions

src/docfx/lib/path/PackagePath.cs
herohua: I tend to prefer `default` because if there is no branch associated with the repo URL, the branch you'll get by `git clone` is called the default branch. What do you think?

freewheel70: I see

freewheel70 (approved):

AngryBerryMS (approved):

Issue #7884, opened 2022-02-11
DocFX crashes on method parameter [Optional][DefaultParameterValue(0)] double PWFAAmount


harriev9:
**Operating System**: Windows

**DocFX Version Used**: Latest from github compiled.

**Template used**: (`default` or `statictoc` or contain custom template)

**Steps to Reproduce**:

1. Make a lib project with one class with content:

using System.Runtime.InteropServices;

namespace TestLib
{
    /// <summary>
    /// This is a class for debugging the problem with [DefaultParameterValue(0)]
    /// DocFX craches. The 0 is internally represented by an object, which seems to be derived from int and than is is tried to cast it to a double.
    /// </summary>
    [ComVisible(true)]
    public class TestClass
    {
        /// <summary>
        /// Register a component as ready.
        /// </summary>
        /// <param name="BatchName">Batch name</param>
        /// <param name="UnitId">Unit</param>
        /// <param name="UniqueId">Unique Id</param>
        /// <param name="ActualAmount">The amount to register</param>
        /// <param name="ErrorCode">Error code</param>
        /// <param name="ExtraAmount">Extra amount</param>
        /// <param name="AmountUOM">UOM of the amount</param>
        ///
        public void Registrate(string BatchName, string UnitId, int UniqueId, double ActualAmount, int ErrorCode, 
            [Optional][DefaultParameterValue(0)] double ExtraAmount, [Optional][DefaultParameterValue("kg")] string AmountUOM)
        {
        }
    }
}

2.  Generate documentation for it.
3.  DocFX will throw an exception "System.InvalidCastException: 'Specified cast is not valid.'" in the method GetLiteralExpressionCore:

on line 1132:

                    return SyntaxFactory.LiteralExpression(
                        SyntaxKind.NumericLiteralExpression,
                        SyntaxFactory.Literal((double)value));

value is an object (according to the debugger object from an int).
casting an object which came from an int to double is not possible.

**Expected Behavior**: Just to create documentation.

**Actual Behavior**: Exception thrown and DocFX stops with an error.

When i change the 0 in [Optional][DefaultParameterValue(0)] into 0d  ([Optional][DefaultParameterValue(0d)]) then DocFX behaves as expected.

PR #8208, opened 2022-10-10
#7884: Fix on DocFX crashes on method parameter [Optional][DefaultPar…

harriev9:
When a default on a double is given as 0 instead of 0d than DocFX fails (exit code <> 0). See the issue number (7884).

On a construct like:

public void Registrate(string BatchName, string UnitId, int UniqueId, double ActualAmount, int ErrorCode,
[Optional][DefaultParameterValue(0)] double ExtraAmount, [Optional][DefaultParameterValue("kg")] string AmountUOM)

DocFX will crash.


src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/Visitors/CSYamlModelGenerator.cs
```diff
@@ -1127,6 +1127,12 @@ public static ExpressionSyntax GetLiteralExpressionCore(object value, ITypeSymbo
                         SyntaxKind.NumericLiteralExpression,
                         SyntaxFactory.Literal((float)value));
                 case SpecialType.System_Double:
+                    if (value is int @int)
```
yufeih, line 1130: Will `value` be a `float` if the default value is `0f`? If so consider use `Convert.ToDouble`

PR #8306, opened 2022-12-21
feat: New extensibility model based on libraries

yufeih:
This is a new extensibility model by shipping docfx as a library. It changes the extensibility model from runtime to build time:
- `Microsoft.DocAsCode` is a new library package with public APIs to build a project with the exact same behavior as executing the `docfx` CLI.
- The `docfx` package is a thin command line wrapper around `Microsoft.DocAsCode`.
- To enable extensibility, create a new EXE project and add `Microsoft.DocAsCode` package
- It should also work seamlessly with more script-oriented build systems like  `dotnet-script` or `cake`.
- Extensions like `memberpage` are shipped as ordinary NuGet libraries, to use an extension, add it to the project using `<PackageReference>`.
- More extensibility options like markdown extensions, template extensions can be exposed through `Microsoft.DocAsCode` with additional APIs for customization.

See `samples/extensions` folder on an EXE project sample.

This approach solves the problems of sharing+consuming extensions and assembly loading issues with the plugin approach.
Loading plugins dynamically was removed with this change.

> NOTE: This PR makes `memberpage` C# code work, but the `memberpage` templates are not yet solved.

The initial APIs exposed by this PR are:

```csharp
namespace Microsoft.DocAsCode;

public class Docset
{
    public static Task Build(string configPath);
}

```

cc @vicancy


src/Microsoft.DocAsCode.App/DocfxProject.cs
```diff
@@ -0,0 +1,100 @@
...
+    {
+        /// <summary>
+        /// Gets a read-only, singleton instance of <see cref="BuildOptions"/> that uses the default configuration.
+        /// </summary>
+        public static BuildOptions Default { get; } = new();
+
+        /// <summary>
+        /// Get or sets a value that indicates whether to export model as JSON as part of the build.
+        /// </summary>
+        public bool Json { get; init; }
```
vicancy, line 24: naming: ExportJsonModel?

question: is this for metadata or build step or both?

src/Microsoft.DocAsCode.App/DocfxProject.cs
```diff
@@ -0,0 +1,100 @@
...
+using System.Threading.Tasks;
+using Microsoft.DocAsCode.Common;
+using Newtonsoft.Json.Linq;
+
+namespace Microsoft.DocAsCode
+{
+    /// <summary>
+    /// Provides options to be used with <see cref="DocfxProject.Build(BuildOptions)"/>.
+    /// </summary>
+    public class BuildOptions
```
vicancy, line 14: Do we need a more specific name, e.g. `DocfxBuildOptions`?

vicancy (commented):
Looks awesome! Some little comments on the API surface

src/Microsoft.DocAsCode.App/DocfxProject.cs
yufeih: It is for build step, do we export JSON model for metadata step?

I went back and forward between `Json` and `ExportJson`, and tentatively named it `json` which looks similar to `typedoc`'s [`json`](https://typedoc.org/guides/options/#json) option:

I do want the name to be the same as command line args similar to `typedoc`. `--exportRawModel` and `--exportViewModel` are the args we have today, I believe `--exportViewModel` is something we want to get rid of with a newer template system because it is tied to the intermediate javascript execution step.

Even as I'm typing this I realize there is a semantic difference in "how `--exportRawModel` behaves today" vs "how I want it to behave". Today it export JSON in addition to HTML, but I would like to only export JSON files under this mode. So I might remove this option for now until we've settled down the semantics. It is used for testing purposes anyway.

src/Microsoft.DocAsCode.App/DocfxProject.cs
yufeih: I would say no, but that is a good question on whether we should rename `DocfxProject` -> `Project`. The term project might be too general so another candidate name is `Docset`, which is widely used in mslearn.

PR #8318, opened 2022-12-31
remove old package references

tibel:
Remove old package references that are not needed by .NET 6.0 anymore.


yufeih (approved):
Thank you!

PR #8361, opened 2023-01-22
Upgrade to Jint 3.0.0-beta-2046

lahma:
Now targeting main for the Jint upgrade.


yufeih:
Test passed, but building the docs folder failed with:

> Error running GetOptions function inside template preprocessor: Property 'forEach' of object is not a function

lahma:
This should now be resolved, sharing a static engine instance is not a good thing as engine is now thread-safe. Changed to pass engine in question to type conversion so now at least "works on my machine".

yufeih (approved):

yufeih:
Awesome thanks!

lahma:
You're welcome,  thanks for accepting the PR!

PR #8371, opened 2023-01-28
Add support for generating multi-level TOC files

### LINKED ISSUE #274: Support for nested namespaces
I believe a very useful feature for DocFX is to support nested namespaces, I realize that this in itself can be quite a challenge, but having the option to turn this feature on would help to generate less confusing and a more compact TOC, especially in much larger projects.

NanoBob:
This PR adds support for generating multi-level (or nested) TOC files for managed reference projects.
This PR Resolves #274 

This PR:
- Adds a `useMultiLevelToc` to the metadata options in `docfx.json 
- Conditionally changes the generation of the toc.yml file, to create a nested TOC when the above option is enabled.
- Moves ordering of TOC items to the `YamlMetadataResolver`
- Prioritises items with children first (so namespaces appear above classes) when using nested TOCs
- Adds a minor CSS change to the default theme, to 
- Adds appropriate tests
- Removes some checks from `LayoutCheckAndCleanup`


yufeih:
It seems the snapshot tests failed, try run [Samples Tests](https://github.com/dotnet/docfx/blob/main/test/docfx.Snapshot.Tests/SamplesTest.cs#L17) locally and commit the changes (if the changes are intentional) to fix the test cases.

NanoBob:
I've committed the files as changed by running the tests. 

Prior to running the tests I did however have to manually copy the build results of the templates to the `bin/debug/dotnet7.0` folders in two separate projects. I'm not sure if there is some automated step that is meant to do this, or some instructions that are missing for this, but after doing this I was able to run the tests successfully.

I am not really certain about these generated changes though, and it seems like committing them hasn't resolved the test failures either.

NanoBob:
Turns out after copying the built templates to one more directory, and re-running the tests resolved the issue. Tests should all be green now.

yufeih:
> Turns out after copying the built templates to one more directory, and re-running the tests resolved the issue. Tests should all be green now.

There should be a symbolic link [here](https://github.com/dotnet/docfx/blob/main/src/Microsoft.DocAsCode.App/templates) that points to template dist folder, something might went wrong here.

NanoBob:
> > Turns out after copying the built templates to one more directory, and re-running the tests resolved the issue. Tests should all be green now.
> 
> There should be a symbolic link [here](https://github.com/dotnet/docfx/blob/main/src/Microsoft.DocAsCode.App/templates) that points to template dist folder, something might went wrong here.

That symlink seems to just be cloned as a file, not as a symlink. That would explain my troubles.
but in current state the PR should be ready for review, as manually creating the template directories allowed me to run the tests and generate the files.

src/Microsoft.DocAsCode.App/Config/MetadataJsonItemConfig.cs
```diff
@@ -52,5 +52,8 @@ internal class MetadataJsonItemConfig
 
         [JsonProperty("disableDefaultFilter")]
         public bool DisableDefaultFilter { get; set; }
+
+        [JsonProperty("useMultiLevelToc")]
```
yufeih, line 56: Consider use an enum over a bool, in the original issue, there are other proposals to remove repeated prefix from nested namespaces, e.g.

```
Microsoft.DocAsCode
    |- Build
    |- Common
```

We don't need to implement it here, but an enum makes room for future enhancements like this.

test/docfx.Snapshot.Tests/SamplesTest.Seed/api/toc.raw.verified.json
```diff
@@ -6,20 +6,25 @@
       "topicHref": "CatLibrary.html",
       "topicUid": "CatLibrary",
       "items": [
+        {
```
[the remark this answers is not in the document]
NanoBob: This is now resolved, the default order is restored.

src/Microsoft.DocAsCode.App/Config/MetadataJsonItemConfig.cs
```diff
@@ -52,5 +52,8 @@ internal class MetadataJsonItemConfig
...
+        public bool UseMultiLevelToc { get; set; }
```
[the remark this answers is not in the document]
NanoBob: I've changed this to be an enum with the options:
`Flattened`, `Nested` and `CompactNested`

src/Microsoft.DocAsCode.Metadata.ManagedReference/Resolvers/YamlMetadataResolver.cs
```diff
@@ -53,6 +50,56 @@ public static class YamlMetadataResolver
...
+
+            MetadataItem root = new MetadataItem()
+            {
+                Type = MemberType.Toc,
+                Items = new()
+            };
+            Dictionary<string, MetadataItem> namespacedItems = new();
+
+            foreach (var member in namespaces
+                .OrderBy(x => x.Value.Name.Where(x => x == '.').Count())
```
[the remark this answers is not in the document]
NanoBob: I've moved the counting of the dots to be outside of the `OrderBy`

src/Microsoft.DocAsCode.Metadata.ManagedReference/Resolvers/YamlMetadataResolver.cs
```diff
@@ -53,6 +50,56 @@ public static class YamlMetadataResolver
...
+                return new MetadataItem
+                {
+                    Type = MemberType.Toc,
+                    Items = namespaces
+                        .Select(x => x.Value)
+                        .OrderBy(x => x.Name)
+                        .ToList(),
+                };
+
+
```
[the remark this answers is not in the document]
NanoBob: This is how it works now, example from the tests:

```
- uid: Samples.Foo
  name: Foo
  items:
  - uid: Samples.Foo.Sub
    name: Sub
    items:
    - uid: Samples.Foo.Sub.SubBar
      name: SubBar
    name.csharp: Samples.Foo.Sub
    name.vb: Samples.Foo.Sub
  - uid: Samples.Foo.Bar
    name: Bar
  name.csharp: Samples.Foo
  name.vb: Samples.Foo
```

src/Microsoft.DocAsCode.App/Config/MetadataJsonItemConfig.cs
NanoBob: I've added a third option to the enum, `CompactNested`, which does exactly this.

```yml
- uid: Samples.Foo
  name: Samples.Foo
  items:
  - uid: Samples.Foo.Sub
    name: Sub
    items:
    - uid: Samples.Foo.Sub.SubBar
      name: SubBar
    name.csharp: Samples.Foo.Sub
    name.vb: Samples.Foo.Sub
  - uid: Samples.Foo.Bar
    name: Bar

```

src/docfx/Models/MetadataCommandOptions.cs
```diff
@@ -43,5 +44,8 @@ internal class MetadataCommandOptions : LogOptions, ICanPrintHelpMessage
 
         [Option("disableDefaultFilter", HelpText = "Disable the default API filter (default filter only generate public or protected APIs).")]
         public bool DisableDefaultFilter { get; set; }
+
+        [Option("tocNamespaceStyle", HelpText = "Determines the style to be used for generating the table-of-contents.")]
```
yufeih, line 48: nit: remove `-`

```suggestion
        [Option("tocNamespaceStyle", HelpText = "Determines the style to be used for generating the table of contents.")]
```

src/docfx/SubCommands/MetadataCommand.cs
```diff
@@ -72,6 +72,7 @@ private MetadataJsonConfig ParseOptions(MetadataCommandOptions options, out stri
                 item.ShouldSkipMarkup |= options.ShouldSkipMarkup;
                 item.DisableGitFeatures |= options.DisableGitFeatures;
                 item.DisableDefaultFilter |= options.DisableDefaultFilter;
+                item.TocNamespaceStyle |= options.TocNamespaceStyle;
```
yufeih, line 75: It appears `|=` doesn't quite make sense here because `TocNamespaceStyle` is not a flags enum. The logic here is to override JSON config with command line options, so consider making `MetadataCommandOptions.tocNamespaceStyle` nullable and override when specified:

```suggestion
                item.TocNamespaceStyle = options.TocNamespaceStyle ?? item.TocNamespaceStyle;
```

src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/ExtractMetadata/TocNamespaceStyle.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        /// <summary>
+        /// Renders the namespaces in a nested tree form
+        /// </summary>
+        Nested,
+
+        /// <summary>
+        /// Renders the namespaces in a nested tree form, compacting namespaces to only use the lowest level
+        /// </summary>
+        CompactNested,
```
yufeih, line 21: I feel a bit hesitated to expose both `nested` and `compactNested` in the first release of the feature. We probably want to pick one recommended nesting style. My intuition is to show only the lowest level as it is more condensed, it matches nested namespace declaration syntax in C# and it fits naturally with the treeview. So consider remove the `CompactNested` option and adjust the behavior of `Nested` to match what `CompactNested` is describing here.

yufeih (approved):
Thanks for your contribution @NanoBob !

gumbarros:
Any estimated date for a release with this feature 👀?

yufeih:
It'll be included in 2.62.0 release which may come in about less than a month or so.

Issue #5599, opened 2020-03-05
How to Generate Docs with primitive type name


wistnki:
**Operation System**: Windows

**DocFX Version Used**: 2.50.0.0

**Template used**: default and little customized

**Steps to Reproduce**:

just follow [Quickstart](https://dotnet.github.io/docfx/tutorial/docfx_getting_started.html) to generate API doc

**Expected Behavior**:

In generated HTML, members whose type is primitive are written with primitive type names such as float.

**Actual Behavior**:

In generated HTML, they are shown with BCL names such as System.Single.

---

How can I get a document with primitive type names?

wistnki:
It looks written names of types are provided by xrefService. But it says the name of System.Single is 'Single'. Therefore the way to overwrite its displayed name with 'float' is needed.

superyyrrzz:
You can provide an xref map to docfx to overwrite this default behavior. But you are right, there is no available config to switch it right now.

wistnki:
yeah, I can provide another xref map. But it looks the xref from microsoft's xrefService is used.

QFSW:
How would you overwrite the full name with the primitive name? (even if this needs to be done manually now)

Or is it not yet possible?

superyyrrzz:
When resolving an xref, docfx should [read from xrefmap](https://github.com/dotnet/docfx/blob/dev/src/Microsoft.DocAsCode.Build.Engine/DocumentBuildContext.cs#L544) first, then [query xref](https://github.com/dotnet/docfx/blob/3c529e22a5d4dc77d2fda2972e40d14a6c377555/src/Microsoft.DocAsCode.Build.Engine/DocumentBuildContext.cs#L563) service.

wistnki:
Thanks for your support.

I wrote xmap yaml file such as:
```
references:
- uid: System.Single
  name: float
```
and added  `"xref": ["xrefmap.yml"]` in build section of docfx.json.
But It would not work.

superyyrrzz:
Thank you for the information. I think this name should be already there with information extracted from Roslyn. I cannot think of an easy workaround for now.

wistnki:
Information extracted from Roslyn? Isn't the infomation used provided from xref map service? such as:
https://xref.docs.microsoft.com/query?uid=System.Single

rhys-vdw:
Is there now a way to use the more familiar names for these types?

PR #8372, opened 2023-01-28
feat: display C# special types using familiar format

yufeih:
Addresses #5599


test/docfx.Snapshot.Tests/SamplesTest.Seed/api/CatLibrary.Cat-2.raw.verified.json
```diff
@@ -840,7 +840,7 @@
         },
```
yufeih, line 840: Why is the `Override` keyword in C#?

yufeih: Turns out there is an UID overwrite
```
uid: CatLibrary.Cat`2.CalculateFood(System.DateTime)
name: "Override CalculateFood Name"
```

PR #8373, opened 2023-01-28
Upgrade to YamlDotNet version 12.3.1

lahma:
This PR contains code changes required to get latest YamlDotNet version to compile and run.

* some `string` usage has changed to more typed, like `TagName` and `AnchorName`
* exception's `Message` does no longer contain the location, need to use `ToString()` to get same format as expected before
* `SerializationOptions` was removed, but as it's configuration data in docfx, easiest to copy old version to the project
* Custom `ExclusiveObjectGraphVisitor` which behaves like the one in 6.x library
* I also added extra caching layer to `EmitTypeInspector` when there are no property extension and can use statically calculated data, this almost halved the memory allocations for `PropertyDescriptor` in my testing project, from 516MB to 260MB

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ docs/api
 *.cache
 .settings
 .DS_Store
+.idea
 launchSettings.json
 package-lock.json
 *.sublime-workspace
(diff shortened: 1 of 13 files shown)


lahma:
Whoa.. the last problem took some hours to debug. There's been a subtle change to `DefaultExclusiveObjectGraphVisitor` in YamlDotNet, basically a bug fix that actually started to check whether to emit default values. Now that it kicked in, some functionality got broken when value was missing.

The change was hidden very well into this large commit: https://github.com/aaubry/YamlDotNet/commit/b7c128c47af392a6bd4947b0cf8befc69ee5cf40#diff-8de8e2e9141d0ba4d8b3027931e674fc5f3952a66104497c82ba570cfb643d57L41-L47

yufeih (approved):

yufeih:
Thanks for your contribution! @lahma

PR #8382, opened 2023-01-30
Fix exception thrown when generating metadata for cast operators

### LINKED ISSUE #8381: [Bug] Exception thrown when trying to generate metadata for codebase that has cast operators
**Describe the bug**
Generating metadata throws an exception when you have multiple cast operators in a single class.

**To Reproduce**
- Add cast operators to a class as follows:
    ```cs
    public static explicit operator uint(Bar value) => 0;
    public static explicit operator string(Bar value) => "";
    ```
- Run docfx to generate metadata
- See exception

**Expected behavior**
I expect it to work properly without execptions.

- OS: Windows
- Docfx version: Most recent main branch

**Additional context**
I suspect this PR caused the issue:
https://github.com/dotnet/docfx/pull/8374

NanoBob:
Resolves #8381 

Removing this debug assertion fixes the above issue for implicit operators.
I have added tests for this case, I do however not know for certain whether removing this debug assertion is enough, or if there is desire for more structural changes.

### DIFF
--- a/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Models/ReferenceItem.cs
+++ b/src/Microsoft.DocAsCode.Metadata.ManagedReference.Common/Models/ReferenceItem.cs
@@ -140,7 +140,6 @@ namespace Microsoft.DocAsCode.Metadata.ManagedReference
                             {
                                 for (int i = 0; i < sourceParts.Count; i++)
                                 {
-                                    Debug.Assert(sourceParts[i].Name == targetParts[i].Name);
                                     targetParts[i].IsExternalPath &= sourceParts[i].IsExternalPath;
                                     targetParts[i].Href = targetParts[i].Href ?? sourceParts[i].Href;
                                 }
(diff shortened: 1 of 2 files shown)


yufeih:
> Resolves #8381
> 
> Removing this debug assertion fixes the above issue for implicit operators. I have added tests for this case, I do however not know for certain whether removing this debug assertion is enough, or if there is desire for more structural changes.

It is better not to remove that assert, `Name` here is an id not a display name, the assert ensures method overloads with the same ID has appears the same in the overloads view. The fix is probably in https://github.com/dotnet/docfx/blob/main/src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/SymbolFormatter.cs#L120 or https://github.com/dotnet/docfx/blob/main/src/Microsoft.DocAsCode.Metadata.ManagedReference.Roslyn/SymbolFormatter.cs#L149

yufeih:
See https://github.com/dotnet/docfx/pull/8384

PR #8416, opened 2023-02-15
fix: cannot search due to lunr not defined

yufeih:


yufeih:
> Which versions of DocFX have this bug?

v2.62.0 published yesterday

PR #8419, opened 2023-02-18
fix: implicitly run dotnet restore

yufeih:
Building a project without any NuGet package reference requires `dotnet restore` in certain cases for `MSBuildWorkspace` to generate a compilation: https://github.com/dotnet/roslyn/issues/36072, otherwise predefined types are not imported:

```
error CS0518: Predefined type 'System.Object' is not defined or imported
```

Run an implicit dotnet restore for `csproj` and `sln`, which can be turned off using the `noRestore` config.


yufeih:
Fixes #8465

PR #8424, opened 2023-02-18
Fix/empty namespace in nested style

NanoBob:
- When an empty namespace is included in a nesting of namespaces it now is present in the toc structure properly
- These namespaces are "compacted" when they have a **single child**, to prevent needless levels of nesting in the toc

Resolves #8417


yufeih (approved):
Thank you @NanoBob

PR #8445, opened 2023-02-24
Group explicit interface implementations into their own section

mikernet:
Puts explicit interface implementations into their own group when using `memberpage`:

image


yufeih (approved):
Thank you @mikernet !

PR #8446, opened 2023-02-24
Add option for including explicit interface implementations

mikernet:
Adds `includeExplicitInterfaceImplementations` metadata option which allows including explicit interface implementations without exposing all private members.

### DIFF
--- a/src/Microsoft.DocAsCode.Dotnet/SymbolFilter.cs
+++ b/src/Microsoft.DocAsCode.Dotnet/SymbolFilter.cs
@@ -94,29 +94,19 @@ namespace Microsoft.DocAsCode.Dotnet
             if (symbol.IsImplicitlyDeclared && symbol.Kind is not SymbolKind.Namespace)
                 return false;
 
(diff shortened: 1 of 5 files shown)


mikernet:
@yufeih Could you review this for suitability before I go too far into testing?

src/Microsoft.DocAsCode.Dotnet/SymbolFilter.cs
```diff
@@ -94,6 +94,20 @@ private bool IsSymbolAccessible(ISymbol symbol)
             if (symbol.IsImplicitlyDeclared && symbol.Kind is not SymbolKind.Namespace)
                 return false;
 
+            if (_config.IncludeExplicitInterfaceImplementations)
+            {
+                bool isEii = symbol.Kind switch
```
yufeih, line 99: consider extract a `bool SymbolHelper.TryGetExplicitInterfaceImplementations(ISymbol symbol, out IEnumerable<ISymbol> eiis)` method and reuse it in line 113

PR #8513, opened 2023-03-14
feat: update docfx to use modern template

yufeih:


sotanakamura:
Very nice theme, looks like microsoft docs. Thanks for your hard work @yufeih

PR #8518, opened 2023-03-15
feat: clickable image

### LINKED ISSUE #8489: [Feature Request] Allow Zoom in full screen when click to image

When I attach this image, I can't quick view detail full screen by double click to image.
image

Right click and Open image in new tab
image

**Additional context**

yufeih:
Fixes #8489

### DIFF
--- a/templates/modern/src/markdown.ts
+++ b/templates/modern/src/markdown.ts
@@ -19,16 +19,30 @@ export function renderMarkdown() {
 }
 
 /**
- * Make images in articles clickable
+ * Make images in articles clickable by wrapping the image in an anchor tag.
+ * The image is clickable only if its size is larger than 200x200 and it is not already been wrapped in an anchor tag.
  */
 function renderClickableImage() {
+  const MIN_CLICKABLE_IMAGE_SIZE = 200
+  const imageLinks = Array.from(document.querySelectorAll<HTMLImageElement>('article a img[src]'))
+
   document.querySelectorAll<HTMLImageElement>('article img[src]').forEach(img => {
-    const a = document.createElement('a')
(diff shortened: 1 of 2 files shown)


yufeih:
Yes, a min image size criteria should help. This is an option per image so a global config may not be ideal.

yufeih:
> This seems to be called from the DOMContentLoaded event. [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/DOMContentLoaded_event):
> 
> > It doesn't wait for other things like images, subframes, and async scripts to finish loading.
> 
> Are `naturalWidth` and `naturalHeight` available if the image hasn't finished loading?

I did test it in chrome devtools with emulated slow internet connection and it seems to have retrieved the value. In theory we need to hook up an event, but there is no mention which is the one to reliably yields the value.

glopesdev:
@yufeih @KalleOlaviNiemitalo I just hit upon this behavior. It cost me hours to understand what exactly was happening and why some images were being wrapped and some weren't. This seems like a terrible decision that assumes something very weird as images wanting to point to themselves?! What use case is this exactly trying to solve?

https://github.com/dotnet/docfx/blob/16b4bc31369bd8483653c7a8db6d7435e1ed07ae/templates/modern/src/markdown.ts#L95C36-L95C36

> It might be best to limit this to images that are the only content of a p element, and even then allow opting out via docfx.json.

@KalleOlaviNiemitalo at the very least this should have been implemented. A global config is absolutely necessary, otherwise I'm now left with these horrible anchors that I can't seem to get rid of, they are breaking my custom DOM handlers and in general create website behaviors that seem completely unwarranted.

PR #8536, opened 2023-03-20
Clean up

RussKie:
* Consolidate common props in root Directory.Build.props
* Use CPM
See https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management
* Resolve warnings
	* Suppress obsolete warnings CS0612 and CS0618
	* Suppress NU1701, see https://learn.microsoft.com/nuget/reference/errors-and-warnings/nu1701
	* Remove unused variables
	* Resolves RS1024: Use 'SymbolEqualityComparer' when comparing symbols

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,0 +1,44 @@
+<Project>
+  <!-- See https://learn.microsoft.com/nuget/consume-packages/Central-Package-Management -->
+  <PropertyGroup>
+    <ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageVersion Include="CommandLineParser" Version="1.9.71" />
+    <PackageVersion Include="HtmlAgilityPack" Version="1.11.46" />
+    <PackageVersion Include="ICSharpCode.Decompiler" Version="8.0.0.7246-preview3" />
+    <PackageVersion Include="iTextSharp" Version="5.5.13.3" />
+    <PackageVersion Include="Jint" Version="3.0.0-beta-2048" />
+    <PackageVersion Include="JsonSchema.Net" Version="4.0.0" />
+    <PackageVersion Include="Markdig" Version="0.31.0" />
+    <PackageVersion Include="Microsoft.Build" Version="17.5.0" />
+    <PackageVersion Include="Microsoft.Build.Locator" Version="1.5.5" />
+    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.5.0" />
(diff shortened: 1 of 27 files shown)


samples/Directory.Build.props
```diff
@@ -2,4 +2,6 @@
   <PropertyGroup>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
+
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```
yufeih, line 6: Do we want samples to inherit root? They are standalone examples and we probably don't want root settings such as target frameworks impacting them.

yufeih (approved):
Thank you @RussKie !

samples/Directory.Build.props
RussKie: I've done it for consistency. Happy to pull it out, if you like.

yufeih: Yes please pull it out.

Directory.Packages.props
```diff
@@ -0,0 +1,44 @@
...
+    <PackageVersion Include="Microsoft.Build.Locator" Version="1.5.5" />
+    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Common" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.VisualBasic.Workspaces" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.VisualBasic" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.MSBuild" Version="4.5.0" />
+    <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
```
WeihanLi, line 25: Maybe this can be specified as a `GlobalPackageReference`

yufeih: Aha, I don't have too much preference. Not using central package version management at all is also good to me because @dependabot takes care of all the jobs and I almost never update these packages myself 😁

RussKie: I actually never seen that property. Though, according to the docs it's appropriate for references that apply to all projects. Not sure, if any of these truly are.

WeihanLi: Looks good from my side, I'm using this in [my project](https://github.com/WeihanLi/dotnet-exec/blob/0.13.0/Directory.Packages.props#L26) and the NuGet package has source link info also https://www.nuget.info/packages/dotnet-execute/0.13.0.

image

Issue #8526, opened 2023-03-20
[Bug] The type initializer for 'Microsoft.DocAsCode.Dotnet.DotnetApiCatalog' threw an exception: No instances of MSBuild could be detected.


RussKie:
DocFx tool is failing with the following exception:

```
System.TypeInitializationException: The type initializer for 'Microsoft.DocAsCode.Dotnet.DotnetApiCatalog' threw an exception.
 ---> System.InvalidOperationException: No instances of MSBuild could be detected.
```

I'm trying to port an existing implementation from explicitly installing DocFx from a zip (source from the Release page) to a .NET tool-based implementation. 
The original code base is also targeting .NET 7, and the code base I'm working on is targeting .NET 8.

I can't see how any of these should be a problem, but at any rate the tool shouldn't fail as it does. If MSBuild is needed, then it can be located with [vswhere tool](https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-public/NuGet/vswhere/overview/3.1.1). If all else fails, a user should be given a meaningful and actionalbe error message - e.g., "Unable to detect MSBuild, please supply it via '--abc' argument".

**To Reproduce**

```xml
# NuGet.config

    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" />
    <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
    <add key="dotnet-tools" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-tools/nuget/v3/index.json" />
    <add key="dotnet8" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8/nuget/v3/index.json" />
    <add key="dotnet8-transport" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet8-transport/nuget/v3/index.json" />

```

```json
// global.json

{
  "sdk": {
    "version": "8.0.100-preview.2.23115.11"
  },
  "tools": {
    "dotnet": "8.0.100-preview.2.23115.11",
    "runtimes": {
      "dotnet/x64": [
        "3.1.32",
        "6.0.13",
        "7.0.2"
      ],
      "aspnetcore/x64": [
        "3.1.32",
        "6.0.13",
        "7.0.2"
      ]
    }
  },
  "msbuild-sdks": {
[...]

PR #8541, opened 2023-03-21
Target .NET 8.0

RussKie:
#8526

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -12,6 +12,7 @@ runs:
       dotnet-version: |
         6.x
         7.x
+        8.0.100-alpha.1.23061.8
 
   - run: npm install
     shell: bash
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,7 +18,6 @@ jobs:
         dotnet-version: |
           6.x
           7.x
-          8.0.100-alpha.1.23061.8
 
     - run: npm install
       working-directory: templates
@@ -39,9 +38,9 @@ jobs:
 
     - name: dotnet publish
       run: |
-        dotnet publish src/docfx -f net8.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r win-x64 -o drop/publish/win-x64
-        dotnet publish src/docfx -f net8.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r linux-x64 -o drop/publish/linux-x64
-        dotnet publish src/docfx -f net8.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r osx-x64 -o drop/publish/osx-x64
+        dotnet publish src/docfx -f net7.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r win-x64 -o drop/publish/win-x64
+        dotnet publish src/docfx -f net7.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r linux-x64 -o drop/publish/linux-x64
+        dotnet publish src/docfx -f net7.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r osx-x64 -o drop/publish/osx-x64
         mkdir -p drop/bin
 
     - run: zip -r ../../bin/docfx-win-x64-${GITHUB_REF_NAME}.zip .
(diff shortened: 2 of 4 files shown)


.github/actions/build/action.yml
```diff
@@ -12,6 +12,8 @@ runs:
       dotnet-version: |
         6.x
         7.x
+        8.0.100-alpha.1.23061.8
```
yufeih, line 15: How often do we expect to keep this version number updated? Or is it appropriate to use latest pre-release?

RussKie:
This seems to have worked. 

Though, I see failures in the log, may I get a guidance on those? https://github.com/dotnet/docfx/actions/runs/4474452471/jobs/7862932362#step:7:107

Directory.Build.props
```diff
@@ -1,7 +1,8 @@
 <Project>
   <PropertyGroup>
-    <TargetFrameworks>net6.0;net7.0</TargetFrameworks>
+    <TargetFrameworks>net6.0;net7.0;net8.0</TargetFrameworks>
```
yufeih, line 3: Consider only build for `net8.0` at this timeframe in CI by checking `ContinuousIntegrationBuild` since most contributors likely don't have .NET 8 SDK.

yufeih:
> This seems to have worked.
> 
> Though, I see failures in the log, may I get a guidance on those? https://github.com/dotnet/docfx/actions/runs/4474452471/jobs/7862932362#step:7:107

Those errors are side effects of running test cases and exist before this change, so we should be good.

Directory.Build.props
```diff
@@ -1,7 +1,8 @@
 <Project>
```
yufeih, line 1: Tests are executed framework by framework in [ci.yml](https://github.com/dotnet/docfx/blob/ee4ab806de415ec47e0ed382615246e49ab52c3f/.github/workflows/ci.yml#L35) to fail fast, consider run tests against net8.0 there. I make tests against non-mandonary TFMs ubuntu only to make CI fast.

.github/actions/build/action.yml
RussKie: I don't expect this number to change until .NET 8 goes GA. Taking the latest SDK is problematic as it burns the version into the artifacts, and that then requires consumers to have at least the same version of runtime installed locally. See https://github.com/microsoft/slngen/issues/456.

.github/workflows/release.yml
```diff
@@ -38,9 +39,9 @@ jobs:
 
     - name: dotnet publish
       run: |
-        dotnet publish src/docfx -f net7.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r win-x64 -o drop/publish/win-x64
-        dotnet publish src/docfx -f net7.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r linux-x64 -o drop/publish/linux-x64
-        dotnet publish src/docfx -f net7.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r osx-x64 -o drop/publish/osx-x64
+        dotnet publish src/docfx -f net8.0 -c Release /p:Version=${GITHUB_REF_NAME#v} --self-contained -r win-x64 -o drop/publish/win-x64
```
yufeih, line 42: This one is for the zip files in GitHub releases, we probably don't want to publish as 8.0 until it GA. The global tool published to NuGet.org in line 32 could include 8.0.

Directory.Build.props
RussKie: Done

RussKie:
Updated.
Is there anything else?

yufeih (approved):
Thank you! @RussKie

yufeih:
It appears .NET 8 tests are not running https://github.com/dotnet/docfx/actions/runs/4474860442/jobs/7863724329?pr=8541 Is it because [ContinuousIntegrationBuild](https://github.com/dotnet/docfx/pull/8541/files#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16eR26) was set after TargetFrameworks checks?

RussKie:
> It appears .NET 8 tests are not running [dotnet/docfx/actions/runs/4474860442/jobs/7863724329?pr=8541](https://github.com/dotnet/docfx/actions/runs/4474860442/jobs/7863724329?pr=8541) Is it because [ContinuousIntegrationBuild](https://github.com/dotnet/docfx/pull/8541/files#diff-9da24614831c308827a1ae533ffea392c97638c261dd42bd0f5226baa136d16eR26) was set after TargetFrameworks checks?

Doh! Fixed :)

RussKie:
Thank you!

Issue #6995, opened 2021-01-18
DocFX doesn't resolve `inheritdocs` for external interfaces


smaillet:
**Operating System**: Windows

**DocFX Version Used**: 
2.56.6

**Template used**: custom template

**Steps to Reproduce**:
1. Addd xrefservice to the docfx.json as described in [docfx doc site](https://dotnet.github.io/docfx/tutorial/links_and_cross_references.html#cross-reference-services)
2. on a type implementing IDisposable add the `/// <inheritdoc/>` comment to the Dispose() Method
2. Build docs

**Expected Behavior**:
Docs would resolve the inheritdocs comment to the official docs for IDispose.Dispose() without errors or warnings

**Actual Behavior**:
`[21-01-17 11:30:05.020]Warning:[MetadataCommand.ExtractMetadata](foo.cs#L50)Could not resolve base documentation for 'Foo.Bar.Dispose'`

**NOTE:** The xrefservice URL for [IDispose.Dispose](https://xref.docs.microsoft.com/query?uid=System.IDisposable.Dispose) does work and retrieves the information, so it's not like the data is missing from the service. It appears that docfx, just can't resolve `inheritdocs` when the docs are inherited from an external interface. (Which is quite a common thing)

peache85:
Any new about this issue ?

MrDave1999:
Any news on this issue? I have the same problem...

mikernet:
As well as external base classes.

arnonax-tr:
Any news or at least a workaround? (other than copy/paste the original docs...)

Crono1981:
I'm using 2.59.4 and this is still happening. :(

PR #8546, opened 2023-03-23
feat: compute mslearn symbol URL

yufeih:
The xref service was introduced to provide fast query by `uid` with a database backend. Microsoft Learn today still maintains and hosts an xref service for the .NET BCL, it is only used by the open-source community today and may be retired in the future.

There are also limitations of using xrefmap as intermediary in resolving .NET API cross references, e.g., missing info in xrefmaps causes [`<inheritdoc />` problems](https://github.com/dotnet/docfx/issues/6995). We’d like to move to a future where API references are resolved directly based on `Symbols`, both the content and URLs. 

URLs can be resolved based on conventions. This first step decouples docfx from mslearn xref service by matching assembly names against a list of assemblies published to mslearn and resolving them based on mslearn URL convention.


paulushub:
Is there a way or API to link to articles (conceptual topics) in the Microsoft Learn Docs?

yufeih:
> Is there a way or API to link to articles (conceptual topics) in the Microsoft Learn Docs?
No at the moment. URL is the way for external site to link to MSLearn docs, most high traffic pages would setup redirection on URL change.

PR #8553, opened 2023-03-25
Add git timeout environment config

### LINKED ISSUE #4094: GitException on GetFileDetail: Timeout (1000ms) exceeded
**DocFX Version Used**: 2.4.10

**Template used**: `default`

I am experiencing a strange issue which I have encountered when using **DocFX** with **Mono** inside of a **docker container**.  When building the HTML files in a local environment or a Linux VM, the "Improve this Doc" appears properly. However if I attempt to build from the same source repository using the same files, the link does not appear.

I receive the following warnings during the build process:
```sh
[19-02-05 09:39:45.787]Warning:[BuildCore.Build Document.Load.TocDocumentProcessor](articles/toc.md)Skipping GetFileDetail. Exception found: Microsoft.DocAsCode.Common.Git.GitException, Message: The result can't be null or empty string
[19-02-05 09:39:45.787]Warning:[BuildCore.Build Document.Load.TocDocumentProcessor](articles/dynamics/toc.md)Timeout (1000ms) exceeded. Killing process /usr/bin/mono-sgen.
```

For some reason the commands defined in `Microsoft.DocAsCode.Common/Git/GitUtility.cs` cannot run within the 1000 ms timeout limit. On the other hand if I run the build process again once or twice, the warnings do not appear and it can build properly, maybe because of some caching.

Would it be possible to somehow increase the timeout to a higher value using a command line parameter? Or did anyone encounter such issues when generating HTML files using DocFX from a Docker Container?

**Steps to Reproduce**:

1. Set up a Docker Container with Mono and download version 2.4.10 of DocFX
2. Clone GitLab repository containing Markdown files and docfx.json
3. Build html files with docfx.exe docfx.json

**Expected Behavior**:
The HTML files should be generated properly as on a local environment or virtual machine when configured correctly.

**Actual Behavior**:
The git information for each file cannot be retrieved for the first time because of the 1000 ms timeout set in `Microsoft.DocAsCode.Common/Git/GitUtility.cs` and the 'Improve this Doc' link does not appear, but it works properly in a local environment without any configuration changes.

Thank you very much in advance!

WeihanLi:
fixes #4094

### DIFF
--- a/src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
+++ b/src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
@@ -13,6 +13,7 @@ namespace Microsoft.DocAsCode.Common.Git;
 
 public static class GitUtility
 {
+    private const string GitTimeoutEnvVarName = "DOVFX_GIT_TIMEOUT";
     private static readonly string CommandName = "git";
     private static readonly int GitTimeOut = 1000;
 
@@ -48,6 +49,15 @@ public static class GitUtility
 
     private static bool? GitCommandExists = null;
     private static object SyncRoot = new();
+
+    static GitUtility()
+    {
+        var gitTimeoutEnvConfig = Environment.GetEnvironmentVariable(GitTimeoutEnvVarName);
+        if (int.TryParse(gitTimeoutEnvConfig, out var gitTimeout) && gitTimeout > 0)
+        {
+            GitTimeOut = gitTimeout;
+        }
+    }
 
     public static GitDetail TryGetFileDetail(string filePath)
     {
--- a/src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
+++ b/src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
@@ -13,9 +13,9 @@ namespace Microsoft.DocAsCode.Common.Git;
 
 public static class GitUtility
 {
-    private const string GitTimeoutEnvVarName = "DOVFX_GIT_TIMEOUT";
+    internal const string GitTimeoutEnvVarName = "DOCFX_GIT_TIMEOUT";
     private static readonly string CommandName = "git";
-    private static readonly int GitTimeOut = 1000;
+    private static readonly int GitTimeOut = GetGitTimeout();
 
     private static readonly string GetRepoRootCommand = "rev-parse --show-toplevel";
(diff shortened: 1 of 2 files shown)


src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.DocAsCode.Common.Git;
 
 public static class GitUtility
 {
+    private const string GitTimeoutEnvVarName = "DOVFX_GIT_TIMEOUT";
```
yufeih, line 16:
```suggestion
    private const string GitTimeoutEnvVarName = "DOCFX_GIT_TIMEOUT";
```

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
```diff
@@ -13,6 +13,7 @@ namespace Microsoft.DocAsCode.Common.Git;
...
     private static readonly string CommandName = "git";
     private static readonly int GitTimeOut = 1000;
```
yufeih, line 18: It is a bit unintuitive to know if the static constructor runs first or this static assignment runs first, consider use a static method.

```suggestion
    private static readonly int GitTimeout = GetGitTimeout();
```

test/Microsoft.DocAsCode.Common.Tests/GitUtilityTest.cs
```diff
@@ -30,6 +33,15 @@ public void Environment_ForBranchName()
         Assert.Equal("special-branch", info.RemoteBranch);
     }
 
+    [Fact]
+    public void Environment_ForGitTimeout()
```
yufeih, line 37: We probably don't need to test it as the logic pretty minor. If we do, make the property `internal` instead of using reflection.

src/Microsoft.DocAsCode.Common/Git/GitUtility.cs
yufeih, line 18: Maybe we can also relax the default 1s a little bit to 10 seconds.

WeihanLi: Updated

yufeih (approved):

yufeih:
Thank you @WeihanLi !

PR #8570, opened 2023-03-29
Update index.md workflow sample

mtirionMSFT:
Indentation of parameters for Deploy in the workflow was wrong, making it fail when copied.


yufeih (approved):
Thank you!

PR #8608, opened 2023-04-08
refactor: update xmlcomment

yufeih:


test/docfx.Snapshot.Tests/SamplesTest.Seed/api/CatLibrary.Cat-2.html.view.verified.json
```diff
@@ -1720,54 +1727,8 @@
             ],
             "id": "CatLibrary_Cat_2_CalculateFood_"
           },
-          "exceptions": [
```
yufeih, line 1723: `null` and `[]` causes different [overwrite merge behaviors](https://github.com/dotnet/docfx/blob/main/src/Microsoft.DocAsCode.Common/EntityMergers/MergeBehaviorOption.cs#L24), it seems `[]` follows list rules but `null` follows other rules.

PR #8609, opened 2023-04-09
Update markdown.md

VaclavElias:
docs: Example added for include

It was difficult to figure out how it works, so I added an example.


yufeih (approved):
Thank you!

PR #8634, opened 2023-04-12
fix: remove dependency on root repo path in unit test

wilkinson-steve:
Unit test SymbolUrlResolverUnitTest previously had "https://github.com/dotnet/docfx" hard-coded in as the repo path.  This change gets the correct (current) repo path from Git.

NB Only tested on Windows.

### DIFF
--- a/test/Microsoft.DocAsCode.Dotnet.Tests/SymbolUrlResolverUnitTest.cs
+++ b/test/Microsoft.DocAsCode.Dotnet.Tests/SymbolUrlResolverUnitTest.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
 using System.Text.RegularExpressions;
 using Microsoft.CodeAnalysis;
 using Xunit;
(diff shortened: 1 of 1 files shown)


wilkinson-steve:
@dotnet-policy-service agree

test/Microsoft.DocAsCode.Dotnet.Tests/SymbolUrlResolverUnitTest.cs
```diff
@@ -84,18 +85,45 @@ public static void GetPdbSourceLinkUrlTest()
...
         Assert.Equal(
-            "https://github.com/dotnet/docfx/blob/*/src/Microsoft.DocAsCode.Dotnet/DotnetApiCatalog.cs",
+            @$"{GetGitRepoOrigin()}/blob/*/src/Microsoft.DocAsCode.Dotnet/DotnetApiCatalog.cs",
             ReplaceSHA(SymbolUrlResolver.GetPdbSourceLinkUrl(compilation, type)));
 
         var method = type.GetMembers(nameof(DotnetApiCatalog.GenerateManagedReferenceYamlFiles)).FirstOrDefault();
         Assert.NotNull(method);
         Assert.Equal(
-            "https://github.com/dotnet/docfx/blob/*/src/Microsoft.DocAsCode.Dotnet/DotnetApiCatalog.cs",
+            @$"{GetGitRepoOrigin()}/blob/*/src/Microsoft.DocAsCode.Dotnet/DotnetApiCatalog.cs",
```
yufeih, line 94: May be we can simply change the assertion to say the expected url starts with `https://github.com/` and ends with `/blob/*/...`?

wilkinson-steve: Agreed - that change is now in the latest commit.

yufeih (approved):

PR #8641, opened 2023-04-14
refactor: CommandLineParser -> Spectre.Console.Cli

yufeih:
BREAKING CHANGES:
- Drop support of `docfx build` and `docfx metadata` commands without using `docfx.json` config. E.g., `docfx build --content ... --resource ...`
- Drop passing metadata through command line args due to usability problems with passing JSON to CLI
- The default `docfx` command executes `metadata`, `build`, and `pdf` in sequence, independent of order defined in `docfx.json`.


yufeih:
> > Drop support of docfx commands without docfx.json config.
> 
> Does that apply to `docfx serve`, too?

No, it is only for `build`, `metadata` commands

PR #8659, opened 2023-04-21
Fixing the location reference for metadata.src.src

StingyJack:
Now this is consistent with the examples and the actual working application


yufeih (approved):
Thank you!

PR #8689, opened 2023-04-26
fix: forward build command args to default command

### LINKED ISSUE #8688: [Bug] Error on unknown command line options (build and help)
**Describe the bug**
We can use `--build` and `-h` before v2.66.1. Now, we can't not use those parameters.
Now, we no need to add `--build` parameter if we want to build. But it'll lead to problem if someone already used this parameter in their pipeline.

**To Reproduce**
1.  docfx "x:\xxx\docfx.json" --build
2. docfx -h

**Expected behavior**

- OS: Windows 10
- Docfx version: [v2.66.1]

**Additional context**

E1

E2

yufeih:
fixes #8688

### DIFF
--- a/src/docfx/Models/BuildCommand.cs
+++ b/src/docfx/Models/BuildCommand.cs
@@ -13,20 +13,13 @@ internal class BuildCommand : Command<BuildCommandOptions>
     {
         return CommandHelper.Run(settings, () =>
         {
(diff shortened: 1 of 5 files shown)


yufeih:
> @yufeih: Thank you very much for the quick fix and the new release! I highly appreciate!

No Problem. Thank you for reporting the issue!

Issue #8742, opened 2023-05-08
[Feature Request] Dynamically load optional template dependencies


yufeih:
Today, the front-end script to process mermaid, MathJs, etc is bundled together in docfx.min.js, which is too big at 6MB. This issue is to track the splitting of the front-end script to only load what is needed for each page.

By splitting the script, we can reduce the size of the bundle that needs to be downloaded by the user, which will improve the overall performance of the site. This change will also allow users to only load the scripts that they need for each page, rather than loading the entire bundle for every page.

paulushub:
How about use of CDN for such third-party libraries?

yufeih:
I am looking at [esbuild code splitting](https://esbuild.github.io/api/#splitting) to automatically split the modules and download only the chunks needed for a given page. This makes it easier to manage, version and type check 3rd party libraries with npm and typescript, and ship minimal js to browser as needed.

PR #8751, opened 2023-05-08
feat: docfx.min.js code splitting

yufeih:
Reduces raw size (no gzip) of `docfx.min.js` from 6MB to 114KB:
- Turn `docfx.min.js` into an ESM module and uses `esbuild` to split it into chunks. 
- Lazy load `highlight.js`, `MathJax`, `mermaid` only when the feature is used on the page.

#8742


Lulalaby:
@yufeih did u forgot to check for the rel?
image

Lulalaby:
Never mind, sorry!

Seems like my folder had a docfx.min.js itself and it overwrote each other

yufeih:
Glad your issue is resolved @Lulalaby

PR #8752, opened 2023-05-08
docs: add info about uid in toc

Lulalaby:
We can use uid in the toc. 
I don't know why it isn't noted, or if it got removed.


Lulalaby:
Will check, test and add if it works!

Lulalaby:
Guessing from comments within the code base I think that this is under rework so I won't be adding more information at the moment.

yufeih (approved):
Thank you @Lulalaby ! Looks good to not include `tocHref` and `topicHref` at the moment.

PR #8761, opened 2023-05-10
fix: updated highlight.js default themes to vs and vs2015 (dark-mode)

wilkinson-steve:
As per discussion https://github.com/dotnet/docfx/discussions/8729, the highlight.js themes have been updated to `vs` and `vs2015` to improve presentation of C# code fragments.


yufeih:
It seems snapshot test cannot push to forked branches due to [cross-organization "Allow edits from maintainers"](https://github.com/orgs/community/discussions/5634).

PR #8780, opened 2023-05-16
fix(templates): modern template uses wrong rel

Lulalaby:
Fixes an issue introduced with #8751

Either I'm dumb rn or docfx, gotta check a few things. Getting weird cors errors


Lulalaby:
Need to check first if the rel is correct

PR #8799, opened 2023-05-28
feat: update default template to allow google analytics from metadata

SeanKilleen:
Bringing #8733 to the default template as well.

(let me know what additional docs etc. need to be updated and I'll be happy to do it.)


yufeih (approved):
Thank you @SeanKilleen ! I'll update the docs.

Issue #8354, opened 2023-01-19
[Bug] Nuget packages not working


zp-zpanda:
**Describe the bug**
The Microsoft.DocAsCode.* packages aren't working as expected.

**To Reproduce**
1. Get a docfx project using the new app model (e.g. the one in /samples/extensions)
2. Use Microsoft.DocAsCode.* packages and remove all `<ProjectReference>`s to `Microsoft.DocAsCode.*.csproj`s

**Expected behavior**
Everything builds cleanly.

**Actual behavior**
1. Only `manifest.json` and `xrefmap.yml` are created
2. These errors come up, sometimes as one, sometimes as two or more:
```
There is no template processing document type(s): Conceptual,ManagedReference,Toc
```

**Context (please complete the following information):**
- OS: Windows
- Docfx version: >2.60.0

**Additional context**
This bug appears with any project.

yufeih:
Fixed by https://github.com/dotnet/docfx/commit/255d4ed33c7be6a4d1f944c212a801cfd0210131

PR #8844, opened 2023-06-06
fix: add settings to include templates as linked content items

filzrev:
This PR is same as #8818. (It is closed before reviewed)

---

**This PR resolve  following problem.**

When using `Microsoft.DocAsCode.App` project with `ProjectReference`.
`templates` content files are not copied to output dir. (#8610)

Currently it require manual operations like belows.
 - Copy `templates/dist` contents to exe's `templates` directory.  
 - Create symbolic link or directory junction.

---
This PR resolve  this problem by importing `templates/dist/**` as `linked contents` .
https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#linkbase

image

**Note**
It seems `LinkBase` approach is used before by `Microsoft.DocAsCode.App` project before.

- https://github.com/dotnet/docfx/issues/8354
- https://github.com/dotnet/docfx/commit/3ed3b86f8ed5e1e0a32d46e820436dee1f8d7dd6#diff-3408df1cfc343f8e7b145f18759311d2c1972f05da8b0be78798f43fd4

This PR set `LinkBase` inside `docfx` project, and  it is not published as NuGet package. 
So similar problems are not expected to happens.


filzrev:
@yufeih 
Sorry to trouble you with my lack of confirmation.

It was my windows environment's problem.
If it is fixed. templates folder copy is executed as you described.

---
In my environment . `Microsoft.DocAsCode.App` project's "templates" symlink is not properly configured.
I fixed symlink with following commands.

1. Run `git config core.symlinks true`.
2. Run `git reset --hard` (with elevated privileges)

PR #8845, opened 2023-06-06
fix: Enable InvalidFileLink check when custom hrefGenerator is used 

filzrev:
**This PR resolve following problem.**
When custom `hrefGenerator` is used.
InvalidFileLink warning check is **always** skipped.

This PR add additional check that `hrefGenerator` returns `new href` or not.
If `href` is not modified. then execute normal InvalidFileLink check path.


yufeih (approved):
Looks great! Thank you @filzrev!

PR #8846, opened 2023-06-06
Add MSBuild target to check symlink existence for windows

filzrev:
Related: #8844

**summary**
Add simple MSBuild target to check `templates\default`  path exists or not at pre-build timing (On Windows only). 
If `templates/default` path is not found. Report as warning and build continues.


yufeih (approved):
Thank you @filzrev ! This is great addition to the contributor experience.

Issue #8847, opened 2023-06-06
[Bug] Unable to apply modern template on fresh DocFX project following quick start guide


jameswilky:
**Describe the bug**
I created a fresh docFX project using the steps provided in the quick start guide, and followed the instructions to supply the 'modern' template, but no template changes are being applied.

**To Reproduce**

1. `dotnet tool update -g docfx`
2. `docfx init --quiet`
3. `docfx docfx_project/docfx.json --serve`
4. update docfx.json with properties specified in [the docs](https://dotnet.github.io/docfx/docs/template.html?tabs=modern)

Noticed that the default docfx file ships with these settings in the docfx file:
```
    "template": [
      "default"
    ],
```
The docs say that it should be `"templates"`, so i tried the following combinations with no success

- `"template":["default","modern"]`
- `"templates":["default","modern"]`
- `"template":["modern"]`
- `"templates":["modern"]`

The site builds fine and runs but its still using the default template:
![image](https://github.com/dotnet/docfx/assets/39325484/72ff243c-6b53-46c6-829f-07797db9bc35)

This is my build output log:

```
PS C:\code\docfx_project> docfx docfx.json --serve
[23-06-06 11:19:00.795]Info:[MetadataCommand.ExtractMetadata]Using msbuild C:\Program Files\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin as inner compiler.
[23-06-06 11:19:00.824]Info:[MetadataCommand.ExtractMetadata]No files are found with glob pattern src/**.csproj, excluding <none>, under directory "C:\code\docfx_project"
[23-06-06 11:19:01.006]Warning:[MetadataCommand.ExtractMetadata]No project detected for extracting metadata.
[23-06-06 11:19:01.009]Info:[MetadataCommand]Completed Scope:MetadataCommand in 260.6692 milliseconds.
[23-06-06 11:19:01.118]Info:[BuildCommand]6 plug-in(s) loaded.
[23-06-06 11:19:01.142]Info:[BuildCommand]No files are found with glob pattern apidoc/**.md, excluding obj/**,_site/**, under directory "C:\code\docfx_project"
[23-06-06 11:19:01.144]Info:[BuildCommand]No files are found with glob pattern images/**, excluding <none>, under directory "C:\code\docfx_project"
[23-06-06 11:19:01.159]Info:[BuildCommand]Markdown engine is markdig
[...]

PR #8849, opened 2023-06-07
fix: documentation of template.md

filzrev:
Documentation fix related to #8847

If `templates`  is specified instead of `template`  in `docfx.json` build config, 
It is ignored by deserializer and fallback to `default` template.

https://github.com/dotnet/docfx/blob/ffabd3dd419bf8c6819dbd9b5c2e13d1f8ba93de/src/Microsoft.DocAsCode.App/Config/BuildJsonConfig.cs#L68-L69

https://github.com/dotnet/docfx/blob/ffabd3dd419bf8c6819dbd9b5c2e13d1f8ba93de/src/Microsoft.DocAsCode.App/RunBuild.cs#L14-L17

### DIFF
--- a/docs/docs/template.md
+++ b/docs/docs/template.md
@@ -4,12 +4,12 @@ Template defines the appearance of the website.
 
 Docfx ships several built-in templates. We recommend using the modern template that matches the look and feel of this site. It supports dark mode, more features, rich customization options and.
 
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Thank you @filzrev ! Nice catch.

Issue #8821, opened 2023-06-02
[Bug] sitemap contains redirect page entries


filzrev:
**Describe the bug**

Sitemap.xml contains [redirect page](https://dotnet.github.io/docfx/docs/config.html#url-redirects)  entries.
Redirect page should not be included in sitemap's entry.

**To Reproduce**

1. Gets sitemap.xml content(`https://dotnet.github.io/docfx/sitemap.xml`)
2. Check redirected page(`https://dotnet.github.io/docfx/tutorial/walkthrough/walkthrough_create_a_docfx_project_2.html`) exists.

**Expected behavior**

redirect page is not appear in sitemap entries.

**How to fix this problem**

It seems `SitemapGenerator` PostProcessor can't access FileModel metadata directly as described below.

> https://dotnet.github.io/docfx/tutorial/howto_add_a_customized_post_processor.html#step3-process-all-the-files-generated-by-docfx
>> Post-processor aims to process the output files, so the FileModel can't be accessed in this phase. If some metadata is needed here,
>> an option is to save it in FileModel.ManifestProperties in build phase, then access it through ManifestItem.Metadata.
>> Another option is to save it somewhere in output files, like HTML's <meta> Tag.

---

I want to making modifications in the following directions, but if you have a better idea, please let us know.

1. Add new HtmlPostProcessor handler  to check `<meta http-equiv="refresh"` tag existence. and save metadata to `OutputFileInfo`.
2. Then SitemapGenerator use this metadata to filter redirect page.

**Pros**
- No need to modify DocumentProcessors.
- No need to open HTML file  at SitemapGenerator.

**Cons**
- It dependent on a specification that `HtmlPostProcessor` is executed in first priority
- It dependent on `<meta http-equiv="refresh"`. 
  For SEO optimization, redirect may be implemented at server side feature.  
  (Example: https://www.statiq.dev/guide/web/content-files/redirects)

yufeih:
I'd prefer we check `redirect_url` metadata over `<meta http-equiv`, as many redirections are implemented server side. 

The `SitemapGenerator` does not have access to metadata, may be we can change the `DocumentType` of redirections to `Redirection` instead of `Conceptual` [here](https://github.com/dotnet/docfx/blob/main/src/Microsoft.DocAsCode.Build.ConceptualDocuments/ConceptualDocumentProcessor.cs#L106)?

filzrev:
Thanks for the reply.

[...]

PR #8892, opened 2023-06-20
Add redirection document type

filzrev:
Add new DocumentType `Redirection` according to comments on #8821.

1. Add `Redirection` document type to `DocumentType`.
2. Add logics to change document type from "Conceptual" to `Redirection`.  If `redirect_url` metadata is specified.
3. Add `Redirection.html.primary.tmpl` to `default` build-in template.  
    -> `Redirection` documents are processed this Mustache template only. (Don't use `_master.tmpl`)
4. Modify `SitemapGenerator.cs` to skip `Redirection` type documents. and some code refactoring.
5. Test code for `redirect_url` .

**Note**
`_master.tmpl`'s conditional statement (`{{#redirect_url}}`...`{{/redirect_url}}`) are left as is.  
Because `redirect_url` can be set via `global.json` and may effects other DocumentTypes.


test/docfx.Tests/DocsetTest.cs
```diff
@@ -166,4 +166,45 @@ public static async Task Build_With_File_Metadata_Files()
...
+            """
+            <!DOCTYPE html>
+            <html>
+              <head>
+                <meta charset="utf-8">
+                <meta http-equiv="refresh" content="0;URL='redirected.html'">
+              </head>
+            </html>
+            """, result);
+
```
yufeih, line 208: Verify that:
- the generated sitemap does not contain redirections
- the content type in the generated manifest file is redirection

filzrev: Fix DocumentType constant definitions in this commit (https://github.com/dotnet/docfx/pull/8892/commits/17fa984d0c1c6c4731e36f0b5f2046469597c233)
And unit tests are added related to `redirect_url` and sitemap generation. (https://github.com/dotnet/docfx/pull/8892/commits/a09b8f83d7740f2e28768953ceb6f0a060750f04)

yufeih (approved):
This is awesome, really appreciate all the hard work you put into this! ❤️❤️❤️

PR #8895, opened 2023-06-20
Set initial startup project to docfx.csproj

filzrev:
Initial VS startup project is determined by defined project order of `.sln`.
This PR set default startup project to `docfx.csproj`.


yufeih (approved):
I didn't realize how this impacts the default vs project selection. Thank you @filzrev !

PR #8896, opened 2023-06-21
Expand active page when it has subpages on modern template

MihkelAuv:
The default template expands its contents in the table of contents (TOC) when the current page has subpages, but the modern template currently does not.

This PR adds an additional check to the modern template to verify whether the currently active page has subpages and expands them.


yufeih (approved):
Thank you @MihkelAuv ! The change looks good.

PR #8898, opened 2023-06-21
Fix scroll into view in TOC

MihkelAuv:
Modern template should scroll last active list item into view: https://github.com/dotnet/docfx/blob/e0f8953c587966476b46a365b3b964725103e69a/templates/modern/src/toc.ts#L39-L43

But currently, list items do not get the class "active" assigned. This PR fixes this.


yufeih (approved):
Aha, nice catch!

Issue #8927, opened 2023-07-03
[Feature Request] Add launch browser feature when `docfx serve` command is executed


filzrev:
**Is your feature request related to a problem? Please describe.**

There is [dotnet-serve](https://github.com/natemcmaster/dotnet-serve) tools that support this feature.
It is useful when using docfx with custom MSBuild project and IDE integration.

**Describe the solution you'd like**
Add `--open-browser` parameter to `docfx serve` and `docfx build --serve` commands.
When this parameter specified. docfx launch new browser instance with default URL (http://localhost:{port}).

Additionally.  when relative path of markdown file is passed as argument.
It'll be resolved to corresponding URL by using `manifest.json`.  
This feature is useful when `docfx serve` command is execute via VSCode tasks.  
(VSCode support passing current file as `${relativeFile}` parameter)

**Additional context**
If above feature request is acceptable. I'll create PR.

yufeih:
Yes! Sounds like a great addition!

filzrev:
Features are Implemented by #8934
Following options are added for `docfx build --serve` and `docfx serve` commands.

```
--open-browser  Open a web browser when the hosted website starts
--openfile      Open a file in a web browser when the hosted website starts.
```

PR #8934, opened 2023-07-04
Add open browser options to docfx-serve command

filzrev:
This PR implement feature described #8927.

**What's is added to docfx CLI command**

Following options are added for  `docfx build` and `docfx serve` commands.

```
--open-browser                Open a web browser when the hosted website starts
--open-browser-relative-path  Specify the relative path to open
```

#### Usage
To launch browser.  pass `--open-browser` switch option to command.
It launch browser with root site URL (e.g. http:localhost:8080).

When `--open-browser-relative-path` optional parameter is passed.  
It try to resolve **output HTML file** by using `manifest.json` file content.
If failed to resolve HTML path. then error log is recorded. and fallback to site root URL.

##### `docfx build --serve`

```pwsh
docfx build --serve --open-browser
docfx build --serve --open-browser --open-browser-relative-path docs/config.html
docfx build --serve --open-browser --open-browser-relative-path docs/config.md 
```

##### Launch with specific page.

```pwsh
docfx serve _site --open-browser
docfx serve _site --open-browser --open-browser-relative-path docs/config .html
docfx serve _site --open-browser --open-browser-relative-path docs/config .md
```

#### Unit Tests
This PR don't contains unit tests.
Because it's difficult to write unit tests that relating OS commands and launch browser.

Tests are executed manually that cover all execution paths.
Excepting `macOS` environment browser launch test.

This PR use same OS command(`open`) that is used by [`dotnet-serve` to open browser](https://github.com/natemcmaster/dotnet-serve/blob/main/src/dotnet-serve/SimpleServer.cs#L175-L203).
I thought, there is no problems on `macOS` environment.

#### Documentation

This PR don't contains documentation updates.
Currently there are few references about `docfx serve` and `docfx build --serve` command in the documentation.

It might be better to update following related docs later.
- https://dotnet.github.io/docfx/#create-a-new-website
- https://dotnet.github.io/docfx/reference/docfx-cli-reference.html

#### Additional Design Notes

**Composing Commands**
This PR add options  as simple options.
`Spectre.Console.Cli` supports [Composing Commands](https://spectreconsole.net/cli/composing).
It might be better to change `--serve` related options as `Composing Commands`.

**Redundant argument names**
`--open-browser-relative-path` is a bit redundant option name. 
`--open-browser:[url]` style option is preferred that is implemented by the `dotnet serve` command.

But It seems`Spectre.Console.Cli` can't handle this option styles.

### DIFF
--- a/src/Microsoft.DocAsCode.App/RunServe.cs
+++ b/src/Microsoft.DocAsCode.App/RunServe.cs
@@ -15,7 +15,7 @@ namespace Microsoft.DocAsCode;
 
 internal static class RunServe
 {
-    public static void Exec(string folder, string host, int? port, bool openBrowser, string openBrowserRelativePath)
+    public static void Exec(string folder, string host, int? port, bool openBrowser, string openFile)
     {
         if (string.IsNullOrEmpty(folder))
             folder = Directory.GetCurrentDirectory();
@@ -48,25 +48,22 @@ internal static class RunServe
             using var app = builder.Build();
             app.UseFileServer(fileServerOptions);
 
-            if (openBrowser)
+            if (openBrowser || !string.IsNullOrEmpty(openFile))
             {
-                string relativePath = openBrowserRelativePath;
+                string relativePath = openFile;
                 var launchUrl = string.IsNullOrEmpty(relativePath)
                     ? url
-                    : Path.Combine(url, ResolveOutputHtmlRelativePath(folder, relativePath));
+                    : ResolveOutputHtmlRelativePath(baseUrl: url, folder, relativePath);
 
                 // Start web server.
                 app.Start();
 
                 // Launch browser process.
-
-                Console.WriteLine($"Launching browser with url: {url}.");
-                using var process = LaunchBrowser(launchUrl);
+                Console.WriteLine($"Launching browser with url: {launchUrl}.");
+                LaunchBrowser(launchUrl);
 
                 // Wait until server exited.
                 app.WaitForShutdown();
-
-                // process object is disposed. (Note:Launched process remain running.after dispose)
             }
             else
             {
@@ -81,12 +78,13 @@ internal static class RunServe
 
     /// <summary>
     /// Resolve output HTML file path by `manifest.json` file.
+    /// If failed to resolve path. return baseUrl.
     /// </summary>
-    private static string ResolveOutputHtmlRelativePath(string folder, string relativePath)
+    private static string ResolveOutputHtmlRelativePath(string baseUrl, string folder, string relativePath)
     {
         var manifestPath = Path.GetFullPath(Path.Combine(folder, "manifest.json"));
         if (!File.Exists(manifestPath))
-            return string.Empty;
+            return baseUrl;
 
         try
         {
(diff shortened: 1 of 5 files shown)


yufeih:
Is it simpler to have an `--open-file` option for opening a specific URL from source file that implies the `--open-browser` switch?

Opens the launch URL: `docfx serve _site --open-browser`
Opens URL from a file:  `docfx serve _site --open-file docs/config.md`

src/Microsoft.DocAsCode.App/RunServe.cs
```diff
@@ -43,11 +47,104 @@ public static void Exec(string folder, string host, int? port)
...
+                var launchUrl = string.IsNullOrEmpty(relativePath)
+                    ? url
+                    : Path.Combine(url, ResolveOutputHtmlRelativePath(folder, relativePath));
+
+                // Start web server.
+                app.Start();
+
+                // Launch browser process.
+
+                Console.WriteLine($"Launching browser with url: {url}.");
```
yufeih, line 63: `launchUrl` or otherwise the URL is already printed in line 47:
```suggestion
                Console.WriteLine($"Launching browser with url: {launchUrl}.");
```

src/Microsoft.DocAsCode.App/RunServe.cs
```diff
@@ -43,11 +47,104 @@ public static void Exec(string folder, string host, int? port)
...
+                using var process = LaunchBrowser(launchUrl);
+
+                // Wait until server exited.
+                app.WaitForShutdown();
+
+                // process object is disposed. (Note:Launched process remain running.after dispose)
```
yufeih, line 69: nit: if we aren't using the `process` object here and disposing it doesn't not terminate the launched process, consider make `LaunchBrowser` return `void`.

src/Microsoft.DocAsCode.App/RunServe.cs
```diff
@@ -43,11 +47,104 @@ public static void Exec(string folder, string host, int? port)
...
+
+            // OSX
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                return Process.Start("open", url);
+            }
+
+            Logger.LogError($"Could not launch the browser process. Unknown OS platform: {RuntimeInformation.OSDescription}");
+        }
+        catch (Exception ex)
```
yufeih, line 143: nit: consider remove this try/catch block and leave it to the top-level exception handler.

src/Microsoft.DocAsCode.App/RunServe.cs
```diff
@@ -43,11 +47,104 @@ public static void Exec(string folder, string host, int? port)
...
+            // Try to resolve output HTML file. (md->html)
+            relativePath = relativePath.Replace('\\', '/'); // Normalize path.
+            var manifestFile = manifest.Files
+                                       .Where(x => FilePathComparer.OSPlatformSensitiveRelativePathComparer.Equals(x.SourceRelativePath, relativePath))
+                                       .FirstOrDefault(x => x.OutputFiles.TryGetValue(".html", out outputFileInfo));
+
+            if (outputFileInfo != null)
+                return outputFileInfo.RelativePath;
+        }
+        catch (Exception)
```
yufeih, line 109: consider remove this `try/catch/throw`, there behavior is almost the same with no exception handling as all exceptions are thrown immediately.

src/Microsoft.DocAsCode.App/RunServe.cs
```diff
@@ -43,11 +47,104 @@ public static void Exec(string folder, string host, int? port)
...
+                    ? url
+                    : Path.Combine(url, ResolveOutputHtmlRelativePath(folder, relativePath));
+
+                // Start web server.
+                app.Start();
+
+                // Launch browser process.
+
+                Console.WriteLine($"Launching browser with url: {url}.");
+                using var process = LaunchBrowser(launchUrl);
```
yufeih, line 64: When `launchUrl` is an empty string, would this call opens the `url` or an empty browser? Opening `url` seems to be a good fallback to me.

src/Microsoft.DocAsCode.App/RunServe.cs
filzrev: As you pointed out. I’ve fixed at latest commit.

src/Microsoft.DocAsCode.App/RunServe.cs
filzrev: Fixed in latest commit.

Originally, I've try to called `process.Kill()` here.
and want to ensure Dispose method called because `Process` might contains unmanaged resources.
Anyway. when serve command completed. docfx.exe is exit. There is no problems change this helper method to return void.

src/Microsoft.DocAsCode.App/RunServe.cs
filzrev: In latest commit.
Changed to fallback to return baseUrl.

src/Microsoft.DocAsCode.App/RunServe.cs
filzrev: In latest commit.
Changed fallback to `url` when fail to resolve relativeUrl.

filzrev:
> Is it simpler to have an --open-file option for opening a specific URL from source file that implies the --open-browser switch?
> Opens the launch URL: docfx serve _site --open-browser
> Opens URL from a file: docfx serve _site --open-file docs/config.md

In latest commit.
Change implementation as you noted.

And founds a bug that using `Path.Combine` to url. It cause test fails on non windows environment.
I've fixed this problems and refactor codes to ensure fallback to baseUrl when failed to resolve relative path.

src/Microsoft.DocAsCode.App/RunServe.cs
filzrev: Changed behavior to log errors. then fallback to return baseUrl.

yufeih (approved):
Looks great now! Thank you @filzrev !

src/docfx/Models/ServeCommand.cs
```diff
@@ -23,10 +23,18 @@ internal class Settings : CommandSettings
         [Description("Specify the port of the hosted website")]
         [CommandOption("-p|--port")]
         public int? Port { get; set; }
+
+        [Description("Open a web browser when the hosted website starts.")]
+        [CommandOption("--open-browser")]
+        public bool OpenBrowser { get; set; }
+
+        [Description("Open a file in a web browser When the hosted website starts,")]
```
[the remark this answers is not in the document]
filzrev: Thanks for your review. It's fixed by #8938

Issue #7736, opened 2021-11-29
Schema file for toc?


klacol:
Hi, I am using VS Code to work on my content in DocFx. Often I have to write proper toc files . In VSCode there is an inline checker against the yaml-schemas, but it looks up a wrong schema and then alle the file is red underlined.

Is there a schema file around for the toc.yml that I can configure for my toc files??

klacol:
The schema file would be something like this:

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "https://dotnet.github.io/docfx/toc.schema.json",
  "title": "Schema for DocFx Table-Of-Content Files (toc.yml)",
  "description": "Reference: https://dotnet.github.io/docfx/tutorial/intro_toc.html",
  "additionalProperties": false,
  "type": "array",
  "items": {
    "type": "object",
    "properties": {
      "name": {
        "description": "Specifies the title of the TOC Item.",
        "type": "string"
      },
      "href": {
        "description": "Specifies the navigation path of the TOC Item. If href is not specified, the TOC Item serves as a container to parent its children TOC Items. If href is an absolute path, the TOC Item navigates to that specified path. If a TOC Item is linking to a relative path, there are three cases: 1. Linking to another TOC File, for example, href: examples/toc.md. 2. Linking to a folder, which means, the value of the link ends with / explicitly, for example, href: examples/ 3. Linking to some local file.",
        "type": "string"
      },
      "items": {
        "description": "Specifies the children TOC Items of current TOC Item.",
        "type": "object",
        "properties": {
          "topicHref": {
            "description": "Specifies the topic href of the TOC Item. It is useful when href is linking to a folder or TOC file or tocHref is used.",
            "type": "string"
          },
          "topicUid": {
            "description": "Specifies the uid of the referenced file. If the value is set, it overwrites the value of topicHref.",
            "type": "string"
          }
        },
        "required": [ "topicHref", "topicUid" ]
      },
      "required": ["name"]
    }
  }
}
```

KalleOlaviNiemitalo:
The schema has to be defined recursively so that TOC.yml can have any depth of nested items.

[...]

PR #8941, opened 2023-07-06
Fix serialization related property name problems

filzrev:

**1.`OriginalHomepage` changes(https://github.com/dotnet/docfx/commit/88104f649f7b17a2f27e52411dd0784423341fe0)**

JSON/YAML Alias names use wrong spelling names. (`OriginallHomepage`)
There is compatibility problems. If these name are changed. 

But this property is added recently (about 4 month ago) and option is `undocumented`. (https://github.com/dotnet/docfx/issues/7736#issuecomment-984613407)

So there is almost no impact by changing these name settings. 

**2. `EnableSourceInfo` changes(https://github.com/dotnet/docfx/commit/88104f649f7b17a2f27e52411dd0784423341fe0)**

Change JSON property name to `enableSourceInfo`.
It seems there is no compatibility problems by this changes.

### DIFF
--- a/src/Microsoft.DocAsCode.DataContracts.Common/TocItemViewModel.cs
+++ b/src/Microsoft.DocAsCode.DataContracts.Common/TocItemViewModel.cs
@@ -88,8 +88,8 @@ public class TocItemViewModel
     [JsonProperty("homepage")]
     public string Homepage { get; set; }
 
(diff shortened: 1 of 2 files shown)


yufeih (approved):
Looks good

PR #8946, opened 2023-07-07
Add FluentAssertions nuget package to test projects

filzrev:
This PR add [FluentAssertions](https://github.com/fluentassertions/fluentassertions) NuGet package to test projects.

**Background**
Currently xUnit's assertions are used for unit tests.
I want to write unit test assertions with `FluentAssertions` assertion helpers. when adding new tests.

`FluentAssertions` is published as `Apache License 2.0` and widely used in [`org:dotnet` repositories](https://github.com/search?q=org%3Adotnet+fluentassertions+language%3AXML&type=code&l=XML)
So I thought there is no major obstacles to adapt this assertion library.

If these changes aren't accepted. feel free to close this PR.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -37,6 +37,7 @@
 
     <!-- Test only -->
     <PackageVersion Include="coverlet.collector" Version="6.0.0" />
+    <PackageVersion Include="FluentAssertions" Version="6.11.0" />
     <PackageVersion Include="Magick.NET-Q16-AnyCPU" Version="13.1.3" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.6.3" />
     <PackageVersion Include="Microsoft.Playwright" Version="1.35.0" />
--- a/test/Directory.Build.props
+++ b/test/Directory.Build.props
@@ -14,6 +14,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="FluentAssertions" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" />
     <!--
       The NuGet.Frameworks package is required to run csproj based tests targeting .NET 7.0.200.


yufeih:
I don't have a lot of personal experience with FluentAssertions. Are there any benefits you see of adopting it except for syntax differences, such as things xunit assertions cannot do easily, or better diagnostic messages?

I'm leaning towards keeping one way of doing things unless there is enough merit where we can say the new approach is the new default.

filzrev:
Thanks for your reply.
> such as things xunit assertions cannot do easily, or better diagnostic messages?

## More readable diagnostic message.
Basically FluentAssertions output more readable diagnostic message.
Because It expose variety of APIs to write assertion.

For example.

**1. FluentAssertions assertion**
```csharp
var results = new[] { "Dummy" };
results.Should().HaveCount(0);
```
**Message:**
> Expected results to contain 0 item(s), but found 1: {"Dummy"}.

**2. xUnit.NET assertion**
```csharp
var results = new[] { "Dummy" };
Assert.True(results.Length== 0);
```
**Message:**
>Assert.True() Failure
>Expected: True
>Actual:   False

It can be confirmed FluentAssertions output contains more assertion context information.

And FluentAssertions have [Subject Identification](https://fluentassertions.com/introduction#subject-identification) feature.
It output *`what's tested`** information when assertion failed.
This feature is useful when writing assertion for object tree.

**For Example**

**Original code**
https://github.com/dotnet/docfx/blob/2fc5c765b298ca34b99bce25ed04bbf193a1fe93/test/docfx.Tests/MetadataCommandTest.cs#L356

**FluentAssertionsassertion**
```csharp
> tocViewModel[2].Items[0].Uid.Should().Be("SubBar");
```
**Message:**
> Expected tocViewModel[2].Items[0].Name to be "SubBarDummy" with a length of 9, but "SubBar" has a length of 6, differs near "r" (index 5).
---

## FluentAssertions features that can't do easily with xunit assertions.

Below is an excerpt  features that I want to use when writing unit tests. 

### [AssertionScope](https://fluentassertions.com/introduction#assertion-scopes)
This feature is useful when writing following tests.

1. Writing multiple assertions in single unit test.  
2. Writing test to multiple items with foreach enumeration.

**Example of `1.`**
- [MetadataCommandTest.cs](https://github.com/dotnet/docfx/blob/2fc5c765b298ca34b99bce25ed04bbf193a1fe93/test/docfx.Tests/MetadataCommandTest.cs#L139-L162)

**Example of `2.`**

```
using var scope = new AssertionScope();
foeach(var path in Directory.EnumerateFiles("TestData"))
{
    // Validate file content using fluent assertins.
}
```
Basically this type of test should be written using xUnit's `InlineData` or `TheoryData`.
But sometimes it's useful writing tests with foreach loop.

## [Object Graph Comparison](https://fluentassertions.com/objectgraphs)

This feature is useful for POCO/DTO class (with no `IEquatable`) data comparison.

```cs
// Act
var actual = JsonUtility.Deserialize<BuildJsonConfig>("<path>")

// Assert
actual.Should().BeEquivalent(new BuildJsonConfig{
  // Set properties to assert.
});
```

## [Extensibility](https://fluentassertions.com/extensibility/#building-your-own-extensions)

It can write custom assertion logics for specific types (e.g. `Manifest`/`OutputFileInfo`,`DirectoryInfo` etc.)
It simplify test assertion codes dramatically. and it can share assertion logics across tests.

yufeih (approved):
Looks great! I like object graph comparison and improved diagnostics.

Issue #8968, opened 2023-07-14
[Feature Request] Add official json schema definition for docfx.json and toc.json


filzrev:
**Is your feature request related to a problem? Please describe.**

When editing `docfx.json` and `toc.yml`.
It'll be helpful intellisense are supported by default. (It require json schema definition)

**Describe the solution you'd like**

Add json schema definition files to docfx repository's following directory.

https://github.com/dotnet/docfx/tree/main/schemas
├ docfx.json
└ toc.json

These json schemas should be generated from C# code to minimize maintenance cost.
(with tool like [NJsonSchema](https://github.com/RicoSuter/NJsonSchema) or [JsonSchema.Net.Generation](https://github.com/gregsdennis/json-everything))
and should be tested by CI automatically.

**Additional context**
There is [unofficial schema definition](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/docfx.json) for `docfx.json`  exists.
and schema for `toc.yml` is discussed at #7736 (`expanded` and `dropdown` options are added for `toc.json` recently)

filzrev:
There are limitations on automatically generated JSON schemas.
And JSON schema generation from C# code is not recommended for **production** usage.
https://docs.json-everything.net/schema/schemagen/schema-generation/#schema-schemagen-disclaimer

So I've decided to manually write JSON schema based on generated schema.

---
Currently It Visual Studio don't supports JSON Schema 2019-09 or later.
It need to write JSON Schema with `Draft-07` version for auto completion.

PR #8969, opened 2023-07-15
Add doc comments to docfx.json related config files.

filzrev:
This PR add doc comments that are related to `docfx.json` configs.
and adding some obsolete attributes.

**Purpose of this PR**
Preparation to create JSON schema document for `docfx.json` from C# code and doc comments. (#8968)

**Add Obsolete attribute**
This PR add `[Obsolete] property to following properties
- `dest` (As commented on #8967).   
   `output` property is added recently. `dest` usage should be deprecated)
- `versions` (It's seems remaining for compatibility purpose)

**Documentation comments**
Almost doc comments are taken from following contents.
- Official document: [docfx.json Configuration Reference](https://dotnet.github.io/docfx/reference/docfx-json-reference.html)
- Help message of docfx.exe.

Some properties are not documented. and no working example found.
So I've added doc comment these property.

** List of undocumented `build` section properties**

|  Name                |  Used for what's purpose
|-------------------|----------------------------------------------|
| `pairing`             |  Used for overwrite folder mappings.
| `xrefTags`           |  Not used in docfx. It's simply exported to xrefmap.yml |
| `tagParameters` |  SchemaDrivenDocumentProcessor handle tags but no usage examples found.
| `versions`           | It seems deprecated. replaced by groups.
| 'groups'             | Used for grouping content/resource files |


src/Docfx.App/Config/BuildJsonConfig.cs
```diff
@@ -18,34 +22,65 @@ internal class BuildJsonConfig
...
     public FileMapping Resource { get; set; }
 
+    /// <summary>
+    /// Contains all the conceptual files which contains yaml header with uid and is intended to override the existing metadata yml files.
+    /// </summary>
     [JsonProperty("overwrite")]
     public FileMapping Overwrite { get; set; }
 
+    /// <summary>
+    /// Specifies pairing that is used for folder redirection rules.
```
yufeih, line 44: We can also obsolete the `Pairing` config, it is to support markdown fragment in schema document processor, which wasn't used by MSLearn and doesn't seem to be used by the community according to https://github.com/search?q=pairing+path%3Adocfx.json&type=code

src/Docfx.App/Config/BuildJsonConfig.cs
```diff
@@ -18,34 +22,65 @@ internal class BuildJsonConfig
...
     public List<ContentPairingInfo> Pairing { get; set; }
 
+    /// <summary>
+    /// Specifies the tags of xrefmap.
+    /// </summary>
     [JsonProperty("xrefTags")]
     public ListWithStringFallback XrefTags { get; set; }
 
+    /// <summary>
+    /// Specifies the urls of xrefmap used by content files. Currently, it supports following scheme: http, https, ftp, file, embedded.
```
yufeih, line 56: It seems we only support HTTP/HTTPS:
https://github.com/dotnet/docfx/blob/main/src/Docfx.Build.Engine/DocumentBuildContext.cs#L301

```suggestion
    /// Specifies the urls of xrefmap used by content files. Supports local file path and HTTP/HTTPS urls.
```

src/Docfx.App/Config/BuildJsonConfig.cs
```diff
@@ -59,51 +94,118 @@ internal class BuildJsonConfig
     [JsonProperty("fileMetadata")]
     public Dictionary<string, FileMetadataPairs> FileMetadata { get; set; }
 
+    /// <summary>
+    /// Specify tag parameters that are used by SchemaDrivenDocumentProcessor's
```
yufeih, line 98: Consider also obsolete this config same reason as the `pairing` config.

filzrev:
I've fixed commented issues with following commits.

- Fix `XRefMaps` comments (https://github.com/dotnet/docfx/pull/8969/commits/5b9936c4f86289c91eef7d081578e9c17af92d0e)
- Add obsolete attributes to `pairning` and `tagParameters` and fix doc comments.
(https://github.com/dotnet/docfx/pull/8969/commits/1d746e93062a5951314e9fd6e383b848742774e2)

> That http/https check is in DocumentBuildContext.LoadXRefMap, which is called by LoadXRefMaps, 
> but does anything call LoadXRefMaps…? It seems these methods are no longer used and 
> the real code is in class XRefMapDownloader...

It seems there are some unused code exists.
I'll create another PR to fix this problems.

yufeih (approved):

filzrev:
Add commit (https://github.com/dotnet/docfx/pull/8969/commits/2a40b8c8a5657599270aed800549ed6097fa465c)
to fix problem that `.gitignore` file created by `docfx init` command don't contains `_site` entry .

PR #8971, opened 2023-07-16
chore: drop Docfx.MarkdigEngine.Validators

yufeih:
The `IMarkdownObjectValidator` was originally designed as an extension mechanism for adding custom markdown validation rules for MSLearn. MSLearn never really used that interface for custom markdown validation and this interface didn't gain much traction in the community.

To add custom markdown validation rules, or custom markdown extensions, use the standard markdig extensions interface, where it exposed the markdown AST at various stages.


filzrev:
> To add custom markdown validation rules, or custom markdown extensions, use the standard markdig extensions interface, where it exposed the markdown AST at various stages.

For `custom Markdown validation rules`.
Docfx seems supporting custom validation by placing `md.style` file to docs root or template directory's `md.styles` folder.

https://github.com/dotnet/docfx/blob/7ab1ad673c2455d9fb129da9a1a95500f7d1d657/src/Docfx.MarkdigEngine/Validation/MarkdownValidatorBuilder.cs#L165-L190

**sample of `md.style` usage in tests**
https://github.com/dotnet/docfx/blob/7ab1ad673c2455d9fb129da9a1a95500f7d1d657/test/Docfx.Build.Engine.Tests/DocumentBuilderTest.cs#L400-L442

This features is not documented. and It seems not be used by any public repository and custom templates.
Is this feature support is continued?
(or is it better to use standard tools like `markdownlint`?)

yufeih:
You are right, the `md.style` file was introduced a long time ago and wasn't documented. It does seem to have low usage and I agree that standard tools like markdown lint is a better alternative. We could drop it as a separate PR.

PR #9015, opened 2023-07-26
Update .NET8 SDK version for CI 

filzrev:

- Update .NET SDK version from `preview.2` to `preview.6`.
- Update `sign` tool version to `0.9.1-beta.23356.1`.

### DIFF
--- a/src/Docfx.Dotnet/DotnetApiCatalog.cs
+++ b/src/Docfx.Dotnet/DotnetApiCatalog.cs
@@ -91,8 +91,24 @@ public static class DotnetApiCatalog
         {
             if (!MSBuildLocator.IsRegistered)
             {
-                var vs = MSBuildLocator.RegisterDefaults() ?? throw new ExtractMetadataException(
-                    $"Cannot find a supported .NET Core SDK. Install .NET Core SDK {Environment.Version.Major}.{Environment.Version.Minor}.x to build .NET API docs.");
+                // Gets non-preview .NET SDKs.
+                var vsInstances = MSBuildLocator.QueryVisualStudioInstances(VisualStudioInstanceQueryOptions.Default)
+                                                .Where(x => !x.MSBuildPath.Contains("-preview."));
+
+                var vs = vsInstances.FirstOrDefault();
+
+                if (vs != null)
+                {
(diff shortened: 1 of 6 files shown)


filzrev:
To avoid `NuGet.Frameworks` related problems (That is referenced by preview version of .NET SDK, But not published to NuGet). 
I've added commit(https://github.com/dotnet/docfx/pull/9015/commits/f311297682c5740585eb9116684872001fba0357) to change behaviors that excluding `preview` version .NET SDK when extracting metadata using MSBuildLocator in `Docfx.Dotnet`.

It means if tests are executed on .NET SDK 8 preview version.   
`Docfx.Dotnet` use non-preview .NET 7 SDK to extract metadata.

Note: Currently. `"-preview."` string is checked to exclude preview version SDK.   
So it may cause same problems when .NET 8 SDK `RC` version released.

src/Docfx.Dotnet/DotnetApiCatalog.cs
```diff
@@ -91,8 +91,24 @@ private static void EnsureMSBuildLocator()
         {
             if (!MSBuildLocator.IsRegistered)
             {
-                var vs = MSBuildLocator.RegisterDefaults() ?? throw new ExtractMetadataException(
-                    $"Cannot find a supported .NET Core SDK. Install .NET Core SDK {Environment.Version.Major}.{Environment.Version.Minor}.x to build .NET API docs.");
+                // Gets non-preview .NET SDKs.
+                var vsInstances = MSBuildLocator.QueryVisualStudioInstances(VisualStudioInstanceQueryOptions.Default)
```
yufeih, line 95: Is it possible to move this custom register logic to test projects? Otherwise we'll ship binaries that cannot build .NET 8 preview projects.

The if check in line 92 checks if a custom MSBuildLocator has been registered upfront.

filzrev: I've modified custom SDK register logics to `docfx.Tests` project's ModuleInitializer.
xUnit execute tests with random order. So add `MSBuildLocator.Unregister();` call if SDK already registered.

yufeih (approved):
Thank you!

PR #9019, opened 2023-07-27
chore: Reorganize editorconfig settings

filzrev:
**What's included in this PR**

- Reorganize `.editorconfig` settings based on `learn.microsoft.com` document defined order.
- Add additional editorconfig settings as `suggestion` level.
- Add file header template

This PR is preparation for following tasks.
- Remove unused code.
- Apply performance related code analysis rules.

PR #9025, opened 2023-07-28
chore: Fix code formatting

filzrev:
In #9019, I've accidently overwrite `csharp_space_between_method_declaration_parameter_list_parentheses` rule to `true`.
and cause diff when running `dotnet format` command.

This PR fix this problem. and re-apply `dotnet format` command to entire solution. 
(It fix namespace orders caused namespace changes from  `Microsoft.*` to `Docfx.*`)


yufeih (approved):
Looks great!

PR #9035, opened 2023-08-02
Fix TOC max height size

MihkelAuv:
Max height of `toc-offcanvas` is calculated by only subtracting header height from viewport size:

https://github.com/dotnet/docfx/blob/a8d1438c20ce62c68d0cb26b687731d3e42a4643/templates/modern/src/layout.scss#L121-L127

This causes a portion of TOC to expand out of the viewport if main part of the page isn't scrolled all the way down:
image

Proposed solution is to also subtract main elements top padding size from viewport height:
image


yufeih (approved):

yufeih:
Thank you @MihkelAuv ！

Issue #2865, opened 2018-06-08
Url prettifier


AndreyAkinshin:
Right now, markdown files (articles) are translated to html files which are available via direct link like `https://mydomain.com/articles/article-name.html`. I want to "prettify" it and get an url like `https://mydomain.com/articles/article-name/`. Actually, I want a copy of `article-name.html` in `https://mydomain.com/articles/article-name/index.html`. Does DocFX provide any API for that?

superyyrrzz:
Not supported for now, but it will be supported in vNext.

AndreyAkinshin:
@superyyrrzz great! Any ETA when it will be available?

yufeih:
We don't have an exact release date yet, it is a work on progress.

yufeih:
This is implemented in v3. 

We use *pretty* url by default: `http://a.com/article-name/`. It can be into `http://a.com/article-name.html` by setting `uglifyUrl: true`

AndreyAkinshin:
@yufeih, awesome, thanks!
By the way, it will be possible to automatically generate redirects from ugly urls to pretty urls? (e.g., http://a.com/article-name.html -> http://a.com/article-name/). With such redirects, all external links to old urls will be valid after prettification.

yufeih:
This is a static switch. `docfx` has no idea when you made the switch. It'll generate redirects for every new file.

Automatic redirect generation is best handled by other tools that know the exact files to be redirected during a change.

arakis:
Today, there's still no v3 released. So, this is a small quick and dirty workarround:
```powershell
$destinationRoot = "path/to/output"

Copy-Item -Path $sourceRoot -Recurse -Destination $destinationRoot -Container

[RegEx]$Search = '(index\.html)"'
$Replace = '"'

[RegEx]$Search2 = '(\.html)"'
$Replace2 = '"'

ForEach ($File in (Get-ChildItem -Path $destinationRoot -Recurse -File)) {
    (Get-Content $File.FullName) -Replace $Search,$Replace -Replace $Search2,$Replace2 | 
        Set-Content $File.FullName
}
```
Live sample: https://www.aximo.games (hostet on github pages)

AndreyAkinshin:
@yufeih is it possible to backport this feature to v2?

yufeih:
Reopen as the feature isn't available in v2.

mjegob:
@yufeih any ETA when this feature will be available?
Really looking forward to it! 
Thanks! 😊

[...]

PR #9036, opened 2023-08-02
feat(modern-template): respect url normalization for toc

Lulalaby:
### The current problem

If you have something like .htaccess url rewrite active, the toc expansion does not work.
Example rewrite:
```htaccess
RewriteEngine On
RewriteBase /


RewriteRule ^index\.html$ / [R=301,L]
RewriteRule ^(.*)/index\.html$ /$1/ [R=301,L]

RewriteCond %{THE_REQUEST} ^[A-Z]{3,}\s([^.]+)\.html [NC]
RewriteRule ^ %1 [R=301,L]

RewriteCond %{REQUEST_FILENAME}.html -f
RewriteRule ^([^\.]+)$ $1.html [NC,L]
```

#### Screenshots
without url rewrite
with url rewrite

### Proposed Solution
By adding a config option called `_urlNormalizationActive` and parsing this in the toc.ts of the `modern` template as `docfx:urlnormalizationactive`, we can achieve the toc expansion.

#### Screenshots

working with url rewrite


yufeih (commented):
This fix feels a bit hacky. Would #2865 address your scenario once it lands?

Lulalaby:
Yeah that's basically what i mean, we should keep in mind for that issue tho, that it needs to work seamless with existing environments if needed.
I agree that this pr is hacky, I'm gonna close it for now.

PR #9038, opened 2023-08-02
fix(modern-template): nav & toc active state breaks with external urls

Lulalaby:
### The current problem

When having an external url defined in the toc.yml, the active state toggler sets all external links & the current page as active

#### Screenshots
invalid active state

### Proposed Solution
By checking whether the items `hostname` or `port` differs from `window.location`, we can set the active state correctly.

#### Screenshots

correct active state


templates/modern/src/nav.ts
```diff
@@ -131,6 +130,9 @@ function findActiveItem(items: (NavItem | NavItemContainer)[]): NavItem {
   let activeItem: NavItem
   let maxPrefix = 0
   for (const item of items.map(i => 'items' in i ? i.items : i).flat()) {
+    if (item.href.hostname !== window.location.hostname || item.href.port !== window.location.port) {
```
yufeih, line 133: consider move `isExternal` to `helper.ts` and reuse the function here?

templates/modern/src/toc.ts
```diff
@@ -54,7 +54,7 @@ export async function renderToc(): Promise<TocNode[]> {
     if (node.href) {
       const url = new URL(node.href, tocUrl)
       node.href = url.href
-      active = normalizeUrlPath(url) === normalizeUrlPath(window.location)
+      active = isExternal(url) ? false : normalizeUrlPath(url) === normalizeUrlPath(window.location)
```
yufeih, line 57: Consider add an external URL node to `samples/seed/articles/toc.yml` as a test case. Snapshot tests should then make sure we don't regress this bug in the future.

templates/modern/src/nav.ts
Lulalaby: Will do

Lulalaby:
@yufeih i'm unsure about how to specially add a test for the "active" variable, but I can't spend more time to figure it out.
Gonna go in a few hours to my doc, kinda emergency surgical intervention.. so ig i'm out for a while. 

I hope that pr is good to go, otherwise I'd be happy if someone could fix it.

Lulalaby:
image
I noticed that it's difficult to test in forks. Can we change that to take the repo / org from var?

Lulalaby:
btw what's up with `/home/runner/work/docfx/docfx/samples/seed/articles/images/seed.jpg: No template bundle found for Resource, model will be ignored.`

I noticed it since upgrading docfx

Sorry for the ci spam also, I tried fixing the snapshot ci

.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches: [ main, feature/*, hotfix/* ]
   push:
     branches: [ main, feature/*, hotfix/* ]
+  workflow_dispatch:
```
yufeih, line 7: What is the purpose of this change?

Lulalaby: Leftover from the try to get the snapshot ci running on pull requests. Sorry about that

yufeih (approved):
Thank you @Lulalaby !

PR #9053, opened 2023-08-08
Fix affix max height size

MihkelAuv:
Similar fix as #9035 for modern template `affix` element ("In this article" section).


yufeih (approved):
Thank you @MihkelAuv !

PR #9066, opened 2023-08-11
feat: add type definition to REST API pages

### LINKED ISSUE #2072: Feature request: Add swagger type definitions to REST API documentation
When generating a REST API documentation from a swagger-file I would expect that the types from the "definitions" section would be included in the documentation, especially if the type is used as a method parameter.

yufeih:
Integrates @mcm-ham 's change to show type definitions for REST API pages: https://github.com/dotnet/docfx/issues/2072#issuecomment-531634718 with a few tweaks:
* Revert the added `in` property
* Fixed array type `[]` inconsistently showed as part of the link in definitions section.

Thank you @mcm-ham !

fixes #2072


templates/default/partials/rest.child.tmpl.partial
```diff
@@ -41,7 +41,18 @@
 {{#parameters}}
     <tr>
       <td><span class="parametername">{{#required}}*{{/required}}{{name}}</span></td>
-      <td>{{type}}</td>
+      <td>
+          {{^schema.cType}}
+            {{schema.type}}
```
mihtjel, line 46: For non-complex types, I think this should rightly be `{{type}}` (and `{{format}}` below). At least that's working for me. I may make a PR for this.

PR #9074, opened 2023-08-14
perf: Reduce lock wait when checking git command exists or not

filzrev:
Modify GitUtility's `ExistGitCommand` method to `Lazy<bool>` property.
And Initialize that property on startup phase.

**Background**
`ExistGitCommand()` takes about 70-100ms.
On Release build. About 3～6 threads are blocked by double-checked locking's  lock when worker thread's start operations.
To reduce lock wait. Add `Lazy<bool>` property initialize logic to `RunBuild.cs`.

### DIFF
--- a/src/Docfx.App/RunBuild.cs
+++ b/src/Docfx.App/RunBuild.cs
@@ -27,7 +27,8 @@ internal static class RunBuild
 
         if (!config.DisableGitFeatures)
         {
-            CheckGitCommandExists();
+            // Initialize Lazy<bool> property by ThreadPool thread.(It takes about 50-100 ms)
+            Task.Run(() => GitUtility.ExistGitCommand.Value);
         }
 
         // TODO: remove BaseDirectory from Config, it may cause potential issue when abused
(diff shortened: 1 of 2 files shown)


src/Docfx.App/RunBuild.cs
```diff
@@ -22,6 +24,12 @@ public static string Exec(BuildJsonConfig config, BuildOptions options, string c
 
         EnvironmentContext.SetGitFeaturesDisabled(config.DisableGitFeatures);
         EnvironmentContext.SetBaseDirectory(Path.GetFullPath(string.IsNullOrEmpty(configDirectory) ? Directory.GetCurrentDirectory() : configDirectory));
+
+        if (!config.DisableGitFeatures)
+        {
+            CheckGitCommandExists();
```
yufeih, line 30: nit: consider use `Task.Run` so we don't need to explicitly catch here:
```suggestion
            Task.Run(() => GitUtility.ExistGitCommand.Value);
```

yufeih (approved):
Looks great! Thanks for spotting the issue.

src/Docfx.App/RunBuild.cs
filzrev: Refactor code  as suggested that using `Task.Run`.

`Task.Run` without await/Wait should not throw any Exception (because thrown exception can be handled at `TaskScheduler.UnobservedTaskException` only) So I've moved try-catch logics to Lazy initializer.

PR #9090, opened 2023-08-15
perf: Optimize templates resource loading peformance

filzrev:
Small perf optimization related to templates resource loading.
- Remove unnecessary `CreateTemplateResource()` call.
- Call `ToArray()` before `IResourceFileReader::Names` property is set.  
  Because Names are enumerated multiple times.
- Add `HashSet<string>` for Names check. Because `ResourceFileReader::GetResourceStream` is called multiple times.
  For example. it is called about 443 times for `samples/seed` project.
  And Names contains 71/93 elements (For default/modern templates).


yufeih (approved):
Aha, improper use of LINQ easily cause perf problems.

PR #9091, opened 2023-08-17
chore: Cleanup tests and fix spellings

filzrev:
- Fix spelling miss in verbose log message from `week type` to `weak type".
- Remove code that explicitly setting `ServicePointManager.SecurityProtocol` to `Tls12`  
   Because there are no supported OS/.NET versions that require this settings.
- Remove `TestSchemaReachedLimits` tests. Because `Json.NET schema` is not used anymore.
- Rename test method signature and traits that using Markdig. 
- Modify skipped tests. that pass tests with small modifications.
- Remove `TestMetadataCommandFromCSProjectWithVsinstalldirEnvSet` test because it's not actually used.

### DIFF
--- a/src/Docfx.Build.Engine/ManifestProcessor.cs
+++ b/src/Docfx.Build.Engine/ManifestProcessor.cs
@@ -77,7 +77,7 @@ internal class ManifestProcessor
 
     private void NormalizeToObject()
     {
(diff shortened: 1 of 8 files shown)


yufeih (approved):
Looks great!

Issue #9067, opened 2023-08-11
[Feature Request] Suppress Codecov indirect coverage changes reports


filzrev:
**Is your feature request related to a problem? Please describe.**
When PR is created.
Codecov randomly reports [indirect coverage changes](https://app.codecov.io/gh/dotnet/docfx/pull/9066/indirect-changes?src=pr&el=tree-more&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=dotnet).  
relating to following two files.

- https://app.codecov.io/gh/dotnet/docfx/pull/9056/blob/src/Docfx.Common/Git/GitUtility.cs
    -> It seems test is executed under `detached head` source randomly?
- https://app.codecov.io/gh/dotnet/docfx/pull/9056/blob/src/Docfx.YamlSerialization/Helpers/ArrayDictionary.cs
    -> It seems related test is not called or failed collect coverage data?

**Describe the solution you'd like**

I can't determine what's cause these coverage problems on CI environment.
Simple solution is to add following attribute to methods that cause coverage diff.

> [ExcludeFromCodeCoverage(Justification = "When running CI Build. Codecov report `Indirect changes`.")]

filzrev:
`GitUtility.cs` related `indirect changes` are suppressed by explicitly set `ExcludeFromCodeCoverage`.

`ArrayDictionary.cs` related `indirect changes` seems occurred by parallel execution.
If set `maxParallelism` to `1`. This branch is not called.

On GitHub hosted runner. there are 2 CPU core.  so race condition occurred occasionally.
On local environment. double-checked locking's lock wait is frequently occurs.
Thought it is not spend large lock-wait times. It might be better to optimize `ArrayDictionary.cs` code to lock-free.

PR #9102, opened 2023-08-18
chore: Add code coverage excusions

filzrev:
**What included in this PR**
Add `ExcludeFromCodeCoverage` attribute that randomly cause CodeCov coverage indirect diff. (See #9067)

I don't know what cause `detached branch` state (Checkout by commitId in some condition?)
But I though it's safe to exclude from code coverage.


yufeih (approved):
Looks great! Thank you @filzrev

PR #9119, opened 2023-08-22
perf: Change existing struct to readonly struct

filzrev:
- Change `struct` model to `readonly struct`.

This change ensure model is not modified after created.
And It can reduce object copy when passing struct as argument. and `defensive copy` of struct object.
(Thought It's not measurable perf improvement.)


filzrev:
There is another `FileInfo` issue that it contains `GroupInfo` reference type.
So `FileLinkInfo` is not considered as immutable object.
And it fails to compare `structure equity` when `GroupInfo` property have value.

https://github.com/dotnet/docfx/blob/d4b18eb9961c1634b86274dc98a1a301ad43800e/test/Docfx.Build.Common.Tests/FileLinkInfoTest.cs#L31-L33

It might to be better to change `FileLinkInfo` to class. And implement `IEquatable<T>` for FileInfo (and GroupInfo).

src/Docfx.Common/FileLinkInfo.cs
```diff
@@ -5,28 +5,28 @@
 
 namespace Docfx.Common;
 
-public struct FileLinkInfo
+public readonly struct FileLinkInfo
```
yufeih, line 8: It might be better to just turn this into a class, because the struct would be boxed when calling `GenerateHref(IFileLinkInfo href);`
```suggestion
public class FileLinkInfo
```

filzrev: I've committed struct -> class changes. and relevant test code. 

> IIRC some methods create the FileLinkInfo instance only if there is an ICustomHrefGenerator, but others don't check.

It seems `FileLinkInfo`  class is used for calculatating Href.

yufeih (approved):

PR #9131, opened 2023-08-24
chore: Change output html file encoding to UTF-8 (No BOM)

filzrev:
- Change output HTML encoding from UTF-8(BOM) to UTF-8 (No BOM).

**Background**
Currently docfx output HTML files using UTF-8 (with BOM) encoding.
And other non-HTML files are using UTF-8 (No BOM).

Currently docfx output HTML with following meta tag. so it seems BOM is not required.

> <meta charset="utf-8">

If there is any reason to use `UTF-8(BOM)` close this PR.
(For example. Snapshot tests library requires BOM)

### DIFF
--- a/src/Docfx.Build.Engine/PostProcessors/HtmlPostProcessor.cs
+++ b/src/Docfx.Build.Engine/PostProcessors/HtmlPostProcessor.cs
@@ -13,6 +13,8 @@ namespace Docfx.Build.Engine;
 
 internal sealed class HtmlPostProcessor : IPostProcessor
 {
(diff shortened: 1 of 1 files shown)


filzrev:
> What's the meta tag? I don't see it in the description.

Sorry it indicate  `<meta charset="utf-8">` (It's contains HTML tags so  it's not rendered when using quote syntax)

**default template**
https://github.com/dotnet/docfx/blob/main/templates/default/partials/head.tmpl.partial#L4

**modern template**
https://github.com/dotnet/docfx/blob/main/templates/modern/layout/_master.tmpl#L9

yufeih (approved):
Looks great!

Issue #8872, opened 2023-06-13
[Feature Request] Expose RunBuild.Exec API and related configs


filzrev:
**Is your feature request related to a problem? Please describe.**

Currently `Microsoft.DocAsCode.App` expose [`Docset.Build`](https://github.com/dotnet/docfx/blob/359ab008cba2d9891a619669e1381227541b8130/src/Microsoft.DocAsCode.App/Docset.cs#L31) API only.

and there are some limitation exists.

- `Docset.Build` API require `docfx.json` file path.   
    It can't directly pass configuration via code.  (Because `BuildJsonConfig` is internal)
- `ConsoleLogListener` is always added on startup.  It can't use custom log listeners only.

**Describe the solution you'd like**

Change `RunBuild.Exec` APIs and related configs (`BuildJsonConfig`) from internal to public.

I thought all internal APIs used by `docfx.exe` should be available also when using `Microsoft.DocAsCode.App`.
It may better to remove `<InternalsVisibleTo Include="docfx" />` settings and modify affected classes.

PR #9140, opened 2023-08-28
chore: Refactor docfx command related code

filzrev:
- ~~Change docfx command related API/Models from internal to public.~~

- ~~Remove `<InternalsVisibleTo Include="docfx" />` settings~~
- Move constant string `docfx.json` definition to `Docfx.DataContracts.Common` project.
- Add `RunMetadata.cs` wrapper to use `DotnetApiCatalog` API.

**Background**
As described at #8872.
Current `Docfx.App` package expose limited API set only.
By this PR, `Docfx.App` user can consume same APIs that is used by `docfx.exe`.

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -12,7 +12,7 @@ runs:
       dotnet-version: |
         6.x
         7.x
-        8.0.100-preview.6.23330.14
+        8.0.100-preview.7.23376.3
 
   - run: npm install
     shell: bash
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,15 +8,15 @@
     <PackageVersion Include="HtmlAgilityPack" Version="1.11.52" />
     <PackageVersion Include="ICSharpCode.Decompiler" Version="8.1.0.7455" />
     <PackageVersion Include="IgnoresAccessChecksToGenerator" Version="0.6.0" />
-    <PackageVersion Include="Jint" Version="3.0.0-beta-2050" />
+    <PackageVersion Include="Jint" Version="3.0.0-beta-2051" />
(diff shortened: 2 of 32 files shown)


yufeih:
Thanks for putting this together. I need some time to examine the public API surface and may likely tweak it a little bit.

yufeih:
Here are some of the public API guidelines for the config objects I have in mind:
1. The C# property name should match the JSON config name, we could enforce it by applying `[JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]` at the class level. We also don't seem to need the `[Serializable]` attributes.
2. Internalize all obsolete properties
3. Make the config object immutable so we don't need to make defensive copies or verify mutability:
    * Use `init` over `set`
    * Use `ImmutableArray<T>` over `List<T>` or `T[]`
    * Use `ImmutableDictionary<TKey, TValue>` over `Dictionary<TKey, TValue>`
4. For types that allows both short-hand and expanded forms, expose only the expanded form with a custom JSON converter applied at property level:
    * `ListWithStringFallback` -> `ImmutableArray<string>`
    * `FileMetadataPairs` -> `ImmutableDictionary<string, object>`
    * `FileItems` -> `ImmutableArray<string>`
    * `FileMapping` -> `ImmutableArray<FileMappingItem>`

filzrev:
Thank you for your explanations about public API guidelines.
But `3.` and `4.` requires additional codes and it seems hard to maintenance.

So I decided to to use `IgnoresAccessChecksToGenerator` to ignore internal modifier on my code.
And modify this PR with following changes only.
- Move constant string docfx.json definition to Docfx.DataContracts.Common project.
- Add RunMetadata.cs wrapper to use DotnetApiCatalog API.

yufeih (approved):
Thank you @filzrev

PR #9164, opened 2023-09-02
chore: Remove unused code

filzrev:
- Remove unused code from DocumentProcessors

It seems these code are not used.  and it can be replaced with `JsonUtility`'s methods.


yufeih (approved):
Awesome!

Issue #661, opened 2016-09-01
Full-search implementation concerns


evil-shrike:
First of all full-search is awesome. Really cool. But let me criticize a bit.

1. Why do we need search-stopwords.json?
lurn already contains built-in stopwords for English. But you remove default stopWordFilter, then load a separate stopwords index file and generate a filter based on it. Why? 
That search-stopwords.json contains the same stopwords as default builtin filter!
Moreover lunr addons for languages (from https://github.com/MihaiValentin/lunr-languages) contains their own stop words.

2. Why not build index in build-time? Why instead do you load json in run-time and then add item by item into index. It can be done (and usually done) in build time. Then in run-time we can just load an index file:
```
$.getJSON("index.json", function (data) {
  engine = lunr.Index.load(data);
})
```
That's all.
I understand that you enrich search results with title and keywords which are absent in `lunr.search`'s result. But it can be done via additional index file.

3. no i18n 
Index should be built with honor of other languages. lunr natively supports only English. For additional languages support we need to add addons (from https://github.com/MihaiValentin/lunr-languages):
- in buildtime:
```
var lunr = require('lunr');
require('./lunr.stemmer.support.js')(lunr);
require('./lunr.ru.js')(lunr);
require('./lunr.multi.js')(lunr);
var lunrIdx = lunr(function() {
  this.use(lunr.multiLanguage('en', 'ru'));
  // config ref/fields
});
```
- in runtime:
```
lunr.multiLanguage('en', 'ru');
engine = lunr.Index.load(data);
``
I can create a template for customization of index building but I think it should possible without template customization. Also please see #650 - these're problems with encoding of extracted keywords for indexing.

qinezh:
Thanks for @evil-shrike 's comments, it's quite reasonable and insightful. I'm glad to share something with your concerns: 

1. For the first question, actually, it's a way to solve the #279 , so users can customize the stop-words to avoid the issue. 
2. For the second one, the `index.json` generated by DocFX is not kind of serialised data which `lunr.Index.load` need. And additional index file would make the process more complicated? 
3. For the third one, I agree with you, we should support other languages. 

Thanks @evil-shrike . Feel free to share it here if you have more concerns.

evil-shrike:
> the index.json generated by DocFX is not kind of serialised data which lunr.Index.load need.
sure, we have to build it - similar to that how it's built in runtime currently and only call `index.toJSON` at the end, and we'll have index json for `Index.load`. 
[...]

PR #9182, opened 2023-09-10
chore: Change search start condition & add onerror handler

filzrev:
- Add search `start condition` based on query string's UTF-8 byte count.
- Add `onerror` log handler for worker.

**Background**
I'm trying to replace `search-worker.min.js` from [Lunr.js](https://lunrjs.com) to [Fuse.js](https://www.fusejs.io) based.
`Fuse.js` supports non-ascii text search by default. so I want to change default condition to start search operation.  
(Currently at least  3-chars required)

And when modifying custom `search-worker.min.js` file.
Sometimes unhandled error is not logged when failed to load `search-worker.min.js` itself. 
(It occurs when using Chrome/Edge browser. When using Firefox. error log is recorded always)
So I've added `onerror` handler.  
(On Chrome/Edge environment `event.message` might be undefined thought).

**Tested Environments**
I've tested on Chrome/Edge/Firefox browsers(on Windows environment)
And confirmed `TextEncoder` API is available on all major platforms.
https://developer.mozilla.org/en-US/docs/Web/API/TextEncoder#browser_compatibility

**Additional Info**
To replacing `Fuse.js` based search. use following  `search-worker.min.js` file.
https://gist.github.com/filzrev/9a046c40f6df63d01f40018d8a19bd47

### DIFF
--- a/templates/modern/src/search.ts
+++ b/templates/modern/src/search.ts
@@ -48,7 +48,7 @@ export function enableSearch() {
   function onSearchQueryInput() {
     query = searchQuery.value
 
-    if (!query || (query.length < 3 && textEncoder.encode(query).length < 3)) {
+    if (query.length < 3 && textEncoder.encode(query).length < 3) {
       document.body.removeAttribute('data-search')
     } else {
       worker.postMessage({ q: query })
--- a/templates/modern/src/search.ts
+++ b/templates/modern/src/search.ts
@@ -11,6 +11,7 @@ type SearchHit = {
   keywords: string
 }
 
+const textEncoder = new TextEncoder()
 let query
 
 /**
@@ -24,6 +25,11 @@ export function enableSearch() {
 
   const relHref = meta('docfx:rel') || ''
   const worker = new Worker(relHref + 'public/search-worker.min.js', { type: 'module' })
+
+  worker.onerror = event => {
+    console.error("Error occurred at search-worker. message: " + event.message);
+  }
+
   worker.onmessage = function(oEvent) {
     switch (oEvent.data.e) {
       case 'index-ready':
@@ -41,7 +47,8 @@ export function enableSearch() {
 
   function onSearchQueryInput() {
     query = searchQuery.value
-    if (query.length < 3) {
+
+    if (!query || (query.length < 3 && textEncoder.encode(query).length < 3)) {
       document.body.removeAttribute('data-search')
     } else {
       worker.postMessage({ q: query })
--- a/templates/modern/src/search.ts
+++ b/templates/modern/src/search.ts
@@ -27,7 +27,7 @@ export function enableSearch() {
   const worker = new Worker(relHref + 'public/search-worker.min.js', { type: 'module' })
 
   worker.onerror = event => {
-    console.error("Error occurred at search-worker. message: " + event.message);
+    console.error('Error occurred at search-worker. message: ' + event.message)
   }
 
   worker.onmessage = function(oEvent) {
--- a/templates/modern/src/search.ts
+++ b/templates/modern/src/search.ts
@@ -11,7 +11,6 @@ type SearchHit = {
   keywords: string
 }
 
-const textEncoder = new TextEncoder()
 let query
 
 /**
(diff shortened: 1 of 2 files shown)


templates/modern/src/search.ts
```diff
@@ -41,7 +47,8 @@ export function enableSearch() {
 
   function onSearchQueryInput() {
     query = searchQuery.value
-    if (query.length < 3) {
+
+    if (!query || (query.length < 3 && textEncoder.encode(query).length < 3)) {
```
yufeih, line 51:
```csharp
    if (!query || (query.length < 3 && textEncoder.encode(query).length < 3)) {
```

Could you help me understand the purpose of this change?

filzrev: > if (!query || (query.length < 3 && textEncoder.encode(query).length < 3)) {

### `!query`
It seems not required in current implementation.  I'll remove this code.  
(This check is required when getting input text from `event.data` that passed as argument)

### `query.length < 3`
It is required to keep original behavior and to skip TextEncoder API call if it's not necessary case(query.length>=3). 

### `textEncoder.encode(query).length < 3`
Encode text to UTF-8 bytes and get byte length.
UTF-8 represents single char with 1-3 bytes (Not exactly, but almost cases) 

`< 3` search start condition is determined to support `single character search` for 3-byte chars (e.g. [CJK character](https://en.wikipedia.org/wiki/CJK_characters)).
For these language. It's meaningful to search with single or double character.

2-byte characters(e.g. Latin/Arabic alphabet) seems not be useful to enable single character search.
so I've explicitly set condition to `< 3`.
As a side effect. When 2-byte characters is used. search starts when 2 chars are inputted thought.

filzrev: I've removed unused `!query` condition.

And summarize search start condition.

| Char category       |  Search start condition         |
|---------------------|--------------------------------|
| ASCII                      | At least 3-chars                    |
| UTF-8 2byte char  | At least 2-chars                    |
| UTF-8 3byte char  | single char                            |

Anyway current `lunr.js` based search don't support  non-ASCII text search.
This PR changes don't affect docfx's search behavior.

yufeih: Do you think it make sense to start searching immediately when there is a char, regardless of language?

filzrev: Is it means changing start search condition to `if (query.length == 0)` ?

If it's acceptable to change existing behavior. (At least 3-chars)
It's more simpler solution.
I'll commit above changes and re-test.

There is concern about performance when too much results hit with single char.
As far as I've tested with Azure docs content (About 2000 HTML files generated)
https://github.com/MicrosoftDocs/azure-docs

It's seems acceptable performance
.
In that case. **index build time** and `search-worker.min.js`'s memory usage is problematic.
 `lunr.js` based search are enabled about 5 seconds after page load, and search-worker consume memory about 225MB.

It might be better to introduce prebuild index discussed at https://github.com/dotnet/docfx/issues/661

yufeih: > In that case. index build time and search-worker.min.js's memory usage is problematic.

Is this a regression introduced by changing `query.length < 3` to `query.length == 0` or a known issue discussed at #661 ? Assuming the latter, we could change the "at least 3 chars" behavior in this PR.

filzrev: It's separated issue.and not related in this PR's changes.
I'll commit above changes later.

filzrev: Change search start condition to `if (query.length == 0)` .  (It is equivalent to `if(query === '')`)

And when testing search with large docset.
Sometimes It failed to hide search results. when query is empty.

It is caused by event handling orders. when search operation take times..
Empty string's event is processed first. then search result is rendered.
So I've added additional check to hide search results when query is empty.

filzrev: Note: this PR modify `modern` template search functionality only.
`default` template's search(`default\styles\docfx.js`) script is not changed.

yufeih (approved):

yufeih:
Thank you @filzrev

PR #9201, opened 2023-09-14
chore: Remove unused localsearch function from default template

### LINKED ISSUE #9193: [Feature Request] Remove `localSearch` function from `default` template'
It seems there is code for worker is not supported environment.
https://github.com/dotnet/docfx/blob/0ce4839028d461e7fb1431414c7bd9820ecebf37/templates/default/styles/docfx.js#L131-L136

But when running this code path.  It failed because `lunr.js` is not defined.

Remove `localSearch` function from `default` template's script.
Because major browsers are already supports web worker APIs already.

filzrev:
Fix #9193.

- Delete `localsearch()` function.
- Modify `window.Worker` check code. (Original code checking `window.worker`. but it's always `undefined`)  
- Remove `!worker` check (Same as `modern template`)
- Add `worker.onerror` event handler that recently added to `modern` template.
- Remove `console.log("using local search");` log

### DIFF
--- a/templates/default/styles/docfx.js
+++ b/templates/default/styles/docfx.js
@@ -128,13 +128,10 @@ $(function () {
       return;
     }
     try {
-      var worker = new Worker(relHref + 'styles/search-worker.min.js');
-      if (!worker && !window.worker) {
-        localSearch();
-      } else {
-        webWorkerSearch();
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Looks great!

PR #9227, opened 2023-09-20
chore: Add IDisposable interface to Manifest class

filzrev:
- Add `IDisposable` interface  to `Manifest` class. 
- Add `using` statement to manifest variables. 
- Move DocumentBuilder's cleanup operations to finally block.
- Add `Dispose()` code that are warned as  [CA2013](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2213) (occurred when enabling this warnings explicitly)

**Background**
When calling `RunBuild.Exec` multiple times inside process.
It seems memory usage is gradually increasing. (When running with full GC collect operations)
It's not large size. but some resources might not be released properly.


yufeih:
> It seems memory usage is gradually increasing.

Do we have a memory profile comparison of what objects are causing the memory creeping and who is pinning them? I want to make sure we are fixing the right thing.

filzrev:
If execute docfx with `--maxParallelism 1` option.
then Memory usage remains constant.

So it seems some thread local data is cached anywhere. (e.g. ArrayPool's buffer)
And take times to stabilize memory usage.

PR #9307, opened 2023-10-18
fix: update mermaid theme on theme change

### LINKED ISSUE #9306: [Feature Request] Dynamically change mermaid theme based on selected appearance mode
**Is your feature request related to a problem?**

Mermaid.js supports different [built-in themes](https://mermaid.js.org/config/theming.html#available-themes). Some are tailored for lighter background (like `forest` or `neutral`) while others are better suited for darker (like `dark`). [On the mermaid.js documentation ](https://mermaid.js.org/syntax/sequenceDiagram.html) if you switch from light appearance mode to dark then a different theme is applied for the diagrams. 

We would like to ask a similar solution from docfx.

The optimal solution would be that the mermaid diagrams' theme are changing  between `default` and `dark` automatically (_if the diagram does not overwrite the site-wide settings_). In other words, if you don't explicitly set the theme then it is appearance mode aware otherwise it is fixed to the chosen theme.

If we could do it by our own inside a [preprocessor](https://dotnet.github.io/docfx/tutorial/intro_template.html#module) in somewhat similar way like [mermaid docs did](https://github.com/mermaid-js/mermaid/blob/33a5fc459d852e2af42c749ac982be38cc9802e8/packages/mermaid/src/docs/.vitepress/theme/Mermaid.vue#L59C12-L59C24) that would also great.
```typescript
const renderChart = async () => {
  const hasDarkClass = document.documentElement.classList.contains('dark');
  const mermaidConfig = {
     ...   
    theme: hasDarkClass ? 'dark' : 'default',
  };
  ...
}
```

**Additional context**

We would like to add more and more [mermaid diagrams to the pollydocs.org](https://www.pollydocs.org/strategies/timeout.html#diagrams). We have temporarily chosen the `dark` theme because it is readable with dark appearance mode but it is not as legible with light mode. 

[We have tried](https://github.com/App-vNext/Polly/pull/1702#discussion_r1361922992) to use the `default` theme for both appearance modes but in dark mode the diagram is terrible.

yufeih:
fixes #9306


peter-csala:
Thank you @yufeih for the quick fix 👍

PR #9341, opened 2023-10-25
chore: add System.Text.Json attributes

filzrev:
**What included in this PR**
This PR add following System.Text.Json's attributes to existing Newtonsoft.Json based models.
  - `JsonPropertyName`
  - `JsonIgnore`
  - `JsonExtensionData`
  - `JsonConstructor`
 
**What's tested on my local environment**
In this PR. attributes are manually added.
So I've verified attributes are aligned with existing `Newtonsoft.Json` value with reflection API.

**Background**
This PR is intended to be used for #8968. (Because `JsonSchema.Net` support only `System.Text.Json` attributes)
So JsonConverter for `System.Text.Json` is not added in this PR.

To fully migrate to  `System.Text.Json`. #8137 to be resolved.

### DIFF
--- a/src/Docfx.App/Config/BuildJsonConfig.cs
+++ b/src/Docfx.App/Config/BuildJsonConfig.cs
@@ -1,6 +1,7 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Text.Json.Serialization;
 using Docfx.Common;
 using Docfx.Plugins;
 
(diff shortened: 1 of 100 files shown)


yufeih (approved):
Looks good.

yufeih:
Does [`OneOf`](https://github.com/dotnet/docfx/blob/main/src/Docfx.Build/ApiPage/ApiPage.cs#L73C12-L73C17) helps with modeling union types for JSON schema generation?

filzrev:
> Does [OneOf](https://github.com/dotnet/docfx/blob/main/src/Docfx.Build/ApiPage/ApiPage.cs#L73C12-L73C17) helps with modeling union types for JSON schema generation?

It can't automatically generate schema from C# code that use custom `JsonConverter`.
So it needs to manually adding schemas for these types.
I thought it's required same operations when using `OneOf` library.

Issue #9203, opened 2023-09-15
[Bug] .NET API bug "must have an implementation part because it has accessibility modifiers."


xavierjohn:
**Describe the bug**
Partial methods results in error "must have an implementation part because it has accessibility modifiers."

**To Reproduce**
1. Create a C# project with the following class
```
using System.Text.RegularExpressions;

internal partial class Class1
{
    [GeneratedRegex("\\A(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?)\\Z", RegexOptions.IgnoreCase, "en-US")]
    private static partial Regex EmailRegEx();
}
```
2. Compile with docfx

**Expected behavior**
Expect the code to compile using DocFx. The code compiles fine in Visual Studio.

**Context (please complete the following information):**
- OS: Windows
- Docfx version: [2.70.3+3968fc5e92ebc2e4a2d97df0105000af2ec7eb82]
- .NET version: [.NET Core SDK 7.0.401]

- `docfx.json` config

```json
// Paste docfx.json config, remove sensitive info
```

- Exceptions

```csharp
// Paste the full exception with stacktrace here, remove sensitive info
```

- Errors and warnings

```csharp
// Paste warnings or errors related to this bug here, remove sensitive info
Class1.cs(7,34): error CS8795: Partial method 'Class1.EmailRegEx()' must have an implementation part because it has accessibility modifiers.
```

- .NET info

```csharp
// Paste `dotnet --info` output here
.NET SDK:
 Version:   7.0.401
 Commit:    eb26aacfec

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\7.0.401\

Host:
  Version:      7.0.11
  Architecture: x64
  Commit:       ecb34f85ec

.NET SDKs installed:
  3.1.426 [C:\Program Files\dotnet\sdk]
  6.0.414 [C:\Program Files\dotnet\sdk]
  7.0.102 [C:\Program Files\dotnet\sdk]
  7.0.401 [C:\Program Files\dotnet\sdk]
```

**Additional context**
This is the actual code I am trying to generate documentation for.
[...]

PR #9348, opened 2023-10-27
chore: Add additional globalization settings

filzrev:

- Add solution-wide `<PredefinedCulturesOnly>false</PredefinedCulturesOnly>` settings.
  - This settings allow to create `new CultureInfo("custom-lang");` when running under `InvariantGlobalization` mode.  
  - This settings is required to use `resx-based localization` on custom plugins DLL
- Add `<SatelliteResourceLanguages>en</SatelliteResourceLanguages>` settings for test projects. to reduce disk space consumption.  
  - This property is set already for `docfx` project (#8813).

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -5,8 +5,12 @@
     <TargetFrameworks Condition=" '$(ContinuousIntegrationBuild)' == 'true' ">$(TargetFrameworks);net8.0</TargetFrameworks>
     <LangVersion>Preview</LangVersion>
     <ImplicitUsings>enable</ImplicitUsings>
-    <InvariantGlobalization>true</InvariantGlobalization>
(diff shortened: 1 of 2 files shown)


yufeih:
Will this fix #9203?

filzrev:
> Will this fix `#9203`?

It seems build error is resolved by this PR.

But it should be noted that there are some behavior differences. 
Because SourceGenerator(Analyzer) is running with created custom culture.
  - Created culture is based on `InvariantCulture`.
  - Created culture don't have `Parent` culture  
    (When creating `en-US` culture. `Parent` property is null by default. And can't customize `GeneratedRegexAttribute`)

yufeih (approved):

yufeih:
Great!

PR #9369, opened 2023-11-02
chore: Add custom file copy logics for template files

### LINKED ISSUE #9183: [Feature Request] Support `Copy if newer` feature to copy resource files to destination
Currently resource files and template files are overwritten every time when executing `docfx build`.
It might be better to Introduce a `Copy if newer` feature like MSBuild.
It  comparing file's `LastWriteTimeUtc` and `File Length` then copy file when diff found.

- Replace file copy operations with `File.Copy` to preserve `LastWriteTimeUtc`.
   (It need to allow some time diffs(about 1 seconds?). to support `output to NAS or synced with OneDrive` scenarios)
- When overwriting resource files. Try to execute `Copy if newer` check.

**Additional context**
Related issue: #8964.
This issue is targeting `Resource files` and `Template's resources`.

filzrev:
This PR intended to resolve #9183 issue.

When running `docfx build` command with multiple times.
There are unnecessaries operations that overwrite identical files.

This PR add features to skip this operations when template files is not changed.
It's required when using something like `dotnet watch` scenarios.

- Add custom template file copy logics for FileStream.
  - Skip file copy operation when following conditions met.
    1. Output file is already exists.
    2. Files length are same.
    3. Files LastWriteTime is same.
  - If above condition is not met. Use `File::Copy` API to preserve original `LastWriteTime`.  
- Add `PreallocationSize` settings for `FileStream` 

**Benchmark Results**
I've tested on `samples/seed` project and measuring `TemplateProcessor::CopyTemplateResources` elapsed times.

1. When it can skip file copy. it's running 10x faster, (e.g.  `86.7` [ms] -> `8.0` [ms])
2. When it needs to `copy new file` and `overwrite files`.  
    It's running slightly slower about 10% (e.g. `86.7` [ms] -> `96.4` [ms])
    I thought it's because of following reasons
     - `File.Copy`  API requires additional operations (e.g. file's attribute copy)
     - It need to open file twice (Open FileStream. then call `File::Copy`)

### DIFF
--- a/src/Docfx.Build/TemplateProcessors/TemplateProcessor.cs
+++ b/src/Docfx.Build/TemplateProcessors/TemplateProcessor.cs
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Collections.Concurrent;
-
+using System.Diagnostics;
 using Docfx.Common;
 using Docfx.Plugins;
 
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Look good. Thank you @filzrev !

Issue #9063, opened 2023-08-10
[Feature Request] Docker Image


JM2K69:
When a release was publish it would be nice to have a docker image for docker available.

JM2K69:
I make some work on the DockerFile

```dockerfile
FROM mcr.microsoft.com/dotnet/sdk:7.0-bullseye-slim

# Add dotnet tools to path.
ENV PATH="${PATH}:/root/.dotnet/tools"

# Install DocFX as a dotnet tool.
RUN dotnet tool update -g docfx && \
    docfx --version

# Install wkhtmltopdf, used by DocFX to generate PDF files.
ARG TARGETARCH
COPY wkhtmltox/wkhtmltox_0.12.6.1-2.bullseye_${TARGETARCH:-amd64}.deb /tmp/wkhtmltox.deb
RUN apt-get update -qq && \
    export DEBIAN_FRONTEND=noninteractive && \
    apt-get install -y -qq --no-install-recommends /tmp/wkhtmltox.deb && \
    rm -rf /var/lib/apt/lists/* /tmp/* && \
    wkhtmltopdf --version

WORKDIR /opt/prj
VOLUME [ "/opt/prj" ]

ENTRYPOINT [ "docfx" ]
```

yufeih:
Github container registry does not require authentication for public container images - a good place to push the container images in absent of a docker organization account.

JM2K69:
With this dockerfile
```
FROM mcr.microsoft.com/dotnet/sdk:7.0-bullseye-slim

# Add dotnet tools to path.
ENV PATH="${PATH}:/root/.dotnet/tools"

# Install DocFX as a dotnet tool.
RUN dotnet tool update -g docfx && \
    docfx --version

# Install dependences for chromium PDF.
RUN apt-get update -qq && apt-get install -y libglib2.0-0 libnss3 libnspr4 \
    libatk1.0-0 libatk-bridge2.0-0 libcups2 libdrm2 libdbus-1-3 libxcb1 \
    libxkbcommon0 libatspi2.0-0 libx11-6 libxcomposite1 \ 
    libxdamage1 libxext6 libxfixes3 libxrandr2 libgbm1 \
    libpango-1.0-0 libcairo2 libasound2

WORKDIR /opt/prj
VOLUME [ "/opt/prj" ]

ENTRYPOINT [ "docfx" ]
```
during a release we need to build the docker and pushi to the github registry container ?

PR #9391, opened 2023-11-07
Add Dockerfile for the project Docfx

JM2K69:
Ad Dockerfile for the docfx projet.

#9063


yufeih:
Arr, I might have just force pushed main. Would you mind recreate the PR?

PR #9394, opened 2023-11-07
Fix `resource` configuration property

MihkelAuv:
Docfx `init` command produces `docfx.json` with `resources` property name. However correct property name for resource files to include in the project output directory is `resource` (singular noun).

This PR fixes `docfx.json` produced by `init` command and corrects the property name in [Config Reference](https://dotnet.github.io/docfx/reference/docfx-json-reference.html#resource) page


MihkelAuv:
@dotnet-policy-service agree

yufeih (approved):
Thank you @MihkelAuv !

PR #9430, opened 2023-11-15
Fix the github actions not working properly in the newest versions of docfx with default settings

ysftulek:
I have updated the actions script because old version was not working properly in the newest versions of DocFX. It now uses official upload pages actions instead of relying on the third party one. And now the default path is correct for website deployment: 'docfx_project/_site' instead of 'docs/_site'.


ysftulek:
@dotnet-policy-service agree

yufeih (approved):
Thank you @ysftulek !

PR #9493, opened 2023-11-30
FIX: Docset Pdf call terminating early

dpvreony:
This fixes an issue where Docset.Pdf was dropping out before the PDF generation was actually running because the action wasn't blocking on the method call.


yufeih (approved):
Thank you @dpvreony !

PR #9513, opened 2023-12-07
feat: Add solution filter file (.slnf) supports

filzrev:
- Add [Solution Filter](https://learn.microsoft.com/en-us/visualstudio/msbuild/solution-filters) (`.slnf`) file supports.
- Fix log message for `unsupported file`.
- Update related documents.


yufeih (approved):
Thank you @filzrev !

PR #9574, opened 2024-01-02
Added extensions configuration point and PlantUml support

### LINKED ISSUE #2347: Support for PlantUML in Markdown
**DocFX Version Used**: 2.29.1
**Template used**: default (`default` or `statictoc` or contain custom template)

In [GitLab it is possible to add PlantUML](https://docs.gitlab.com/ee/administration/integration/plantuml.html) Code to markdown files with: 

\```plantuml
Bob -> Alice : hello
Alice -> Bob : Go Away
\```

The result looks as in the following image:

image

For us it would be a big benefit to remove all our commited binaries (jpgs exported from Enterprise Architect) in our repositories.

For docfx it would mean that it needs some kind of pre processor which posts the data to a public or self hosted [PlantUML Server](https://github.com/plantuml/plantuml-server), save the image to a defined folder and replace markdown code with the resulting image. 

Gitlab posts the image to the PlantUML Server and uses the image link in the markdown file which is not a good option for offline documentation. 

Is some kind of feature planned or what would be the best way to implement it?

### LINKED ISSUE #8763: [Bug] Packages DocFx.Plugins.PlantUml
It seems that the latest version of docfx is not executing the Packages DocFx.Plugins.PlantUml any more.
With Version 1.52.x the PlantUML charts are rendered correct.
With the latest version it seems to stop working.

Did any body faced similar issue?

### LINKED ISSUE #9566: [Feature Request] Add PlantUML support to enable migrations onto latest version of docfx
We have been using plantuml and the [docfx plantuml plugin](https://github.com/KevReed/DocFx.Plugins.PlantUml) to create of our technical documentation.  The latest versions of docfx does not support plantuml with support being limited to mermaid diagrams. 

To continue using the latest versions of docfx we need plantuml support.  We are willing to write a plugin that supports the feature set, but we are unclear as to where to start, what extensibility points to use etc.  Can somebody please provide us with the relevant hooks/mechanisms to implement/override to enable this?  It seens like we need to create a Markdig diagrams extension?  The old plugin is based on the DFM syntax which I believe is not supported going forward.  Any pointers will be much appreciated.

cjlotz:
As discussed in #9566 , this pull request aims to add support for rendering PlantUml diagrams using the latest docfx toolchain.  This will unblock all existing users who want to continue using docfx and PlantUml.  

[This comment](https://github.com/dotnet/docfx/issues/9566#issuecomment-1872550779) covers the approach taken to implement the `PlantUmlExtension`.  Some basic tests were added for the extension points touched.  [As mentioned](https://github.com/dotnet/docfx/issues/9566#issuecomment-1873252393) some tests in the `QuoteSectionNoteTest` are failing which seems to be related to CR-LF return issues which I am unable to resolve.   Help for getting this resolved will be much appreciated.

fixes #2347 fixes #8763 fixes #9566


cjlotz:
@dotnet-policy-service agree [company="Pragma"]

cjlotz:
@dotnet-policy-service agree

src/Docfx.MarkdigEngine.Extensions/PlantUml/FormatterFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+    {
+        this.settings = settings;
+    }
+
+    internal IOutputFormatter CreateOutputFormatter()
+    {
+        switch (OutputFormat)
+        {
+            case OutputFormat.Svg:
+                return new SvgOutputFormatter();
```
yufeih, line 21: nit: seems like an overkill to extract the `IOutputFormatter` interface, could this be turned into a simple function in `CustomCodeBlockRenderer`?

src/Docfx.Plugins/MarkdownServiceParameters.cs
```diff
@@ -35,6 +35,13 @@ public class MarkdownServiceProperties
     [JsonProperty("alerts")]
     [JsonPropertyName("alerts")]
     public Dictionary<string, string> Alerts { get; set; }
+
+    /// <summary>
+    /// PlantUml extension configuration parameters
+    /// </summary>
+    [JsonProperty("plantUml")]
+    [JsonPropertyName("plantUml")]
+    public Dictionary<string, string> PlantUml { get; set; }
```
yufeih, line 44: Can this be turned into a strongly-typed object so we know what properties to expect for PlantUml? The `alerts` is a string dictionary because alerts maps from name to css class.

src/Docfx.MarkdigEngine.Extensions/MarkdownContext.cs
```diff
@@ -51,6 +57,11 @@ public class MarkdownContext
     /// </summary>
     public GetImageLinkDelegate GetImageLink { get; }
 
+    /// <summary>
+    /// Get the configuration for a given extension
+    /// </summary>
+    public GetExtensionConfigurationDelegate GetExtensionConfiguration { get; }
```
yufeih, line 63: Consider keep the existing pattern of passing plant UML options to `UsePlantUml` and `UseDocfxExtensions` as additional arguments. `IReadOnlyDictionary<string, string>` isn't a good abstraction for extension configurations - we may in the future have extension configs that are more than string -> string maps.

test/Docfx.MarkdigEngine.Extensions.Tests/QuoteSectionNoteTest.cs
```diff
@@ -275,12 +303,12 @@ public void TestDfmVideo_MixWithNote()
 
     > code text 1-1
     > [!NOTE]
-    > code text 1-2  
+    > code text 1-2
```
yufeih, line 306: don't remove these trailing whitespaces, they represent softbreaks in markdown.

src/Docfx.Plugins/MarkdownServiceParameters.cs
cjlotz: I would also prefer going this route, but the problem is that the configuration is owned by the extension and there is no reference to the extension from within the Plugins assembly, so there is no opportunity to use a strongly typed object.

src/Docfx.MarkdigEngine.Extensions/MarkdownContext.cs
cjlotz: For me the extension owns/understands its configuration.  The consumer of the extension should not be burdened by knowing how to configure the extension.  This is internal knowledge to the extension.  The extension will pull the relevant configuration using the delegate that provides a map (key/value pair) of configuration parameters from the `MarkdownContext`.  The consumer just passes the map on and the extension takes care of the rest

test/Docfx.MarkdigEngine.Extensions.Tests/QuoteSectionNoteTest.cs
cjlotz: Mmm, i tried doing that but somehow my editor keeps on removing it.  Will try again

src/Docfx.MarkdigEngine.Extensions/MarkdownContext.cs
yufeih: I mean `IReadOnlyDictionary<string, string>` cannot represent the config if it uses `bool`, `number`, `array` or nested objects such as:

```json
{
  "my-extension": {
    "showTable": false,
    "tableOptions": {
      "displayAs": "table"
    }
  }
}
```

cjlotz: Are you proposing `Dictionary<string, object>` rather?

cjlotz: I've changed the delegate signature to return an `object` 

```c#
    /// <summary>
    /// Allows configuration of extensions
    /// </summary>
    /// <param name="extension">Name of the extension being configured</param>
    public delegate object GetExtensionConfigurationDelegate(string extension);
```

The extension can then pull it using code such as:

```c#
    public PlantUmlExtension(MarkdownContext context)
    {
        _context = context;
        _settings = new();

        if (_context.GetExtensionConfiguration("PlantUml") is Dictionary<string, string> config)
            _settings = new DocfxPlantUmlSettings(config);
    }
```

This will enable any .NET intrinsic type to be used.

test/Docfx.MarkdigEngine.Extensions.Tests/QuoteSectionNoteTest.cs
cjlotz: This is now fixed

yufeih (approved):
Thank you @cjlotz . This is a great addition and addresses one of the most asked features #2347

cjlotz:
> Thank you @cjlotz . This is a great addition and addresses one of the most asked features #2347

@yufeih  Glad to assist.  I will be away on some leave soon, so let me know if there is anything left to do to enable you to merge the pull request.  Looking forward to using the latest version of docfx with this included.

test/docfx.Tests/JsonConverterTest.cs
```diff
@@ -49,6 +50,51 @@ public void TestJObjectDictionaryToObjectDictionaryConverterSerializeAndDeserial
...
+                                  "REVIEW": "alert alert-primary"
+                                },
+                                "plantUml": {
+                                  "outputFormat": "svg",
+                                  "remoteUrl": "https://www.plantuml.com/plantuml"
+                                }
+                              },
+                              "disableGitFeatures": true
+                            }
+                            """;
```
filzrev, line 72: It needs to add `.ReplaceLineEndings();` to compare platform invariant string assertion.
(Because current `.cs` source code use `CRLF` line separator)

PR #9575, opened 2024-01-03
chore: Fix picture image click event handling

### LINKED ISSUE #9571: [Bug] Images that use HTML `<picture>` tag is not works in some condition
**Describe the bug**
When using `<picture>` tag to display images.
It's not displayed as expected when following condition met.
Following condition

- Display image using `<picture>` element conditionally.
- Image file width/height is over 200px.

**To Reproduce**
1. Add following HTML content to markdown document. 
```html
<picture>
  <source srcset="images/example_dark.png" media="(max-width: 600px)">
  <img src="images/example.png">
</picture>
```
2. And image files to `images` directory.
3. Build content.
4. Display rendered HTML on docfx.

**Expected behavior**
When browser window resized.
Image should be switched automatically based on `media selector` condition.

But when hosting HTML on docfx.
Image is not changed (It seems temporary changed. But immediately reverted to original `src`)

- Browser: [e.g. chrome, edge]
- Form factor: [e.g. desktop]
- Docfx version: [e.g. 2.74.1]
- Template: [e.g. default, modern]

**Additional context**
It seems above behaviors are caused by `makeClickable` script.
https://github.com/dotnet/docfx/blob/main/templates/modern/src/markdown.ts#L97-L132

This script wrap `` tag with `<a>` anchor tag.
But it seems [`<picture>` tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture) is not expected anchor child element.

It need to be fixed with of following workaround.

1. Disable `makeClickable` function for `picture` based images.   
    (e,g, Add `if(img.parentElement.tagName=='PICTURE'){return false}`)
or
2. Wrap entire `picture` element with anchor.   
    And replace `src` to [HTMLMediaElement::currentSrc](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentSrc)

filzrev:
Fix #9571

**What's changed this PR**
Add `<picture>` tag special handling to `makeClickable()` function.

`<picture>` tag's `img.currentSrc` is changed dynamically based on context.
So it need to add `onClick` event handler to picture tag itself.

## Test Code

```html
<div>
<picture>
    <source srcset="images/example_dark.png" media="(max-width: 600px)" alt=""/>
    <img src="images/example.png" alt=""/>
</picture>
</div>
```
## What's tested manually
1. When window size is over 600px. `images/example.png` is displayed. and this image opened when clicking.
2. When window size is <= 600 px. `images/example_dark.png"` is displayed. and this image opened when clicking.
3. Other `` tags  behavior is not changed (wrapped by anchor tag)

### DIFF
--- a/templates/modern/src/markdown.ts
+++ b/templates/modern/src/markdown.ts
@@ -114,9 +114,18 @@ async function renderClickableImage() {
       const a = document.createElement('a')
       a.target = '_blank'
       a.rel = 'noopener noreferrer nofollow'
-      a.href = img.src
-      img.replaceWith(a)
-      a.appendChild(img)
+
+      if (img.parentElement.tagName === 'PICTURE') {
+        const picture = img.parentElement
+        picture.addEventListener('click', () => {
+          a.href = img.currentSrc
+          a.click()
+        })
+      } else {
+        a.href = img.src
+        img.replaceWith(a)
+        a.appendChild(img)
+      }
     }
 
     function shouldMakeClickable(): boolean {


yufeih (approved):
Thank you @filzrev !

PR #9587, opened 2024-01-06
build(deps): update nuget dependencies

filzrev:
This PR manually update following package dependencies.

- `Microsoft.NET.Test.Sdk`
- `xunit`
- `xunit.runner.visualstudio`

Note: I don't know why dependabot don't create PR automatically for xunit packages.
(It seems worked before [November 20](https://github.com/dotnet/docfx/pull/9442))

---
To resolve following warning.
I've modified `ModelAttributeHandlerTest.cs`.

> Warning	xUnit2027	Comparing an instance of List<string> with an instance of HashSet<string> has undefined results, because the order of items in the set is not predictable. Create a stable order for the set (i.e., by using OrderBy from Linq).


yufeih (approved):
Thank you @filzrev !

Issue #9619, opened 2024-01-17
[Bug] Using Roslyn Source Generators with attribute class fails


paulushub:
**Describe the bug**
Trying to document a source code which has attribute classes generated by the source generators.
- Without generating attribute classes, `docfx` works without any error. 
- With attribute class, the build fails with error below
> MyClass.cs(11,6): error CS0246: The type or namespace name 'ElementFactoryAttribute' could not be 
> found (are you missing a using directive or an assembly reference?)

**To Reproduce**
1. Create any sample Generator project and have it generate an attribute class; say `ElementFactoryAttribbute`
2. Create a project to use the generated classes and use the attribute on a class.
3. Create a docfx project and serve.

**Expected behavior**
The build to work with a generated attribute class.

**Context:**
- OS: Windows Desktop
- Docfx version: 2.75.1

**Additional context**
```cs
namespace HelloWorldGenerated
{
    /// <summary>
    /// Enter description here for class X.
    /// ID string generated is "T:MyNamespace.MyClass".
    /// </summary>
    [ElementFactory]
    public class MyClass
    {
        /// <summary>
        /// Enter description here for the first constructor.
        /// ID string generated is "M:MyNamespace.MyClass.#ctor".
        /// </summary>
        public MyClass() { }

        /// <summary>
        /// Enter description here for the second constructor.
        /// ID string generated is "M:MyNamespace.MyClass.#ctor(System.Int32)".
        /// </summary>
        /// <param name="i">Describe parameter.</param>
        public MyClass(int i) { }

        /// <summary>
        /// Enter description here for field Message.
        /// ID string generated is "F:MyNamespace.MyClass.Message".
        /// </summary>
        public string Message;
     }
}

filzrev:
On my environment.
It can be resolved by executing command with following orders.

1. `dotnet build -c Release`
2. `docfx metadata`

**Background**
`docfx metadata` command build projects by using  `MSBuildWorkspace` with `Release` configuration.
But it doesn't generate physical DLL files.

Then project's analyzer is resolved to `UnresolvedAnalyzerReference`.
And it's silently skipped.

paulushub:
@filzrev Thanks for the help.
The following did resolve the issue as you said (without `dotnet build -c Release`, the `docfx metadata` failed): 
1. dotnet build -c Release
2. docfx metadata
3. docfx docfx.json

If there is no plan to directly support this use case, then I will close the issue.

PR #9623, opened 2024-01-18
chore: Add warnings for UnresolvedAnalyzerReference

filzrev:
Add warning logs for `UnresolvedAnalyzerReference`.
(That occurs when analyzer DLL can't resolved when running `dotnet metadata` command.
See: https://github.com/dotnet/docfx/issues/9619

**Test**
I've tested PR manually with solution that contains SourceGenerator project.

**Warning message examples**
> warning: There is .NET Analyzer that can't be resolved. Path: C:\Projects\SampleApp\src\SampleSourceGenerator\bin\Release\netstandard2.0\SampleSourceGenerator.dll

### DIFF
--- a/src/Docfx.Dotnet/DotnetApiCatalog.Compile.cs
+++ b/src/Docfx.Dotnet/DotnetApiCatalog.Compile.cs
@@ -7,6 +7,7 @@ using Microsoft.Build.Construction;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Logging;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.MSBuild;
 
 #nullable enable
@@ -132,6 +133,11 @@ partial class DotnetApiCatalog
                     await Process.Start("dotnet", $"restore \"{path}\"").WaitForExitAsync();
                 }
                 project = await workspace.OpenProjectAsync(path, msbuildLogger);
+
+                foreach (var unresolvedAnalyzer in project.AnalyzerReferences.OfType<UnresolvedAnalyzerReference>())
+                {
+                    Logger.LogWarning("There is .NET Analyzer that can't be resolved. Path: " + unresolvedAnalyzer.FullPath);
+                }
             }
 
             if (!project.SupportsCompilation)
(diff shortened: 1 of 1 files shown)


paulushub:
@filzrev Is it possible to detect such case and automatically fix it (re-running)?

src/Docfx.Dotnet/DotnetApiCatalog.Compile.cs
```diff
@@ -132,6 +133,11 @@ await LoadCompilationFromProject(project.AbsolutePath) is { } compilation)
                     await Process.Start("dotnet", $"restore \"{path}\"").WaitForExitAsync();
                 }
                 project = await workspace.OpenProjectAsync(path, msbuildLogger);
+
+                foreach (var unresolvedAnalyzer in project.AnalyzerReferences.OfType<UnresolvedAnalyzerReference>())
+                {
+                    Logger.LogWarning("There is .NET Analyzer that can't be resolved. Path: " + unresolvedAnalyzer.FullPath);
```
yufeih, line 139: shall we hint user to try explicitly run a dotnet build before running docfx?

filzrev: > shall we hint user to try explicitly run a dotnet build before running docfx?

How about the following text.
> There is .NET Analyzer that can't be resolved. If this analyzer DLL is .NET Source Generator project. Try build with `dotnet build -c Release` before running docfx command.  Path: ";

paulushub: OR if you have a Q&A section in the documentation, it can be referenced. 
Since you are running the `docfx metadata`, I assume you also already have information on how to run `dotnet build` on multi-projects defined in the `docfx.json`.

yufeih: @filzrev Looks good.

@paulushub
> Since you are running the docfx metadata, I assume you also already have information on how to run dotnet build on multi-projects defined in the docfx.json.
It might be problematic to load a project in memory to detect if `dotnet build` is needed, run `dotnet build` and then reload the project to memory. It seems a hint is good enough for now. We can revisit if this issue become more widely spread.

yufeih (approved):

PR #9634, opened 2024-01-22
feat: Enforce `GlobMatcherOptions.AllowDotMatch` option for file's `excludePatterns`

### LINKED ISSUE #8913: [Bug] "exclude" is not works when path fragments contains folder that starts with '.'
**Describe the bug**

"exclude" is ignored when path fragments contains names that starts with `.` 

**Example**
`_site/.attachments/file.txt` should be excluded. but docfx process as resource.

``` json
    "resource": [
      {
        "files": [
          "**/.attachments/*.txt" // Azure DevOps wiki style attachments folder name
        ],
        "exclude": [
          "_site/**"
        ]
      }
    ],
```

**To Reproduce**
1. Clone sample test project (https://github.com/filzrev/docfx_glob_tests)
2. Run `docfx build`.  `articles\.attachments\file.txt` are handled as resource.   
    and copied to `_site\articles\.attachments/file.txt` is copied.
4. Run `docfx build`.  `_site\articles\.attachments/file.txt`  are handled as resource.   
    and copied to `_site\_site\articles\.attachments/file.txt` . 

**Expected behavior**
`"exclude": [ "_site/**" ] `  glob settings works as expected.

- OS: Windows
- Docfx version: 2.67.5

**Additional context**

It seems `Microsoft.DocAsCode.Glob` handle folders that starts with '.' as special `hidden folder`.

https://github.com/dotnet/docfx/blob/b03e06e6631c8dacfe8df1780e3a3b9081619713/src/Microsoft.DocAsCode.Glob/GlobMatcher.cs#L406-L416
https://github.com/dotnet/docfx/blob/971df085ad24ba26b3c48371631841b8240e3db8/test/Microsoft.DocAsCode.Glob.Tests/GlobFileTest.cs#L19-L50

If there are no special requirements for file globbing rules.
it might be better to consider to replace globbing library to `Microsoft.Extensions.FileSystemGlobbing`.

filzrev:
This PR intended to fix #8913 issue.

By default. docfx exclude directories/files that name starts with `dot` char .
And there are two options available to include these files.

- **Option 1:** Explicitly specify path that contains `dot` char.
- **Option 2:** Set [FileMappings `dot` property](https://github.com/dotnet/docfx/blob/ad6db161e426413aeea52f3cd42c65a7e60e99e7/src/Docfx.Common/FileMappingItem.cs#L107-L113) to `true`

But when using **Option1**.
There is problem that can't exclude some paths by using `exclude` settings.
Because  default `AllowDotMatch:false` settings is also applied to `exclude` filter rules.

This PR enforce `AllowDotMatch` options for `exclude` filter rules.

### DIFF
--- a/src/Docfx.Glob/FileGlob.cs
+++ b/src/Docfx.Glob/FileGlob.cs
@@ -19,7 +19,7 @@ public class FileGlob
         var globArray = patterns.Select(s => new GlobMatcher(s, options)).ToArray();
         var excludeGlobArray = excludePatterns == null ?
             Array.Empty<GlobMatcher>() :
-            excludePatterns.Select(s => new GlobMatcher(s, options)).ToArray();
+            excludePatterns.Select(s => new GlobMatcher(s, options | GlobMatcherOptions.AllowDotMatch)).ToArray();
         return GetFilesCore(cwd, globArray, excludeGlobArray);
     }
 
--- a/test/Docfx.Glob.Tests/GlobFileTest.cs
+++ b/test/Docfx.Glob.Tests/GlobFileTest.cs
@@ -2,6 +2,8 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using Docfx.Tests.Common;
+using FluentAssertions;
+using FluentAssertions.Execution;
 using Xunit;
 
 namespace Docfx.Glob.Tests;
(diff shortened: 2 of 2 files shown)


yufeih:
I'm leaning towards gradually retire the various glob config options (`dot`, `noExpand`, `noNegate`, `noGlobStar`, `case`) and move to a standard glob behavior, made it easier to switch the underlying glob implementation in the future.

This PR seems to go against that direction by introducing different behaviors for `include` vs `exclude`. 

It seems TypeScript also does not glob dot files: https://github.com/microsoft/TypeScript/issues/49555

filzrev:
> I'm leaning towards gradually retire the various glob config options (dot, noExpand, noNegate, noGlobStar, case) and move to a standard glob behavior

I agree with that direction.

I takes approach to not including that starts with root wildcard.
(e.g. Use `articles/**/..attachments/**` instead of `**/.attachments/**`)

PR #9637, opened 2024-01-23
pref: Optimize xrefmap download performance

### LINKED ISSUE #9558: [Feature Request] Download `xrefmap.json` file on `docfx build` is significantly slower
I'm trying to use `xref:` links for .NET types inside **markdown document**.
And add `xref` settings with following instruction.
https://dotnet.github.io/docfx/docs/links-and-cross-references.html?q=xref#cross-reference-to-net-basic-class-library

But when running `docfx build` command.
It takes too long time to download `xrefmap` file. (about 10 minutes on my environment. And It's disconnected from server sometimes)
When downloading same file with browser. It's completed about `a few seconds`. 

It seems there are some problems in [XRefMapDownloader](https://github.com/dotnet/docfx/blob/27da5531b3b0fed9253fce66c4458e05d085ff3e/src/Docfx.Build/XRefMaps/XRefMapDownloader.cs#L128-L135)

**1. File size of `xrefmap.json`**
 `xrefmap.json` file size is about `316MB`. And this file is downloaded everytime when running `dotnet build` command.
  - I thought. it's better to set `AutomaticDecompression = System.Net.DecompressionMethods.All` for `HttpClient` handler to enable HTTP response compression. (When compressed with gzip. File size is shrieked to about16MB)
  -  It might be better to cache downloaded `xrefmap` file locally with `ETag`.  
     And use `ETag` to skip download when server-side document is not updated.

**2. `StreamReader`'s default buffer size(1KB) is too small to synchronously processing stream**
Yaml deserialization is executed synchronously. So it frequently blocking file download process. 
As far as I've confirmed. It's significantly faster when increasing buffer size (1MB). 

Download `xrefmap.json`file manually (or `docfx download`) before running `docfx build`.
And specify local `xrefmap.zip` file in `docfx.json` file.

**Additional context**
~~There is `docfx download` command exists.~~
~~But it's not documented and It seems not workings (There is no way to pass download URL from command line parameter?)~~

`docfx download -x https://learn.microsoft.com/en-us/dotnet/.xrefmap.json xrefmap.zip`

filzrev:
This PR is intended to resolve #9558

- Add settings to use HTTP Compression (if available)
- Change HttpClient's timeout to 30 minutes (Default: 100 seconds)
- Change StreamReader's buffer size to 80KB (Default: 1024 byte)
  -  `80KB` is default buffer size that used by `Stream::CopyTo`. 

**What's tested**

I've manually tested xrefmap download with following configuration.
```
  "build": {
    "xref": [
      "https://learn.microsoft.com/en-us/dotnet/.xrefmap.json"
    ]
  }
```

And confirmed server returns following  HTTP Response headers. And download size is reduced. (`316MB` -> `16MB`)

```
HTTP/1.1 200 OK
Content-Type: application/json
Content-Length: 15978441
Content-Encoding: gzip
```

**TODO Tasks**
I'll create another PR later relating #9558.

- Add special deserialization code path for JSON file.
   (Because YamlDotNet based JsonDeserializer is 5x slower than `System.Text.Json` based deserialization)
- Add support  for loading `xrefmap compressed by gzip` (`xrefmap.json.gz`) from local disk.

### DIFF
--- a/src/Docfx.Build/XRefMaps/XRefMapDownloader.cs
+++ b/src/Docfx.Build/XRefMaps/XRefMapDownloader.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Net;
 using Docfx.Common;
 
 using EnvironmentVariables = Docfx.DataContracts.Common.Constants.EnvironmentVariables;
@@ -127,11 +128,15 @@ public class XRefMapDownloader
 
         using var httpClient = new HttpClient(new HttpClientHandler()
         {
+            AutomaticDecompression = DecompressionMethods.All,
             CheckCertificateRevocationList = !EnvironmentVariables.NoCheckCertificateRevocationList,
-        });
+        })
+        {
+            Timeout = TimeSpan.FromMinutes(30), // Default: 100 seconds
+        };
 
         using var stream = await httpClient.GetStreamAsync(uri);
-        using var sr = new StreamReader(stream);
+        using var sr = new StreamReader(stream, bufferSize: 81920); // Default :1024 byte
         var map = YamlUtility.Deserialize<XRefMap>(sr);
         map.BaseUrl = baseUrl;
         UpdateHref(map, null);


yufeih (approved):
Looks great! Thank you @filzrev

PR #9638, opened 2024-01-23
fix: When `disableDefaultFilter` is enabled. `filterConfig.yml' is not loaded.

filzrev:
This PR fixes the issue that it can't use `"disableDefaultFilter": true' settings with custom `filterConfig.yml'.

I've tested this PR manually with following configurations.

```
  "metadata": [
    {
      "filter": "filterConfig.yml",
      "disableDefaultFilter": true
    }
  ],
```


yufeih (approved):
Thank you @filzrev !

PR #9648, opened 2024-01-24
fix: Corrected class name in inheritedMembers block

BattlefieldDuck:
## Description
This pull request addresses a bug found in the `inheritedMembers` block of the code. The class name was previously set as `derived`, which was incorrect. This PR updates the class name to `inheritedMembers`.

## Changes
- Updated the class name from `derived` to `inheritedMembers` in the `inheritedMembers` block.


yufeih (approved):
Thank you @BattlefieldDuck

Issue #9618, opened 2024-01-17
[Bug] `docfx metadata` does not generate all metadata on Ubuntu


AristurtleDev:
**Describe the bug**
We are using DocFX to generate markdown API documentation based on two projects.  The generation is run through a GitHub Action running on `ubuntu-latest`.  It's been discovered that when running on Ubuntu, the `dotnet docfx metadata` command will skip all except the last glob listed in the `metadata.src.files` array.

For example, here is the **docfx.json** configuration file.

```json
{
    "metadata": [
      {
        "outputFormat": "markdown",
        "src": [
          {
            "src": ".",
            "files": [
              "external/MonoGame/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj",
              "external/MonoGame/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj"
            ]
          }
        ],
        "dest": "./content/api",
        "includePrivateMembers": false,
        "disableGitFeatures": false,
        "disableDefaultFilter": false,
        "noRestore": false,
        "namespaceLayout": "flattened",
        "memberLayout": "samePage",
        "EnumSortOrder": "alphabetic"
      }
    ]
}
```

With this configuration file, when run on `windows-latest` and `macOS-latest` GitHub runners, the following is output 

```
dotnet docfx metadata
Using .NET Core SDK 8.0.100
Loading project /Users/aristurtle/Desktop/monogame-11ty/external/MonoGame/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj
  Determining projects to restore...
  Restored /Users/aristurtle/Desktop/monogame-11ty/external/MonoGame/MonoGame.Framework/MonoGame.Framework.DesktopGL.csproj (in 1.08 sec).
Loading project /Users/aristurtle/Desktop/monogame-11ty/external/MonoGame/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj
  Determining projects to restore...
  Restored /Users/aristurtle/Desktop/monogame-11ty/external/MonoGame/MonoGame.Framework.Content.Pipeline/MonoGame.Framework.Content.Pipeline.csproj (in 443 ms).
  1 of 2 projects are up-to-date for restore.

...

Export succeed: 460 items
```

However on `ubunutu-latest` the following is output

```
dotnet docfx metadata
Using .NET Core SDK 8.0.100
[...]

PR #9650, opened 2024-01-25
fix: ApiPage is not generated when multiple namespaces defined across projects

filzrev:
This PR is intended to fix https://github.com/dotnet/docfx/issues/9618.

On current implementation.
`CreateNamespaceToc` method don't call `yield return node` if `node.containsLeafNodes` is `false`.
So if empty namespace is processed first. other items that share same namespaces are not included as docfx output.

This PR change logics to call `yield return node` with following 2 condition. 
1. `TocNode` is created & items are added.
2. `TocNode` is exists but has no items & items are added.

### DIFF
--- a/src/Docfx.Dotnet/DotnetApiCatalog.Toc.cs
+++ b/src/Docfx.Dotnet/DotnetApiCatalog.Toc.cs
@@ -102,6 +102,8 @@ partial class DotnetApiCatalog
                     });
                 }
 
+                var existingNodeHasNoLeafNode = idExists && !node.containsLeafNodes;
+
                 node.items ??= new();
                 node.symbols.Add((symbol, compilation));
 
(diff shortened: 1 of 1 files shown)


src/Docfx.Dotnet/DotnetApiCatalog.Toc.cs
```diff
@@ -120,9 +122,12 @@ IEnumerable<TocNode> CreateNamespaceToc(INamespaceSymbol ns)
                 }
 
                 node.containsLeafNodes = node.items.Any(i => i.containsLeafNodes);
-                if (!idExists && node.containsLeafNodes)
+                if (node.containsLeafNodes)
```
yufeih, line 125: It seems that the root cause is we assume in line 96 if a TOC node exists in `tocNodes` it'll be included in the final. Does the following code fixes the problem?

```suggestion
                if (!node.containsLeafNodes)
                {
                    tocNodes.Remove(node);
                }

                if (!idExists && node.containsLeafNodes)
                {
                    yield return node;
                }
```

yufeih (approved):
Looks good @filzrev 👍

Issue #5233, opened 2019-10-25
Invalid link: '<a href="#...">...</a>'. The file index.md doesn't contain a bookmark named '...'


davikor:
**Operation System**: `Windows`

**DocFX Version Used**: 2.46 

**Template used**: `default`

**Steps to Reproduce**:

1. download extract [attached example](https://github.com/dotnet/docfx/files/3772235/docfx_project.zip)
2. run `docfx.exe docfx.json`

**Expected Behavior**:

Build succeeds without warnings

**Actual Behavior**:

Build succeeds with one warning: Invalid link: `'<a href="#anchor-with-umlaut-%C3%A4">Link to anchor</a>'. The file index.md doesn't contain a bookmark named 'anchor-with-umlaut-%C3%A4'`

DocFx seems to url-encode special characters (e.g. german Umlaute) in links but does not url encode the anchor names that are being generated for each heading, therefore the warning I guess. Firefox seems to be okay with that (it can navigate the links) but of course it is not obvious anymore whether ones link is really invalid or whether it just contains characters that got url encoded.

davikor:
this issue is still valid with 2.48.1

AndreasGloeckner:
We wanted to use markdig as our markdown engine and ran into the same error with it. Everything is fine with

robog3:
Hi,
We have this issue with French characters too (é, è, à, ...) which is a major problem for us even though the redirection works. Our CI/CD pipeline regenerates the documentation website at each commit if there are no warnings (for broken links), which makes it impossible for us to use accents in anchors as long as this issue is not solved.
Docfx v2.56.7.
Thank you!

DavidFeldhoff:
Any news on this? We're also waiting for it.
Using Docfx v2.57.2

NilsGruettner:
Hi,

I am also encountering that problem with german Umlaute.
It is really hard to find actual non-working links. In larger projects I get >50 false warnings.

Docfx v2.58.9.0

patware:
For french characters (é, è, à, ', ...), I tried saving the .md as UTF8 with and without BOM.  Same warnings.

Docfx v2.58.9.0

```
# foo

Visit [Québec](#québec).

## Québec

The province or the city.
```

[...]

PR #9660, opened 2024-01-29
fix: Warnings that occurs when bookmark link contains non-ASCII chars

filzrev:
This PR intended to fix issue #5233
By adding `URI normalization logics` to validate bookmark URLs .

---

**Examples markdown that contains non-ASCI chars**
```markdown
# foo

Visit [Québec](#québec).

## Québec

The province or the city.
```

**Generated output HTML**
```html
<h2 id="foo">foo</h2>
<p>Visit <a href="#qu%C3%A9bec">Québec</a>.</p>
<h2 id="québec">Québec</h2>
<p>The province or the city.</p>
```

**Background**
Markdown link `[Québec](#québec)` is converted to url-encoded anchor link (`#qu%C3%A9bec`)
[This behavior is according to the CommonMark spec](https://github.com/xoofx/markdig/issues/380#issuecomment-547699642).

Instead, `## Québec` heading is converted to `<h2 id="québec">`
Because  it's generated by `AutoIdentifier` markdig extension.
And docfx explicitly specify `AutoIdentifierOptions.GitHub` options.

So bookmark URI validation should be executed by `Normalized URI` comparison.
(In this PR. This normalization is executed by `Uri.UnescapeDataString` API).

It might be better to use API that comply with the following standard specifications
- RFC 3987 - Internationalized Resource Identifiers (IRIs)
- [WHATWG URL Standard](https://url.spec.whatwg.org)

### DIFF
--- a/src/Docfx.Build/PostProcessors/ValidateBookmark.cs
+++ b/src/Docfx.Build/PostProcessors/ValidateBookmark.cs
@@ -43,7 +43,7 @@ sealed class ValidateBookmark : HtmlDocumentHandler
              {
                  Title = node.InnerText,
                  Href = TransformPath(outputFile, decodedLink),
-                 Bookmark = bookmark,
+                 Bookmark = Uri.UnescapeDataString(bookmark),
                  SourceFragment = WebUtility.HtmlDecode(node.GetAttributeValue("data-raw-source", null)),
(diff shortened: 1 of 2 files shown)


yufeih (approved):
Looks good. Thank you @filzrev !

Issue #9649, opened 2024-01-24
[Bug] EmptyTocItemName Warnings When using Xrefs 


groogiam:
**Describe the bug**
As of 2.75.1+6aa697f56975e85e82c5a6b81b71157c77302270 I am getting warnings when I use uid's in my toc.

**To Reproduce**

toc.md
```
# <xref:custom-entity-add-new>

```

page.md

```
---
uid: custom-entity-add-new
---

# Add New Domain Entity

My Content
```
This results in the following warning

`warning: EmptyTocItemName: TOC item (TopicUid:custom-entity-add-new ) with empty name found. Missing a name?`

**Expected behavior**
The referenced file has a header # so I would expect this warning to not be shown as was done with previous versions.

**Context (please complete the following information):**
- OS: Windows
- Docfx version: 2.75.1+6aa697f56975e85e82c5a6b81b71157c77302270

filzrev:
This issue seems be fixed by 9665.

groogiam:
@filzrev Thanks for the update. Is this released in [2.75.3](https://www.nuget.org/packages/docfx/2.75.3)? If not any idea when the next release will be? Thanks.

filzrev:
It seems be included in v2.75.3.
Please let me know if there are any problems.

groogiam:
Everything appears to be working. Thanks for your help!

PR #9665, opened 2024-01-30
fix: Toc warnings when toc item with TopicUid but name is not specified

filzrev:
This PR intended to fix ~~#9664~~ #9649.
By adding lines to skip warnings when `TopicUid` value exists.

Because if `TocItem` contains a `TopicUid` value.
It seems `node.Name` is automatically resolved by `TocDocumentProcessor`.

https://github.com/dotnet/docfx/blob/277fb48c3e2ad7b68e1c23d010a0c6670e9d7f8b/src/Docfx.Build/TableOfContents/TocDocumentProcessor.cs#L105-L118


yufeih (approved):
Looks good. Thank you @filzrev !

Issue #8427, opened 2023-02-20
[Bug] Custom namespace filter rules not working anymore


Krusty93:
**Describe the bug**
I use DocFx to document a specific namespace in my dotnet project.

To achieve my goal I have a `filterConfig.yml` file where I exclude everything but what I need:
```yml
apiRules:
- exclude:
      hasAttribute:
          uid: Newtonsoft.Json.JsonIgnoreAttribute
- exclude:
      uid: .*
      type: Method
- exclude:
      uid: .*
      type: Field
- exclude:
      uid: .*
      type: Interface
- include:
      uidRegex: ^MyApp\.Domain\.Models
- exclude:
      uidRegex: ^MyApp\.Domain
```

But unfortunately, the produced documentation is empty because the last `exclude` rule is removing everything.

I cloned the source code to attach the debugger and I saw that the class `SymbolFilter` (method `IncludeApi`) is correctly validating my rules, but then it checks for the `ContainingSymbol` as well. And being the `Models` a child of the `Domain` namespace, this obviously breaks my rules invalidating the `include` one.

This didn't happen with the v2.61.0 and [according to the documentation](https://dotnet.github.io/docfx/docs/dotnet-api-docs.html#filter-by-uid) this approach shall work.

**To Reproduce**
Create a dotnet  console app `MyApp` and two classes in two different namespaces:
1. MyApp.Domain.Models.Class1
2. MyApp.Domain.Class2

Run `docfx init --quiet`

Fill the `metadata` property of the `docfx.json` file as follow:
```json
  "metadata": [
    {
      "src": [
        {
          "files": [
            "**.csproj"
          ],
          "exclude": [
            "**/bin/**",
            "**/obj/**",
            "_site/**",
          ],
          "src": "../"
        }
      ],
      "dest": "api",
      "filter": "filterConfig.yml",
      "includePrivateMembers": false,
      "disableGitFeatures": false,
      "disableDefaultFilter": false
    }
  ],
```
And create in the same folder a `filterConfig.yml` file:
```yml
apiRules:
- include:
      uidRegex: ^ConsoleApp1\.Domain\.Models
- exclude:
      uidRegex: ^ConsoleApp1\.Domain
```

Run the `serve` command. Generated documentation will be empty

**Expected behavior**
Generated documentation shall contain documentation for the `ConsoleApp1.Domain.Models` namespace (`Class1`).

To have a test, perform the same steps as above but with the 2.61.0 version installed.

**Context:**
[...]

PR #9666, opened 2024-01-30
fix: filterconfig exclude rule is not works as documented

### LINKED ISSUE #9404: API Filtering Yml Cannot Include Specific Namespace
**Describe the bug**
The api filtering yml mechanism does not allow keeping a specific sub namespace as per https://dotnet.github.io/docfx/docs/dotnet-api-docs.html#filter-by-uid.

**To Reproduce**

The following api rules yml does filters out all items in the `MyProject.Server` namespace.

```yml
apiRules:
- include:
    uidRegex: ^MyProject\.Server\.Config
- exclude:
    uidRegex: ^MyProject\.Server
```

**Expected behavior**
Based on the documentation this should generate docs for only MyProject.Server.Config

It seems like the excludes are taking precedence over the includes.

- OS: Windows 10 Pro
- Docfx version: 2.73.1+1d7b96c23465e14eafbcb0dd765029a87b5312d4

filzrev:
This PR intended to fix #9404

Currently `filterconfig.yml`'s `exclude` rules are not works as documented.
https://dotnet.github.io/docfx/docs/dotnet-api-docs.html?q=filterconfig#filter-by-uid

When `exclude` rule is defined without `Type` attribute.
`Namespace`  symbol type is included as exclusion target.
So types defined at this namespace are not included.

> API filter are hierarchical, if a namespace is excluded, all types/members defined in the namespace would also be excluded. Similarly, if a type is excluded, all members defined in the type would also be excluded.

This PR change `exclude` rule's default kind to `ExtendedSymbolKind.Type`.

Don't change current behaviors. Fix documentation to explicitly specify `type: Type` for excude rules.


filzrev:
I've found issue #8427 that relating to this PR.
It seems  documented sample filter rules are working on older version of docfx.

yufeih (approved):
Thank you @filzrev !

PR #9690, opened 2024-02-10
chore: Add fail-fast logics  when `--serve` option enabled & port is already used

filzrev:
**What included in this PR**
Add fail-fast logics when  `--serve` option enabled and specified TCP port is already used.

**Background**
`docfx build` commands takes some times to complete.
When combined with `--serve` options.
Command fails on `RunServe::Exec` phase.
And need to run the `docfx serve' command separately.

This PR add TCP port availability check before start `metadata`/`build` commands.
And fail-fast if port is already used.

**Test**
I've manually confirmed behaviors with following `--hostname` patterns

- `null`/ `localhost` 
  - It's bound to loopback addressed (both `127.0.0.1` and `::1`)
- `*`
  -  It's bound to all available IPv4/IPv6 addressed
- IPAddress: (e.g. `127.0.0.1` / `127.0.0.2`)
  - I’ve confirmed that docfx can successfully host both sites simultaneously.

### DIFF
--- a/src/docfx/Models/BuildCommand.cs
+++ b/src/docfx/Models/BuildCommand.cs
@@ -14,6 +14,12 @@ internal class BuildCommand : Command<BuildCommandOptions>
     {
         return CommandHelper.Run(settings, () =>
         {
+            if (settings.Serve && CommandHelper.IsTcpPortAlreadyUsed(settings.Host, settings.Port))
+            {
+                Logger.LogError($"Serve option specified. But TCP port {settings.Port ?? 8080} is already being in use.");
+                return;
+            }
+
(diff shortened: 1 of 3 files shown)


filzrev:
> What if we do something like this?

Personally, I prefer the `fail-fast approach' to the proposed approach.
For the following reasons.

- If BuildCommand is run twice based on the same `docfx.config`.  
  `_site` contents are overwritten if `fail-fast` check not exists.  
  And the content hosted on `port:8080' will be modified also.
  In this case. it makes no sense to host contents with alternative port.
- When acceptting user input. It needs additional consideration about non-UserInteractive environment.

yufeih (approved):
Thank you @filzrev !

PR #9700, opened 2024-02-15
fix: InvalidInputFile error occurs if file contains URI escaped charactors

### LINKED ISSUE #5132: Unable to process files with url-escaped symbols in names
**Operation System**: (`Windows` or `Linux` or `MacOS`)
tested on Windows

**DocFX Version Used**: 
2.39.2.0 (reproducable at 2.45.1 as well)

`default`

**Steps to Reproduce**:
1. name the file with %2D or any other url-escape code, e.g. `im%2Descaped.md` (as azure devops wiki does)
2. use `"files": [ "*.md" ]` in docfx.json
3. try to build

**Expected Behavior**:
document builds

**Actual Behavior**:
```
Build failed.
[19-09-24 08:42:50.944]Error:System.IO.DirectoryNotFoundException: Unable to find part of path "D:\md\azurewiki2docfx\sources\im%2Aescaped.md".
   в System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
```

[Minimal reproducible example (ZIP)](https://github.com/dotnet/docfx/files/3646270/docfx-percent-naming-bug.zip).

filzrev:
This PR intended to fix #5132.

- Fix `PathUtility.MakeRelativePath` methods to use `Path.GetRelativePath` method if specified path is file protocol. 
  -  It seems `file://` protocol can't be handled by this `PathUtility.MakeRelativePath`.   
     So add fallback to use existing implementation (`Uri::MakeRelativeUri`)
- Add additional tests for `PathUtility.MakeRelativePath` method.

### DIFF
--- a/test/Docfx.Common.Tests/PathUtilityTest.cs
+++ b/test/Docfx.Common.Tests/PathUtilityTest.cs
@@ -9,7 +9,7 @@ namespace Docfx.Common.Tests;
 public class PathUtilityTest
 {
     [Theory]
-    [MemberData(nameof(TestData.AdditionalTests), MemberType = typeof(TestData), DisableDiscoveryEnumeration = false)]
+    [MemberData(nameof(TestData.AdditionalTests), MemberType = typeof(TestData))]
     public void TestMakeRelativePath(string basePath, string targetPath, string expected)
     {
         // Act
@@ -20,7 +20,7 @@ public class PathUtilityTest
     }
 
     [Theory]
-    [MemberData(nameof(TestData.EscapedPaths), MemberType = typeof(TestData), DisableDiscoveryEnumeration = true)]
+    [MemberData(nameof(TestData.EscapedPaths), MemberType = typeof(TestData))]
     public void TestMakeRelativePathWithEncodedPath(string inputPath)
     {
         // Arrange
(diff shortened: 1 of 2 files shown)


test/Docfx.Common.Tests/PathUtilityTest.cs
```diff
@@ -0,0 +1,55 @@
...
+
+using FluentAssertions;
+using Xunit;
+
+namespace Docfx.Common.Tests;
+
+public class PathUtilityTest
+{
+    [Theory]
+    [MemberData(nameof(TestData.AdditionalTests), MemberType = typeof(TestData), DisableDiscoveryEnumeration = false)]
```
yufeih, line 12: nit: is there a reason why this marks `DisableDiscoveryEnumeration=false` and the other true? also consider `[InlineData]`?

yufeih (approved):

test/Docfx.Common.Tests/PathUtilityTest.cs
filzrev: > is there a reason why this marks DisableDiscoveryEnumeration=false

No. It's not necessary settings.
I've created commit that removed `DisableDiscoveryEnumeration=false` settings. (https://github.com/dotnet/docfx/pull/9700/commits/2a32f896a337e33d7caa77511374fc031d14bdf0)

That settings are originally added for test data that can't serialized.
But it's no more needed because test argument is changed to primitive strings.

> would [InlineData] work here?

It can rewrite tests to use `[InlineData]`.
But I', personally prefer to separating test data definitions for following reasons.

- Readability. When using `TheoryData`. It can manually align formatting of test data.
- It can reuse test data set with multiple tests.
- [InlineData] support only primitive types.  
  When using `MemberData`.  It can use dedicated types. And it can change name that displayed on TestExplorer.

Issue #8028, opened 2022-05-24
DefineConstants / MSBuild Properties not working with non-csproj configurations.


Lachee:
**Operating System**: (`Windows` or `Linux` or `MacOS`)
Windows

**DocFX Version Used**: 
- docfx 2.44.0.0
- docfx 2.59.2.0

**Template used**: (`default` or `statictoc` or contain custom template)
default

**Steps to Reproduce**:
_Via Example_
1. `git clone https://github.com/Lachee/docfx-define-constants-issue.git` - Repository highlighting issue
2. `cd docfx-define-constants-issue`
3. `docfx Documentation/docfx.json`

**Expected Behavior**:
The `DefineConstants` property gets applied to the `MSBuild` and classes within a `#if` condition have their documentation generated

**Actual Behavior**:
Anything within a `#if` gets ignored.

**Other**:

I have generated a sample project that experiences the issue here: https://github.com/Lachee/docfx-define-constants-issue but the main project that has an issue is my [Unity Utilities](https://github.com/Lachee/unity-utilities). It is a Unity Package so doesnt directly have .csproj itself.

Here is the metadata i used
```json
    "metadata": [
        {
            "src": [
                {
                    "src": "..",
                    "files": [ "Scripts/**/*.cs" ]
                }
            ],
            "globalNamespaceId": "Global",
            "dest": "api",
            "properties": {
                "DefineConstants": "SOME_DEFINE=1"
            }
        }
    ],
```

KalleOlaviNiemitalo:
`SOME_DEFINE=1` is odd because C# conditional compilation defines cannot have values.

[...]

PR #9707, opened 2024-02-20
feat: Add DefineConstants supports for source file based build

filzrev:
This PR intended to fix https://github.com/dotnet/docfx/issues/8028.

When building API metadata from C#/VB source files.
The `DefineConstants` properties defined in `docfx.json` are not currently used.

This PR add logics to read `DefineConstants` MSBuild properties and set values to `CSharpParseOptions`/ `VisualBasicParseOptions`.


yufeih (approved):
Looks great thank you @filzrev !

Issue #9659, opened 2024-01-29
[Announcement]📢Microsoft Learn xref service to be retired on March 8, 2024


yufeih:
[Microsoft Learn](https://learn.microsoft.com/) will retire the Learn xref service on March 8, 2024. 

Both of these 2 APIs for querying .NET xrefs published to Learn will be retired after the target date:

* https://learn.microsoft.com/api/xref/query?uid={uid}
* https://xref.docs.microsoft.com/query?uid={uid}

**Actions Required:** if you are using the above URLs to resolve xref reference to learn, change your `docfx.json` to use [this xrefmap file](https://github.com/dotnet/docfx/blob/main/.xrefmap.json): 

```json
{
  "build": {
    "xref": [
      "https://github.com/dotnet/docfx/raw/main/.xrefmap.json"
    ]
  }
}
```

We understand that this may cause inconvenience to some of our users, and we apologize for any disruption this may cause.

filzrev:
Uploaded `.xrefmap.json` file size is about `131MB`.
And it's downloaded when running `docfx build` command every times.

I though it's better to used pre-download `xrefmap` files. 

**Steps to use**
1. Download xrefmap as `XRefArchive` (Zip Archived YAML file) with following command
> docfx download xrefmap.zip --xref https://github.com/dotnet/docfx/raw/main/.xrefmap.json
2. Set `docfx.json` to use local zip fie.
```
"xref": [ "xrefmap.zip"]
```

---
In addition. When downloading `xrefmap` from `https://github.com/dotnet/docfx/raw/main/.xrefmap.json`
It's redirected to` https://media.githubusercontent.com/media/dotnet/docfx/main/.xrefmap.json`
(It's managed by git LFS.)

And HTTP compression seems not supported when downloading `.xrefmap.json` file.

filzrev:
> docfx download xrefmap.zip --xref https://github.com/dotnet/docfx/raw/main/.xrefmap.json

`docfx download` command that is commented above is not works as I've expected.
Because docfx repositories `.xrefmap.json` file use `baseUrl: https://learn.microsoft.com/dotnet/api/`.
And `href` value is set as relative URL from `baseUrl`. 

But `docfx download` command rewrite `baseUrl` to `https://github.com/dotnet/docfx/raw/main/`
And `href` values are converted to absolute URL.

glopesdev:
[...]

PR #9721, opened 2024-02-22
fix: `docfx download` command behaviors

filzrev:
This PR intended to fix problems commented at https://github.com/dotnet/docfx/issues/9659#issuecomment-1957018027

**Background**
When running `docfx download` with following command.
> docfx download xrefmap.zip --xref https://github.com/dotnet/docfx/raw/main/.xrefmap.json

Xrefmap is downloaded as `xrefmap.zip`.
However, the contents of `xrefmap.yml` are modified as follows.
1. `baseUrl` (`https://learn.microsoft.com/dotnet/api/`) is updated to `https://github.com/dotnet/docfx/raw/main/`.
2. `href` with relative URLs are rewritten to absolute URL that based on `https://github.com/dotnet/docfx/raw/main/`.

This PR fix above behaviors by removing relating code.
And this changes. `docfx download` command simply execute following tasks. 
- Download `.json` or `.yml` file from specified URL.
- Load downloaded content as XRefMap model. And serialize data to `xrefmap.yml`.
- Sort `XRefMap`'s references by UID to optimize performance.
- Save zip archive as specified file name. (`xrefmap.zip`)
---

- Remove `baseUrl`/`href` update related logics that are used when `docfx download` command.
- Remove unused code. (Relating `CreateMinor` method)
- Add documentation comment for `XRefMap` type.
- Add warning messages when specify zip archive to download URL. (Currently it seems silently ignored. See: #2805)
- Sort references by UID to optimize performance (for not sorted xrefmap), 
   and to support not properly sorted xrefmap. (sorted:true but not properly sorted) (See: https://github.com/NormandErwan/UnityXrefMaps/issues/4)

**Test**
I've tested `docfx download` command with following xrefmaps.
And It works as expected.
- `https://github.com/dotnet/docfx/raw/main/.xrefmap.json`
- `https://normanderwan.github.io/UnityXrefMaps/xrefmap.yml`

**Additional notes**
This PR contains some `docfx download` command behavior changes.
It might be better to list changelog as `BREAKIN CHANGES`.

### DIFF
--- a/src/Docfx.Build/XRefMaps/XRefArchiveBuilder.cs
+++ b/src/Docfx.Build/XRefMaps/XRefArchiveBuilder.cs
@@ -24,7 +24,7 @@ public class XRefArchiveBuilder
             try
             {
                 using var xa = XRefArchive.Open(outputFile, XRefArchiveMode.Create);
-                await DownloadCoreAsync(uri, xa, true);
+                await DownloadCoreAsync(uri, xa);
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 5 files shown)


src/Docfx.Build/XRefMaps/XRefArchiveBuilder.cs
```diff
@@ -36,88 +36,32 @@ public async Task<bool> DownloadAsync(Uri uri, string outputFile)
...
+            Logger.LogWarning($"Download an xref archive, or reference to an xref archive is not supported. URI: {uri}");
             return null;
         }
-        if (map.Redirections?.Count > 0)
-        {
-            await RewriteRedirections(uri, xa, map);
-        }
-        if (map.References?.Count > 0 && map.HrefUpdated != true)
-        {
-            if (string.IsNullOrEmpty(map.BaseUrl))
```
yufeih, line 54: Does it make sense that URLs are relative to the `baseUrl` in xrefmap if exist, otherwise relative to the download location?

filzrev: I've added following logics for xrefmaps that don't define `baseUrl` and using relative urls. (https://github.com/dotnet/docfx/pull/9721/commits/6e665d3d0d590fe9c52aa09bcb16909c978dd84e)
1. If `baseUrl` is not defined. Set `download location` as `baseUrl`.
2. If `baseUrl` is not defined. Try to resolve href that are using relative URL to absolute URL.

In addition. I've added `launchSettings.json` settings to test `docfx download` command from Visual Studio.

yufeih (approved):
Looks good!

Issue #9296, opened 2023-10-16
[Feature Request] Disable TOC in modern template


cdervout:
Hello

I want to keep the affix area but disable TOC on one page.
As with the default template, it would be interesting to make the following available : `_disableToc` and `_disableAffix`.
Today, the only option is to create a custom template.

paulushub:
> As with the default template, it would be interesting to make the following available : `_disableToc` and `_disableAffix`.

I do not know if I understand your point. The above options are available:
https://dotnet.github.io/docfx/reference/docfx-json-reference.html

cdervout:
Yes. The ones I'm interested in are only available for the default template : https://dotnet.github.io/docfx/docs/template.html?q=_layout&tabs=default#template-metadata

yufeih:
The the `_layout: landing` option would hide both the TOC and the Affix, as well as removing the extra margin/padding allowing you full control of the content.

cdervout:
Exactly, but I'm only interested in deactivating the TOC, not the affix (for a single page).

PR #9737, opened 2024-02-27
feat: Add additional layout options to modern template

### LINKED ISSUE #8686: [Bug] Modern Template Cannot Disable Breadcrumbs
I cannot seem to disable the breadcrumb bar with the modern template, using docfx version 2.64. I have added `"_disableBreadcrumb": true` to the `"_globalMetadata"` of my docfx.json config. Indeed, using the default template, the expected behavior is observed. However, the breadcrumb bar is not disabled when using the modern template.

filzrev:
This PR intended to fix #8686.
By adding following additional layout options to `modern` template.
- `_disableNavbar`
- `_disableBreadcrumb`
- `_disableToc`
- `_disableAffix`

**Test**
It's manually tested on my local environment.
And confirms that the following checks have been passed. 
- When setting `_disable*` option. The corresponding elements is not be rendered in HTML.
- No JavaScript Error occurred when features are disabled.

**Background**
It seems some users are requesting above options. (#8686, #9116, #9296)
It can partially resolved by using `_layout: landing` option. (This option disable both toc and affix)
But there are feature requests that hide individual elements separately.

### DIFF
--- a/docs/docs/template.md
+++ b/docs/docs/template.md
@@ -40,6 +40,10 @@ Name         | Type    | Description
 `_gitContribute`        | object  | Defines the `repo` and `branch` property of git links.
 `_gitUrlPattern`        | string  | URL pattern of git links.
 `_disableNewTab`        | bool    | Whether to render external link indicator icons and open external links in a new tab.
+`_disableNavbar`        | bool    | Whether to show the navigation bar.
+`_disableBreadcrumb`    | bool    | Whether to show the breadcrumb.
+`_disableToc`           | bool    | Whether to show the TOC.
+`_disableAffix`         | bool    | Whether to show the right rail.
 `_disableNextArticle`   | bool    | Whether to show the previous and next article link.
 `_disableTocFilter`     | bool    | Whether to show the table of content filter box.
 `_googleAnalyticsTagId` | string  | Enables Google Analytics web traffic analysis.
--- a/templates/modern/layout/_master.tmpl
+++ b/templates/modern/layout/_master.tmpl
@@ -60,6 +60,7 @@
 
   <body class="tex2jax_ignore" data-layout="{{_layout}}{{layout}}" data-yaml-mime="{{yamlmime}}">
     <header class="bg-body border-bottom">
+      {{^_disableNavbar}}
       <nav id="autocollapse" class="navbar navbar-expand-md" role="navigation">
         <div class="container-xxl flex-nowrap">
           <a class="navbar-brand" href="{{_appLogoUrl}}{{^_appLogoUrl}}{{_rel}}index.html{{/_appLogoUrl}}">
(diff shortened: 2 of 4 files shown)


filzrev:
Snapshot diff seems be occurred by following reasons.

`_disableToc` metadata variable is overriden by [`conceptual.html.primary.js`]. (https://github.com/dotnet/docfx/blob/main/templates/default/conceptual.html.primary.js#L12C1-L12C99)
with following condition.
>  model._disableToc = model._disableToc || !model._tocPath || (model._navPath === model._tocPath);`

It seems `index.md` root file met above conditions. 
And `_disableToc` is overwritten to `true`.

yufeih:
Thank you @filzrev - does the manual test include mobile views as well?

filzrev:
> does the manual test include mobile views as well?
It seems works with smaller window size layout.

And I've removed `min-height` style from `actionbar` element.
Because it display blank element when using `_disableBreadcrumb: true` settings.
https://github.com/dotnet/docfx/pull/9737/commits/e5b5ec94be36f7883d6c4830a20094ba2b367c94

By this changes. default `actionbar` element height is reduced from 40px->37px.
I thought this changes is not a problem for usability (To tap hamburger menu of TOC on mobile view)

yufeih (approved):
Looks good, thank you @filzrev !

filzrev:
It's need to update snapshots.

There is concerning about `breaking changes`.
Currently docfx implicitly set `_disableToc` by following condition.

> model._disableToc = model._disableToc || !model._tocPath || (model._navPath === model._tocPath);`

But existing modern template render TOC elements regardless `disableToc` options. 
By this PR. This behavior is aligned to `default` template.

if `model._navPath === model._tocPath`. Then TOC is not displayed.
And it can't be overwritten by `docfx.json` or file metadata.

VaclavElias:
Hmm, it was unexpected.

I had to add this to get the toc back on the home page. Most likely, when you deploy your docs, you might have the same issue, missing toc on the home page.

```
    "fileMetadata": {
      "_disableToc": {
        "index.md": false
      }
    }
```

filzrev:
It's breaking changes for `modern` template user (that is commented at https://github.com/dotnet/docfx/pull/9737#issuecomment-1968030785).

From version 2.76.0. 
`default` template compatible layout attribute supports are added.
And default behavior is aligned to `default` template. (TOC is not displayed at homepage by default)

So who don't use `_layout: landing settings. And want to display TOC at homepage.
It' required to set `_disbleToc: false` setting to YAML header explicitly.

VaclavElias:
> setting to YAML header explicitly.

Do you mean a front matter of `index.md`?

Like so?

```
---
_disableToc: false
---
```

filzrev:
> Do you mean a front matter of index.md?
Sorry for the confusion. It means YAML Frontmatter as you indicated.

> Probably, it needs to be added also here? https://github.com/dotnet/docfx/blob/main/docs/index.md

Official docs index page don't match following conditions. 
> model._navPath === model._tocPath

So there is no need to set `_disableToc: false` explicitly.

VaclavElias:
@filzrev, thank you for clarification.

PR #9741, opened 2024-02-28
chore: Don't overwrite `_disabledToc` when metadata is explicitly defined

filzrev:
This PR fix problem occurred by #9737.

**What's changed in this PR**
- Set `_disableToc` metadata only when `_disableToc` is not defined yet.
- Set `_disableToc` to `index.md` file to suppress HTML snapshot diffs.


filzrev:
I'll add commit to remove `_disableToc` metadata from `seed` site.
Because it specify `_layout: landing` so  `_disableToc`  settings is not works anyway.

yufeih (approved):

Issue #9730, opened 2024-02-24
[Bug] Print styles being ignored


cjlotz:
**Describe the bug**
Custom styles created for printing does not seem to get applied.  I'm trying to change the background-color of the code blocks as the background color used for web output does not match the printed output.  I've added the following CSS to my custom `main.css` file

```css
@media print {
    section[role="tabpanel"] {
        display: block !important;
    }

    .hljs {
        background-color: aliceblue;
    }
}
```

The section[role] gets applied, but the background-color overide does not get applied.  I've verified that the .hljs override does work when applied to the whole site and not only to print media selector.

yufeih:
PDF does not print background graphics by default, we could add an option to opt-in background graphics.

cjlotz:
@yufeih is background-color classified as background graphics?  All I'm looking for is for the area assigned to the code fence to have a different background color so that it is visible for the user from where they should be selecting the text to paste and execute outside of the document.

yufeih:
> Is background-color classified as background graphics? 

You can save the page as PDF in the browser and test with "Background Graphics" off:

<img width="931" alt="Screenshot 2024-02-26 at 9 06 38 PM" src="https://github.com/dotnet/docfx/assets/511355/69eeed0d-7ca2-4137-b927-6cc1124cbdcd">

cjlotz:
I can confirm that I turn the Background graphics setting on, the background-color is indeed rendered correctly.  Having a setting to opt into this behaviour will be great.

PR #9742, opened 2024-02-28
Added pdfPrintBackground setting

### LINKED ISSUE #9740: Print background when creating PDF
Is it possible, that the PDF creating is made with option PrintBackground enabled?
This would result for example in more beautiful Alerts in created PDF file:
```markdown
> [!NOTE]
> Information the user should notice even if skimming
```
Source: https://github.com/dotnet/docfx/blob/a58fec70cdaa98d6348d1cb3e0185f19a5ca6c48/src/Docfx.App/PdfBuilder.cs#L167C17-L167C18
Suggested change:
```cs
return await page.PdfAsync( new PagePdfOptions()
{
	PrintBackground = true
});
```

cjlotz:
Added pdfPrintBackground setting to include background graphics as part of pdf printing.  Fixes #9740 #9730

### DIFF
--- a/docs/docs/pdf.md
+++ b/docs/docs/pdf.md
@@ -79,6 +79,10 @@ Indicates whether to include a "Table of Contents" pages at the beginning.
 
 A path to an HTML page relative to the root of the output directory. The HTML page will be inserted at the beginning of the PDF file as cover page.
 
+### `pdfPrintBackground`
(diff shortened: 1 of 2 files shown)


yufeih (approved):
Looks great! Thank you @cjlotz !

Issue #9736, opened 2024-02-27
[Bug] Broken HTML for <code> from inherited xml-comment


bkoelman:
**Describe the bug**
When using `/// <inheritdoc />` on a class property that implements an interface, its example code is escaped twice, resulting in broken HTML.

**To Reproduce**
1. Download the zip file below, which contains a minimal repro project
    [DocfxEscapeBugRepro.zip](https://github.com/dotnet/docfx/files/14412694/DocfxEscapeBugRepro.zip)
2. Run `dotnet tool restore`
3. Run `dotnet docfx docs/docfx.json --serve`
4. Open `http://localhost:8080` in your browser
5. Click **API** > **IJsonApiOptions**
6. Observe the example HTML is rendered properly
    ![image](https://github.com/dotnet/docfx/assets/10324372/88cc7783-8028-4410-ab9f-ede4067c278f)
8. Click **JsonApiOptions**
9. Observe the example HTML is broken
    ![image](https://github.com/dotnet/docfx/assets/10324372/7e3bf054-622d-435b-82a2-815cd1801240)

**Expected behavior**
Example code is rendered properly on both pages.

**Context (please complete the following information):**
- OS: Windows 11
- Docfx version: 2.75.3
- .NET version: .NET 8
- `docfx.json` config: as produced by `dotnet docfx init`, see zip file
- Exceptions: none
- Errors and warnings: none
- .NET info:
```
.NET SDK:
 Version:           8.0.201
 Commit:            4c2d78f037
 Workload version:  8.0.200-manifests.e575128c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22631
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.201\

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.2
  Architecture: x64
  Commit:       1381d5ebd2

.NET SDKs installed:
  8.0.200 [C:\Program Files\dotnet\sdk]
  8.0.201 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.App 6.0.27 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
  Microsoft.AspNetCore.App 8.0.2 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.App]
[...]

Issue #9495, opened 2023-12-02
Broken examples after update


bkoelman:
**Describe the bug**

After update from 2.74.0 to 2.74.1, examples are broken. See https://github.com/json-api-dotnet/JsonApiDotNetCore/pull/1415#issuecomment-1837126296.

**To Reproduce**
1. Clone the repo at https://github.com/json-api-dotnet/JsonApiDotNetCore
2. `cd docs`
3. `./build-dev.ps1`
4. In the browser, navigate to **Getting Started** > **API** > **JsonApiOptions**

**Expected behavior**
No broken HTML.

yufeih:
Seems to be working just fine:

<img width="1328" alt="image" src="https://github.com/dotnet/docfx/assets/511355/5a95411f-a7af-4269-8a85-9d05b2b2a5ca">

bkoelman:
Did you try on Ubuntu?

You can see the diff by downloading and extracting the documentation artifact from https://github.com/json-api-dotnet/JsonApiDotNetCore/actions/runs/7055097802 and https://github.com/json-api-dotnet/JsonApiDotNetCore/actions/runs/7208009175.

The next screenshot shows the diff:
![image](https://github.com/dotnet/docfx/assets/10324372/6468c006-fd27-44bf-8819-e9306e57ab63)

Here's what it looks like in Chrome:
![image](https://github.com/dotnet/docfx/assets/10324372/a8731c6b-e3e7-483d-864e-a73a536ca425)

bkoelman:
Well, it does not render entirely correct in v2.74.0 on Windows either:
```c#
    /// <summary>
    /// Whether to use relative links for all resources. <c>false</c> by default.
    /// </summary>
    /// <example>
    /// <code><![CDATA[
    /// options.UseRelativeLinks = true;
    /// ]]></code>
    /// <code><![CDATA[
    /// {
    ///   "type": "articles",
    ///   "id": "4309",
    ///   "relationships": {
    ///      "author": {
    ///        "links": {
    ///          "self": "/api/shopping/articles/4309/relationships/author",
    ///          "related": "/api/shopping/articles/4309/author"
    ///        }
    ///      }
    ///   }
    /// }
    /// ]]></code>
    /// </example>
    bool UseRelativeLinks { get; }
```

![image](https://github.com/dotnet/docfx/assets/10324372/7d766e80-ad0f-41cf-9f9a-e87e3d472f7f)

bkoelman:
[...]

PR #9754, opened 2024-03-03
fix: Broken examples problems when using inheritdoc tag

filzrev:
This PR intended to fix problems that are reported by #9736, #9495.

When using `example` tag with `<inheritdoc />`.
`symbol.GetDocumentationComment` API returns slightly different format xml.
And it cause wrong metadata generation.

**Problems of parsed results**
**1. Parsed example comments contains heading spaces.**

It cause rendered as markdown as `Fenced code blocks`.

**2. HTML `<pre>` element without preceding line break cause wrong HTML output. ** 

This may be due to CommonMark specifications.  
Because CommonMark/Markdig/GFM have same rendered result.
By this problem. it must be formatted as indented XML.

**Example markdown code that reproduce problems**
```
DUMMY<pre>{
  indended text
}</pre>
```

```
DUMMY
<pre>{
  indended text
}</pre>
```

### DIFF
--- a/src/Docfx.Dotnet/Parsers/XmlComment.cs
+++ b/src/Docfx.Dotnet/Parsers/XmlComment.cs
@@ -112,6 +112,10 @@ internal class XmlComment
         }
         try
         {
+            // Format xml with indentation.
+            // It's needed to fix issue (https://github.com/dotnet/docfx/issues/9736)
+            xml = XElement.Parse(xml).ToString(SaveOptions.None);
+
             return new XmlComment(xml, context ?? new());
         }
         catch (XmlException)
--- a/test/Docfx.Dotnet.Tests/XmlCommentUnitTest.cs
+++ b/test/Docfx.Dotnet.Tests/XmlCommentUnitTest.cs
@@ -28,8 +28,7 @@ public class XmlCommentUnitTest
         Assert.Equal(
             """
             a
-            <p>b</p>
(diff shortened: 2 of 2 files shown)


filzrev:
**Unit test code**
Following unit test code should be added to  `samples/seed/dotnet/project/Project/Inheritdoc.cs`.
(It need to add snapshot of `*.html.verified.json` files. But these files should be created by CI)

```
    public class Issue9736
    {
        public interface IJsonApiOptions
        {
            /// <summary>
            /// Whether to use relative links for all resources. <c>false</c> by default.
            /// </summary>
            /// <example>
            /// <code><![CDATA[
            /// options.UseRelativeLinks = true;
            /// ]]></code>
            /// <code><![CDATA[
            /// {
            ///   "type": "articles",
            ///   "id": "4309",
            ///   "relationships": {
            ///      "author": {
            ///        "links": {
            ///          "self": "/api/shopping/articles/4309/relationships/author",
            ///          "related": "/api/shopping/articles/4309/author"
            ///        }
            ///      }
            ///   }
            /// }
            /// ]]></code>
            /// </example>
            bool UseRelativeLinks { get; }
        }

        public sealed class JsonApiOptions : IJsonApiOptions
        {
            /// <inheritdoc />
            public bool UseRelativeLinks { get; set; }
        }
    }
```

yufeih (approved):

PR #9759, opened 2024-03-05
feat: Support DOCFX_SOURCE_REPOSITORY_URL

filzrev:
This PR intended to fix issue #9743.

When running **SamplesTest** on a **forked** docfx repository.
A lot of diffs are raised regarding git-related path differences.

This PR rewrites these paths if `DOCFX_SOURCE_REPOSITORY` is set.

---

~~This PR rewrites these paths with the following conditions.~~
1. ~~When `EnvironmentContext.IsSnapshotTestContext=true` is explicitly set.  
    (Used for in-process tests)~~
2. ~~When `DOCFX_IS_SNAPSHOT_TEST_CONTEXT` environment variable is set.
    (Used for tests that launch docfx process)~~

> [!NOTE]
> ~~`DOCFX_IS_SNAPSHOT_TEST_CONTEXT` value is cached by `Lazy<bool>`.~~
> ~~Because `GitUtility` methods are called so many times. And I want to reduce `Environment.GetEnvironmentVariable`~~
> ~~system API call overhead. ~~
>
> ~~`InternalVisiblesTo` attributes are added to `Docfx.Plugins` project.~~
> ~~To reference `EnvironmentContext.IsSnapshotTestContext` internal property.~~
~~


src/Docfx.Common/Git/GitUtility.cs
```diff
@@ -38,9 +39,13 @@ record Repo(string path, string url, string branch);
         if (repo is null)
             return null;
 
+        var repoUrl = EnvironmentContext.IsSnapshotTestContext
```
yufeih, line 42: Can repo URL be configured from the `DOCFX_SOURCE_REPO_URL` environment similar to how `DOCFX_SOURCE_BRANCH_NAME` works (and how branch are unified for PR builds)?

filzrev:
> Can repo URL be configured from the DOCFX_SOURCE_REPO_URL environment similar to how DOCFX_SOURCE_BRANCH_NAME works (and how branch are unified for PR builds)?

Changed environment variable name to  `DOCFX_SOURCE_REPOSITORY` and set `dotnet/docfx` for snapshot test.
(Remove `_URL` postfix because it need to support non-GitHub resository URLs )

**Other changes**
- Add `GitUtilityWithSourceRepoUrlTest` to test organization/repository names override.
- Remove Environment variable caching. Because performance difference seem relatively small.
   (It's called about 1000 times when generating metadata for `samples/seed` projects)

docs/reference/docfx-environment-variables-reference.md
```diff
@@ -14,6 +14,11 @@ If set true. Keep following debug info in output HTML.
 
 Used to override git branch name.
 
+## `DOCFX_SOURCE_REPOSITORY`
+
+Used to override git organization and repository names.
+It must be defined in the `{organization}/{repository_name}` format (e.g.`dotnet/docfx`).
```
yufeih, line 20: `{owner}/{repo}` is a pattern for GitHub but doesn't seem to fit every other repo management systems. 

As this is a configuration exposed to public, I would prefer it to be `DOCFX_SOURCE_REPOSITORY_URL` but we only support GitHub URLs (`https://github.com/{owner}/{repo}`) as a starting point. With that you can pattern match the URL in `ResolveDocfxSourceRepoUrl` method and return the original URL if  it is not github.

filzrev: I've changed environment variable name  to `DOCFX_SOURCE_REPOSITORY_URL` by commit (fa41fed).

URL is replaced when
- `DOCFX_SOURCE_REPOSITORY_URL` environment variable is defined.
- URL host name is matched to `DOCFX_SOURCE_REPOSITORY_URL`.
- Hostname matches one of the following URLs (that use `{org_name}/{repo_name}` patterns)
  - `github.com`
  - `bitbucket.org`
  - `dev.azure.com`

yufeih (approved):
Thank you @filzrev !

PR #9762, opened 2024-03-06
feat: Support for all emphasis supported by MarkDig

EMaderbacher:
Currently docfx only supports emphasis Strikethrough.
This PR add support for all emphasis supported by Markdig

PR #9772, opened 2024-03-08
feat: Enable smileys in "emojis and smileys extension"

EMaderbacher:
Enable smileys in "emojis and smileys extension"

### DIFF
--- a/src/Docfx.MarkdigEngine.Extensions/MarkdownExtensions.cs
+++ b/src/Docfx.MarkdigEngine.Extensions/MarkdownExtensions.cs
@@ -28,7 +28,7 @@ public static class MarkdownExtensions
             .UseDFMCodeInfoPrefix()
             .UseQuoteSectionNote(context, notes)
             .UseXref()
-            .UseEmojiAndSmiley(false)
+            .UseEmojiAndSmiley()
             .UseTabGroup(context)
             .UseMonikerRange(context)
             .UseInteractiveCode()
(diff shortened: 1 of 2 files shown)


EMaderbacher:
@dotnet-policy-service agree

EMaderbacher:
Updated unit tests too because ` :*` will result in :kissing:
I'm not sure whether you'll classify this as breaking change...

filzrev:
> Updated unit tests too because  :* will result in 😗
> I'm not sure whether you'll classify this as breaking change...

When setting `enableSmileys=true`, following text are additionally replaced to emoji.

https://github.com/xoofx/markdig/blob/fcb56fb03701cc1a07a52769908579adb8927ee2/src/Markdig/Extensions/Emoji/EmojiMapping.cs#L1651-L1729

It it's set by default. Existing document might be affected. 
I thought it' better this setting as `opt-in` features. (And #9762 EmphasisExtras features also)
By explicitly setting following configs.

```
"build": {
    "markdownEngineProperties": {
      "markdigExtensions": [
          "EmphasisExtras",
          "Emoji"
      ]
    }
  }
```

Currently above config is not works.
Because optional extension is skipped when same extension type is already registered.
I'll try to fix this problem at another PR.

EMaderbacher:
Will moving `builder.UseOptionalExtensions(extensions);` before `builder.UseDocfxExtensions(...` (in `MarkdigMarkdownService`) have any negative impact?

filzrev:
> Will moving builder.UseOptionalExtensions(extensions); before builder.UseDocfxExtensions(... (in MarkdigMarkdownService) have any negative impact?

Some markdig extensions seems to have specific order dependencies.
It need to preserve extension orders as possible.

I've created PR #9780 to override existing markdig extension with default settings.

yufeih:
Thank you @filzrev for pointing out and yes this is a breaking change and needs opt-in. Closing in favor of #9780

PR #9780, opened 2024-03-11
feat: Add feature to reset markdig extension settings that are added by default

filzrev:
This PR intended to add functionality to use Markdig extensions with default settings. 

**Background**
There are some PRs(#9762/#9772)  are created that request changing markdig extensions setting to support advanced syntax.
But these change might break existing markdown documents.

So these changes are should be enabled as `opt-in` features.

**What changed in this PR**
This PR add functionality to reset settings when existing **known** markdown extension names are specified.
(And it preserve extension orders.)

**Usage**
When specify existing "markdigExtensions" names. 
These markdig extensions are replaced with markdig default settings.

```json
{
  "build": {
    "markdownEngineProperties": {
      "markdigExtensions": [
        "EmphasisExtras",
        "Emojis"
      ]
    }
  }
}
```


src/Docfx.MarkdigEngine.Extensions/MarkdownExtensions.cs
```diff
@@ -56,9 +57,34 @@ public static class MarkdownExtensions
...
 
+        ReplaceExistingExtensions();
+
         pipeline.Configure(string.Join("+", optionalExtensions));
 
         return pipeline;
+
+        // Some markdig extensions are added by UseDocfxExtensions with non-default settings.
+        // This methods replace these extensions with default settings.
+        void ReplaceExistingExtensions()
```
yufeih, line 68: These feel like configuration of existing extensions, it might be more future proof if we configure it such as:

```
markdigExtensions: [
    'other-extensions'
    { emoji: { enableSmileys: true }],
    { autoidentifier: 'default' }
]
```

filzrev:
Close this PR.

When resolving PR conflicts. Unintended file changes are pushed.
To clearer change histories.  
I've created another PR (#9820).

src/Docfx.MarkdigEngine.Extensions/MarkdownExtensions.cs
filzrev: I've added commented features to override extension properties by PR #9820.

**Example configs**

```
markdigExtensions: [
    "FootNotes",
    { "Emojis": { "options": {  "enableSmileys": true } },
    { "AutoIdentifiers": { "options": 'default '} },
    { "MediaLinks": { "options": { "width": 800, "height": 400 }} 
]
```

Note: I've added `options` key. Because it might required to support additional property
when implementing **`remove` markdig extensions that are enabled by default** feature.(Discussed at #7833)

PR #9746, opened 2024-02-29
chore: Add detailed error message for specific exceptions

filzrev:
This PR is intended to help diagnose the CI error which occurred in #9744.

**Error Message**
> /home/runner/work/docfx/docfx/docs/extensions/tools.yml: error InvalidInputFile: Unable to load file 'extensions/tools.yml' via processor 'Dashboard': Index was out of range. Must be non-negative and less than the size of the collection. (Parameter 'index')

It seems to be throwing an `ArgumentOutOfRangeException`. but there is no StackTrace information contained in log.
So I've added full exception message to following exception types.

- `ArgumentException` derived class
- `NullReferenceException`  
   (It occurred at #9146)

PR #9807, opened 2024-03-21
chore: Change JsonSchema EvaluationOptions instance to shared object

filzrev:
This PR intended to avoid error that reported at #9746.

When `EvaluationOptions` new  instance is passed to `JsonSchema::Evaluate` method.
JsonSchema.NET clear internal constraints cache and rebuild constraints.
And it cause errors when running Evaluate from multiple thread simultaneously.

In this PR. I've changed `EvaluationOptions` to `static readonly` shared object.


yufeih (approved):
🔥

PR #9810, opened 2024-03-22
Update markdown documentation regarding MarkDig extension

tbolon:
Add a link to markdig extension names, because there are cases (GenericAttributesExtension vs attributes) where the name is different from the class / extension name.


yufeih (approved):
Thank you @tbolon !

PR #9812, opened 2024-03-25
chore: Correct `docfx template` page title

peitschie:


yufeih (approved):
Thank you @peitschie !

Issue #9808, opened 2024-03-21
[Bug] EnumSortOrder.DeclaringOrder in docfx.json doesn't work


Culter:
**Describe the bug**
Enum values are always shown in alphabetical order, regardless of the requested order in docfx.json.

**To Reproduce**
1. Add any of the following attempts to an existing `docfx.json` file:
    * `"enumSortOrder": "declaringOrder"`
    * `"enumSortOrder": "DeclaringOrder"`
    * `"EnumSortOrder": "declaringOrder"`
    * `"EnumSortOrder": "DeclaringOrder"` 
2. Note that this property name is not allowed by the published schema at https://json.schemastore.org/docfx.json, but let's try to build the site anyway.
3. docfx docfx.json --serve

**Expected behavior**
1. Enum values should be shown in declaring order.
    * Alternatively, if the current version of the docfx tool doesn't recognize a property specified in the docfx.json file, or if it rejects the syntax of a property, it should output a warning on the command line when it builds the site.
2. https://json.schemastore.org/docfx.json should be maintained, so that editing a docfx.json file in a modern editor (Visual Studio) helps the author to find the right syntax without trial-and-error.
3. https://dotnet.github.io/docfx/reference/docfx-json-reference.html#enumsortorder should also be updated with the correct capitalization, whichever it is.

**Context (please complete the following information):**
- OS: Windows 10 Pro
- Docfx version:
```
>dotnet --version
8.0.200

>docfx --version
4.75.3+a5c719410037cc288c7adff010b1abf3c0f2e581
```

**Additional context**
This PR, merged August 2023:
* https://github.com/dotnet/docfx/pull/9093/

...purports to fix this issue from December 2017:
* https://github.com/dotnet/docfx/issues/2285

BarkenBark:
- Where in you _docfx.json_ are you adding `"enumSortOrder"`? For me, the following works:
```json5
  "metadata": [
    {
      // Other stuff
      "enumSortOrder": "declaringOrder"
    }
  ],
```
- Your `docfx` version seems strange. Isn't the latest release 2.76?

rpallares:
Hi, I'm using DocFx@2.78.2 and this doesn't work:
`"enumSortOrder": "declaringOrder"`

filzrev:
I've confirmed `"enumSortOrder": "declaringOrder"` settings works as expected on both `mref`/`apipage` output formats.   
(Environment: Windows + docfx v2.78.2)

If problems remain.
[...]

PR #9813, opened 2024-03-25
chore: Update `docfx.json` config reference document

filzrev:

**1. Update `docfx-json-reference.md` file contents**
  - Fix some lint warnings.
  - Change `rules` heading level
  - Add following missing property document
    -  `customLinkResolver`
    -  `groups`
    -  `references`
    -  `codeSourceBasePath`
  - Fix `enumSortOrder` naming (reported at #9808)

**2. Fix `MetadataJsonConfig.cs` file.**
  - Remove unreferenced `Raw` property.
  - Add doccomment for `References` property.   
    (Currently this property is not works for `source-file` based metadata generation. I'll create PR for this issue later)
  - Update doccomment for `group` property.  
    (This property accepts group metadata. But it seems not used in current implementation)


yufeih (approved):
Thank you @filzrev !

PR #9824, opened 2024-03-29
perf: Optimize xrefmap.json file deserialization performance

filzrev:
**What's changed in this PR**

When XrefMap is loaded from local file or network.
The current implementation using `YamlDotNet`'s deserialize method.

For JSON file deserialization.
It seems to be faster to use `Newtonsoft.json` for JSON deserialization.

**Benchmark**

|             |                |               |
|---------|-----------|-----------|
| YamlDotnet         | 8781[ms] |
| Newtonsoft.Json | 2304[ms] |
| System.Text.Json | 1687[ms] |  Note: It's not included in this PR. I'll create separated PR later.

### DIFF
--- a/src/Docfx.Build/XRefMaps/XRefMapDownloader.cs
+++ b/src/Docfx.Build/XRefMaps/XRefMapDownloader.cs
@@ -111,13 +111,25 @@ public class XRefMapDownloader
     private static IXRefContainer ReadLocalFile(string filePath)
     {
         Logger.LogVerbose($"Reading from file: {filePath}");
-        if (".zip".Equals(Path.GetExtension(filePath), StringComparison.OrdinalIgnoreCase))
-        {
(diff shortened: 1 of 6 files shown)


paulushub:
Is file-based database (like sqlite) possible as an option? query might be faster.

filzrev:
> Is file-based database (like sqlite) possible as an option? query might be faster.

I though faster xrefmap loading is important. 
Because loading/deserializing xrefmap over 100MB size JSON file take some times and take memory footprint.
And this operation is executed on every `docfx build`. 
 
So it might be better to use file-based database or some key-value storage as suggested.

Some of the required characteristics are.

1. File-based database/storage.
2. It can use `Compressed Database` file and that size should be lower than 100MB for .NET xrefmap.  
   (100MB is GitHub file size limit without using Git LFS)
  (When compressing .NET `xrefmap.json` with gzip. It can reduce file size from `131MB` to `8.6`MB)

Issue #9706, opened 2024-02-19
[Feature Request] Add support to reference additional DLLs when build metadata with C# source files


filzrev:
**Is your feature request related to a problem? Please describe.**

It seems `C# file based` API metadata generation are used by users who use `Unity`.

- https://github.com/dotnet/docfx/issues/9675
- https://github.com/NormandErwan/DocFxForUnity/issues/28

In this case. It needs to reference additional DLLs that are provided by Unity. (e.g. `UnityEngine.dll`)
But It seems currently not possible by `docfx.json` config. 

It can generate API metadata partially. When setting `"allowCompilationErrors": true`. 
But some metadata that depends on external DLLs are not generated.

**Describe the solution you'd like**
Allow additional DLL file references when using C# file based build.

**Additional context**
Currently `CreateCompilationFromCSharpFiles` don't support additional references.
But `CreateCompilationFromCSharpCode` that used test code supports additional DLL references.
https://github.com/dotnet/docfx/blob/841fcdfe3d97b958caeaf78b86c6d183f20f854f/src/Docfx.Dotnet/CompilationHelper.cs#L55-L71

filzrev:
Close this issue.
#9825 is merged and docfx v2.76.0 is released.

PR #9825, opened 2024-03-29
feat: Add  DLL references support for file-based API metadata generation

filzrev:
**What's changed in this PR**
Currently metadata's [references](https://dotnet.github.io/docfx/reference/docfx-json-reference.html?q=references#references)  config is used only when generating API metadata from DLL files.

This PR intended to add same functionality to C#/VB source-file based build. (#9706)

**Other Changes**
Add `using`  keyword for `MSBuildWorkspace`. Because it's disposable.


yufeih (approved):
Thank you @filzrev ! This is great!

PR #9828, opened 2024-03-29
chore: Add a warning about relative path resolution.

voroninp:
It's not obvious whether paths are resolved based on current working directory where docfx is called or based on location of `docfx.json` file.

It's worth to be explicit about it in the documentation.


yufeih (approved):
Thank you @voroninp !

PR #9820, opened 2024-03-27
feat: Add feature to configure markdig extension by docfx.json configs.

filzrev:
This PR supersede #9780.

This PR intended to add functionality to **override settings** for some specific Markdig extensions.

**Changed behaviors**

1. When markdig extension name **ONLY**  specified.  
-> Then try to add or replace extension with **markdig default options**. (Used to reset docfx default settings)
2. When markdig extension name with `options` are specified.
-> Then try to add or replace extension with **specified options**.

Other markdig extensions that have not customizable configs are processed same as before..  
(By using `MarkdownPipelineBuilder.Configure` method ).

---

**What's changed in this PR**

1. Add **MarkdigExtensionSetting.cs** file.
  -  This class represent extension name with custom options.
  - `options` are hold with `System.Text.JsonElement`
2. Add **MarkdigExtensionSettingConverter.cs** file.
  - This class convert json config to `MarkdigExtensionSetting` object. 
3. Modify `MarkdownExtensions.cs` logics.
  - If known extension names are passed. It try to add or replace markdig extension with specified options. 
4. Modify `MarkdownServiceParameters.cs`
  - `MarkdigExtensions` propety type is changed from `string[]` to `MarkdigExtensionSetting`.
5. Update public API contracts file (`Api.verified.cs`)
6. Add unit tests for markdig extension that can specify custom option settings.

**Additional Notes**
Original reported feature requests (#9762 / #9772) are resolved by this PR.
By using following configurations.

```
{
  "build": {
    "markdownEngineProperties": {
      "markdigExtensions": [
        "EmphasisExtras",
        "Emojis"
      ]
    }
  }
}
```

If it need to specify options.
Pass `options` properies with extension name.

```
markdigExtensions: [
    "FootNotes",
    { "Emojis": { "options": {  "enableSmileys": true } },
    { "AutoIdentifiers": { "options": "default" } },
    { "MediaLinks": { "options": { "width": 800, "height": 400 }} 
]
```

PR #9830, opened 2024-04-01
chore: Remove options from MarkdigExtensionSetting

yufeih:
Some minor tweak to #9820 

1. Simplify config by removing the `options` property
2. Change emoji config from bool to an enum

```json
{
  "markdigExtensions": [
    "FootNotes",
    {"Emojis": "default"},
    {"AutoIdentifiers": "default"},
    {"MediaLinks": {"width": 800, "height": 400}}
  ]
}
```

cc @filzrev


filzrev:
Thanks for updates.
I think this PR has made the config simpler and easier to understand.

Additionally, it might be possible to support a `remove' default extension with a `remove' parameter in the future.
(Discussed at https://github.com/dotnet/docfx/pull/7833)

yufeih:
> Additionally, it might be possible to support a remove' default extension with a remove' parameter in the future.
(Discussed at https://github.com/dotnet/docfx/pull/7833)

That can be perhaps be configured as `{ "extensionName": false }`

PR #9832, opened 2024-04-01
perf: Change serializer for XrefMap from NewtonsoftJson to System.Text.Json

filzrev:
This PR is successor of #9824.
Change XrefMap serializer from `Newtonsoft.Json` to `System.Text.Json` for perf optimization.

1. Add `SystemTextJsonUtility.cs` and related converter for generic JSON serialization/deserialization.
- This file is intended to be called from `JsonUtility.cs` wrapper class in future (By #9727).
   (And existing `JsonUtility.cs` logics will be moved to `NewtonsoftJsonUtility.cs`)

2. Refactor `XRefMapDownloader.cs`
  - Change class to `public sealed` class.
  - Change to async-based API.
  - Change virtual methods that is not referenced from external to private. 
  - Remove unused `UpdateHref` method.

3. Fix wrong `XRefMapRedirection.cs` Href attribute that introduced by #9824.


test/docfx.Tests/Api.verified.cs
```diff
@@ -2057,6 +2053,16 @@ public static class Build
             public const string EmptyInputFiles = "EmptyInputFiles";
         }
     }
+    public static class SystemTextJsonUtility
```
yufeih, line 2056: Can this be made internal?

src/Docfx.Common/Json/System.Text.Json/SystemTextJsonUtility.cs
```diff
@@ -0,0 +1,101 @@
...
+    /// </summary>
+    public static readonly JsonSerializerOptions IndentedSerializerOptions;
+
+    static SystemTextJsonUtility()
+    {
+        DefaultSerializerOptions = new JsonSerializerOptions()
+        {
+            // DefaultBufferSize = 1024 * 16, // TODO: Set appropriate buffer size based on benchmark.(Default: 16KB)
+            AllowTrailingCommas = true,
+            DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull | JsonIgnoreCondition.WhenWritingDefault,
```
yufeih, line 32: Feels a bit safer to only use `JsonIgnoreCondition.WhenWritingNull`. `JsonIgnoreCondition.WhenWritingDefault` can ignore default enum values which may not be desirable.

src/Docfx.Common/Json/System.Text.Json/SystemTextJsonUtility.cs
```diff
@@ -0,0 +1,101 @@
...
+            PropertyNameCaseInsensitive = true,
+            PropertyNamingPolicy = JsonNamingPolicy.CamelCase,
+            // DictionaryKeyPolicy = JsonNamingPolicy.CamelCase, // This setting is not compatible to `Newtonsoft.Json` serialize result.
+            NumberHandling = JsonNumberHandling.AllowReadingFromString,
+            Converters =
+            {
+                new JsonStringEnumConverter(),
+                new ObjectToInferredTypesConverter(), // Required for `Dictionary<string, object>` type deserialization.
```
yufeih, line 40: Can we change `Dictionary<string, object> Others` -> `Dictionary<string, JsonElement> Others` if that is the only place using this converter?

test/docfx.Tests/Api.verified.cs
filzrev: I've changed to internal with commit.(https://github.com/dotnet/docfx/pull/9832/commits/8a3be9460c9800988b2727032f947f7b672ec90e)

InternalVisibleTo attributes are temporary required to test both SystemTextJson/NewtonsoftJson.

src/Docfx.Common/Json/System.Text.Json/SystemTextJsonUtility.cs
filzrev: This setting intended to align options used by following code.
(Because `JsonSerializerOptions` instance is recommended to be shared when same option used)
https://github.com/dotnet/docfx/blob/b7285fcdfb04e6f4132d702217588de6236bfa4c/src/Docfx.Build/ApiPage/ApiPage.cs#L149

As you noted.
It's seems better to use only `WhenWritingNull`. I've removed `WhenWritingDefault` (https://github.com/dotnet/docfx/pull/9832/commits/194506188c8d0b86a9d54c09173bced59fb04853)

filzrev:
I've failed to rebase to latest main branch.
and other PR commit are displayed as diffs.

For clear diffs for code review. I'll recreate this PR later.
And fixing following remaining review comment.

> Can we change Dictionary<string, object> Others -> Dictionary<string, JsonElement> Others if that is the only place using this converter?

PR #9840, opened 2024-04-06
chore: Add `InternalVisibleTo` attribute for benchmark project

filzrev:
This PR add `InternalsVisibleTo` attribute for benchmark projects.

I'll planning to add benchmark project to `benchmarks/Docfx.Benchmarks` in separated PR.
after this PR content is published to NuGet. 

**Previous PR discussion about `InternalsVisibleTo` attributes**
https://github.com/dotnet/docfx/pull/9652

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -3,5 +3,6 @@
 
   <ItemGroup>
     <InternalsVisibleTo Include="$(AssemblyName).Tests" />
+    <InternalsVisibleTo Include="Docfx.Benchmarks" />
   </ItemGroup>
 </Project>


yufeih:
Is there a reason why we need to publish to NuGet first before adding the benchmark project?

filzrev:
> Is there a reason why we need to publish to NuGet first before adding the benchmark project?

It's because I want to write benchmarks that targeting following two versions in single benchmark.
1. NuGet published version (`Baseline`)
2. Local source build version

Note: 
It's not tested yet, that it can referencing same assemblies with different versions aliases (by referencing using `extern alias`).
If it's not possible. It need to switch two version by using `csproj` settings.

And it's also can be used to compare multiple NuGet versions benchmark.([Example](https://github.com/dotnet/BenchmarkDotNet/blob/a24d689361c44356b12892424d25bdf27d13d8da/samples/BenchmarkDotNet.Samples/IntroNuGet.cs#L16))

yufeih:
Make sense

yufeih (approved):

PR #9850, opened 2024-04-08
feat: Enable PreferCSSPageSize option for PDF generation

### LINKED ISSUE #9842: [Feature Request] Print PDF's in landscape orientation and using specific page size
How difficult will it be to extend to the existing PDF printing logic to specify the orientation and page size of the page that should be used?  Thinking of adding `pdfOrientation: landscape | portrait`  setting.  If not specified it should default to `portrait`.  For size, using something like `pdfPageSize:  A0 | A1 .. | Letter `   

If possible, any pointers where to start looking at adding the support to the existing code base?

filzrev:
This PR intended to fix #9842.

By default, docfx generating PDF files with following settings. 
- `PageSize`:  Letter (215.9 x 279.4) 
- `PageOrientation`: Portrait

This PR enable PagePdfOptions's `PreferCSSPageSize` settings.
So It can customize PDF option by custom CSS. 

**Example CSS settings**
```
@media print {
  @page {
    size: A4 landscape
  }
}
```

### DIFF
--- a/src/Docfx.App/PdfBuilder.cs
+++ b/src/Docfx.App/PdfBuilder.cs
@@ -166,7 +166,11 @@ static class PdfBuilder
                     Logger.LogWarning($"Timeout waiting for page to load, generated PDF page may be incomplete: {url}");
                 }
 
-                return await page.PdfAsync(new PagePdfOptions { PrintBackground = outline.pdfPrintBackground });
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Thank you @filzrev !

PR #9851, opened 2024-04-08
fix: PDF `Producer` document information

### LINKED ISSUE #9843: [Bug] PDF producer's `version` information is missing on PDF that generated by `docfx pdf` command
**Describe the bug**
PDF `producer` property is not written as expected.

**To Reproduce**
1. Open PDF file generated by docfx.
2. Confirm PDF document property.
3. `docfx ()` text is displayed as `producer`.

**Expected behavior**
PDF generator exe name with version information is displayed (e.g. 'docfx (2.76.0)`)

**Additional context**
Currently `producer` property is set on following line.
https://github.com/dotnet/docfx/blob/464b947169eee016e4ca166cbae3f6a82f6178eb/src/Docfx.App/PdfBuilder.cs#L273

But `AssemblyVersionAttribute` is missing on `Docfx.App` assembly.

It might be better to use name of `Assembly.GetEntryAssembly().Name` as producer name.
And use `AssemblyFileVersionAttribute` version instead.

filzrev:
This PR intended to fix #9843.

Currently PDF files that generated by docfx has wrong `Producer` document information.
(It's resolved to `docfx ()` text always).

This PR add following changes to set `Producer` document information.
- ~~Get application name from EntryAssembly's name. (for custom application that use `Docfx.App` package)~~
- Get assembly version from `AssemblyFileVersionAttribute`  information.

By this PR changes.
PDF `Producer` document information information is displayed as `docfx (2.76.0.0)`

### DIFF
--- a/src/Docfx.App/PdfBuilder.cs
+++ b/src/Docfx.App/PdfBuilder.cs
@@ -270,7 +270,7 @@ static class PdfBuilder
         if (numberOfPages is 0)
             return;
 
-        var producer = $"docfx ({typeof(PdfBuilder).Assembly.GetCustomAttribute<AssemblyVersionAttribute>()?.Version})";
+        var producer = $"docfx ({typeof(PdfBuilder).Assembly.GetCustomAttribute<AssemblyFileVersionAttribute>()?.Version})";
 
         using var output = File.Create(outputPath);
         using var builder = new PdfDocumentBuilder(output);


src/Docfx.App/PdfBuilder.cs
```diff
@@ -270,7 +270,12 @@ static string ExpandTemplate(string? pdfTemplate, int pageNumber, int totalPages
         if (numberOfPages is 0)
             return;
 
-        var producer = $"docfx ({typeof(PdfBuilder).Assembly.GetCustomAttribute<AssemblyVersionAttribute>()?.Version})";
+        var entryAssembly = Assembly.GetEntryAssembly() ?? typeof(PdfBuilder).Assembly;
```
yufeih, line 273: Do we want to use entry assembly if it is called using `Docfx.App` package?

filzrev: > Do we want to use entry assembly if it is called using Docfx.App package?
There is no particular reason to stick to it.
It just curious about showing `Docfx.App` assembly file version as `docfx` version.

I'll fix new commit about this.

yufeih (approved):
Thank you @filzrev !

PR #9858, opened 2024-04-09
chore: Skip unstable PlantUML tests

filzrev:
This PR intended to temporary skip `PlantUmlTest.TestRenderSvg_SequenceDiagram` tests.
Because this test **randomly** fails with following error.

```
2024-04-09T07:46:33.2553752Z    Assert.Equal() Failure: Strings differ
2024-04-09T07:46:33.2554761Z                                   ↓ (pos 251)
2024-04-09T07:46:33.2555876Z Expected: ···"none" style="width:113px;height:120px;bac"···
2024-04-09T07:46:33.2557067Z Actual:   ···"none" style="width:109px;height:120px;bac"···
2024-04-09T07:46:33.2558032Z                                   ↑ (pos 251)
```

### DIFF
--- a/test/Docfx.MarkdigEngine.Extensions.Tests/PlantUmlTest.cs
+++ b/test/Docfx.MarkdigEngine.Extensions.Tests/PlantUmlTest.cs
@@ -7,7 +7,7 @@ namespace Docfx.MarkdigEngine.Tests;
 
 public class PlantUmlTest
 {
-    [Fact]
+    [Fact(Skip = "Unstable test")]
     public void TestRenderSvg_SequenceDiagram()
     {
         var source = """
--- a/test/Docfx.MarkdigEngine.Extensions.Tests/PlantUmlTest.cs
+++ b/test/Docfx.MarkdigEngine.Extensions.Tests/PlantUmlTest.cs
@@ -1,13 +1,14 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using FluentAssertions;
 using Xunit;
 
 namespace Docfx.MarkdigEngine.Tests;
 
 public class PlantUmlTest
 {
-    [Fact(Skip = "Unstable test")]
+    [Fact]
     public void TestRenderSvg_SequenceDiagram()
     {
         var source = """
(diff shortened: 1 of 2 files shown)


filzrev:
When using `docker version` of PlantUML Server.
It's return consistent SVG contents (`width:109px`).

So it seems to be `https://www.plantuml.com/plantuml` side problems.
It seems HTTP response header always returns `X-Powered-By: PlantUML Version 1.2024.4` in both type of response.
Though returned SVG content is inconsistent.

Anyway external service dependencies within unit tests can cause unintended test failures. (e.g. timeout, server busy)
So it's reasonable to merge this PR to skip this test. (To avoid CI error on other PR)

test/Docfx.MarkdigEngine.Extensions.Tests/PlantUmlTest.cs
```diff
@@ -7,7 +7,7 @@ namespace Docfx.MarkdigEngine.Tests;
 
 public class PlantUmlTest
 {
-    [Fact]
+    [Fact(Skip = "Unstable test")]
```
yufeih, line 10: instead of skipping, can we relax the assertion to say that we expect the result to start with `<div class="lang-plantUml"><?xml version="1.0" encoding="us-ascii" standalone="no"?><svg` and end with `</svg></div>`?

filzrev: OK. I'll trying to fix assertions.

filzrev: It seems a bit complicated to reuse existing `TestUtility.VerifyMarkup` to test startWith/endsWith.

How about temporary merge this PR.
And revert changes after the PlantUML Online server response is stabilized?

yufeih: Make `VerifyMarkup` return `actualHtml` and pass `html: null` and verify the result?

filzrev: If it's acceptable to change existing `VerifyMarkup` method signature.
It can easily verify markup result,

I've created commit by https://github.com/dotnet/docfx/pull/9858/commits/ac73209c5bd111f7d064ad66547ee10db865390d

yufeih (approved):
Thank you @filzrev !

Issue #9866, opened 2024-04-12
[Bug] Markdown xref broken for external xrefmap.yml


glopesdev:
**This is a regression in 2.76.0. Confirmed that the same reference resolves fine in 2.75.3 so we are downgrading for the time being.**

Markdown cross-references in articles pointing to UIDs defined in external `xrefmap.yml` files are resolved as if they are internal API references. For example:

`[OtherClass](xref:External.Namespace.OtherClass)` will resolve to `mydomain/api/External.Namespace.OtherClass.html` instead of `externaldomain/api/External.Namespace.OtherClass.html`.

**To Reproduce**
1. Add `xref` section to `docfx.json` containing a URL to a `xrefmap.yml` in another domain.
2. Add a markdown cross-reference to a UID in the external YML file.

**Expected behavior**
Cross-references should resolve to links pointing to the external website.

**Context (please complete the following information):**
- OS: Windows
- Docfx version: [e.g. 2.76.0]

**Additional context**
Downgrading to 2.75.3 fixes the problem.

filzrev:
@glopesdev
Are you using `redirection` entry in `xrefmap.json`?.
If so. it'll be fixed by #9832.  (`Fix wrong XRefMapRedirection.cs Href attribute that introduced`)

It's expected be released same timing with #9824 changes.
But it's not merged yet. (Because I've mixed bugfix in perf optimization PR)

If it's not using `redirection` entry.
Please provide example xrefmap that cause problems.

glopesdev:
@filzrev this doesn't use redirection (I didn't even know this existed until now), both projects use mostly default docfx settings.

Example xrefmap: https://horizongir.github.io/ZedGraph/xrefmap.yml

filzrev:
@glopesdev 
Where is the `externaldomain` setting is defined?

It seems xrefmap define `href` as relative URL. and no `baseUrl` setting is defined.
So it's expected behaviors that are resolved to `mydomain` based URL.

glopesdev:
@filzrev I have updated the issue to clarify that `externaldomain` is part of the example resolved URL, and not a setting. The external websites and `xrefmap.yml` were generated by DocFX built with default settings.

> So it's expected behaviors that are resolved to mydomain based URL.

I am not sure I understand the rationale for this. These are not local `xrefmap.yml` files, they are added as external additional references using a URL which is outside of the domain, so why should they be assumed to be part of the local project?

[...]

PR #9869, opened 2024-04-13
fix: Xrefmap baseUrl problem reported at #9866

filzrev:
This PR intended to fix bug reported at #9866.

It's a regression that introduced by PR #9721

**Current behaviors**
If following conditions are met. 
`<xref:>` link is resolved based on current site. It should be resolved to xrefmap download file based path.
- xrefmap is specified with URL.
- Xrefmap don't define baseUrl. And using `relative URL` for href.

**What's changed in this PR**

1. `XRefArchiveBuilder`
  1.1. Remove `baseUrl` set logics. Because it's set at downloader by this PR changes.
  1.2.  `map.UpdateHref` call is also removed.  Because it's expected to be called by [BasicXRefMapReader]( https://github.com/dotnet/docfx/blob/main/src/Docfx.Build/XRefMaps/BasicXRefMapReader.cs#L17-L24)
  1.3. By this change. downloaded `xrefmap.zip` content are changed. (href's relative url are preserved as is)
2. `XrefArchiveDownloader`
  2.1. Add logics to set `baseUrl`  if baseUrl is not defined.
  2.2. Remove `UpdateHref` method. (It's not used by docfx)
3. Add additional XrefmapDownloader tests with following patterns
  3.1 `baseUrl` is not defined, and href is relative URL (In this case. `baseUrl` is used to resolve relative url)
  3.2. `baseUrl` is not defined, and href is absolute URL (In this case. absolute URL is used.)

Note: `docfx download` related command tests are executed manually.

### DIFF
--- a/src/Docfx.Build/XRefMaps/XRefArchiveBuilder.cs
+++ b/src/Docfx.Build/XRefMaps/XRefArchiveBuilder.cs
@@ -47,16 +47,6 @@ public class XRefArchiveBuilder
             return null;
         }
 
-        // If BaseUrl is not set. Use xrefmap file download url as basePath.
-        if (string.IsNullOrEmpty(map.BaseUrl))
-        {
-            var baseUrl = uri.GetLeftPart(UriPartial.Path);
-            baseUrl = baseUrl.Substring(0, baseUrl.LastIndexOf('/') + 1);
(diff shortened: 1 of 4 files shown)


yufeih (approved):
Thank you @filzrev !

PR #9872, opened 2024-04-14
perf: Change serializer for XrefMap from NewtonsoftJson to System.Text.Json

filzrev:
This PR is recreated instead of #9832.

This PR contains cherry picked commit from above PR.
And resolve conflicts that are occurred at following codes.
-  `DownloadFromWebAsync` method of `XRefMapDownloader.cs`
-  `Api.verified.cs`.

Following review comment is remaining that need to be resolved.
 ('https://github.com/dotnet/docfx/pull/9832#discussion_r1561946897)

>Can we change Dictionary<string, object> Others -> Dictionary<string, JsonElement> Others if that is the only place using this converter?

### DIFF
--- a/src/Docfx.Build/XRefMaps/XRefMapDownloader.cs
+++ b/src/Docfx.Build/XRefMaps/XRefMapDownloader.cs
@@ -10,7 +10,7 @@ using EnvironmentVariables = Docfx.DataContracts.Common.Constants.EnvironmentVar
 
 namespace Docfx.Build.Engine;
 
-public class XRefMapDownloader
+public sealed class XRefMapDownloader
 {
     private readonly SemaphoreSlim _semaphore;
     private readonly IReadOnlyList<string> _localFileFolders;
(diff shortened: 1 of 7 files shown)


filzrev:
> Can we change `Dictionary<string, object>` Others -> `Dictionary<string, JsonElement>` Others if that is the only place using this converter?

This property is used by `YamlDotnet` also. so it can't simply changed to JsonElement.

---

I also tried to create `JsonConverter<Dictionary<string, object>>` for this type (That similar to `JObjectDictionaryToObjectDictionaryConverter`)

But it seems not possible to use custom JsonConverter with  `[JsonExtensionData] attribute.

Because when using `JsonExtensionData`.
It can't use custom JsonConverter. (serialization works as expected, but deserialize don't use custom converter)
https://github.com/dotnet/runtime/issues/60560

yufeih (approved):
Thank you @filzrev !

PR #9909, opened 2024-05-08
feat: Enable view transitions feature

filzrev:
This PR intended to support ViewTransition for modern template.

**Background**
Chrome 126 (It'll be released next month) supports Multi-Page-Application(MPA) ViewTransition by default.
- https://chromestatus.com/roadmap
- [View Transitions Same-Origin Navigation](https://chromestatus.com/feature/5118874666663936)

This PR enable view transition by adding related CSS styles
that based on https://github.com/WICG/view-transitions/blob/main/cross-doc-explainer.md#declarative-opt-in

By enabling MPA view transition.
TOC/breadcrumb regions are more smoothly transitioned when navigating between pages.

**How to test ViewTransition features**
- Manually enable `chrome://flags/#view-transition-on-navigation` feature flags or.
- Use [Chrome Canary](https://www.google.com/intl/en/chrome/canary/) build.

PR #9910, opened 2024-05-09
chore: Add detailed error message for specific exceptions

filzrev:
This PR intended to get detailed error messages for https://github.com/dotnet/docfx/pull/9909#issuecomment-2102880480

It's expected to be caused by dependent libraries that similar to #9746.
And it's hard to reproducing problems.
So add detailed error logs by this PR.


filzrev:
Close this PR because it can reproduce on local environment.

This error is occurred when using JsonSchema.Net `7.x`.
See: `https://github.com/gregsdennis/json-everything/issues/726`

Issue #9745, opened 2024-02-28
[Feature Request] Support building with preview versions of the .NET SDK


martincostello:
**Is your feature request related to a problem? Please describe.**

Testing applications out with the .NET 9 Preview 1 SDK, it's not possible to build sites using docfx that have a `net9.0` target framework.

This appears to be due to the code below, which requires that the SDK used is the same version as the runtime version docfx uses.

https://github.com/dotnet/docfx/blob/bebcd35323da35e253125c5b3cfd12e7a4553696/src/Docfx.Dotnet/CompilationHelper.cs#L135-L159

Even if the .NET 8 SDK is installed, in addition to .NET 9's, and `DOTNET_ROLLFORWARD=Major` is set, docfx will still and try to use the .NET 8 SDK to build for .NET 9, which fails.

```console
Using .NET Core SDK 8.0.200
Loading project 
/runner/_work/MyProject/MyProject/src/MyProject/MyProject.csproj
  Determining projects to restore...
Error: /usr/share/dotnet/sdk/8.0.200/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.TargetFrameworkInference.targets(166,5): error NETSDK1045: The current .NET SDK does not support targeting .NET 9.0.  Either target .NET 8.0 or lower, or use a version of the .NET SDK that supports .NET 9.0. Download the .NET SDK from https://aka.ms/dotnet/download [/runner/_work/MyProject/MyProject/src/MyProject/MyProject.csproj::TargetFramework=net9.0]
```

**Describe the solution you'd like**

docfx allows the use of newer .NET SDKs if installed in the execution environment.

**Describe alternatives you've considered**

Wait until November until .NET 9 is released and ignore my failing documentation builds.

**Additional context**

None.

martincostello:
It's a shame that roll-forward isn't causing that behaviour to change - that's often a good-enough workaround when I test pre-release versions of .NET where I use global tools that don't yet support them.

filzrev:
If `DOTNET_ROLLFORWARD=Major` is specified.
Runtime will not be rolled forward if the target runtime is already installed.
https://learn.microsoft.com/en-us/dotnet/core/versions/selection#values

To force to use .NET9 runtime.
You must set the following environment variables.
- `DOTNET_ROLL_FORWARD=LatestMajor`
- `DOTNET_ROLL_FORWARD_TO_PRERELEASE=1`

martincostello:
[...]

PR #9911, opened 2024-05-10
chore: Drop .NET 7 SDK supports

filzrev:

- Drop `.NET 7 SDK` supports. (Supports ends at May 14, 2024)
- Add `.NET 9 SDK preview` supports to nightly build.  
   (It's relating to #9745)
- Update `dotnet-sign` tool version
- Update related docs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,9 +32,6 @@ jobs:
 
     - run: dotnet test -c Release -f net8.0 --no-build --collect:"XPlat Code Coverage"
 
-    - run: dotnet test -c Release -f net7.0 --no-build --collect:"XPlat Code Coverage"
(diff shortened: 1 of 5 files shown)


filzrev:
It seems PR that relating GitHub Actions can't test with CI until merged to main.
Because CI triggers are changed to  `pull_request_target`.

yufeih (approved):
Looks great!

filzrev:
> [!NOTE]
> When using `DOCFX_PREVIEW_BUILD` environment variable on `non-CI` environment.
> It need to execute `dotnet restore` command **explicitly** before running build.
>
> Otherwise, It cause build error that relating `targetframework` mismatch.
> It seems `obj/project.assets.json` file is not updated by **implicit restore** that executed on `dotnet build` command.

PR #9914, opened 2024-05-12
chore: Add detailed error log message for specific exception type

filzrev:
This PR is successor of #9910.

On [latest CI build](https://github.com/dotnet/docfx/actions/runs/9050786668)
Another multi-threading related exception (`InvalidOperationException`) is recorded.

To capture detailed logs. Add  `InvalidOperationException` to detailed log targets. 

And multi-thread related errors may occur in the future when updating dependent libraries.
So add also `IndexOutOfRangeException` types as detailed logging target.

**Exception message**

>D:\a\docfx\docfx\docs\extensions\packages.yml: error InvalidInputFile: Unable to load file 'extensions/packages.yml' via processor 'Dashboard': Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.

### DIFF
--- a/src/Docfx.Build/HostServiceCreator.cs
+++ b/src/Docfx.Build/HostServiceCreator.cs
@@ -63,7 +63,8 @@ class HostServiceCreator
             {
                 if (e is not DocumentException)
                 {
-                    var message = e is ArgumentException or NullReferenceException
+                    // Log detailed exception messages for specific type of exception.
+                    var message = e is ArgumentException or NullReferenceException or IndexOutOfRangeException or InvalidOperationException
                         ? e.ToString()
                         : e.Message;
 


src/Docfx.Build/HostServiceCreator.cs
```diff
@@ -63,7 +63,8 @@ public HostServiceCreator(DocumentBuildContext context)
             {
                 if (e is not DocumentException)
                 {
-                    var message = e is ArgumentException or NullReferenceException
+                    // Log detailed exception messages for specific type of exception.
```
yufeih, line 66: Consider for all exception types:
1. Log an error message with `e.Message`
2. Log a verbose message with `e.ToString()`.

filzrev:
Multi-thread concurrency issue seems to be resolved by `JsonSchema.Net` 7.0.2.

I'll create separate PR to update library and remove temporary detailed logging code for diagnostics.

Issue #9941, opened 2024-05-22
[Bug] .NET API bug - Nullable project throws error telling I should enable nullable


gumbarros:
**Describe the bug**
`The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.` error is throw at GitHub actions when generating API reference.

**To Reproduce**
Just `docfx docfx.json` at GitHub Actions or `docfx metadata`

**Expected behavior**
API should be generated.

**Context (please complete the following information):**
- OS: Windows latest at GitHub actions
- Docfx version: 2.76.0
- .NET version: .NET 8

- `docfx.json` config
https://github.com/ncalc/ncalc/blob/master/docs/docfx.json

- Exceptions
https://github.com/ncalc/ncalc/actions/runs/9183960851/job/25255512510

- Errors and warnings
https://github.com/ncalc/ncalc/actions/runs/9183960851/job/25255512510

- .NET info
Not applicable, CI/CD.

**Additional context**
The error only happens at GitHub Actions.

filzrev:
It seems following warnings are logged before CS8632 errors.

> warning: [Failure] Msbuild failed when processing the file 'D:\a\ncalc\ncalc\src\NCalc\NCalc.csproj' with message: The "Nerdbank.GitVersioning.Tasks.GetBuildVersion" task failed unexpectedly.
Nerdbank.GitVersioning.GitException: Shallow clone lacks the objects required to calculate version height. Use full clones or clones with a history at least as deep as the last version height resetting change.

And by adding [`fetch-depth: 0` options](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/cloudbuild.md#github-actions) to `actions/checkout action`. CS8632  errors seems resolved.

```
      - name: Checkout
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
```

---
Another problem I've noticed is that the `docfx metadata` command returns `-1` as an exit code.
But next line of script is executed.

It's because GitHub windows runner use `pwsh` as default shell.
And pwsh don't stop on native command error by default. 
(It need to set `PSNativeCommandUseErrorActionPreference = $true` to change behaviors)

gumbarros:
I also tried to using Linux before Windows and had the same error, except Linux completely broke my pipeline, Windows at least publish my articles

gumbarros:
Weird, adding `fetch-depth: 0` solved the problem, thank you very much. Feel free to close the issue if it's expected behavior

PR #9945, opened 2024-05-22
chore: Add error handling logics

filzrev:
This PR intended to fix issue reported by #9941.

In the current implementation.
When `WorkspaceDiagnosticKind.Failure` error occurred when calling [MSbuildWorkSpace::OpenProjectAsync](https://github.com/dotnet/docfx/blob/12ddaf6e53a193238eb8172b3dbec2ddecaf17a5/src/Docfx.Dotnet/DotnetApiCatalog.Compile.cs#L150)
Code analysis is interrupted and it's recorded as **warning** at following callback method.
https://github.com/dotnet/docfx/blob/12ddaf6e53a193238eb8172b3dbec2ddecaf17a5/src/Docfx.Dotnet/DotnetApiCatalog.Compile.cs#L44

It’s confusing that root cause of issue is reported as **warning**.
So this PR changes this behavior to record `WorkspaceDiagnosticKind.Failure` as an error. (Only when AllowCompilationErrors is not specified)

---

Additionally. This PR add `fail-fast` logics to `DefaultCommand`.
(Currently build/serve command is executed even if an error has occurred in the previous process.

### DIFF
--- a/src/Docfx.Dotnet/DotnetApiCatalog.Compile.cs
+++ b/src/Docfx.Dotnet/DotnetApiCatalog.Compile.cs
@@ -41,7 +41,19 @@ partial class DotnetApiCatalog
         });
 
         using var workspace = MSBuildWorkspace.Create(msbuildProperties);
-        workspace.WorkspaceFailed += (sender, e) => Logger.LogWarning($"{e.Diagnostic}");
+        workspace.WorkspaceFailed += (sender, e) =>
+        {
+            var message = e.Diagnostic.ToString();
+            switch (e.Diagnostic.Kind)
+            {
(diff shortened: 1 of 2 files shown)


yufeih:
> It’s confusing that root cause of issue is reported as warning.

This may result in warnings be reported as errors due to: https://github.com/dotnet/roslyn/pull/66710

filzrev:
> This may result in warnings be reported as errors due to `https://github.com/dotnet/roslyn/pull/66710`

Thanks for indicating related Roslyn issue.

Close this PR.
Because it can't be fixed until related roslyn PR is merged.
And `fail-fast` logics for `DefaultCommand` isn't important feature.

PR #9957, opened 2024-05-27
feat(xml): better error when link not found

Patrick8639:
Hi!

When a link is not correct in an XML comment, the error message doesn't show where the error is.
This change adds more information in this case.


src/Docfx.Dotnet/Parsers/XmlComment.cs
```diff
@@ -376,7 +376,17 @@ private void ResolveCrefLink(XNode node, string nodeSelector, Action<string, str
...
+                {
+                    var memberName = (string) doc.Element("member")?.Attribute("name");
+
+                    if (! string.IsNullOrEmpty(memberName))
+                    {
+                        detailedInfo.Append(", member name is ");
+                        detailedInfo.Append(memberName);
+                    }                }
+
+                Logger.Log(LogLevel.Warning, $"Invalid cref value \"{cref}\" found in XML documentation comment{detailedInfo}.", code: "InvalidCref");
```
yufeih, line 389: The whitespace between `comment` and `{detailedInfo}` is removed, will this change the output if condition on line 379 was not hit?

src/Docfx.Dotnet/Parsers/XmlComment.cs
```diff
@@ -376,7 +376,17 @@ private void ResolveCrefLink(XNode node, string nodeSelector, Action<string, str
...
-                Logger.Log(LogLevel.Warning, $"Invalid cref value \"{cref}\" found in XML documentation comment {detailedInfo}.", code: "InvalidCref");
+                if (detailedInfo.Length == 0 && node is XDocument doc)
+                {
+                    var memberName = (string) doc.Element("member")?.Attribute("name");
+
+                    if (! string.IsNullOrEmpty(memberName))
+                    {
+                        detailedInfo.Append(", member name is ");
+                        detailedInfo.Append(memberName);
+                    }                }
```
yufeih, line 387: Please move the second closing bracket to the next line and intend properly.

src/Docfx.Dotnet/Parsers/XmlComment.cs
Patrick8639: If you look at lines 367 and 372, you with see that `detailedInfo` always starts with a space, which leads to 2 spaces in the message.
Removing the space in line 389 just removes the second space.

yufeih (approved):
Thank you @Patrick8639

PR #9960, opened 2024-05-28
chore: Add PolySharp libarary to use latest C# syntax

filzrev:
This PR add [PolySharp](https://github.com/Sergio0694/PolySharp) package reference to projects.

**Background**

Currently docfx use multi target frameworks `net6` and `net8`.
So some latest C# feature usage is limited. (e.g. [required modifier](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/required))

By this changes. 
It can use latest C# 12 features when targeting `.NET 6` runtime.  
And it can remove `#ifdef` from metadata generation when using `markdown`/`apiPage` output format.

---

PolySharp generate source-only polyfills. So runtime library dependency is not needed.
The only concern is the increase of build time due to running source generator.
But it's acceptable increase rate.

**CI build time**

| OS | Original | [This PR CI](https://github.com/dotnet/docfx/actions/runs/9263366790)
|----|----|-----|
| Windows| 10m18s | 9m2s
| Ubuntu    |11m10s | 12m12s
| macOS     | 6m 33s | 6m41s

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -46,6 +46,13 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="PolySharp">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>
+
+  <ItemGroup>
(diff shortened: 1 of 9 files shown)


yufeih (approved):
Wow, this is so great to see! Thank you @filzrev

PR #9966, opened 2024-05-30
feat: Add support for gzipped xrefmap that stored as local file

filzrev:
This PR add support for `gzip compressed` xrefmap file.

**Background**

`xrefmap` files can exceed 100 MB in size in some cases.
For example. [.NET API's xrefmap](https://github.com/dotnet/docfx/blob/main/.xrefmap.json) takes `132MB`. (And `8.5`MB when using gzip compression)

And when specifying xrefmap with URL schema.
docfx download xrefmap file on every build.

So it's recommended to use pre downloaded local file for large xrefmap.
But in this case, xrefmap can't be stored on GitHub repository. (By default it's limited to50MB and It requires git LFS for larger file)

And currently there is zipped xrefmap support that compressed by `zip`.
But it support only `yml` format that is created by `dotnet download` command.
And yml file processing run 4x slower than json.

This PR add support for gzipped xrefmap file to resolve described above issue.


yufeih:
will an `.xrefmap.json.gz` be uploaded in a separate PR?

filzrev:
> will a `.xrefmap.json.gz` be uploaded in a separate PR?

This PR intended to support gzipped files that saved on **local** files.
For network files. gzip support is not implemented.
Because it should use `HTTP Compression` if available.

For `https://github.com/dotnet/docfx/raw/main/.xrefmap.json` file. 
It's stored on Git LFS. and don't support  `HTTP Compression`. (https://github.com/dotnet/docfx/issues/9659#issuecomment-1914522207)

In this case. 
It need manually download xrefmap and compress as gzip.

Alternatively. It can use xrefmap provided by `learn.microsoft.com`  (Though It might not officially supported)
> https://learn.microsoft.com/en-us/dotnet/.xrefmap.json

yufeih (approved):
Looks good

PR #9972, opened 2024-05-31
Feat/private symbols

Patrick8639:
When specifying `"includePrivateMembers" : "true"` in `docfx.js`, the private classes of an assembly are documented, but not the private members of a class.

This fix, modified how the members are enumerated, in order to include the private members of a class, if needed.
Unfortunately, the method used to enumerate the members, using Roslyn, doesn't provide a way to enumerate also the private, so I have to set a flag using reflection to do this.

To be correct, code has also been added to not document the class constructor, if any, or compiler-generated members.

### DIFF
--- a/src/Docfx.Dotnet/CompilationHelper.cs
+++ b/src/Docfx.Dotnet/CompilationHelper.cs
@@ -1,6 +1,8 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+
+using System.Reflection;
 using Docfx.Common;
 using Docfx.Exceptions;
 using ICSharpCode.Decompiler.Metadata;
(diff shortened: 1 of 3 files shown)


src/Docfx.Dotnet/CompilationHelper.cs
```diff
@@ -102,12 +104,22 @@ public static Compilation CreateCompilationFromVBCode(string code, IDictionary<s
             references: GetDefaultMetadataReferences("VB").Concat(references ?? []));
     }
 
-    public static (Compilation, IAssemblySymbol) CreateCompilationFromAssembly(string assemblyPath, params MetadataReference[] references)
+    public static (Compilation, IAssemblySymbol) CreateCompilationFromAssembly(string assemblyPath, ExtractMetadataConfig? config = null, params MetadataReference[] references)
     {
+        var options = new CS.CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary);
```
filzrev, line 109: It seems there is constructor that accept `metadataImportOptions` named parameter.
Is it able to use this overload instead of reflection?

Patrick8639: Good catch, I missed it!

yufeih (approved):
Thank you @Patrick8639 .

PR #9974, opened 2024-05-31
chore: Fix exception thrown on TOC loading internally

filzrev:
This PR is intended to fix internal exception that occurred when deserializing `toc.yml` file.

Current docfx implementation use `YamlDeserializerWithFallback` class to deserialize `toc.yml` file.
And following operation executed.

1. Try to deserialize yml as `List<TocItemViewModel>`.
2. If YamlException is thrown. then deserialize as `TocItemViewModel`.

---

This PR intended to fix following problems.
- On current implementation `try-catch` logics are used when deserializing `TocItemViewModel` as `List<TocItemViewModel>` .  
   -> Exceptions should not be thrown in the normal execution flow.
- When deserializing as `TocItemViewModel`. It requires to open safe file twice.


src/Docfx.Build/TableOfContents/TocHelper.cs
```diff
@@ -86,7 +86,45 @@ public static TocItemViewModel LoadSingleToc(string file)
             Logger.LogError(message, code: ErrorCodes.Toc.InvalidTocFile);
             throw new DocumentException(message, e);
         }
+    }
+
+    private static bool IsListTocItems(StreamReader reader)
```
yufeih, line 91: This looks a bit complicated. For handling YAML, I'm leaning towards we move to a future where YAML are parsed into a JSON node tree, and then deserialized to objects with standard JsonConverter.

Similar to https://github.com/dotnet/docfx/blob/v3/src/docfx/lib/json/YamlUtility.cs#L112 but with System.Text.Json.

filzrev: > This looks a bit complicated.
Remove stream relating unwind logics.
Instead, Read **entire YAML** text first, and then determine TOC type by YAML Parser.

vicancy (approved):

PR #9985, opened 2024-06-04
Feat/source code enhancements

Patrick8639:
Source-code enhancements:
- new C# syntax;
- static lambda;
- remove not used local variables;
- remove implicit calls to base class;
- use base class for static method, when needed.

This is a first batch of corrections, more to come (still 2159 warnings...)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,7 +15,7 @@
     <PackageVersion Include="Newtonsoft.Json" Version="13.0.3" />
     <PackageVersion Include="OneOf" Version="3.0.271" />
     <PackageVersion Include="OneOf.SourceGenerator" Version="3.0.271" />
-    <PackageVersion Include="PdfPig" Version="0.1.9-alpha-20240601-65a18" />
+    <PackageVersion Include="PdfPig" Version="0.1.9-alpha-20240609-affc1" />
     <PackageVersion Include="PlantUml.Net" Version="1.4.80" />
     <PackageVersion Include="PolySharp" Version="1.14.1" />
     <PackageVersion Include="Spectre.Console" Version="0.49.1" />


yufeih:
@Patrick8639  build failed.

src/Docfx.Build.Common/ModelAttributeHandlers/HandleModelAttributesContext.cs
```diff
@@ -8,17 +8,17 @@ namespace Docfx.Build.Common;
...
-    public bool SkipMarkup { get; set; }
-    public bool EnableContentPlaceholder { get; set; }
-    public string PlaceholderContent { get; set; }
-    public bool ContainsPlaceholder { get; set; }
-    public HashSet<string> Dependency { get; set; } = new HashSet<string>();
-    public FileAndType FileAndType { get; set; }
-    public HashSet<string> LinkToFiles { get; set; } = new HashSet<string>(FilePathComparer.OSPlatformSensitiveStringComparer);
-    public HashSet<string> LinkToUids { get; set; } = new HashSet<string>();
-    public List<UidDefinition> Uids { get; set; } = new List<UidDefinition>();
+    internal int                 NestedLevel              { get; set; } = 0;
```
yufeih, line 11: Consider disable adding spaces to align property names. The rest of the codebase does not follow this style pattern.

src/Docfx.App/Helpers/DocumentBuilderWrapper.cs
```diff
@@ -134,11 +134,11 @@ private static List<DocumentBuildParameters> ConfigToParameter(BuildJsonConfig c
 
         if (config.PostProcessors != null)
         {
-            parameters.PostProcessors = config.PostProcessors.ToImmutableArray();
+            parameters.PostProcessors = [..config.PostProcessors];
```
filzrev, line 137: Personally, I though CodeFixer rule for [IDE0305](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0305) should not be applied unconditionally

Because it decrease readability.
And no perf improvement by using collection initializer syntax.
(See: `https://github.com/dotnet/roslyn/issues/70656`)

Patrick8639:
Hello @yufeih and @filzrev ,
Thank you for your comments.

I will rebase to main and start again, taking into account the remarks.

What do you think would be the better to push the changes?
- One commit / pull request per file.
- One commit / pull request per type of change.

> Consider disable adding spaces to align property names. The rest of the codebase does not follow this style pattern.

Yes, I will take attention to this, even if I think that the code is more readable when the `=` are aligned...

PR #9986, opened 2024-06-04
Ensure BuildOptions can configure the output directory instead of forcing inline rendering

rmannibucau:


rmannibucau:
@dotnet-policy-service agree

filzrev:
`BuildJsonConfig`  has `Output` property that can be used to change `output directory` by setting this property.

https://github.com/dotnet/docfx/blob/4f6f77945758ceb9d706b25dc33b04dabdbc6180/src/Docfx.App/Config/BuildJsonConfig.cs#L54-L60

Is there any missing features that can not be covered above solution?

---
Note: 
There is following behavior differences. So it might need to explicitly set `full path`. 
- When using `BuildJsonConfig`'s `Output` property. relative path is resolved based on `docfx.json` config path.
- When using `outputDirectory` parameter. It's resolved based on current directory.

rmannibucau:
Hi

One issue is the configuration is internal so programmatically it would require to generate a json file, I'd prefer something closer to docfx project api if possible.

filzrev:
> One issue is the configuration is internal

I've forgotten that `BuildJsonConfig` can't access programmatically because it's declared as `internal`. 
I've requested to change these options to `public` before at #8872.
But currently there is no progress.

As a temporary workaround.
By using `IgnoresAccessChecksToGenerator`.
It can use `RunBuild.Exec` method directly instead of `Docset.Build` API.

**Example configs of csproj**

```
  <ItemGroup>
    <PackageReference Include="IgnoresAccessChecksToGenerator" Version="0.7.0" PrivateAssets="All" />
    <InternalsAssemblyName Include="Docfx.App" />
    <InternalsAssemblyName Include="Docfx.Dotnet" />
  </ItemGroup>
```

rmannibucau:
@filzrev it would work but does it make sense to not expose what doxfx project uses or should docfx not use this output param anymore then?

rmannibucau:
guess we can drop this PR is https://github.com/dotnet/docfx/pull/10010 will reach main repo anytime soon, wdyt?

rmannibucau:
[auto] outdated PR, guess there is no will to get it so closing as a housekeeping cycle

PR #10007, opened 2024-06-12
chore: Fix nightly build errors that occurred on .NET 9 SDK Preview 5

filzrev:
This PR intended to fix nightly build error that occurred after CI build is updated to .NET 9 SDK Preview 5.

**Warning Message**
> CS9220
> One or more overloads of method 'Create' having non-array params collection parameter might be applicable only in expanded form which is not supported during dynamic dispatch.	Docfx.Build.SchemaDriven (net9.0)

Above warning occurred when passing `dynamic` variable to `ImmutableArray.Create`.
So I've added type cast logics.

### DIFF
--- a/src/Docfx.Build/Conceptual/ConceptualDocumentProcessor.cs
+++ b/src/Docfx.Build/Conceptual/ConceptualDocumentProcessor.cs
@@ -104,9 +104,14 @@ class ConceptualDocumentProcessor : DisposableDocumentProcessor
             FileLinkSources = model.FileLinkSources,
             UidLinkSources = model.UidLinkSources,
         };
-        if (model.Properties.XrefSpec != null)
+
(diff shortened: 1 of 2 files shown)


src/Docfx.Build/Conceptual/ConceptualDocumentProcessor.cs
```diff
@@ -104,9 +104,14 @@ public override SaveResult Save(FileModel model)
             FileLinkSources = model.FileLinkSources,
             UidLinkSources = model.UidLinkSources,
         };
-        if (model.Properties.XrefSpec != null)
+
+        if (((IDictionary<string, object>)model.Properties).TryGetValue("XrefSpec", out var value))
```
filzrev, line 108: `TryGetValue` logic is required to suppress **Internal** exception when accessing `model.Properties.XrefSpec` first time.

yufeih (approved):
Thank you @filzrev

Issue #2297, opened 2017-12-12
docfx watch


vicancy:
Monitor web page request, build the page with its dependencies.

netroby:
+1 need this feature,

bryancostanich:
As an organization looking at docfx to power our developer site, I have to say; from a productivity perspective, this is the first issue we've run into that's a major deal. 

We currently use github pages (jekyll + ruby), and when you run it locally, changes are automatically picked up and rendered. Without that feature, it's impossible to easily view our changes in situ; and have to rely on our markdown editor's render view. 

That coupled with the fact that I [can't actually terminate the server from mac terminal](https://github.com/dotnet/docfx/issues/3870), makes it very difficult to work on conceptual documentation in docfx. 

Autoupdates would improve the situation massively.

superyyrrzz:
@bryancostanich According to our [roadmap](https://github.com/dotnet/docfx/blob/dev/Roadmap.md), v2 will not support `watch`, and v3 will. Does [DocFX preview plugin](https://marketplace.visualstudio.com/items?itemName=ms-docfx.DocFX#review-details) help?

bryancostanich:
I can't actually get that plugin to work. It throws server errors. `Request failed with status code 405`.

I'll work on it with some folks to see if we can fix and evaluate.

bryancostanich:
Watch being a v3 feature makes complete sense to me, especially given how far along it is. By the way, i think your roadmap is out of date. It mentions december stuff as future. curious as to where you're currently at on v3? Can you provide an update on that issue, please?

yufeih:
@bryancostanich , I've updated the v3 roadmap [here](https://github.com/dotnet/docfx/blob/v3/docs/roadmap.md), we are currently busy working on docs integration and docs migration.

BennyKok:
A quick alternative I tried, I am able to use npm-watch to simulate a 'hot-reload' by retriggering `docfx build --serve` when the file changes. Incase this will be useful for someone.
![image](https://user-images.githubusercontent.com/18395202/99903967-e334b980-2d02-11eb-8e74-0b945950c41b.png)

prog-rajkamal:
To future visitors: here is a sample package.json to simulate watch as @BennyKok  suggested :

```json
  "scripts": {
        "watch": "npm-watch build",
        "build": "docfx build --serve"
    },
    "watch": {
        "build": {
            "patterns":[
                "api",
                "app",
                "toc",
                "index"
            ],
[...]

PR #10010, opened 2024-06-14
[docfx] adding --watch option to build command

rmannibucau:
Follow-up of my tests (https://github.com/dotnet/docfx/pull/9986) and https://github.com/dotnet/docfx/discussions/9992

Not 100% sure of the following so happy to get a feedback on all these points:

1. didn't see much tests of the build/serve mode so does it need a test (regarding project compliance)?
2. I'm not very clear why default command is not build command but a kind of fork not doing exactly the same thing so maybe I didn't patch the right location
3. the glob to pattern logic is very simple to start (can always be enhanced later), would be happy to get an ack it is sufficient to start or not from your experience (I'm very new to docfx and don't know all the subtleties yet)
4. I opened docfx.json model, hope it is ok (I assumed thanks to #9986 it was maybe an option but not 100% sure)


src/Docfx.App/Config/BuildJsonConfig.cs
```diff
@@ -13,7 +13,7 @@ namespace Docfx;
 /// BuildJsonConfig.
 /// </summary>
 /// <see href="https://dotnet.github.io/docfx/reference/docfx-json-reference.html#12-properties-for-build"/>
-internal class BuildJsonConfig
```
filzrev, line 16: This change seems no behavior differences.
Access to this class from `docfx` is already allowed by `InternalsVisibleTo` attribute.

rmannibucau: Right but it helps https://github.com/dotnet/docfx/pull/9986 too and makes this PR more "readable" since it does not need to check the `InternalsVisibleTo` so think it is worth it to include it - but if blocking I can remove it, trying to explain the rational.

filzrev:
In my personal opinion.
This feature should be implemented as separated `docfx watch` command (#2297).
And existing `docfx` and `docfx build` commands are leave as is.

Because `docfx watch` feature requires different set of options/configs.
And implementation can be simpler for future command extensibilities.

For example.
- Watch `_site` contents and reload browser window (LiveReload/BrowserRefresh)  similar to `dotnet watch` command.
- Support `Ctrl+R` hotkey to force rebuild. similar to `dotnet watch` command.
- Support pseudo incremental build (Build only changed document)

src/Docfx.App/Config/BuildJsonConfig.cs
filzrev: Default class visibility is `internal` when declared under namespace.
It thought it need to specify `public` for consistency with PR comments. (or wait #8872 is resolved)

https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/access-modifiers#class-and-struct-accessibility

> Classes and structs declared directly within a namespace (aren't nested within other classes or structs) can be either public or internal. internal is the default if no access modifier is specified.

rmannibucau:
>Because docfx watch feature requires different set of options/configs.
>And implementation can be simpler for future command extensibilities.

Hmm, it has serve capabilities so adding watch which is a serve companion or concurrent is not shocking to me.
To be concrete if we do a watch command and follow the reasoning it means watch will not serve right? But then we need yet another watch+serve command so not sure it is great.
I'm coming from a projects where commands are `build` (plain build no serve or watch command) and `serve` or `watch` which both handle both features (optionally with live reload support) so I'm surely biased but my experience with docfx is I have no idea how `serve` can work since to work on the doc I needed to setup `inotify` and a plain `http-server` was replacing `serve` command quite quickly so maybe I'm missing some usage there - happy to learn if so.

>
>For example.
>
> Watch _site contents and reload browser window (LiveReload/BrowserRefresh) similar to dotnet watch command.

Kind of requires serve to be there as well no?

> Support Ctrl+R hotkey to force rebuild. similar to dotnet watch command.
> Support pseudo incremental build (Build only changed document)

+1 (it is the last `else` block where watcher is started without serve feature) and I'm happy to do a watch only feature extracted from from this PR but we still need an all in one command, where would it fit if not build one which is already an all in one?
Also what about the default command convergence with build or should it be used as all in one command?

filzrev:
Thank you for confirming my PR comments

I'm expecting following behaviors by `docfx watch` command. (If it's added by this PR)

When `docfx watch` command executed.
By default, following processes are executed on separated `HostedService` thread.
1. Serve `_site` directory content with ASP.NET. (And inject `livereload-js` scripts to HTML file)
2. Watch changes of `_site` directory files. When changes are detected. Send LiveReload command to browser.
3. Watch changes of docs files. And run build when files are modified.
4. Watch keyboard input. And When `Ctrl+R` key pressed. Request docfx build.


And by adding `watch` section to `docfx.json`.
It can change behaviors.
- `"Serve": false` option suppress `1.` and `2.` behaviors.
- `"Watch": false` option suppress `3.` behavior.

---

If basic `docfx watch` command infrastructure is added by this PR.
I'm also willing creating PR that relating to this features. (`Ctrl+R` hotkey support and LiveReload supports)

rmannibucau:
@filzrev I'm in the process of moving build to watch command (restoring old build command), hope I got it right

rmannibucau:
Anything I can do to help getting this onboard?

filzrev:
> Anything I can do to help getting this on board?

To supporting `docfx watch` command.
I thought It need to add following statement to `Program.cs`.

> config.AddCommand<WatchCommand>("watch");

And for `WatchCommandOptions`.
Following options expected to be enabled by default (that similar to `dotnet watch` command).
- `Watch`
- `Serve`
- `OpenBrowser`

Currently it inherit `BuildCommandOptions` tought
Personally I though `dotnet watch` command should have different set of command line parameters.
(e.g. `--no-watch`)

@yufeih 
I'm also want to add feature that relating `dotnet watch` command.
Is it able to create dedicated `feature branch`?

rmannibucau:
Yes addcommand is needed but was awaiting to ensure solution was ok.

About build command i guess it would be neat to extract settings in a shared base setting class to ensure it is consistent accross command and nlt duplicated, wdyt?

src/docfx/Models/WatchCommandOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Docfx;
+
+[Description("Generate client-only website combining API in YAML files and conceptual files and watch them for changes")]
+internal class WatchCommandOptions : DefaultBuildCommandOptions
+{
+    [Description("Should directory be watched and website re-rendered on changes.")]
+    [CommandOption("-w|--watch")]
+    [DefaultValue("true")]
+    public bool Watch { get; set; }
```
yufeih, line 15: The `--watch` option looks unnecessary, when would we want to execute `docfx watch` without watching?

src/docfx/Models/WatchCommandOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+
+    [Description("Host the generated documentation to a website")]
+    [CommandOption("-s|--serve")]
+    [DefaultValue("true")]
+    public bool Serve { get; set; }
```
yufeih, line 20: Since `docfx watch` serves content by default, use `docfx watch --no-serve` to disable serve.

src/docfx/Models/WatchCommand.cs
```diff
@@ -0,0 +1,179 @@
...
+            lock (cancellation)
+            {
+                ResetToken(cancellation);
+                cancellation[0] = token;
+            }
+
+            await Task.Delay(100, token.Token);
+            if (!token.IsCancellationRequested)
+            {
+                onChange();
```
yufeih, line 104: if the build invoked by `onChange` method takes a minute to finish and files changes each second, would we trigger multiple build or wait for the last build to complete before triggering the next?

yufeih:
@filzrev @rmannibucau

Implemented as a standalone `watch` command looks good. I added a `feature/watch` branch for you to work towards completion, also left some comments.

src/docfx/Models/WatchCommand.cs
```diff
@@ -0,0 +1,179 @@
...
+    internal static IDisposable Watch(string baseDir, BuildJsonConfig config, Action onChange)
+    {
+        FileSystemWatcher watcher = new(baseDir)
+        {
+            IncludeSubdirectories = true,
+            NotifyFilter = NotifyFilters.Attributes | NotifyFilters.Size | NotifyFilters.FileName |
+                           NotifyFilters.DirectoryName | NotifyFilters.LastWrite
+        };
+
+        if (WatchAll(config))
```
yufeih, line 71: We can start with a simple deterministic watch implementation that watches all files changes in the `docfx.json` directory, except for the output directory (`_site`)

rmannibucau:
[auto] outdated PR, guess there is no will to get it so closing as a housekeeping cycle

PR #10012, opened 2024-06-17
Fix #9990 header and footer template files

### LINKED ISSUE #9990: HTML template not used for header and footer
In my previous issue #9463 I requested the use of header and footer to be added through a template.
This has now been added, but not in a way that is very usable.
I have several issues with the current implementation.

# Template is not using html files
When I try to use the pdfHeaderTemplate, I cannot use it like the cover page.
I have tried the following:
```yml
pdf: true
pdfTocPage: true
pdfFileName: TestDoc.pdf
pdfCoverPage: articles/TestDoc/0-cover.html
pdfHeaderTemplate: articles/TestDoc/0-header.html
pdfFooterTemplate: articles/TestDoc/0-footer.html
items:
- name: Introduction
  href: 1-introduction.md
- name: Some amazing content
  href: 2-some-amazing-content.md
- name: Conclusion
  href: 3-conclusion.md
```
When I generate this, the cover page is properly added, but the header becomes (very tiny in the top left corner):
image
The same goes for the footer

In the toc.yml file, if I add the full html of the example/default I get an error, since the html contains a colon (:) since html works that way.

# The header/footer are not added to the TOC page
Nothing is added on the TOC page.
image

# What I would like
In the end it would be great if I can add a template for both header and footer in html file (or md which generates html files).
I have added the following zip, that contains all files that could be turned into a pdf file.
This PDF would contain 3 chapters, a TOC and the front page.
Each of these files would receive the header and footer as provided.

[TestDoc.zip](https://github.com/user-attachments/files/15614077/TestDoc.zip)

Foitn:
Add reading of file, instead of only allowing the template to be added to the toc.yml


src/Docfx.App/PdfBuilder.cs
```diff
@@ -217,6 +217,13 @@ static string ExpandTemplate(string? pdfTemplate, int pageNumber, int totalPages
                     .Replace("<span class='totalPages'></span>", $"<span>{totalPages}</span>")
                     .Replace("<span class=\"totalPages\"></span>", $"<span>{totalPages}</span>");
             }
+
+            static string? GetHeaderFooter(string? template)
+            {
+                return File.Exists(template)
+                    ? File.ReadAllText(template)
```
yufeih, line 224: Where is the current directory if this is a relative path? Consider explicitly state the base directoy to relative to

yufeih:
Close in favor of #10099

Issue #2532, opened 2018-03-19
Error extracting metadata for project : An item with the same key has already been added


darylrobbins:
**DocFX Version Used**: 2.33

**Template used**: `default`

**Steps to Reproduce**:

Paths have been anonymized to protect the innocent.

1. Execute `docfx metadata C:\Users\someuser\Code\SomeProject\Module\SomeLibrary.Shared\SomeLibrary.Shared.vbproj -f`

**Expected Behavior**:
Successfully generates metadata

**Actual Behavior**:
Metadata generation fails with the following error:
```
[18-03-19 07:58:56.263]Info:[ExtractMetadata]Generating metadata for each project... 
[18-03-19 07:59:04.842]Error:Error extracting metadata for C:\Users\someuser\Code\SomeProject\Module\SomeLibrary.Shared\SomeLibrary.Shared.vbproj: Error extracting metadata for project "C:\Users\someuser\Code\SomeProject\Module\SomeLibrary.Shared\SomeLibrary.Shared.vbproj": An item with the same key has already been added.
```

**Notes:**
This is the first time trying to run docfx on our project. I saw the same error on other VB projects. I tried earlier versions and saw similar/additional errors.

vicancy:
Hi @darylrobbins Could you share a minimum project that can repro the issue to us?

michaelp:
I have the same issue with version 2.35.4 
_Is it possible to increase verbosity and get the docfx log ?_

vicancy:
Hi @michaelp Yes good suggestion. Meanwhile, could you provide some minimum project for us to repro the issue? Is there duplicate references in the project?

JulienM28:
Hello,
I just got the same issue with 2.3.5 version.

Impossible to build the documentation.

Projects are using .Net Framework 4.5.2.

michaelp:
problem happens with the tests projects 
The following snippet of code is the source of the issue. It happens on 6th projectis getting into the dictionary
checking if there is anything special about those projects. 

```csharp
private static async Task<ConcurrentDictionary<string, AbstractCompilation>> GetProjectCompilationAsync(ConcurrentDictionary<string, AbstractProject> projectCache)
        {
            var compilations = new ConcurrentDictionary<string, AbstractCompilation>();
            var sb = new StringBuilder();
            foreach (var project in projectCache)
            {
                try
                {
                    var compilation = await project.Value.GetCompilationAsync();
                    compilations.TryAdd(project.Key, compilation);
                }
[...]

PR #10023, opened 2024-06-21
fix: `docfx metadata` command throw `ArgumentException` when referencing empty namespace by doc comment

filzrev:
This PR intended to fix issue reported at https://github.com/dotnet/docfx/issues/2532#issuecomment-2181542930

**Details of problem**

`docfx metadata` command failed with `ArgumentException` at [this line](https://github.com/dotnet/docfx/blob/cc5754aa8c4d7fb083ed1de8514265804b58ce76/src/Docfx.Dotnet/ManagedReference/Resolvers/YamlMetadataResolver.cs#L105)

```
ArgumentException: An item with the same key has already been added. Key: Root.Datastore
  at bool TryInsert(TKey key, TValue value, InsertionBehavior behavior)
  at void Add(TKey key, TValue value)
  at MetadataItem GenerateNestedTocStructure(IEnumerable<KeyValuePair<string, MetadataItem>> namespaces,
     Dictionary<string, ReferenceItem> allReferences) in YamlMetadataResolver.cs:105
```

This problem is occurred when following conditions are met

1. Referencing empty namespace with `<see cref="">` tag.
2. Using "namespaceLayout": "nested"
3. Using "outputFormat": "mref"

**What's changed in this PR**

This PR changes `allReferences.Add` logic to `allReferences.TryAdd` method to skip when key already exists.

For example.  When `/// <see cref="Root.Datastore"/>` comment exists.
`allReferences` contains this information as `ReferenceItem` (`N:Root.Datastore`).

Additionally I've added post-process logics to overwrite **empty namespace** reference.
It's required because **Empty namespace** don't generate HTML page.
So `<see cref=""/>` link is not works for empty namespace.

**Test**
I've manually tested following tests by using [DocFxUndefinedReferenceBugDemo](https://github.com/scott-ainsworth/DocFxUndefinedReferenceBugDemo) content.

1. `docfx metadata` command successfully completed.
2. `<see cref=""` link to **empty namespace** is rendered as plain text.
3. `<see cref=""` link to **other namespace** is rendered as normal link.

### DIFF
--- a/src/Docfx.Dotnet/ManagedReference/Resolvers/YamlMetadataResolver.cs
+++ b/src/Docfx.Dotnet/ManagedReference/Resolvers/YamlMetadataResolver.cs
@@ -102,7 +102,7 @@ internal static class YamlMetadataResolver
                         };
                         missingNamespace.DisplayNames.Add(SyntaxLanguage.Default, partialParentNamespace);
                         namespacedItems[partialParentNamespace] = missingNamespace;
-                        allReferences.Add(partialParentNamespace, new());
+                        allReferences.TryAdd(partialParentNamespace, new());
 
                         if (!partialParentNamespace.Contains('.'))
                         {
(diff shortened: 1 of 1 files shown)


scott-ainsworth:
This branch correctly builds both the minimal sample I provided and the original 600+ *.cs files on which I first encountered the crash.

Thanks for the really fast resolution!

R/Scott Ainsworth

yufeih (approved):
Thank you @filzrev

Issue #10028, opened 2024-06-21
[Feature Request] Enable SystemMetadataGenerator to never use a relative path but always an absolute path


rmannibucau:
**Is your feature request related to a problem? Please describe.**
I want any asset or page to use absolute links from the root of the website (to be configured in this mode).
Goal is to be able to configure the base of the website on github pages (or any similar system).
Instead of having for subfolders something like `../../public/docfx.min.js` I'd like to get `/myproject/public/docfx.min.js` and so on.

This issue pops obviously for pages not stored in the URL path folder. A common example is `404.html` - created from `404.md` - which will be served when hitting `/myproject/foo/bar/whatever/not/existing`.

**Describe the solution you'd like**
Ideally I'd like to put in header metadata on `404.md` a flag to use absolute path for this page - but if global for all pages I'm happy too.

**Describe alternatives you've considered**
For now I added a js script in the page to redirect on the actual 404 page url when the subfolder is not matching the hosting but this changes the url and is not satisfying.

filzrev:
> This issue pops obviously for pages not stored in the URL path folder. A common example is 404.html

I've tested `Rewriting 404.html file on deploying GitHub Pages` approach.
And it seems to works as expected.

**PowerShell steps to rewrite HTML**
https://github.com/filzrev/docfx_sample_404_notfound/blob/main/.github/workflows/build_docs.yml#L78C1-L93C47

**Sample `404.md` file that simulate default 404: NotFound page**
https://raw.githubusercontent.com/filzrev/docfx_sample_404_notfound/main/docs/404.md

**Sample render result page**
https://filzrev.github.io/docfx_sample_404_notfound/foo/bar/whatever/not/existing

Is there other scenario that needs **Root relative URL**?
If there is such a requirements, I thought it need to add following features.
1.  Add `<base href="/{baseURL}/" />` tag inside `<head>` tag.
2.  Rewrite URLs to  **Root relative URL** by custom PostProcessor.

rmannibucau:
@filzrev `<base>` is a workaround indeed but it has side effects (in particular when embedding js in pages) so it is always better to avoid it. Indeed a custom implementation will make it work, this is why this issue is about ensuring default generation is compatible with common frontend middleware error page mapping without redirection hacks or breaking the site js.

filzrev:
> `<base>` is a workaround indeed but it has side effects (in particular when embedding js in pages) so it is always better to avoid it. 

[...]

PR #10032, opened 2024-06-24
chore: Fix `navrel`/`tocrel` URL resolve behaviors when using `Site Relative URL`

filzrev:
This PR intended to fix some problem that discussed at #10028

On current docfx modern template implementation.
`navrel` / `tocrel` absolute URL is resolve by using following code.

>  `new URL('toc.json', window.location.href)

So when I've modified docfx generated pages URLs to `Site Relative URL` (That starts with '/') based URL.
JavaScript errors are occurred on non-site-root pages.

This PR fix this issue by adding dedicated code path for `Site Relative URL`.

**Test**
I've manually confirmed following patterns and it works as expected.
1. Default page (That use relative URLs)
2. Rewriting HTML page URLs to `Site Relative URL` .
3. Rewriting HTML page URLs to `Absolute URL`.

And it also tested on `default` template environment.
It ssems `default` template processing `navrel` / `tocrel` URL as plain string. so no modification needed.

### DIFF
--- a/templates/modern/src/nav.ts
+++ b/templates/modern/src/nav.ts
@@ -73,7 +73,9 @@ export async function renderNavbar(): Promise<NavItem[]> {
       return []
     }
 
-    const navUrl = new URL(navrel.replace(/.html$/gi, '.json'), window.location.href)
+    const navUrl = navrel.startsWith('/')
+      ? new URL(navrel.replace(/.html$/gi, '.json'), document.baseURI)
+      : new URL(navrel.replace(/.html$/gi, '.json'), window.location.href)
     const { items } = await fetch(navUrl).then(res => res.json())
     return items.map((a: NavItem | NavItemContainer) => {
       if ('items' in a) {
--- a/templates/modern/src/toc.ts
+++ b/templates/modern/src/toc.ts
@@ -23,7 +23,9 @@ export async function renderToc(): Promise<TocNode[]> {
 
   const disableTocFilter = meta('docfx:disabletocfilter') === 'true'
 
-  const tocUrl = new URL(tocrel.replace(/.html$/gi, '.json'), window.location.href)
+  const tocUrl = tocrel.startsWith('/')
+    ? new URL(tocrel.replace(/.html$/gi, '.json'), document.baseURI)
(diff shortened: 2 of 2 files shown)


templates/modern/src/nav.ts
```diff
@@ -73,7 +73,9 @@ export async function renderNavbar(): Promise<NavItem[]> {
       return []
     }
 
-    const navUrl = new URL(navrel.replace(/.html$/gi, '.json'), window.location.href)
+    const navUrl = navrel.startsWith('/')
```
yufeih, line 76: `document.baseURL` is `window.location.href` when `<base>` is not set. Does this work?

`const navUrl = new URL(navrel.replace(/.html$/gi, '.json'), document.baseURL)`

rmannibucau:
Not sure it can fix #10028 since #10028 is really about loading assets, so fixing only the js will make it a bit less broken but still broken.
Guess a more transversal and not related to a particular theme solution is more desired, no?

templates/modern/src/nav.ts
filzrev: > document.baseURL is window.location.href when <base> is not set. Does this work?

Sorry for the confusion.
This PR intended to fix toc loading issue when **custom** `<base>` tag is used on HTML file.

If `<base>` tag is not exists on HTML file.
There is no behavior differences.

filzrev:
Close this issue.

Because I've changed approach on my code not to use `<base>` tags.
(Instead rewrite entire HTML URLs for `404.md` to `Site Relative URL` on site deployment timing)

- Workflow: https://github.com/filzrev/docfx_sample_404_notfound/blob/main/.github/workflows/build_docs.yml#L91-L97
- Script to rewrite URLs: https://github.com/filzrev/docfx_sample_404_notfound/blob/main/.github/scripts/Rewrite404Page.csx

This PR contents might be useful for people who want to use `<base>` tag on custom docfx layout page.
Thought currently there are not such users at this time. So close this PR.

rmannibucau:
Side note/warning: this is limited to default template usage AFAIK. It will not handle all the links in meta (rdf and friends) and `data-xxx` attributes. So customizations to the csx can be needed. (mainly for future readers, not asking any actions)

PR #10051, opened 2024-06-29
Remove some enum boxing in GlobMatcher

lahma:
Traces are showing `GlobMatcherOptions` being boxed and causing allocations.

image


yufeih (approved):
Thank you @lahma

PR #10055, opened 2024-07-01
feat: Support `search.min.js` script customize functionality

filzrev:
This PR intended to support `search.min.js` script customization.

Currently docfx `modern` template bundle scripts to `docfx.min.js`.
So it can't customize search scripts with custom template.

This PR exclude `search.ts` file from bundling by adding following settings

- Set 'modern/src/search.ts' as `entryPoints` to separately create `search.min.js`.
- Rewrite `./search`  import statement settings by custom esbuild plugin.
  - Set `external` property to `true` to exclude from `esbuild bundle`
  - Rewrite `path` from `./search` to `./search.min.js`

Additionally this PR fix .NET runtime version to .NET 8 (that used by `npm run start`)

### DIFF
--- a/templates/build.js
+++ b/templates/build.js
@@ -51,12 +51,24 @@ async function buildModernTemplate() {
     entryPoints: [
       'modern/src/docfx.ts',
       'modern/src/search-worker.ts',
+      'modern/src/search.ts',
     ],
     external: [
       './main.js'
     ],
     plugins: [
-      sassPlugin()
+      sassPlugin(),
+      {
+        name: 'Exclude `./search` script from bundle',
+        setup(build) {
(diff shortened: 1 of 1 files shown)


yufeih:
Should we instead expose a callback method in `DocfxOption` for custom search implementation? That `options` object is the pattern for customizing JS behaviors.

filzrev:
> Should we instead expose a callback method in DocfxOption for custom search implementation? That options object is the pattern for customizing JS behaviors.

Currently `DocfOption` settings are configuration is supported by `main.js` customization only.

On my use case. I want to switch search engine from `lurn` based search to `PageFind` based search by `custom template`.
So I hope it can be customize search by `custom template`.

vicancy (approved):

PR #10066, opened 2024-07-05
feat: Remove Node.js runtime dependencies from docfx tools

### LINKED ISSUE #9396: Exploded package size and quality concerns
Package size at https://www.nuget.org/packages/docfx changed from 51 MB to 199 MB since v2.73.0. Is that really needed?

I suspect this is related to the new PDF rendering, which I'm not interested in. However, looking at the output diff, every HTML page now contains the following extra line:
```
<meta name="loc:downloadPdf" content="Download PDF">
```

Do you guys ever diff the HTML output with the previous version before release?
Or even run it through an HTML validator? For example, why does every file contain `</article>` without an opening tag?

I've been getting updates for docfx frequently since I turned on Dependabot updates, but now I'm considering turning that off. It just takes too much time to verify everything and report back. For example, see https://github.com/dotnet/docfx/issues/9337#issuecomment-1798876066 which was supposed to be fixed, but apparently was never tested. I feel like being the beta tester of your product, this is not very professional.

filzrev:
This PR intended to resolve #9396.

**Background**

Playwright `1.41` or later supports following environment variables. (It's not documented though)

- `PLAYWRIGHT_NODEJS_PATH`
- `PLAYWRIGHT_DRIVER_SEARCH_PATH`

So it can remove Node.js runtime from `docfx` package.
Instead.  it can use locally installed Node.js runtime instead. 

By this changes. `docfx` packages size can be reduced from `208MB` -> `56 MB`

**BREAKING CHANGES**
This PR remove `Node.js` runtime bundles.
So it may introduce **breaking changes** for some users who using docfx without Node.js installation.
(e.g. docker image users)

I'll try to fix official docker image later.
https://github.com/dotnet/docfx/blob/main/Dockerfile

### DIFF
--- a/src/Docfx.App/Helpers/PlaywrightHelper.cs
+++ b/src/Docfx.App/Helpers/PlaywrightHelper.cs
@@ -25,7 +25,7 @@ internal static class PlaywrightHelper
             throw new DocfxException("Node.js executable is not found. Try to install Node.js or set the `PLAYWRIGHT_NODEJS_PATH` environment variable.");
         }
 
-        Logger.LogInfo($"Using custom Node.js {nodeVersion} executable.");
+        Logger.LogInfo($"Using Node.js {nodeVersion} executable.");
         Logger.LogVerbose($"Path: {exePath}");
 
         Environment.SetEnvironmentVariable("PLAYWRIGHT_NODEJS_PATH", exePath, EnvironmentVariableTarget.Process);
--- a/src/Docfx.App/Helpers/PlaywrightHelper.cs
+++ b/src/Docfx.App/Helpers/PlaywrightHelper.cs
@@ -1,0 +1,80 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+using System.Runtime.InteropServices;
+using Docfx.Common;
+using Docfx.Exceptions;
+
+#nullable enable
+
+namespace Docfx;
+
+internal static class PlaywrightHelper
+{
+    public static void EnsurePlaywrightNodeJsPath()
+    {
+        // Skip if playwright environment variable exists.
+        if (Environment.GetEnvironmentVariable("PLAYWRIGHT_DRIVER_SEARCH_PATH") != null)
+            return;
+
+        if (Environment.GetEnvironmentVariable("PLAYWRIGHT_NODEJS_PATH") != null)
+            return;
+
+        if (!TryFindNodeExecutable(out var exePath, out var nodeVersion))
+        {
+            throw new DocfxException("Node.js executable is not found. Try to install Node.js or set the `PLAYWRIGHT_NODEJS_PATH` environment variable.");
+        }
+
+        Logger.LogInfo($"Using custom Node.js {nodeVersion} executable.");
+        Logger.LogVerbose($"Path: {exePath}");
+
+        Environment.SetEnvironmentVariable("PLAYWRIGHT_NODEJS_PATH", exePath, EnvironmentVariableTarget.Process);
+    }
+
+    private static bool TryFindNodeExecutable(out string exePath, out string nodeVersion)
+    {
+        // Find Node.js executable installation path from PATHs.
+        string exeName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "node.exe" : "node";
+
+        var pathEnv = Environment.GetEnvironmentVariable("PATH");
+        if (pathEnv == null)
+            throw new DocfxException("Failed to get `PATH` environment variable.");
+
+        var paths = pathEnv.Split(Path.PathSeparator);
+        foreach (var path in paths)
+        {
+            string fullPath = Path.GetFullPath(Path.Combine(path, exeName));
+
+            if (File.Exists(fullPath))
+            {
+                exePath = fullPath;
+                nodeVersion = GetNodeVersion(exePath);
+                return true;
+            }
+        }
+
+        exePath = "";
+        nodeVersion = "";
+        return false;
+    }
+
+    /// <summary>
+    /// Returns `node --version` command result
+    /// </summary>
+    private static string GetNodeVersion(string exePath)
+    {
+        using var memoryStream = new MemoryStream();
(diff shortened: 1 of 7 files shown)


src/Docfx.App/Helpers/PlaywrightHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+    {
+        // Skip if playwright environment variable exists.
+        if (Environment.GetEnvironmentVariable("PLAYWRIGHT_DRIVER_SEARCH_PATH") != null)
+            return;
+
+        if (Environment.GetEnvironmentVariable("PLAYWRIGHT_NODEJS_PATH") != null)
+            return;
+
+        // Find Node.js executable installation path from PATHs.
+        string exeName = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) ? "node.exe" : "node";
```
yufeih, line 22: Does this work here ? https://github.com/dotnet/docfx/blob/main/src/Docfx.Common/CommandUtility.cs#L110C24-L110C36

src/Docfx.App/Helpers/PlaywrightHelper.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            if (File.Exists(fullPath))
+            {
+                Logger.LogInfo($"Using custom Node.js executable: {fullPath}");
+                Environment.SetEnvironmentVariable("PLAYWRIGHT_NODEJS_PATH", fullPath, EnvironmentVariableTarget.Process);
+                return;
+            }
+        }
+
+        throw new DocfxException("Failed to find Node.js executable from `PATH`.");
```
yufeih, line 41: consider add a hint to the error message to install NodeJS or set `PLAYWRIGHT_NODEJS_PATH`

filzrev: I've added error messages in latest commit.
> Node.js executable is not found. Try to install Node.js or set the `PLAYWRIGHT_NODEJS_PATH` environment variable.

src/Docfx.App/Helpers/PlaywrightHelper.cs
filzrev: On latest commit.
Refactor code to use `CommandUtility`.

Additionally. I've added log for Node.js version.

src/Docfx.App/Helpers/PlaywrightHelper.cs
```diff
@@ -0,0 +1,68 @@
...
+            return;
+
+        if (Environment.GetEnvironmentVariable("PLAYWRIGHT_NODEJS_PATH") != null)
+            return;
+
+        if (!TryFindNodeExecutable(out var exePath, out var version))
+            throw new DocfxException("Node.js executable is not found. Try to install Node.js or set the `PLAYWRIGHT_NODEJS_PATH` environment variable.");
+
+        Logger.LogInfo($"Using custom Node.js {version} executable.");
+        Logger.LogVerbose($"Path: {exePath}");
```
filzrev, line 29: I've changed code to print Node.js version by default.
And exe path information is outputted as verbose log.

filzrev:
When using `CommandUtility.ExistCommand` helper methods.
It need to handle stdout result parse logics.
Because of following reasons.
 
- On GitHub runner environment multiple node.js installations found by `where` command (Windows)  
 - it's possible to search `node` command on current directory by where command.
- `node is` prefix string appended by `type` command (Linux/mac)

So I've reverted code to `PATH` based search logics.

yufeih (approved):
Thank you @filzrev !

JanVargovsky:
Hi @yufeih, is it expected that everyone has to install the node.js locally even if we don't use any PDF related feature?

filzrev:
Node.js installation is required when running following commands.
- `docfx` (Default command)
- `docfx pdf`

It might be better to modify `Prerequisite` section of following document.
I'll try to create PR later.
- https://github.com/dotnet/docfx/blob/main/docs/index.md?plain=1#L9-L12

georg-jung:
I use the default command with `"pdf": false` in my `docfx.json`. I run docfx inside docker, image `mcr.microsoft.com/dotnet/sdk:8.0-jammy`. With 2.78.x, docfx started breaking my builds with

```log
TypeInitializationException: The type initializer for 'Docfx.Pdf.PdfBuilder' 
threw an exception.
     DocfxException: Node.js executable is not found. Try to install Node.js or 
     set the `PLAYWRIGHT_NODEJS_PATH` environment variable.                     
       at void EnsurePlaywrightNodeJsPath() in PlaywrightHelper.cs:25           
       at cctor() in PdfBuilder.cs:53                                           
  at Task CreatePdf(string outputFolder)                                        
  at void <Execute>b__0() in DefaultCommand.cs:53                               
  at int Run(LogOptions options, Action run) in CommandHelper.cs:48             
  at int Execute(CommandContext context, Options options) in DefaultCommand.cs: 
     31                                                                         
  at Task<int> Execute(CommandContext context, CommandSettings settings) in     
     CommandOfT.cs:40                                                           
  at async Task<int> Execute(CommandTree leaf, CommandTree tree, CommandContext 
     context, ITypeResolver resolver, IConfiguration configuration) in          
     CommandExecutor.cs:166
```

Could it be a possibility to not require Node.js if PDF output is disabled?

JanVargovsky:
Workaround is simply by setting `PLAYWRIGHT_DRIVER_SEARCH_PATH` or `PLAYWRIGHT_NODEJS_PATH` to non empty value. For example in CMD:
> set PLAYWRIGHT_NODEJS_PATH=anyvalue && docfx
My build then succeeded as it used to. Or install node.js even if it's not required (unless you are using pdf).

filzrev:
I've created PR(#10406) to fix a reported problem with the default `docfx` command.

georg-jung:
FWIW in the meantime, 
```dockerfile
RUN PLAYWRIGHT_NODEJS_PATH="workaround https://github.com/dotnet/docfx/pull/10066#issuecomment-2486269657" docfx tech/docfx.json --disableGitFeatures=true
```
works in my Dockerfile. Thanks @JanVargovsky and of course @filzrev for the super fast fix!

PR #10068, opened 2024-07-05
make verified files as lf and utf8

SimonCropp:
https://github.com/VerifyTests/Verify?tab=readme-ov-file#text-file-settings


yufeih (approved):
Thank you @SimonCropp

PR #10072, opened 2024-07-08
feat: UseClrTypeNames: new option to indicate that the output must use the CLR  type names instead of the language specific aliases

Patrick8639:
The option can be set in `docfx.json`configuration file, under the `metadata` object.

If the option is not set or its value is `false`, docfx works as usual: the output uses the language aliases.
For example, in C#, it displays `int`.

If the option value is `true`, docfx output will use the CLR type names.
For example, in C#, it displays `Int32`.

### DIFF
--- a/docs/reference/docfx-cli-reference/docfx-metadata.md
+++ b/docs/reference/docfx-cli-reference/docfx-metadata.md
@@ -95,6 +95,15 @@ Run `docfx metadata --help` or `docfx -h` to get a list of all available options
   - `SeparatePages`
     - Place members in separate pages.
 
+- **--useClrTypeNames**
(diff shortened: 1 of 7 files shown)


filzrev:
Is it able to modify following related document/schema in this PR?

- https://github.com/dotnet/docfx/blob/main/docs/reference/docfx-cli-reference/docfx-metadata.md
- https://github.com/dotnet/docfx/blob/main/schemas/docfx.schema.json#L574

yufeih (approved):

yufeih:
Thank you @Patrick8639 ! I'll update and merge the other PR after this is in.

PR #10088, opened 2024-07-08
use target typed new

SimonCropp:
where type is obvious


yufeih (approved):
Thank you @SimonCropp ! Really excited to more folks contributing to this project and making it better 🎉

Issue #10089, opened 2024-07-08
[Feature Request] Include `type` on `toc.yml`


frarees:
### Use case

I want to filter out TOC entries based on their member type. For example, to list only namespaces and classes.

### Current behavior

`toc.yml` doesn't contain type information.

> [!NOTE]
> When using `memberLayout: "separatePages"`, the additional TOC entries pushed to the model contain this information.

### How to reproduce

For example, given this source:

```csharp
namespace Lib
{
    public class Lib
    {
        public enum MyEnum
        {
            Value1,
            Value2,
        }

        public class SubLib
        {
            public float Foo { get; private set; }
        }

        public MyEnum MyEnumField;
        public float Foo { get; private set; }
        public float Bar { get; set; }
        public int Baz;

        public void MyMethod()
        {
            System.Console.WriteLine("Hello, World!");
        }
    }
}
```

The generated model looks like this (pre-transformations):

![model](https://github.com/dotnet/docfx/assets/220249/0c43e511-dc53-47cf-bd3d-79980c871e8d)

And the HTML TOC looks like this:

![toc](https://github.com/dotnet/docfx/assets/220249/79d6e64a-27c7-450f-bca3-af9ba455a1dd)

### Proposal

See #10090.

> [!NOTE]
> This might require some attention to the templates, since there's assumptions based on entries having defined the `type` property - which makes the TOC render like this:

![image](https://github.com/dotnet/docfx/assets/220249/bec4b3ae-1fff-44a0-81e7-e174812fcefa)

frarees:
Addressed in #10090.

PR #10090, opened 2024-07-08
feat: expose type through the toc

frarees:
See #10089.

### DIFF
--- a/src/Docfx.DataContracts.Common/TocItemViewModel.cs
+++ b/src/Docfx.DataContracts.Common/TocItemViewModel.cs
@@ -92,6 +92,11 @@ public class TocItemViewModel
     [JsonPropertyName("order")]
     public int? Order { get; set; }
 
+    [YamlMember(Alias = Constants.PropertyName.Type)]
+    [JsonProperty(Constants.PropertyName.Type)]
+    [JsonPropertyName(Constants.PropertyName.Type)]
+    public string Type { get; set; }
+
     [YamlIgnore]
     [Newtonsoft.Json.JsonIgnore]
     [System.Text.Json.Serialization.JsonIgnore]
--- a/src/Docfx.DataContracts.Common/TocItemViewModel.cs
+++ b/src/Docfx.DataContracts.Common/TocItemViewModel.cs
@@ -115,7 +115,7 @@ public class TocItemViewModel
     [ExtensibleMember]
     [Newtonsoft.Json.JsonIgnore]
     [System.Text.Json.Serialization.JsonIgnore]
-    public Dictionary<string, object> Metadata { get; set; } = new Dictionary<string, object>();
+    public Dictionary<string, object> Metadata { get; set; } = new();
 
     [EditorBrowsable(EditorBrowsableState.Never)]
     [YamlIgnore]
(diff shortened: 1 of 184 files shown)


frarees:
@dotnet-policy-service agree

src/Docfx.DataContracts.Common/TocItemViewModel.cs
```diff
@@ -92,6 +92,11 @@ public class TocItemViewModel
     [JsonPropertyName("order")]
     public int? Order { get; set; }
 
+    [YamlMember(Alias = Constants.PropertyName.Type)]
+    [JsonProperty(Constants.PropertyName.Type)]
+    [JsonPropertyName(Constants.PropertyName.Type)]
+    public string Type { get; set; }
```
yufeih, line 98: The name `type` is too general for TOC as TOC is language independent unlike in `MetadataItem`. How about `dotnetType`?

frarees: > language independent 

Does it make sense then that the name is not .NET-related? i.e., we don't include `dotnet` or alike as part of the name

> The name `type` is too general 

Makes sense. Thinking `symbolType` is more descriptive (similar to symbols in Roslyn's semantic model). More accurate than `memberType`, since values like `namespace` are not members.

In any case this might also mean that the `type` that's currently being exposed on inner nodes when using `memberLayout: "separatePages"` should also use `symbolType` instead of `type` for consistency - which might be seen as a breaking change?

I'll explore what changes would be needed to normalize into `symbolType`.

frarees: So consistently moving from `type` to `symbolType` (or alike) means quite some breaking changes, specially for templates. I'm not a good judge of whether this is a good or a bad idea - but by keeping `type` we can skip / delay the problem altogether (which seems reasonable for backwards compatibility to me).

From what I see `type` is generally used to refer to different concepts e.g., `ApiNames` or `MemberType`, so it's generally ambiguous as it stands now. It's also found referenced from `Constants` and other times as a literal - this is making current maintainability more laborous than needed (I don't have enough context on the reasons why it is currently like this, so take this with a grain of salt).

yufeih: Yep lets keep the name `type` to avoid breaking change. I thought `type` is a new fields in the TOC but it was already exposed.

The changes looks good as proposed. We can merge after publishing the PR and test pass.

frarees:
@yufeih any idea why the tests fail?

I've run them locally (just for .NET8, I cannot get some to work on .NET6 on my machine) and they look fine
image

The GitHub Action is returning with `Error: Process completed with exit code 1.`

filzrev:
> The GitHub Action is returning with `Error: Process completed with exit code 1.`

On GitHub Action CI. `TestPublicApiContract` test seems failed.

> 2024-07-10T09:23:06.0375360Z [xUnit.net 00:00:01.74]     Docfx.Tests.PublicApiContractTest.TestPublicApiContract [FAIL]

It needs following tasks to resolve CI error.
1. Run `TestPublicApiContract` test locally. 
2. Commit updated `test\docfx.Tests\Api.verified.cs` file.

filzrev:
PublicApiContractTest error seems be resolved by latest commit.

But PlantUML tests that use external PlantUML server failed with transient error.
It need to re-run CI.

yufeih (approved):
Thank you @frarees !

PR #10092, opened 2024-07-09
chore: Reduce `dotnet test` log messages

filzrev:
This PR change output log level of `dotnet test` commands.

Currently a lot of messages are outputted on `dotnet test` command.
And it's hard to find failed tests. (Need to grep `[FAIL]` from logs)

So this PR change MSBuild log level from `Normal` to `Minimal` (Output warnings/errors only)
And additionally output test results summary when test finished.

One problem of this PR is that **normal logs** are not output to the GitHub Actions logs.
If detailed test logs are needed in future.
It can be resolved by adding additional loggers (e.g. file logger or `trx` logger)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,9 +31,9 @@ jobs:
       shell: bash
       working-directory: templates
 
-    - run: dotnet test -c Release -f net8.0 --no-build --collect:"XPlat Code Coverage"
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Does anyone know an easy way of showing .NET test reports in GH actions similar to https://learn.microsoft.com/en-us/azure/devops/pipelines/test/review-continuous-test-results-after-build?view=azure-devops#view-test-results-in-build?

In absence of that, this PR looks good, and yes it indeed is difficult to see what test fails.

PR #10093, opened 2024-07-09
chore: Update docker file

filzrev:
This PR include following changes.

1. Specify install docfx version explicitly  
  Without this setting. Docker try to use `cached layer` if exists. In this case it need to manually clear cache.
2. Remove unused files to reduce docker image size
  2.1. `.nupkg` files
  2.2. Executables for .NET 6 
3. Additional Node.js installation (Required for #10066)
4. Pre-Install Chromium browser  

Note:  By `#10066`) changes. directory structures are not matched to `playwright.ps1` expecting structure.
So It need to create **temporary** symlink to run `playwright.ps1`.


yufeih (commented):
💯

yufeih (approved):

Issue #10087, opened 2024-07-08
[Bug] Footer and page size incorrect for custom printed layout


cjlotz:
**Describe the bug**
Using the new #9850 feature to specify a custom layout, the footer containing page numbers is not positioned correctly.  The page content is also not scaled correctly to fit onto the page

**To Reproduce**
1. Add custom CSS as in #9850 to force custom print layout
```css
@media print {
    @page {
        size: A4 landscape;
    }    
}
```
3. Build and run docfx to create output that includes pdf settings like the following sample `globalMetadata`
```json
    "globalMetadata": {
      "_appTitle": "On Key Release Notes",
      "_appFooter": "On Key Plus 2024.13.0 (Build 0)",
      "_appFaviconPath": "images/favicon.png",
      "_appLogoPath": "images/onkey_icon.png",
      "_enableNewTab": true,
      "_enableSearch": true,
      "_disableNavBar": "true",
      "_disableBreadcrumb": "true",
      "_disableToc": "false",
      "_disableAffix": "false",
      "pdf": true,
      "pdfTocPage": true,
      "pdfFileName": "onkey_releasenotes.pdf",
      "pdfPrintBackground": true
    },
```

**Generated output**

![image](https://github.com/dotnet/docfx/assets/1300847/32a643aa-b7b6-4103-960f-288ba94aa1da)

Notice that the page content is not scaled to fix onto the page (nr 2) and that the footer is not positioned correctly (nr 1) as illustrated above

**Context (please complete the following information):**
- OS: Windows
- Docfx version: 2.77.0

filzrev:
@cjlotz
What is the size of the generated PDF file page?

On my **Windows** environment.
Generated page size is displayed as `373 x 288 mm`. 
And it's not match `A4 landscape` (297 x210mm).

I've confirmed same issues when using Playwright `1.4.0` and without `PreferCSSPageSize` setting.
So I suspect environment-specific problems.

filzrev:
Thanks for your confirmation.

Page size problem is not related reported issues.
I've confirmed It's occurred on **Edge PDF Viewer**  and not occurred on Chrome/Acrobat Reader.

---

> the footer is still not positioned at the right side of the page being printed in A4 landscape orientation.

It seems PDF header/footer rendering is not using `PreferCSSPageSize` settings currently.
So default letter size setting is used for header/footer.
https://github.com/dotnet/docfx/blob/main/src/Docfx.App/PdfBuilder.cs#L203-L208

[...]

PR #10100, opened 2024-07-10
fix: PDF Header/Footer pages size

filzrev:
This PR intended to fix pdf footer problem reported at #10087

**Background**
PDF header/footer pages are separately generated from content page PDF. and merged with content pages.
And currently header/footer pages always using default `Letter` page size.
So when using custom page size. header/footer is displayed on wrong locations.

This PR change logics to use `Content Page PDF`'s page size for header/footer page generation.

**Test**
I've confirmed it works as expected with following page sizes.
- `size: A4 portrait;`
- `size: A4 landscape;`
- `size: 460mm 220mm`  (Custom size)
- `size: 220mm 460mm` (Custom size)

### DIFF
--- a/src/Docfx.App/PdfBuilder.cs
+++ b/src/Docfx.App/PdfBuilder.cs
@@ -18,6 +18,7 @@ using Spectre.Console;
 
 using UglyToad.PdfPig;
 using UglyToad.PdfPig.Actions;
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Looks great!

PR #10114, opened 2024-07-15
fix: missing delegate group name

frarees:
As part of #10090 `type` was introduced to the TOC.
However I've missed the `delegate` group name.

There's other names that are not present, but I think they are never exposed to the TOC?

* Variable
* Function
* Type Alias
* Member


yufeih (approved):
Looks good!

PR #10154, opened 2024-08-14
chore: Add logger methods

filzrev:

## 1. Add  `ResetCount` method to `Logger` class.

When running build with **multiple** times (e.g. `docfx watch` command). warning/error counts are accumulated.
Currently  it seems no way to reset these counters.

## 2.  Add `GetAllListeners`/`RegisterListeners` to `Logger`/`CompositeLogListener` class

These methods are required to **temporary** replace LogListeners.


yufeih (approved):
Thank you @filzrev

PR #10175, opened 2024-08-30
feat: add `AllowUnsafeBlocks` supports for source code based metadata generation

### LINKED ISSUE #10174: [Feature Request] Add support for `AllowUnsafeBlocks` in `properties`
Hi,
I would like to request a feature enhancement to add support for the `"AllowUnsafeBlocks"` within the `"properties"` section of `docfx.json`.

According to the [documentation](https://dotnet.github.io/docfx/reference/docfx-json-reference.html#properties), the `"properties"` section:

> Specifies an optional set of MSBuild properties used when interpreting project files. These are the same properties that are passed to MSBuild via the /property:name=value command line argument.

However, it seems that `"AllowUnsafeBlocks"` is currently not supported. From my understanding, adding support for this would require a small modification in the following file:

https://github.com/dotnet/docfx/blob/73d827970fdf183cb619b5bf344d5364cd4c2c29/src/Docfx.Dotnet/CompilationHelper.cs#L205-L212

Here’s a minimal working example that demonstrates the issue:

- `Program.cs`

```csharp
namespace Test
{
    public static class Program
    {
        public static unsafe void Main() { }
    }
}
```

- `docfx.json`

```json
{
  "metadata": [
    {
      "src": [{"src": "./", "files": ["Program.cs"]}],
      "dest": "api",
      "properties": { "AllowUnsafeBlocks": true }
    }
  ]
}
```

**Expected behavior:**

The `api` documentation should generate correctly without any errors or warnings.

**Current behavior:**

The build fails with errors/warnings, (unless one set `"allowCompilationErrors": true`).

---

Thank you for considering this request. I believe it would be a valuable addition.

Best,
Andrzej

filzrev:
This PR intended to fix #10174.

When generating metadata from **source files**.
MSBuild properties  need to be mapped to  related property. 

Currently `DefineConstants` is supported.
This PR add `AllowUnsafeBlocks` property support for C# source build (VB don't support `unsafe`)

### DIFF
--- a/docs/reference/docfx-json-reference.md
+++ b/docs/reference/docfx-json-reference.md
@@ -397,6 +397,13 @@ Specifies an optional set of MSBuild properties used when interpreting project f
 > [!Note]
 > Make sure to specify `"TargetFramework": <one of the frameworks>` in your docfx.json when the project is targeting for multiple platforms.
 
+> [!Note]
+> When generating metadata from source code files.
+> Supported properties are limited to the following.
+>   - `DefineConstants`
+>   - `AllowUnsafeBlocks`
+> If other properties are specified. These properties are ignored silently.
+
 ### `noRestore`
 
 Do not run `dotnet restore` before building the projects.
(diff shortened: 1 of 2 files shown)


yufeih (approved):
Thank you @filzrev !

PR #10185, opened 2024-09-04
feat: Add `CancellationToken` support for `RunBuild.Exec`

filzrev:
This PR add `CancellationToken` support to`RunBuild.Exec` method.
It's expected to be used by `docfx watch` command.

**What's Changed in this PR**

- Add `CancellationToken` parameter to methods and use token on operations that take times.
  - `for`/`foreach` operations
  - `Parallel.ForEach` operations
  - `AsParallel()` operations

- `DocumentBuildContext.cs`
  - Add `CancellationToken` property.
  - Change some constructor's access modifiers from `public` to `internal/private` that used by tests.

- `DocumentExceptionExtensions.cs`
  - Add logics to handle `AggregateException` on cancellation.

**Breaking Changes**

This PR change some public method signatures.
The user who using `Docfx.Plugins` package  need to update versions.

**Test**
Manually run cancellation tests with various delay patterns.
And confirmed operation is canceled with `OperationCanceledException` based exception (including `TaskCanceledException`)

### DIFF
--- a/src/Docfx.App/Helpers/DocumentBuilderWrapper.cs
+++ b/src/Docfx.App/Helpers/DocumentBuilderWrapper.cs
@@ -15,7 +15,7 @@ internal static class DocumentBuilderWrapper
 {
     private static readonly Assembly[] s_pluginAssemblies = LoadPluginAssemblies(AppContext.BaseDirectory).ToArray();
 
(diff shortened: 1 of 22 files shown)


yufeih (approved):
Thank you @filzrev !

PR #10213, opened 2024-09-18
Add Publish from Azure DevOps

kale-swapnil:


kale-swapnil:
@dotnet-policy-service agree

vanillajonathan:
Looks good!

There could also be instructions for deploying to Azure Static Web Apps instead of GitHub Pages.

yufeih (approved):
Thank you @kale-swapnil !

docs/docs/publish-azure-devops.md
```diff
@@ -0,0 +1,100 @@
+# Setting up DocFX with Azure DevOps to Deploy to GitHub Pages
```
yufeih, line 1: Do we also want this to appear in the [Table of Content](https://github.com/dotnet/docfx/blob/main/docs/docs/toc.yml)

kale-swapnil: I think so. Let me know if that needs to be done, and I will update it.

yufeih: Yes please

vanillajonathan:
You can also publish to Azure Static Web Apps.

https://learn.microsoft.com/en-us/azure/static-web-apps/get-started-portal?tabs=vanilla-javascript&pivots=azure-devops

filzrev:
> You can also publish to Azure Static Web Apps.

Azure Static Web Apps has [`Total File Count` quota](https://learn.microsoft.com/en-us/azure/static-web-apps/quotas) to `15,000`.
- `https://github.com/Azure/static-web-apps/issues/520`

So there is concern about reaching in this quota when deploying large document sites.  
(Especially when using `"memberLayout": "separatePages"` options)

PR #10217, opened 2024-09-20
feat: Add System.Text.Json serialization/deserialization supports

filzrev:
This draft PR intended to add functionality for **gradual migration** from `Newtonsoft.Json` to `System.Text.Json`. (Related issue #9727)

#### 1. chore: refactoring JsonUtility code to support newtonsoftjson/systemtxtjson switching based on type (fda9353)
 - Refactor `JsonUtility` related code to support switching NewtonsoftJson/SystemTextJson by type.

#### 2. chore: add json/yaml serialization test infrastructure (92a47a6)
 -  Add test infrastructure code to validate serialization/deserialization results.
   - Validate JSON serialization results has no diffs between Newtonsoft.Json/SystemTextJson. 
   - Validate JSON serialize/deserialize roundtrip result object comparison between Newtonsoft.Json/SystemTextJson. 
   - Validate YAML related roundtrip results. (`Object->YAML->Object` and `Object->JSON->Object`) 

#### 3. chore: switch XRefMap serialization logics to use System.Text.Json (643c89b)
- Switch `XRefMap` serialization/deserialization to use JsonUtility and `System.Text.Json`.
- Add  `XRefMap`  serialization/deserialization tests.

#### 4. chore: add json roundtrip test with JsonUtility (2fa823a)
- Add JSON serialization/deserialization round trip tests with JsonUtility

#### 5. chore: Add MarkdigExtensionSetting json roundtrip tests (7b3dd78)
- Add System.Text.Json based serialization/deserialization for `MarkdigExtensionSetting`.

#### 6. chore: add ManagedReference YAML roundtrip tests (9492ef8)
- Add System.Text.Json based serialization/deserialization for `ManagedReference.PageViewModel` 
  -  Modify CompositeDictionary related property to add `private init` and  `JsonInclude` attribute
     (It's required because `System.Text.Json` can't deserialize getter-only property)

#### 7. chore: add UniversalReference YAML roundtrip tests (12c7394)
- Add System.Text.Json based serialization/deserialization for `UniversalReference.PageViewModel` 

#### 8. chore: add ApiPage YAML roundtrip tests (448c933)
- Add Yaml serialization/deserialization tests for `ApiPage` 

#### 9. chore: add FilterConfig YAML roundtrip tests (a9fde45)
- Add Yaml serialization/deserialization tests for `FilterConfig`
- Add ‘[JsonIgnore]‘ attribute to `Rule` getter-only property.

#### 10. chore: Add Manifest model json roundtrip tests (9c13117)
[...]

PR #10232, opened 2024-09-24
chore: Add `Equals`/`GetHashCode` method override to ApiPage's models

filzrev:
This PR add `object.Equals`/`object.GetHashCode` method override for following ApiPage types.
- `OneOfBase` derived class that have array reference type template parameter.
-  `class`  that is used by `OneOfBase` template parameter.
- `struct` that contains reference type properties.

**Background**
These changes are required for testing [object graph comparison with  FluentAssertions' `BeEquivalentTo` method](https://fluentassertions.com/objectgraphs)).

`OneOf` library override `object.Equals` method. And that [compare value with `object.Equals`](https://github.com/mcintyre321/OneOf/blob/6e02dbe75d0f20f198c640d4c04190a85c5ac9e6/OneOf/OneOfBaseT1.generated.cs#L103-L124)
So when `OneOf` template parameter contains `reference type`. 
It's compared with `reference equality` by default.

This PR intended to change these behaviors to compare members by value.

### DIFF
--- a/src/Docfx.Build/ApiPage/ApiPage.cs
+++ b/src/Docfx.Build/ApiPage/ApiPage.cs
@@ -1,6 +1,7 @@
 ﻿// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 using OneOf;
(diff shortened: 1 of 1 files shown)


yufeih:
This feels like it adds extra burden on maintaining this one of types. If it is just for testing, would it be ok to compare the JSON serialized strings for object equivalent tests?

filzrev:
> If it is just for testing, would it be ok to compare the JSON serialized strings for object equivalent tests?

Yes. It's expected to be used for testing purpose only. that test using FluentAssetions `BeEquivalentTo`.
Certainly this PR change would include `extra code` for testing.

So I've changed test code to compare serialize YAML results.
It's not ensure `serialization results YAML` is expected to be same as original data. (It might to be dropped some metadata)
But for almost data, we can conclude that there is no problem by comparing serialized results .

https://github.com/dotnet/docfx/pull/10217/commits/fe75b7d6cd3c7b06085e336dbf51e4b1d140a1af#diff-7f2ad43badc5aed8fd8543eda978a49074baf670d7b9cd3f5b2d4640109612faR52-R63

PR #10233, opened 2024-09-24
Update Version in "Use the NuGet Library"

RiPont:
Update the version in the example for "Use the NuGet Library".

Add warning to use only the exact version.

Add necessary line to example for GenerateManagedReferenceYamlFiles.


yufeih (approved):
Thank you @RiPont !

PR #10234, opened 2024-09-25
chore: Remove `OneOf.SourceGenerator` runtime dependency

filzrev:
This PR remove runtime dependency to `OneOf.SourceGenerator` of following packages.

- https://www.nuget.org/packages/Docfx.Build#dependencies-body-tab
- https://www.nuget.org/packages/Docfx.DotNet#dependencies-body-tab


yufeih (approved):
Thank you @filzrev

PR #10239, opened 2024-09-28
feat: Add .NET 9 support (and drop .NET 6 support)

filzrev:
This PR including following changes.
- Add .NET 9 support (.NET 9,0 RC is already released as `go-live`)
- Drop .NET 6 support (Support ends at `November 12, 2024`)

Note: 
The SDK for Self-Contained exe zips will remain .NET 8.
Because it's .NET 8 is `LTS` release and longer support lifetime.


paulushub:
For a tool, is there any real need to even drop support for the framework before the support ends?
Many applications and libraries will still support the framework after the official end of support.

filzrev:
I'm expecting this PR is merged and newer version is released after .NET 9 RTM release.  
(or .NET 9 version of Roslyn packages are released).

The reason I've dropped `.NET 6` supports is following.

1. `Microsoft.CodeAnalysis.Workspaces.MSBuild` supports only latest 2 versions (.NET 6 support is already dropped)  
    So when continuing .NET 6 supports. It need to reference older versions and it introduce additional complexity.
2. Published NuGet tool package size issue (https://github.com/dotnet/docfx/issues/9396).  
    .NET tools package contains files per `TargetFramework`.  
    So it increase package size when continuing support .NET 6.
3. Local disk usage issue
   Currently when building/publishing on local dev environment. It consume a lot of disks spaces.
   (On my environment it takes over 10GB disk spaces).

paulushub:
> 1. `Microsoft.CodeAnalysis.Workspaces.MSBuild` supports only latest 2 versions (.NET 6 support is already dropped)
If I understand it well, this normally should not be an issue. Reference the version that supports the .NET 6 for the .NET 6 build.

> 2. Published NuGet tool package size issue ([Exploded package size and quality concerns #9396]
> 3. Local disk usage issue
>    Currently when building/publishing on local dev environment. It consume a lot of disks spaces.
>    (On my environment it takes over 10GB disk spaces).

Just updated my version to `2.77.0` and here are the stats of the installation:
**Size**: 0.99 GB (1,063,444,125 bytes)
**Size on disk**: 0.99 GB (1,065,295,872 bytes)
**Contains**: 969 Files, 113 Folders

If data are later downloaded, I will assume these are not Framework dependent. Again, I might be missing something.

yufeih:
I agree with dropping unsupported .NET versions to reduce unneeded maintainance effort. This will to be merged and released after November 12, 2024.

filzrev:
> 3. Local disk usage issue

Please ignore the above comments.
It's **build environment** issues. and this problem is now mitigated by #10066.

For docfx users. keeping .NET 6 support  requires about `37MB` extra local disk spaces. and increase  about `16MB` on `nupkg` file size.

In any case, I would like to drop .NET 6 support to lower maintenance costs with following reasons

- It takes extra LocalBuild/CI test time. 
- Remove Conditional NuGet package references (for Roslyn APIs)
- Reduce `#ifdef` directive usage
- Using newer .NET APIs (e.g. span-based overload, [Frozen Collection](https://learn.microsoft.com/en-us/dotnet/api/system.collections.frozen))

---

For users who continue to using .NET 6. It need to use version that supports .NET 6(`2.77.0`)

yufeih (approved):

martincostello:
What's the ETA on a new version including this change being published to NuGet.org?

yufeih:
@martincostello  A new release has been published to NuGet.org today with .NET 9 support.

PR #10242, opened 2024-09-30
Update ci.yml for potential security issue.

markjulmar:
Replace pull_request_target with pull_request to mitigate potential vulnerability.


yufeih (approved):
This would break percy visual regression testing for PRs from forked repo, but yes lets update the security problem first.

PR #10245, opened 2024-10-01
chore: Fix failed  test for PlantUML

filzrev:
This PR intended to fix CI test errors that occurred since today.

PlantUML Online Server sometimes returns additional scripts tags when requesting SVG data.
So adding temporary workaround code to remove these script tags.

See: `https://github.com/plantuml/plantuml-server/issues/342`


filzrev:
Close this issue.
Because it can no longer be reproduced at this time.

 PR(#10243 #10244) failed tests will be passed by re-running CI workflow

PR #10272, opened 2024-10-16
fixed article a/an in config.md

MorkovnySok:


yufeih (approved):
Thank you @MorkovnySok

PR #10279, opened 2024-10-17
remove redundant dictionary lookups

SimonCropp:


yufeih (approved):

src/Docfx.Dotnet/ManagedReference/Resolvers/ResolveReference.cs
```diff
@@ -52,14 +52,11 @@ public void Run(MetadataModel yaml, ResolverContext context)
 
     private static void TryAddReference(ResolverContext context, MetadataItem page, List<ReferenceItem> addingReferences, string key)
     {
-        if (!page.References.ContainsKey(key))
+        if (context.References.TryGetValue(key, out ReferenceItem item))
```
yufeih, line 55: looks like a test fails due to removing the page references dictionary check

paulushub:
From 
```cs 
 if (!mergeTo.ContainsKey(pair.Key))
  {
      mergeTo[pair.Key] = pair.Value;
  }
```
```cs
mergeTo.TryAdd(pair.Key, pair.Value);
```
The implementations are different, which is faster?

yufeih:
> From:
> 
> ```cs
> if (!mergeTo.ContainsKey(pair.Key))
> {
>       mergeTo[pair.Key] = pair.Value;
> }
> ```
> 
> To:
> 
> ```cs
> mergeTo.TryAdd(pair.Key, pair.Value);
> ```
> 
> The implementations are different, which is faster?

`TryAdd` is faster, it computes hash code once.

PR #10282, opened 2024-10-17
simplify some exception handling

SimonCropp:


src/Docfx.Plugins/DocumentExceptionExtensions.cs
```diff
@@ -23,8 +23,9 @@ public static TResult[] RunAll<TElement, TResult>(this IReadOnlyList<TElement> e
                 results[i] = func(elements[i]);
             }
             catch (DocumentException ex)
+                when (firstException == null)
             {
-                firstException ??= ex;
+                firstException = ex;
```
yufeih, line 28: Does this change the exception handling behavior when multiple DocumentExceptions are thrown? Before the change the second DocumentException is ignored.

SimonCropp: ahh yes. i need to swallow the subsequent. i will revert this part. given the previous code would be cleaner than adding the swallow

yufeih (approved):

PR #10285, opened 2024-10-17
use some pattern matching

SimonCropp:


src/docfx/Models/BuildCommand.cs
```diff
@@ -110,7 +110,7 @@ void SetGlobalMetadataFromCommandLineArgs()
 
             static (string key, object value) ParseMetadata(string metadata)
             {
-                if (metadata.IndexOf('=') is int i && i < 0)
+                if (metadata.IndexOf('=') is int i and < 0)
```
yufeih, line 113: This change leads to:
> error CS0165: Use of unassigned local variable 'i' [D:\a\docfx\docfx\src\docfx\docfx.csproj::TargetFramework=net8.0]

PR #10300, opened 2024-10-19
fix: change link as plaintext if API is excluded

filzrev:
This PR intended to fix problem reported at #10287.

When class is excluded. link target document is not exists. 
So link to excluded symbol should be rendered as **plain text**. 

**What's changed in this PR**
It needs `SymbolFilter::IncludeApi` instance to check specific symbol is excluded or not.
So it need to pass `SymbolFilter` instance to API that generate URL.

- Add  `SymbolFilter filter` parameter to several APIs. 
- Change `url` generation logics to check `filter.IncludeApi(symbol)` and set null when symbol is excluded.
- Remove `SymbolVisitorAdapter adapter` parameter (that seems not be used).

**Test**
Tested on local environment with following output formats.
- `mref`
- `markdown`
- `apiPage`

### DIFF
--- a/src/Docfx.Dotnet/SymbolFormatter.cs
+++ b/src/Docfx.Dotnet/SymbolFormatter.cs
@@ -138,9 +138,7 @@ internal static partial class SymbolFormatter
             {
                 Name = overload ? VisitorHelper.GetOverloadId(symbol) : VisitorHelper.GetId(symbol),
                 DisplayName = part.ToString(),
-                Href = filter.IncludeApi(symbol)
-                        ? SymbolUrlResolver.GetSymbolUrl(symbol, compilation, memberLayout, urlKind, allAssemblies)
-                        : null,
+                Href = SymbolUrlResolver.GetSymbolUrl(symbol, compilation, memberLayout, urlKind, allAssemblies, filter),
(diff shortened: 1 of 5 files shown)


src/Docfx.Dotnet/SymbolFormatter.cs
```diff
@@ -138,7 +138,9 @@ LinkItem ToLinkItem(SymbolDisplayPart part)
             {
                 Name = overload ? VisitorHelper.GetOverloadId(symbol) : VisitorHelper.GetId(symbol),
                 DisplayName = part.ToString(),
-                Href = SymbolUrlResolver.GetSymbolUrl(symbol, compilation, memberLayout, urlKind, allAssemblies),
+                Href = filter.IncludeApi(symbol)
+                        ? SymbolUrlResolver.GetSymbolUrl(symbol, compilation, memberLayout, urlKind, allAssemblies)
```
yufeih, line 142: Should we add the `SymbolFilter` to the `GetSymbolUrl` method to ensure it’s always included?

filzrev: I've moved `SymbolFilter` check logics to `GetSymbolUrl` by db035fe

yufeih (approved):

yufeih:
Thank you for the fix!

PR #10318, opened 2024-10-23
test: Update test snapshot

yufeih:


yufeih:
Snapshot testing seems broken

filzrev:
These diffs are caused by changing JSON serialize from NewtonsoftJson to `System.Text.Json`.

```
var value = 1.0d;
Console.WriteLine(System.Text.Json.JsonSerializer.Serialize(value));   // Output: `1`
Console.WriteLine(Newtonsoft.Json.JsonConvert.SerializeObject(value)); // Output: `1.0`
```

Both result are correct JSON representation.
Though it might be better to serialize as `1.0`. (by adding custom JSON converter)

Because when JSON deserialized as `Dictionary<string, object>`.
Deserialized value type is inferred by number text representation.

https://github.com/dotnet/docfx/blob/af68cbdd73b803a010fd8c22ad64148bf6337a8f/src/Docfx.Common/Json/System.Text.Json/ObjectToInferredTypesConverter.cs#L30-L35

yufeih:
I understand where the diff is coming from, what I'm trying to figure out is why these diffs didn't fail the PRs that introduced them.

filzrev:
As comment at #10156
Currently CI don't validate snapshot files because `SNAPSHOT_TEST` is not set.

PR #10341, opened 2024-11-03
use char base startswith and endswith

SimonCropp:
has better perf

### DIFF
--- a/src/Docfx.Build.OverwriteDocuments/OverwriteUtility.cs
+++ b/src/Docfx.Build.OverwriteDocuments/OverwriteUtility.cs
@@ -23,7 +23,7 @@ public static class OverwriteUtility
             throw new ArgumentException("OPathString cannot be null or empty.", nameof(OPathString));
         }
 
-        if (OPathString.EndsWith("/", StringComparison.Ordinal))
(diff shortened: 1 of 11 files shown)


filzrev:
These modifications seems to corresponding to following `Code Analyzer Rules`
- [CA1865/CA1866](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1865-ca1867)

I thought it might be better to change these rules category from `suggestion` to `warning` at `.editorconfig` file.
By these changes. It can enforce rules for future code modifications.
And it can automatically refactoring code by executing `dotnet format analyzers` command.

**Example settings for .editorconfig**
```
#----------------------------------------------
# CodeAnalyzer Rules
#----------------------------------------------
[*.{cs}]

# Performance rules
dotnet_diagnostic.CA1865.severity = warning # CA1865: Use 'string.Method(char)' instead of 'string.Method(string)' for string with single char
dotnet_diagnostic.CA1866.severity = warning # CA1866: Use 'string.Method(char)' instead of 'string.Method(string)' for string with single char
```

yufeih (approved):

PR #10353, opened 2024-11-04
avoid cast in loop and null ref warning in TestKey

SimonCropp:


src/Docfx.Build.SchemaDriven/Processors/Merger.cs
```diff
@@ -135,10 +135,13 @@ private static bool TestKey(object source, object overrides, BaseSchema schema)
...
             return false;
         }
+
+        var sourceDictionary = (IDictionary<string, object>)source;
+        var overridesDictionary = (IDictionary<string, object>)overrides;
         return schema?.Properties != null && schema.Properties.Any(p => p.Value.MergeType == MergeType.Key) && schema.Properties.Where(p => p.Value.MergeType == MergeType.Key).All(p =>
         {
-            (source as IDictionary<string, object>).TryGetValue(p.Key, out var s);
-            (overrides as IDictionary<string, object>).TryGetValue(p.Key, out var o);
+            sourceDictionary.TryGetValue(p.Key, out var s);
```
yufeih, line 143: test failed and it appear this cast is only valid after `schema?.Properties != null && schema.Properties.Any(p => p.Value.MergeType == MergeType.Key)`

PR #10395, opened 2024-11-17
feat: Use PlantUML Online Server as default configs

filzrev:
This PR change `PlantUmlOptions` default settings.

**Background**
When using `PlantUML` diagrams.
It need to explicitly set  `build/markdownEngineProperties/plantuml` configs.
https://dotnet.github.io/docfx/docs/markdown.html#settings

Currently when above config is not defined. following exception is thrown.

> error PlantUmlExtension: Value cannot be null. (Parameter 'uriString')

This PR change default settings to use PlantUML Online server as default options.


yufeih (approved):
Looks great!

PR #10407, opened 2024-11-20
chore: Update `Dockerfile`

filzrev:
This PR update [Dockerfile](https://github.com/dotnet/docfx/blob/main/Dockerfile)

**What's changed**

1. Update base OS image version.
2. Update docfx version to `2.78.1`
3. Remove `apt-get update` command. And use `apt-get install` with `--update` option instead (that is added on Ubuntu 24)
4. Change library dependencies installation to use `--with-deps` options.
5. Remove temporary `symbolic link` steps. (Currently `.playwright` directory is placed under bin)

**Test**
I've confirmed docker image successfully build.
And `docfx pdf` command is works as expected with following steps.

```
docker build ./ -t docfx

docker run -it --rm --entrypoint /bin/bash docfx
docfx init --yes
docfx build
docfx pdf
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,26 @@
-FROM mcr.microsoft.com/dotnet/sdk:8.0-bookworm-slim
+FROM mcr.microsoft.com/dotnet/sdk:8.0-noble
 
 # Add dotnet tools to path.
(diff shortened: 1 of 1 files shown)


yufeih (approved):
Thank you @filzrev

PR #10436, opened 2024-11-28
feat: Allow extensionless URL HTML file access

filzrev:
This PR is intended to add support for extensionless HTML URLs for `docfx serve` command.

**Background**
`GitHub Pages` support extensionless URLs. (e.g. https://dotnet.github.io/docfx/docs/basic-concepts)
And other static HTML hosting site also have similar features.

So I thought it's reasonable to support extensionless URLs with the `docfx serve` command.

Extensionless URL support is partially related to #2865.
It might be better to add options to output extensionless URLs for docfx internal links.

### DIFF
--- a/src/Docfx.App/RunServe.cs
+++ b/src/Docfx.App/RunServe.cs
@@ -2,14 +2,18 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 using Docfx.Common;
 using Docfx.Plugins;
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.FileProviders;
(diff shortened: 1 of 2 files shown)


src/Docfx.App/RunServe.cs
```diff
@@ -44,6 +48,7 @@ public static void Exec(string folder, string host, int? port, bool openBrowser,
             Console.WriteLine($"Serving \"{folder}\" on {url}");
             Console.WriteLine("Press Ctrl+C to shut down");
             using var app = builder.Build();
+            app.UseExtensionlessHtmlUrl();
```
yufeih, line 51: Should this be opt-in? I think the default behavior should remain a standard static file server so it can be hosted by a variety of web servers.

filzrev: This settings is works on `docfx serve` command only.
And not affect generated HTML files.

So I though it's not a problems by enabled default. (And not support config to disable)
When supporting feature that rewrite URL to `extensionless URL` automatically.
It should be supported as user's opt-in features.

yufeih (approved):

Issue #9870, opened 2024-04-13
[Bug] docfx pdf command don't accept `Ctrl+C` command interruption


filzrev:
**Describe the bug**
When running `docfx pdf` command is executed.
progress rate is displayed by using `AnsiConsole.Progress()`.

And in this state. `Ctrl+C` command interruption does not seem to be accepted.

**To Reproduce**
1. Run `docfx pdf` command
2. Try to interrupt command with `Ctrl+C`

**Expected behavior**
Command operation is immediately stopped.

**Context (please complete the following information):**
- OS: Windows
- Docfx version: [e.g. 2.76.0]

filzrev:
Close this issue.
Because it's resolved by #10460

PR #10460, opened 2024-12-14
chore: Add cancellation support for PdfBuilder operation

filzrev:
This PR intended to fix issue #9870.

Currently when running `docfx pdf` command.
`Ctrl+C` command interruption is not accepted.
So I've added codes to  manually cancel running operations.

> [!NOTE]
> Playwright don't supports cancellationToken parameters.
> So it need manually cancel operation by dispose playwright context.
>
> `Specre.Console` don't support native cancellation support .
> So PdfCommand/DefaultCommand is not modified. (It need to migrate to `AsyncCommand` in future)
> `https://github.com/spectreconsole/spectre.console/issues/701`

**What's changed in this PR**
1. Add cancellationToken parameter to PdfBuilder's public APIs
2. Add `Console.CancelKeyPress` event handlers to cancel running operations
3. Add `cancellationToken` parameter to following methods
  3.1. WebApplication.Start
  3.2. `ParallelOptions` parameter of `Parallel.Foreach`
4. Call `cancellationToken.ThrowIfCancellationRequested()` inside loop.
5. Rename `outputPath` parameter to `pdfOutputPath`.

### DIFF
--- a/src/Docfx.App/PdfBuilder.cs
+++ b/src/Docfx.App/PdfBuilder.cs
@@ -51,17 +51,17 @@ static class PdfBuilder
         public string? pdfFooterTemplate { get; init; }
     }
 
-    public static Task Run(BuildJsonConfig config, string configDirectory, string? outputDirectory = null)
+    public static Task Run(BuildJsonConfig config, string configDirectory, string? outputDirectory = null, CancellationToken cancellationToken = default)
     {
         var outputFolder = Path.GetFullPath(Path.Combine(
             string.IsNullOrEmpty(outputDirectory) ? Path.Combine(configDirectory, config.Output ?? "") : outputDirectory,
             config.Dest ?? ""));
 
         Logger.LogInfo($"Searching for manifest in {outputFolder}");
-        return CreatePdf(outputFolder);
+        return CreatePdf(outputFolder, cancellationToken);
     }
 
-    public static async Task CreatePdf(string outputFolder)
+    public static async Task CreatePdf(string outputFolder, CancellationToken cancellationToken = default)
     {
         var stopwatch = Stopwatch.StartNew();
         var pdfTocs = GetPdfTocs().ToDictionary(p => p.url, p => p.toc);
@@ -82,7 +82,7 @@ static class PdfBuilder
         using var app = builder.Build();
         app.UseServe(outputFolder);
         app.MapGet("/_pdftoc/{*url}", TocPage);
-        await app.StartAsync();
+        await app.StartAsync(cancellationToken);
 
         baseUrl = new Uri(app.Urls.First());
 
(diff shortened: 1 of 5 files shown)


src/Docfx.App/PdfBuilder.cs
```diff
@@ -94,27 +94,56 @@ public static async Task CreatePdf(string outputFolder)
...
+                    pageNumbers => pdfPageNumbers[url] = pageNumbers,
+                    cts.Token);
 
                 task.Value = task.MaxValue;
                 task.StopTask();
             });
         });
 
+        // Wait pdfBuildTask completed or cancelled.
+        await Task.WhenAny(pdfBuildTask, Task.Delay(Timeout.Infinite, cts.Token));
```
yufeih, line 123: Is there a reason we are not relying on the cancellation token passed to `pdfBuildTask` to gracefully cancel the PDF build job? This extra manual shutdown code looks suspicious to me.

src/Docfx.App/PdfBuilder.cs
```diff
@@ -641,4 +681,16 @@ private static StringComparison GetStringComparison()
             ? StringComparison.OrdinalIgnoreCase
             : StringComparison.Ordinal;
     }
+
+    private static CancellationTokenRegistration SubscribeCancelKeyPressEvent(CancellationTokenSource cts)
```
yufeih, line 685: This seems to apply to all build jobs, do we want to also enable it on the build command say in the future?

src/Docfx.App/PdfBuilder.cs
filzrev: It's required because `AnsiConsole.Progress().StartAsync` and Playwright APIs doesn't support `CancellationToken` parameter.

So when operation is cancelled via CancellationToken.
It takes some seconds(about 5-10 seconds) until operations are cancelled.

So I need to use `Task.WhenAny`  for stopping operations immediately.
And additional dipose/await code are required.
Because It can't output logs until `AnsiConsole.Progress().StartAsync` is completed  
(Otherwise console log is outputted to wrong places)

src/Docfx.App/PdfBuilder.cs
filzrev: I'm expecting same cancellation token should be passed to Build command also in future.

filzrev:
**What's changed on latest commit.**

**1. `CancellableCommandBase.cs`**
Add abstract `CancellableCommandBase` class to support `CancellationToken`.
It subscribe following signal events and pass cancellation token to command. 
 - `SIGINT`: It happens on `Ctrl+C` key pressed.
 - `SIGQUIT`: It happens on `Ctrl+Break` key pressed.
 - `SIGTERM`: It used on `docker stop` for example (On Windows environment it's not used)

See details: `https://github.com/spectreconsole/spectre.console/issues/701`

**2. `PdfBuilder.cs`**
2.1. Remove `Console.CancelKeyPress` event subscription. Use `SIGINT` signal instead.
2.2. Modify code to use `cancellationToken` that is passed as argument.
2.3. Modify code to use `.WaitAsync(cancellationToken) instead of `Task.When`.

**3. DefaultCommand.cs**
**4. PdfCommand.cs**
Modify code to  use `CancellableCommandBase`.

**TODO**
Following tasks are not scope of this PR.
- Implement **Graceful Shutdown**.
- Add `CancellationToken` support for non-PDF relating commands (Build/Metadata)

yufeih (approved):

PR #10375, opened 2024-11-08
chore: Update xUnit related packages version to `v3`

filzrev:
This is **Draft** PR to migrate unit test projects to `xUnit v3`.
(`xUnit v3` is expected to be released in the near future)

### What's changed to this PR

**1. Switch xUnit.NET related NuGet packages to v3.**
  - `Verify.Xunit` -> `Verify.XunitV3`
  - `xunit` -> `xunit.v3`

**2. Add additional output log types (`*.log`/`*.ctrf`)**

**3. Change unittest's projects `OutputType from `Library` to `Exe`.**

**4. Enable `Invariant Globalization` mode for test projects.**
xUnit.v3 run unit tests as standard exe.
So it can configure setting with `InvariantGlobalization` MSBuild property.

**5. Add `Microsoft.Testing.Platform` related settings**
[Microsoft.Testing.Platform](https://learn.microsoft.com/en-us/dotnet/core/testing) is alternative unit test execution layers.
Currently It' used when running tests with following ways. 
  - When running test with `dotnet test` command. 
  - When running tests with [Visual Studio **Preview**](https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-platform-vs-vstest#communication-protocol-preview).

**6.Add `xunit.runner.json` config file for test project.**

**7.Pass `TestContext.Current.CancellationToken` that require Cancellation token.**

**8.Modify `MetadataCommandTest.cs` And add steps to create dummy `Directory.Build.props` file.**

PR #10474, opened 2024-12-19
chore: Update xUnit related package version to v3

filzrev:
This is **Draft** PR to migrate unit test projects to `xUnit v3`. (Prev draft PR is #10375)
Currently test execution on VS is unstable

### What's changed to this PR

**1. Switch xUnit.NET related NuGet packages to v3.**
  - `Verify.Xunit` -> `Verify.XunitV3`
  - `xunit` -> `xunit.v3.mtp-v2`

**2. Add additional output log types (`*.log`/`*.ctrf`)**

**3. Change unittest's projects `OutputType` from `Library` to `Exe` and Explicitly set `IsTestProject` property**

**4. Add `Microsoft.Testing.Platform` related settings**
[Microsoft.Testing.Platform](https://learn.microsoft.com/en-us/dotnet/core/testing) is alternative unit test execution layers.

**6. Add `xunit.runner.json` config file for test project.**

**7. Pass `TestContext.Current.CancellationToken` that require Cancellation token.**

**8. Modify `MetadataCommandTest.cs` And add steps to create dummy `Directory.Build.props` file.**
It's required to ignore project directories `Directory.Build.props` settings

**9. Add empty  `Directory.Build.props` file to samples directory**

**10. ~~Modify CI workflow to use custom `seed` value to reproduce issue on CI~~**
~~By default. xUnit determine `seed` on **build time** and seed value is different between assemblies. ~~
~~So to fix seed value. Use `github.run_number` as a seed value. (It's incremented from 1)~~
seed parameter is seems not used for deterministic test orders.
On current xUnit implementation `ModuleVersionId` is used instead.

**11. Add `UseCustomBranchNameAttribute` that derived from `BeforeAfterTestAttribute`**
When running `SeedMarkdown` tests before other snapshot tests.
It cause snapshot diffs because `GitUtility.cs` caching **actual branch name**.
To simplify environment variable set/restore.
Add custom `BeforeAfterTestAttribute` for this purpose.

**12. Add `[assembly:CaptureConsole]` attribute to test assemblies.**

**13. Add **ProcessHelper** class to invoke external command (e.g. dotnet, docfx).**
When using `Process.Start` to launch external command that stdout.
Running tests is not completed when running tests on Test Explorer.
So I've added utility code to ignoring stdout/stdout.

**15. Modify `dotnet test --filter Stage=Percy` command. to use `-- --filter-trait "Stage=Percy"`**
When `TestingPlatformDotnetTestSupport` is enabled. `dotnet test` command's  `--filter` argument is silently ignored. (Same behavior as that reported at `https://github.com/microsoft/testfx/issues/4401`)
It need modify command to pass additional argument(`-- --filter-trait "Stage=Percy"`)

**16. Separate `Chromium Headless Shell` installation from test steps**
Chromium is installed on when running `docfx pdf` command first time.
It take times and it affects test execution times.
So creating separate chromium install step that is executed before running `dotnet test`.

https://github.com/dotnet/docfx/pull/10474/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR43-R48

**17. Add `-nodereuse:false` and `--no-dependencies` parameter to `dotnet build` command inside tests**
After migrated to xUnit v3.
`SamplesTest::Extensions` `dotnet build` command takes extra 15 minutes on Windows CI.
It seems relating to `https://github.com/dotnet/sdk/issues/9452` so I've added  `-nodereuse:false`  parameter to resolve issue.
And adding `--no-dependencies` parameter to suppress `Docfx.App` project build during test.

**18. Add `dotnet-coverage` command related settings.**
Currently Microsoft.Testing.Platform don't support coveret-based codecoverage with `dotnet test` command.

**19. Modify custom Fact  constructors to suppress xUnit3003 warnings.**

**20. Add `Directory.Build.targets` and `tests/Directory.Build.targets`**
These files are used for settings that need to be set after loading project file. 

**21. Add `global.json` and enable MTP mode.**
This setting is required to use `Microsoft.Testing.Platform` with `dotnet test`.
https://learn.microsoft.com/en-us/dotnet/core/testing/unit-testing-with-dotnet-test#microsofttestingplatform-mtp-mode-of-dotnet-test

**22. Modify .gitattributes file to use LF for verified files**
Latest version of `Verify.XunitV3` enforcing to use LF for verified files by default,
So I've added related settings to `.gitattributes`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -7,3 +7,6 @@ templates/** eol=lf
 
 # VerifyTests
 *.verified.txt text eol=lf working-tree-encoding=UTF-8
+*.verified.json text eol=lf working-tree-encoding=UTF-8
+*.verified.yml text eol=lf working-tree-encoding=UTF-8
+*.verified.md text eol=lf working-tree-encoding=UTF-8
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,9 @@ on:
     branches: [ main, feature/*, hotfix/* ]
   workflow_dispatch:
 
+env:
+  NO_COLOR: true
+
 jobs:
   test:
     runs-on: ${{ matrix.os }}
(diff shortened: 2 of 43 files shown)


filzrev:
- Update xunit.net version to 4.0.1
- Remove `--no-progress` parameter. (It's no more needed because `--progress auto` is added on MTP side)
- Modify CTRF related command parameter names. (It's changed on xunit.v3 4.0.0 but it's not updated)

filzrev:
On Windows CI.
[Pack tool for distribution tests](https://github.com/dotnet/docfx/blob/main/.github/workflows/ci.yml#L65) step taking too much time(over 18 min) and it cause workflow timeout (30min)

I don't know what changes are causing this behavior.

#### Log
```
Run dotnet pack src/docfx -c Release --no-restore /p:Version=0.0.0-packaging-test -o drop/package-test
  Docfx.YamlSerialization -> D:\a\docfx\docfx\src\Docfx.YamlSerialization\bin\Release\net10.0\Docfx.YamlSerialization.dll
  Docfx.Plugins -> D:\a\docfx\docfx\src\Docfx.Plugins\bin\Release\net10.0\Docfx.Plugins.dll
Attempting to cancel the build...
Error: The operation was canceled.
```

filzrev:
I've added following temporary workaround code to avoid `dotnet pack` command cause timeout issue.
> dotnet build-server shutdown

Root cause of this issue is still unknown.
I've confirmed binlog on `dotnet pack` and it looks like freezed at `ResolveProjectReferences` step.

vicancy (approved):

PR #10540, opened 2025-02-14
feat: Add `.slnx` solution format metadata generation support

filzrev:
This PR add support metadata generation for `.slnx` solution format.

Note:  
It requires `Microsoft.Build` package version `17.13.9` or later. (Requires .NET 9).   
So it works only when running docfx with .NET 9.

**Test**
It's manually tested by following steps.
1. Convert `samples\seed\dotnet\solution\BuildFromSolution.sln` to `.slnx` format by using Visual Studio.  
2  Running `docfx metadata` command with generated `.slnx` solution file.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -25,6 +25,12 @@
     <PackageVersion Include="YamlDotNet" Version="15.3.0" />
   </ItemGroup>
 
+  <!-- .slnx solution format is supported Microsoft.Build 17.13.9 or later. -->
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.Build" Version="[17.11.4]" Condition="'$(TargetFramework)' == 'net8.0'"/>
+    <PackageVersion Include="Microsoft.Build" Version="17.13.9"   Condition="'$(TargetFramework)' != 'net8.0'"/>
+  </ItemGroup>
+
   <ItemGroup>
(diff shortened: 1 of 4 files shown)


yufeih (approved):

yufeih:
Thank you @filzrev

Directory.Packages.props
```diff
@@ -25,6 +25,12 @@
     <PackageVersion Include="YamlDotNet" Version="15.3.0" />
   </ItemGroup>
 
+  <!-- .slnx solution format is supported Microsoft.Build 17.13.9 or later. -->
+  <ItemGroup>
+    <PackageVersion Include="Microsoft.Build" Version="[17.11.4]" Condition="'$(TargetFramework)' == 'net8.0'"/>
```
WeihanLi, line 30: Is it possible to use `17.13.9` for `net8.0` so that it can also be supported for `net8.0`?

filzrev: `Microsoft.Build` 17.13.9 package is published as .NET 9 target. 
https://www.nuget.org/packages/Microsoft.Build/17.13.9#supportedframeworks-body-tab

So It's not possible to reference this version from .NET 8 applications.

WeihanLi: 👍 thanks for the info

PR #10561, opened 2025-02-26
chore: add .net10 supports for nightly build

filzrev:
This PR add .NET 10 target for nightly build CI.

**What changed in this PR**

#### 1. `.github/workflows/nightly.yml`
- Add build/test settings for .NET 10 (and .NET 8 (LTS))
- Remove `Report failed tests` step. 

#### 2. `.github/workflows/reports.yml`
- Add `nightly` build CI as trigger

#### 3. `Directory.Build.props`
- Modify nightly build CI target framework to .NET8(LTS) / .NET 10

#### 4. Other changes
When build project with .NET SDK 10 Preview.
[NU1511 warnings](https://learn.microsoft.com/en-us/nuget/reference/errors-and-warnings/nu1511) are reported for following packages. 

- System.Collections.Immutable
- System.Text.Json
- System.Formats.Asn1

Currently there is no need to use latest NuGet packages.
So I've removed these references from projects.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -15,6 +15,11 @@ jobs:
     permissions:
       packages: write
     steps:
+    - uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: |
+          10.x
+
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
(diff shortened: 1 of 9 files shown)


yufeih (approved):

yufeih:
Thank you @filzrev

Issue #10552, opened 2025-02-21
Reflecting the Experimental attribute


kordys:
**Is your feature request related to a problem? Please describe.**
We started using the [experimental attribute](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-12.0/experimental-attribute) in our code base lately. Members decorated by this attribute are not identified in the API documentation by any sign.

**Describe the solution you'd like**
It would be great if the generator could detect members decorated by this attribute and mark such members somehow in the generated API documentation including the message provided in the attribute.

filzrev:
On latest version of docfx (2.78.4).
`ExperimentalAttribute` is displayed as `Preview` badge when using `outputFormat: ApiPage`.

**Example Page**:  
  https://dotnet.github.io/docfx/seed/apipage/BuildFromProject.Class1.IIssue8948.html

Currently, `ExperimentalAttribute` is ignored when using `mref` or `markdown` outputFormat.

PR #10566, opened 2025-02-26
fix: Include [Experimental] attribute support by default

filzrev:
This PR intended to fix issue #10552

`outputFormat: ApiPage` supports `ExperimentalAttribute` (It's identified by name)
And render HTML with `Preview` badge with alert text.
Example: https://dotnet.github.io/docfx/seed/apipage/BuildFromProject.Class1.html

But when using `System.Diagnostics.CodeAnalysis.ExperimentalAttribute` attribute.
It's filtered out by default filterconfig settings.
Because when `namespace` is excluded. all types inside namespace are also excluded.

This PR modify `defaultfilterconfig.yml` rules and enable ExperimentalAttribute by default.


yufeih (approved):
Thank you @filzrev

Issue #9207, opened 2023-09-17
[Feature Request] Reduce code analyzer's suggestion messages.


filzrev:
**Is your feature request related to a problem? Please describe.**
When building docfx solution.
Currently about 2918 messages are shown on Visual Studio `Error List` window (When enable Build + Intellisense).

I want to reduce this `suggestion messages` because a lot of unneeded suggestions are reported.
And want to change some useful suggestion message level to warning. (e.g, performance related rules.)

**Describe the solution you'd like**
It needs following tasks to reduce suggestion messages.
- Fix code by refactoring
- Suppress message by code comment
- Suppress unrequired message by changing log level with `.editorconfig` settings.

I'll send PR to fix this issue.

PR #10578, opened 2025-03-08
chore: Reduce suggestion analyzer message

filzrev:
This PR intended to reduce analyzer suggestion messages (It's tracked by #9207)

**Suggestion messages that suppressed on solution level**
Following suggestion messages raise a lot of message.
But it's not useful. So these messages are suppressed on solution level.  
- IDE0058: Remove unnecessary expression value
- IDE0130: Namespace does not match folder structure

**Suggestion messages that suppressed on test project level**
Following **performance related ** suggestion messages are suppressed on test projects.
- CA1816: Call GC.SuppressFinalize correctly
- CA1861: Avoid constant arrays as arguments
- CA1869: Cache and reuse 'JsonSerializerOptions' instances


yufeih (approved):
Thank you @filzrev

Issue #10332, opened 2024-10-30
[Bug] Generic methods / types are not ordered correctly together with their non-generic versions


mikernet:
**Describe the bug**
Title says it all. Examples:

![Image](https://github.com/user-attachments/assets/8040782d-2654-4b39-987d-ff5fb0fe7021)

![Image](https://github.com/user-attachments/assets/fdf0de37-63a3-4c66-ab3f-8dba72043e30)

**Expected behavior**
Generic types/methods should follow immediately after non-generic ones with the same name.

**Context (please complete the following information):**
- OS: Windows
- Docfx version: 2.77.0
- .NET version: .net 8

filzrev:
It seems to be occurred when using `"outputFormat": "mref"` options.
When using `apiPage` or `markdown` options. It works as expected.

I though it's introduced by enabling `InvariantGlobalization` mode. (#8893)
I'll try to create PR to fix reported issue later.

**Background**
Currently toc items are sorted by `uid`.
And when item contains `generic type parameter` it's represented as "`1". 

But when toc items are sorted with `Ordinal` order.
The order of results is changed from what is expected.

**Sample TOC file**
```yaml
  - uid: BuildFromProject.IRoutedView
    name: IRoutedView
  - uid: BuildFromProject.IRoutedViewModel
    name: IRoutedViewModel
  - uid: BuildFromProject.IRoutedView`1
    name: IRoutedView<T>
```

mikernet:
So based on your comment, does that mean that there is a workaround that I can use for now (something that won't radically change how I need to generate the docs)?

filzrev:
A temporary workaround is to disable [globalization nvariant mode](https://learn.microsoft.com/en-us/dotnet/core/runtime-config/globalization#invariant-mode).

**1. Steps to disable`InvariantMode`settings**
1.  Open following file. (On Windows environment)
    > `%UserProfile%\.dotnet\tools\.store\docfx\2.77.0\docfx\2.77.0\tools\net8.0\any\docfx.runtimeconfig.json` 
2. Modify `System.Globalization.Invariant` setting to `false`.

**2. Steps to confirm behavior changes**
1. Run `docfx metadata` command.
2. Run `docfx build --serve` command
3. Confirm TOC  displayed orders.

---
To fix the root cause problems.
It need to create `Custom IComparer<string>` that compatible with `InvariantCulture`(with `System.Globalization.Invariant: false`)

PR #10585, opened 2025-03-11
fix: TOC items order issue

filzrev:
This PR intended to fix issue #10332

**Background**
When using `Globalization Invariant Mode`. string is sorted with `ordinal` orders.
And it will cause some order problems, which seems not to be natural orders. 

**What's changed in this PR**

#### 1. Add `SymbolStringComparer.cs`
Add custom StringComparer that sort ASCII chars.
It's similar to `StringComparer.InvariantCulture`'s order but have following differences.
- Non ASCII chars are sorted with ordinal order.
- Non-printable chars are compared. (built-in string comparer seems ignoring these chars)
- Char orders of ',' and  ')' are inverted. (To handle .NET overloaded method parameter orders)

#### 2. Modify `YamlViewModelExtensions.cs`
Modify toc /enum items sort logics to use custom `SymbolStringComparer`.

#### 3. Add test class to that contains test for Issue10332
Confirmed generated web site symbols are ordered as expected.
And `enumSortOrder:alphabetic` behavior is confirmed by modifying `seed/docfx.json` manually.

Additional apply formatter to `Class1.cs` file content.


test/Docfx.Dotnet.Tests/Docfx.Dotnet.Tests.csproj
```diff
@@ -1,4 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <!-- Use following setting when running tests withoug Globalization Invariant Mode. -->
```
yufeih, line 4:
```suggestion
    <!-- Use following setting when running tests without Globalization Invariant Mode. -->
```

src/Docfx.Dotnet/Helpers/SymbolStringComparer.cs
```diff
@@ -0,0 +1,216 @@
...
+
+/// <summary>
+/// StringComparer that simulate <see cref="StringComparer.InvariantCulture"> behavior for ASCII chars.
+/// </summary>
+/// <remarks>
+/// .NET StringComparer ignores non-printable chars on string comparison
+/// (e.g. StringComparer.InvariantCulture.Compare("\x0000 ZZZ \x0000"," ZZZ ")) returns 0).
+/// This feature is not implement by this comparer.
+/// </remarks>
+internal sealed class SymbolStringComparer : IComparer<string>
```
yufeih, line 16: >It seems to be occurred when using "outputFormat": "mref" options.
When using apiPage or markdown options. It works as expected.

The apiPage output does not use a custom string comparer and seems to [sort by `name` ](https://github.com/dotnet/docfx/blob/main/src/Docfx.Dotnet/DotnetApiCatalog.Toc.cs#L216) directly, does it work to sort by name also for `mref` output?

filzrev: ApiPage resove name with `SymbolFormatter.GetName(symbol, SyntaxLanguage.CSharp)`
So name is represented as followings.

```
Class1
Class1.IRoutedView
Class1.IRoutedView<T>
Class1.IRoutedViewModel
Class1.IRoutedViewModelBase
```
So generics are sorted as expected when it's sorted by ordinal .
Because `<` char code is smaller than alphabet char.

On the other hand, when using `Mref` output format.
name is represented as followings. 
```
BuildFromProject.Class1
BuildFromProject.Class1.IRoutedView
BuildFromProject.Class1.IRoutedViewModel
BuildFromProject.Class1.IRoutedViewModelBase
BuildFromProject.Class1.IRoutedView`1
```

So it requires SymbolStringComparer to get expected results.
I thought SymbolStringComparer should be also applied to ApiPage/Markdown output format.

src/Docfx.Dotnet/Helpers/SymbolStringComparer.cs
```diff
@@ -0,0 +1,216 @@
...
+                // Gets custom char order
+                var xOrder = AsciiCharSortOrders[xChar];
+                var yOrder = AsciiCharSortOrders[yChar];
+
+                var result = xOrder.CompareTo(yOrder);
+                if (result == 0)
+                    continue;
+
+                // Custom order logics for method parameters.
+                if ((xChar == ',' && yChar == ')') || (xChar == ')' && yChar == ','))
```
vicancy: For 
```cs
public class A { public class B { } }
public class B { }

public class C
{
    public void M(A a) { }
    public void M(A a, B b) { }
    public void M(A.B ab) { }
}
```

what is the order for `M(A) → M(A,B) → M(A.B)`?

filzrev: Since the **comma** comes before the **period** in ASCII code, the behavior should be the same.
`M(A)` → `M(A,B)` → `M(A.B)`

**TestCode**
```csharp
	string[] values = ["M(A)", "M(A,B)", "M(A.B)"];
	values.OrderBy(x=>x, StringComparer.CurrentCulture).Dump();
	values.OrderBy(x=>x, StringComparer.Ordinal).Dump();
```

vicancy:
```cs
var comparer = SymbolStringComparer.Instance;

comparer.Compare("M(A)", "M(A,B)");   // -9
comparer.Compare("M(A,B)", "M(A.B)"); // -5
comparer.Compare("M(A.B)", "M(A)");   // -4

```
so `M(A) < M(A,B) < M(A.B) < M(A)`?

PR #10723, opened 2025-06-20
Fix spelling error

bart-vmware:


yufeih:
Thank you @bart-vmware

PR #10827, opened 2025-09-27
fix: handle NRE

bdovaz:
`basePath` can be null, and if it is, it causes the `Path.Combine` method to throw an exception.

https://learn.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=net-9.0

> Exceptions
> [ArgumentNullException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentnullexception?view=net-9.0)
path1 or path2 is null.


yufeih (approved):
Thank you @bdovaz

Issue #10850, opened 2025-10-17
[Bug] Nonsense links are rendered when source generators are used


bkoelman:
**Describe the bug**
When a public class internally uses the logger source generator, nonsense links are rendered in the output html of API docs.

**To Reproduce**
1. Create a class library that contains a public type, whose implementation uses the .NET logger source generator
2. Run docfx to generate the website

**Expected behavior**
Correct links are rendered, or the links are omitted entirely (which is what earlier docfx versions did).

**Actual behavior**
Nonsense links are rendered, for example:
```html
<a class="header-action link-secondary" title="View source" href="https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/dev/src/JsonApiDotNetCore/obj/Release/net8.0/Microsoft.Extensions.Logging.Generators/Microsoft.Extensions.Logging.Generators.LoggerMessageGenerator/LoggerMessage.g.cs/#L6"><i class="bi bi-code-slash"></i></a>
```

**Context (please complete the following information):**
- OS: Windows
- Docfx version: v2.79.0
- .NET version: .NET 9

bkoelman:
What could help here are some heuristics when the source location for a type is defined in multiple files, in order to choose the proper one:
- Skip over file names that end in `.g.cs`
- Skip over anything with `/obj/` in the path
- Skip over files that contain the well-known `// <autogenerated />` header

Alternatively, use the same rules Roslyn uses to determine whether a file is auto-generated.

filzrev:
I've also confirmed reported issue can be reproduced with latest `main` code base.

It seems there is behavior changes between Roslyn `4.12.0` -> `4.14.0`.

When using Roslyn `4.12.0`.
[syntaxNode.SyntaxTree.FilePath](https://github.com/dotnet/docfx/blob/364f5af2ca3c4a749924215658560098fc08d398/src/Docfx.Dotnet/ManagedReference/Visitors/VisitorHelper.cs#L165) property return **relative** path.
So [it fails to compare path with git base path and `null` value is set to `source.Remote`](https://github.com/dotnet/docfx/blob/364f5af2ca3c4a749924215658560098fc08d398/src/Docfx.Dotnet/ManagedReference/Visitors/VisitorHelper.cs#L169).

When Roslyn updated to `4.14.0`, it return **absolute** path for generated files.
So it's handled same as normal file and resolved URL is outputted.

PR #10854, opened 2025-10-21
chore: Add logics to avoid generating invalid link

filzrev:
This PR intended to fix issue reported by #10850.

**What's changed in this PR**
Modify `TryGetFileDetail` to return `null` when input file path contains `/obj/` path.

It seems Roslyn API don't expose public function to get flag that current `SyntaxNode` is generated from Source Generator or not.
So this PR simply checking  path has `/obj/` or not.

**Other Changes**
Update snapshot files with `snapshot.yml` workflow.
This PR affects existing page contents that are generated by `outputFormat: markdown`.
And invalid `Edit this page` links are removed.

**Background**
Mref/ApiPage link is resolved when running `docfx metadata` command.

In contrast, Markdown output format generate **plain** markdown on `docfx metadata` command.
And `Edit this page` link is resolved on `docfx build` command [based on markdown file path](https://github.com/dotnet/docfx/blob/053e041f31774a7a078a8aceb01ea23809d62dda/src/Docfx.Build.Common/MarkdownReader.cs#L27-L30).
So current existing markdown pages contains invalid links that can't accessible.

**Example URLs**
- `Edit this page` link: https://dotnet.github.io/docfx/seed/md/BuildFromAssembly.html


filzrev:
Current snapshot test don't contains files that are created by Source Generator.
So it can't detect regression on existing tests.

I'll expecting to add required test files on another PR after this PR merged.

Issue #10553, opened 2025-02-21
[Regression] Links are no longer clickable after update to 2.78.3


bkoelman:
**Describe the bug**
Links in v2.78.2 were rendered as clickable links, which is no longer the case after updating to v2.78.3.

Before:

![Image](https://github.com/user-attachments/assets/e42b43ce-fd18-43d4-87e4-d6cc72339d34)

After:

![Image](https://github.com/user-attachments/assets/b59193b4-1b08-4c39-8488-4c127949291e)

Source code this was produced from:
```c#
using JetBrains.Annotations;
using Microsoft.AspNetCore.Mvc.Filters;

namespace JsonApiDotNetCore.Middleware;

/// <summary>
/// Converts action result without parameters into action result with null parameter.
/// <example>
/// <code><![CDATA[
/// return NotFound() -> return NotFound(null)
/// ]]></code>
/// </example>
/// This ensures our formatter is invoked, where we'll build a JSON:API compliant response. For details, see:
/// https://github.com/dotnet/aspnetcore/issues/16969
/// </summary>
[PublicAPI]
public interface IAsyncConvertEmptyActionResultFilter : IAsyncAlwaysRunResultFilter;
```

**To Reproduce**
1. Clone the repo at https://github.com/json-api-dotnet/JsonApiDotNetCore
2. `cd docs`
3. `.\build-dev.ps1 -NoBuild`
4. Browser opens, navigate to http://localhost:8080/api/JsonApiDotNetCore.Middleware.IAsyncConvertEmptyActionResultFilter.html
5. `dotnet tool update docfx`
6. Continue from step 3 and observe change in browser

**Expected behavior**
No change in clickable links.

**Context (please complete the following information):**
- OS: Windows
- Docfx version: 2.78.3
- .NET version: .NET 9 SDK v9.0.103, runtimes: 8.0.13, 9.0.2

- `docfx.json` config

```json
{
  "metadata": [
    {
      "properties": {
        "ProduceReferenceAssembly": "true"
      },
      "src": [
        {
          "files": [
            "**/JsonApiDotNetCore.csproj",
            "**/JsonApiDotNetCore.Annotations.csproj"
          ],
          "src": "../"
        }
      ],
      "dest": "api",
      "disableGitFeatures": false
    }
  ],
  "build": {
    "content": [
      {
        "files": [
          "api/**.yml",
          "api/index.md",
          "getting-started/**.md",
          "getting-started/**/toc.yml",
          "usage/**.md",
[...]

PR #10855, opened 2025-10-23
chore: Modify logics converting xml doc comment to markdown

filzrev:
This PR intended to fix issue #10553.

### Background
Markdown can mix `HTML Block` inside markdown content.
When mix `HTML Block` inside markdown.
It need to add `Empty Line` between `HTML Element` and markdown content to parse HTML Block correctly. 
([CommonMark Spec](https://spec.commonmark.org/0.31.2/#html-blocks)

#### 1. `XmlComment.Extensions.cs`
Add **partial class** that definition of `GetMarkdownText` method and related helper methods/fields.
 `GetMarkdownText` method returns markdown that insert EmptyLine before/after HTMLBlock when required.

#### 2. `XmlComment.cs`

- Replace `XPathNavigator` to use `XDocument` instead. 
- Remove temporary workaround logics to insert new line before `<pre>` and preserving indent.   
  (These logics are moved to `GetMarkdownText` method)
- Cleanup existing code.

#### 3. `XmlCommentTest.*.cs`
Add test patterns for XML->Markdown conversion. 

### Update Snapshots
Snapshot files are updated by `snapshot.yml` and following changes applied.

- Extra `\n` is added at `description` content. ([It's caused by new line is inserted before [`<para>` tag](https://github.com/dotnet/docfx/blob/053e041f31774a7a078a8aceb01ea23809d62dda/samples/seed/dotnet/solution/CatLibrary/Class1.cs#L329-L331) 

- Empty line inserted before `<pre>`,`<ul>`,`<table>`,`<p>` tag when previous node is markdown content.  
   (If prev node is HTML element.  empty line is not added)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         os: [windows-latest, macos-latest, ubuntu-latest]
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
       with:
         ref: ${{github.event.pull_request.head.ref}}
         repository: ${{ github.event.pull_request.head.repo.full_name }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,7 @@ jobs:
     if: github.event.pull_request.draft == false   
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - uses: actions/setup-dotnet@v5
         with:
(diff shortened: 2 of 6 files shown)


yufeih (approved):

filzrev:
It seems `dotnet format  --verify-no-changes` command failed on Ubuntu environment.
So I've switched lint runner to Windows as temporary workaround.

It’ll be reverted after following `dotnet format` issue is resolved.
`https://github.com/dotnet/sdk/issues/52249`

PR #10892, opened 2025-11-11
fix: Navigation layout issue on mobile display size

### LINKED ISSUE #10835: [Bug] Navbar menu on mobile incorrectly positioned
Use a mobile device, open the navbar menu.

If a doc site uses a longer app title, then it pushes the menu further to the right and may overflow the page. Ideally the menu should open under the title and be full width.

filzrev:
This PR intended to fix #10835.

When using default `$enable-important-utilities: true` setting that break layout on mobile display size.
So it need revert `$enable-important-utilities: false` setting.

### DIFF
--- a/templates/modern/src/docfx.scss
+++ b/templates/modern/src/docfx.scss
@@ -5,6 +5,7 @@
 
 @use "mixins";
 @use "bootstrap/scss/bootstrap" with (
+  $enable-important-utilities: false,
   $container-max-widths: (
     xxl: 1768px
   )


xPaw:
Why not remove flex-nowrap from the navbar?

filzrev:
> Why not remove flex-nowrap from the navbar?

There is no particular reason.
The purpose of this PR is to revert the site's behavior to that of the previous version.

If cleaning up the HTML markup is preferable, it will be addressed in another PR.

In any case, the menu button behavior in the mobile view has not been tested.
First, it might need to add Percy snapshot tests to avoid regressions.

PR #10927, opened 2025-12-14
Custom build for linq2db

MaceWindu:
- update to Roslyn 5
- partial C# 14 extensions support : no proper rendering, just make build not fails
- custom uid resolve logic to avoid conflicts

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -36,13 +36,13 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageVersion Include="Microsoft.CodeAnalysis" Version="4.14.0" />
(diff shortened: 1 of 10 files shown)


MaceWindu:
Whoops, sorry. Didn't noticed I opened it on main repo. Intended to open it on fork. It's project-specifix workaround to make it work for us with C#14 before docfx not updated yet and custom conflict resolution logic to solve existing issue.

In any case, probably conflict resolution idea could be used with some modifications to solve issue itself. Only issue I noticed - xref links from articles lead to non-prefixed page (page works, but not highligted in menu) and urls contain assembly name.

We actually trying to fix two conflict types here:
- different projects have same namespace
- same projects but targeting different version of EF - so both namespace and assembly name matches

PR #10943, opened 2025-12-22
chore: fix build failure issue when using c# 14 extension members

filzrev:
This PR add **limited** support for C# 14 Extension members syntax.

Currently it throw exception when some C# features are used.
By this PR it can use C# 14 syntax without error.

**What's changed in this PR**
- Add code to handle `TypeKind.Extension` enum value.
- Add code to **skip** `Extension Members` symbol as temporary workaround for issue described below.
- Add C# 13/14 syntax test code under `samples/csharp/src` directory.

**Background**
When using C# extension members 
`INamedTypeSymbol` is generated per extension blocks.
And following property is exposed.
- `ExtensionGroupingName` // e.g. `<G>$8B58B811E742D8E9EA7E14F878F87B0F`
- `ExtensionMarkerName`     // e.g. `<M>$2C37A6F24442AF359D03A7723186221C`
- `ExtensionParameter`:       // `IParameterSymbol` that specified at extension block.

These names are used for DocumentationCommentId also.
And it contains `<`, `>` chars so need to be escape when output YAML files. 

**Limitation**
- Extension Members are not shown on target type's extension method/property.
- `extension blocks` are not shown on `static class` extensions.

### DIFF
--- a/samples/csharp/docfx.json
+++ b/samples/csharp/docfx.json
@@ -3,20 +3,37 @@
     {
       "src": [
         "src/*.csproj"
+      ],      
+      "outputFormat": "mref",
+      "memberLayout": "separatePages",
+      "output": "obj/api"
+    },
+    {
+      "src": [
+        "src/*.csproj"
+      ],      
+      "outputFormat": "apiPage",
+      "memberLayout": "separatePages",
+      "output": "obj/apipage"
+    },
+    {
+      "src": [
+        "src/*.csproj"
       ],
-      "dest": "api"
-    }
+      "outputFormat": "markdown",
+      "memberLayout": "separatePages",
+      "output": "obj/md"
+    },
   ],
   "build": {
     "content": [
-      {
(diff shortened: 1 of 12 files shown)


mikernet:
I tried building the docs for https://github.com/Singulink/Singulink.UI/ using this branch and it no longer throws an exception during doc build but it also doesn't show the extension members.

The generated docs: https://www.singulink.com/Docs/Singulink.UI/api/Singulink.UI.Navigation.ViewModelExtensions.html

The source file: https://github.com/Singulink/Singulink.UI/blob/main/Source/Singulink.UI.Navigation/ViewModelExtensions.cs

mikernet:
I also tried it with the modern template just to make sure it wasn't something template related and it's not showing there either.

filzrev:
I've renamed PR title.
As described at description. This PR intended to fix build failure issue when C# 14 syntax used.

PR #10951, opened 2025-12-28
chore: Fix dependabot.yml

filzrev:
It seems #10910 changes are not works as expected and causing validation errors.
And it's not possibly to define multiple nuget sources with same `target-branch`/ `directory` pairs.

This PR include following changes
1. Remove section for `Verify.*` packages
2. Add ignore settings for `Verify.*` packages (It's expected to be updated manually)


vicancy:
Covered by #11150

PR #10964, opened 2026-01-14
deps: Update JsonSchema.Net dependency to 8.0.5

filzrev:
This PR update JsonSchema.Net dependency to 8.0.5.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -7,7 +7,7 @@
     <PackageVersion Include="HtmlAgilityPack" Version="1.12.4" />
     <PackageVersion Include="ICSharpCode.Decompiler" Version="9.1.0.7988" />
     <PackageVersion Include="Jint" Version="4.6.3" />
-    <PackageVersion Include="JsonSchema.Net" Version="7.4.0" />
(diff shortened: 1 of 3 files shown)


filzrev:
I've temporary  fixed JsonSchema.Net dependency version to `8.0.5`.
Because [JsonSchema.Net  planning to change `End User License Agreement` from February 2026](https://github.com/json-everything/json-everything/issues/981).

And when using newer version of packages. it might need to modify [THIRD-PARTY-NOTICES.TXT](https://github.com/dotnet/docfx/blob/main/THIRD-PARTY-NOTICES.TXT).

filzrev:
> And when using newer version of packages. it might need to modify [THIRD-PARTY-NOTICES.TXT](https://github.com/dotnet/docfx/blob/main/THIRD-PARTY-NOTICES.TXT)

As far as I've confirmed [FAQ of Open Source Maintenance Fee](https://opensourcemaintenancefee.org)
It seems no problem to distribute JsonSchema.Net binaries with MIT license as docfx dependency.

Because docfx is distributed as OSS and no maintenance fee is required. 

But license related issue is complicated. 
I would like to keep the version fixed for the time being.

vicancy (approved):

Issue #10901, opened 2025-11-15
[Bug] CS8256 triggered for CSPROJ targetting .NET 10


smaillet:
**Describe the bug**
'error CS8652: The feature 'field keyword' is currently in Preview and *unsupported*. To use Preview features, use the 'preview' language version.'

Is triggered during metadata generation for any use of the "field" keyword even though the CSPROJ file targets the .NET 10 framework or has `<LangVersion>14</LangVersion>`

**To Reproduce**
1. Add  `<LangVersion>14</LangVersion>` to the CSPROJ 
2. Add a valid use of the new C# 14 'field' keyword
Example:
``` C#
private int SomeProperty
{
    get
    {
        ObjectDisposedException.ThrowIf( value < 0, this );
        return field;
    }

    set;
}
```

**Expected behavior**
Docs are generated without errors

**Context (please complete the following information):**
- OS: Windows 11 25H2
- Docfx version: 2.78.4+ba9dfcbb96a4ceed2ac6d3ba93e12e16550b3e2a

**Additional context**
Occurs even if Global.json is set to use .NET 10
``` JSON
{
    "sdk": {
        "version": "10.0.100",
        "rollForward": "latestMinor",
        "allowPrerelease": false
    }
}
```

BadMagic100:
I hit the same thing recently, it seems that the component doing metadata generation from projects needs to update to a newer version of the Roslyn API. In the meantime I've worked around it by building from assemblies as seen in https://github.com/BadMagic100/ItemChanger.Core/commit/1e688b3ca67c263338f207b4b7969a8a4360da4f

hexxone:
I am pretty sure I just experienced a similar error?

However, the code in my case is "CS8652" and not "CS8256"

`error CS8652: The feature 'unbound generic types in nameof operator' is currently in Preview and *unsupported*. To use Preview features, use the 'preview' language version.`

What does this mean? What is the "preview" language version and how can I use it? The message is not really useful. I tried searching for several things to no avail.

I am also using this exact dotnet global.json:

```json
{
    "sdk": {
        "version": "10.0.0",
        "rollForward": "latestMajor",
        "allowPrerelease": false
    }
}
```

filzrev:
Close this issue.
Reported problem is expected to be resolved by docfx `v2.78.5`.

Issue #10840, opened 2025-10-10
[Bug] .NET Maui projects fail to build properly through docfx


LeadAssimilator:
**Describe the bug**
Maui ios binding projects fail to build properly through docfx - the binding generator never runs leading to missing files in the output. While not directly shown in the repo below, this almost always results in a build failure since most binding projects have additional code that refer to the auto generated files that were never created.

**To Reproduce**
1. create a maui library project targeting net9.0-ios
2. set `<IsBindingProject>true</IsBindingProject>`
3. using sharpie, generate and including binding files for a framework
* `<ObjcBindingApiDefinition Include="Platforms/iOS/Bindings/ApiDefinitions.cs">`
* `<ObjcBindingCoreSource Include="Platforms/iOS/Bindings/StructsAndEnums.cs">`
4. init docfx project referring to maui library project with all defaults
5. run docfx to build the project
6. observe the binding files are never generated in obj/Release/net9.0-ios/iOS for example
7. build the project manually in release
8. observe binding files are generated in obj/Release/net9.0-ios/iOS as they should be

Non-Maui ios binding projects also fail:
1. dotnet new iosbinding
2. Put the following in ApiDefinitions.cs
`[BaseType (typeof (Foundation.NSObject))]
        interface Widget {
        }`
3. Same as above 4-8

**Expected behavior**
obj/Release/net9.0-ios/iOS contains generated files

**Context (please complete the following information):**
- OS: MacOS 15.7
- Docfx version: 2.78.3
- .NET version: .NET 9.0

Is there a way to pass -bl to the build that docfx does so a binary build log can be produced? Not sure how to debug this without said log.

filzrev:
> Is there a way to pass -bl to the build that docfx does so a binary build log can be produced.

Currently it's not possible to output additional error logs.
Because  when using Roslyn 4.9.2 or later, MSBuildWorkspace API ignore custom loggers.

[Binary logger support is expected to be added](https://github.com/dotnet/roslyn/issues/72202) on next version of Roslyn(`v5.0.0`)
But it's merged a few days ago, so it's not released yet.

filzrev:
> Binary logger support is expected to be added on next version of Roslyn(v5.0.0)

This commit seems not included on Roslyn v5.0.0.
It need to wait next Roslyn releases.

Issue #10808, opened 2025-09-11
[Feature Request] Support C# 14


AArnott:
When I'm using C# 14 extension members syntax, the docfx build fails with errors suggesting that my global.json selected SDK version and Directory.Build.props selected LangVersion is not being used.

Why is this? How/when can docfx support using C# 14 and/or the SDK and langversion specified by the user?

filzrev:
docfx using Roslyn API by referencing `Microsoft.CodeAnalysis.*` packages.
Latest version of docfx using [4.14.0](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp/4.14.0) version.

C# 14  Extensions feature support seems to be added on  v5.x. (Currently preview status)
And it'll updated by PR #10859

After above PR merged.
It need to add code to handle C# 14 extensions features.

- Handle `TypeKind.Extension` and extract defined extensions. 
- Handle  `AssociatedSymbol` and extract information about extended method/property/user defined operator
- Add page rendering logics for extensions

bkoelman:
I don't know if docfx shows user-defined operators, but if it does, should also take into account the new += operator which was added in C# 14.

AArnott:
Getting a docfx release that at least doesn't totally break on C# 14 syntax would be helpful. Actually including C# 14 extension members in the docs could be a separate release, if that allows the unblocking release to come sooner.

gumbarros:
Any ETA on this? This is affecting us to migrate to .NET 10.

Marv51:
I saw that Roslyn 5.0 packages were released 2 weeks ago (e.g. https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp).

Would that allow some progress on this issue? 

Seems like @dependabot already created #10906

WeihanLi:
also https://github.com/dotnet/docfx/pull/10859

rent-a-developer:
When can a new release be expected that supports the C# 14 features?
I can't use many of the new C# features, because docfx can't handle them yet, which is very unfortunate.

**Also: Thank you all for your hard work, highly appreciated!**

TheBoxyBear:
Is there no workaround to get the metadata to generate while ignoring the problematic extension members? I tried adding ```json
          "exclude": [
            "**/Extensions.cs"
          ]
[...]

Issue #10781, opened 2025-08-13
'No instances of MSBuild could be detected - Try calling RegisterInstance or RegisterMSBuildPath to manually register one' Error


chrisbrain:
Since yesterday, a build pipeline which uses docfx 2.70.4 - run via the command line - and which has been running successfully for a couple of years started failing with:

```
No instances of MSBuild could be detected.
Try calling RegisterInstance or RegisterMSBuildPath to manually register one.
```

We are using:
```
vmImage: 'windows-2022'
```

In our Azure pipeline and I suspect this issue _might_ be related to the underlying image changing from `20250727.1.0` to `20250803.1.0` in the last couple of days.

The only thing I can spot in the diff between these versions:
https://github.com/actions/runner-images/compare/releases/win22/20250727...releases/win22/20250803

Is perhaps this:
<img width="933" height="172" alt="Image" src="https://github.com/user-attachments/assets/90bb78d3-d5b8-4fdf-a441-3595cf4a5558" />

Was wondering if you have any thoughts on this and how we could resolve it?

I am not really sure how to manually call RegisterInstance or RegisterMSBuildPath - haven't seen anything in the CLI docs.

I have tried updating to the very latest version but then I start getting a new error so am wondering if there is a more minimal fix possible here.

Thanks for any help!

chrisbrain:
I have also tried configuring to use the DLLs rather than the .vbproj files with something like this:
```
 "metadata": [
        {
            "src": [
                {
                    "files": [ 
                        "NameOfDll1.dll", 
                        "NameOfDll2.dll" 
                    ],
                    "src": "../../../Source-Code/NameOfProject/bin/release"
                }
            ],
            "dest": "reference",
            "disableGitFeatures": false,
            "disableDefaultFilter": false,
            "filter": "filterConfig.yml"
        },
        ...
```

Which I thought should prevent it from trying to do any building itself via docfx, but I get the same error.

filzrev:
> In our Azure pipeline and I suspect this issue might be related to the underlying image changing from 20250727.1.0 to 20250803.1.0 in the last couple of days.

Is seems .NET 6 SDK preinstallation is dropped from `20250803.1.0`.
So it might be resolved by manually install .NET 6 SDK.

filzrev:
Close this issue.
Because it reported issue seems to be docfx `2.70.4` specific issue. (That using Roslyn 4.7.0 and MSBuildLocator)

[...]

PR #10969, opened 2026-01-16
Fix: use MsBuild Locator rather than embed

### LINKED ISSUE #10848: DocFX bundles roslyn rather than using the version from the SDK
Related https://github.com/dotnet/docfx/issues/10808

We noticed this when running against projects targeting .NET 10.0.  They cannot use new language features, and starting with RC2 they stopped loading the source-generators built into the framework.

I found this was because docfx was using a copy of roslyn bundled inside it.  Instead it should be either a) locating the roslyn from the SDK used by the project or b) allowing the user to specify a path as an escape hatch.

dpvreony:
This is a fairly fundamental change to MsBuild + Roslyn, so appreciate might want to take some time looking at it.

This changes DocFX to use Microsoft.Build.Locator to find the latest applicable version of the .NET SDK or Visual Studio. So the TFM used for the docfx runtime will only find the latest version it can use. (for example the NET8 version complains if you try to scan a net10 project with it as it can only locate net8 etc.) Also added a ForceBuild for cases where some output is not tied to Roslyn, or DesignTime builds.

Fixes #10848 #10808 #10781 #10731 #10901 #11046 #10542

potentially sorts #10772 and the underlying error seen in #10840 (but not the actual piece about enabling binary logging, though that could be done via a Directory.Build.rsp file next to a project \ solution and the ForceBuild option)

also would give the answer "It just works as long as you use the right TFM (or latest) version of the DocFX runtime" to https://github.com/dotnet/docfx/discussions/8251

Thanks :)


Directory.Packages.props
```diff
@@ -26,6 +26,16 @@
     <PackageVersion Include="Microsoft.Build" Version="[17.11.48]" Condition="'$(TargetFramework)' == 'net8.0'" />
     <PackageVersion Include="Microsoft.Build" Version="[17.14.28]" Condition="'$(TargetFramework)' == 'net9.0'" />
     <PackageVersion Include="Microsoft.Build" Version="18.0.2"     Condition="'$(TargetFramework)' != 'net8.0' AND '$(TargetFramework)' != 'net9.0'" />
+    <PackageVersion Include="Microsoft.Build.Framework" Version="[17.11.48]" Condition="'$(TargetFramework)' == 'net8.0'" />
```
filzrev: I thought these items should be grouped by `<ItemGroup>`'s Condition per TargetFramework.

```
<ItemGroup Condition="'TargetFramework' == 'net8.0'">
  <PackageVersion Include="Microsoft.Build" Version="[17.11.48]" />
  <PackageVersion Include="Microsoft.Build.Framework" Version="[17.11.48]" />
  <PackageVersion Include="Microsoft.Build.Tasks.Core"  Version="[17.11.48]" />
  <PackageVersion Include="Microsoft.Build.Utilities.Core"  Version="[17.11.48]" />
</ItemGroup>
```

src/Docfx.Dotnet/MetadataJsonConfig.cs
```diff
@@ -265,6 +265,13 @@ internal class MetadataJsonItemConfig
     [JsonProperty("useClrTypeNames")]
     [JsonPropertyName("useClrTypeNames")]
     public bool UseClrTypeNames { get; init; }
+
+    /// <summary>
+    /// Forces building the project before metadata extraction.
+    /// </summary>
+    [JsonProperty("forceBuild")]
+    [JsonPropertyName("forceBuild")]
+    public bool ForceBuild { get; set; }
```
filzrev: Is it able to modify following files also?
- JsonSchema: https://github.com/dotnet/docfx/blob/main/schemas/docfx.schema.json#L578
- Config docs: https://github.com/dotnet/docfx/blob/main/docs/reference/docfx-json-reference.md#metadata
- CLI docs: https://github.com/dotnet/docfx/blob/main/docs/reference/docfx-cli-reference/docfx-metadata.md

filzrev:
docfx using [Microsoft.CodeAnalysis.Workspaces.MSBuild](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Workspaces.MSBuild) package for metadata extractions.
And metadata extraction is executed on separated process.

Is registered MSBuild SDK on this PR is actually used for metadata extraction?

dpvreony:
> docfx using [Microsoft.CodeAnalysis.Workspaces.MSBuild](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Workspaces.MSBuild) package for metadata extractions. And metadata extraction is executed on separated process.
> 
> Is registered MSBuild SDK on this PR is actually used for metadata extraction?

most the code remains the same for metadata, it's just the locator sets up MSBuild rather than it being included in the bin folder. the details are available in https://learn.microsoft.com/en-us/visualstudio/msbuild/find-and-use-msbuild-versions 

i've tested metadata extraction using https://github.com/dpvreony/Vetuviem with the release version of docfx it fails to "build" the projects due to source generators being involved. with this PR:

* NET8 and NET9 versions of the docfx tool complain about being unable to Target NET10
* NET10 works and produces full metadata, including for generated code.

i'll look at the other adjustments above

Directory.Packages.props
```diff
@@ -22,13 +22,29 @@
...
+    <PackageVersion Include="Microsoft.Build.Framework" Version="[17.11.48]" />
+    <PackageVersion Include="Microsoft.Build.Tasks.Core" Version="[17.11.48]" />
+    <PackageVersion Include="Microsoft.Build.Utilities.Core" Version="[17.11.48]" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'net9.0'">
+    <PackageVersion Include="Microsoft.Build" Version="[17.14.28]" />
+    <PackageVersion Include="Microsoft.Build.Framework" Version="[17.14.28]" />
+    <PackageVersion Include="Microsoft.Build.Tasks.Core" Version="[17.14.28]" />
+    <PackageVersion Include="Microsoft.Build.Utilities.Core" Version="[17.14.28]" />
```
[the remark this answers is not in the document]
dpvreony: the package references were to try and resolve the MSBL001 warning. i need to look and see if one of the CodeAnalysis packages is causing the MSBuild locator to warn about Microsoft.Build.Framework without it being in there as a reference.

A-Loot:
What is the current status of this PR?

A-Loot:
Sorry to ask again, but will this be merged anytime soon?

uo-uhbc:
@dpvreony Thank you very much for creating such a comprehensive solution for this problem.
I'm not sure, how this is working in this repository, but one required test for macos in the pipeline failed.
Maybe @yufeih waits for that to be fixed, before this PR is reviewed and approved.

dpvreony:
The macos test issue has a pending fix in #11074 it's a download test that's failing.

PR #10995, opened 2026-02-16
feat(pdf): Add pdfTocSource option to generate TOC from document head…

tsondergaard:
…ings

Add new configuration options to control PDF Table of Contents source:
- pdfTocSource: 'toc' (default) or 'headings' to extract from document content
- pdfTocHeadingDepth: maximum heading level to include (1-6, default 3)

When pdfTocSource is set to 'headings', the PDF TOC is generated by extracting h1-h6 headings from the rendered HTML pages instead of using the toc.yml structure. This is particularly useful for print-style publications using docfx for markdown-to-PDF generation, where the TOC should reflect the document's internal heading structure rather than website navigation.

Fixes #10994

### DIFF
--- a/src/Docfx.App/PdfBuilder.cs
+++ b/src/Docfx.App/PdfBuilder.cs
@@ -37,6 +37,15 @@ static class PdfBuilder
 {
     private static readonly SearchValues<char> InvalidPathChars = SearchValues.Create(Path.GetInvalidPathChars());
 
+    class HeadingInfo
+    {
+        public string Text { get; init; } = "";
+        public string Id { get; init; } = "";
+        public int Level { get; init; }
+        public Uri PageUrl { get; init; } = null!;
+        public int PageNumber { get; set; }
+    }
+
     class Outline
     {
         public string name { get; init; } = "";
@@ -51,6 +60,9 @@ static class PdfBuilder
 
         public string? pdfHeaderTemplate { get; init; }
         public string? pdfFooterTemplate { get; init; }
+
+        public string? pdfTocSource { get; init; }
+        public int pdfTocHeadingDepth { get; init; } = 3;
     }
 
     public static Task Run(BuildJsonConfig config, string configDirectory, string? outputDirectory = null, CancellationToken cancellationToken = default)
@@ -93,6 +105,7 @@ static class PdfBuilder
 
         Uri? baseUrl = null;
         var pdfPageNumbers = new ConcurrentDictionary<string, Dictionary<Outline, int>>();
+        var pdfHeadings = new ConcurrentDictionary<string, List<HeadingInfo>>();
 
         using var app = builder.Build();
         app.UseServe(outputFolder);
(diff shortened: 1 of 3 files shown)


tsondergaard:
@dotnet-policy-service agree

tsondergaard:
@yufeih, I see in the pull request history that you are the most active reviewer recently for this repo. Could you have a look at my PR or maybe point me to another suitable reviewer?

tsondergaard:
@filzrev, #10005 lists you as an active participant of this project. Could you have a look at my PR?

e212156822000:
Hi @tsondergaard , @yufeih,
Thank you for all your contribution. I just tested this feature in my local machine. It works like a charm.
I really hope this feature can get merged to the main branch. 
image

p.s. The title sequence in my screenshot is generated by vs code extension 'markdown all in one'.

src/Docfx.App/PdfBuilder.cs
```diff
@@ -379,12 +484,39 @@ static async Task CreatePdf(
...
+        // so we can determine the exact page within the merged PDF.
+        foreach (var heading in allHeadings)
+        {
+            var cleanUrl = CleanUrl(heading.PageUrl);
+            if (pagesByUrl.TryGetValue(cleanUrl, out var dests))
+            {
+                var resolved = false;
+                foreach (var (node, namedDests) in dests)
+                {
+                    if (namedDests.TryGet(heading.Id, out var dest) && dest is not null)
```
vicancy: This lookup misses headings with non-ASCII IDs because Chromium stores the destination names with URL encoding. I tried `中文` and `café` on pages 3 and 4, and the TOC showed page 2 for both, even though the links went to the right pages. Could we use the same encoded fragment that `HandleUriAction` uses here?

src/Docfx.App/PdfBuilder.cs
```diff
@@ -637,6 +788,34 @@ static HtmlTemplate TocHtmlTemplate(Uri baseUrl, Outline node, Dictionary<Outlin
...
+            var href = new UriBuilder(heading.PageUrl) { Fragment = heading.Id }.Uri;
+
+            var pageNumberHtml = heading.PageNumber > 0
+                ? Html($"<span class='spacer'></span> <span class='page-number'>{heading.PageNumber}</span>")
+                : default;
+
+            // Use data-level attribute for CSS styling of indentation
+            var item = Html($"""
+                <li data-level='{heading.Level}'>
+                  <a href='{href}'>{System.Web.HttpUtility.HtmlEncode(heading.Text)}
```
vicancy: `Html(...)` already escapes string values, so this encodes the heading twice. With a heading like `R&D <T>`, the PDF TOC literally shows `R&amp;D &lt;T&gt;`. Passing `heading.Text` directly should fix it.

src/Docfx.App/PdfBuilder.cs
```diff
@@ -247,6 +271,45 @@ IResult TocPage(string url)
...
+                    var level = int.Parse(tagName[1].ToString());
+                    var id = await element.GetAttributeAsync("id") ?? "";
+                    var text = (await element.InnerTextAsync()).Trim();
+
+                    // Skip headings without id or text
+                    if (string.IsNullOrEmpty(id) || string.IsNullOrEmpty(text))
+                        continue;
+
+                    // Clean up text (remove source link icons, etc.)
+                    var cleanText = text.Split('\n')[0].Trim();
```
vicancy: This also drops part of a heading if it contains a line break. For example, `<h2 id="example">First line<br>Second line</h2>` becomes just `First line` in the TOC. Could we keep the full text and turn line breaks into spaces instead?

vicancy (commented):
I tried this with a small PDF and found the issues below. Could you add tests for these cases too?

PR #11012, opened 2026-02-25
Remove links to non-existent pages for parent namespaces

### LINKED ISSUE #10588: [Bug] Namespace links created for non-existing parent namespaces
**Describe the bug**
For types, the generated pages contain a `Namespace` value. This consists of several links. However, this does not respect the project's `RootNamespace` anymore since version 2.65.2. Instead it creates a link for each namespace part, including non-existing parent namespaces. For instance, if my root namespace is `Foo.Bar.Baz`, the namespace contains the following links:
* `Foo` -> `Foo.html`
* `Bar` -> `Foo.Bar.html`
* `Baz` -> `Foo.Bar.Baz.html`

The problem is that `Foo.html` and `Foo.Bar.html` don't exist, leading to 404 errors.

Maybe I'm missing some configuration, but I tried several options including `globalNamespaceId` but the issue is not fixed. Downgrading to version 2.65.1 isn't possible because of #9047 and #9153.

**To Reproduce**
Minimal `docfx.json`:
```json
{
  "metadata": [
    {
      "src": [
        {
          "files": [
            "**.csproj"
          ],
          "src": "../Foo.Bar.Baz"
        }
      ],
      "dest": "api"
    }
  ],
  "build": {
    "content": [
      {
        "files": [
          "api/**.yml"
        ]
      },
      {
        "files": [
          "toc.yml"
        ]
      }
    ],
    "dest": "_site"
  }
}
```
The `Foo.Bar.Baz` project descriptor contains this:
```xml
<RootNamespace>Foo.Bar.Baz</RootNamespace>
```
Run `docfx docfx.json` from the directory where `docfx.json` is located.
This successfully generates the API pages but with the invalid links.

**Expected behavior**
* `Foo` is not a link
* `Bar` is not a link
* `Baz` is a link to `Foo.Bar.Baz.html`

**Screenshots**
Generated using version 2.65.1:
Image

Generated using the latest version:
Image

- Docfx version: 2.78.3
- Template: default

TimHess:
Fixes https://github.com/dotnet/docfx/issues/10588


bart-vmware:
This is great!

image

Would love to get this merged.
/cc @filzrev @yufeih

TimHess:
@vicancy can you take a look at this PR please?

vicancy:
Thanks for working on this! I found a regression in the new cleanup pass: **a reference UID being absent from `context.Members` does not necessarily mean its href is broken.** Constructed generic members can have valid links to their original definitions.

### Minimal counterexample

Generate `mref` metadata for:

```csharp
namespace Foo.Bar
{
    public class Base<T>
    {
        public void Method(T value) { }
        public T Value { get; set; }
    }

    public class Derived : Base<int> { }
}
```

Then include a Markdown page alongside the generated API pages in the build, containing:

```html
<xref uid="Foo.Bar.Base{System.Int32}.Method(System.Int32)" text="Constructed method"/>
<xref uid="Foo.Bar.Base{System.Int32}.Value" text="Constructed property"/>
```

On `7bfb15a`, the generated HTML is:

```html
<a class="xref" href="https://learn.microsoft.com/dotnet/api/system.int32">Constructed method</a>
<span class="xref">Constructed property</span>
```

With only the new cleanup pass disabled, both links point to their existing targets:

```text
Foo.Bar.Base-1.html#Foo_Bar_Base_1_Method__0_
Foo.Bar.Base-1.html#Foo_Bar_Base_1_Value
```

The condition at [ResolveReference.cs:77-82](https://github.com/dotnet/docfx/blob/7bfb15ac15a682ba88111dacaa6a8349cdc2060e/src/Docfx.Dotnet/ManagedReference/Resolvers/ResolveReference.cs#L77-L82) clears the member's valid href because `context.Members` contains its definition, not its constructed UID. `YamlViewModelExtensions.GetHref()` then picks the next non-null href from `NameParts`, which is the parameter's `System.Int32` URL. For the property, no href remains.

Could we restrict this cleanup to missing namespace components, or account for the original definition before removing a member link? A regression test covering these constructed-member xrefs would help.

There is also a smaller edge case in the `StartsWith("http")` check: with only `namespace httpTools.Child { public class Client { } }`, the local parent href `httpTools.html` is treated as an external URL and retained, even though that page does not exist. Checking actual HTTP/HTTPS schemes rather than the bare prefix would avoid this.

TimHess:
> Thanks for working on this! I found a regression in the new cleanup pass: **a reference UID being absent from `context.Members` does not necessarily mean its href is broken.** Constructed generic members can have valid links to their original definitions.
> 
> ### Minimal counterexample
> Generate `mref` metadata for:
> 
> ```cs
> namespace Foo.Bar
> {
>     public class Base<T>
>     {
>         public void Method(T value) { }
>         public T Value { get; set; }
>     }
> 
>     public class Derived : Base<int> { }
> }
> ```
> 
> Then include a Markdown page alongside the generated API pages in the build, containing:
> 
> ```
> <xref uid="Foo.Bar.Base{System.Int32}.Method(System.Int32)" text="Constructed method"/>
> <xref uid="Foo.Bar.Base{System.Int32}.Value" text="Constructed property"/>
> ```
> 
> On `7bfb15a`, the generated HTML is:
> 
> ```
> <a class="xref" href="https://learn.microsoft.com/dotnet/api/system.int32">Constructed method</a>
> <span class="xref">Constructed property</span>
> ```
> 
> With only the new cleanup pass disabled, both links point to their existing targets:
> 
> ```
> Foo.Bar.Base-1.html#Foo_Bar_Base_1_Method__0_
> Foo.Bar.Base-1.html#Foo_Bar_Base_1_Value
> ```
> 
> The condition at [ResolveReference.cs:77-82](https://github.com/dotnet/docfx/blob/7bfb15ac15a682ba88111dacaa6a8349cdc2060e/src/Docfx.Dotnet/ManagedReference/Resolvers/ResolveReference.cs#L77-L82) clears the member's valid href because `context.Members` contains its definition, not its constructed UID. `YamlViewModelExtensions.GetHref()` then picks the next non-null href from `NameParts`, which is the parameter's `System.Int32` URL. For the property, no href remains.
> 
> Could we restrict this cleanup to missing namespace components, or account for the original definition before removing a member link? A regression test covering these constructed-member xrefs would help.
> 
> There is also a smaller edge case in the `StartsWith("http")` check: with only `namespace httpTools.Child { public class Client { } }`, the local parent href `httpTools.html` is treated as an external URL and retained, even though that page does not exist. Checking actual HTTP/HTTPS schemes rather than the bare prefix would avoid this.

Thanks for the quick review, those look like good catches. I pushed fixes: constructed generic members now normalize to their original definition, and the http-prefix check is swapped for the existing `IsExternalPath` flag. Also added a regression test based on your `Base<T>`/`Derived` repro.

PR #11141, opened 2026-09-17
feat: add configurable View Source exclusions

### LINKED ISSUE #11138: [Bug] Source links for generated PDB documents under obj/ still point to 404
**Describe the bug**
When generating API documentation from compiled assemblies, source-generated members whose PDB document is under `obj/` can still get a "View Source" link pointing to a non-existent repository file.
This happens on DocFX 2.78.5 even though #10854 added a guard against generating invalid links for files under `obj/`.
In this case, the PDB/source metadata itself is valid and `sourcelink test` passes, but DocFX still emits a GitHub link to the generated `.g.cs` file under `obj/`, which returns 404.
**To Reproduce**
1. Build a project that exposes public members generated by a Roslyn source generator and produces a portable PDB with Source Link enabled.
   I reproduced this with NetCord and the generated method:
```c#
NetCord.Rest.MessageProperties.WithContent(string?)
```

2. Generate API documentation from the compiled assembly/PDB with DocFX 2.78.5.
3. Inspect the PDB documents:
`/NetCord/Rest/MessageProperties.cs` has a valid Source Link URL: https://raw.githubusercontent.com/jedrek0429/NetCord/43e938735f37050dcbfbedab5edbf35224d63ac9/NetCord/Rest/MessageProperties.cs, while the generated document `NetCord/obj/Release/net10.0/MethodsForPropertiesGenerator/MethodsForPropertiesGenerator.MethodsForPropertiesGenerator/NetCord.Rest.MessageProperties.g.cs` has no Source Link URL.
4. Verify that the Source Link tests pass.
5. Open the generated documentation for MessageProperties.WithContent.
    DocFX generates this “View Source” URL: https://github.com/jedrek0429/NetCord/blob/43e938735f37050dcbfbedab5edbf35224d63ac9/NetCord/obj/Release/net10.0/MethodsForPropertiesGenerator/MethodsForPropertiesGenerator.MethodsForPropertiesGenerator/NetCord.Rest.MessageProperties.g.cs, which returns 404.

**Expected behavior**

DocFX should not generate a repository “View Source” link for a generated PDB document under obj/ when that document has no valid Source Link URL.

For this example:

* MessageProperties.Content should continue linking to the handwritten MessageProperties.cs
* MessageProperties.WithContent should have no source link rather than a broken 404 link

**Context**

* OS: macOS
* Docfx version: 2.78.5
* .NET version: 10

**Additional context**

This appears related to #10854.

The existing /obj/ guard is in GitUtility.TryGetFileDetail, which covers source paths resolved directly from syntax trees.

When documentation is generated from compiled assemblies, VisitorHelper.GetSourceDetail can instead use `SymbolUrlResolver.GetPdbSourceLinkUrl(compilation, symbol)`

That path appears to bypass the existing /obj/ protection.

A possible fix would be to apply the same exclusion when resolving PDB document names in SymbolUrlResolver.SourceLink.cs, although there may be a better way to detect that a PDB document has no repository-backed source.

vicancy:
Fixes #11138.

Adds `metadata.sourceLinkExclude` to hide **View Source** links matching explicit source-path rules, without removing API documentation.

```json
{
  "metadata": [{
    "src": [{ "files": ["**/bin/Release/**/MyLibrary.dll"] }],
    "dest": "api",
    "sourceLinkExclude": ["**/ExampleGenerator/**"]
  }]
}
```

Rules match compiler source paths or document paths recorded in portable PDBs—not URLs, local directory searches, or paths relative to `docfx.json`. The default is `[]`, preserving existing behavior. API references and relative code includes remain unchanged.

The automatic generated-code flag and detector have been removed. This new setting only applies explicit path rules; it does not inspect attributes or comments to classify generated code.

Includes generic documentation examples and real source-generator/portable-PDB regressions. All 249 local metadata component tests pass, with no skips; the Release build succeeds.


docs/docs/dotnet-api-docs.md
```diff
@@ -64,6 +64,76 @@ Docfx examines the assembly and tries to load the reference assemblies from with
...
+All these APIs remain in the generated documentation. In particular, the handwritten `Content` property keeps its link, while a generated `WithContent` method in the excluded document does not get one. Windows PDB paths using `\` match the same `/`-separated patterns.
+
+#### Exclude one document
+
+To keep links for the other documents from that generator, replace `sourceLinkExclude` in the example above with this more specific setting:
+
+```json
+{
+  "sourceLinkExclude": [
+    "**/NetCord.Rest.MessageProperties.g.cs"
```
vwxyzh: maybe we can check [GeneratorAttribute](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.generatorattribute?view=roslyn-dotnet-5.9.0) first

vicancy: Added `metadata.excludeGeneratedSourceLinks` (default `false`) so generated-output markers can be checked before falling back to explicit path rules.

`GeneratorAttribute` marks the generator implementation rather than the API it generates, so this checks `System.CodeDom.Compiler.GeneratedCodeAttribute` on the output and leading C#/VB `<auto-generated>` comments instead. For DLL inputs, it reads the output attributes and, when available, the source embedded in the portable PDB. NetCord's generator at the reported commit emits the comment header but does not add `GeneratedCodeAttribute`.

The option is intentionally conservative: no inference from `obj`, `.g.cs`, `CompilerGeneratedAttribute`, or merely having embedded source. Unknown documents retain their links, and `sourceLinkExclude` remains available for them. Source partial declarations are handled separately; a compiled type-level attribute is propagated to its members only when the type has a single PDB document, since merged metadata cannot identify which partial declaration carried the attribute otherwise. Enabling the option can hide an otherwise valid link for marked generated code, which is why it is opt-in.

Added real generator/DLL/PDB and C#/VB field/event regressions, off/on/off configuration checks, handwritten and partial-type controls, and relative code-include coverage. After updating the branch from main, all 306 local component tests pass, and Windows/macOS/Ubuntu CI (including the actual package-distribution gate) plus lint pass on `78d89dfa85e159937d2a551de36e54b759685788`.

vicancy: Update: I've narrowed this PR to `sourceLinkExclude` path rules and removed the automatic flag and detector described in my earlier reply. `GeneratorAttribute` identifies the generator, while output markers are optional and do not establish that a repository link is unavailable. Explicit rules keep the behavior predictable and unchanged by default; automatic classification can be considered separately.

vwxyzh (approved):

Issue #11147, opened 2026-09-18
[Bug] Markdown Table With one Column missing alignment fails to convert


iJungleboy:
**Describe the bug**

In the latest release (just a few hours ago) something broke converting tables.
In the following example, the first table converts as expected the second does not

```markdown
This works:

| Field       | PersonShared  | Person      |
| ---         | ---           | ---         |
| Name        | 🔊 Master     | 🔗 Inherit  |

While this fails - note the missing "---" under the second column "PersonShared"

| Field       | PersonShared  | Person      |
| ---         |               | ---         |
| Name        | 🔊 Master     | 🔗 Inherit  |
```

The first table converts as expected, the second results as the markdown being added to the page:

```html
<p>While this fails - note the missing &quot;---&quot; under the second column &quot;PersonShared&quot;</p>
<p>| Field       | PersonShared  | Person      |
| ---         |               | ---         |
| Name        | 🔊 Master     | 🔗 Inherit  |</p>
```

**Expected behavior**

Both should convert - this did work before.

**Context (please complete the following information):**

- OS: Windows
- Docfx version: 2.78.6

**Additional context**

I'm quite certain this is a regression from #11088

vicancy:
Thanks for reporting this. We traced the change to the Markdig update from 1.0.0 to 1.1.0 in Docfx 2.78.6. The relevant upstream change is [Markdig's separator-validation fix](https://github.com/xoofx/markdig/commit/7959e3b9129f9ad5e7b5b4c2bad149060381c01c).

The updated parser follows the [GFM table specification](https://github.github.com/gfm/#tables-extension-): each separator cell needs hyphens (`-`), while the colons used to specify alignment are optional. Earlier versions accepted empty separator cells.

We recommend updating the affected tables by adding `---` to those empty cells:

```markdown
| Field | PersonShared | Person |
| --- | --- | --- |
| Name | 🔊 Master | 🔗 Inherit |
```

We appreciate that this changes behavior for existing documents. Would updating the affected tables be practical for your documentation? If you have concerns or constraints that would make this difficult, please share them so we can discuss further.

iJungleboy:
@vicancy I already fixed my docs, so it doesn't matter.

But I'm pretty sure that others may not expect this, and I'm almost certain that this is a mistake.

[...]

PR #11160, opened 2026-09-21
Bump Markdig to 1.4.0

vicancy:
Upgrade Markdig from 1.3.2 to 1.4.0 to fix partially empty pipe-table separators. Add regression coverage while preserving Docfx's existing table parsing behavior.


vicancy:
Fixes https://github.com/dotnet/docfx/issues/11147

PR #11163, opened 2026-09-21
fix: resolve local namespace links through xrefs

### LINKED ISSUE #10588: [Bug] Namespace links created for non-existing parent namespaces

vicancy:
Fixes #10588.

Thanks @TimHess for #11012 and the tests. This keeps those test cases but tries a different fix for the namespace link regression from #8588.

### Differences

#11012 generates URLs first, then clears links whose UIDs are missing from the metadata.

This PR leaves local namespace `href`s unset. `docfx build` resolves them instead. This only changes `mref`; `apiPage` and Markdown keep their current behavior.

### Trade-offs

#11012 preserves existing `href`s where possible. But the cleanup can also remove valid links, like `Outer<int>.Inner`. The new `ConstructedNestedTypesKeepUrls` test covers this case.

This PR avoids that cleanup. But namespace `href`s won’t be available right after metadata generation. I’m not sure if any workflows depend on them at that stage.

@TimHess, what do you think? Any cases we might be missing?


TimHess:
I agree with prevention over cleanup. Running your `Outer<int>.Inner` case against our branch produces the same failure as the method/property one, just a different symbol kind.

Nothing concrete to add on anything missing, but one half-formed thought: your tests are all single-build, and this is the first time local namespace links go through the same xref resolution as external ones. Should it be tested against an external xrefmap.yml (separate site referencing this one)?

I'm fine with closing #11012 in favor of this approach

